PR #48, opened 2019-10-23
Add cassandra state store

### LINKED ISSUE #45: Cassandra state store
## Describe the proposal

Implement Cassandra as a state store and transactional state store.

yaron2:
closes #45


state/cassandra/cassandra.go
```diff
@@ -0,0 +1,304 @@
...
+
+// Set saves state into cassandra
+func (c *Cassandra) Set(req *state.SetRequest) error {
+	var bt []byte
+	b, ok := req.Value.([]byte)
+	if ok {
+		bt = b
+	} else {
+		bt, _ = jsoniter.ConfigFastest.Marshal(req.Value)
+	}
```
youngbupark: I think we used this logic multiple times. Wouldn't it be good to have utils package in component-contrib ?

yaron2: Yes, but lets save that for a separate refactor

state/cassandra/cassandra.go
```diff
@@ -0,0 +1,304 @@
...
+	return nil
+}
+
+// Get retrieves state from cassandra with a key
+func (c *Cassandra) Get(req *state.GetRequest) (*state.GetResponse, error) {
+	query := fmt.Sprintf("SELECT value FROM %s WHERE key='%s'", c.table, req.Key)
+
+	session := c.session
+
+	if req.Options.Consistency == state.Strong {
```
youngbupark: Isn't there any chance that user passes unsupported consistency ? e.g. typo. evntual

state/cassandra/cassandra.go
```diff
@@ -0,0 +1,304 @@
...
+		sess, err := c.createSession(gocql.Any)
+		if err != nil {
+			return err
+		}
+		defer sess.Close()
+		session = sess
+	}
+
+	err := session.Query(query, req.Key, bt).Exec()
+	return err
```
youngbupark: I would just return session.Query(x) without the assignment.
```
return session.Query(query, req.Key, bt).Exec()
```

state/cassandra/cassandra.go
yaron2: Yes. And then the default consistency level is used

state/cassandra/cassandra.go
yaron2: Alright...

yaron2: Fixed with commit

Issue #58, opened 2019-10-26
No CI pipeline (either azure or github actions)


sayboras:
Not sure if any plan to do so for this repo.

sayboras:
As other repo dapr/dapr just migrated from azure pipeline to github action, I will proceed with github action then, please help to review #59

youngbupark:
@sayboras Can you please add badge to README.md ? I think you can find the reference in dapr repo.

sayboras:
@youngbupark Good point 💯

sayboras:
All items are addressed, closing PR :)

PR #59, opened 2019-10-26
Add github action for CI

sayboras:
# Description
To add github action for CI. This is the first version of CI, so I am not sure what is required in term of releasing process. 
1. Running `go test` as metioned in #5 
2. Running `golangci-lint run` for linter check. As the first cut, I used the same `.golangci.yaml` as https://github.com/dapr/dapr/blob/master/.golangci.yml 

## Issue reference
Related issue #58  #5 

## Checklist
Not sure why the github action didn't run for this repo (maybe I don't have admin/write permission in this repo), please refer to output here https://github.com/sayboras/components-contrib/pull/1

* [x] Code compiles correctly
* [ ] Created/updated tests


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/dapr/components-contrib
 
-go 1.12
+go 1.13
```
sayboras: Upgrade to 1.13 as per dapr/dapr repo, we can keep it as 1.12 if you guys are having any concern about compatibility.

state/state_store.go
```diff
@@ -5,8 +5,8 @@
 
 package state
 
-// StateStore is an interface to perform operations on store
-type StateStore interface {
+// Store is an interface to perform operations on store
+type Store interface {
```
sayboras: rename the interface as package is already `state` (e.g. to avoid state.StateStore from consumer point of view)

state/retry_test.go
```diff
@@ -1,89 +1,88 @@
...
-					Threshold: 3,
-					Pattern:   "linear",
-				},
-			},
-		})
-		assert.Equal(t, 3, counter, "should execute 3 times")
-	})
-
-}
+// ------------------------------------------------------------
```
sayboras: just carriage character.

state/retry.go
```diff
@@ -1,98 +1,98 @@
...
-				}
-			}
-			return fmt.Errorf("failed to set value after %d retries", req.Options.RetryPolicy.Threshold)
-		}
-		return method(req)
-	default:
-		return fmt.Errorf("unrecognized retry patter '%s'", req.Options.RetryPolicy.Pattern)
-	}
-}
+// ------------------------------------------------------------
```
sayboras: just carriage character.

state/redis/redis_test.go
```diff
@@ -1,86 +1,86 @@
...
-		slaves := store.parseConnectedSlaves("# Replication\r\nrole:master\r\nconnected_slaves:2\r\n")
-		assert.Equal(t, 2, slaves, "connected slaves must be 2")
-	})
-
-	t.Run("connectedSlaves is 1", func(t *testing.T) {
-		slaves := store.parseConnectedSlaves("# Replication\r\nrole:master\r\nconnected_slaves:1")
-		assert.Equal(t, 1, slaves, "connected slaves must be 1")
-	})
-}
+// ------------------------------------------------------------
```
sayboras: just carriage character.

state/state_store.go
```diff
@@ -6,6 +6,7 @@
 package state
 
 // StateStore is an interface to perform operations on store
+// nolint:golint
```
sayboras: I would love to rename this struct to just `State`, but need to refactor in all implementation for redis, cosmodb, etc. So just leave it for now.

go.mod
youngbupark: We use 1.13.3 in CI. it should be fine. cc/ @yaron2

state/state_store.go
youngbupark: Agree. let's keep it as is because it requires dapr change as well.

yaron2: This can be changed and will not affect Dapr until its updated to use the new version of contrib.
However, I would advise against using the name `State` for this interface, as I don't think its descriptive enough in regards to the methods that make up the interface.

For me, `State` might represent the actual state and is mutable. A `StateStore` is more representative of the entity that gets, saves, deletes, etc. the state itself.

sayboras: oh actually, what I meant is to change it to `Store`, previous comment is typo.

sayboras: @yaron2 @youngbupark I assume you guys are happy to make it `Store` :)

youngbupark: @yaron2 

> This can be changed and will not affect Dapr until its updated to use the new version of contrib.

It is true, but someone who want to develop with the component-contrib could fail because of this renaming.. :) 

@yaron2 @sayboras Yeah I overlooked @sayboras' comment on my phone. I thought it was `store` because package name is `state`. 

For renaming stuff, I would use the dedicated PR which is only for renaming because if somethings goes wrong, it is easier to revert the commit...

.github/workflows/components-contrib.yml
```diff
@@ -0,0 +1,61 @@
...
+            target_arch: arm
+          - os: macOS-latest
+            target_arch: arm
+    steps:
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v1
+      - name: Set up Go ${{ env.GOVER }}
+        uses: actions/setup-go@v1
+        with:
+          go-version: ${{ env.GOVER }}
```
youngbupark: This needs to be the first step because it will set up the environment variables for go and checkout step will checkout in the go module directory based on env var.

Makefile
```diff
@@ -0,0 +1,64 @@
...
+export GOOS ?= $(TARGET_OS_LOCAL)
+
+ifeq ($(GOOS),windows)
+BINARY_EXT_LOCAL:=.exe
+GOLANGCI_LINT:=golangci-lint.exe
+export ARCHIVE_EXT = .zip
+else
+BINARY_EXT_LOCAL:=
+GOLANGCI_LINT:=golangci-lint
+export ARCHIVE_EXT = .tar.gz
```
youngbupark: nit: I do not think we need this

Makefile
```diff
@@ -0,0 +1,64 @@
...
+   TARGET_OS_LOCAL = darwin
+else
+   TARGET_OS_LOCAL ?= windows
+endif
+export GOOS ?= $(TARGET_OS_LOCAL)
+
+ifeq ($(GOOS),windows)
+BINARY_EXT_LOCAL:=.exe
+GOLANGCI_LINT:=golangci-lint.exe
+export ARCHIVE_EXT = .zip
```
youngbupark: nit: I do not think we need this

state/state_store.go
sayboras: you are right, I leave it as it is now, will make the changes in subsequent PR.

sayboras:
> I left some nitpicked comments :) Thanks for your contribution!
> 
> @yaron2 please double-check this PR :)

Thanks, I have incorporated all of your comments 💯

state/state_store.go
youngbupark: > you are right, I leave it as it is now, will make the changes in subsequent PR.

@sayboras Please open the issue first to track to rename it.

go.mod
yaron2: Should be ok

state/state_store.go
sayboras: Here it is https://github.com/dapr/components-contrib/issues/67, I will start working on this after this PR is merged.

Issue #67, opened 2019-10-28
To rename StateStore interface in state package


sayboras:
## Describe the proposal

As discussed in https://github.com/dapr/components-contrib/pull/59#discussion_r339658021, the `StateStore` interface can be renamed to `Store` for simplicity.

youngbupark:
cc/ @yaron2 @Haishi2016 @amanbha @shalabhms @LMWF

yaron2:
I edited the title and body to reflect interface and not struct.

PR #70, opened 2019-10-29
Rename the StateStore interface

sayboras:
# Description

Fixes https://github.com/dapr/components-contrib/issues/67

## Issue reference
As mentioned in https://github.com/dapr/components-contrib/issues/67, we would like to rename interface `state.StateStore` to just `state.Store`.

I also noticed that there is inconsistency in naming for different store implementation such as `cosmodb` and `redis` compared to others (e.g. etcd, cassandra, consul etc). Hence, I have done some refactor as well. This might be breaking change, so would love to hear your inputs.

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


state/cosmosdb/cosmosdb.go
```diff
@@ -16,8 +16,8 @@ import (
 	"github.com/a8m/documentdb"
 )
 
-// StateStore is a CosmosDB state store
-type StateStore struct {
+// CosmosDB is a CosmosDB state store
+type CosmosDB struct {
```
youngbupark: package name is cosmosdb and you renamed StateStore to CosmosDB. cosmosdb.CosmosDB doesn't look good to me. Do you have the better idea for this ? maybe StateStore would be better.. cc/ @yaron2

sayboras: Oh i take a look on other implementation (eg etcd), and would like to make it consistent. Will be easier for consumer

sayboras: @shalabhms @youngbupark noted and thanks guys, let me update the PR.

PR #82, opened 2019-10-30
upgrade logrus to v1.4.2

thinkerou:
# Description

- upgrade logrus to v1.4.2
- remove `depguard` from golintci.yml
- use `go test ./...` instead of `go test ./... -mod=vendor`

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


Makefile
```diff
@@ -51,12 +51,12 @@ endif
 ################################################################################
 .PHONY: test
 test:
-	go test ./... -mod=vendor
+	go test ./...
```
youngbupark: we need to keep -mod=vendor. can you run `go mod vendor` in your change ?

cc/ @yaron2

thinkerou: @youngbupark Yes.
we use `go mod` and `GOPROXY`, so I suggest we not use `vendor`.

youngbupark: > @youngbupark Yes.
> we use `go mod` and `GOPROXY`, so I suggest we not use `vendor`.

@thinkerou Sorry for the late reply. This pr is to upgrade logurus to 1.4.2, which is not related to remove vendor.  So, let's keep using vendor now. Instead, please create the issue to remove vendor option in dapr/dapr.

cc/ @yaron2

thinkerou: @youngbupark OK, thanks!

thinkerou:
closing the pr.

Issue #92, opened 2019-11-01
Azure Service Bus pubsub implementation


jjcollinge:
## Azure ServiceBus PubSub
Add an Azure ServiceBus PubSub implementation. Opening this as a reference issue for an incoming PR. Proposal is fairly self explanatory.

Thought it might be useful to some folks to use a hosted message bus for pubsub rather than redis or the newly added NATS implementation. Apologies I didn't open this sooner to discuss, if we don't think this is a good idea then no problem!

Related: #73 
PR: #93

jjcollinge:
#93

Issue #73, opened 2019-10-29
Azure Service Bus Topics


alanwales:
Hello,

Is there a way to subscribe to a topic? I have tried the sample and can connect to a queue successfully but not to a topic (even with the /dapr/subscriptions endpoint added):

https://github.com/dapr/docs/blob/master/concepts/bindings/specs/servicebusqueues.md

Also I traced the http response and can't seem to get the message properties from the service bus message, just the content. I would like at least to get the CorrelationId from the message headers if that is supported? Thanks

yaron2:
Hey @alanwales , Azure Service Bus is a binding, meaning that in order to get events as input bindings there's no topic subscription needed, as you'd have with Pub-Sub.

alanwales:
Thanks for the reply, it's a little bit confusing the separation between a Binding and PubSub since most of the bindings have a pub-sub flavour to them, but I see that there is a pr open for Azure Service Bus PubSub added a few days ago so that will be really useful when it lands.

https://github.com/dapr/components-contrib/issues/92

Cheers

PR #93, opened 2019-11-01
Azure ServiceBus PubSub Implementation

jjcollinge:
# Description

I've added an Azure ServiceBus pubsub provider.

## Issue reference

#92, #73 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

## Notes:
I've ran the pubsub sample against this implementation. I'm leaving it as a draft whilst I clean up some bits and try to add more test coverage. I also have some open questions below:

- If you subscribe to a topic that does not exists, should we create it?
- Should a top level context be provided by dapr to the pubsub providers?
- Due to checking for existing topics/subscriptions this will be quite chatty - I can optimise this by caching some of the used subscriptions and topics rather than looking them up each time. Do you want me to include this optimisation in this PR or add it as a follow up?

## TODO:
- Rebase
- Documentation
- Dapr repo integration


yaron2:
hey @jjcollinge, thanks for this PR, much appreciated and we've gotten several asks about Cloud based Pub-Sub message buses, Azure Service Bus in particular.

> If you subscribe to a topic that does not exists, should we create it?

Yes, if possible: that logic should be part of the "as a service" offered by the abstraction.

> Should a top level context be provided by dapr to the pubsub providers?

In terms of ownership of the context, since Timeout Context are mostly created from metadata timeouts defined in the spec of the implementation, it makes sense contexts are created in the implementation and not handed over from the outside.

> Due to checking for existing topics/subscriptions this will be quite chatty - I can optimise this by caching some of the used subscriptions and topics rather than looking them up each time. Do you want me to include this optimisation in this PR or add it as a follow up?

Follow up. Smaller PRs are better. :)

yaron2:
Test failing on:

```
##[error]pubsub/azureservicebus/azureservicebus.go:31:16: undefined: servicebus
##[error]pubsub/azureservicebus/azureservicebus.go:32:16: undefined: servicebus
```

pubsub/azureservicebus/azureservicebus.go
```diff
@@ -0,0 +1,256 @@
...
+		err = a.createTopicEntity(topic)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (a *azureServiceBus) ensureSubscription(name string, topic string) error {
+	err := a.ensureTopic(topic) // TODO: should we create the topic if it doesn't exist?!
```
yaron2: Yes

alanwales:
> If you subscribe to a topic that does not exists, should we create it?

I think later down the line some people will object to this suggestion as they want to typically scope their access to [Read for consumers of a topic, whereas Manage rights](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-sas#rights-required-for-service-bus-operations) will need to be granted to create a topic which also allows enumerating and deleting other topics (perhaps too powerful). It would be great to be able to create a subscription if it doesn't exist but make this opt-in (therefore requires Manage access).

alanwales:
If you add creating a subscription dynamically (would be great), would also be nice to configure some properties of the subscription like AutoDeleteOnIdle which we use extensively during test runs:

```c#
var subscriptionDescription = new SubscriptionDescription(topicName, subscriptionName)
            {
                AutoDeleteOnIdle = TimeSpan.FromMinutes(30)
            };
```

jjcollinge:
> 
> 
> > If you subscribe to a topic that does not exists, should we create it?
> 
> I think later down the line some people will object to this suggestion as they want to typically scope their access to [Read for consumers of a topic, whereas Manage rights](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-sas#rights-required-for-service-bus-operations) will need to be granted to create a topic which also allows enumerating and deleting other topics (perhaps too powerful). It would be great to be able to create a subscription if it doesn't exist but make this opt-in (therefore requires Manage access).

Agree we want to try and allow people to use the minimum suitable scope. Just to clarify, you are suggesting a flag (i.e. `ManageEntities`) that if false will require all the topics and subscription to exist prior to dapr running and assume you're using either `send` or `listen` claims? I'd have to defer to @yaron2 for whether that is a desirable option.

> If you add creating a subscription dynamically (would be great), would also be nice to configure some properties of the subscription like AutoDeleteOnIdle which we use extensively during test runs:

Happy to expose more of the subscription config in the metadata (currently just `MaxDeliveryCount`). It's probably worth considering others from the [available settings](https://github.com/Azure/azure-service-bus-go/blob/54b2faa53e5216616e59725281be692acc120c34/subscription_manager.go#L101) too. Again I'll defer to @yaron2 whether this is suitable.

youngbupark:
> Happy to expose more of the subscription config in the metadata (currently just `MaxDeliveryCount`). It's probably worth considering others from the [available settings](https://github.com/Azure/azure-service-bus-go/blob/54b2faa53e5216616e59725281be692acc120c34/subscription_manager.go#L101) too. Again I'll defer to @yaron2 whether this is suitable.

`LockDuration` is one of useful options in Service Bus if multiple subscriber clients listens to the same topic and the app takes longer than 1 minute.

yaron2:
> > > If you subscribe to a topic that does not exists, should we create it?
> > 
> > 
> > I think later down the line some people will object to this suggestion as they want to typically scope their access to [Read for consumers of a topic, whereas Manage rights](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-sas#rights-required-for-service-bus-operations) will need to be granted to create a topic which also allows enumerating and deleting other topics (perhaps too powerful). It would be great to be able to create a subscription if it doesn't exist but make this opt-in (therefore requires Manage access).
> 
> Agree we want to try and allow people to use the minimum suitable scope. Just to clarify, you are suggesting a flag (i.e. `ManageEntities`) that if false will require all the topics and subscription to exist prior to dapr running and assume you're using either `send` or `listen` claims? I'd have to defer to @yaron2 for whether that is a desirable option.
> 
> > If you add creating a subscription dynamically (would be great), would also be nice to configure some properties of the subscription like AutoDeleteOnIdle which we use extensively during test runs:
> 
> Happy to expose more of the subscription config in the metadata (currently just `MaxDeliveryCount`). It's probably worth considering others from the [available settings](https://github.com/Azure/azure-service-bus-go/blob/54b2faa53e5216616e59725281be692acc120c34/subscription_manager.go#L101) too. Again I'll defer to @yaron2 whether this is suitable.

We have quite a few imlementations (especially state stores) that allow users to tweak configuration via the metadata fields in the YAML spec, and then a default behavior if that configuration is NOT required. So this is fine.

yaron2:
The way I'd do it is create topics and subscriptions automatically unless found in the spec metadata.

yaron2:
Thank you for this contribution @jjcollinge

PR #96, opened 2019-11-02
Add TLS connection option to Redis pub sub

### LINKED ISSUE #95: TLS config for Redis client
## Describe the proposal
Hello, I'm trying to use Dapr with Digital Ocean managed Redis which is using secured connection. I'm getting following message:

```
time="2019-11-02T17:45:45Z" level=warning msg="error initializing pub sub pubsub.redis: redis streams: error connecting to redis at private-dapr01-do-user-2603541-0.db.ondigitalocean.com:25061: EOF"
```

I suspect it is caused by the missing TLS configuration here:

https://github.com/dapr/components-contrib/blob/master/pubsub/redis/redis.go#L69

```go
	client := redis.NewClient(&redis.Options{
		Addr:            m.host,
		Password:        m.password,
		DB:              0,
		MaxRetries:      3,
		MaxRetryBackoff: time.Second * 2,
	})
```

Could you confirm or refute whether the current implementation of Redis pub sub supports secure connection?

Thanks in advance!

yaron2:
This PR allows Redis Pub/Sub to established a secured connection to TLS enabled Redis servers.

Closes #95.

### DIFF
--- a/pubsub/redis/redis.go
+++ b/pubsub/redis/redis.go
@@ -79,9 +79,10 @@ func (r *redisStreams) Init(metadata pubsub.Metadata) error {
 		MaxRetryBackoff: time.Second * 2,
 	}
 
+	/* #nosec */
 	if r.metadata.enableTLS {
 		options.TLSConfig = &tls.Config{
-			InsecureSkipVerify: r.metadata.enableTLS, /* #nosec */
+			InsecureSkipVerify: r.metadata.enableTLS,
 		}
 	}
 
--- a/pubsub/redis/redis.go
+++ b/pubsub/redis/redis.go
@@ -81,7 +81,7 @@ func (r *redisStreams) Init(metadata pubsub.Metadata) error {
 
 	if r.metadata.enableTLS {
 		options.TLSConfig = &tls.Config{
-			InsecureSkipVerify: r.metadata.enableTLS,
+			InsecureSkipVerify: r.metadata.enableTLS, /* #nosec */
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


yaron2:
@youngbupark we need to disable `gosec` from the linter. It's failing on valid code, in this case `InsecureSkipVerify` set to true.

pubsub/redis/redis.go
```diff
@@ -58,16 +69,23 @@ func (r *redisStreams) Init(metadata pubsub.Metadata) error {
...
 		Password:        m.password,
 		DB:              0,
 		MaxRetries:      3,
 		MaxRetryBackoff: time.Second * 2,
-	})
+	}
+
+	if r.metadata.enableTLS {
+		options.TLSConfig = &tls.Config{
+			InsecureSkipVerify: r.metadata.enableTLS,
```
youngbupark: Add `/* #nosec */` here to disable gosec 
```
InsecureSkipVerify: r.metadata.enableTLS, /* #nosec */
```

youngbupark:
@yaron2 will you add tls option to redis state component ?

yaron2:
> @yaron2 will you add tls option to redis state component ?

Yes, that will go into a deeper change to use the `go-redis/redis` package in the state component instead of `redispipe`.

yaron2:
> We can use /* #nosec */ to bypass the gosec

Added it, still fails.

yaron2:
Alright, got it to pass. should have gone at the IF block scope.

PR #100, opened 2019-11-04
Couchbase state store implementation

### LINKED ISSUE #98: Provide a state store implementation for Couchbase
## Describe the proposal

Provide a state store implementation for [Couchbase](https://docs.couchbase.com/home/index.html)

abhirockzz:
# Description

State store implementation for Couchbase

## Issue reference

Closes #98 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


abhirockzz:
This is work in progress. Ideally, the implementation should take into consistency, concurrency etc. I have posted questions in Issue #98 to gain more clarity

state/couchbase/couchbase.go
```diff
@@ -0,0 +1,180 @@
...
+	} else {
+		value, err = cbs.json.MarshalToString(req.Value)
+	}
+
+	if err != nil {
+		return fmt.Errorf("couchbase error: failed to convert value %v", err)
+	}
+	//Under RBAC, bucket-passwords are no longer used
+	//see https://docs.couchbase.com/go-sdk/1.6/sdk-authentication-overview.html#authenticating-with-legacy-sdk-versions
+	bucket, err := cbs.cbCluster.OpenBucket(cbs.bucket, "")
```
yaron2: is `bucket` a cheap pass through object? If not, wouldn't it be better to create one on init? in the future when we enable a bucket to be specified per-call in the metadata, it could create an new bucket if needed.

abhirockzz: sure I think that can be done. bu that brings up another (generic) question - how does one close resources (connections etc.) spawned inside `Init()`?

abhirockzz:
I left some notes in the original issue #98 w.r.t consistency

state/couchbase/couchbase.go
yaron2: Dapr doesn't dispose of components unless the process is dead, so generally there's no external condition under which it would be needed to keep a reference to a component but close any of its internal resources.

Thinking about non-Dapr users, it might be useful in the future to add a `Close()` method for components.

abhirockzz: bucket is now created during `Init()`

abhirockzz:
can be merged?

Issue #97, opened 2019-11-03
Add Cloud Firestore for state


yokawasa:
## Describe the proposal

Add support for Cloud Firestore (in datastore mode) as a state component.

yokawasa:
I started working on this

yaron2:
Thanks, will be great to see this.

PR #110, opened 2019-11-07
Added state store: Cloud Firestore 

yokawasa:
# Description

Add Cloud Firestore for state

## Issue reference

Please reference the issue this PR will close: #97 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


yaron2:
Thanks for this PR.

Is this adding support for Cloud Datastore or Firestore?

The package used here is for Datastore.
It looks like Firestore has its own package:

https://firebase.google.com/docs/firestore/quickstart

yokawasa:
@yaron2 

My intension is to support `Cloud Firestore in in datastore mode`. Google Cloud Firestore is going to replace the legacy Google Cloud Datastore soon. One then has the choice between using Cloud Firestore in "native mode" or in "datastore mode". The former allows access to Firestore through the usual Firestore SDK while the latter allows usage of the old Cloud Datastore SDK (which has no Web/Mobile APIs).

I think this is very confusing point. If you think it's not appropriate to use `firestore` for the package name, I'll change it to `datastore`, `clouddatastore`, or something else. 

The reason I picked "datastore mode" is that I thoguht mobile and web client libraries with real-time and offline features not needed for dapr state store. 

Here is an official guide from google.
> We recommend the following when choosing between database modes:
> - **Use Cloud Firestore in Datastore mode for new server projects**: Cloud Firestore in Datastore mode allows you to use established Cloud Datastore server architectures while removing fundamental Cloud Datastore limitations. Datastore mode can automatically scale to millions of writes per second.
> - **Use Cloud Firestore in Native mode for new mobile and web apps**: Cloud Firestore offers mobile and web client libraries with real-time and offline features. Native mode can automatically scale to millions of concurrent clients.

See this for more deail 
- https://cloud.google.com/firestore/docs/firestore-or-datastore
- https://cloud.google.com/datastore/docs/upgrade-to-firestore

yaron2:
You're right, with what you're describing now the choice for Dapr is indeed "datastore" mode and not the native firestore.

As indeed Datastore will be rebranded as Firestore, lets keep the package name firestore.

Thanks!

state/Readme.md
```diff
@@ -11,6 +11,7 @@ Currently supported state stores are:
 * Azure CosmosDB
 * Memcached
 * Zookeeper
+* Cloud Firestore
```
yaron2: Can we change this to `Cloud Firestore (Datastore mode)`?

yokawasa: @yaron2  Sure, I can

yaron2:
LGTM, thank you @yokawasa!

Issue #87, opened 2019-10-31
Add transactions support to MongoDB


yaron2:
## Describe the proposal

Implement the Transactions interface for MongoDB.
That would make MondoDB be able to operate multi-operation transactions and be certified for an Actor state store for Dapr.

PR #111, opened 2019-11-07
Adding transactions support to MongoDB

shalabhms:
# Description
Implementing the Transactions interface for MongoDB.
That would make MongoDB be able to operate multi-operation transactions 

## Issue reference

#87 

Please reference the issue this PR will close: #87

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/state/mongodb/mongodb.go
+++ b/state/mongodb/mongodb.go
@@ -113,6 +113,19 @@ func (m *MongoDB) Init(metadata state.Metadata) error {
 
 // Set saves state into MongoDB
 func (m *MongoDB) Set(req *state.SetRequest) error {
+	ctx, cancel := context.WithTimeout(context.Background(), m.operationTimeout)
+	defer cancel()
+
+	err := m.setInternal(ctx, req)
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (m *MongoDB) setInternal(ctx context.Context, req *state.SetRequest) error {
 	var vStr string
 	b, ok := req.Value.([]byte)
 	if ok {
@@ -120,9 +133,6 @@ func (m *MongoDB) Set(req *state.SetRequest) error {
 	} else {
 		vStr, _ = json.MarshalToString(req.Value)
 	}
-
-	ctx, cancel := context.WithTimeout(context.Background(), m.operationTimeout)
-	defer cancel()
 
 	// create a document based on request key and value
 	filter := bson.M{id: req.Key}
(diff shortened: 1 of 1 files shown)


state/mongodb/mongodb.go
```diff
@@ -113,6 +113,19 @@ func (m *MongoDB) Init(metadata state.Metadata) error {
 
 // Set saves state into MongoDB
 func (m *MongoDB) Set(req *state.SetRequest) error {
+	ctx, cancel := context.WithTimeout(context.Background(), m.operationTimeout)
```
youngbupark: JFYI - I'm worried that we use context.Background() in many places. In DotNet, cancellationtoken is used to cooperate cancellation among thread/thread pools/work items/Tasks and normally we pass this token down to function calls. 

In Go, Incoming request has its own context and its context should be properly propagated to function call chains during the request life-cycle.  I created the issue pointing out this - https://github.com/dapr/dapr/issues/292

state/mongodb/mongodb.go
```diff
@@ -196,6 +216,53 @@ func (m *MongoDB) BulkDelete(req []state.DeleteRequest) error {
...
+				return fmt.Errorf("error during transaction, aborting the transaction: %s", err)
+			}
+		} else if o.Operation == state.Delete {
+			req := o.Request.(state.DeleteRequest)
+			err := m.deleteInternal(sessCtx, &req)
+
+			if err != nil {
+				sessCtx.AbortTransaction(sessCtx)
+				return fmt.Errorf("error during transaction, aborting the transaction: %s", err)
+			}
```
youngbupark: nit: this block is duplicated with line 248-251. I think we can check the error outside of `if-statement`.

state/mongodb/mongodb.go
youngbupark: We do not need to handle this issue now :)

yaron2: For components specifically, in most cases contexts are created when a metadata property for a timeout is given. That means the context is created by a timeout defined by the user. If a transient connection occurs between the client and Dapr which the user didn't mean to cancel, it will cancel an underlying operation also even though the user did not mean to. That will hurt some of Dapr's reliability guarantees.

yaron2:
@shalabhms have you tested this locally against MongoDB?

shalabhms:
> @shalabhms have you tested this locally against MongoDB?

It is in progress.

yaron2:
> > @shalabhms have you tested this locally against MongoDB?
> 
> It is in progress.

OK. I'll approve once verified tested.

shalabhms:
> > > @shalabhms have you tested this locally against MongoDB?
> > 
> > 
> > It is in progress.
> 
> OK. I'll approve once verified tested.

It is now tested locally.

yaron2:
LGTM

Issue #49, opened 2019-10-24
Add dynamodb for state


sayboras:
## Describe the proposal

To have support for dynamodb for state.

sayboras:
If no one is working on this, I can pick up this.

amanbha:
Sure, I have assigned it to you. 
 Thanks.

yaron2:
@sayboras anything you need help with?

sayboras:
> @sayboras anything you need help with?

I am just back from travelling this week, will plan sometime to wrap it up.

sayboras:
@yaron2 need your input on how to handle `key` in dynamodb. Basically, we might need to have `key_col_num` during Set and Get methods. I can put this column name in `Metadata` of `GetRequest` or `SetRequest`, but feel like it's a hacky way.

``` golang
type GetRequest struct {
	Key      string            `json:"key"`
	Metadata map[string]string `json:"metadata"`
	Options  GetStateOption    `json:"options,omitempty"`
}
```
```golang
getItemInput := &dynamodb.GetItemInput{
		Key: map[string]*dynamodb.AttributeValue{
			"key_col_name": {
				S: aws.String(ID),
			},
		},
		TableName:      aws.String(ddb.cfg.DBTableProducts),
		ConsistentRead: aws.Bool(true),
	}
```

yaron2:
> @yaron2 need your input on how to handle `key` in dynamodb. Basically, we might need to have `key_col_num` during Set and Get methods. I can put this column name in `Metadata` of `GetRequest` or `SetRequest`, but feel like it's a hacky way.
> 
> Another appoach is to use fixed key col num (e.g. `key`), however this will give very limited flexibility for consumer.
> 
> ```go
> type GetRequest struct {
> 	Key      string            `json:"key"`
> 	Metadata map[string]string `json:"metadata"`
> 	Options  GetStateOption    `json:"options,omitempty"`
> }
> ```
> 
> Code snippet on how to retrieve item.
> 
> ```go
> getItemInput := &dynamodb.GetItemInput{
> 		Key: map[string]*dynamodb.AttributeValue{
> 			"key_col_name": {
> 				S: aws.String(req.Key),
> 			},
> 		},
> 		TableName:      aws.String("table_name"),
> 		ConsistentRead: aws.Bool(true),
> 	}
> ```

We can use our own struct as the data object for DynamoDB which had two fields: Key and Data.

Then in GetItem requests we can select by the Key field.

[...]

PR #116, opened 2019-11-11
Feature/state dynamodb

sayboras:
# Description

Addressed https://github.com/dapr/components-contrib/issues/49

## Issue reference
Please reference the issue this PR will close: https://github.com/dapr/components-contrib/issues/49

## Checklist
Current coverage is 71.2%

* [ ] Code compiles correctly
* [ ] Created/updated tests


yaron2:
@sayboras any updates?

yaron2:
Thanks for the updates, I see this is still WIP. 

Loking forward to this being ready for review!

yaron2:
@sayboras do you want plan to continue, or do you want someone else to pick this up?

sayboras:
@yaron2 I just pushed initial implemetation for your early review. Happy to address your comment. 

Note: integration testing is still pending.

state/aws/dynamodb/dynamodb.go
```diff
@@ -0,0 +1,168 @@
...
+		Item:      item,
+		TableName: &d.table,
+	}
+
+	_, e := d.client.PutItem(input)
+	return e
+}
+
+// BulkSet performs a bulk set operation
+func (d *StateStore) BulkSet(req []state.SetRequest) error {
```
sayboras: not completely happy with this implementation, I might change it to WriteBatchRequest later.

go.sum
```diff
@@ -30,7 +30,6 @@ github.com/Azure/azure-pipeline-go v0.1.8/go.mod h1:XA1kFWRVhSK+KNFiOhfv83Fv8L9a
 github.com/Azure/azure-pipeline-go v0.2.1 h1:OLBdZJ3yvOn2MezlWvbrBMTEUQC72zAftRZOMdj5HYo=
 github.com/Azure/azure-pipeline-go v0.2.1/go.mod h1:UGSo8XybXnIGZ3epmeBw7Jdz+HiUVpqIlpz/HKHylF4=
 github.com/Azure/azure-sdk-for-go v21.3.0+incompatible/go.mod h1:9XXNKU+eRnpl9moKnB4QOLf1HestfXbmab5FXxiDBjc=
-github.com/Azure/azure-sdk-for-go v29.0.0+incompatible h1:CYPU39ULbGjQBo3gXIqiWouK0C4F+Pt2Zx5CqGvqknE=
```
artursouza: Any reason we are removing these?

state/aws/dynamodb/dynamodb.go
artursouza: Please, create an issue after merging.

state/aws/dynamodb/dynamodb.go
```diff
@@ -0,0 +1,168 @@
...
+			},
+		},
+		TableName: aws.String(d.table),
+	}
+	_, err := d.client.DeleteItem(input)
+	return err
+}
+
+// BulkDelete performs a bulk delete operation
+func (d *StateStore) BulkDelete(req []state.DeleteRequest) error {
```
artursouza: Add this to the tech debt: batch operation for deletes too.

state/aws/dynamodb/dynamodb.go
```diff
@@ -0,0 +1,168 @@
...
+		return nil, err
+	}
+	return &state.GetResponse{
+		Data: []byte(output),
+	}, nil
+}
+
+// Set saves a dynamoDB item
+func (d *StateStore) Set(req *state.SetRequest) error {
+	item := map[string]*dynamodb.AttributeValue{
```
artursouza: You are not handling conditional put.

state/aws/dynamodb/dynamodb.go
```diff
@@ -0,0 +1,168 @@
...
+		"key": {
+			S: aws.String(req.Key),
+		},
+		"value": {
+			S: aws.String(fmt.Sprintf("%v", req.Value)),
```
artursouza: I wonder how this would work for other types too. Double, boolean, byte[] and even JSON (since DynamoDB has native support for JSON). This is tricky in the unit test but would be good to know if all those would work with this implementation.

go.sum
sayboras: oh just got clean up by `go mod tidy`

yaron2:
@sayboras have you tested this? meaning, called all the interface methods and verified the desired behavior? (saving/getting/deleting state)

sayboras:
@artursouza @yaron2 unfortunately it's not working for all data type. I am thinking about to define own Interface something like below
~~~
type Interface interface {
      Marshal() ([]byte, error)
      Unmarshel([]byte) interface{}, err
} 
~~~

The downside is that we delegate to consumer, we can make it easily by leveraging the exsiting Endcoder and Decoder from aws dynamo sdk.

state/aws/dynamodb/dynamodb.go
youngbupark: @sayboras [Redis implementation](https://github.com/dapr/components-contrib/blob/37344917183400724f1382a3b314a1ed7d720b5b/state/redis/redis.go#L203-L208) is always serializing req.Value. dapr runtime will pass req.value object to this. it would be good if you can check the type of object and then save it differently. otherwise, let's serialize req.value like redis implementation. 

cc/ @yaron2  @artursouza

youngbupark:
> @artursouza @yaron2 unfortunately it's not working for all data type. I am thinking about to define own Interface something like below
> 
> ```
> type Interface interface {
>       Marshal() ([]byte, error)
>       Unmarshel([]byte) interface{}, err
> } 
> ```
> 
> The downside is that we delegate to consumer, we can make it easily by leveraging the exsiting Endcoder and Decoder from aws dynamo sdk.



> @artursouza @yaron2 unfortunately it's not working for all data type. I am thinking about to define own Interface something like below
> 
> ```
> type Interface interface {
>       Marshal() ([]byte, error)
>       Unmarshel([]byte) interface{}, err
> } 
> ```
> 
> The downside is that we delegate to consumer, we can make it easily by leveraging the exsiting Endcoder and Decoder from aws dynamo sdk.

@sayboras [Redis implementation](https://github.com/dapr/components-contrib/blob/37344917183400724f1382a3b314a1ed7d720b5b/state/redis/redis.go#L203-L208) is always serializing req.Value. dapr runtime will pass req.value object to this. it would be good if you can check the type of object and then save it differently. otherwise, let's serialize req.value like redis implementation. 

cc/ @yaron2  @artursouza

state/aws/dynamodb/dynamodb.go
artursouza: The only way out for this might be to save it as byte[]. We cannot determine what is the format and structured used in req.Value. It might be JSON or not.

yaron2:
@artursouza is this PR in merge condition or is there anything else needs to be done? Would be great if you could re-review

state/aws/dynamodb/dynamodb.go
gmusial-obj: what worries me with batch operations done in a loop is the high probability of exceeding DynamoDB's provisioned throughput for writes. The issue is hidden in a single-element write too, because if we start calling state store too frequent from our app even with throughput autoscaling we are still going to experience 4XX responses due to insufficient provisioned throughput for a short period of time. The issue does not affect for example redis. And my question is, is there any storage supervisor that will handle retries or should the store be self-contained and do retries too.

yaron2: If the dynamo API has a bulk/batch operation that's the preferable route.

It should be assumed that retries will be handled globally by Dapr.

yaron2:
OK @sayboras looks like we're good to go!

artursouza:
Follow up items are in: #245 #246 #247

PR #118, opened 2019-11-12
Servicediscovery

LMWF:
# Description
## Issue reference

This ports service discovery into a component, so it can be pluggable.
https://github.com/dapr/dapr/issues/753

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


go.mod
```diff
@@ -28,38 +25,29 @@ require (
 	github.com/coreos/go-semver v0.3.0 // indirect
 	github.com/coreos/go-systemd v0.0.0-20190719114852-fd7a80b32e1f // indirect
 	github.com/coreos/pkg v0.0.0-20180928190104-399ea9e2e55f // indirect
-	github.com/eapache/go-resiliency v1.2.0 // indirect
+	github.com/dapr/dapr v0.1.0
```
youngbupark: component-contrib should not have the dependency of dapr.

LMWF: Yep, forgot to replace call to the function that was moved into components-contrib.

servicediscovery/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,27 @@
...
+
+type Resolver struct {
+}
+
+func NewKubernetesResolver() *Resolver {
+	return &Resolver{}
+}
+
+func (z *Resolver) ResolveID(req *servicediscovery.ResolveRequest) (string, error) {
+	namespace := os.Getenv("NAMESPACE")
```
youngbupark: I would define const for "NAMESPACE"

servicediscovery/requests.go
```diff
@@ -0,0 +1,12 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+package servicediscovery
+
+type ResolveRequest struct {
+	ID       string
+	GrpcPort int
```
youngbupark: I think this could be any port :) wouldn't it be good to rename to `port`?

servicediscovery/mdns/mdns.go
```diff
@@ -0,0 +1,66 @@
...
+	port, err := LookupPortMDNS(req.ID)
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("localhost:%v", port), nil
+}
+
+// LookupPortMDNS uses mdns to find the port of a given service entry on a local network
+func LookupPortMDNS(id string) (int, error) {
+	resolver, err := zeroconf.NewResolver(nil)
```
youngbupark: Do we need to create resolver instance every time? can't we cache it?

servicediscovery/kubernetes/kubernetes.go
yaron2: I would get Namespace from req.Metadata, and have Dapr pass it in. Because its the Dapr runtime that is better off pulling environment variables and passing them to the component, which should expect to get it.

If Namespace is nil or empty, it should default to `default`.

servicediscovery/requests.go
yaron2: I agree, rename to `Port`.

servicediscovery/mdns/mdns.go
yaron2: It should be possible to cache this.

servicediscovery/kubernetes/kubernetes_test.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+func TestResolve(t *testing.T) {
+	resolver := NewKubernetesResolver()
+	request := servicediscovery.ResolveRequest{ID: "myid", GrpcPort: 1234}
+
+	// ResolveID(), the function we are testing, formats this environment variable into part of the returned string.
+	// So as not to affect the local environment, we'll just read in whatever value the local environment
+	// has it set to (probably nothing, or empty string) and verify the string format vs that.
+	namespace := os.Getenv("NAMESPACE")
```
yaron2: Best if we don't rely or reference env vars from tests. If this can be any value, just mock one up.

servicediscovery/requests.go
LMWF: Yep

servicediscovery/kubernetes/kubernetes.go
LMWF: Done

servicediscovery/kubernetes/kubernetes_test.go
LMWF: Removed since namespace is getting passed in request now.

servicediscovery/requests.go
```diff
@@ -0,0 +1,12 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+package servicediscovery
+
+type ResolveRequest struct {
+	Id       string
+	GrpcPort int
```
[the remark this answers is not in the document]
LMWF: Changed to "Port"

servicediscovery/mdns/mdns.go
LMWF: Put inside a once.Do

servicediscovery/mdns/mdns.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/dapr/components-contrib/servicediscovery"
+	"github.com/grandcat/zeroconf"
+)
+
+func NewmDNSResolver() *MDNSResolver {
```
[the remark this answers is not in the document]
LMWF: Re-cased

servicediscovery/mdns/mdns.go
```diff
@@ -0,0 +1,66 @@
...
+		return "", err
+	}
+	return fmt.Sprintf("localhost:%v", port), nil
+}
+
+// LookupPortMDNS uses mdns to find the port of a given service entry on a local network
+func LookupPortMDNS(id string) (int, error) {
+	resolver, err := zeroconf.NewResolver(nil)
+	if err != nil {
+		return -1, fmt.Errorf("Failed to initialize resolver: %e", err)
```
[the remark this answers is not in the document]
LMWF: I think you may be looking at an earlier version.  That was fixed earlier.

servicediscovery/mdns/mdns.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	port := -1
+	entries := make(chan *zeroconf.ServiceEntry)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*1)
```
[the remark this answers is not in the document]
LMWF: This was copied from the dapr repo where it was 1 sec.  But I don't think there's a "default"?  I only see 15 sec in zeroconf sample code, but that's not a default.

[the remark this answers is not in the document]
LMWF: Discussed with Yaron, it's set to 1 on purpose.  The time it takes varies a lot between OSes and   15 is too long to wait locally.

servicediscovery/mdns/mdns.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/grandcat/zeroconf"
+)
+
+func NewMDNSResolver() *Resolver {
+	return &Resolver{}
+}
+
+type Resolver struct {
+	Once     sync.Once
+	Resolver *zeroconf.Resolver
```
youngbupark: We do not need to expose these members to the external package. I would make these member variable names lower-case.

servicediscovery/requests.go
```diff
@@ -0,0 +1,19 @@
...
+
+type ResolveRequest struct {
+	ID        string
+	Namespace string
+	Port      int
+	Data      map[string]string
+}
+
+func NewResolveRequest() *ResolveRequest {
+	return &ResolveRequest{Namespace: "default"}
```
youngbupark: nit: you might want to use DefaultNamespace const instead of "default"

servicediscovery/mdns/mdns.go
```diff
@@ -0,0 +1,75 @@
...
+	var err error
+	var t *zeroconf.Resolver
+	z.Once.Do(func() {
+		t, err = zeroconf.NewResolver(nil)
+		z.Resolver = t
+	})
+
+	if err != nil {
+		return -1, fmt.Errorf("failed to initialize resolver: %e", err)
+	}
```
youngbupark: I would move this block inside once.do {}

servicediscovery/mdns/mdns.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("localhost:%v", port), nil
+}
+
+// LookupPortMDNS uses mdns to find the port of a given service entry on a local network
+func (z *Resolver) LookupPortMDNS(id string) (int, error) {
+	var err error
+	var t *zeroconf.Resolver
```
youngbupark: we do not need to declare the variable here. instead we can use `:=` assigner inside once.do {}

servicediscovery/requests.go
LMWF: It's meant to be 'default' like the k8s default namespace.

youngbupark: You defined `DefaultNamespace` const in this code. why not use that const?

servicediscovery/mdns/mdns_test.go
```diff
@@ -0,0 +1,28 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+package mdns
+
+import (
+	//	"fmt"
```
youngbupark: nit: I would remove this line

servicediscovery/mdns/mdns.go
LMWF: It's outside because this function needs to return if there's an error, and the function has different return parameters than the function literal.

servicediscovery/mdns/mdns.go
LMWF: This needs to be outside for the same reason as the comment above.

servicediscovery/requests.go
LMWF: Oh, right.  Didn't replace, done.

servicediscovery/mdns/mdns.go
```diff
@@ -0,0 +1,75 @@
...
+	z.once.Do(func() {
+		t, err = zeroconf.NewResolver(nil)
+		z.resolver = t
+	})
```
youngbupark: What about this scneario? 

if NewResolver throws the error at the first call, then this block will never be called by the next call. then err is not caught in line 43-46 then it will make panic.

LMWF: Refactored

servicediscovery/mdns/mdns.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/dapr/components-contrib/servicediscovery"
+	"github.com/grandcat/zeroconf"
+)
+
+func NewMDNSResolver() *Resolver {
```
yaron2: Last comment, then lgtm:

It's better to return the `Resolver` interface here, and not the Resolver struct. You can make the Resolver struct unexported (private) by changing it to `resolver`.

yaron2:
LGTM

Issue #105, opened 2019-11-05
Add RabbitMQ for Pub/Sub


lukekim:
Support RabbitMQ as a Pub/Sub component.

yaron2:
This is for pub/sub, right? (We have a RabbitMQ binding) - I suggest removing the word bindings from the title.

fbeltrao:
Is anyone working on this already?
If not, I would love to contribute.

markrussinovich:
I was planning on working on it this weekend, but if you want to take this, let me know.

fbeltrao:
If you don't mind I would like to take it. It is a good first item for me to get familiar with the project internals.

markrussinovich:
Okay, it's yours.

amanbha:
https://github.com/dapr/components-contrib/tree/master/pubsub/rabbitmq

PR #124, opened 2019-11-15
Add RabbitMQ pub/sub

fbeltrao:
# Description

Add RabbitMQ as pub/sub provider.
This is how the components file looks like:
```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: messagebus
spec:
  type: pubsub.rabbitmq
  metadata:
  - name: host
    value: amqp://guest:guest@localhost:5672/
```

## Issue reference

#105 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

## Additional information

- I plan to add e2e tests once there is guidelines on how to do it
- I am learning Go, all feedback is welcome!
- Once this is accepted I will submit the PR to dapr/dapr to register RabbitMQ as a pub/sub


yaron2:
thans @fbeltrao! Review will lag a bit as some of us are at Kubecon this week :)

LMWF:
I'll review this in the next few days.

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -0,0 +1,176 @@
...
+		if err != nil {
+			requeue := r.metadata.requeueInFailure && !d.Redelivered
+
+			log.Debugf("%s nacking message '%s' from topic '%s', requeue=%t", logMessagePrefix, d.MessageId, topic, requeue)
+			if err = r.channel.Nack(d.DeliveryTag, false, requeue); err != nil {
+				log.Errorf("%s error nacking message '%s' from topic '%s', %s", d.MessageId, logMessagePrefix, topic, err)
+			}
+		} else {
+			log.Debugf("%s acking message '%s' from topic '%s'", logMessagePrefix, d.MessageId, topic)
+			if err = r.channel.Ack(d.DeliveryTag, false); err != nil {
```
yaron2: Should only ack here if the `handler` method from line 139 returned nil

fbeltrao: Hi @yaron2,

Thanks for the review.
I think I am missing something here. Thinking about the scenarios:

1. Auto ack == true
  In this case we don't ack or nack since the message was already acked during dequeue. This option was by default true, but I am changed to false to better handle scenarios where daprd is running before the consumer is ready.

2. AutoAck == false and err != nil
  Flow goes to lines:
    - 141 (log error)
    - 147 (block, including logging and nack)

3. AutoAck == false and err == nil
  Flow goes to line:
    - 155 (block, including logging and ack)

fbeltrao:
@yaron2 being new to golang I don't know how to fix the check failure caused by my last commit. It seems to be something related to go modules. Any help would be appreciated.

yaron2:
no I'll take a look

yaron2:
@fbeltrao can you try to run `go mod tidy` locally and push any changes?
Also, are you sure you don't have any stashed or uncommitted changes?

yaron2:
Thanks for your contribution @fbeltrao!

Issue #56, opened 2019-10-25
The names of the packages need to be consistent with name of the company


msfussell:
for example 

- pubsub should be gcp-pubsub 
- sns should be aws-sns 
- blobstorage should be azure-blobstorage

PR #128, opened 2019-11-18
Changes to include Bindings Package names to be consistent with company name

shalabhms:
Changing the names of the packages directory so to be consistent with name of the company .

It is to be noted that package name can not have hyphen "-" in the name.
So changing the directory names and putting under company names.

For example - dynamodb  package would be

dapr/components-contrib/bindings/aws/dynamodb

## Issue reference

#56 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


youngbupark:
I thought we would use prefix instead of creating sub directory. for example, `azure-keyvault`

shalabhms:
> I thought we would use prefix instead of creating sub directory. for example, `azure-keyvault`

I tried that first, but looks like we can not have "-" in the package name. I highlighted this in the PR description.

Issue #147, opened 2019-12-01
Support Azure SignalR output binding


fbeltrao:
Add output binding for Azure SignalR hubs, enabling scenarios where user notifications can be sent directly from Dapr.

Binding definition:
```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: signalr_out
spec:
  type: bindings.azure.signalr
  metadata:
  - name: connectionString
    value: Endpoint=https://<your-azure-signalr>.service.signalr.net;AccessKey=<your-access-key>;Version=1.0;
```

Calling from code:
```javascript
const daprPort = process.env.DAPR_HTTP_PORT || 3500;
const bindingUrl = `http://localhost:${daprPort}/v1.0/bindings`;
fetch(`${bindingUrl}/signalr_out`, {
    method: 'POST',
    body: JSON.stringify({
        data: {
            Target: "newMessage",
            Arguments: [
                {
                    sender: "dapr",
                    text: "notification from dapr"
                }]
        },
        metadata: {
            hub: "chat"
        }
    }),
    headers: {
        "Content-Type": "application/json"
    }
})
...
```

davidfowl:
Very cool! @yaron2 are output bindings always an explicit call or can they also be invoked by returning a JSON object from a dapr invocation?

yaron2:
> Very cool! @yaron2 are output bindings always an explicit call or can they also be invoked by returning a JSON object from a dapr invocation?

They can also be invoked with a special return JSON, yes.

davidfowl:
@yaron2 this should be closed

yaron2:
closed via https://github.com/dapr/components-contrib/pull/148

PR #148, opened 2019-12-01
Add Azure SignalR binding

fbeltrao:
# Description

Add support to Azure SignalR as output binding

## Issue reference

#147 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/bindings/azure/signalr/signalr.go
+++ b/bindings/azure/signalr/signalr.go
@@ -75,13 +75,10 @@ func (s *SignalR) Init(metadata bindings.Metadata) error {
 				if s.endpoint[len(s.endpoint)-1] == '/' {
 					s.endpoint = s.endpoint[:len(s.endpoint)-1]
 				}
-				break
 			case "AccessKey":
 				s.accessKey = connectionValue[i+1:]
-				break
 			case "Version":
 				s.version = connectionValue[i+1:]
-				break
 			}
 		}
 	}
--- a/bindings/azure/signalr/signalr.go
+++ b/bindings/azure/signalr/signalr.go
@@ -14,7 +14,7 @@ import (
 	"time"
 
 	"github.com/dapr/components-contrib/bindings"
-	"github.com/gbrlsnchs/jwt/v3"
+	"github.com/dgrijalva/jwt-go"
 	"github.com/pkg/errors"
 
 	log "github.com/Sirupsen/logrus"
@@ -66,7 +66,7 @@ func (s *SignalR) Init(metadata bindings.Metadata) error {
 	// Expected: Endpoint=https://<servicename>.service.signalr.net;AccessKey=<access key>;Version=1.0;
 	connectionValues := strings.Split(strings.TrimSpace(connectionString), ";")
 	for _, connectionValue := range connectionValues {
-		if i := strings.Index(connectionValue, "="); i != -1 {
+		if i := strings.Index(connectionValue, "="); i != -1 && len(connectionValue) > (i+1) {
 			k := connectionValue[0:i]
 			switch k {
 			case "Endpoint":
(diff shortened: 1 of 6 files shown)


yaron2:
cc @davidfowl

bindings/azure/signalr/signalr.go
```diff
@@ -0,0 +1,195 @@
...
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/dapr/components-contrib/bindings"
+	"github.com/gbrlsnchs/jwt/v3"
```
davidfowl: Is this the defactor JWT package for golang projects?

yaron2: If there'a a "defacto" package for JWT for Go I'd say its [this](https://github.com/dgrijalva/jwt-go) one..

@fbeltrao would it be possible to make the switch unless there are special reasons to use the current one?

fbeltrao: Wasn’t sure about that one too. Thanks for the tip.

bindings/azure/signalr/signalr.go
```diff
@@ -0,0 +1,195 @@
...
+	version    string
+	hub        string
+	tokens     map[string]signalrCachedToken
+	httpClient *http.Client
+}
+
+// Init is responsible for initializing the SignalR output based on the metadata
+func (s *SignalR) Init(metadata bindings.Metadata) error {
+	connectionString, ok := metadata.Properties[connectionStringKey]
+	if !ok || connectionString == "" {
```
davidfowl: for my edification, `connectionString == ""` or `len(connectionString) == 0`? I see both patterns in this file.

bindings/azure/signalr/signalr.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	// Expected: Endpoint=https://<servicename>.service.signalr.net;AccessKey=<access key>;Version=1.0;
+	connectionValues := strings.Split(strings.TrimSpace(connectionString), ";")
+	for _, connectionValue := range connectionValues {
+		if i := strings.Index(connectionValue, "="); i != -1 {
+			k := connectionValue[0:i]
+			switch k {
+			case "Endpoint":
+				s.endpoint = connectionValue[i+1:]
```
jjcollinge: Is it worth checking i+1 exists before indexing into the array? i.e. `Endpoint=;`

bindings/azure/signalr/signalr.go
fbeltrao: According to what I [read](https://stackoverflow.com/questions/18594330/what-is-the-best-way-to-test-for-an-empty-string-in-go) the compiler will do make both [equal](https://stackoverflow.com/questions/18594330/what-is-the-best-way-to-test-for-an-empty-string-in-go/54804858#54804858), so we should use the one that is more readable.

Will change to == as I find more readable

bindings/azure/signalr/signalr.go
fbeltrao: Good catch, thanks @jjcollinge

bindings/azure/signalr/signalr.go
```diff
@@ -0,0 +1,194 @@
...
+	return url, nil
+}
+
+func (s *SignalR) sendMessageToSignalR(url string, token string, data []byte) error {
+	httpReq, err := http.NewRequest("POST", url, bytes.NewBuffer(data))
+	if err != nil {
+		return err
+	}
+
+	httpReq.Header.Add("Authorization", "Bearer "+token)
```
davidfowl: Set vs Add?

yaron2: In this case, it doesn't matter.

Here are the differences:

```
Set sets the header entries associated with key to the single element value. It replaces any existing values associated with key. The key is case insensitive; it is canonicalized by textproto.CanonicalMIMEHeaderKey. To use non-canonical keys, assign to the map directly.
```

```
Add adds the key, value pair to the header. It appends to any existing values associated with key. The key is case insensitive; it is canonicalized by CanonicalHeaderKey.
```

davidfowl: Let’s call set for both

yaron2: I missed the Set call. Should have consistency, so either set or add for both.

PR #158, opened 2019-12-20
Add sql server state

fbeltrao:
# Description

**SQL Server State Store**

This is an experimental implementation of a state based on SQL Server. 

## Why SQL Server?

**PROS**

- Operational features: backup, restore
- Database model that might be more appealing to some users in comparison to key/value and/or nosql models
- Rich ecosystem, querying capabilities

**CONS**

- No built to be a key/value store

## Implementation details

Data is stored in a table with the following definition:

```sql
CREATE TABLE [sampleappv3].[users](
	[Key] [nvarchar](200) NOT NULL,
	[Data] [nvarchar](max) NOT NULL,
	[InsertDate] [datetime2](7) NOT NULL,
	[UpdateDate] [datetime2](7) NULL,
	[RowVersion] [timestamp] NOT NULL
) 
```

As you can see the data is saved as text. With SQL Server we can promote properties from JSON to become columns, allowing indexing. For instance, adding favorite beverage and pets count:

```sql
CREATE TABLE [sampleappv3].[users](
	[Key] [nvarchar](200) NOT NULL,
	[Data] [nvarchar](max) NOT NULL,
	[InsertDate] [datetime2](7) NOT NULL,
	[UpdateDate] [datetime2](7) NULL,
	[FavoriteBeverage]  AS (CONVERT([nvarchar](100),json_value([Data],'$.FavoriteBeverage'))) PERSISTED,
	[PetsCount]  AS (CONVERT([int],json_value([Data],'$.PetsCount'))) PERSISTED,
	[RowVersion] [timestamp] NOT NULL
) 
```

The column values will be resolved by SQL Server at insert/update time.

### Key

Using the type `NVARCHAR(200)` as key in some scenarios is not ideal (integers or uuids). The experimental implementation has support for int or uniqueidentifiers as key. The implementation cannot be used yet because the current api is [adding the app.id as the key prefix](https://github.com/dapr/dapr/blob/master/pkg/http/api.go#L345).

### Bulk delete

Bulk delete is performed with the help of this stored procedure:

```sql
CREATE PROCEDURE [sampleappv3].[sp_BulkDelete_users] 
    @itemsToDelete [sampleappv3].users_Table READONLY
AS 
    DELETE [sampleappv3].[users]
    FROM [sampleappv3].[users] x
    JOIN @itemsToDelete i ON i.[Key] = x.[Key] AND (i.[RowVersion] IS NULL OR i.[RowVersion] = x.[RowVersion])
```

### Upsert

Inserts and updates are performed with the following stored procedure:

```sql
CREATE PROCEDURE [sampleappv3].[sp_Upsert_users] (
    @Key 			NVARCHAR(200),
    @Data 			NVARCHAR(MAX),
    @RowVersion 	BINARY(8))
AS
    IF (@RowVersion IS NOT NULL)
    BEGIN
        UPDATE [users]
        SET [Data]=@Data, UpdateDate=GETDATE()
        WHERE [Key]=@Key AND RowVersion = @RowVersion

        RETURN
    END
    
    BEGIN TRY
        INSERT INTO [users] ([Key], [Data]) VALUES (@Key, @Data);
    END TRY

    BEGIN CATCH
        IF ERROR_NUMBER() IN (2601, 2627) 
        UPDATE [users]
        SET [Data]=@Data, UpdateDate=GETDATE()
        WHERE [Key]=@Key AND RowVersion = ISNULL(@RowVersion, RowVersion)
    END CATCH
```

### State declaration

Simplest example:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: statestore
spec:
  type: state.azure.sqlserver
  metadata:
  - name: connectionString
    value: server=localhost;user id=sa;password=your-password;port=1433;database=dapr_test;
  - name: schema
    value: sampleappv3
  - name: tableName
    value: users
```

Example with property promotions

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: statestore
spec:
  type: state.azure.sqlserver
  metadata:
  - name: connectionString
    value: server=localhost;user id=sa;password=your-password;port=1433;database=dapr_test;
  - name: schema
    value: sampleappv3
  - name: tableName
    value: users
  - name: indexedProperties
    value: '[{ "column":"FavoriteBeverage", "property":"FavoriteBeverage", "type":"nvarchar(100)"}, { "column":"PetsCount", "property":"PetsCount", "type": "INTEGER"}]'
```

### Interesting features/needs further investigation

Features that might be interesting to experiment:

- Using memory optimized tables
- How to migrate different versions (i.e. adding new indexed property)

## Disclaimer

- No stress/load test has been performed

## Issue reference

Haven't created one. Can do if needed.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

## Questions
- In order to run the tests I need to have an image of mssql running. I modified the make file to start it locally. Another option is to start it programmatically using testcontainers (or something similar). Would appreciate guidance on it.
- Would it be possible to rethink the need to suffix the state key with the application id? For SQL Server it prevents the usage of native id types (unique identifier, int) as the key.


yaron2:
Can you remove or comment out the test that relies on SQL Server using Docker? our build agents do not have Docker installed and this fails the test..

fbeltrao:
@yaron2 I have skipped the integration tests. 

Two remarks:
- Is there plan in the future to add docker support in build agents to enable the integration tests?
- It would be great is to support different key data types, which is limited now to string due to the concatenation with the application-id.

yaron2:
> Is there plan in the future to add docker support in build agents to enable the integration tests?

That's a question for GitHub, really. If they do have this option, then yes, I think we'll want to enable that.

> It would be great is to support different key data types, which is limited now to string due to the concatenation with the application-id.

It's a string mostly due to the nature of KV stores which operate on string keys almost exclusively.

yaron2:
Does this PR have anything specific to Azure SQL Server or can this target any SQL Server instance?

state/azure/sqlserver/sqlserver.go
```diff
@@ -0,0 +1,527 @@
...
+
+// IndexedProperty defines a indexed property
+type IndexedProperty struct {
+	ColumnName string `json:"column"`
+	Property   string `json:"property"`
+	Type       string `json:"type"`
+}
+
+// StateStore defines a Sql Server based state store
+type StateStore struct {
```
yaron2: Can this be renamed to `SQLServer`?

fbeltrao: Yes, sure

fbeltrao:
> Does this PR have anything specific to Azure SQL Server or can this target any SQL Server instance?

It does work with Microsoft SQL Server (versions 2016+, as it requires user defined tables). Integration tests for instance, runs against a local server.

yaron2:
Is the key column currently configured as NVARCHAR?

fbeltrao:
> Is the key column currently configured as NVARCHAR?

By default is NVARCHAR(200). Would you prefer a different default?

yaron2:
Nope, that's fine.

I'm going to approve the PR, do you see any reason to not merge or are there any limitations on this other than what we've already discussed?

fbeltrao:
> One change request:
> 
> Every place that has `defer db.Close()` needs to have the defer line before returning err, right after the DB variable is getting assigned.

Hi @yaron2, all samples I have seen are doing it after checking that Open did not fail. Like [this doc](https://golang.org/pkg/database/sql/#Conn.ExecContext)

```golang
	// A *DB is a pool of connections. Call Conn to reserve a connection for
	// exclusive use.
	conn, err := db.Conn(ctx)
	if err != nil {
		log.Fatal(err)
	}
	defer conn.Close() // Return the connection to the pool.
```

Is that incorrect?

fbeltrao:
> Nope, that's fine.
> 
> I'm going to approve the PR, do you see any reason to not merge or are there any limitations on this other than what we've already discussed?

For SQL Server it would be nice to have the key without the owner id prefix, allowing other native types as key (INT, UNIQUEIDENTIFIER). But, it remains useful the way it is.

yaron2:
> > Nope, that's fine.
> > I'm going to approve the PR, do you see any reason to not merge or are there any limitations on this other than what we've already discussed?
> 
> For SQL Server it would be nice to have the key without the owner id prefix, allowing other native types as key (INT, UNIQUEIDENTIFIER). But, it remains useful the way it is.

We'll look into that regardless inside `dapr/dapr`. I was looking to see that it's still useful :)

yaron2:
> > One change request:
> > Every place that has `defer db.Close()` needs to have the defer line before returning err, right after the DB variable is getting assigned.
> 
> Hi @yaron2, all samples I have seen are doing it after checking that Open did not fail. Like [this doc](https://golang.org/pkg/database/sql/#Conn.ExecContext)
> 
> ```go
> 	// A *DB is a pool of connections. Call Conn to reserve a connection for
> 	// exclusive use.
> 	conn, err := db.Conn(ctx)
> 	if err != nil {
> 		log.Fatal(err)
> 	}
> 	defer conn.Close() // Return the connection to the pool.
> ```
> 
> Is that incorrect?

Actually it depends on the use case. Looking into the `Close()` function, looks like this ok OK behavior, so no changes needed.

yaron2:
Approved, will be merged soon.

I have to say this is a very high quality work and PR, well done.

Issue #178, opened 2020-01-07
Azure ServiceBus allow concurrent consumers on pubsub


jjcollinge:
As discussed in Gitter, the current azure servicebus pubsub implementation receives messages from topics one at a time in a serial fashion. This is a issue is similar to #171. 

I propose adding an optional `numConcurrentConsumers` setting which allows users to configure this.
By exposing this setting you can get synchronous behaviour by setting it to 1 or asynchronous by setting it to something higher?

yaron2:
That's a good issue to raise, thanks for this.

I think by default we should dispatch asynchronous and open a goroutine for every message received, and use  `numConcurrentConsumers` to limit.

yaron2:
Closed with #178

PR #179, opened 2020-01-07
Add concurrency to Azure ServiceBus pubsub

jjcollinge:
# Description

Updated the receive behaviour to support multiple consuming workers (gorountines).
We might need to consider exposing other behavioural toggles such as ASB sessions for FIFO, etc. 

__NOTE:__ I'm super tired and need to look at this with fresh eyes tomorrow!

## Issue reference
#178

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


pubsub/azure/servicebus/servicebus.go
```diff
@@ -84,6 +86,15 @@ func parseAzureServiceBusMetadata(meta pubsub.Metadata) (metadata, error) {
 		}
 	}
 
+	m.NumConcurrentConsumers = runtime.NumCPU()
```
yaron2: As mentioned in the issue, I think we should by default dispatch on a new goroutine for every message (instead of limiting to number of CPUs, as Go can schedule hundreds of thousands of Goroutines per core), and use `numConcurrentConsumers` to limit.

jjcollinge: Ok makes sense. The way the sdk is designed means a call to `Receive()` is blocking with an internal message pump which pushes the messages onto a channel which are then pulled and handled by a function you provide on a separate go routine. [Their example for handling concurrency](https://github.com/Azure/azure-service-bus-go/blob/1839bfc2a41a03ac130d55a1715188d3c0552f07/queue_examples_test.go#L113) works like this, where you create `n` listeners. I'll have a look at this tomorrow and see if we can handle this slightly differently to create a go routine per message. Welcome thoughts.

yaron2: Looking at that example, we might not need lines 125 - 136.

We can create a new goroutine from the handler, something like this: (very much pseudo code)

```go
var concurrentHandler servicebus.HandlerFunc = func(ctx context.Context, msg *servicebus.Message) error {
		go func() {
                      ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
		      defer cancel()

                         // send message to the user app
                       err := handler(msg)
                       if err == nil {
                          msg.Complete(ctx)
                      }
                 }()
		return nil
}
```

yaron2: Also, concurrency limit can be a global setting for Pub/Sub in Dapr instead of implemented individually.. see here:

https://github.com/dapr/dapr/issues/940#issuecomment-571789035

jjcollinge: Thanks for the feedback @yaron2 - interested to understand the "global setting" approach a bit more. Are you referring to adding a property to the `Config` struct in `dapr/runtime/config.go` that differs to `MaxConcurrency` such as `PubSubMaxConsumers` and then passing the runtime config into the pubsub in `initPubSub` , or did you have something else in mind?

pubsub/azure/servicebus/servicebus.go
```diff
@@ -200,11 +210,41 @@ func (a *azureServiceBus) getHandlerFunc(topic string, handler func(msg *pubsub.
...
+			// TODO: What should the correct time out on handling a message be?
+			ctx, cancel := context.WithTimeout(context.Background(), time.Second*time.Duration(a.metadata.TimeoutInSec))
+			defer cancel()
+
+			err := handlerFunc(ctx, msg)
+			if err != nil {
+				log.Errorf("%s error handling message from topic '%s', %s", errorMessagePrefix, topic, err)
+			}
+
+			if limitConcurrency {
```
jjcollinge: TODO: Move to defer

pubsub/azure/servicebus/servicebus.go
yaron2: It'll be another `spec.metadata` property on Pub Sub components, named `concurrencyLimit`.
The runtime will then enforce it in `daprd` for every pub/sub implementation.

yaron2: So my suggestion is to remove concurrency limits completely from the Azure Service bus code.

jjcollinge: Is it best to close this PR then?

yaron2: Well, looking into the ASB implementation now, unfortunately we should keep this PR, as ASB waits for the handler function to finish before executing a new handler, so it itself enforces concurrency, meaning the only way to allow for more than one execution is to create N goroutines.

Concurrency limit will still be handled in the Dapr runtime as I outlined above, but that's under the expectation that individual implementations by default do not enforce concurrency.

The question now is how to calculate the default concurrency limit.
Since in Go gourtines != threads and the Go scheduler is smart enough to grow/shrink the stack size and reallocate routines on threads, I think we can safely start with sane defaults of 1000 goroutines per CPU core.

yaron2: I would still leave the metadata property for an option for power users.

jjcollinge: Just to clarify - for the ASB impl we keep an additional metadata property (numConcurrentConsumers or equivalent) and use a default of 1000 per core. There will be an additional mechanism in the runtime that will use global property limitConcurrency to limit other pubsubs but will be ignored by this impl. Feels like this will need to be very clear in the docs to stop people tripping up.

yaron2: The Dapr runtime expects implementations to fire as many events in parallel as they can, with respects to individual configurations.

The Dapr runtime setting to limit will always take the rate of incoming events from a pub/sub components and limit it, regardless of what happens in the implementation itself.

The components are unaware of what the Dapr runtimes does or not. the `numConcurrentConsumers` in ASB is not to limit concurrency, its essentially there to work around the fact that the ASB itself is serially executing a handler message per worker.

jjcollinge: Ok, so in the current implementation - the handler is effectively async (which bypasses error handling etc. in the sdk) but the message iterator is serial. Is it best we revert to this PRs original implementation which ran 'n' (in this case we think 1000 per core) workers each which processing messages serially? Maybe rename `numConcurrentConsumers` to `numListeners` and then leave the throttling to the Dapr runtime.

jjcollinge: Ok so; `limitConcurrency` in runtime will effectively throttle the number of concurrent requests dapr makes to your app. `numConcurrentConsumers` would effectively be a limit on the number of concurrent handlers (and thus goroutines) running in the pubsub component. Is that about right? Maybe renamed `numConcurrentConsumers` to `numConcurrentHandlers`. We leave that without a default (leaving it to the go runtime) and allow power users to set it to effectively control the number of goroutines the pubsub component itself can spin up.

yaron2: That is correct. +1 on suggested property names.

jjcollinge: ok great I'll resolve this thread - any additional comments can be made on a new review

yaron2:
lgtm.

Are we good to merge?

jjcollinge:
> 
> 
> lgtm.
> 
> Are we good to merge?

I don't have any additional changes to add to this PR. If there's no more comments then I think we're good to merge.

Issue #182, opened 2020-01-09
Modify the AppResponse struct to provide the store name along with state requests


shalabhms:
Need to change the  AppResponse struct to have the store name and state requests separately to support multiple state stores support in Dapr.

This is needed to support this requirement : https://github.com/dapr/dapr/issues/636

PR #183, opened 2020-01-09
Modify the AppResponse struct to have the store name 

shalabhms:
Need to change the AppResponse struct to have the store name and state requests separately to support multiple state stores support in Dapr.

This is needed to support this requirement : dapr/dapr#636

# Description

Added the store name in the AppResponse and requests separately.

## Issue reference

#182 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/bindings/responses.go
+++ b/bindings/responses.go
@@ -19,6 +19,7 @@ type ReadResponse struct {
 type AppResponse struct {
 	Data        interface{}        `json:"data"`
 	To          []string           `json:"to"`
-	State       []state.SetRequest `json:"state"`
+	StoreName   string             `json:"storeName"`
+	Requests    []state.SetRequest `json:"requests"`
 	Concurrency string             `json:"concurrency"`
 }
--- a/bindings/responses.go
+++ b/bindings/responses.go
@@ -17,9 +17,14 @@ type ReadResponse struct {
 
 // AppResponse is the object describing the response from user code after a bindings event
 type AppResponse struct {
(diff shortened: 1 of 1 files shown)


bindings/responses.go
```diff
@@ -19,6 +19,7 @@ type ReadResponse struct {
 type AppResponse struct {
 	Data        interface{}        `json:"data"`
 	To          []string           `json:"to"`
-	State       []state.SetRequest `json:"state"`
+	StoreName   string             `json:"storeName"`
+	Requests    []state.SetRequest `json:"requests"`
```
artursouza: There is no reason to change this, right? Also, this is seems to be a response object, so calling it requests reads odd (IMO). Can you just add the State attribute and be done?

shalabhms: Though this is app response, but that can have state request to store after binding event. 
I did not get the State attribute part as how to achieve through the attribute..

bindings/responses.go
```diff
@@ -17,8 +17,14 @@ type ReadResponse struct {
 
 // AppResponse is the object describing the response from user code after a bindings event
 type AppResponse struct {
```
artursouza: My comment was to do just this:
```
type AppResponse struct {
	Data        interface{}        `json:"data"`
	To          []string           `json:"to"`
	StoreName   string             `json:"storeName"`
	State       []state.SetRequest `json:"state"`
        Concurrency string             `json:"concurrency"`
}
```

Issue #192, opened 2020-01-16
Proposal: Rate Limiting Middleware


jjcollinge:
## Proposal: Rate Limiting Middleware
Add a middleware that can throttle the rate of requests.

Potential rate limiters to leverage:

- https://godoc.org/golang.org/x/time/rate
- https://github.com/juju/ratelimit
- https://github.com/uber-go/ratelimit

We'd probably want to support 2 modes of operation, one that throws a 429 and expects the user to retry and one that blocks until a token is available.

yaron2:
Dapr supports rate limiting:
https://github.com/dapr/docs/blob/master/howto/control-concurrency/README.md

However it's interesting to move that logic into a middleware.

yaron2:
Ok, so the above rate limits the user app, I gather this is to rate limit Dapr?

jjcollinge:
Sorry, I completely missed the existing rate limiting - I just did a quick search and didn't find anything so proposed this 🤦‍♂️. If you think having a middleware option is of value then we can pursue this a bit more - if not feel free to close this!

jjcollinge:
Although looking at the existing control I think the middleware could be used to compliment or extend it. Like you say the middleware would allow you to throttle requests coming into Dapr. We could look to isolate it to just Dapr ingress and then you could use concurrency controls to control appplication ingress. Or something to that effect.

jjcollinge:
A bug in Dapr https://github.com/dapr/dapr/issues/948 made me think the middleware behaved differently to how I expected, once that's been fixed the middleware will only be invoked once per client call. Therefore we can use this to throttle client and app calls into Dapr. Whereas the existing concurrency control is just Dapr calling the app. Hope that makes sense.

PR #193, opened 2020-01-16
Add Rate Limiting Middleware

jjcollinge:
# Description

A middleware that can be used to rate limit the number of requests passing through the HTTP pipeline.

Example component
```yaml
 apiVersion: dapr.io/v1alpha1
 kind: Component
 metadata:
   name: ratelimit
 spec:
   type: middleware.http.ratelimit
   metadata:
   - name: maxRequestsPerSecond
     value: 10
```
Example configuration
```yaml
 apiVersion: dapr.io/v1alpha1
 kind: Configuration
 metadata:
   name: pipeline
 spec:
   httpPipeline:
     handlers:
     - name: ratelimit
       type: middleware.http.ratelimit
```

_NOTE:_ The middleware will be invoked on any HTTP ingress request to Dapr and therefore this will also throttle the user application's calls.

## Issue reference

Please reference the issue this PR will close: #192 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


jjcollinge:
Done some initial very rough testing: 
```
MaxRequestsPerSecond=10
Block=false
---
Iter 1
---
Total: 1000
Successful: 109
Rate Limited: 891
Failed: 0

real    0m9.863s
user    0m4.233s
sys     0m1.163s
---
Iter 2
---
Total: 1000
Successful: 108
Rate Limited: 892
Failed: 0

real    0m9.840s
user    0m4.436s
sys     0m1.043s
---
Iter 3
---
Total: 1000
Successful: 109
Rate Limited: 891
Failed: 0

real    0m9.897s
user    0m4.278s
sys     0m1.189s
```
This is roughly about what I'd expect given the test is very primitive.
1000 in 10 seconds = 100 rps
We're throttling at 10 rps so would expect 10% (100 requests) to be successful.

go.sum
```diff
@@ -526,6 +530,8 @@ go.uber.org/atomic v1.4.0/go.mod h1:gD2HeocX3+yG+ygLZcrzQJaqmWj9AIm7n08wl/qW/PE=
 go.uber.org/multierr v1.1.0/go.mod h1:wR5kodmAFQ0UK8QlbwjlSNy0Z68gJhDJUG5sjR94q/0=
 go.uber.org/multierr v1.2.0 h1:6I+W7f5VwC5SV9dNrZ3qXrDB9mD0dyGOi/ZJmYw03T4=
 go.uber.org/multierr v1.2.0/go.mod h1:wR5kodmAFQ0UK8QlbwjlSNy0Z68gJhDJUG5sjR94q/0=
+go.uber.org/ratelimit v0.1.0 h1:U2AruXqeTb4Eh9sYQSTrMhH8Cb7M0Ian2ibBOnBcnAw=
```
youngbupark: does juju/ratelimit use uber's ratelimiter? 

I have a concern about the license of `juju/ratelimit`. it is licensed under `LGPL v3 License`.

youngbupark: @yaron2 @amanbha any issues to use LGPL v3 licence module ?

jjcollinge: @youngbupark it doesn't use uber's ratelimiter (AFAIK that only does blocking rate limiting) but definitely worth checking the license obligations of LGPL v3.

youngbupark: Normally, (at least Microsoft), production team normally excludes any LGPL/GPL licensed OSS. if it is required, it may require the review with LCA. So we need to be careful here.

jjcollinge: It’s a good catch - I will remove the reference to Uber/ratelimit anyway (I was testing it and left it in). If we can’t use juju ratelimit, then I could try alternatives like https://github.com/didip/tollbooth

jjcollinge:
Ok migrating from [juju/ratelimit](https://github.com/juju/ratelimit) (LGPLv3) to [tollbooth](https://github.com/didip/tollbooth) (MIT). No longer support blocking rate limiting, so only 429 behaviour.

```
MaxRequestsPerSecond=10
---
Iter 1
---
requests: 1000
success: 116
failed: 0
limited: 884

real    0m10.708s
user    0m4.360s
sys     0m1.650s
---
Iter 2
---
requests: 1000
success: 108
failed: 0
limited: 892

real    0m9.849s
user    0m4.219s
sys     0m1.088s
---
Iter 3
---
requests: 1000
success: 117
failed: 0
limited: 883

real    0m10.770s
user    0m4.687s
sys     0m1.375s
```

Again this test is not run under strict conditions so is just a very rough estimate of whether it behaves as expected rather than the accuracy of the rate limit.

Also, just a note I've had to introduce the same function used in the [Prometheus WIP work](https://github.com/dapr/dapr/pull/962) to map from fasthttp handlers to net/http handlers. 

cc: @youngbupark

middleware/http/ratelimit/ratelimit_middleware.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+		
+		h(&c)
+
+		c.Response.Header.VisitAll(func(k []byte, v []byte) {
+			w.Header().Add(string(k), string(v))
+		})
+		c.Response.BodyWriteTo(w)
+	})
+}
```
youngbupark: Can you please add unit-test for this ? I think we can create the fake Request and Response to test this middleware.

jjcollinge: Yes, I'm currently writing some tests for this (although I will probably leave it for next week now).

jjcollinge:
Worth noting the adaptor isn't zero cost - I'll do some profiling and see if we can understand the overhead a bit better.

yaron2:
> Worth noting the adaptor isn't zero cost - I'll do some profiling and see if we can understand the overhead a bit better.

That's important to figure out..

jjcollinge:
Moving the handle wrapper outside the inner handler has made the throttling a lot more consistent.

```
MaxRequestsPerSecond=10
---
Iter 1
---
requests: 1000
success: 10
failed: 0
limited: 990

real    0m9.861s
user    0m4.294s
sys     0m1.248s
---
Iter 2
---
requests: 1000
success: 10
failed: 0
limited: 990

real    0m10.035s
user    0m4.316s
sys     0m1.400s
---
Iter 3
---
requests: 1000
success: 10
failed: 0
limited: 990

real    0m9.782s
user    0m4.173s
sys     0m1.253s
```
I'll try do some profiling later today.

jjcollinge:
This is a CPU profile whilst running with ratelimiting at 10 rps, whilst receiving 100 rps. Do you need anything else to move this PR forward?
dapr_07022020_cpu
[CPU Profile PDF](https://github.com/dapr/components-contrib/files/4172084/dapr_07022020_cpu.pdf)

yaron2:
Yeah looks fine. The build tests are failing

jjcollinge:
Fixed

yaron2:
> Fixed, do you think we need any additional test coverage?

lgtm

PR #200, opened 2020-01-29
Add CloudState CRDT

yaron2:
This PR adds CloudState integration as a state store, allowing for fast low latency in-memory replicated cache.


yaron2:
Done with the wonderful help of @jroper @viktorklang from CloudState!

PR #185, opened 2020-01-13
[WIP] Adding support for Manage Identities in the Azure Key Vault Secret Store

cmendible:
# Description

Adding support for Manage Identities in the Azure Key Vault Secret Store

Code in _authutils.go_ now extends _keyvaultSecretStore_ in such a way that calling the _GetAuthorizer()_ method result in using a client certificate to call the key vault if all the needed configuration is provided or fall back to using Managed Identities pretty much like the Azure SDK does when calling _GetAuthorizer_  in the autorest code found [here](https://github.com/Azure/go-autorest/blob/master/autorest/azure/auth/auth.go)

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #184

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

If you think this is going in the right direction then I'll be glad to finish the implementation, update the tests and extend the documentation.

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Extended the documentation

PR #206, opened 2020-02-04
Fixed issue preventing authentication fallback to MSI

cmendible:
# Description

Wrong condition was preventing the Key Vault secret store to fallback to MSI authentication when needed.

* [ x] Code compiles correctly
* [ x] Created/updated tests

### DIFF
--- a/secretstores/azure/keyvault/authutils.go
+++ b/secretstores/azure/keyvault/authutils.go
@@ -31,17 +31,17 @@ type CertConfig struct {
 // GetClientCert creates a config object from the available certificate credentials.
 // An error is returned if no certificate credentials are available.
 func (s EnvironmentSettings) GetClientCert() (CertConfig, error) {
-	certFilePath := s.Values[componentSPNCertificateFile]
-	certBytes := []byte(s.Values[componentSPNCertificate])
+	certFilePath, certFilePathPresent := s.Values[componentSPNCertificateFile]
+	certBytes, certBytesPresent := s.Values[componentSPNCertificate]
 	certPassword := s.Values[componentSPNCertificatePassword]
 	clientID := s.Values[componentSPNClientID]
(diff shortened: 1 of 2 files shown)


cmendible:
Hi @shalabhms, @yaron2 I was testing edge builds in a k8s cluster to add the docs requested in #185 and the key vault secret store was not falling back to MSI cause of a wrong if statement. This PR fixes that.

secretstores/azure/keyvault/authutils.go
```diff
@@ -37,7 +37,7 @@ func (s EnvironmentSettings) GetClientCert() (CertConfig, error) {
 	clientID := s.Values[componentSPNClientID]
 	tenantID := s.Values[componentSPNTenantID]
 
-	if certFilePath == "" && certBytes == nil {
```
youngbupark: can `s.Values[componentSPNCertificate]` be nil?  then certBytes is always zero byte array?

cmendible: `Values` is `map[string]string` which does not allows for null and returns an empty string when the given key is not present.

Guess a better check would be:

``` 
        certBytes, bytesPresent := s.Values[componentSPNCertificate]
	certPassword := s.Values[componentSPNCertificatePassword]
	clientID := s.Values[componentSPNClientID]
	tenantID := s.Values[componentSPNTenantID]

	if certFilePath == "" && !bytesPresent {
		return CertConfig{}, fmt.Errorf("missing client secret")
	}
```

thoughts?

youngbupark: looks good to me :)

cmendible: Fixed in f3e67d1

youngbupark:
@cmendible Thanks for your contribution!

PR #216, opened 2020-02-10
Add params for mongoDB

### LINKED ISSUE #215: Support additional params for the MongoDB state component
## Describe the proposal

The [current MongoDB state component does not support additional parameters](https://github.com/dapr/components-contrib/blob/master/state/mongodb/mongodb.go#L259).

For some (managed) MongoDB databases, it is required to provide them.

For example for Azure CosmosDB with MongoAPI the connection string looks like ```[name].mongo.cosmos.azure.com:10255/?ssl=true&replicaSet=globaldb&maxIdleTimeMS=120000&appName=@[name]@```

Would be great if these additional params can be supplied.

yaron2:
Closes #215

### DIFF
--- a/state/mongodb/mongodb.go
+++ b/state/mongodb/mongodb.go
@@ -34,6 +34,7 @@ const (
 	writeConcern     = "writeConcern"
 	readConcern      = "readConcern"
 	operationTimeout = "operationTimeout"
+	params           = "params"
 	id               = "_id"
 	value            = "value"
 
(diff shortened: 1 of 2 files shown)


state/mongodb/mongodb.go
```diff
@@ -256,11 +258,15 @@ func (m *MongoDB) doTransaction(sessCtx mongo.SessionContext, operations []state
...
 }
 
+func getMongoURI(metadata *mongoDBMetadata) string {
+	return fmt.Sprintf(connectionURIFormat, metadata.username, metadata.password, metadata.host, metadata.databaseName, metadata.params)
+}
+
 func getMongoDBClient(metadata *mongoDBMetadata) (*mongo.Client, error) {
 	var uri string
 
 	if metadata.username != "" && metadata.password != "" {
```
amanbha: If username and password is empty (they can be optional), the url will not be generated.

yaron2: Correct. That was the logic in the original implementation and I didn't want to change it.

@shalabhms can advise here.

PR #221, opened 2020-02-12
Reduce Redis Streams block time

### LINKED ISSUE #220: Remove block for Redis Streams
As reported by @MohsenM, Redis Streams does not contain a non block attribute and will thus timeout after the connection pool has timed out, as reported [here](https://github.com/go-redis/redis/issues/1002).

yaron2:
This PR adds a `100ms` block timeout to Redis Streams to reduce blocking time between XRead commands without hammering Redis.

Closes #220


yaron2:
Actually, I'm going to take the chance to also update the `redis-go` pkg. I'll close this PR in favor of a new one.

Issue #235, opened 2020-02-27
Use the same logger pkg of dapr


youngbupark:
## Describe the proposal

Use the same logger pkg of dapr.

PR #236, opened 2020-02-27
Use Dapr logger

youngbupark:
# Description

Use Dapr logger

## Issue reference

#235 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


yaron2:
I think we should inject `Logger` to the `Init()` method and thus save all the duplicate work of `NewLogger` in each components individually.

PR #249, opened 2020-03-05
extend envelop to include a subject field

Haishi2016:
This extention is to allow carrying correlation ID in cloud event envelop.


LMWF:
Does the proto need an update too?

Haishi2016:
> Does the proto need an update too?

This is not a protobuf type

yaron2:
@youngbupark is GH drunk again?

Haishi2016:
> @youngbupark is GH drunk again?

No, my bad, didn't update test cases

Issue #232, opened 2020-02-24
Azure Service Bus Component - Add "decode base64 message" option to input bindings


subesokun:
## Describe the proposal

When connecting an Azure service bus queue or Azure storage queue (maybe also event hub) to a storage account to receive e.g. the blob created events, all queue messages/events are base64 encoded. This will cause an error in the application in case the POST route handler applies an automated JSON parsing of the body (because the headers have the content-type `application/json`) so a special middleware has to be added to decode the message before it gets parsed.

Instead, it'd be great if there is an option like "decode base64 message" so that the application doesn't need to implement the decoding logic.

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: <name>
spec:
  type: bindings.azure.servicebusqueues
  metadata:
  - name: connectionString
    value: https://***
  - name: queueName
    value: queue1
  - name: decodeMessage
    value: true
```

artursouza:
@youngbupark @yaron2 We can start by implementing this in the specific component mentioned here. If this proves to be a common need for other components, then it might be worth to bake it into the runtime.

yaron2:
I agree.

CarlosLanderas:
Hello @subesokun!
As I work quite a lot with Azure storage and Service bus I wanted to recreate your environment to contribute with a PR and add the decodeMessage feature.

Could you add some information about the steps you are following? I created and Azure event grid subscription from Azure storage with blob events redirected to Azure service bus queues and it is not base64 encoded.

Are you using az functions?

CarlosLanderas:
I have this change working in a local branch, however there is something I would like to ask. The dapr api method onOutputBindingMessage, is marshalling the data request payload to []uint8 and it is including the string quotes in the target data message:

https://github.com/dapr/dapr/blob/master/pkg/http/api.go#L254

When you try to decode the base64 it is invalid so you need to trim the string payload:

```go
if a.metadata.DecodeBase64 == "true" {
			if decoded, err := decodeBase64Data(msg.Data); err == nil {
				msg.Data = decoded
```

```go
func decodeBase64Data(data []byte) ([]byte, error) {
	data = bytes.Trim(data, "\"") //Need to trim the byte array
	decoded := make([]byte, base64.StdEncoding.DecodedLen(len(data)))
	_, err := base64.StdEncoding.Decode(decoded, data)

	if err != nil {
		return nil, err
	}

	return decoded, nil
}
```
[...]

PR #251, opened 2020-03-05
[Azure Storage Queue Binding] Add option to decode base64 payloads #232

CarlosLanderas:
# Description

Allow Azure Storage Queues binding to decode base64 payloads.

If you create and event trigger from an storage account (blob created, deleted, directory created, deleted) you receive base64 payloads.

By configuring the component option decodeBase64 to "true" the component binding will decode the payload for you:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: mybindingname
spec:
  type: bindings.azure.storagequeues
  metadata:
  - name: storageAccount
    value: name
  - name: storageAccessKey
    value: key
  - name: queue
    value: test1
  - name: decodeBase64
    value: "true"
```
Issue #232

## Checklist

* [X] Code compiles correctly

PD: I want to create the Azure Storage Queue binding documentation that is missing, but I am waiting to see if this is merged to add the documentation updated with the change


CarlosLanderas:
Let me fix tests and add a new one

CarlosLanderas:
@yaron2 the actual existing storagequeues are not working.

```go
func (m *MockHelper) Read(ctx context.Context, consumer *consumer) error {
	return nil
}
```

Is not calling the consumer callback, so the assertions inside all the handlers are not being executed, the tests sleeps for a while and passes.

I've fixed the tests and added a new one with the included logic but none of them are being executed.

Maybe you already knew it.

Thanks

bindings/azure/storagequeues/storagequeues.go
```diff
@@ -128,7 +146,12 @@ func (a *AzureStorageQueues) Init(metadata bindings.Metadata) error {
 	}
 	a.metadata = meta
 
-	err = a.helper.Init(a.metadata.AccountName, a.metadata.AccountKey, a.metadata.QueueName)
+	decodeBase64 := false
+	if a.metadata.DecodeBase64 == "true" {
```
artursouza: Can this be a boolean directly from Yaml?

https://yaml.org/type/bool.html

CarlosLanderas: Answered here: 

https://github.com/dapr/components-contrib/pull/251#pullrequestreview-369620175

CarlosLanderas:
@artursouza, extending a little bit more my previous comment, it looks like the same signature is used in the runtime repository:

https://github.com/dapr/dapr/blob/d8aa4008346735aa8e4746626c0a7a5eb4e9a9fa/pkg/runtime/runtime.go#L637

https://github.com/dapr/dapr/blob/d17e9243b308e830649b0bf3af5f6e84fd543baf/tests/config/dapr_redis_state.yaml#L18

bindings/azure/storagequeues/storagequeues.go
yaron2: > > Only a small question.
> 
> Hello @artursouza . I tried that in the first place but the properties metadata signature is:
> 
> ```go
> type Metadata struct {
> 	Name       string
> 	Properties map[string]string `json:"properties"`
> }
> ```
> 
> So you get unmarshal error because of the map[string] string signature:
> 
> https://github.com/dapr/components-contrib/blob/master/bindings/metadata.go#L11
> 
> I've seen that other components are doing it the same way, maybe because of the properties field signature:
> 
> https://github.com/dapr/components-contrib/blob/ee97c3c56ada888e0740497131c2161f6a0a5539/secretstores/hashicorp/vault/vault.go#L121
> 
> cc @yaron2

That's correct, it's a `string[string]` map and thus has to be converted;

CarlosLanderas: yaron2. Then I think this is ok. I saw you already approved it.  Thanks!

Issue #243, opened 2020-03-02
Hazelcast PubSub Support


mesutcelik:
## In what area(s)?

> /area runtime

## Describe the feature

Hazelcast has a feature called Hazelcast Topic for pub/sub messaging model and async communication. That can be a good candidate for Dapr Pubsub Integration.
https://docs.hazelcast.org/docs/4.0/manual/html-single/#topic

Current list of PubSub Modules can be found here.
https://github.com/dapr/components-contrib/tree/master/pubsub

youngbupark:
moving to component-contribs

peterjot:
Could someone assign me to that issue?

youngbupark:
> Could someone assign me to that issue?

Thanks in advance. let me assign it to you.

PR #254, opened 2020-03-06
Add hazelcast pubsub support

peterjot:
# Description

Hazelcast pubsub support

## Issue reference

Issue https://github.com/dapr/components-contrib/issues/243

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/pubsub/hazelcast/hazelcast.go
+++ b/pubsub/hazelcast/hazelcast.go
@@ -1,0 +1,102 @@
+package hazelcast
+
+import (
+	"errors"
+	"fmt"
+	"github.com/dapr/components-contrib/pubsub"
+	"github.com/dapr/dapr/pkg/logger"
+	"github.com/hazelcast/hazelcast-go-client"
+	hazelcastCore "github.com/hazelcast/hazelcast-go-client/core"
+	"strings"
+)
+
+const (
+	hazelcastServers = "hazelcastServers"
+)
+
+type Hazelcast struct {
+	client hazelcast.Client
+	logger logger.Logger
+}
+
+// NewHazelcastPubSub returns a new hazelcast pub-sub implementation
+func NewHazelcastPubSub(logger logger.Logger) pubsub.PubSub {
+	return &Hazelcast{logger: logger}
+}
+
+func parseHazelcastMetadata(meta pubsub.Metadata) (metadata, error) {
+	m := metadata{}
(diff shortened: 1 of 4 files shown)


mesutcelik:
Thanks @peterjot ! I see you should update the [README](https://github.com/dapr/components-contrib/blob/master/pubsub/Readme.md) too.

Hi @yaron2 @youngbupark ,
Can we please get reviews from Dapr Team? It would be appreciate to see feedback from you guys!

yaron2:
Sure, we'll review. 

This is a great PR, thanks @peterjot.
We are now working towards the `0.5.0`. This PR will be part of `0.6.0`!

pubsub/hazelcast/metadata.go
```diff
@@ -0,0 +1,5 @@
+package hazelcast
+
+type metadata struct {
+	hazelcastServers string
```
artursouza: How is authentication done in hazelcast? The other components have some authentication mechanism but here there is only a list of servers.

pubsub/hazelcast/hazelcast.go
```diff
@@ -0,0 +1,108 @@
...
+	pubsubHandler func(msg *pubsub.NewMessage) error
+}
+
+func (l *hazelcastMessageListener) OnMessage(message hazelcastCore.Message) error {
+	var msg []byte
+	m, ok := message.MessageObject().([]byte)
+	if ok {
+		msg = m
+	} else {
+		msg, _ = json.Marshal(message.MessageObject())
```
artursouza: Why do you need this fallback to parse the message? I would expect for all messages to have the save structure.

pubsub/hazelcast/metadata.go
peterjot: Same as in the existing Hazelcast integration for state. Here https://github.com/dapr/components-contrib/blob/master/state/hazelcast/hazelcast.go#L48 we use default authentication credentials.

pubsub/hazelcast/hazelcast.go
peterjot: I looked a second time. You right, I actually dont need it

yaron2: @peterjot once you make that change, we can merge this PR. otherwise it looks good to me.

Issue #268, opened 2020-03-10
Enable the Azure Event hubs binding to be able to read/write from a single partition


msfussell:
There are scenarios in Azure Events hubs where you need to read from a particular partition. Currently the EventHubs binding reads from all partitions and then forwards the message. 

This issue to update the binding here https://github.com/dapr/docs/blob/master/reference/specs/bindings/eventhubs.md to enable reads and write to a particular partition for an Event Hub instance.

ksivamuthu:
@msfussell @yaron2  Can you please assign this to me, I'll work on this.

yaron2:
Thanks!

PR #269, opened 2020-03-11
Updated EventHub bindings to read and write from same partition

ksivamuthu:
# Description

### Write to Partition
Added `partitionID` and `partitionKey` fields in metadata.

#### PartitionID
If the partitionID is configured in metadata,  the eventhub is created with PartitionedSender options with the configured partitionID. `Write` method will send the event to the configured partition id.

#### PartitionKey
If the partitionKey is configured in metadata, or from the request metadata, the partitionKey is set to the event object.

```
event := &eventhub.Event{
	Data: req.Data,
}
...
event.PartitionKey = &partitionKey
...
err := a.hub.Send(context.Background(), event)
```

### Read from partition

If the partitionID is configured in metadata, setting up the event receiver on the configured partition. I see the EPH implementation is done in existing binding. 

I created a new partitioned receiver implementation without using EPH. 

From what I've researched, there is no option to claim ownership of the specific partition by EPH available in `azure-eventhubs-go` SDK similar to NodeJS SDK. 

https://github.com/Azure/azure-sdk-for-js/blob/98644cada2d8949f103dbf1c0ad4098a0adc31dc/sdk/eventhub/event-hubs/src/eventHubConsumerClient.ts#L418-L422

Please let me know if there is a better way to implement it.

## Issue reference

Please reference the issue this PR will close: #268

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/bindings/azure/eventhubs/eventhubs.go
+++ b/bindings/azure/eventhubs/eventhubs.go
@@ -32,6 +32,10 @@ const (
 	storageAccountKey    = "storageAccountKey"
 	storageContainerName = "storageContainerName"
 
+	// optional
+	partitionKeyName = "partitionKey"
+	partitionIDName  = "partitionID"
+
 	// errors
 	missingConnectionStringErrorMsg     = "error: connectionString is a required attribute"
 	missingStorageAccountNameErrorMsg   = "error: storageAccountName is a required attribute"
@@ -54,6 +58,12 @@ type azureEventHubsMetadata struct {
 	storageAccountName   string
 	storageAccountKey    string
 	storageContainerName string
+	partitionID          string
+	partitionKey         string
+}
+
+func (m azureEventHubsMetadata) partitioned() bool {
+	return m.partitionID != ""
 }
 
 // NewAzureEventHubs returns a new Azure Event hubs instance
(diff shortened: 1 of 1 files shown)


yaron2:
Looks good. Have you tested this in some way?

ksivamuthu:
I've tested by creating Azure eventhub with 4 partitions. 

Tested by creating different bindings for each partition and made sure the corresponding endpoint is called when publishing. Tested publishing by using two ways.

1. Publish events using dapr endpoint
2. Publish natively directly to eventhub to specific partitions

```
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: receive-event-1
spec:
  type: bindings.azure.eventhubs
  metadata:
  - name: connectionString 
    value: "CONN_STRING"
  - name: storageAccountName    # Azure Storage Account Name
    value: "STORAGE_ACCT_NAME"
  - name: storageAccountKey     # Azure Storage Account Key
    value: "STORAGE_ACCT_KEY"
  - name: storageContainerName  # Azure Storage Container Name
    value: "test"
  - name: consumerGroup
    value: "test"  
  - name: partitionID
    value: 0
```

yaron2:
LGTM.

This is great, thanks.
Can you please update the Event Hubs binding spec yaml [here](https://github.com/dapr/docs/blob/master/reference/specs/bindings/eventhubs.md)?

PR #271, opened 2020-03-14
Allow non-managed identity for keyvault

yaron2:
This PR brings back the removed support for non-managed identities for Azure Key Vault.

### DIFF
--- a/secretstores/azure/keyvault/keyvault.go
+++ b/secretstores/azure/keyvault/keyvault.go
@@ -13,6 +13,7 @@ import (
 	"github.com/dapr/dapr/pkg/logger"
 
 	kv "github.com/Azure/azure-sdk-for-go/profiles/latest/keyvault/keyvault"
+	"github.com/Azure/go-autorest/autorest"
 )
 
 // Keyvault secret store component metadata properties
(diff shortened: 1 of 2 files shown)


youngbupark:
Please revert test code as well.

secretstores/azure/keyvault/keyvault.go
```diff
@@ -47,13 +48,36 @@ func (k *keyvaultSecretStore) Init(metadata secretstores.Metadata) error {
...
-		k.vaultClient.Authorizer = authorizer
+	certFilePath := metadata.Properties[componentSPNCertificateFile]
+	certBytes := []byte(metadata.Properties[componentSPNCertificate])
+	certPassword := metadata.Properties[componentSPNCertificatePassword]
+	clientID := metadata.Properties[componentSPNClientID]
+	tenantID := metadata.Properties[componentSPNTenantID]
+
+	var authorizer autorest.Authorizer
+	var err error
+	if certFilePath != "" && len(certBytes) > 0 && certPassword != "" && clientID != "" && tenantID != "" {
```
youngbupark: I reviewed the code again. we do not need this code change.

I think this current component code should support both non-MSI mode and MSI mode. so this is a bug when refactoring the code. Let's review it next Monday. 

This condition is covered by this

https://github.com/dapr/components-contrib/blob/cf077cca65694b53088dc726603d625f28135faa/secretstores/azure/keyvault/authutils.go#L33-L46

and this:

https://github.com/dapr/components-contrib/blob/cf077cca65694b53088dc726603d625f28135faa/secretstores/azure/keyvault/authutils.go#L143-L149

yaron2:
After careful.check, this PR is not needed. Non MSI and MSI both work.

Issue #213, opened 2020-02-10
Support Redis with SSL, e.g. for Azure Managed Redis Cache


RicardoNiepel:
## Describe the proposal

Currently Azure Managed Redis Cache is only supported, if ["SSL only" is deactivated](https://github.com/dapr/docs/blob/master/howto/setup-state-store/setup-redis.md#creating-an-azure-managed-redis-cache) and a non-SSL connection will be used.

Would be great if it is possible to use also a SSL connection.

yaron2:
I just opened an issue to track it, but will close it in favor of this one. #214

yaron2:
Today the Redis state store is using https://github.com/joomcode/redispipe which does not support TLS client options.

I propose to move to https://github.com/go-redis/redis/ which supports TLS client options.

youngbupark:
@yaron2 shouldn't we ship this change in 0.4.0? 

cc/ @msfussell @artursouza

yaron2:
> @yaron2 shouldn't we ship this change in 0.4.0?
> 
> cc/ @msfussell @artursouza

No, this can wait for 0.5/0.6.

msfussell:
Yes, we can hold off on this for a while and review all the the asks for the next release.

RicardoNiepel:
@pruthvidhodda are you already working on this? Can I help somehow?

pruthvidhodda:
Already working on this, will reach out if I need help. Thanks for offering though.

RicardoNiepel:
@pruthvidhodda @yaron2 Please reopen.

See below what is still open:
- [ ] Enable TLS at [Redis Binding Component](https://github.com/dapr/components-contrib/blob/master/bindings/redis/redis.go)
  - [ ] Document it [here](https://github.com/dapr/docs/blob/master/reference/specs/bindings/redis.md)
- [X] Enable TLS at [Redis State Store Component](https://github.com/dapr/components-contrib/blob/master/state/redis/redis.go)
  - [ ] Document it [here](https://github.com/dapr/docs/blob/master/howto/setup-state-store/setup-redis.md#configuration)
- [X] Enable TLS at [Redis PubSub Component](https://github.com/dapr/components-contrib/blob/master/pubsub/redis/redis.go)
  - [ ] Document it [here](https://github.com/dapr/docs/blob/master/howto/setup-pub-sub-message-broker/setup-redis.md#configuration)

youngbupark:
[...]

PR #283, opened 2020-03-27
Change redis client library to go-redis for redis component

pruthvidhodda:
# Description

Changed the redis client library to go-redis for redis component

## Issue reference

Please reference the issue this PR will close: [#_[213]_](https://github.com/dapr/components-contrib/issues/213)

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


pruthvidhodda:
I haven't tested the changes but please go through these changes and give suggestions if you find anything wrong.

state/redis/redis.go
```diff
@@ -68,28 +69,28 @@ func (r *StateStore) Init(metadata state.Metadata) error {
...
+	var tlsConfig *tls.Config
+	err = json.Unmarshal(b, tlsConfig)
 	if err != nil {
 		return err
 	}
 
-	r.client = &redis.SyncCtx{
-		S: conn,
+	opts := redis.Options{
+		DB:        0,
```
amanbha: it would need host, specify it as:
` Addr: redisCreds.Host`

state/redis/redis.go
```diff
@@ -68,28 +69,28 @@ func (r *StateStore) Init(metadata state.Metadata) error {
 		return err
 	}
 
-	ctx := context.Background()
-	opts := redisconn.Opts{
-		DB:       0,
-		Password: redisCreds.Password,
-	}
-	conn, err := redisconn.Connect(ctx, redisCreds.Host, opts)
+	var tlsConfig *tls.Config
```
amanbha: read enableTLS from metadata and then use it here. Users would be able to set it in their components yaml file if they are talking to a secured redis endpoint with ssl.
Please see the usage in redis pubsub component:
https://github.com/dapr/components-contrib/blob/master/pubsub/redis/redis.go#L85

yaron2: +1 on this, also `tls.Config` will probably not unmarshall. 

To enable TLS if the user specified it on the YAML, its probably:

```
&tls.Config{
	InsecureSkipVerify: true,
}
```

As reported [here](https://github.com/go-redis/redis/issues/756).

state/redis/redis.go
shalabhms: +1. Also we need to consider taking MaxRetries and MaxBackoff configuration.

state/redis/redis.go
```diff
@@ -68,28 +69,28 @@ func (r *StateStore) Init(metadata state.Metadata) error {
...
-	r.client = &redis.SyncCtx{
-		S: conn,
+	opts := redis.Options{
+		DB:        0,
+		Password:  redisCreds.Password,
+		TLSConfig: tlsConfig,
 	}
 
+	r.client = redis.NewClient(&opts)
+
```
shalabhms: You might need to check for connection in init() through client.Ping().Result().

state/redis/redis.go
```diff
@@ -245,19 +246,20 @@ func (r *StateStore) BulkSet(req []state.SetRequest) error {
...
 func (r *StateStore) Multi(operations []state.TransactionalRequest) error {
-	redisReqs := []redis.Request{}
+
+	pipe := r.client.TxPipeline()
 	for _, o := range operations {
 		if o.Operation == state.Upsert {
 			req := o.Request.(state.SetRequest)
 			b, _ := r.json.Marshal(req.Value)
-			redisReqs = append(redisReqs, redis.Req("SET", req.Key, b))
+			pipe.Set(req.Key, b, 0)
```
shalabhms: Please avoid hardcoding, make it as defaultExpirationTime =0 and use it through variable .
Also consider reading the key expiration property from the request options, if it is not there use default..

yaron2:
Can you detail what is missing to remove WIP status?

state/redis/redis.go
pruthvidhodda: Oh right, I have added the host now. 



> +1. Also we need to consider taking MaxRetries and MaxBackoff configuration.

I have added those both now but how do we decide what all options should we consider?

state/redis/redis.go
pruthvidhodda: Thanks for the pointer, changed it now.

state/redis/redis.go
pruthvidhodda: Added the check now.

yaron2:
Also, let's not add any more configurable options in this PR as that can be done later. The most important thing is to replace what we have today with the new package for the purpose of TLS.

state/redis/redis.go
pruthvidhodda: (y) Added the constant now.

> Also consider reading the key expiration property from the request options, if it is not there use default..

I assume we need to add that to state.SetRequest struct first?

pruthvidhodda:
> Can you detail what is missing to remove WIP status?

As of now I have made all the changes except [this](https://github.com/dapr/components-contrib/pull/283#discussion_r399676931) one which is a configurable option so I believe we can skip that for now per your comment above.

What's left is to test it with a sample. Will remove WIP status once that's done.

state/redis/redis.go
yaron2: Lets not change `state.SetRequest` at this time.

amanbha: @pruthvidhodda I would suggest to get the existing functionality working with TLS and new redis package. More configs can be added later.

pruthvidhodda:
Verified the component with (@amanbha @youngbupark ):
i. local redis store **without** TLS
ii. azure redis cache **with** TLS

Please review the PR

state/redis/redis.go
shalabhms: Thats fine.

state/redis/redis.go
```diff
@@ -52,34 +55,77 @@ func NewRedisStateStore(logger logger.Logger) *StateStore {
...
+	m.enableTLS = false
+	if val, ok := meta.Properties[enableTLS]; ok && val != "" {
+		tls, err := strconv.ParseBool(val)
+		if err != nil {
+			return m, fmt.Errorf("redis store error: can't parse enableTLS field: %s", err)
+		}
+		m.enableTLS = tls
+	}
+
+	m.maxRetries = 3
```
shalabhms: Please create const DefaultMaxRetries = 3 , and set m.maxRetries = DefaultMaxRetries ..

state/redis/redis.go
```diff
@@ -52,34 +55,77 @@ func NewRedisStateStore(logger logger.Logger) *StateStore {
...
+	if val, ok := meta.Properties[maxRetries]; ok && val != "" {
+		parsedVal, err := strconv.ParseInt(val, defaultBase, defaultBitSize)
+		if err != nil {
+			return m, fmt.Errorf("redis store error: can't parse maxRetries field: %s", err)
+		}
+		m.maxRetries = int(parsedVal)
+	}
+
+	m.maxRetryBackoff = time.Second * 2
```
shalabhms: same above comment applicable here..

state/redis/redis.go
```diff
@@ -52,34 +55,77 @@ func NewRedisStateStore(logger logger.Logger) *StateStore {
...
+	m, err := parseRedisMetadata(metadata)
 	if err != nil {
 		return err
 	}
+	r.metadata = m
+
+	opts := &redis.Options{
+		Addr:            m.host,
+		Password:        m.password,
+		DB:              0,
```
shalabhms: consider this also moving to defaultDB = 0 , and assign here through variable..

state/redis/redis.go
```diff
@@ -7,43 +7,50 @@ package redis
 
 import (
 	"context"
-	"encoding/json"
+	"crypto/tls"
```
amanbha: Could you please include the integration test file as @youngbupark suggested (clean it up a bit and remove the credentials from it). It would serve as a guide for other contributors. If its going to take longer then please finish this PR and add that as a separate PR after this release.

pruthvidhodda: I will add it as a separate PR but will try to hit it for this release.

PR #290, opened 2020-04-01
Added Azure Event Grid Bindings

lynn-orrell:
# Description

Added Azure Event Grid bindings

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


bindings/azure/eventgrid/eventgrid.go
```diff
@@ -0,0 +1,186 @@
...
+					Data: bodyBytes,
+				})
+				if err != nil {
+					a.logger.Error(err.Error())
+				}
+			}
+		}
+	}
+
+	fasthttp.ListenAndServe(":8080", m)
```
RicardoNiepel, line 96: Should the port not be configurable?

bindings/azure/eventgrid/eventgrid.go
```diff
@@ -0,0 +1,186 @@
...
+				EndpointType: eventgrid.EndpointTypeWebHook,
+				WebHookEventSubscriptionDestinationProperties: &eventgrid.WebHookEventSubscriptionDestinationProperties{
+					EndpointURL: &a.metadata.SubscriberEndpoint,
+				},
+			},
+		},
+	}
+
+	a.logger.Debugf("Attempting to create or update Event Grid subscription. scope=%s endpointURL=%s", scope, a.metadata.SubscriberEndpoint)
+	result, err := subscriptionClient.CreateOrUpdate(context.Background(), scope, a.metadata.EventGridSubscriptionName, eventInfo)
```
RicardoNiepel, line 170: Should there not be WebHook ValidationCode handshake after that?
https://docs.microsoft.com/en-us/azure/event-grid/security-authentication#webhook-event-delivery

bindings/azure/eventgrid/eventgrid.go
lynn-orrell: Good catch, @RicardoNiepel! Meant to add this and missed it. Just added.

bindings/azure/eventgrid/eventgrid.go
lynn-orrell: No, that's the old way. We are using the CloudEvents method now which is what the OPTIONS callback is for... More information here: https://github.com/cloudevents/spec/blob/v1.0/http-webhook.md

RicardoNiepel: Thanks, learned something new - docs are outdated ....

RicardoNiepel:
/cc @yaron2 
> can the Dapr middleware be used for the webhook endpoint to use AAD auth for securing the Webhook endpoint?

Your thoughts on it?

lynn-orrell:
> can the Dapr middleware be used for the webhook endpoint to use AAD auth for securing the Webhook endpoint?
> 
> Your thoughts on it?

@RicardoNiepel, I believe the current OAuth middleware only support the Authorization Code Grant Flow (per the docs). I think it would need to also support the Client Credential Flow in order for this to be a possibility.

bindings/azure/eventgrid/eventgrid.go
```diff
@@ -0,0 +1,191 @@
...
+	a.logger.Debug("Read() called...")
+
+	err := a.createSubscription()
+	if err != nil {
+		return err
+	}
+
+	m := func(ctx *fasthttp.RequestCtx) {
+		switch string(ctx.Path()) {
+		case "/api/events":
```
artursouza, line 74: If this is the only case, it can be an if statement instead.

bindings/azure/eventgrid/eventgrid.go
```diff
@@ -0,0 +1,191 @@
...
+			if string(ctx.Method()) == "OPTIONS" {
```
artursouza, line 75: These two ifs may be a switch.

bindings/azure/eventgrid/eventgrid.go
```diff
@@ -0,0 +1,191 @@
...
+				ctx.Response.Header.Add("WebHook-Allowed-Origin", string(ctx.Request.Header.Peek("WebHook-Request-Origin")))
+				ctx.Response.Header.Add("WebHook-Allowed-Rate", "*")
+				ctx.Response.Header.SetStatusCode(fasthttp.StatusOK)
+				_, err = ctx.Response.BodyWriter().Write([]byte(""))
+				if err != nil {
+					a.logger.Error(err.Error())
```
artursouza, line 81: Does it mean we are swallowing errors?

bindings/azure/eventgrid/eventgrid.go
```diff
@@ -0,0 +1,191 @@
...
+	request.SetRequestURI(a.metadata.TopicEndpoint)
+	request.SetBody(req.Data)
+
+	response := fasthttp.AcquireResponse()
+	defer fasthttp.ReleaseResponse(response)
+
+	client := &fasthttp.Client{WriteTimeout: time.Second * 10}
+	err := client.Do(request, response)
+	if err != nil {
+		a.logger.Error(err.Error())
```
artursouza, line 121: Shouldn't this be handled just like the case for statuscode != OK?

bindings/azure/eventgrid/eventgrid.go
lynn-orrell: Unfortunately, I don't believe there is anything we can do here besides log the error.

bindings/azure/eventgrid/eventgrid.go
lynn-orrell: Now properly returning the error.

Issue #288, opened 2020-03-31
Support nested partition keys in cosmosdb binding


programaka:
Cosmosdb allows you to enter a nested partition key when creating a container (collection) in the portal.

For example: /headers/salesNumber
`{
    "headers": {
        "salesNumber": "ZCW666"
    }
}`

With Dapr you can only pass a top level property as a partition key.

## Proposal
Support nested partition keys.
Mention in the documentation whether nested partition keys are supported or not.

## Dapr version
v0.5.0

yaron2:
Thanks, that makes sense.

ksivamuthu:
@yaron2 @youngbupark Can you please assign this to me, if it is free to take up? Thanks.

youngbupark:
> @yaron2 @youngbupark Can you please assign this to me, if it is free to take up? Thanks.

Thanks in advance. This is yours :)

PR #296, opened 2020-04-05
Cosmos - Nested PartitionKey Support

ksivamuthu:
# Description

Read the property value from the request data using jsonpath to support nested partition keys.

Format for configuring nested partition key is

"address.zip" - fields separated by dot - similar to a .NET Bulk executor library

Example: 

Configuration: 
```
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: send-event
spec:
  type: bindings.azure.cosmosdb
  metadata:
  - name: url 
    value: URL
  - name: masterKey    # Azure Storage Account Name
    value: MASTER_KEY
  - name: database     # Azure Storage Account Key
    value: DB
  - name: collection  # Azure Storage Container Name
    value: CONTAINER
  - name: partitionKey
    value: "address.zip"
```

JSON: 
```
{
    "data": {
    	"id": "1",
        "name": "test",
        "address": {
             "zip": "35220"
        }
    }
}
```
image

## Issue reference

PR will close: #288 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/bindings/azure/cosmosdb/cosmosdb.go
+++ b/bindings/azure/cosmosdb/cosmosdb.go
@@ -8,11 +8,11 @@ package cosmosdb
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/a8m/documentdb"
 	"github.com/dapr/components-contrib/bindings"
 	"github.com/dapr/dapr/pkg/logger"
-	"github.com/yalp/jsonpath"
 )
 
 // CosmosDB allows performing state operations on collections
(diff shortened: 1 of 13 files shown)


bindings/azure/cosmosdb/cosmosdb.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/a8m/documentdb"
 	"github.com/dapr/components-contrib/bindings"
 	"github.com/dapr/dapr/pkg/logger"
+	"github.com/yalp/jsonpath"
```
yaron2, line 15: Is there a way to do this without using this specific package? it looks unmaintained.

yaron2: In general, we want as little external dependencies if possible..

ksivamuthu: That makes sense. I will programmatically parse the fields and navigate to the nested map.

ksivamuthu: Done the changes. Please review.

yaron2:
Thanks!

Issue #309, opened 2020-04-19
Expand the secret store implementation to provide a method to retrieve all secrets from a store


cmendible:
## Describe the proposal

Some use cases require to load all secrets from a store, so in order to simplify the customer code the `SecretStore`  interface should be expanded with a new GetSecrets method as a previous step before extendending the Dapr Secrets API:

``` golang
type SecretStore interface {
	// Init authenticates with the actual secret store and performs other init operation
	Init(metadata Metadata) error
	// GetSecret retrieves a secret using a key and returns a map of decrypted string/string values
	GetSecret(req GetSecretRequest) (GetSecretResponse, error)
        // GetSecrets retrieves all secrets in the store and returns a map of decrypted string/string values
	GetSecrets(req GetSecretsRequest) (GetSecretsResponse, error)
}
```

PR #310, opened 2020-04-20
Implementing GetSecrets

cmendible:
# Description

Expanding the secret store implementation providing a method to retrieve all secrets from a store

## Issue reference

Please reference the issue this PR will close: #309 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


secretstores/aws/secretmanager/secretmanager.go
```diff
@@ -86,6 +88,11 @@ func (s *smSecretStore) GetSecret(req secretstores.GetSecretRequest) (secretstor
 	return resp, nil
 }
 
+// GetSecrets retrieves all secrets in the store and returns a map of decrypted string/string values
+func (s *smSecretStore) GetSecrets(req secretstores.GetSecretsRequest) (secretstores.GetSecretsResponse, error) {
+	return secretstores.GetSecretsResponse{Data: nil}, status.Errorf(codes.Unimplemented, "method SaveState not implemented")
```
youngbupark, line 93: Do you have a reason why you use grpc status for error?

youngbupark (dismissed):
Thanks for your contribution!

secretstores/aws/secretmanager/secretmanager.go
cmendible: Obviously a copy & paste error. I'll push a fix.

yaron2:
@cmendible is there anything else planned for this PR?

cmendible:
> @cmendible is there anything else planned for this PR?

@yaron2 I still have to develop the `GetSecrets` method for Google's `secretmanager`

msfussell:
@cmendible - This PR has been here for a few months now. Any updates?

cmendible:
Hi  @msfussell sorry for the delay. I pushed some changes with the `GetSecrets` implementation for Google's `secretmanager`.

Implementation for Google's KMS is pending cause I noticed the component is not registered in daprd as a secret store [here](https://github.com/dapr/dapr/blob/fbfe2ebfbf2241a0c5953394d74b616a6c230faf/cmd/daprd/main.go#L136). Should I open an issue and PR for this?

cmendible:
PR https://github.com/dapr/components-contrib/pull/496 must be merged to rebase this one and finish the work.

secretstores/secret_store.go
```diff
@@ -11,4 +11,6 @@ type SecretStore interface {
 	Init(metadata Metadata) error
 	// GetSecret retrieves a secret using a key and returns a map of decrypted string/string values
 	GetSecret(req GetSecretRequest) (GetSecretResponse, error)
```
youngbupark, line 13: @yaron2 Do we still need single secret get in secret component? it seems like it is redundant.

yaron2: We have in Dapr a Single Get API and a Bulk Get API. We can change the implementation to use the Bulk and return the first item, I guess.

cmendible: `GetSecret` performance will always be better than `GetSecrets` when retrieving a single secret cause `GetSecrets` usually requires on call to get all the keys and then one call for each found key.

This becomes relevant specially when using `secretKeyRef` in the components. which should load as fast as possible, while the use case for  `GetSecrets`  is bulk loading from a dedicated store when loading your app or lazy loading when needed.

yaron2: That's a good point.

youngbupark: ok make sense.

Then please keep naming convention consistent with the other bulk api. In the case of store, we use `Bulk` prefix. It would be easier to recognize suffix `s`. 
https://github.com/dapr/components-contrib/blob/master/state/store.go#L15

cmendible: Changed the API to use the Bulk prefix...

yaron2:
@cmendible are we good to merge this from your end?

cmendible:
> @cmendible are we good to merge this from your end?

@yaron2 yes

yaron2 (approved):
lgtm

PR #313, opened 2020-04-21
Unescape Azure blob text data

benc-uk:
# Description

Azure Blob storage doesn't support JSON objects natively (it's not an object store), but currently the data is written "as-is" from the binding request. This means the value that is stored in the blob is an unescaped JSON string, surrounded with quotes e.g.  

If the user sent the following to the output bindings API
```go
bindingOutput := `{
  "metadata": {
    "ContentType": "text/plain"
  },
  "data": "Hello World\nBye"
}`
```

The value in the blob will literally be this, including the quotes
```
"Hello World\nBye"
```

This also means it's impossible to add newlines to any text you output

Note. This change means if the user *did* want to store JSON value in the blob, it would need to be stringified / marshalled first and sent over as a string value. I believe this is consistent with the Blob API and expected behaviour.

I don't know if there's a wider issue here.  
Should all output binding providers be "content type aware" and the Dapr bindings API should have a content-type field on the payload? (Note. I'm not talking about the ContentType metadata field in the example above, as that effectively does nothing). Or perhaps at least expose the expected content-type somewhere? otherwise users are left guessing what the value of `data` should hold

## Issue reference

None

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


amanbha:
@benc-uk Could you please open issue as well for the PR.

yaron2:
> # Description
> Azure Blob storage doesn't support JSON objects natively (it's not an object store), but currently the data is written "as-is" from the binding request. This means the value that is stored in the blob is an unescaped JSON string, surrounded with quotes e.g.
> 
> If the user sent the following to the output bindings API
> 
> ```go
> bindingOutput := `{
>   "metadata": {
>     "ContentType": "text/plain"
>   },
>   "data": "Hello World\nBye"
> }`
> ```
> 
> The value in the blob will literally be this, including the quotes
> 
> ```
> "Hello World\nBye"
> ```
> 
> This also means it's impossible to add newlines to any text you output
> 
> Note. This change means if the user _did_ want to store JSON value in the blob, it would need to be stringified / marshalled first and sent over as a string value. I believe this is consistent with the Blob API and expected behaviour.
> 
> I don't know if there's a wider issue here.
> Should all output binding providers be "content type aware" and the Dapr bindings API should have a content-type field on the payload? (Note. I'm not talking about the ContentType metadata field in the example above, as that effectively does nothing). Or perhaps at least expose the expected content-type somewhere? otherwise users are left guessing what the value of `data` should hold
> 
> ## Issue reference
> None
> 
> ## Checklist
> Please make sure you've completed the relevant tasks for this PR, out of the following list:
> 
> * [x]  Code compiles correctly
> * [ ]  Created/updated tests
> * [ ]  Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

We're now solving the content type in service invocation by passing it along from the caller to the callee.
We'll next be looking at how we serialize state and pass data to bindings.

So far, we've made it explicit that for the HTTP APIs, everything sent must be JSON serializable.
If we decide to shift responsibility more towards the implementations (and users) and provide hints for the content type, we would not need to enforce JSON serialization.

PR #322, opened 2020-04-30
Upgrade Go to 1.14.2 and lint to 1.25.1

youngbupark:
# Description

Upgrade Go to 1.14.2 and lint to 1.25.1

## Issue reference

n/a

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/.github/workflows/components-contrib.yml
+++ b/.github/workflows/components-contrib.yml
@@ -21,11 +21,11 @@ jobs:
     name: Build ${{ matrix.target_os }}_${{ matrix.target_arch }} binaries
     runs-on: ${{ matrix.os }}
     env:
-      GOVER: 1.13.7
+      GOVER: 1.14.2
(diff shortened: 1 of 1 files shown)


mchmarny:
I think this is probably the right thing to do but.... do we want to do it now? Or should we wait for 0.7 to land?

yaron2:
> I think this is probably the right thing to do but.... do we want to do it now? Or should we wait for 0.7 to land?

considering the slew of new linting violations from the test above, we should wait :)

youngbupark:
> I think this is probably the right thing to do but.... do we want to do it now? Or should we wait for 0.7 to land?

Technically, no impact to dapr. We can wait. but CI shows the same errors as you got on your local machine. I recommend you to install older version of linter to fix the lint for now :)

PR #329, opened 2020-05-07
Fix issue with ASB pubsub throttling

jjcollinge:
# Description
Rather than just applying concurrency controls to the message handlers, I've update it to handle it in the receiver as well. This way dapr doesn't pull a message it cannot immediately process.
More details in the original [issue](https://github.com/dapr/dapr/issues/1526).

I've smoke tested this locally, I'd like to check this resolves @programaka issue before merging.

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/1526

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/pubsub/azure/servicebus/metadata.go
+++ b/pubsub/azure/servicebus/metadata.go
@@ -13,7 +13,7 @@ type metadata struct {
 	TimeoutInSec                   int    `json:"timeoutInSec"`
 	HandlerTimeoutInSec            int    `json:"handlerTimeoutInSec"`
 	LockRenewalInSec               int    `json:"lockRenewalInSec"`
-	MaxActiveMessage               int    `json:"maxActiveMessages"`
+	MaxActiveMessages              int    `json:"maxActiveMessages"`
 	MaxActiveMessagesRecoveryInSec int    `json:"maxActiveMessagesRecoveryInSec"`
 	DisableEntityManagement        bool   `json:"disableEntityManagement"`
 	MaxDeliveryCount               *int   `json:"maxDeliveryCount"`
--- a/pubsub/azure/servicebus/metadata.go
+++ b/pubsub/azure/servicebus/metadata.go
@@ -8,16 +8,18 @@ package servicebus
 // Reference for settings:
 // https://github.com/Azure/azure-service-bus-go/blob/54b2faa53e5216616e59725281be692acc120c34/subscription_manager.go#L101
 type metadata struct {
-	ConnectionString              string `json:"connectionString"`
-	ConsumerID                    string `json:"consumerID"`
-	TimeoutInSec                  int    `json:"timeoutInSec"`
-	HandlerTimeoutInSec           int    `json:"handlerTimeoutInSec"`
-	LockRenewalInSec              int    `json:"lockRenewalInSec"`
-	DisableEntityManagement       bool   `json:"disableEntityManagement"`
-	MaxDeliveryCount              *int   `json:"maxDeliveryCount"`
-	LockDurationInSec             *int   `json:"lockDurationInSec"`
-	DefaultMessageTimeToLiveInSec *int   `json:"defaultMessageTimeToLiveInSec"`
-	AutoDeleteOnIdleInSec         *int   `json:"autoDeleteOnIdleInSec"`
-	NumConcurrentHandlers         *int   `json:"numConcurrentHandlers"`
-	PrefetchCount                 *int   `json:"prefetchCount"`
+	ConnectionString               string `json:"connectionString"`
+	ConsumerID                     string `json:"consumerID"`
+	TimeoutInSec                   int    `json:"timeoutInSec"`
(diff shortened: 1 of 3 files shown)


jjcollinge:
When leaving the subscribers idle for a while I'm seeing the below error. I'll investigate.
```
== DAPR == time="2020-05-07T13:02:04.9222981+01:00" level=error msg="azure service bus error: error receiving from topic B, link detached, reason: *Error{Condition: amqp:link:detach-forced, Description: The link 'G27:1364404:O_hltrHw6G9XfAN6RBYhhSB3hZPFEPNVVfHb0Ai6admvA2gt-_eaIw' is force detached. Code: consumer(link614583). Details: InnerMessageReceiver was closed. TrackingId:d98a4c330002001d000960b75eb3f215_G27_B2, SystemTracker:daprdev2:Topic:b|node-subscriber, Timestamp:2020-05-07T12:02:04, Info: map[]}" instance=DESKTOP-8VUOI7P scope=dapr.contrib type=log ver=edge

== DAPR == time="2020-05-07T13:02:04.9407386+01:00" level=error msg="azure service bus error: error closing subscription to topic B, link detached, reason: *Error{Condition: amqp:link:detach-forced, Description: The link 'G27:1364404:O_hltrHw6G9XfAN6RBYhhSB3hZPFEPNVVfHb0Ai6admvA2gt-_eaIw' is force detached. Code: consumer(link614583). Details: InnerMessageReceiver was closed. TrackingId:d98a4c330002001d000960b75eb3f215_G27_B2, SystemTracker:daprdev2:Topic:b|node-subscriber, Timestamp:2020-05-07T12:02:04, Info: map[]}" instance=DESKTOP-8VUOI7P scope=dapr.contrib type=log ver=edge
```

jjcollinge:
When running with `numConcurrentHandlers=1`, `maxDeliveryCount=1`, `lockDurationInSec=60` and a 45 second message handler, I see dead lettering and get the following error:
```
ERRO[0341] azure service bus error: error renewing active message lock(s) for topic C, %!(EXTRA common.Retryable=unhandled error link e109e4ca-86b0-4dbb-ab56-d983ac9e37f8: status code 410 and description: The lock supplied is invalid. Either the lock expired, or the message has already been removed from the queue. Reference:cf7b4117-f4b5-4994-9fe4-c35db5bea881, TrackingId:0694cfab-2f18-4a5c-94d7-6d51eb0cfb6e_B1, SystemTracker:daprdev2:Topic:c|python-subscriber, Timestamp:2020-05-07T18:53:04)  app_id=python-subscriber instance=DESKTOP-8VUOI7P scope=dapr.contrib type=log ver=edge
```
This error is because the server has dead lettered the message before we process it. My suspicion is that the underlying sdk it prefetching and those messages are expiring in an internal buffer before we get them. We have explicitly set the prefetch to both 0 (and we get no messages) and 1 (get the same behaviour).

jjcollinge:
Ok this was caused by the prefetch, I've changed the behaviour to start renewing the locks on all prefetched messages. The prefetch count defaults to 1 but you can configure it using `prefectCount` property or disable it using `prefetchCount=0`. `numConcurrentHandlers` will then throttle the actual processing of these messages as expected. Currently you can set the lock renewal rate using `lockRenewalInSec` property (defaults to 20) - this could be moved to per message using `LockedUntil` at a future date. In my testing using your parameters this has worked as expected with no dead lettering.

jjcollinge:
@yaron2 by default there will be no throttling - as we keep track of every message that is being or going to be processed for lock renewal. You can add throttling or disable lock renewal in high bandwidth apps - however, do you think we need a sensible default throttle to stop this dictionary growing unbounded and potentially crashing the app if lock renewal is enabled.

jjcollinge:
Note to self: at the moment if lock renewal is enabled - the activeMessages dictionary is effectively unbounded. There is no throttle on how quick messages are consumed and therefore this could grow and potentially crash the app. I should add a way to reject messages efficiently when under pressure and assume the server can delay resend.

jjcollinge:
I've added `maxActiveMessages` and `maxActiveMessagesRecoveryInSec` to allow you to throttle the subscriber. Therefore you can use this to control the receiving of messages and `numConcurrentMessages` to control the handling of messages and `lockRenewalInSec` to keep all active messages refreshed.

jjcollinge:
Todo: fix typo - abondonMessage -> abandonMessage

davidfowl:
Isn’t there an API that lets you control the receive loop? cc @clemensv

jjcollinge:
@davidfowl as far I can tell from the [go sdk](https://github.com/Azure/azure-service-bus-go), with subscriptions you can call `Recieve` where you can't throttle the receive loop and `RecieveOne` where you can (as I've added in this PR). If there's something additional I've missed that'd be great! Keen to hear if @clemensv has any ideas.

yaron2:
@jjcollinge are you good to merge?

jjcollinge:
@yaron2 yeh - I'll open an issue to update the docs

Issue #339, opened 2020-05-17
aws components should be able to use default aws credentials resolution


trondhindenes:
## Expected Behavior
One of the great things with AWS is the well-documented and consistent way all sdks resolve credentials. For the aws go libraries, this process is documented here: https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html, but it really is the same for all sdk languages.
It seems like at least the sqs binding in dapr expects a hard-coded credential. This breaks with the above flow and becomes very problematic for organizations that rely on projects such as kube2iam or kiam to inject aws credentials into pods.

I would ask that:
1. aws credential resoultion in dapr is properly documented
2. any config items in dapr yaml files are listed as either optional or required so that it's possible for users to understand what is a minimum viable configuration
3. for aws bindings and modules, dapr respects the default credential flow as describe above.

## Actual Behavior
```
== DAPR == time="2020-05-17T14:22:11.591678113+02:00" level=error msg="failed to init output binding sqs (bindings.aws.sqs): EmptyStaticCreds: static credentials are empty" app_id=messagehandler instance=<computer> scope=dapr.runtime type=log ver=0.7.1
```

trondhindenes:
fyi we now have a generic "aws credentials" abstraction that behaves correctly, but I'll leave this open until we've switched to it in all aws clients

trondhindenes:
@yaron2 I was hoping we'd have updated docs in place, including how to document optional vs required params before closing this.

PR #340, opened 2020-05-17
aws sqs: Use implicit creds and region by default, explicit if specified

trondhindenes:
# Description
This PR adjusts the sqs binding so that it supports implicit credentials and region config. I've described the problem in more detail in #339. Once this PR is merged, all components using aws services should be updated to similar logic. Once the docs repo support separating between optional and required config variables, we can indicate that access key/secret and region are optional but not required.

The functionality of this PR can be tested by sending environment vars into dapr stand-alone instead of configuring the region, access key id and secret access key variables, or by using a local aws profile if configured:

`AWS_PROFILE=myprofile dapr run ...`

or

`AWS_ACCESS_KEY_ID=123 AWS_SECRET_ACCESS_KEY=abc AWS_REGION=eu-west-1 dapr run ...`

## Issue reference

#_339_ (is this the correct format? Maybe fix the PR template?)

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


yaron2:
@trondhindenes thanks, this is great!

in order to not duplicate the logic here, can you move it to a new dir at the root of the repo so that all AWS components can reference it from there?

`authentication/aws/aws.go`

The `getClient` method could become `GetClient(accessKey, secretKey, region string)`

trondhindenes:
Good idea. I will certainly try! Not a go dev but ill give it an honest try.

trondhindenes:
@yaron2 I think I've done that now :-)

yaron2:
So the linter is failing because in Go you shouldn't use an underscore in a package name.

Can you restructure like this?

`/authentication/aws/aws.go`.

And then the package name in `aws.go` is `aws`.

Inside the SQS code, you can create an alias: `aws_auth github.com/dapr-components-contrib/authentication/aws`

trondhindenes:
oh nice. Will try that. Again, I'm a total golang newb so I appreciate the pointers :-)

yaron2:
Alright, we are almost done. Seems like your IDE does not do `goimports`.

Open your terminal, and run `goimports bindings/aws/sqs/sqs.go`

Take the formatted output and replace it in `sqs.go`.

trondhindenes:
yup, I think I have the build nailed down now. Thanks for the patience :-)

trondhindenes:
thanks! I'd be happy to PR the docs change but as mentioned I think there's a need for a "standard way" of separating required from optional parameters, now that access key/secret id/region are optional in some situations.

PR #341, opened 2020-05-18
Update mDNS resolver to respect service IP address

jjcollinge:
# Description
In combination with https://github.com/dapr/dapr/pull/1577 this PR moves from relying on localhost to respecting a service's IP address if provided.

## Issue reference

Please reference the issue this PR will close: [1506](https://github.com/dapr/dapr/issues/1506)

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/servicediscovery/mdns/mdns.go
+++ b/servicediscovery/mdns/mdns.go
@@ -39,7 +39,7 @@ func lookupAddressMDNS(id string) (string, error) {
 	}
 
 	port := -1
-	ip := "localhost" // default
+	var addr string
 	entries := make(chan *zeroconf.ServiceEntry)
 
 	go func(results <-chan *zeroconf.ServiceEntry) {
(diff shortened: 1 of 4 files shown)


servicediscovery/mdns/mdns.go
```diff
@@ -24,28 +24,34 @@ type resolver struct {
...
+// lookupAddressMDNS uses mdns to find the port of a given service entry on a local network
+func lookupAddressMDNS(id string) (string, error) {
 	resolver, err := zeroconf.NewResolver(nil)
 	if err != nil {
-		return -1, fmt.Errorf("failed to initialize resolver: %e", err)
+		return "", fmt.Errorf("failed to initialize resolver: %e", err)
 	}
 
 	port := -1
+	ip := "localhost" // default
```
youngbupark, line 42: will `ip` be overwritten all the time ?

jjcollinge: I would expect it to be yes but this was just a safety in case the entry didn’t have a registered IPv4 or IPv6 address for some reason. I will add an additional condition to the entry for a default case and leave this uninitialized to reduce alloc.

youngbupark:
@yaron2 @LMWF looks good to me. I will merge this pr.

PR #347, opened 2020-05-23
Use shared aws auth package

### LINKED ISSUE #339: aws components should be able to use default aws credentials resolution

trondhindenes:
# Description
This PR builds on the changes made in #340 , using the shared credentials resolver package for all aws clients. I also discovered that the `sqssns` pubsub package implements an "endpoint" parameter, so I added that to the auth abstraction package, and implemented it in all aws packages. This is especially powerful when developing locally against a local DynamoDB container instead of an actual DynamoDB Table running at AWS - should allow some new and interesting integration test scenarios aswell, using libraries such as https://github.com/spulec/moto to mock various aws services.

The only package I'm unsure of is the `sqssns` pubsub package. It uses different param names than all other aws-related packages in dapr, and the best would probably be to get everything aligned now that the project is still in alpha, but this would introduce a breaking change. I'd like the maintainers' thoughts on this.

## Issue reference

Fixes #339 - docs should probably be updated before that issue is closed.

## Checklist
* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/authentication/aws/aws.go
+++ b/authentication/aws/aws.go
@@ -6,7 +6,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 )
 
-func GetClient(accessKey string, secretKey string, region string) (*session.Session, error) {
+func GetClient(accessKey string, secretKey string, region string, endpoint string) (*session.Session, error) {
 	awsConfig := aws.NewConfig()
 
 	if region != "" {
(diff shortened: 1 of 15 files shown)


trondhindenes:
@yaron2 i think merging this was a bit premature. I was hoping to have a discussion about the sqssns package.

PR #350, opened 2020-05-27
Add bi-directional bindings support

yaron2:
This PR does the following:

1. Update the output binding interface to support bi-directional bindings
2. Updates all existing bindings to support the new interface and return a list of operations

### DIFF
--- a/bindings/Readme.md
+++ b/bindings/Readme.md
@@ -31,7 +31,7 @@ Each output binding can decide which operations it supports. This information is
 type OutputBinding interface {
 	Init(metadata Metadata) error
 	Invoke(req *InvokeRequest) error
-	Operations() []bindings.OperationType
+	Operations() []OperationKind
 }
 ```
 A spec is also needed in [Dapr docs](https://github.com/dapr/docs/tree/master/reference/specs/bindings).
--- a/bindings/Readme.md
+++ b/bindings/Readme.md
@@ -31,7 +31,7 @@ Each output binding can decide which operations it supports. This information is
 type OutputBinding interface {
 	Init(metadata Metadata) error
 	Invoke(req *InvokeRequest) error
-	Operations() []string
+	Operations() []bindings.OperationType
 }
 ```
 A spec is also needed in [Dapr docs](https://github.com/dapr/docs/tree/master/reference/specs/bindings).
--- a/bindings/Readme.md
+++ b/bindings/Readme.md
@@ -24,10 +24,14 @@ type InputBinding interface {
 
 Output binding:
 
+An output binding can be used to invoke an external system and also to return data from it.
+Each output binding can decide which operations it supports. This information is communicated to the caller via the `Operations()` method.
+
 ```go
 type OutputBinding interface {
 	Init(metadata Metadata) error
-	Write(req *WriteRequest) error
+	Invoke(req *InvokeRequest) error
+	Operations() []string
 }
 ```
 A spec is also needed in [Dapr docs](https://github.com/dapr/docs/tree/master/reference/specs/bindings).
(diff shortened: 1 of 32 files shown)


bindings/Readme.md
```diff
@@ -24,10 +24,14 @@ type InputBinding interface {
...
 Output binding:
 
+An output binding can be used to invoke an external system and also to return data from it.
+Each output binding can decide which operations it supports. This information is communicated to the caller via the `Operations()` method.
+
 ```go
 type OutputBinding interface {
 	Init(metadata Metadata) error
-	Write(req *WriteRequest) error
+	Invoke(req *InvokeRequest) error
```
youngbupark, line 33: I think `Invoke` needs a response, doesn't it?

bindings/Readme.md
```diff
@@ -24,10 +24,14 @@ type InputBinding interface {
...
+	Operations() []string
```
youngbupark, line 34: Can we define the type for Operation types instead of `string` ? e.g. BindingOperationType

bindings/Readme.md
yaron2: Oh yeah, good catch. Added.

bindings/Readme.md
yaron2: Added `OperationType` since there's "binding" already in the package name.

yaron2:
@youngbupark please recheck.

bindings/requests.go
```diff
@@ -5,8 +5,20 @@
...
 
-// WriteRequest is the object given to an dapr output binding
-type WriteRequest struct {
-	Data     []byte            `json:"data"`
-	Metadata map[string]string `json:"metadata"`
+// InvokeRequest is the object given to a dapr output binding
+type InvokeRequest struct {
+	Data      []byte            `json:"data"`
+	Metadata  map[string]string `json:"metadata"`
+	Operation string            `json:"operation"`
```
youngbupark, line 12: OperationType ?

bindings/requests.go
```diff
@@ -5,8 +5,20 @@
...
 }
+
+// OperationType defines an output binding operation
+type OperationType string
+
+// Non exhaustive list of operations. A binding can add operations that are not in this list.
+const (
+	GetOperation    = "get"
```
youngbupark, line 20: I think we need to specify the exact type for strict typing..

e.g.
```
GetOperation OperationType = "get"
CreateOperation OperationType = "create"
...
```

bindings/requests.go
yaron2: Changed

bindings/requests.go
yaron2: Changed

bindings/requests.go
yaron2: I used `OperationKind` as it stutters less

bindings/alicloud/oss/oss.go
```diff
@@ -50,7 +50,11 @@ func (s *AliCloudOSS) Init(metadata bindings.Metadata) error {
...
 }
 
-func (s *AliCloudOSS) Write(req *bindings.WriteRequest) error {
+func (s *AliCloudOSS) Operations() []bindings.OperationKind {
+	return []bindings.OperationKind{bindings.CreateOperation}
+}
+
+func (s *AliCloudOSS) Invoke(req *bindings.InvokeRequest) (*bindings.InvokeResponse, error) {
 	key := ""
 	if val, ok := req.Metadata["key"]; ok && val != "" {
```
vinayada1, line 59: Don't we need to check the OperationKind in the Invoke method? Why are we assuming that Invoke is for Write?

Issue #225, opened 2020-02-13
[StateStore] PostgreSQL statestore


youngbupark:
## Describe the proposal

This is to create PostgreSQL statestore

youngbupark:
@brooke-hamilton will work on this issue.

youngbupark:
@brooke-hamilton Can you please update the progress?

brooke-hamilton:
Thanks for checking - I'm ready to submit a PR for review and feedback.

Apart from submitting the review PR, remaining work includes implementing the TransactionalStore interface, consistency flags, benchmark tests, and likely some renaming of things internal to the package.

yaron2:
> Thanks for checking - I'm ready to submit a PR for review and feedback.
> 
> Apart from submitting the review PR, remaining work includes implementing the TransactionalStore interface, consistency flags, benchmark tests, and likely some renaming of things internal to the package.

Thanks, submitting the PR as soon as possible would be great :)

youngbupark:
@brooke-hamilton Thanks for the update. I moved this issue to 0.9.0 milestone. 

cc/ @yaron2 @amanbha @msfussell @orizohar

orizohar:
@brooke-hamilton - Thanks for this contribution. On top of this PR, can you please add a PR for the docs repo that provides guidance on using this component here https://github.com/dapr/docs/tree/master/howto/setup-state-store ?
There is a template for these howtos you can find here https://github.com/dapr/docs/blob/master/contributing/howto-template.md

orizohar:
@brooke-hamilton - are you opening a PR on the docs repo for the "how to" for this?
We'd really want to get this into 0.9 and code freeze is end of this week 7/10/2020

brooke-hamilton:
> @brooke-hamilton - are you opening a PR on the docs repo for the "how to" for this?
> We'd really want to get this into 0.9 and code freeze is end of this week 7/10/2020

@orizohar, I will submit the PR today. Thanks for the pointer on the date.

brooke-hamilton:
@orizohar, the PR for documentation is here: https://github.com/dapr/docs/pull/675

PR #353, opened 2020-06-02
[WIP] PostgreSQL State Store

brooke-hamilton:
# Description

Work in Progress - please review for feedback. This PR implements a state store in PostgreSQL.

Remaining work:
- [x] Implement consistency flags
- [x] Implement TransactionalStore
- [x] ~~Implement limits and validation on key size~~ Dapr does not put a limit on key size, so changed PostgreSQL type to text
- [x] Parameter validation for request structs
- [x] Use logger where appropriate
- [x] Manual integration testing with a full Dapr application
- [ ] Run benchmark tests
- [ ] Extend documentation (Including https://github.com/dapr/components-contrib/issues/225#issuecomment-643373578)

Package tests cover 81.6%. Non covered code is mostly error returns from calling other functions.

## Issue reference

Please reference the issue this PR will close: #225

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

@youngbupark, @yaron2


brooke-hamilton:
I'm looking at build failures. One timeout and two failures at the go.mod check-diff step.

yaron2:
Are you using Go 1.14?

yaron2:
If so,

Do `go mod tidy` and push the changes.

brooke-hamilton:
> Are you using Go 1.14?
Yes
```go version go1.14.2 linux/amd64```

Thanks for the feedback on ```go mod tidy```

**Comment by yaron2** [comment]
Thanks for this PR!

> All keys are treated as strings by this store. I can add configuration to instruct Postgres on which data type to use for the key, and validate based on that type, similar to how the SQL Server state store implements it. Key lookups for integer keys would be faster.

I think its Ok to add a data type metadata which is optional and then validated.

> Does Dapr impose a limit on key size? I currently have a limit of 200 characters and treat all keys as strings.

Dapr does not impose a limit on key size.

**Comment by brooke-hamilton** [comment]
The TransactionalStore interface is now implemented.

**Comment by youngbupark** [comment]
@brooke-hamilton can you please update the progress on description ? 
JFYI - I added checkbox for each remaining work items. :)

**Comment by brooke-hamilton** [comment]
@youngbupark I updated the progress in the description. 
One question: the CONTRIBUTING.md document states "All third-party code must be placed in the `vendor/` folder." I don't see any specific configuration that puts dependencies into the vendor folder. Is there an additional step I need to take, or is this an automatic part of the build?

**Comment by youngbupark** [comment]
> @youngbupark I updated the progress in the description.
> One question: the CONTRIBUTING.md document states "All third-party code must be placed in the `vendor/` folder." I don't see any specific configuration that puts dependencies into the vendor folder. Is there an additional step I need to take, or is this an automatic part of the build?

CONTRIBUTING.md doc is wrong, we will update it. We no longer use /vendor after switching go mod. 3rd party pkg will be downloaded when we build the code. So you do not need to care about it.

**Comment by yaron2** [comment]
@brooke-hamilton are there any more outstanding items for this PR?

File: state/postgresql/postgresdbaccess.go [hunk 1]
```diff
@@ -0,0 +1,289 @@
...
+	if err != nil {
+		p.logger.Error(err)
+		return err
+	}
+
+	p.db = db
+
+	pingErr := db.Ping()
+	if pingErr != nil {
+		p.logger.Error(pingErr)
```
yaron2, line 67: The dapr runtime will log any error from `Init`, so this can be removed here.

state/postgresql/postgresdbaccess.go
```diff
@@ -0,0 +1,289 @@
...
+		return pingErr
+	}
+
+	err = p.ensureStateTable(tableName)
+	if err != nil {
+		p.logger.Error(err)
```
yaron2, line 73: The dapr runtime will log any error from `Init`, so this can be removed here.

state/postgresql/postgresdbaccess.go
```diff
@@ -0,0 +1,289 @@
...
+	if req.Key == "" {
+		return fmt.Errorf("missing key in set operation")
+	}
+
+	var valueBytes []byte
+
+	// Convert to json string
+	valueBytes, err = json.Marshal(req.Value)
+	if err != nil {
+		p.logger.Error(err)
```
yaron2, line 103: Should not log errors on a per request operation as this may flood the session.

state/postgresql/postgresdbaccess.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+
+	err = tx.Commit()
+	return err
+}
+
+// Verifies that the sql.Result affected only one row and no errors exist
+func (p *postgresDBAccess) returnSingleDBResult(result sql.Result, err error) error {
+	if err != nil {
+		p.logger.Error(err)
```
yaron2, line 228: Please change the log level to `Debug` for this method

brooke-hamilton:
> @brooke-hamilton are there any more outstanding items for this PR?

@yaron2, I have only one more potential change I would like to get your opinion on. Currently the PostgreSQL table name is hard coded, but I would like to have the table name as a configuration item. However, PostgreSQL does not allow table names to be used as parameters in SQL statements, which forces me into using string concatenation to build the table name into each query. If the table name becomes a user configurable item, then that would potentially expose the code to SQL injection attacks. I am still looking for a good alternative, like a golang function that will check a string for malicious input, but haven't found a solution. Any ideas? How do you feel about the table name being hard coded?

yaron2:
> > @brooke-hamilton are there any more outstanding items for this PR?
> 
> @yaron2, I have only one more potential change I would like to get your opinion on. Currently the PostgreSQL table name is hard coded, but I would like to have the table name as a configuration item. However, PostgreSQL does not allow table names to be used as parameters in SQL statements, which forces me into using string concatenation to build the table name into each query. If the table name becomes a user configurable item, then that would potentially expose the code to SQL injection attacks. I am still looking for a good alternative, like a golang function that will check a string for malicious input, but haven't found a solution. Any ideas? How do you feel about the table name being hard coded?

Most of our state stores have the table name hard-coded. This is fine as long as it's mentioned in the docs.

brooke-hamilton:
@yaron2, I implemented the requested changes. (Thank you for the review!) I'm not sure if I pressed the right buttons to correctly update this PR, but the changes are there.

yaron2:
This is merged and will make it to next week's release.

Thanks a bunch @brooke-hamilton!

youngbupark:
@brooke-hamilton Thanks a lot for this contribution!

brooke-hamilton:
@yaron2, should I submit a PR in the dapr/dapr repo for cmd/daprd/main.go to integrate the postgres state store?

yaron2:
Yes, for sure!

Issue #311, opened 2020-04-20
OnInvoke Slow Performance on Local Machine


XavierGeerinck:
/area runtime

## What version of Dapr?

> 0.6.0

## Expected Behavior

I **expect ms** performance when `Invoking` a service through the **gRPC** binding

## Actual Behavior

Currently I see ~1s delay between call starting and ending

## Steps to Reproduce the Problem

I'm including a Python example here, but it's not the language itself causing the issue. When logging, I see the following output:

```
== APP == [Client][DaprInvoke][debug] Creating Envelope 2020-04-20 14:10:47.693447
== APP == [Client][DaprInvoke][debug] Creating Envelope 2 2020-04-20 **14:10:47.693627**
== APP == [Client][DaprInvoke][debug] Creating Envelope 3 2020-04-20 **14:10:48.698124**
```

When analysing this in more detail, we can see that between Envelope 2 and 3 there is a delay of around 1000ms happening. This is happening while executing the `client.InvokeService(envelope)` function. 

After logging on the **server** to see for execution and response, the following was generated:

```
== APP == OnInvoke: debug -- 2020-04-20 14:10:48.696802
== APP ==         Finished @ 2020-04-20 14:10:48.697117
```

Which allows me to conclude the following:
1. The server response in `ms` magnitude
2. More importantly, the server start processing the message after around 1s (exactly it start processing at `14:10:48.696802` while the message was generated at `14:10:47.693627`)

The above makes me suspect something happening in the Dapr engine which - I assume - is locating the service before it is able to invoke a method on it. **Is someone able to assist with this?**

To reproduce, the following example can be utilized: https://github.com/dapr/python-sdk/tree/master/examples/invoke-simple and modifying it to include logging through the code included below.

```python
from datetime import datetime
import logging
logging.basicConfig(level=logging.INFO, format='%(message)s')
logger = logging.getLogger('MyLogger')


# Start a gRPC client
port = os.getenv('DAPR_GRPC_PORT')
channel = grpc.insecure_channel(f"localhost:{port}")
client = dapr_services.DaprStub(channel)
print(f"Started gRPC client on DAPR_GRPC_PORT: {port}")

data = Any(value='ACTION=1'.encode('utf-8'))
logger.info(f"[Client][DaprInvoke][debug] Creating Envelope {datetime.now().strftime('%Y-%m-%d %H:%M:%S.%f')}")
[...]

PR #362, opened 2020-06-16
Cancel context when service is discovered

youngbupark:
# Description

Cancel context when service is found. 

The current implementation is always waiting for 1 second even if service is discovered. So we can always see 1 second delay for service invocation. This will cancel earlier to fix this issue when service is discovered.

Before (using helloworld python app)
```
ℹ️  Updating metadata for app command: python3 app.py
✅  You're up and running! Both Dapr and your app logs will appear here.

== APP == Time elapsed: 1.0576

== APP == Time elapsed: 1.0166
```

After this fix
```
== APP == Time elapsed: 0.0151

ℹ️  Updating metadata for app command: python3 app.py
✅  You're up and running! Both Dapr and your app logs will appear here.

== APP == Time elapsed: 0.0085

== APP == Time elapsed: 0.0097

== APP == Time elapsed: 0.0097

== APP == Time elapsed: 0.0100

== APP == Time elapsed: 0.0096
```

## Issue reference

#311 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/servicediscovery/mdns/mdns.go
+++ b/servicediscovery/mdns/mdns.go
@@ -15,6 +15,9 @@ import (
 	"github.com/grandcat/zeroconf"
 )
 
+const browseTimeout = time.Second * 1
+
+// NewMDNSResolver creates the instance of mDNS service discovery resolver.
 func NewMDNSResolver(logger logger.Logger) servicediscovery.Resolver {
 	return &resolver{logger: logger}
 }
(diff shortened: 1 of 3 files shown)


servicediscovery/mdns/mdns.go
```diff
@@ -42,6 +46,8 @@ func lookupAddressMDNS(id string) (string, error) {
 	var addr string
 	entries := make(chan *zeroconf.ServiceEntry)
 
+	ctx, cancel := context.WithTimeout(context.Background(), browseTimeout)
```
mukundansundar, line 49: can we leave the `defer cancel()` as such? From docs it seems that multiple calls to the function after the first call does nothing. 
https://golang.org/src/context/context.go?s=7999:8021#L214

youngbupark: ok for the safety.

yaron2:
late lgtm

Issue #364, opened 2020-06-17
ASB lock renewal not isolated per subscription


jjcollinge:
## Expected Behavior
When subscribed to multiple ASB topics, only messages belonging to that subscription entity should be renewed. 

## Actual Behavior
Currently lock renewal attempts to renew all locks on every subscription which leads to the following error:
```
time="2020-06-17T15:58:57.4641242+01:00" level=error msg="azure service bus error: error renewing active message lock(s) for topic B, %!(EXTRA common.Retryable=unhandled error link 6b69e1d7-9b98-43c0-8977-582d3bde1a71: status code 410 and description: The lock supplied is invalid. Either the lock expired, or the message has already been removed from the queue. Reference:860b9c70-97da-4438-a39b-a362860bb97a, TrackingId:50419ebc-96fb-42da-a30a-ff6c8451afac_B1, SystemTracker:daprdev:Topic:b|node-subscriber, Timestamp:2020-06-17T14:58:57)" app_id=node-subscriber instance=XPS15 scope=dapr.contrib type=log ver=edge
```

## Steps to Reproduce the Problem
Subscribe to multiple topics and cause sufficient back pressure for message locks to require renewal.

jjcollinge:
I’ll fix this up ASAP please assign me

yaron2:
Assigned

PR #365, opened 2020-06-17
ASB lock renewal per topic

jjcollinge:
# Description
This PR does the following:
- Handles best effort lock renewal on a per subscription basis
- Updates ASB sdk version
- Refactors subscription logic out of ASB namespace and entity management
- Demotes lock renewal error to warning (this will happen but is expected)
- Tracks async handler cancelFunc to invoke on close
- Handle connection recovery

## Issue reference

Please reference the issue this PR will close: #364 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


jjcollinge:
@yaron2 few things I'm unsure on here:
1. We currently don't propagate a context in the [publish func](https://github.com/dapr/dapr/blob/b529b552dd16a11514c50c630d42ca355b15f4d4/pkg/runtime/runtime.go#L342). Should we do this so we can properly timeout the app handler and do graceful cleanup of async calls.
2. We also don't currently have a pattern for restarting a subscription to a topic in the case of fatal errors i.e. server kills the connection due to a property update or topic/subscription being deleted for instance. In this current build, the subscription is cleaned up and then stops. How do you think we should handle this?

jjcollinge:
@youngbupark any thoughts on the above questions?

youngbupark:
> @yaron2 few things I'm unsure on here:
> 
> 1. We currently don't propagate a context in the [publish func](https://github.com/dapr/dapr/blob/b529b552dd16a11514c50c630d42ca355b15f4d4/pkg/runtime/runtime.go#L342). Should we do this so we can properly timeout the app handler and do graceful cleanup of async calls.

You can find many parts which we are not populating context. While working on service invocation, we populated context in the service invocation call flow, but not the others. I would create the separate work item to populate context. 

> 2. We also don't currently have a pattern for restarting a subscription to a topic in the case of fatal errors i.e. server kills the connection due to a property update or topic/subscription being deleted for instance. In this current build, the error is logged, the subscription is cleaned up (although async calls currently can't be cancelled, see issue 1.) and then dapr continues but since the connection is closed no more messages will be received. How do you think we should actually handle this?

Good points. One of users filed [the issue](https://github.com/dapr/dapr/issues/1688) to reconnect dependency if it is disconnected or reconnected. not all client supports reconnection. So we need to consider to review each component behavior for this cases and improve it. I think conformance tests should test this negative scenario. (cc/ @mukundansundar)

I would track this problem as a separate issue. 

cc/ @orizohar @amanbha @msfussell @yolember

jjcollinge:
@yaron2 please review and test. https://github.com/dapr/docs/pull/642 is the associated docs update.

jjcollinge:
Found an edge case that stops the message processing from functioning correctly once recovering from multiple connection errors. Investigating at the moment.

jjcollinge:
I've raised an issue on the underlying SDK for the above: https://github.com/Azure/azure-service-bus-go/issues/173
I'll try to see if I can fix this later this week.

jjcollinge:
@yaron2 looks like the change downstream in `azure-service-bus-go` requires a change in `go-amqp` which could have a fairly big impact on existing users so might be slow to integrate. I'll back out my hack for now and leave the original issue in. We can merge this PR and open a new issue for the specific edge case. Seem reasonable?

yaron2:
Yes

PR #371, opened 2020-06-26
Handle errors in AzureEventHubs.Read

nachtmaar:
# Description

Handle errors in AzureEventHubs.Read.

While following [this article](https://dev.to/azure/learn-how-to-use-azure-event-hubs-with-the-dapr-framework-333h) on how to use dapr with Azure EventHubs, I was facing the problem that the `storageAccountKey` was in the wrong format (only base64 value was required, but I was using this format instead `DefaultEndpointsProtocol=https;AccountName=daprazure;AccountKey=<the key which was needed as base64>==;EndpointSuffix=core.windows.net`.

The problem is that the error was not visible in the logs of the dapr sidecar.

Now they are:
```
time="2020-06-26T09:05:12.490481193Z" level=error msg="error reading from input binding eventhubs-input: illegal base64 data at input byte 24" app_id=goapp-input-bindingapp instance=goapp-input-binding-7cf48cfc46-jhvsj scope=dapr.runtime type=log ver=edge
```

Credits go also to @k15r

## Issue reference

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


yaron2:
Thanks for this.

Issue #152, opened 2019-12-09
Add transactional capabilities to cloud-based state stores: CosmosDB


yaron2:
## Describe the proposal

In order to support actors in Azure, AWS and GCP environments, we need to enable transactional capabilities for the cloud-native state stores.

Related to https://github.com/orgs/dapr/projects/12

anuragprafulla:
@yaron2 Can I give this a try?

yaron2:
> @yaron2 I would like to take this up!

But of course. Starting with CosmosDB will be best. It needs to support multi operation transactions, meaning update/deletes in the same transaction.

It's ok to do it with stored procedures.

amanbha:
@anuragprafulla Given that the issue is little old, so just checking if you are planning to pick it up

msfussell:
This is to add other transactional state stores that can be used by  Actors. Today only Redis and Mongo DB are supported and ideally we need a cloud state store in each public cloud.

PR #372, opened 2020-06-26
Implement the Multi() interface method adding transaction support in …

LMWF:
# Description

Implement the Multi() interface method adding transaction support in the cosmos db component.  The implementation uses a stored procedure to do this.  The stored procedure is registered in the component's Init().

This is a breaking change because the current implementation uses the "id" field as the partition key and id must be unique. 
 However transactions can only be performed over items with the same partition key.  Therefore this pr requires that the cosmos db container be created with a partition key of `/partitionkey`.  For actors, daprd will use a partition key value of 

```
appid + actortype + actorid
```

to ensure all actor state for the same actor ends up in the same partition.  This doesn't apply to reminders because reminders are not saved under Multi() (also reminder state is saved by actor type and some other info today), so the partition key for that sort of data will simply have the value as the id field.

This document will be updated with the new cosmos setup instructions:
https://github.com/dapr/docs/blob/master/howto/setup-state-store/setup-azure-cosmosdb.md

## Issue reference

Please reference the issue this PR will close: #152 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/state/azure/cosmosdb/cosmosdb.go
+++ b/state/azure/cosmosdb/cosmosdb.go
@@ -8,11 +8,11 @@ package cosmosdb
 import (
 	"encoding/json"
 	"fmt"
-
-	"github.com/dapr/dapr/pkg/logger"
-	jsoniter "github.com/json-iterator/go"
+	"strings"
 
 	"github.com/dapr/components-contrib/state"
+	"github.com/dapr/dapr/pkg/logger"
+	jsoniter "github.com/json-iterator/go"
 
 	"github.com/a8m/documentdb"
 )
@@ -22,6 +22,7 @@ type StateStore struct {
 	client     *documentdb.DocumentDB
 	collection *documentdb.Collection
 	db         *documentdb.Database
+	sp         *documentdb.Sproc
 
 	logger logger.Logger
 }
@@ -36,9 +37,20 @@ type credentials struct {
 // CosmosItem is a wrapper around a CosmosDB document
 type CosmosItem struct {
 	documentdb.Document
-	ID    string      `json:"id"`
-	Value interface{} `json:"value"`
+	ID           string      `json:"id"`
+	Value        interface{} `json:"value"`
+	PartitionKey string      `json:"partitionkey"`
 }
+
+type storedProcedureDefinition struct {
+	ID   string `json:"id"`
+	Body string `json:"body"`
+}
+
+const (
+	storedProcedureName  = "__dapr__"
+	metadataPartitionKey = "partitionkey"
+)
 
 // NewCosmosDBStateStore returns a new CosmosDB state store
 func NewCosmosDBStateStore(logger logger.Logger) *StateStore {
@@ -47,6 +59,8 @@ func NewCosmosDBStateStore(logger logger.Logger) *StateStore {
 
 // Init does metadata and connection parsing
 func (c *StateStore) Init(metadata state.Metadata) error {
+	c.logger.Debugf("Cosmos init start")
+
 	connInfo := metadata.Properties
 	b, err := json.Marshal(connInfo)
 	if err != nil {
@@ -87,12 +101,37 @@ func (c *StateStore) Init(metadata state.Metadata) error {
 	if err != nil {
 		return err
 	} else if len(colls) == 0 {
-		return fmt.Errorf("collection %s for CosmosDB state store not found", creds.Collection)
+		return fmt.Errorf("collection %s for CosmosDB state store not found.  This must be created before Dapr uses it", creds.Collection)
 	}
 
 	c.collection = &colls[0]
 	c.client = client
 
+	sps, err := c.client.ReadStoredProcedures(c.collection.Self)
+	if err != nil {
+		return err
+	}
+
+	// get a link to the sp
+	for i := 0; i < len(sps); i++ {
+		if sps[i].Id == storedProcedureName {
+			c.sp = &sps[i]
+		}
+	}
+
+	if c.sp == nil {
+		// register the stored procedure
+		createspBody := storedProcedureDefinition{ID: storedProcedureName, Body: spDefinition}
+		c.sp, err = c.client.CreateStoredProcedure(c.collection.Self, createspBody)
+		if err != nil {
+			// if it already exists that is success
+			if !strings.HasPrefix(err.Error(), "Conflict") {
+				return err
+			}
+		}
+	}
+
+	c.logger.Debug("cosmos Init done")
 	return nil
 }
 
@@ -100,8 +139,10 @@ func (c *StateStore) Init(metadata state.Metadata) error {
 func (c *StateStore) Get(req *state.GetRequest) (*state.GetResponse, error) {
 	key := req.Key
 
+	partitionKey := populatePartitionMetadata(req.Key, req.Metadata)
+
 	items := []CosmosItem{}
-	options := []documentdb.CallOption{documentdb.PartitionKey(req.Key)}
+	options := []documentdb.CallOption{documentdb.PartitionKey(partitionKey)}
 	if req.Options.Consistency == state.Strong {
 		options = append(options, documentdb.ConsistencyLevel(documentdb.Strong))
 	}
@@ -139,7 +180,8 @@ func (c *StateStore) Set(req *state.SetRequest) error {
 		return err
 	}
 
-	options := []documentdb.CallOption{documentdb.PartitionKey(req.Key)}
+	partitionKey := populatePartitionMetadata(req.Key, req.Metadata)
+	options := []documentdb.CallOption{documentdb.PartitionKey(partitionKey)}
 
 	if req.ETag != "" {
 		options = append(options, documentdb.IfMatch((req.ETag)))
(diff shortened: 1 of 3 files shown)


state/azure/cosmosdb/cosmosdb.go
```diff
@@ -208,3 +265,57 @@ func (c *StateStore) BulkDelete(req []state.DeleteRequest) error {
...
+func (c *StateStore) Multi(operations []state.TransactionalRequest) error {
+	upserts := []CosmosItem{}
+	deletes := []CosmosItem{}
+
+	partitionKey := "UNKNOWN"
+	for _, o := range operations {
+		if o.Operation == state.Upsert {
+			req := o.Request.(state.SetRequest)
+
+			partitionKey = populatePartitionMetadata(req.Key, req.Metadata)
```
amanbha, line 279: does it needs a check that partition key for all requests should be same and then return err if anyone of them is different because cosmos will not guarantee transaction  if partitionkeys are different

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -179,9 +221,21 @@ func (c *StateStore) Delete(req *state.DeleteRequest) error {
...
 
-	selfLink := fmt.Sprintf("dbs/%s/colls/%s/docs/%s", c.db.Id, c.collection.Id, req.Key)
+	partitionKey := populatePartitionMetadata(req.Key, req.Metadata)
+	options := []documentdb.CallOption{documentdb.PartitionKey(partitionKey)}
 
-	options := []documentdb.CallOption{documentdb.PartitionKey(req.Key)}
+	items := []CosmosItem{}
+	_, err = c.client.QueryDocuments(
+		c.collection.Self,
+		documentdb.NewQuery("SELECT * FROM ROOT r WHERE r.id=@id", documentdb.P{Name: "@id", Value: req.Key}),
```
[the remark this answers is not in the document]
LMWF: The delete api takes in a self link to the item to delete (rather than the key) and I don't think it's possible derive get this without querying it.  All the examples show this method.  The original code which formatted the self link based on a format string did not work.  The actual self link has the format string values hashed and if you provide them unhashed it doesn't do it for you.  For example this 
dbs/a/colls/b/docs/c 

might turn into something like 

dbs/dL3mAA==/colls/dE5mAJ-kY9w=/docs/dL9mAJ-kY7xPBQAAAAABBA==/

>> If we have to query use count(*) so that all the data is not returned only 1 integer is returned and then if the value is 0, it doesnt exist.
I'm not sure I understand what you're saying here, our Delete interface expects a single item for a single key.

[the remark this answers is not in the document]
amanbha: thanks for explaining, that sounds fine.

state/azure/cosmosdb/cosmosdb.go
LMWF: Cosmos will fail the stored procedure if the partition keys differ, but I added an earlier check here.

state/azure/cosmosdb/storedprocedure.go
```diff
@@ -0,0 +1,118 @@
...
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+package cosmosdb
+
+const spDefinition string = `// upserts - an array of objects to upsert
+// deletes - an array of objects to delete
+
+function sample(upserts, deletes) {
```
youngbupark, line 11: can we use the meaningful name for this procedure?

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -87,21 +103,49 @@ func (c *StateStore) Init(metadata state.Metadata) error {
...
 	c.collection = &colls[0]
 	c.client = client
 
+	sps, err := c.client.ReadStoredProcedures(c.collection.Self)
+	if err != nil {
+		return err
+	}
+
+	// get a link to the sp
+	for i := 0; i < len(sps); i++ {
```
youngbupark, line 118: nit: we can use `range` to simplify it
```
for i, sps := range sps {
   if sps.Id == storeProcedureName {
   }
}
```

state/azure/cosmosdb/storedprocedure.go
```diff
@@ -0,0 +1,118 @@
...
+        var isAccepted = container.queryDocuments(collectionLink, query, requestOptions, function (err, retrievedDocs, responseOptions) {
+            if (err) {
+                throw err;
+            }
+
+            if (retrievedDocs == null) {                
+                response.setBody(JSON.stringify("success"));
+            } else if (retrievedDocs.length > 0) {                
+                tryDelete(retrievedDocs);
+            } else if (responseOptions.continuation) {
```
youngbupark, line 83: look like this code is for the pagination. so that this needs to be called even if retrievedDocs.length > 0.

Can you please double-check the behavior of this code?

state/azure/cosmosdb/cosmosdb.go
LMWF: Done

state/azure/cosmosdb/storedprocedure.go
LMWF: Yeah, thanks I didn't notice I didn't change it since the name isn't used when calling it.  Changed it to dapr_multi().

state/azure/cosmosdb/storedprocedure.go
LMWF: Removed the pagination code.  We actually don't want pagination/continuation because then it will require multiple sp invocations and it won't be atomic.

youngbupark: What’s the max items per page ?

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -36,17 +38,31 @@ type credentials struct {
 // CosmosItem is a wrapper around a CosmosDB document
 type CosmosItem struct {
 	documentdb.Document
-	ID    string      `json:"id"`
-	Value interface{} `json:"value"`
+	ID           string      `json:"id"`
+	Value        interface{} `json:"value"`
+	PartitionKey string      `json:"partitionkey"`
```
yaron2, line 43: Should be `partitionKey`?

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -36,17 +38,31 @@ type credentials struct {
...
 }
 
+type storedProcedureDefinition struct {
+	ID   string `json:"id"`
+	Body string `json:"body"`
+}
+
+const (
+	storedProcedureName  = "__dapr__"
+	metadataPartitionKey = "partitionkey"
```
yaron2, line 53: same as above

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -36,17 +38,31 @@ type credentials struct {
...
+)
+
 // NewCosmosDBStateStore returns a new CosmosDB state store
 func NewCosmosDBStateStore(logger logger.Logger) *StateStore {
 	return &StateStore{logger: logger}
 }
 
 // Init does metadata and connection parsing
 func (c *StateStore) Init(metadata state.Metadata) error {
+	c.logger.Debugf("Cosmos init start")
```
yaron2, line 64: Change to `CosmosDB init start`

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -208,3 +268,70 @@ func (c *StateStore) BulkDelete(req []state.DeleteRequest) error {
...
+	deletes := []CosmosItem{}
+
+	partitionKey := unknownPartitionKey
+	previousPartitionKey := unknownPartitionKey
+
+	for _, o := range operations {
+		if o.Operation == state.Upsert {
+			req := o.Request.(state.SetRequest)
+
+			partitionKey = populatePartitionMetadata(req.Key, req.Metadata)
```
yaron2, line 284: Lines 284 - 288 and 300 - 304 are essentially the same and can move up to line 281.

state/azure/cosmosdb/storedprocedure.go
LMWF: It depends on configuration which dapr won't have direct control over.  The user will have to configure this when they create the container to suit their needs -  https://docs.microsoft.com/en-us/azure/cosmos-db/request-units

state/azure/cosmosdb/storedprocedure.go
```diff
@@ -0,0 +1,115 @@
...
+    if (typeof upserts === "string") {
+        throw new Error("first arg is a string, expected array of objects");
+    }
+
+    if (typeof deletes === "string") {
+        throw new Error("second arg is a string, expected array of objects");
+    }
+
+    // create the query string used to look up deletes
+    var query = "select * from n where n.id = ";
```
youngbupark, line 25: Can we use `IN` query keyword? - https://docs.microsoft.com/en-us/azure/cosmos-db/sql-query-keywords#in

it will reduce the length of query. Do we know the length of single query?

KaiWalter: `IN` would be applicable here and regarding query string length you would safe all the " or = " string segments.

youngbupark: https://docs.microsoft.com/en-us/azure/cosmos-db/concepts-limits#sql-query-limits 256KB ... I think it should be a problem in normal case.

LMWF: Modified it to use `in`.

youngbupark: *it should not be a problem :)

PR #384, opened 2020-07-07
Fixe for double envelope when published payload is already a cloud event

mchmarny:
Given the findings in [Dapr PR 1741](https://github.com/dapr/dapr/issues/1741) it makes sense to unmarshal the content into actual CloudEventEnvelope vs `interface{}`. 

This A) prevents an existing CloudEvent content from being wrapped into yet another develop, and B) allows for submission of v1.0 CloudEvents 

#387 

* [x] Code compiles correctly
* [x] Created/updated tests


pubsub/envelope.go
```diff
@@ -30,29 +30,55 @@ type CloudEventsEnvelope struct {
...
 
-	return &CloudEventsEnvelope{
-		ID:              id,
-		Source:          source,
-		Type:            eventType,
-		Data:            i,
-		SpecVersion:     CloudEventsSpecVersion,
-		DataContentType: contentType,
-		Subject:         subject,
+	if ce.Type == "" {
```
amanbha, line 59: if the published event is already a cloud event, we should just return it as, no need to modify the properties which are in the original cloud event.

mchmarny: This is defensive posture to allow for CE optional fields to be set by the invoking Dapr code (HTTP or gRPC pub handlers). I was also trying to adhere to existing test to ensure I don't break something so if the published cloud event is not including eventType for example some of these would tail.

amanbha: Once we enable this functionality, IMO, as a user, if I gave a cloud event, I would not expect Dapr to modify it, it should just take it and publish. Tests can be modified.

pubsub/envelope.go
```diff
@@ -15,7 +15,7 @@ const (
 	// CloudEventsSpecVersion is the specversion used by Dapr for the cloud events implementation
 	CloudEventsSpecVersion = "0.3"
 	//ContentType is the Cloud Events HTTP content type
-	ContentType = "application/cloudevents+json"
+	ContentType = "application/json"
```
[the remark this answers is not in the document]
mchmarny: fixed

pubsub/envelope.go
mchmarny: Refactored to return the CE as is if provided by the user. Also added CE content tests to make sure we are not altering the user defined values

pubsub/envelope.go
```diff
@@ -30,29 +33,47 @@ type CloudEventsEnvelope struct {
 	Subject         string      `json:"subject"`
 }
 
-// NewCloudEventsEnvelope returns a new CloudEventsEnvelope
+// NewCloudEventsEnvelope returns CloudEventsEnvelope from data or a new one when data content was not
+// already an existing CE. Invoked by either the HTTP or gRPC pug handles where:
```
LMWF, line 37: pug?  is that supposed to be "publish"?

pubsub/envelope.go
```diff
@@ -30,29 +33,47 @@ type CloudEventsEnvelope struct {
...
+		DataContentType: "application/json",
 		Source:          source,
 		Type:            eventType,
-		Data:            i,
-		SpecVersion:     CloudEventsSpecVersion,
-		DataContentType: contentType,
 		Subject:         subject,
 	}
+
+	// if content was not JSON, set tata and its type to text
```
LMWF, line 72: "data"

pubsub/envelope.go
```diff
@@ -30,29 +33,47 @@ type CloudEventsEnvelope struct {
...
+//  id = new UUID
+// 	source = handling app ID
+//  eventType = const DefaultCloudEventType from this package ("com.dapr.event.sent")
+//  subject = diagnostics span context W3C as string
 func NewCloudEventsEnvelope(id, source, eventType, subject string, data []byte) *CloudEventsEnvelope {
-	if eventType == "" {
-		eventType = DefaultCloudEventType
+	var ce CloudEventsEnvelope
+	err := jsoniter.Unmarshal(data, &ce)
+	if err == nil && ce.ID != "" {
```
LMWF, line 45: Should this case be an error?  I'm reading this as: the user provided a CE but it's an invalid CE because id was not specified.

mchmarny: No, this is a valid CloudEvent (no error and it did unmarshal into CloudEvent). There are 3 scenarios this code covers when the user posts content: 

1. content is not JSON (deserialization error), set type to TXT and create a brand new CloudEvents envelop
2. content is valid JSON (no deserialization error) but the unmarshalling didn't match any fields from the CloudEvent so set type to JSON and set fields on the new CloudEvents envelop
3. content is valid JSON in CloudEvent format (no deserialization error), return the unmarshalled struct as is

pubsub/envelope.go
mchmarny: fixed

pubsub/envelope.go
mchmarny: fixed

pubsub/envelope.go
LMWF: I see, thanks for explaining.  In number 2 there is the possibility that the non-CE json happens to have an "ID" field, which seems like a field a lot of objects may have, which would make this go down the wrong path.  Perhaps using a different field like "SpecVersion" may decrease the chance of this.

mchmarny: Good call! I will combine SpecVersion, Type, and Source to be more defensive on this.

mchmarny: Come to think of it, I included all of the CloudEvent required fields (ID, SpecVersion, Type, Source, DataContentType). PTAL

Issue #366, opened 2020-06-20
Add output support to Twitter binding


mchmarny:
Input binding works for query stream that was defined in the binding YAML. Add support for runtime output binding where query can be passed in as part of of the input parameter metadata

PR #385, opened 2020-07-07
Adds output support to Twitter binding #366

mchmarny:
# Description

Currently the Twitter binding supports only Input which works for query stream that was defined in the binding YAML. This PR adds support for runtime output binding where query can be passed in as part of of the input parameter metadata

## Issue reference

#366 

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation ([doc](https://github.com/dapr/docs/pull/652))


bindings/twitter/twitter.go
```diff
@@ -58,36 +54,34 @@ func (t *twitterInput) parseMetadata(metadata bindings.Metadata) error {
 	if !f || as == "" {
 		return fmt.Errorf("accessSecret not set")
 	}
+
+	// query set only on input binding
+	// in output binding scenario the query is provided on create
 	q, f := metadata.Properties["query"]
-	if !f || q == "" {
-		return fmt.Errorf("query not set")
+	if !f && q != "" {
```
yaron2, line 61: shouldnt this be `if f &&`?

mchmarny: good catch, Fixed. 

```go
if f {
  t.query = q
}
```

Given that `t.query` is already a string so even if we reset it to "" it won't matter.

PR #389, opened 2020-07-13
Fizes JSON array envelope parsing, adds unit tests

mchmarny:
This fixes the issue with string array not being parsed as valid JSON and adds unit tests based on what's tested in end to end tests to catch early

### DIFF
--- a/pubsub/envelope.go
+++ b/pubsub/envelope.go
@@ -95,7 +95,6 @@ func NewCloudEventsEnvelope(id, source, eventType, subject string, data []byte) 
 		Subject:         subject,
 		Data:            string(data),
 	}
-
 }
 
 func getStrVal(m map[string]interface{}, key string) string {
--- a/pubsub/envelope.go
+++ b/pubsub/envelope.go
@@ -35,17 +35,7 @@ type CloudEventsEnvelope struct {
 
 // NewCloudEventsEnvelope returns CloudEventsEnvelope from data or a new one when data content was not
 func NewCloudEventsEnvelope(id, source, eventType, subject string, data []byte) *CloudEventsEnvelope {
-	var ce CloudEventsEnvelope
-	err := jsoniter.Unmarshal(data, &ce)
-
-	if err == nil && ce.ID != "" && ce.SpecVersion != "" && ce.Type != "" && ce.Source != "" && ce.DataContentType != "" {
-		// data was already CloudEvent
-		// assuming the likelihood of other structures having all these
-		//CloudEvent-specific fields is very unlikely
-		return &ce
-	}
-
-	// ensure valid input parameters
+	// defaults
 	if id == "" {
 		id = uuid.New().String()
 	}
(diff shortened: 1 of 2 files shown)


pubsub/envelope.go
```diff
@@ -35,17 +35,7 @@ type CloudEventsEnvelope struct {
...
-	if err == nil && ce.ID != "" && ce.SpecVersion != "" && ce.Type != "" && ce.Source != "" && ce.DataContentType != "" {
-		// data was already CloudEvent
-		// assuming the likelihood of other structures having all these
-		//CloudEvent-specific fields is very unlikely
-		return &ce
-	}
-
-	// ensure valid input parameters
+	// defaults
 	if id == "" {
```
yaron2, line 39: shouldn't lines 49 - 50 go inside the if statement in line 55?

mchmarny: Used again in 95 so for readability reasons, the code checks everything up front. Does not change the execution logic as this value is only set if not provided up stream. Our current unit tests pass bunch of empty strings so not to change the tests, I'm validating on entry.

yaron2: ok

amanbha:
@mchmarny Could you please open an issue for this and tag it for 0.9.0 project.

mchmarny:
Fixes #390

youngbupark:
I am not sure this can fix the issue... 
`data` in CE payload is `message-xxx` as a plain text.. it is not array.. let me debug it more...

https://github.com/dapr/dapr/blob/159c16945d4e9da1e63bb134ceaa902866972a11/tests/e2e/pubsub/pubsub_test.go#L76

pubsub/envelope.go
```diff
@@ -55,22 +45,63 @@ func NewCloudEventsEnvelope(id, source, eventType, subject string, data []byte)
...
-	ce = CloudEventsEnvelope{
+	// content was JSON but not a valid CloudEvent, make one
+	return &CloudEventsEnvelope{
 		ID:              id,
 		SpecVersion:     CloudEventsSpecVersion,
 		DataContentType: "application/json",
 		Source:          source,
 		Type:            eventType,
 		Subject:         subject,
+		Data:            string(data),
```
youngbupark, line 96: This must be `j` which is the deserialized object. Otherwise, data will be treated as normal string in serialization.

please see the original code where it took the deserialized object `i`.

mchmarny:
The test that was failing before was this (abbreviated) 

```shell
Error Trace: pubsub_test.go:149 
275 pubsub_test.go:200 
276 Error: Not equal: 
277 expected: []string{"message-100", "message-101", "message-102"} 
278 actual : []string{"\"message-100\"", "\"message-101\"", "\"message-102\""}
```
The unit tests included with this PR test for this scenario and the PR passes. 

Are you seeing this error in end to end test **AFTER** this PR was applied?

pubsub/envelope.go
youngbupark: https://github.com/dapr/components-contrib/pull/389/files#r453886658 <- original code

mchmarny: This make sense. Made that change but had to change our existing unit tests. 

We are now testing for: 

* JSON obj
* JSON string
* Non-JSON content 
* Existing CloudEvent

PTAL

Issue #388, opened 2020-07-13
Azure Blob Storage  Component- Add "decode base64" option to output binding


mtaghavi2005:
## Describe the proposal

In order to save some file types with binary content on Azure Blob Storage we need to decode base64 payload to corresponding object array. Otherwise, they will be stored on Blob Storage with base64 encoding content and get corrupted. 

It would be great if there is an option (decodeBase64) just like what is already done for Azure Service Bus Component ( #232 ).

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: <NAME>
  namespace: <NAMESPACE>
spec:
  type: bindings.azure.blobstorage
  metadata:
  - name: storageAccount
    value: myStorageAccountName
  - name: storageAccessKey
    value: ***********
  - name: container
    value: container1
  - name: decodeBase64
    value: "true"
```

mtaghavi2005:
I've already done it in following branch:
[https://github.com/mtaghavi2005/components-contrib/tree/add-decode-base64-to-azure-blob-storage](https://github.com/mtaghavi2005/components-contrib/tree/add-decode-base64-to-azure-blob-storage)

If it is applicable, I will submit a PR.
Thank you in advance.

yaron2:
Yes, this can be a good addition.
Feel free to submit a PR.

/cc @lynn-orrell.

philliphoff:
For the sake of consistency, should this capability also be applied to the Azure blob-backed state component?

youngbupark:
> For the sake of consistency, should this capability also be applied to the Azure blob-backed state component?

@philliphoff That's fair. I will create separate issue to handle this. 

cc/ @mchmarny @msfussell @yaron2

youngbupark:
@philliphoff you already created the issue :)

youngbupark:
@mtaghavi2005 Can you please double-check your change with the latest Dapr edge build ? 

Download Link - https://github.com/dapr/dapr/actions/runs/268901219

1. Download correct os and arch binary
2. replace ~/.dapr/bin/daprd with downloaded daprd.
3. Test

mtaghavi2005:
> @mtaghavi2005 Can you please double-check your change with the latest Dapr edge build ?
> 
> Download Link - https://github.com/dapr/dapr/actions/runs/268901219
> 
> 1. Download correct os and arch binary
> 2. replace ~/.dapr/bin/daprd with downloaded daprd.
> 3. Test

[...]

PR #393, opened 2020-07-14
add decodeBase64 option to azure blob storage component

mtaghavi2005:
# Description

Allow Azure Blob Storage binding to decode base64 file content.

In order to save some file types with binary content on Azure Blob Storage we need to decode base64 payload to corresponding object array. Otherwise, they will be stored on Blob Storage with base64 encoding content and get corrupted. 

By configuring the component option decodeBase64 to "true" the component will decode the file content for you:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: <NAME>
  namespace: <NAMESPACE>
spec:
  type: bindings.azure.blobstorage
  metadata:
  - name: storageAccount
    value: myStorageAccountName
  - name: storageAccessKey
    value: ***********
  - name: container
    value: container1
  - name: decodeBase64
    value: "true"
```

## Issue reference

Please reference the issue this PR will close: #388 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


bindings/azure/blobstorage/blobstorage.go
```diff
@@ -135,9 +136,25 @@ func (a *AzureBlobStorage) Invoke(req *bindings.InvokeRequest) (*bindings.Invoke
...
+	unescapedData, unescapeError := strconv.Unquote(string(req.Data))
 
-	_, err := azblob.UploadBufferToBlockBlob(context.Background(), []byte(unescapedData), blobURL, azblob.UploadToBlockBlobOptions{
+	if unescapeError != nil {
+		return nil, unescapeError
+	}
+
+	var data []byte
+
+	if a.metadata.DecodeBase64 == "true" {
```
tcnghia, line 147: Is `"true"` the only allowed positive value? What about other variations like `"True"`, etc?

mtaghavi2005: That's true. Currently, no other positive variations are valid. I have implemented it just like other components like [azure storagequeue](https://github.com/dapr/components-contrib/blob/04ce9b829946707fcf7abcb55b6a5ea6e749c86d/bindings/azure/storagequeues/storagequeues.go#L160) due to consistency. But I agree with you, that it could be more convenient to support other variations as well.

bindings/azure/blobstorage/blobstorage.go
```diff
@@ -135,9 +136,25 @@ func (a *AzureBlobStorage) Invoke(req *bindings.InvokeRequest) (*bindings.Invoke
...
 	// Unescape data which will still be a JSON string
-	unescapedData, _ := strconv.Unquote(string(req.Data))
+	unescapedData, unescapeError := strconv.Unquote(string(req.Data))
 
-	_, err := azblob.UploadBufferToBlockBlob(context.Background(), []byte(unescapedData), blobURL, azblob.UploadToBlockBlobOptions{
+	if unescapeError != nil {
+		return nil, unescapeError
+	}
+
+	var data []byte
```
tcnghia, line 145:
```suggestion
	data := []byte(unescapedData)
```
(avoiding an uninitialized var)

bindings/azure/blobstorage/blobstorage.go
```diff
@@ -135,9 +136,25 @@ func (a *AzureBlobStorage) Invoke(req *bindings.InvokeRequest) (*bindings.Invoke
...
+
+	if a.metadata.DecodeBase64 == "true" {
+		decoded, decodeError := b64.StdEncoding.DecodeString(unescapedData)
+		if decodeError != nil {
+			return nil, decodeError
+		}
+		data = decoded
+	} else {
+		data = []byte(unescapedData)
+	}
```
tcnghia, line 155:
```suggestion
	}
```

tcnghia (changes requested):

bindings/azure/blobstorage/blobstorage.go
mtaghavi2005: You 're right. But what if I extend documentation for this component instead of commenting code? Which one would be better?

mtaghavi2005: Done!

tcnghia (approved):

youngbupark (approved):
lgtm. thanks for your contribution.

Issue #397, opened 2020-07-18
sql store component uses too many connections


LMWF:
From gitter.  In a load test using sql store the user hit port exhaustion.

The component should not be creating a new connection for each call.

## Steps to Reproduce the Problem

Use sql store

Perform many store operations

Issue #398, opened 2020-07-18
sql store component cannot deserialize data when using grpc


LMWF:
## Steps to Reproduce the Problem

This can be reproed by running the state sample using the .net or java sdk.

PR #399, opened 2020-07-18
Sql component should save db object on Init instead of creating a new

LMWF:
# Description

Save the db object during Init() so it can be used for operations - don't recreate it every time.

Also fixes an issue when using grpc - the data should not be marshalled if it's already []byte.

## Issue reference

Please reference the issue this PR will close: #397, #398

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/state/sqlserver/sqlserver.go
+++ b/state/sqlserver/sqlserver.go
@@ -97,6 +97,7 @@ type SQLServer struct {
 	deleteWithoutETagCommand string
 
 	logger logger.Logger
+	db     *sql.DB
 }
 
 func isLetterOrNumber(c rune) bool {
@@ -234,6 +235,11 @@ func (s *SQLServer) Init(metadata state.Metadata) error {
 	s.deleteWithETagCommand = mr.deleteWithETagCommand
 	s.deleteWithoutETagCommand = mr.deleteWithoutETagCommand
 
+	s.db, err = sql.Open("sqlserver", s.connectionString)
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


youngbupark:
What happens if db client is disconnected? In this case, reconnection is required. So I would create helper method to check if existing client connection is valid first and then use the existing connection. otherwise, I would reconnect it first.

LMWF:
I don't think the user needs to manage that themselves:

https://golang.org/pkg/database/sql/#Open
```
The returned DB is safe for concurrent use by multiple goroutines and maintains its own pool of idle connections. Thus, the Open function should be called just once. It is rarely necessary to close a DB.
```

Also it looks like everything internally calls the conn() method which will manage connections for you:
https://golang.org/src/database/sql/sql.go?s=46419:46453#L1133

youngbupark:
> I don't think the user needs to manage that themselves:
> 
> https://golang.org/pkg/database/sql/#Open
> 
> ```
> The returned DB is safe for concurrent use by multiple goroutines and maintains its own pool of idle connections. Thus, the Open function should be called just once. It is rarely necessary to close a DB.
> ```
> 
> Also it looks like everything internally calls the conn() method which will manage connections for you:
> https://golang.org/src/database/sql/sql.go?s=46419:46453#L1133

if the client has reconnection feature by itself, then we do not need to care the connection status.

state/sqlserver/sqlserver.go
```diff
@@ -478,9 +450,16 @@ type dbExecutor interface {
 }
 
 func (s *SQLServer) executeSet(db dbExecutor, req *state.SetRequest) error {
-	json, err := json.Marshal(req.Value)
-	if err != nil {
-		return err
+	var err error
```
youngbupark, line 453: is this same behavior of redis component?

LMWF: Yes.  Other components may have a similar issue as well when handing data received by the grpc api.

yaron2:
lgtm

PR #401, opened 2020-07-20
Influx binding

gbaeke:
# Description

Output binding for InfluxDB 2.0 that uses the line protocol to add data. Component spec takes a Url, Token, Organization and Bucket to connect to InfluxDB 2.0. Tested against local InfluxDB on localhost and InfluxDB cloud (https).

Create operation expects data as { "measurement": "...", "tags":"...", "values": "..." }

## Issue reference

None

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


yaron2:
Thanks, we'll be reviewing soon.

bindings/influx/influx.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	line := fmt.Sprintf("%s,%s %s", jsonPoint["measurement"], jsonPoint["tags"], jsonPoint["values"])
+
+	// write the point
+	err = i.writeAPI.WriteRecord(context.Background(), line)
+	if err != nil {
+		return nil, errors.New("Influx Error: Cannot write point")
+	}
+	i.client.Close()
```
yaron2, line 105: If this closes the channel, it means a second operation will fail, correct?

yaron2:
Sorry the review took so long!

yaron2:
@gbaeke left a review comment, would like to merge this for 0.10 :)

bindings/influx/influx.go
gbaeke: Yes, indeed. I was not quite sure where the connection should be closed to be honest. Looking at other components such as MQTT, it's disconnecting in invoke as well. On second thought, probably should use defer here...

yaron2: Shouldn't close at all, otherwise the component will only ever work for one call to Invoke without restarting it.

gbaeke: I tested the component and it did work well for a stream of data for several hours.

So I should just remove the close or is there a method to implement where I should close the connection?

gbaeke: It looks like Close does not actually close the connection (which is just HTTP) but makes sure the data is written. This would explain why the subsequent calls succeed.

yaron2: Oh ok, that makes sense.

yaron2:
@gbaeke can you open a PR to register this on `dapr/dapr` repo?

In the dapr runtime repository, do the following:

1. Run `go get -u github.com/dapr/components-contrib@master`
2. Add the binding registration code to https://github.com/dapr/dapr/blob/8ffa6151e16160dbd533ed2a790fd8c3f611e415/cmd/daprd/main.go#L298

PR #419, opened 2020-07-24
Update to Cloud Events to 1.0

yaron2:


LMWF:
Is that the only changed needed?

yaron2:
> Is that the only changed needed?

Yes, I went over the specs for both versions and confirmed.

PR #423, opened 2020-08-03
RethinkDB state store support

mchmarny:
# Description

This PR implements the transactional RethinkDB state store outlined in issue #322 (`state/rethinkdb/rethinkdb.go`). This state store has been validated with actors. 

In addition, this PR also includes input binding for change notification from RethinkDB state store (`bindings/rethinkdb/statechange/statechange.go`). Using this binding users can capture state over time for specific objects (e.g. Dapr Actor). A demo of how to use this binding to publish all state store changes to a pub/sub topic is available [here](https://github.com/mchmarny/dapr-state-store-change-handler)

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Extended the documentation - WIP
* [ ] Added in dapr/dapr cmd/main - WIP


youngbupark:
@mchmarny can you resolve merge conflict ?

state/rethinkdb/rethinkdb.go
```diff
@@ -0,0 +1,364 @@
...
+			if !ok {
+				return errors.Errorf("invalid request type (expected DeleteRequest, got %t)", v.Request)
+			}
+			deletes = append(deletes, r)
+		default:
+			return errors.Errorf("invalid operation type: %s", v.Operation)
+		}
+	}
+
+	// best effort, no transacts supported
```
youngbupark, line 270: rethinkdb doesn't support any transaction levels?

state/rethinkdb/rethinkdb.go
```diff
@@ -0,0 +1,364 @@
...
+
+// BulkSet performs a bulk save operation
+func (s *RethinkDB) BulkSet(req []state.SetRequest) error {
+	docs := make([]*stateRecord, len(req))
+	for i, v := range req {
+		docs[i] = &stateRecord{
+			ID:   v.Key,
+			TS:   time.Now().UTC().UnixNano(),
+			Hash: v.ETag,
+			Data: v.Value,
```
youngbupark, line 183: how does rethinkdb client serialize value? Can you please double-check if this implementation has no same issue as [this](https://github.com/dapr/components-contrib/pull/460) ?

bindings/rethinkdb/statechange/statechange.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+	}()
+
+	done := <-b.stopCh
+	b.logger.Errorf("done: %b", done)
+	defer cursor.Close()
+	return nil
+}
+
+func metadataToConfig(cfg map[string]string, logger logger.Logger) (StateConfig, error) {
```
youngbupark, line 109: Is there any good way to not duplicate the same code ?

state/rethinkdb/rethinkdb.go
```diff
@@ -0,0 +1,364 @@
...
+	"time"
+
+	r "github.com/dancannon/gorethink"
+	"github.com/dapr/components-contrib/state"
+	"github.com/dapr/dapr/pkg/logger"
+	"github.com/pkg/errors"
+)
+
+const (
+	stateTableName          = "daprstate"
```
youngbupark, line 22: looks like table name is fixed. as the initial pr, it is fine. maybe it would be good to allow users to use custom table name via metadata in CRD

youngbupark (commented):
I left some comments. I will finish the review by tomorrow.

state/rethinkdb/rethinkdb.go
mchmarny: no

state/rethinkdb/rethinkdb.go
mchmarny: As you know, that double serialization issue was global (as in Dapr-level) and not component specific. Also, the `With struct data` test covers that scenario.

bindings/rethinkdb/statechange/statechange.go
mchmarny: These aren't exactly the same.... Also, I'm not sure we want to encourage direct inter-component dependencies. Increased blast radius on subsequent changes and potential resulting (unintended) side effects, probably not worth to prevent dup of 10-20 lines of generic code. 

Maybe the answer here is some global "helper" package in the "components-contrib" itself that can be used for common/duplicative tasks like parsing metadata that components can use if they want to?

state/rethinkdb/rethinkdb.go
mchmarny: added parameterized table option

mchmarny:
Addresses #472

youngbupark (dismissed):

bindings/rethinkdb/statechange/statechange.go
```diff
@@ -0,0 +1,187 @@
...
+
+	r "github.com/dancannon/gorethink"
+	"github.com/dapr/components-contrib/bindings"
+	"github.com/dapr/dapr/pkg/logger"
+	"github.com/pkg/errors"
+)
+
+// Binding represents RethinkDB change change state input binding which fires handler with
+// both the previous and current state store content each time there is a change.
+type Binding struct {
```
mukundansundar, line 22: isn't the name Binding too generic? Can it be something else?

state/rethinkdb/rethinkdb.go
```diff
@@ -0,0 +1,374 @@
...
+		}
+	}
+
+	// best effort, no transacts supported
+	if err := s.BulkSet(upserts); err != nil {
+		return errors.Wrap(err, "error saving records to the database")
+	}
+
+	if err := s.BulkDelete(deletes); err != nil {
+		return errors.Wrap(err, "error saving records to the database")
```
mukundansundar, line 279: error deleting records

youngbupark: nice catch :)

bindings/rethinkdb/statechange/statechange.go
mukundansundar, line 22: I think the name Binding is too generic since the struct is also exported.

state/rethinkdb/rethinkdb.go
mchmarny: fixed

bindings/rethinkdb/statechange/statechange.go
mchmarny: It is in a Rethink package so this is a good way to prevent stuttering `rethinkdb.RethinkDB`. `rethinkdb.Binding` seems cleaner, but I realize this is the convention we've been following so I'll fix this once it lands.

youngbupark: To be honest, each component has the different convention :)  In Go convention, Binding is recommended. But the best convention is following the existing one.

youngbupark (approved):

mchmarny:
https://github.com/dapr/components-contrib/issues/472

Issue #446, opened 2020-08-28
Cosmosdb component should be looking for metadata["partitionid"] in ExecuteStateTransactionRequest not inside each individual request


LMWF:
## Expected Behavior

Starting v0.10.0 the cosmosdb component should be looking for `metadata["partitionid"]` in `ExecuteStateTransactionRequest` not inside each individual request.

This doesn't affect actor usage, only the "non-actor" transaction api.

## Actual Behavior

The cosmos component is looking for `metadata["partitionid"]` in each individual request so unless this is done you may get `all objects used in Multi() must have the same partition key` when calling the transaction api.

A workaround is to specify the same partition id in each element of each request in the transaction.

LMWF:
https://github.com/dapr/components-contrib/blob/5fdcbe9fe5d22c466b37697156b5a5f76b8255a2/state/azure/cosmosdb/cosmosdb.go#L310

PR #452, opened 2020-09-04
Use metadata at top level instead of per request

LMWF:
# Description

As in title.

## Issue reference

Please reference the issue this PR will close: #446

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/state/azure/cosmosdb/cosmosdb.go
+++ b/state/azure/cosmosdb/cosmosdb.go
@@ -8,7 +8,6 @@ package cosmosdb
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


yaron2:
Why do I have a feeling this is going to break stuff..?

LMWF:
It doesn't break for actors because the changes in 0.10.0 (for the non-actor transaction api) + here are invisible to the user.

It may break usage of the non-actor transaction api shipped a few weeks ago in 0.10.0, but that's only if you were using the workaround described in the linked issue.  That's what this pr aims to solve and make more uniform.

yaron2 (approved):
LGTM

PR #456, opened 2020-09-05
feat(middleware): add opa middleware

RichiCoder1:
# Description

Adds a new Open Policy Agent HTTP Middleware that allows for evaluating requests against rego snippets.

## Issue reference

https://github.com/dapr/dapr/issues/1617

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Created issue in the https://github.com/dapr/docs/ repo: https://github.com/dapr/docs/issues/786 (WIP)

### Example Usage

Adapted from https://www.openpolicyagent.org/docs/latest/http-api-authorization/#8-optional-use-json-web-tokens-to-communicate-policy-data
```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: authz
  namespace: default
spec:
  type: middleware.http.opa
  metadata:
  - name: rego
    value: |
        package http
        import input.request

        default allow = false

        # Allow users to get their own salaries.
        allow {
            some username
            request.method == "GET"
            path = ["finance", "salary", username]
            token.payload.user == username
        }

        # Allow managers to get their subordinate' salaries.
        allow {
            some username
            request.method == "GET"
            path = ["finance", "salary", username]
            token.payload.subordinates[_] == username
        }

        # Allow HR members to get anyone's salary.
        allow {
            request.method == "GET"
            path = ["finance", "salary", _]
            token.payload.hr == true
        }

        path = path {
            path := split(trim(input.path, "/"), "/")
        }

        token = {"payload": payload} {
            auth_header := request.headers["authorization"]
            [_, jwt] := split(auth_header, " ")
            [_, payload, _] := io.jwt.decode(input.token)
        }        
```

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: authz
  namespace: default
spec:
  type: middleware.http.opa
  metadata:
  - name: rego
    value: |
        package http
        import input.request

        default allow = true

        allow = {
            "allow": false, 
            "status_code": 403, 
            "additional_headers": { "www-authenticate": "Bearer realm=\"my_api\"" } 
        } {
            not token.payload.aud["my_api"]
        }

        token = {"payload": payload} {
            auth_header := request.headers["authorization"]
            [_, jwt] := split(auth_header, " ")
            [_, payload, _] := io.jwt.decode(input.token)
        }

```


yaron2:
Can you add a small description of how this middle ware is used from the user's perspective?

/cc @mchmarney

RichiCoder1:
Added some (admittedly contrived) examples to the description.

RichiCoder1:
Thanks to someone else who pointed out an example, it looks like the rest of OPA is packaged up enough that it wouldn't be too hard to do a follow on that supports other OPA capabilities later too: https://github.com/open-policy-agent/example-api-authz-go/blob/master/internal/opa/opa.go

yaron2:
@RichiCoder1 you marked this PR as a draft, which other changes are you expecting to push before this is ready for review?

RichiCoder1:
> @RichiCoder1 you marked this PR as a draft, which other changes are you expecting to push before this is ready for review?

I think I just wanted to make sure everything was working and that the general idea was correct before I marked it read. Should be good now I think

yaron2:
> > @RichiCoder1 you marked this PR as a draft, which other changes are you expecting to push before this is ready for review?
> 
> I think I just wanted to make sure everything was working and that the general idea was correct before I marked it ready. Should be good now I think
> 
> Edit: Will keep the `[WIP]` flag though until I get some confirmation I'm on the right path/people are on board with how this is being implemented as an MVP

Sounds good, thanks.

go.mod
```diff
@@ -25,6 +25,7 @@ require (
 	github.com/apache/pulsar-client-go v0.1.0
 	github.com/aws/aws-sdk-go v1.25.0
 	github.com/baiyubin/aliyun-sts-go-sdk v0.0.0-20180326062324-cfa1a18b161f // indirect
+	github.com/bmizerany/assert v0.0.0-20160611221934-b7ed37b82869
```
yaron2, line 28: Can you use `"github.com/stretchr/testify/assert"`? this is the assertion package we use through-out Dapr.

yaron2 (changes requested):

go.mod
RichiCoder1: Hmm. My bad, must have been an autocomplete fail.

yaron2:
@tcnghia can you please review this PR?

middleware/http/opa/opa_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// RegoResult is the expected result from rego policy
+type RegoResult struct {
+	Allow             bool              `json:"allow"`
+	AdditionalHeaders map[string]string `json:"additional_headers,omitempty"`
+	StatusCode        int               `json:"status_code,omitempty"`
+}
+
+// GetHandler retruns the HTTP handler provided by the middleware
```
tcnghia, line 37:
```suggestion
// GetHandler returns the HTTP handler provided by the middleware
```

middleware/http/opa/opa_middleware.go
```diff
@@ -0,0 +1,178 @@
+package opa
```
tcnghia, line 1: Since we are already in package `opa`, can you please rename the file to just `middleware.go`?

middleware/http/opa/opa_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+	"github.com/valyala/fasthttp"
+)
+
+type opaMiddlewareMetadata struct {
+	Rego            string `json:"rego"`
+	IncludedHeaders string `json:"includedHeaders"`
+}
+
+// NewOpaMiddleware returns a new Open Policy Agent middleware
+func NewOpaMiddleware(logger logger.Logger) *Middleware {
```
tcnghia, line 21: Change to `NewMiddleWare`? That would remove the stutter in downstream call (i.e. `opa.NewMiddleWare()`)

middleware/http/opa/opa_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+
+	"github.com/dapr/components-contrib/middleware"
+	"github.com/dapr/dapr/pkg/logger"
+	"github.com/open-policy-agent/opa/rego"
+	"github.com/valyala/fasthttp"
+)
+
+type opaMiddlewareMetadata struct {
+	Rego            string `json:"rego"`
+	IncludedHeaders string `json:"includedHeaders"`
```
tcnghia, line 17: Can this be array instead of what appears to be some sort of delimited list?

I think that has the bonus of grouping the `strings.Split(...)` call on line #68 closer to the parsing logic.

middleware/http/opa/opa_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+	"errors"
+	"strings"
+
+	"github.com/dapr/components-contrib/middleware"
+	"github.com/dapr/dapr/pkg/logger"
+	"github.com/open-policy-agent/opa/rego"
+	"github.com/valyala/fasthttp"
+)
+
+type opaMiddlewareMetadata struct {
```
tcnghia, line 15: Change to `middlewareMetadata`?

middleware/http/opa/opa_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+		},
+	}
+
+	results, err := query.Eval(context.TODO(), rego.EvalInput(input))
+
+	if err != nil {
+		m.opaError(ctx, err)
+		return false
+	} else if len(results) == 0 {
+		m.opaError(ctx, errors.New("received no results back from rego policy. Are you setting data.http.allow?"))
```
tcnghia, line 105: Can you please extract this into a var outside of this function, like `errRegoNoResult`

middleware/http/opa/opa_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+		return func(ctx *fasthttp.RequestCtx) {
+			if handled := m.evalRequest(ctx, meta, &query); handled {
+				return
+			}
+			h(ctx)
+		}
+	}, nil
+}
+
+func (m *Middleware) evalRequest(ctx *fasthttp.RequestCtx, meta *opaMiddlewareMetadata, query *rego.PreparedEvalQuery) bool {
```
tcnghia, line 66: Can you please document the meaning of the output `bool` here?

middleware/http/opa/opa_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+
+	input := map[string]interface{}{
+		"request": map[string]interface{}{
+			"method":    string(ctx.Method()),
+			"path":      string(ctx.Path()),
+			"raw_query": string(ctx.QueryArgs().QueryString()),
+			"query":     queryArgs,
+			"headers":   headers,
+			"scheme":    string(ctx.Request.URI().Scheme()),
+			//TODO: allow opting into body support? Reading body isn't efficient
```
tcnghia, line 94: Please file tracking issues for these TODO as these are very hard to keep track and follow up later.

After that, we can replace `TODO` with `TODO(#issue)` or `TODO(<link-to-issue-comment>)`.

middleware/http/opa/opa_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+			//TODO: flow parsed token from other http middlewares?
+		},
+	}
+
+	results, err := query.Eval(context.TODO(), rego.EvalInput(input))
+
+	if err != nil {
+		m.opaError(ctx, err)
+		return false
+	} else if len(results) == 0 {
```
tcnghia, line 104:
```suggestion
	if err != nil {
		m.opaError(ctx, err)
		return false
	}
	
	if len(results) == 0 {
```

middleware/http/opa/opa_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+		m.opaError(ctx, errors.New("received no results back from rego policy. Are you setting data.http.allow?"))
+		return false
+	} else if allowed := m.handleRegoResult(ctx, results[0].Bindings["result"]); !allowed {
+		return false
+	}
+	return true
```
tcnghia, line 110:
```suggestion
	}
	return m.handleRegoResult(ctx, results[0].Bindings["result"])
```

middleware/http/opa/opa_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func (m *Middleware) handleRegoResult(ctx *fasthttp.RequestCtx, result interface{}) bool {
+	if allowed, ok := result.(bool); ok {
+		if !allowed {
+			ctx.Error(fasthttp.StatusMessage(fasthttp.StatusUnauthorized), fasthttp.StatusUnauthorized)
+			return false
+		}
+		return true
```
tcnghia, line 119:
```suggestion
		if !allowed {
			ctx.Error(fasthttp.StatusMessage(fasthttp.StatusUnauthorized), fasthttp.StatusUnauthorized)
		}
		return allowed
```

middleware/http/opa/opa_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	if _, ok := result.(map[string]interface{}); !ok {
+		m.opaError(ctx, errors.New("got an invalid type back from repo policy. Only a boolean or map is valid"))
```
tcnghia, line 123: Can you please extract this err to a var outside of this func, e.g. `errInvalidRegoPolicyType`?

middleware/http/opa/opa_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+
+	// Is it expensive to marshal back and forth? Should we just manually pull out properties?
+	marshaled, err := json.Marshal(result)
+	if err != nil {
+		m.opaError(ctx, err)
+		return false
+	}
+
+	regoResult := RegoResult{
+		StatusCode:        -1,
```
tcnghia, line 135: Can you please either link to rego documentation to explain what StatusCode -1 mean? Is there a constant with descriptive name that we can use? If not, can you please extract this to a constant with a descriptive name?

middleware/http/opa/opa_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+	} else if len(results) == 0 {
+		m.opaError(ctx, errors.New("received no results back from rego policy. Are you setting data.http.allow?"))
+		return false
+	} else if allowed := m.handleRegoResult(ctx, results[0].Bindings["result"]); !allowed {
+		return false
+	}
+	return true
+}
+
+func (m *Middleware) handleRegoResult(ctx *fasthttp.RequestCtx, result interface{}) bool {
```
tcnghia, line 113: Please add a short doc comment to mention what the returned `bool` means in case of `true` or `false`.

middleware/http/opa/opa_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+		AdditionalHeaders: make(map[string]string),
+	}
+
+	if err = json.Unmarshal(marshaled, &regoResult); err != nil {
+		m.opaError(ctx, err)
+		return false
+	}
+
+	if !regoResult.Allow {
+		if regoResult.StatusCode >= 100 {
```
tcnghia, line 145: Please add a comment explaining what this means.

middleware/http/opa/opa_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+		for key, value := range regoResult.AdditionalHeaders {
+			ctx.Response.Header.Set(key, value)
+		}
+		return false
+	}
+
+	for key, value := range regoResult.AdditionalHeaders {
+		ctx.Request.Header.Set(key, value)
+	}
+	return true
```
tcnghia, line 157:
```suggestion
	for key, value := range regoResult.AdditionalHeaders {
		ctx.Request.Header.Set(key, value)
	}
	if !regoResult.Allow && regoResult.StatusCode >= 100 {
		ctx.Response.SetStatusCode(regoResult.StatusCode)
	}
	return regoResult.Allow
```

middleware/http/opa/opa_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func (m *Middleware) opaError(ctx *fasthttp.RequestCtx, err error) {
+	ctx.Error(fasthttp.StatusMessage(fasthttp.StatusUnauthorized), fasthttp.StatusUnauthorized)
+	ctx.Response.Header.Set("x-dapr-opa-error", "1")
```
tcnghia, line 162: Please extract `x-dapr-opa-error` into a const, or reuse if there is already one somewhere? Also, can you please add some comments explaining why "1"? Is that intended to be a boolean (in which case can we use "true" instead)?

middleware/http/opa/opa_middleware_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/dapr/components-contrib/middleware"
+	"github.com/dapr/dapr/pkg/logger"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	fh "github.com/valyala/fasthttp"
+)
+
+func mockedRequestHandler(ctx *fh.RequestCtx) {}
+
+func TestBasicOpaPolicy(t *testing.T) {
```
tcnghia, line 16: These tests are very similar in structure. May be consider writing using table-driven test https://dave.cheney.net/2019/05/07/prefer-table-driven-tests and reuse much of the skeleton? Thanks!

tcnghia (changes requested):

RichiCoder1:
@tcnghia Thanks for all the feedback! Will dig in and address short. For at least some of it though, specifically the duplication in naming, I was mostly following what the other Http midlewares were doing. I'm fine (and actually prefer) removing the duplicate naming, but I wanted to make sure it was ok to break with the other middleware first.

middleware/http/opa/opa_middleware.go
RichiCoder1: Is that something that the middleware metadata will allow? If so, I'll make this swap, I'd much rather have it be an array.

middleware/http/opa/opa_middleware.go
RichiCoder1: Will do on extracting the constant name. The value being `1` is an http-ism for truthyness. I'm cool doing `true` too.

middleware/http/opa/opa_middleware_test.go
RichiCoder1: Oooo, that's neat. I'll get the tests converted, certaintly a lot of duplication hapening.

middleware/http/opa/opa_middleware.go
RichiCoder1: Wanted to make sure the `opa` middleware was a go before I extracted these out. If consensus is yes on adding the middleware, I can get these marshaled to issues.

middleware/http/opa/opa_middleware.go
yaron2: Metadata for components only supports key/Val's, so this does in fact need to be a delimited string..

middleware/http/opa/opa_middleware.go
yaron2: This was discussed in a maintainers meeting and was decided to be included, provided the PR is approved and that there's a docs update for the middleware :)

RichiCoder1: @yaron2 Cool beans! I'll spin out these tickets and start a docs pr then

tcnghia:
@RichiCoder1 thanks for addressing the feedback! I will leave it to @yaron2 to comment about the overall theme.

middleware/http/opa/opa_middleware.go
tcnghia: Even if the metadata value is a string, when parsing it we can use array in this intermediate structure (`opaMiddlewareMetadata`) so that no consumer of such data structure need to do the string split.

middleware/http/opa/opa_middleware.go
tcnghia: I see, thanks for explaining. The two blocks looks so similar and I overlooked the difference. Thanks for adding a comment.

middleware/http/opa/middleware.go
```diff
@@ -0,0 +1,196 @@
...
+// RegoResult is the expected result from rego policy
+type RegoResult struct {
+	Allow             bool              `json:"allow"`
+	AdditionalHeaders map[string]string `json:"additional_headers,omitempty"`
+	StatusCode        int               `json:"status_code,omitempty"`
+}
+
+const (
+	opaErrorHeaderKey         = "x-dapr-opa-error"
+	opaErrorNoResult          = "received no results back from rego policy. Are you setting data.http.allow?"
```
tcnghia, line 41: you can call errors.New() here and avoid calling when use. I don't think errors.New() decorate things with stacktrace, so there is no need to create new one every time. 

If you do so, it is customary to name an error like `errFoo`.

middleware/http/opa/middleware.go
```diff
@@ -0,0 +1,196 @@
...
+		StatusCode:        meta.DefaultStatus,
+		AdditionalHeaders: make(map[string]string),
+	}
+
+	if err = json.Unmarshal(marshaled, &regoResult); err != nil {
+		m.opaError(ctx, err)
+		return false
+	}
+
+	// If the result isn't allowed, set the response status and
```
tcnghia, line 160: This reads really well now, thanks a lot!

tcnghia (changes requested):
Looking good! One more comment.

middleware/http/opa/middleware.go
RichiCoder1: Awesome! I wasn't sure, will fix that up now.

RichiCoder1: Fixed up

tcnghia:
@RichiCoder1 thanks a lot for adding more tests. awesome.

@yaron2 I don't have anything else to add, this looks good to me.

middleware/http/opa/opa_middleware.go
yaron2: That's possible, but basically there are no other consumers of the metadata struct, it's internal to the Middleware which gets registered in the core Dapr runtime with an interface.

yaron2 (approved):
/lgtm

yaron2:
Thanks @tcnghia and @RichiCoder1.

RichiCoder1:
🥳 Whoop! Happy to have contributed. Is there anything else I need to do to get this integrated aside from docs?

yaron2:
> 🥳 Whoop! Happy to have contributed. Is there anything else I need to do to get this integrated aside from docs?

Yep. Please open a PR in the main Dapr repo to register the middleware. It's pretty straightforward, see here: https://github.com/dapr/dapr/blob/1e3b7aa3202e268759bce6cf9b30b6d95471ab8c/cmd/daprd/main.go#L387

To get the latest version of contrib with the OPA middleware, run `go get -u github.com/dapr/components-contrib@master` and then make the changes in `main.go`.

Issue #455, opened 2020-09-05
Add Qos to rabbitMQ binding


abserari:
## Describe the proposal
make rabbitMQ input binding could configure the Qos of Consumer.

## task
- [x] update code
- [ ] update doc to notify the Qos arguments. I find no doc about the input binding Yaml so I don't know how to update it.

## Refer
https://github.com/rabbitmq/rabbitmq-server/pull/1408
https://www.rabbitmq.com/confirms.html#channel-qos-prefetch
https://www.rabbitmq.com/consumer-prefetch.html

## Reason
Yanzhi Li
@Li-Yanzhi
Sep 02 11:33
DAPR won’t distribute message from input binding to multiple instances of consumer?

I have the following scenario and issue:

1) External system publish 500 TODO job message to RabbitMQ
2) Consumer Pod use input binding (AMQP protocol) to receive messages from RabbitMQ, and Consumer Pod is running on demand (if no message, then no instances running)
2) KEDA scale the Consumer Pod from 0 to 5 based on custom metrics of RabbitMQ queueLength
3) Kubernetes create 5 instances of Consumer Pod based on scale rule (incoming 500 messages)
4) The daprd sidecar of first instances of Consumer Pod will get all 500 messages and processed them one by one, each message will use 20 seconds to complete, at meantime, the status of all messages in RabbitMQ changed from Ready to Unacked - UnAcknoledged
5) The other 4 instances of consumer Pod did NOT get any un-processed messages from RabbitMQ and free all time

I need all 5 instances of Consumer Pod parallelly process all non-complete messages in RabbitMQ, What configuration I missed or did wrong?

Yaron Schneider
@yaron2
Sep 03 01:50
What you are trying to accomplish should be supported. I'll look into the RabbitMQ binding

abserari:
I completed the code here with two-line codes. https://github.com/abserari/components-contrib

yaron2:
@abserari can you open a PR?

abserari:
hello, I find no way to write the doc about how to set up an input binding component.

yaron2:
> hello, I find no way to write the doc about how to set up an input binding component.

https://github.com/dapr/docs/blob/master/reference/specs/bindings/rabbitmq.md

PR #457, opened 2020-09-06
make rabbitMQ input binding could configure the Qos of Consumer.

abserari:
# Description

make rabbitMQ input binding could configure the Qos of Consumer.

## Issue reference

https://github.com/dapr/components-contrib/issues/455

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/bindings/rabbitmq/rabbitmq.go
+++ b/bindings/rabbitmq/rabbitmq.go
@@ -34,8 +34,9 @@ type rabbitMQMetadata struct {
 	Host             string `json:"host"`
 	Durable          bool   `json:"durable,string"`
 	DeleteWhenUnused bool   `json:"deleteWhenUnused,string"`
-	PrefetchCount    int    `json:"prefetchCount"`
+	PrefetchCount    string `json:"prefetchCount"`
 	defaultQueueTTL  *time.Duration
+	prefetchCount    int
 }
 
 // NewRabbitMQ returns a new rabbitmq instance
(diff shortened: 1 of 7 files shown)


Li-Yanzhi (commented):
if user doesn't provide the PrefetchCount parameter,  the default value should be 0.

abserari:
> if the user doesn't provide the PrefetchCount parameter, the default value should be 0.

when doesn't provide PrefetchCount, the value would be default 0 because go init int with 0 value.

bindings/rabbitmq/rabbitmq.go
```diff
@@ -34,7 +35,9 @@ type rabbitMQMetadata struct {
 	Host             string `json:"host"`
 	Durable          bool   `json:"durable,string"`
 	DeleteWhenUnused bool   `json:"deleteWhenUnused,string"`
+	PrefetchCount    string `json:"prefetchCount"`
```
yaron2, line 38: Make "PrefetchCount    string `json:"prefetchCount"`" into PrefetchCount    int `json:"prefetchCount"` and remove `prefetchCount    int`

yaron2 (changes requested):

bindings/rabbitmq/rabbitmq.go
abserari: It seems JSON.Unmarshal Couldn't resolve the string to int because metadata is map[string]string.
`&json.UnmarshalTypeError{Value:"string", Type:(*reflect.rtype)(0x1236720), Offset:84, Struct:"rabbitMQMetadata", Field:"prefetchCount"}`

abserari: 2C3928DC421EE14E4530CA9AAEABFE73
 I check the source code of JSON, and unless we change the int to other, It may not be Unmarshal success.

abserari: I update the way to use metadate["prefetch"] to retrieve the data instead of the JSON decode now.

yaron2 (approved):
lgtm

PR #460, opened 2020-09-08
Properly handle serialization deserizliation for redis transaction api.

mukundansundar:
# Description

Found an issue for Transaction API via Java Integration Tests. 

When the state is stored via an **upsert** call made and the got back via the normal getState API call, it can be observed that there is an extra Json Marshal happening in the code. 

Example:
```json
data: "\"{\\\"propertyA\\\":\\\"data in property A\\\",\\\"propertyB\\\":\\\"data in property B\\\"}\""
```
Instead of 
```json
data: "{\"propertyA\":\"data in property A\",\"propertyB\":\"data in property B\"}"
````
The second output is what we have when we have a normal save and get state call done via Dapr. 

The code change here, checks if the incoming request value is a byte array or not and then properly serializes the value. 
## Issue reference

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/state/azure/cosmosdb/cosmosdb.go
+++ b/state/azure/cosmosdb/cosmosdb.go
@@ -306,6 +306,7 @@ func (c *StateStore) Multi(request *state.TransactionalStateRequest) error {
 		if o.Operation == state.Upsert {
 			req := o.Request.(state.SetRequest)
 
+			// Value need not be marshaled here. It is handled by cosmosdb client.
 			upsertOperation := CosmosItem{
 				ID:           req.Key,
 				Value:        req.Value,
--- a/state/postgresql/postgresdbaccess.go
+++ b/state/postgresql/postgresdbaccess.go
@@ -93,14 +93,15 @@ func (p *postgresDBAccess) setValue(req *state.SetRequest) error {
 		return fmt.Errorf("missing key in set operation")
 	}
 
-	var valueBytes []byte
-
 	// Convert to json string
-	valueBytes, err = json.Marshal(req.Value)
-	if err != nil {
-		return err
+	var bt []byte
+	b, ok := req.Value.([]byte)
+	if ok {
+		bt = b
+	} else {
+		bt, _ = json.Marshal(req.Value)
 	}
-	value := string(valueBytes)
+	value := string(bt)
 
 	var result sql.Result
 
--- a/state/postgresql/postgresdbaccess.go
+++ b/state/postgresql/postgresdbaccess.go
@@ -14,6 +14,7 @@ import (
 	"strconv"
 
 	"github.com/dapr/components-contrib/state"
+	"github.com/dapr/components-contrib/state/utils"
 	"github.com/dapr/dapr/pkg/logger"
 
 	// Blank import for the underlying PostgreSQL driver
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: state/redis/redis.go [hunk 1]
```diff
@@ -298,8 +298,16 @@ func (r *StateStore) Multi(request *state.TransactionalStateRequest) error {
 	for _, o := range request.Operations {
 		if o.Operation == state.Upsert {
 			req := o.Request.(state.SetRequest)
-			b, _ := r.json.Marshal(req.Value)
-			pipe.Set(req.Key, b, defaultExpirationTime)
+
```
mukundansundar, line 301: changes similar to https://github.com/dapr/components-contrib/pull/460/files#diff-ae10e12c8f4b65b24b405cfe5a7ed386R255

state/redis/redis.go
```diff
@@ -298,8 +298,16 @@ func (r *StateStore) Multi(request *state.TransactionalStateRequest) error {
...
+			var bt []byte
+			b, ok := req.Value.([]byte)
+			if ok {
+				bt = b
+			} else {
+				bt, _ = r.json.Marshal(req.Value)
+			}
```
tcnghia, line 308:
```suggestion
			bt, ok := req.Value.([]byte)
			if !ok {
				bt, _ = r.json.Marshal(req.Value)
			}
```

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -306,6 +306,7 @@ func (c *StateStore) Multi(request *state.TransactionalStateRequest) error {
 		if o.Operation == state.Upsert {
 			req := o.Request.(state.SetRequest)
 
+			// Value need not be marshaled here. It is handled by cosmosdb client.
```
mukundansundar, line 309: https://github.com/a8m/documentdb/blob/59a8c720ee4a28e9aed5fabde2949a85bb3a3238/client.go#L166

state/postgresql/postgresdbaccess.go
```diff
@@ -93,14 +93,15 @@ func (p *postgresDBAccess) setValue(req *state.SetRequest) error {
 		return fmt.Errorf("missing key in set operation")
 	}
 
-	var valueBytes []byte
-
 	// Convert to json string
-	valueBytes, err = json.Marshal(req.Value)
-	if err != nil {
-		return err
+	var bt []byte
```
youngbupark, line 97: if this code is used in multiple places, it would be great to create shared util.

youngbupark (changes requested):

youngbupark (changes requested):

state/utils/utils.go
```diff
@@ -0,0 +1,17 @@
...
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+package utils
+
+func Marshal(val interface{}, marshaler func(interface{}) ([]byte, error)) ([]byte, error) {
+	var err error
+	bt, ok := val.([]byte)
+	err = nil
```
youngbupark, line 11: no need to set error to nil. in go, default error value is nil.

if you want to set explicitly, 

I would change line 9.
```
var err error = nil
```

youngbupark (changes requested):

artursouza (approved):

youngbupark (approved):

PR #468, opened 2020-09-23
Bi-directional output binding for PostgreSQL

mchmarny:
There has been a few issues open now against Dapr for some form of a more granular state query capability (e.g. [state filtering](https://github.com/dapr/dapr/issues/1339) or [SQL against state](https://github.com/dapr/dapr/issues/925)). Also, in the current implementation, Dapr doesn't allow for sharing of state between applications. There are some good reason for this (microservices, preventing key collision etc), and while there is an [ongoing discussion](https://github.com/dapr/dapr/issues/809) WRT enabling state grouping to solve for that, I've put together a more targeted solution to help us validate these use-cases using binding. 

This bi-directional output binding for PostgreSQL supports 3 operation types (kinds):

* `exec` for `DDL` operations, like table creation, as well as `INSERT`, `UPDATE`, `DELETE` operations which return only metadata (e.g. number of affected rows)
* `query` for `SELECT` operations which returns the metadata as well as data in a form of array of rows
* `close` for explicitly closing the DB connection and returning it to the pool

## Config 

Binding expects only one required parameter in the init metadata `url`. The URL can include any number of connection and connection pool configurations (e.g. `connect_timeout=5`)

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: mydb
spec:
  type: bindings.postgres
  metadata:
  - name: url
    value: postgres://daprtest@localhost:5432/daprtest?application_name=test&connect_timeout=5
```

## Caveats 

* No parameterized queries, all SQL needs to be passed as string in metadata
* `Metadata` vs `Data` on `bindings.InvokeRequest` to to simplify the use and not require users to serialize their intent on each operation 
* Resulting `bindings.InvokeResponse` includes metadata with things like duration of query or number of rows affected. Only in case of a `query` operation is the response `Data` populated with an array of rows returned from DB
* Binding uses connection pool. To explicitly close the connection and return it to the pool users need to use the `close` operation type 
* Binding itself doesn't prevent SQL injection, like with any DB-based application, validate the input before executing query. 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


mchmarny:
Addresses dapr/#[2088](https://github.com/dapr/dapr/issues/2088)

bindings/postgres/postgres.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+var _ = bindings.OutputBinding(&Binding{})
+
+// NewBinding returns a new PostgreSQL output binding
+func NewBinding(logger logger.Logger) *Binding {
+	return &Binding{logger: logger}
+}
+
+// Init initializes the Twitter binding
```
mukundansundar, line 46: nitpick: Postgres binding

bindings/postgres/postgres.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	poolConfig, err := pgxpool.ParseConfig(url)
+	if err != nil {
+		errors.Wrap(err, "error opening DB connection")
+	}
+
+	b.db, err = pgxpool.ConnectConfig(context.Background(), poolConfig)
+	if err != nil {
+		errors.Wrap(err, "unable to ping the DB")
```
mukundansundar, line 60: return
```suggestion
		return errors.Wrap(err, "unable to ping the DB")
```

bindings/postgres/postgres.go
```diff
@@ -0,0 +1,176 @@
...
+// Init initializes the Twitter binding
+func (b *Binding) Init(metadata bindings.Metadata) error {
+	url, ok := metadata.Properties[ConnectionURLKey]
+	if !ok || url == "" {
+		return errors.Errorf("required metadata not set: %s", ConnectionURLKey)
+	}
+
+	poolConfig, err := pgxpool.ParseConfig(url)
+	if err != nil {
+		errors.Wrap(err, "error opening DB connection")
```
mukundansundar, line 55:
```suggestion
		return errors.Wrap(err, "error opening DB connection")
```

bindings/postgres/postgres.go
```diff
@@ -0,0 +1,176 @@
...
+// List of operations.
+const (
+	ExecOperation  bindings.OperationKind = "exec"
+	QueryOperation bindings.OperationKind = "query"
+	CloseOperation bindings.OperationKind = "close"
+
+	ConnectionURLKey = "url"
+	CommandSQLKey    = "sql"
+
+	CommandTimeoutDefault = 5
```
mukundansundar, line 30: should this be a configurable property?

bindings/postgres/postgres.go
```diff
@@ -0,0 +1,176 @@
...
+
+	endTime := time.Now().UTC()
+	resp.Metadata["end-time"] = endTime.Format(time.RFC3339Nano)
+	resp.Metadata["duration"] = endTime.Sub(startTime).String()
+
+	return resp, nil
+}
+
+// QueryResults is
+type QueryResults struct {
```
mukundansundar, line 132: not used?

bindings/postgres/postgres_test.go
```diff
@@ -0,0 +1,111 @@
...
+	t.Log("POSTGRES_RUN_LIVE_TEST set, running live DB test")
+
+	// live DB test
+	m := bindings.Metadata{}
+	m.Properties = map[string]string{
+		ConnectionURLKey: os.Getenv("POSTGRES_TEST_CONN_URL"),
+	}
+
+	logger := logger.NewLogger("test")
+	// logger.SetOutputLevel("debug")
```
mukundansundar, line 54: can be removed?

go.sum
```diff
@@ -587,6 +587,8 @@ github.com/lib/pq v1.2.0 h1:LXpIM/LZ5xGFhOpXAQUIMM1HdyqzVYM13zNdjCEEcA0=
 github.com/lib/pq v1.2.0/go.mod h1:5WUZQaWbwv1U+lTReE5YruASi9Al49XbQIvNi/34Woo=
 github.com/lib/pq v1.3.0 h1:/qkRGz8zljWiDcFvgpwUpwIAPu3r07TDvs3Rws+o/pU=
 github.com/lib/pq v1.3.0/go.mod h1:5WUZQaWbwv1U+lTReE5YruASi9Al49XbQIvNi/34Woo=
+github.com/lib/pq v1.8.0 h1:9xohqzkUwzR4Ga4ivdTcawVS89YSDVxXMa3xJX3cGzg=
```
youngbupark, line 590: I can see mulitple `pq` pkg versions. can you do go mod tidy?

bindings/postgres/postgres.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/dapr/components-contrib/bindings"
+	"github.com/dapr/dapr/pkg/logger"
+	"github.com/pkg/errors"
+
+	"github.com/jackc/pgx/v4/pgxpool"
+)
+
+// List of operations.
+const (
+	ExecOperation  bindings.OperationKind = "exec"
```
youngbupark, line 23: these constants need to start with lower case. because these will be used in this package.

bindings/postgres/postgres.go
```diff
@@ -0,0 +1,176 @@
...
+	CommandTimeoutDefault = 5
+)
+
+// Binding represents PostgreSQL output binding
+type Binding struct {
+	logger logger.Logger
+	db     *pgxpool.Pool
+}
+
+var _ = bindings.OutputBinding(&Binding{})
```
youngbupark, line 39: why do we need this ?

youngbupark (changes requested):

bindings/postgres/postgres.go
```diff
@@ -0,0 +1,163 @@
...
+
+	p.db, err = pgxpool.ConnectConfig(context.Background(), poolConfig)
+	if err != nil {
+		return errors.Wrap(err, "unable to ping the DB")
+	}
+
+	return nil
+}
+
+// Operations returns list of operations supported by twitter binding
```
mukundansundar, line 65: nitpick: twitter

bindings/postgres/postgres.go
```diff
@@ -0,0 +1,163 @@
...
+
+	connectionURLKey = "url"
+	commandSQLKey    = "sql"
+)
+
+// Postgres represents PostgreSQL output binding
+type Postgres struct {
+	logger     logger.Logger
+	db         *pgxpool.Pool
+	cmdTimeout time.Duration
```
mukundansundar: not used

bindings/postgres/postgres.go
mchmarny: done

bindings/postgres/postgres.go
mchmarny: fixed

bindings/postgres/postgres.go
mchmarny: fixed

bindings/postgres/postgres.go
mchmarny: Actually not needed at all, the connection string can have all these properties so already configurable.

bindings/postgres/postgres.go
mchmarny: removed

bindings/postgres/postgres_test.go
mchmarny: removed

go.sum
mchmarny: This was a result of GH merge, fixed

bindings/postgres/postgres.go
mchmarny: fixed

bindings/postgres/postgres.go
mchmarny: that's a trick to validates adherence to interface, 0 impact at runtime

bindings/postgres/postgres.go
mchmarny: fixed

bindings/postgres/postgres.go
```diff
@@ -0,0 +1,161 @@
...
+	return []bindings.OperationKind{
+		execOperation,
+		queryOperation,
+		closeOperation,
+	}
+}
+
+// Invoke handles all invoke operations
+func (p *Postgres) Invoke(req *bindings.InvokeRequest) (resp *bindings.InvokeResponse, err error) {
+	if req == nil || req.Metadata == nil {
```
mukundans91, line 75: metadata is not required for close operation.

mukundansundar, line 75: metadata is not required for close operation.

mchmarny: fixed

mukundansundar (approved):

bindings/postgres/postgres.go
youngbupark: good to learn :)

youngbupark (approved):

PR #473, opened 2020-09-23
Add get operation for azure blob storage binding

### LINKED ISSUE #469: support Get operation for Azure Blob Storage output binding
Currently Azure Blob Storage output binding only supports `create` operation. We'd like to have also `get` operation supported by the output binding.

## background

We have a scenario with payloads >1MB which cannot be put on Azure Service Bus queues (not even with Premium plan). Instead we want to put payload in blob storage, some metadata into the queue, let the queue trigger a Dapr input binding and with the information in the metadata from there lookup payload from blob storage.

yaron2:
Closes #469

### DIFF
--- a/bindings/azure/blobstorage/blobstorage.go
+++ b/bindings/azure/blobstorage/blobstorage.go
@@ -6,6 +6,7 @@
 package blobstorage
 
 import (
+	"bytes"
 	"context"
 	b64 "encoding/base64"
 	"encoding/json"
@@ -28,6 +29,7 @@ const (
 	contentLanguage    = "ContentLanguage"
 	contentDisposition = "ContentDisposition"
 	cacheControl       = "CacheControl"
+	getBlogRetryCount  = 10
 )
 
 // AzureBlobStorage allows saving blobs to an Azure Blob Storage account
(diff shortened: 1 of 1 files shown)


bindings/azure/blobstorage/blobstorage.go
```diff
@@ -28,6 +29,7 @@ const (
 	contentLanguage    = "ContentLanguage"
 	contentDisposition = "ContentDisposition"
 	cacheControl       = "CacheControl"
+	getBlogRetryCount  = 10
```
mukundansundar, line 32: getBlobRetryCount? unused?

yaron2: Fixed

yaron2:
@mukundansundar please re-review, we'd like to merge this for 0.11.0.

bindings/azure/blobstorage/blobstorage.go
```diff
@@ -28,6 +29,7 @@ const (
 	contentLanguage    = "ContentLanguage"
 	contentDisposition = "ContentDisposition"
 	cacheControl       = "CacheControl"
+	getBlobRetryCount  = 10
```
youngbupark, line 32: I would rename from getBlobRetryCount to defaultGetblobRetryCount. 

wouldn't it make this as metadata to configurable ?

youngbupark (changes requested):

bindings/azure/blobstorage/blobstorage.go
yaron2: Sure, we can do that.

yaron2: Check latest commit.

bindings/azure/blobstorage/blobstorage.go
```diff
@@ -90,6 +91,10 @@ func (a *AzureBlobStorage) parseMetadata(metadata bindings.Metadata) (*blobStora
 	if err != nil {
 		return nil, err
 	}
+
+	if m.GetBlobRetryCount == 0 {
```
yaron2, line 95: This basically means there will *always* be retries, which I think is desired.

youngbupark (approved):

Issue #470, opened 2020-09-23
return effective URL of Azure Blob Storage document created when using an output binding


KaiWalter:
When no `blobName` is specified, [the component auto generates an UUID as blob filename](https://github.com/dapr/docs/blob/master/reference/specs/bindings/blobstorage.md#additional-information). To use the document for further processing it would be nice to have the URL and/or the blob filename of the created document available in the response of the component.

PR #474, opened 2020-09-23
Add blob url to azure blob create

yaron2:
This PR adds the following functionality to the CREATE operation for an Azure Blob using the Azure Blob Storage binding:

* The create operation now returns a response message containing the URL of the created blob.

Closes https://github.com/dapr/components-contrib/issues/470.


youngbupark (dismissed):

bindings/azure/blobstorage/blobstorage.go
```diff
@@ -156,7 +160,17 @@ func (a *AzureBlobStorage) create(blobURL azblob.BlockBlobURL, req *bindings.Inv
 		Metadata:        req.Metadata,
 		BlobHTTPHeaders: blobHTTPHeaders,
 	})
-	return nil, err
+	resp := createResponse{
+		BlobURL: blobURL.String(),
+	}
+	b, err := json.Marshal(resp)
+	if err != nil {
+		return nil, fmt.Errorf("error marshalling create response for azure blob: %s", err)
```
youngbupark, line 168: I am not sure we start using pkg/errors pkg here. do we ? then 
I would use `errors.Errorf` instead of fmr.Errorf

yaron2: We already use it in other places, but not globally. Introducing a change here will need a change in the entire file. What do you think?

youngbupark: I am ok with either way. we can do everything later.

youngbupark (approved):

Issue #480, opened 2020-09-25
Apple Push Notification Service Output Binding


mfcollins3:
## Overview

I would like to propose creating an output binding component for sending push notifications to mobile devices using the Apple Push Notification Service. The output binding component would use the APNs HTTP/2 protocol to communicate directly with the Apple Push Notification Service and not require third-party subscriptions to Azure Notification Hubs or AWS SNS.

## About the Apple Push Notification Service Protocol

Apple's Push Notification Service implements an HTTP/2 API that allows services to send push notifitications to devices that have subscribed to receive push notifications. When an iOS or Mac application registers to receive notifications, Apple generates a device token that the application will send to a service to manage subscriptions. To send push notifications back to applications, the service will need to invoke the Apple Push Notification Service once for each device to receive the notification.

Authorization to send push notifications is done using a JSON Web Token (JWT) that is passed with each HTTP request to send a push notification. The JWT would need to be generated by the service that is sending the push notification and the JWT needs to be cryptographically signed using a private key that is generated by Apple. The JWT has a lifespan of one hour and can be reused for multiple push notifications.

A push notification is represented as a JSON object. A simple push notification is shown below:

```json
{
    "aps": {
        "alert": {
            "title": "New Updates!",
            "body": "New updates are available for your review."
        }
    }
}
```

The `aps` child object contains the push notification details that are used by the Apple Push Notification Service to generate the push notification event for the target application. Additional data can be added as sibling values or objects to the `aps` object and can be used to pass data that will be extracted and used by the target application that receives the push notification. The description of the structure and possible fields for a push notification are documented [here](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/generating_a_remote_notification).

The APNs protocol accepts other metadata about a push notification that are sent via HTTP headers:

* `apns-push-type`: Identifies the content of the notification payload. Can be on of `alert`, `background`, `voip`, `complication`, `fileprovider`, and `mdm`.
* `apns-id`: A UUID value that uniquely identifies the push notification. This can be used to track notifications.
* `apns-expiration`: The date/time at which the notification is no longer necessary. This value is specified as the number of seconds since January 1, 1970 at 00:00 UTC.
* `anps-priority`: If `10`, the notification is sent immediately. If `5`, the notification is sent when possible based on power conditions of the destination device.
* `apns-topic`: The topic for notification. Typically the topic is the bundle identifier of the target app.
[...]

PR #481, opened 2020-09-25
Apns output binding

mfcollins3:
# Description

I implemented an output binding described in #480 to send push notifications to Apple devices using Apple's Push Notification Service.

## Issue reference

Please reference the issue this PR will close: #480 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#830


yaron2:
Welcome to the Dapr community @mfcollins3, I haven't reviewed yet but this looks great.

Be mindful that the documentation will eventually go here: https://github.com/dapr/docs/tree/master/reference/specs/bindings

bindings/apns/apns.go
```diff
@@ -0,0 +1,160 @@
...
+	if value, ok := metadata.Properties[keyIDKey]; ok && value != "" {
+		a.authorizationBuilder.keyID = value
+	} else {
+		return errors.New("the key-id parameter is required")
+	}
+
+	return nil
+}
+
+func (a *APNS) extractTeamID(metadata bindings.Metadata) error {
```
mchmarny, line 132: Would this be more readable? 

```go
func (a *APNS) extractKeyID(metadata bindings.Metadata) error {
	if value, ok := metadata.Properties[keyIDKey]; ok && value != "" {
		a.authorizationBuilder.keyID = value
                return nil
	}
	return errors.New("the key-id parameter is required")
}
```

mfcollins3: Yes. I agree. I corrected it in the code.

bindings/apns/apns.go
```diff
@@ -0,0 +1,159 @@
+package apns
```
youngbupark, line 1: Can you please add license as comment in each code ?

bindings/apns/create_operation.go
```diff
@@ -0,0 +1,148 @@
...
+
+	urlPrefix           string
+	deviceToken         string
+	httpRequest         *http.Request
+	httpResponse        *http.Response
+	authorizationHeader string
+}
+
+func (op *createOperation) run() (*bindings.InvokeResponse, error) {
+	op.getDeviceToken()
```
youngbupark, line 29: What's the benefit of splitting the code to granular methods ?

youngbupark (commented):
Overall looks great :) Thanks for your contribution. I left some comments.

bindings/apns/create_operation.go
yaron2: One benefit I see to breaking down to multiple methods is easier testability, even if we don't unit test all of them.
As long as each function is doing it's own dedicated thing, I think this is fine.

bindings/apns/apns.go
mfcollins3: On the header, do I apply the standard with the Microsoft copyright (from the Kafka binding):

```
// ------------------------------------------------------------
// Copyright (c) Microsoft Corporation.
// Licensed under the MIT License.
// ------------------------------------------------------------
```

or do I leave off the Microsoft Copyright and just say "Licensed under the MIT License"?

yaron2: The headers may eventually blanket change under some governance changes that the project is undergoing, but for now please keep the MS header.

bindings/apns/create_operation.go
youngbupark: I think that pipeline pattern approach is good to handle the data stream or enrich the complicated data. I also used to use this pattern for processing and enriching the data. 

But this PR uses three different methods only to construct http request and save the values into each member variable. We need to maintain member variables in createOperation struct. I am not sure that it is good for maintainability in this use-case. For example, createOperation has `err` which is global error variable in this struct.

There are three parts in this code:
1. construct http header
2. send push notification
3. wraps response

I do not like to have everything in the single method, either :) It seems to have too granular methods. current error handling is not easy to understand the flow of the code.

mfcollins3: The `createOperation.err` variable is used to capture an error that occurs in the pipeline so that I can skip the rest of the steps and properly handle the error at the end and return it. I originally wrote it as a single method inside of `apps.go`, but it just seemed big and bulky and this approach seemed more readable. I thought it might be more acceptable for the PR review for my first contribution. I'm ok with condensing it back into a single function.

How do you want me to proceed?

youngbupark: I definitely agree to have createOperation struct and `run()` method which is good pattern for the other components. But, IMO, having pipeline pattern, granular methods, and member variables definitions in struct would not be easy to understand the code and maintain it. 

We can remove the members and run can help the control of the flow, which would be easier to understand code and maintain. 

```
        req                  *bindings.InvokeRequest
	res                  *bindings.InvokeResponse
	err                  error

	httpRequest         *http.Request
	httpResponse        *http.Response
```

then run() can call the below methods... 

1. makeHTTPRequest
2. sendPushNotification
3. create response based on `sendPushNotification`

@mfcollins3 what do you think?

youngbupark: I am adding the example... The below example would be simpler than using pipeline pattern. If we need to do the series of data processing, then pipeline pattern would be great. but this case is single http request call, so I recommend this way. 

```
type createOperation struct {
	authorizationBuilder *authorizationBuilder
	client               *http.Client
	logger               logger.Logger
}

func (op *createOperation) run() (*bindings.InvokeResponse, error) {
   req, err := op.makeHTTPRequest(...)
   body, err := op.sendPushNotification(req)
   if err != nil {
      return nil, err
   } else {
      // wraps result as bindings.invokeresponse
      return response, nil
   }
}

...

```

mfcollins3: I removed the pipeline code and merged it back into the main output binding type. Please take a look and let me know if you have any further thoughts. Thanks!

youngbupark (changes requested):
Thanks for the change. I have additional small comments.

bindings/apns/apns.go
```diff
@@ -0,0 +1,243 @@
...
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+package apns
+
+import (
+	"bytes"
+	"context"
+	"crypto/x509"
+	"encoding/json"
```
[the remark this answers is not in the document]
mfcollins3: I made the change to jsoniter.

bindings/apns/authorization_builder.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	token := jwt.NewWithClaims(jwt.SigningMethodES256, claims)
+	token.Header["kid"] = a.keyID
+	signedToken, err := token.SignedString(a.privateKey)
+	if err != nil {
+		return "", err
+	}
+
+	a.authorizationHeader = "bearer " + signedToken
+	a.tokenExpiresAt = now.Add(time.Minute * 55)
```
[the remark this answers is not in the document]
mfcollins3: I turned it into a constant and added a comment describing the purpose of it.

youngbupark (approved):
LGTM :) Thanks for your contribution!

yaron2:
@mfcollins3 the next step is to register this on the `dapr/dapr` repo.

Clone the repo, and execute the following:

`go get -u github.com/components-contrib@master`

Then go [here](https://github.com/dapr/dapr/blob/c871b603362e06f9d72d83fee4a6a7d81d50322c/cmd/daprd/main.go#L313) and register the new output binding.

Issue #486, opened 2020-09-29
Force HTTPS on final redirect from oauth2 middleware


TenSt:
**Case:** A web-app which is exposed over HTTPS and access API that are behind darp + oauth2 middleware.

**Problem:** 'Location' header in the final redirect from dapr oauth2 middleware contains HTTP and redirect request is blocked by browser with mixed-content error.

**Solution:** add new key to metadata. If key is true - force HTTPS on final redirect to not get blocked by browser.

**Justification:** A case when web-app communicates with api or some other backend over HTTPS is pretty common. dapr does a very good job on protecting API, especially in conjunction with Oauth2 middleware. The issue is that dapr doesn't operate over HTTPS. HTTPS request to dapr can be handled by Nginx tls configuration or something similar (it verifies that connection is secure and then changes scheme to HTTP), but in the final redirect dapr oauth2 middleware returns 'Location' header with 'http' as a scheme and that leads to mixed-content errors in browser.

**Code**: I already have a POC written and verified in the k8s environment, so can create pull request to cover this.

yaron2:
Thanks for raising this, a PR would be welcome.

TenSt:
Thanks @yaron2. I've created new PR and issue in the docs repo to cover it.

PR #487, opened 2020-09-29
Adding force https metadata key

TenSt:
# Description

Added `ForceHTTPS` key to metadata and simple if statement that will handle forcing https in the Location header of the final redirect from oauth2 middleware. 

## Issue reference

Reference to the issue this PR will close: #486

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests - there were no tests for this middleware, so I didn't created any new ones.
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#836
* [x] Tested in the k8s environment with the latest stable dapr release v0.10.0.


middleware/http/oauth2/oauth2_middleware.go
```diff
@@ -81,6 +82,9 @@ func (m *Middleware) GetHandler(metadata middleware.Metadata) (func(h fasthttp.R
 			} else {
 				authState := session.GetString(savedState)
 				redirectURL := session.GetString(redirectPath)
+				if meta.ForceHTTPS == "true" {
```
youngbupark, line 85: I would change ForceHTTP to lowercase first and compare with `"true"`

middleware/http/oauth2/oauth2_middleware.go
```diff
@@ -81,6 +82,9 @@ func (m *Middleware) GetHandler(metadata middleware.Metadata) (func(h fasthttp.R
...
+					redirectURL = "https://" + string(ctx.Request.Host()) + redirectURL
```
youngbupark, line 86: Can we make "http://" const?

middleware/http/oauth2/oauth2_middleware.go
```diff
@@ -81,6 +82,9 @@ func (m *Middleware) GetHandler(metadata middleware.Metadata) (func(h fasthttp.R
...
+				}
```
youngbupark, line 87: Please add unit-test for this change.

youngbupark (changes requested):
Thanks for your contribution! I left some comments. please add unit-test as well.

middleware/http/oauth2/oauth2_middleware.go
TenSt: Done.

middleware/http/oauth2/oauth2_middleware.go
TenSt: Done. Added const `https`.

middleware/http/oauth2/oauth2_middleware.go
TenSt: Working on it.

TenSt: Hi @youngbupark. Could you please give me advice on how unit test for this should be done? This is a Authorization Code Flow of OAuth2 protocol which should be done via browser to pass user credentials directly to Authorization Provider without getting them to the app. When I was testing it locally I've setup whole infrastructure in order to do that. There is no existing unit tests for this middleware, so I can't mimic existing tests or extend them. Maybe I'm looking at the wrong direction and don't see some obvious ways to do this. Advice or help would be appreciated.

youngbupark: Thanks for your investigation and your time. The initial ask is to test the below code by separating it as separate method or using mock:

```
redirectURL := session.GetString(redirectPath)
				if meta.ForceHTTPS == "true" {
					redirectURL = "https://" + string(ctx.Request.Host()) + redirectURL
				}
```

If it is difficult to write the test without the actual backend service support and you validate it locally, I am ok to merge it now.  @yaron2 any objection?

JFYI - We are planning to have conformance test. we will provide the common test infra for middleware test later. @mukundansundar

youngbupark (approved):

yaron2:
Thanks for this @TenSt

TenSt:
@yaron2 it was my pleasure

Issue #515, opened 2020-11-03
Azure blob storage binding doesn't save JSON values


yaron2:
First reported by @KaiWalter 

When saving a JSON payload to an Azure Blob Storage binding, an error is returned for "invalid syntax".

PR #516, opened 2020-11-03
Fix azure blob storage json payload

yaron2:
Closes https://github.com/dapr/components-contrib/issues/515

### DIFF
--- a/bindings/azure/blobstorage/blobstorage.go
+++ b/bindings/azure/blobstorage/blobstorage.go
@@ -12,6 +12,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/url"
+	"strconv"
 
 	"github.com/Azure/azure-storage-blob-go/azblob"
 	"github.com/dapr/components-contrib/bindings"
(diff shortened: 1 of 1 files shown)


bindings/azure/blobstorage/blobstorage.go
```diff
@@ -138,25 +138,21 @@ func (a *AzureBlobStorage) create(blobURL azblob.BlockBlobURL, req *bindings.Inv
 		delete(req.Metadata, cacheControl)
 	}
 
-	// Unescape data which will still be a JSON string
-	unescapedData, unescapeError := strconv.Unquote(string(req.Data))
-
-	if unescapeError != nil {
-		return nil, unescapeError
+	d, err := strconv.Unquote(string(req.Data))
```
youngbupark, line 141: If json serializer serializes `string` data, it adds double-quotes `"string"`. If we want to save only `string`, not a json-serialized one, then this is correct. is this what we want ?

yaron2: Yes. If a JSON string is sent, it's saved as is. I validated it.

youngbupark (approved):

PR #525, opened 2020-11-17
Various AWS improvements

trondhindenes:
# Description
- Implemented support for AWS session token (https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html) in the shared `authentication/aws` function
- Cleaned up the sns/sqs pubsub attributes so that its configuration attribute set matches the rest of the aws resources (created an alias lookup thing so that the old attribute names still work, added tests for that)

As far as I can see, the existing sns/sqs pubsub binding simply cannot have worked - the way it mixes account id and access key ist just... strange. In any case, this PR attempts to fix all of that so that all aws-related things can be configured in the same way. I'm not sure if the aliasing is really needed, as I strongly doubt anyone using it in its current state, but I figure it never hurts to be backwards-compatible anyway.

I've functinally tested both the `sqssns` pubsub, and a few of the bindings as well with various aws auth methods (explicit, implicit, etc), and everything now seems to work as it should.

I'll create documentation in a separate PR - I want to create a shared "aws auth" documentation page for all aws components to refer to.
## Checklist
* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#934

### DIFF
--- a/pubsub/aws/snssqs/snssqs.go
+++ b/pubsub/aws/snssqs/snssqs.go
@@ -23,7 +23,7 @@ type snsSqs struct {
 	topicHash map[string]string
 	// key is the topic name, value holds the ARN of the queue and its url
 	queues        map[string]*sqsQueueInfo
-	awsAcctID     string
+	// awsAcctID     string
 	snsClient     *sns.SNS
 	sqsClient     *sqs.SQS
 	metadata      *snsSqsMetadata
--- a/pubsub/aws/snssqs/snssqs.go
+++ b/pubsub/aws/snssqs/snssqs.go
@@ -41,17 +41,15 @@ type snsSqsMetadata struct {
 	sqsQueueName string
 
 	// aws endpoint for the component to use.
-	awsEndpoint string
-	// aws account ID to use for SNS/SQS. Required
-	awsAccountID string
-	// aws secret corresponding to the account ID. Required
-	awsSecret string
-	// aws token to use. Required
-	awsToken string
-	// aws sessino token to use.
-	awsSessionToken string
-	// aws region in which SNS/SQS should create resources. Required
-	awsRegion string
+	Endpoint string
+	// access key to use for accessing sqs/sns
+	AccessKey string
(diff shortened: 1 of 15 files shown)


trondhindenes:
Also fixes missing logger initialization in `bindings/aws/sqs/sqs.go` (https://github.com/dapr/components-contrib/issues/529)

yaron2:
@trondhindenes could you build this into `daprd` and test at least several AWS components here to make sure we don't have any regressions? we are working on an automated infrastructure for this, but currently we need to test this manually.

trondhindenes:
@yaron2 yup, I already did. Tested a few bindings, and also a few of the other components using multiple auth methods. Everything seems a-ok.

pubsub/aws/snssqs/snssqs.go
```diff
@@ -41,23 +40,23 @@ type snsSqsMetadata struct {
...
-	awsEndpoint string
-	// aws account ID to use for SNS/SQS. Required
-	awsAccountID string
-	// aws secret corresponding to the account ID. Required
-	awsSecret string
-	// aws token to use. Required
-	awsToken string
-	// aws region in which SNS/SQS should create resources. Required
-	awsRegion string
+	Endpoint string
```
mukundansundar, line 43: i think changing the name here makes this a breaking-change. 
cc @yaron2 ?
the names make it consistent with other aws components but the consistency is lost within the same struct. 
Should we consider capitalizing the other properties also?

yaron2: If a field needs to be exported, it should be capitalized. that's the Go rule..

trondhindenes: @mukundansundar yes it's a breaking change. However, I've implemented an aliasing thingy so Component manifests using the old attribute names should still work.

yaron2 (approved):
lgtm

PR #547, opened 2020-12-04
Add watch method into stateStore

cvictory:
# Description

Maybe we need add watch method into StateStore.  I think using gRpc server streaming is the better way.

## Issue reference

https://github.com/dapr/dapr/issues/2531

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


pkedy (commented):
For StateStores that don't implement `Watch`, there should not one that returns an "unimplemented" error. Instead, I would recommend looking at how [BulkStateStore](https://github.com/dapr/components-contrib/blob/master/state/store.go#L18-L23) is broken out from the basic [`StateStore`](). I don't think that the pattern should be followed exactly because `DefaultBulkStore` is used to make [`BulkSet`](https://github.com/dapr/components-contrib/blob/master/state/store.go#L50-L60) fallback to calling `Set` for each key. `Watch` is a completely different concept/feature. Speaking of features, it might make since to add a `Watch` as a feature. @artursouza WDYT?

artursouza:
> For StateStores that don't implement `Watch`, there should not one that returns an "unimplemented" error. Instead, I would recommend looking at how [BulkStateStore](https://github.com/dapr/components-contrib/blob/master/state/store.go#L18-L23) is broken out from the basic `StateStore`. I don't think that the pattern should be followed exactly because `DefaultBulkStore` is used to make [`BulkSet`](https://github.com/dapr/components-contrib/blob/master/state/store.go#L50-L60) fallback to calling `Set` for each key. `Watch` is a completely different concept/feature. Speaking of features, it might make since to add a `Watch` as a feature. @artursouza WDYT?

The Feature list is only used if that affects how the state store is used - I don't think this feature should be optional. Instead, it should be implemented via busy waiting for the "unsupported" state stores. Although busy waiting is not ideal in production, it can still be useful for dev environments. This is important because "GA" components cannot partially implement the API.

state/zookeeper/zk.go
```diff
@@ -169,6 +171,33 @@ func (s *StateStore) Get(req *state.GetRequest) (*state.GetResponse, error) {
...
+	if _, _, events, err = s.conn.GetW(s.prefixedKey(req.Key)); err != nil {
+		return err
+	}
+	go func() {
+		for e := range events {
+			if req.Key != e.Path {
+				continue
+			}
+			if resp, err := s.Get(req); err != nil {
+				handler(&state.GetResponse{
```
artursouza, line 189: Shouldn't this be in some sort of infinite loop to recover in case of error.

Also, this feels a lot like a PubSub - how can the subscriber stop this? For example, what if the client (app) restarts?

pkedy:
> I don't think this feature should be optional. Instead, it should be implemented via busy waiting for the "unsupported" state stores. Although busy waiting is not ideal in production, it can still be useful for dev environments. This is important because "GA" components cannot partially implement the API.

@artursouza I'm thinking the more idiomatic approach would be to have a `Watcher` interface that embeds the `StateStore` interface so that is it not necessary for components that do not implement watching (like etcd) do not have to implement it. The default response via the API should be implemented by testing if the components implements `Watcher`. If it does not, then there is a default response returned to the application and it is coded in a single place.

@artursouza I also agree that `Watch` looks a lot like pubsub or input bindings. That got me thinking... Not saying that this should implemented BUT given the "interface driven" approach above, it would be possible for a component to implement 1 or many component types. `runtime.go` might be the only thing to change to make that happen. An interesting thought experiment if nothing else.

artursouza:
> > I don't think this feature should be optional. Instead, it should be implemented via busy waiting for the "unsupported" state stores. Although busy waiting is not ideal in production, it can still be useful for dev environments. This is important because "GA" components cannot partially implement the API.
> 
> @artursouza I'm thinking the more idiomatic approach would be to have a `Watcher` interface that embeds the `StateStore` interface so that is it not necessary for components that do not implement watching do not have to implement it. The default response via the API should be implemented by testing if the components implements `Watcher`. If it does not, then there is a default response returned to the application and it is coded in a single place.
> 

I like that approach. It is similar to how transactional feature is implemented. Feature list in state store can be used here to define if the component has this feature and invoke the Watch method vs using the out of the box watch implementation.

> @artursouza I also agree that `Watch` looks a lot like pubsub or input bindings. That got me thinking... Not saying that this should implemented BUT given the "interface driven" approach above, it would be possible for a component to implement 1 or many component types. `runtime.go` might be the only thing to change to make that happen. An interesting thought experiment if nothing else.
That might be possible. I think the change would go beyond runtime.go because the components are organized by type, so breaking that might require reorganizing the code (not 100% of that yet). The way to avoid "reorganizing" the code would be to pick a "primary" building block type to keep the component code. Optionally we can still keep the per-building-block struct but just as an "alias" to the main code - this can improve discoverability in the code.

This actually made me think of another similar approach. What if we don't implement the Watch functionality with a new interface but via PubSub interface instead. Then, we would also need a wrapper that converts any state store (that does not implement PubSub interface) to become one (with busy waiting probably).

pkedy:
@artursouza So I understand correctly, we are saying to both:

1) Break out the `Watch` method into a separate interface AND
2) Add a feature so that there is a "metadata-based" way to determine if `Watch` is supported by the component?

Just wanted to double check we are on the same page because before you commented:

> The Feature list is only used if that affects how the state store is used - I don't think this feature should be optional.

but since the default response would be handled in the API layer that is addressed. Correct?

artursouza:
> @artursouza So I understand correctly, we are saying to both:
> 
> 1. Break out the `Watch` method into a separate interface AND
> 2. Add a feature so that there is a "metadata-based" way to determine if `Watch` is supported by the component?
> 
> Just wanted to double check we are on the same page because before you commented:
> 
> > The Feature list is only used if that affects how the state store is used - I don't think this feature should be optional.
> 
> but since the default response would be handled in the API layer that is addressed. Correct?

Let's proceed with that approach for now. We don't need to implement Watch on all GA components because a component does not need to have every feature implemented to be considered GA.

pkedy:
@cvictory Just a friendly ping. :)  Are you able to make the changes described in the above comments?

pkedy:
@cvictory Ping :)

Issue #560, opened 2020-12-17
Add prefetch count to RabbitMQ pub/sub


yaron2:
Dapr today has prefetch count implemented for the RabbitMQ binding. The same functionality should be applied to the RabbitMQ pub/sub implementation.

/cc @trondhindenes

trondhindenes:
I'll try my hand at this one.

trondhindenes:
sweet! I'll get on the doc PR this week.

yaron2:
Closed via #562

PR #562, opened 2020-12-17
Add support for prefetch

trondhindenes:
# Description
Adds support for the "prefetch" parameter in rabbitmq, allowing clients to only prefetch a portion of the available messages, instead of all of them.

I can do a separate PR to update rabbitmq docs according to https://v1-rc2.docs.dapr.io/operations/components/component-schema/

I'm not sure if I should just switch to intsince that's what the Qos function expects natively. I figured I'd use the same numeric format as other parameters in the same module. Happy to change that.

I'm also not sure if it should default to 0 (which disables the prefetch functionality) or 1 (which IMHO is the safe and "correct" setting). 

## Issue reference
https://github.com/dapr/components-contrib/issues/560

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/pubsub/rabbitmq/metadata.go
+++ b/pubsub/rabbitmq/metadata.go
@@ -15,8 +15,8 @@ type metadata struct {
 	autoAck          bool
 	requeueInFailure bool
 	deliveryMode     uint8 // Transient (0 or 1) or Persistent (2)
-	reconnectWait    time.Duration
 	prefetchCount    uint8 // Prefetch deactivated if 0
+	reconnectWait    time.Duration
 }
 
 // createMetadata creates a new instance from the pubsub metadata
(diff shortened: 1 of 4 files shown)


trondhindenes:
I'm not sure if I should just switch to `int`since that's what the `Qos` function expects natively. I figured I'd use the same numeric format as other parameters in the same module. Happy to change that.

trondhindenes:
I can do a separate PR to update rabbitmq docs according to https://v1-rc2.docs.dapr.io/operations/components/component-schema/

yaron2:
@trondhindenes the default setting for RabbitMQ is not to enable prefetching, that is true for all of its SDKs. How important do you think it is we set this to 1? I can see the benefits, however I can also see some expected performance hits for users that just moved their default RabbitMQ code to Dapr.

trondhindenes:
yup - lets default to zero, and then I can update docs with "recommended settings for production" or something. The rabbitmq documentation covers this very well, so we can probably just link there for more info.

yaron2 (approved):
LGTM

PR #565, opened 2020-12-23
TTL in PubSub.

artursouza:
# Description

Added TTL in PubSub.

## Issue reference

Please reference the issue this PR relates to: https://github.com/dapr/dapr/issues/2216

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

### DIFF
--- a/metadata/utils.go
+++ b/metadata/utils.go
@@ -3,7 +3,7 @@
 // Licensed under the MIT License.
 // ------------------------------------------------------------
 
-package bindings
+package metadata
 
 import (
 	"fmt"
@@ -18,7 +18,7 @@ const (
 	TTLMetadataKey = "ttlInSeconds"
 )
 
-// TryGetTTL tries to get the ttl (in seconds) value for a binding
+// TryGetTTL tries to get the ttl (in seconds) value for pubsub, binding and any other building block.
 func TryGetTTL(props map[string]string) (time.Duration, bool, error) {
 	if val, ok := props[TTLMetadataKey]; ok && val != "" {
 		valInt, err := strconv.Atoi(val)
--- a/metadata/utils.go
+++ b/metadata/utils.go
@@ -7,6 +7,7 @@ package metadata
 
 import (
 	"fmt"
+	"math"
 	"strconv"
 	"time"
 
@@ -18,19 +19,25 @@ const (
 	TTLMetadataKey = "ttlInSeconds"
 )
 
-// TryGetTTL tries to get the ttl (in seconds) value for pubsub, binding and any other building block.
+// TryGetTTL tries to get the ttl as a time.Duration value for pubsub, binding and any other building block.
 func TryGetTTL(props map[string]string) (time.Duration, bool, error) {
 	if val, ok := props[TTLMetadataKey]; ok && val != "" {
-		valInt, err := strconv.Atoi(val)
+		valInt64, err := strconv.ParseInt(val, 10, 64)
 		if err != nil {
 			return 0, false, errors.Wrapf(err, "%s value must be a valid integer: actual is '%s'", TTLMetadataKey, val)
 		}
 
-		if valInt <= 0 {
-			return 0, false, fmt.Errorf("%s value must be higher than zero: actual is %d", TTLMetadataKey, valInt)
+		if valInt64 <= 0 {
+			return 0, false, fmt.Errorf("%s value must be higher than zero: actual is %d", TTLMetadataKey, valInt64)
 		}
 
-		return time.Duration(valInt) * time.Second, true, nil
+		duration := time.Duration(valInt64) * time.Second
+		if duration < 0 {
+			// Overflow
+			duration = math.MaxInt64
+		}
+
+		return duration, true, nil
 	}
 
 	return 0, false, nil
--- a/pubsub/Readme.md
+++ b/pubsub/Readme.md
@@ -25,3 +25,44 @@ type PubSub interface {
 	Subscribe(req SubscribeRequest, handler func(msg *NewMessage) error) error
 }
 ```
+
+### TTL (or Time To Live)
+
+Time to live is implemented by default in Dapr. A publishing application can add the `expiration` attribute to a CloudEvent object and Dapr runtime will drop the message if the expiration is due at the subscriber's end, prior to invoking subscriber applications. The `expiration` attribute is Dapr specific and extends the CloudEvent spec. This is a convenience to subscriber applications, so Dapr does not expire messages from pub subs that do not have subscribers via Dapr.
+
+Alternatively, the publisher application can set the expiration by publishing with the `ttlInSeconds` metadata. Components that support TTL should parse this metadata attribute. Dapr runtime will also automatically populate the `expiration` attribute in the CloudEvent object if `ttlInSeconds` is present - in this case, Dapr will add 1 second to allow component implementations to expire the message before Dapr consumes it.
+
+As per the CloudEvent spec, `expiration` is formatted using RFC3339.
+
+Example:
+```json
+{
+	"specversion" : "1.0",
+	"type" : "com.github.pull.create",
+	"source" : "https://github.com/cloudevents/spec/pull",
+	"subject" : "123",
+	"id" : "A234-1234-1234",
+	"expiration" : "2018-04-06T17:31:00Z",
+	"datacontenttype" : "text/xml",
+	"data" : "<much wow=\"xml\"/>"
+}
+```
+
+If the pub sub component implementation can handle TTL natively without relying on Dapr, consume the `ttlInSeconds` metadata in the component implementation for the Publish function.
+
+Example:
+```go
+import contrib_metadata "github.com/dapr/components-contrib/metadata"
+
+//...
+
+func (c *MyComponent) Publish(req *pubsub.PublishRequest) error {
+	//...
+	ttl, hasTTL, _ := contrib_metadata.TryGetTTL(req.Metadata)
+	if hasTTL {
+		//... handle ttl for component.
+	}
+	//...
+	return nil
+}
+```
--- a/pubsub/Readme.md
+++ b/pubsub/Readme.md
@@ -26,29 +26,11 @@ type PubSub interface {
 }
 ```
 
-### TTL (or Time To Live)
+### Message TTL (or Time To Live)
 
-Time to live is implemented by default in Dapr. A publishing application can add the `expiration` attribute to a CloudEvent object and Dapr runtime will drop the message if the expiration is due at the subscriber's end, prior to invoking subscriber applications. The `expiration` attribute is Dapr specific and extends the CloudEvent spec. This is a convenience to subscriber applications, so Dapr does not expire messages from pub subs that do not have subscribers via Dapr.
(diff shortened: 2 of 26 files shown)


pubsub/Readme.md
```diff
@@ -25,3 +25,44 @@ type PubSub interface {
...
+
+Example:
+```json
+{
+	"specversion" : "1.0",
+	"type" : "com.github.pull.create",
+	"source" : "https://github.com/cloudevents/spec/pull",
+	"subject" : "123",
+	"id" : "A234-1234-1234",
+	"expiration" : "2018-04-06T17:31:00Z",
```
artursouza, line 45: To make this feature easier to understand, I am considering to rename `expiration` to `daprExpiration` and not document it as something users can set from application side (or SDK). So, expiration would always come from the `ttlInSeconds` metadata.

wcs1only: Agreed, I don't think there is a good reason to have both expiration and ttlInSeconds exposed to the user. They are conceptually easy to get mixed up.

pubsub/Readme.md
```diff
@@ -25,3 +25,44 @@ type PubSub interface {
 	Subscribe(req SubscribeRequest, handler func(msg *NewMessage) error) error
 }
 ```
+
+### TTL (or Time To Live)
+
+Time to live is implemented by default in Dapr. A publishing application can add the `expiration` attribute to a CloudEvent object and Dapr runtime will drop the message if the expiration is due at the subscriber's end, prior to invoking subscriber applications. The `expiration` attribute is Dapr specific and extends the CloudEvent spec. This is a convenience to subscriber applications, so Dapr does not expire messages from pub subs that do not have subscribers via Dapr.
+
+Alternatively, the publisher application can set the expiration by publishing with the `ttlInSeconds` metadata. Components that support TTL should parse this metadata attribute. Dapr runtime will also automatically populate the `expiration` attribute in the CloudEvent object if `ttlInSeconds` is present - in this case, Dapr will add 1 second to allow component implementations to expire the message before Dapr consumes it.
```
artursouza, line 33: This 1 second delay is an arbitrary value to give time between the calculation of `expiration` in Dapr and the moment the event is published to the underlying component. If there is a big delay between those, Dapr will "see" expired messages.

wcs1only: So if I publish a message with ttlInSeconds = 1, will I be guaranteeing that it will never be delivered?

wcs1only: I'm unclear with exactly what is meant by
```
Dapr will add 1 second to allow component implementations to expire the message before Dapr consumes it.
```
Do I have this backwards? Will we end up delivering messages that should have been expired?

wcs1only: If we were to use ms precision, would we be able to avoid this fudging of the timestamp?

metadata/utils.go
```diff
@@ -18,7 +18,7 @@ const (
 	TTLMetadataKey = "ttlInSeconds"
 )
 
-// TryGetTTL tries to get the ttl (in seconds) value for a binding
+// TryGetTTL tries to get the ttl (in seconds) value for pubsub, binding and any other building block.
```
rynowak, line 21:
```suggestion
// TryGetTTL tries to get the ttl as a time.Duration value for pubsub, binding and any other building block.
```

Comment is stale 😆

pubsub/Readme.md
```diff
@@ -25,3 +25,44 @@ type PubSub interface {
 	Subscribe(req SubscribeRequest, handler func(msg *NewMessage) error) error
 }
 ```
+
+### TTL (or Time To Live)
+
+Time to live is implemented by default in Dapr. A publishing application can add the `expiration` attribute to a CloudEvent object and Dapr runtime will drop the message if the expiration is due at the subscriber's end, prior to invoking subscriber applications. The `expiration` attribute is Dapr specific and extends the CloudEvent spec. This is a convenience to subscriber applications, so Dapr does not expire messages from pub subs that do not have subscribers via Dapr.
```
rynowak, line 31: > A publishing application can add the `expiration` attribute to a CloudEvent object and Dapr runtime will drop the message if the expiration is due at the subscriber's end, prior to invoking subscriber applications. The `expiration` attribute is Dapr specific and extends the CloudEvent spec.

Where is this documented for **applications**? IIRC - Applications don't use cloud events to *send* pub/sub messages - they receive cloud events as the when handling a pub/sub message.

pubsub/Readme.md
rynowak: I'm a little confused about why we'd use `daprExpiration` rather than `expiration` - and also who's setting it in this example - is this and example of the payload produced by a pub/sub component?

artursouza: @rynowak Applications can publish in two ways: send Dapr a CloudEvent (which, in this case, the application could also set the `expiration` attribute) or send the payload and let Dapr automatically wrap it into a CloudEvent. So, `expiration` can be set by Application. Dapr also supports (as part of this PR) the `ttlInSeconds` metadata, so that could also set the `expiration` attribute in Dapr.

pubsub/Readme.md
artursouza: This has nothing to do with clock precision. We use `time.Now()`, which as per Golang's documentation: `A Time represents an instant in time with nanosecond precision.`

The 1s delay to expire is to avoid a situation where the component and Dapr are in a "race" to expire the message. Imagine that there is a ttl of 1 minute and th at message is being published to ServiceBus with a ttl of 1 minute. Dapr with calculate an expiration of now() + 1 minute and pass that message to the ServiceBus component that will publish the event with a ttl param value of 1 minute. Now, will Dapr get the message right before ServiceBus expires it or will Service Bus expire the message first? The 1s delay is to give Dapr a second priority when both the component and Dapr  can expire a message.

pubsub/Readme.md
artursouza: @rynowak Actually, Dapr supports both. If the payload is a CloudEvent, Dapr will not double wrap it. There was an issue about that a while ago.

metadata/utils.go
artursouza: OK. Will do.

pubsub/Readme.md
withinboredom: > Dapr will not double wrap it.

It does remove the `time` attribute though, dunno if that's a bug.

pubsub/envelope.go
```diff
@@ -110,6 +116,29 @@ func NewCloudEventsEnvelope(id, source, eventType, subject string, topic string,
...
+
+	return false
+}
+
+// ApplyMetadata will process metadata to modify the cloud event accordingly
+func (cloudEvent *CloudEventsEnvelope) ApplyMetadata(metadata map[string]string) {
+	ttl, hasTTL, _ := contrib_metadata.TryGetTTL(metadata)
+	if hasTTL {
+		// Adds an extra second to allow components to expire message before Dapr.
+		cloudEvent.Expiration = time.Now().UTC().Add(ttl).Add(time.Second).Format(time.RFC3339)
```
withinboredom, line 138: Potential corner-case: when the time is set to the maximum possible time, adding one second would overflow to the past. Programmers tend to use constants like that to indicate that a message should never expire.

pubsub/Readme.md
artursouza: I was fixing that in this PR too but changed my mind to keep the scope within the issue here. I do believe that is a bug. I will create an issue for that.

pubsub/envelope.go
artursouza: Good point.

pubsub/envelope.go
```diff
@@ -110,6 +116,29 @@ func NewCloudEventsEnvelope(id, source, eventType, subject string, topic string,
...
+		expiration, err := time.Parse(time.RFC3339, cloudEvent.Expiration)
+		if err != nil {
+			return false
+		}
+
+		return expiration.UTC().Before(time.Now().UTC())
+	}
+
+	return false
+}
```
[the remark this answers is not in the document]
artursouza: We are not implementing deadletter queue in Dapr yet, just TTL. I like the idea of Dapr not doing any manual expiration if component supports it natively. I wonder what would be the best way to implement that. It would be a change in how Dapr implements building blocks, where the common logic in Dapr changes based on the component.

pubsub/Readme.md
artursouza: https://github.com/dapr/dapr/issues/2611

pubsub/envelope.go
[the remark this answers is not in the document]
yaron2: > It would be a change in how Dapr implements building blocks, where the common logic in Dapr changes based on the component.

We are already doing this with Bulk Get. If a component natively supports a Bulk Get operation, the component takes over. if not, it signals to Dapr that it doesn't support it and then Dapr operates it's own logic.

This would be the same here: if the component natively supports TTL it would do it, and if not it would signal to Dapr it doesn't have this capability and so Dapr will expire the message when receiving it.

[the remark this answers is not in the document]
artursouza: OK. I like this approach.

pubsub/Readme.md
artursouza: I kept `expiration` as the name. It is less verbose.

artursouza: Also, `expiration` is not dropped from the CloudEvent if the applications injects it. It is only set by Dapr via `ttlInSeconds` metadata.

pubsub/Readme.md
artursouza: This logic is gone now. If component implements message TTL feature, Dapr does not.

pubsub/envelope.go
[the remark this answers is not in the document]
artursouza: This logic has been implemented now. Dapr will not do any expiration if component implementation handles that.

pubsub/envelope.go
artursouza: I actually found an overflow issue in the logic to parse Duration from seconds. It is virtually impossible for an overflow to happen here because the max duration is `maxInt64`, which is not enough to overflow `Now().Add(ttl)` for looong time.

pubsub/envelope.go
```diff
@@ -110,6 +115,40 @@ func NewCloudEventsEnvelope(id, source, eventType, subject string, topic string,
...
+		expiration := now.Add(ttl)
+		// The maximum duration (maxInt64) is not enough to overflow max time, for now.
+		// As of the time this code was written (2020 Dec 28th),
+		// the maximum time of now() adding maxInt64 is ~ "2313-04-09T23:30:26Z".
+		// Max time in golang is currently 292277024627-12-06T15:30:07.999999999Z.
+		// So, we have some time before the overflow below happens.
+		if (ttl.Seconds() > 0) && expiration.Before(now) {
+			// Overflow
+			expiration = time.Unix(1<<63-62135596801, 999999999).UTC()
+		}
```
wcs1only, line 147: Would it be better to throw an error of some kind rather than silently set this to max time? If a users code is overflowing, that's probably an indicator of a bug in the user code. Shouldn't we expose this error sooner, rather than make the user investigate why their messages are never expiring?

Also, is there a constant available for these magic numbers?

wcs1only: After carefully re-reading this comment, is the assertion that this is code we will never actually hit? If so, shouldn't we panic here?

artursouza: This code will "never" run. I actually thought of leaving the comment explaining why the overflow will unlikely happen and remove this if clause.

wcs1only (approved):
LGTM

artursouza:
I am merging this as I need this change to merge my PR in the dapr/dapr repo. If there is any other comments, I can address in a separate PR. This change is aligned to what we discussed in our last maintainers meeting.

PR #566, opened 2020-12-30
Add method to check if payload is a cloudevent

yaron2:
This PR adds a dedicated method to check if a payload is a Cloud Event.

### DIFF
--- a/pubsub/envelope.go
+++ b/pubsub/envelope.go
@@ -78,29 +78,6 @@ func NewCloudEventsEnvelope(id, source, eventType, subject string, topic string,
 		}
 	}
 
-	// handle CloudEvent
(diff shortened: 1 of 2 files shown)


artursouza:
Can we use content type header too? Maybe we can skip the parsing if content header is correct.

"content-type":"application/cloudevents+json"


/cc @rynowak

yaron2:
> Can we use content type header too? Maybe we can skip the parsing if content header is correct.
> 
> "content-type":"application/cloudevents+json"
> 
> /cc @rynowak

I will do this check from `daprd`, would rather keep this method focused on the spec parsing.

artursouza (approved):

PR #567, opened 2020-12-30
Replace content cloud event check, add settraceid

yaron2:
This PR replaces the body check to determine if a payload is a CloudEvent or not by relying on `daprd` to make the decision based on a user supplied information, either via a header on HTTP or a special field in proto.

Determining a cloud event by inspecting the body has the following drawbacks:

1. Performance hit in terms of CPU
2. The method to determine if a Cloud Event is compliant to the spec will need to be maintained long term to keep up with any changes to the spec

In addition, the PR corrects a long time error in treating the `subject` field as a way to store the trace ID.
Based on this change, trace IDs will be saved in a `daprtraceid` extension field.
A method has been added to add the trace field to existing CE json based payloads.

### DIFF
--- a/pubsub/envelope.go
+++ b/pubsub/envelope.go
@@ -6,6 +6,7 @@
 package pubsub
 
 import (
+	"fmt"
 	"time"
 
 	contrib_metadata "github.com/dapr/components-contrib/metadata"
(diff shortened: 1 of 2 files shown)


pubsub/envelope.go
```diff
@@ -24,6 +24,7 @@ const (
 	DefaultCloudEventSource = "Dapr"
 	// DefaultCloudEventDataContentType is the default content-type for the data attribute
 	DefaultCloudEventDataContentType = "text/plain"
+	DaprTraceIDField                 = "daprtraceid"
```
artursouza, line 27: We have chosen not to name `expiration` as `daprexpiration`. Why is this field different?

pubsub/envelope.go
```diff
@@ -89,28 +89,26 @@ func NewCloudEventsEnvelope(id, source, eventType, subject string, topic string,
...
+		DaprTraceID:     traceID,
 	}
 }
 
-// IsCloudEvent returns a bool to indicate if a payload is a CloudEvents compliant format
-func IsCloudEvent(data []byte) bool {
-	var j interface{}
-	err := jsoniter.Unmarshal(data, &j)
+// SetTraceID adds the daprtraceid field to an existing cloudevents JSON
+func SetTraceID(data []byte, traceID string) ([]byte, error) {
```
artursouza, line 97: This method seems to be useful only for tests, please move it to a test specific place. I would like to avoid production code messing with the payload after it has been serialized to bytes.

pubsub/envelope.go
yaron2: I'll change it to match

artursouza: Which field will change?

pubsub/envelope.go
```diff
@@ -53,70 +41,53 @@ func NewCloudEventsEnvelope(id, source, eventType, subject string, topic string,
...
+		"traceid":         traceID,
 	}
 }
 
-// IsCloudEvent returns a bool to indicate if a payload is a CloudEvents compliant format
-func IsCloudEvent(data []byte) bool {
-	var j interface{}
-	err := jsoniter.Unmarshal(data, &j)
+// SetTraceID adds the daprtraceid field to an existing cloudevents JSON
+func SetTraceID(data []byte, traceID string) ([]byte, error) {
```
artursouza, line 69: Now that cloudevent is just a map, is this method even needed?

artursouza (approved):

PR #568, opened 2020-12-31
Add new cloudevents creation method

yaron2:
This PR does a small refactor that introduces a new method that returns a Cloud Event map based on an existing Cloud Event.

### DIFF
--- a/pubsub/envelope.go
+++ b/pubsub/envelope.go
@@ -74,6 +74,7 @@ func FromCloudEvent(cloudEvent []byte, traceID string) (map[string]interface{}, 
 	}
 
 	setTraceContext(m, traceID)
+
 	return m, nil
 }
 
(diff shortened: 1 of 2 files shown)


pubsub/envelope.go
```diff
@@ -65,8 +65,19 @@ func NewCloudEventsEnvelope(id, source, eventType, subject string, topic string,
...
-func SetTraceContext(cloudEvent map[string]interface{}, traceID string) {
+// FromCloudEvent returns a map representation of an existing cloudevents JSON
+func FromCloudEvent(cloudEvent []byte, traceID string) (map[string]interface{}, error) {
+	var m map[string]interface{}
+	err := jsoniter.Unmarshal(cloudEvent, &m)
+	if err != nil {
+		return m, err
+	}
+
+	setTraceContext(m, traceID)
```
artursouza, line 76: Separate this. Keep this method as simple as deserializing.

artursouza (dismissed):

artursouza (approved):

PR #570, opened 2021-01-04
Extend custom cloud event

yaron2:
This PR adds additional fields to the custom Cloud Event creation method, needed by `daprd`.


wcs1only (approved):
LGTM

PR #571, opened 2021-01-05
fix data serialization in cloudevent

yaron2:
This PR fixes a regression in JSON serialization


artursouza (approved):

tcnghia:
> This PR fixes a regression in JSON serialization

Can we please add a test to detect future regressions?

yaron2:
> > This PR fixes a regression in JSON serialization
> 
> Can we please add a test to detect future regressions?

This was discovered in `daprd` tests

yaron2:
> > This PR fixes a regression in JSON serialization
> 
> Can we please add a test to detect future regressions?

Also, the functionality change + the change in the test will now account for that

PR #579, opened 2021-01-09
Treat etag errors across all state stores

yaron2:
This PR consolidates error handling for possible etag mismach and invalid etags across all state stores.

Partially closes https://github.com/dapr/dapr/issues/2619

### DIFF
--- a/state/azure/tablestorage/tablestorage.go
+++ b/state/azure/tablestorage/tablestorage.go
@@ -95,7 +95,12 @@ func (r *StateStore) Init(metadata state.Metadata) error {
 func (r *StateStore) Delete(req *state.DeleteRequest) error {
 	r.logger.Debugf("delete %s", req.Key)
 
-	return r.deleteRow(req)
+	err := r.deleteRow(req)
+	if req.ETag != "" {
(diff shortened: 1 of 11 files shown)


artursouza (approved):

state/azure/tablestorage/tablestorage.go
```diff
@@ -122,7 +127,12 @@ func (r *StateStore) Get(req *state.GetRequest) (*state.GetResponse, error) {
 func (r *StateStore) Set(req *state.SetRequest) error {
 	r.logger.Debugf("saving %s", req.Key)
 
-	return r.writeRow(req)
+	err := r.writeRow(req)
+	if req.ETag != "" {
+		return state.NewETagError(state.ETagMismatch, err)
+	}
```
rynowak, line 133: @yaron2 - can you explain how this works? I'm not familiar with those code but it looks like this would return `state.NewETagError` every time an ETag is specified.

yaron2: Theres an err check missing here, I'll push a fix later

PR #591, opened 2021-01-12
Fixed #590

### LINKED ISSUE #590: SecretStore BulkGetSecret method flattens all of secrets into a single map
## Expected Behavior

Should BulkGetSecret responses return a `map[string]map[string]string`?

## Actual Behavior

BulkGetSecret responses returns a `map[string]string` instead of  `map[string]map[string]string` which means that when using Kubernetes we are losing the original secret key / name cause the method is flattening the results.

Should we keep the method as is? Or should we change it to return `map[string]map[string]string`?

Check: https://github.com/dapr/dotnet-sdk/pull/511#discussion_r554204277 for more details.

## Release Note

RELEASE NOTE:

cmendible:
# Description

`BulkGetSecret` method now returns a `map[string]map[string]string`

## Issue reference

Please reference the issue this PR will close: #590

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#1089

### DIFF
--- a/secretstores/aws/secretmanager/secretmanager.go
+++ b/secretstores/aws/secretmanager/secretmanager.go
@@ -85,9 +85,9 @@ func (s *smSecretStore) GetSecret(req secretstores.GetSecretRequest) (secretstor
 }
 
 // BulkGetSecret retrieves all secrets in the store and returns a map of decrypted string/string values
-func (s *smSecretStore) BulkGetSecret(req secretstores.BulkGetSecretRequest) (secretstores.GetSecretResponse, error) {
-	resp := secretstores.GetSecretResponse{
-		Data: map[string]string{},
+func (s *smSecretStore) BulkGetSecret(req secretstores.BulkGetSecretRequest) (secretstores.BulkGetSecretResponse, error) {
+	resp := secretstores.BulkGetSecretResponse{
+		Data: map[string]map[string]string{},
 	}
 
 	search := true
@@ -99,21 +99,19 @@ func (s *smSecretStore) BulkGetSecret(req secretstores.BulkGetSecretRequest) (se
 			NextToken:  nextToken,
 		})
 		if err != nil {
-			return secretstores.GetSecretResponse{Data: nil}, fmt.Errorf("couldn't list secrets: %s", err)
+			return secretstores.BulkGetSecretResponse{Data: nil}, fmt.Errorf("couldn't list secrets: %s", err)
 		}
 
 		for _, entry := range output.SecretList {
 			secrets, err := s.client.GetSecretValue(&secretsmanager.GetSecretValueInput{
 				SecretId: entry.Name,
-				// VersionId:    versionID,
-				// VersionStage: versionStage,
 			})
 			if err != nil {
-				return secretstores.GetSecretResponse{Data: nil}, fmt.Errorf("couldn't get secret: %s", *entry.Name)
+				return secretstores.BulkGetSecretResponse{Data: nil}, fmt.Errorf("couldn't get secret: %s", *entry.Name)
 			}
 
 			if entry.Name != nil && secrets.SecretString != nil {
-				resp.Data[*entry.Name] = *secrets.SecretString
+				resp.Data[*entry.Name] = map[string]string{*entry.Name: *secrets.SecretString}
 			}
 		}
 
--- a/secretstores/azure/keyvault/keyvault.go
+++ b/secretstores/azure/keyvault/keyvault.go
@@ -84,21 +84,21 @@ func (k *keyvaultSecretStore) GetSecret(req secretstores.GetSecretRequest) (secr
 }
 
 // BulkGetSecret retrieves all secrets in the store and returns a map of decrypted string/string values
-func (k *keyvaultSecretStore) BulkGetSecret(req secretstores.BulkGetSecretRequest) (secretstores.GetSecretResponse, error) {
+func (k *keyvaultSecretStore) BulkGetSecret(req secretstores.BulkGetSecretRequest) (secretstores.BulkGetSecretResponse, error) {
 	vaultURI := k.getVaultURI()
 
 	maxResults, err := k.getMaxResultsFromMetadata(req.Metadata)
 	if err != nil {
-		return secretstores.GetSecretResponse{}, err
+		return secretstores.BulkGetSecretResponse{}, err
 	}
 
 	secretsResp, err := k.vaultClient.GetSecretsComplete(context.Background(), vaultURI, maxResults)
 	if err != nil {
-		return secretstores.GetSecretResponse{}, err
+		return secretstores.BulkGetSecretResponse{}, err
 	}
 
-	resp := secretstores.GetSecretResponse{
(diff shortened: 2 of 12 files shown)


yaron2:
hey @rynowak can you please review this? it's about the unflattening of Bulk Get for secrets.

secretstores/aws/secretmanager/secretmanager.go
```diff
@@ -109,11 +109,11 @@ func (s *smSecretStore) BulkGetSecret(req secretstores.BulkGetSecretRequest) (se
 				// VersionStage: versionStage,
 			})
 			if err != nil {
-				return secretstores.GetSecretResponse{Data: nil}, fmt.Errorf("couldn't get secret: %s", *entry.Name)
+				return secretstores.BulkGetSecretResponse{Data: nil}, fmt.Errorf("couldn't get secret: %s", *entry.Name)
```
rynowak, line 112: Not changed in this PR but what's the rationale for `VersionId` and `VersionStage` to be commented out on L107 & L108? This seems like the kind of thing that should really be documented with a comment rather than "spooky ghost code"

rynowak: 👻

secretstores/aws/secretmanager/secretmanager.go
```diff
@@ -109,11 +109,11 @@ func (s *smSecretStore) BulkGetSecret(req secretstores.BulkGetSecretRequest) (se
...
 			}
 
 			if entry.Name != nil && secrets.SecretString != nil {
-				resp.Data[*entry.Name] = *secrets.SecretString
+				resp.Data[*entry.Name] = map[string]string{*entry.Name: *secrets.SecretString}
```
rynowak, line 116: It would be great to see a unit test or integration test added for this functionality. The `GetSecret` function already has tests so it should be a small amount of work.

secretstores/azure/keyvault/keyvault.go
```diff
@@ -107,15 +107,15 @@ func (k *keyvaultSecretStore) BulkGetSecret(req secretstores.BulkGetSecretReques
...
+			return secretstores.BulkGetSecretResponse{}, err
 		}
 
 		secretValue := ""
 		if secretResp.Value != nil {
 			secretValue = *secretResp.Value
 		}
 
-		resp.Data[secretName] = secretValue
+		resp.Data[secretName] = map[string]string{secretName: secretValue}
```
rynowak, line 118: The logic for `secretName` is slightly different here than from the `GetSecret` codepath.. `GetSecret` uses the name that was passed in, rather than inspecting the `ID` that was returned from the server and normalizing it to a name.

This makes sense because in the `GetSecret` case the user provided the name and in the `BulkGetSecret` the user didn't. I wonder if there are subtle normalization differences that will catch people out (mainly casing).

I think regardless you've probably done the most reasonable thing here, just making an observation in case it's interesting to drill into more.

secretstores/hashicorp/vault/vault.go
```diff
@@ -217,12 +217,12 @@ func (v *vaultSecretStore) BulkGetSecret(req secretstores.BulkGetSecretRequest)
 	// Set X-Vault-Request header
 	httpReq.Header.Set(vaultHTTPRequestHeader, "true")
 	if err != nil {
-		return secretstores.GetSecretResponse{Data: nil}, fmt.Errorf("couldn't generate request: %s", err)
+		return secretstores.BulkGetSecretResponse{Data: nil}, fmt.Errorf("couldn't generate request: %s", err)
```
rynowak, line 220: It seems like there's a bug here that this error isn't handled immediately after the call to `NewRequestWithContext`, and could result in a nullref. L216 & L218 both interact with `httpReq` after a possible error.

secretstores/local/env/envstore.go
```diff
@@ -39,15 +39,15 @@ func (s *envSecretStore) GetSecret(req secretstores.GetSecretRequest) (secretsto
...
 // BulkGetSecret retrieves all secrets in the store and returns a map of decrypted string/string values
-func (s *envSecretStore) BulkGetSecret(req secretstores.BulkGetSecretRequest) (secretstores.GetSecretResponse, error) {
-	r := map[string]string{}
+func (s *envSecretStore) BulkGetSecret(req secretstores.BulkGetSecretRequest) (secretstores.BulkGetSecretResponse, error) {
+	r := map[string]map[string]string{}
 
 	for _, element := range os.Environ() {
 		envVariable := strings.Split(element, "=")
-		r[envVariable[0]] = envVariable[1]
+		r[envVariable[0]] = map[string]string{envVariable[0]: envVariable[1]}
```
rynowak, line 47: This seems like it has a bug for cases where an env-var value contains an `=`.  You could end up with a slice of more than 2 elements. This should use `SplitN`

rynowak: For an example, think about base64 encoded text that uses `=` as padding. A the trailing `=`s would be removed, and a strict parser would fail to decode it.

secretstores/local/env/envstore.go
```diff
@@ -39,15 +39,15 @@ func (s *envSecretStore) GetSecret(req secretstores.GetSecretRequest) (secretsto
...
 	}
 
-	return secretstores.GetSecretResponse{
+	return secretstores.BulkGetSecretResponse{
```
rynowak, line 50: Would be great to see a test for `BulkGetSecret` - already has a test for `GetSecret`

secretstores/local/file/filestore.go
```diff
@@ -84,9 +84,15 @@ func (j *localSecretStore) GetSecret(req secretstores.GetSecretRequest) (secrets
...
-		Data: j.secrets,
+func (j *localSecretStore) BulkGetSecret(req secretstores.BulkGetSecretRequest) (secretstores.BulkGetSecretResponse, error) {
+	r := map[string]map[string]string{}
+
+	for k, v := range j.secrets {
+		r[k] = map[string]string{k: v}
+	}
+
+	return secretstores.BulkGetSecretResponse{
+		Data: r,
```
rynowak, line 95: Would be great to see a test for BulkGetSecret - already has some existing tests

secretstores/aws/secretmanager/secretmanager.go
cmendible: I removed the comments, indeed it was👻 code!

secretstores/local/env/envstore.go
cmendible: Good one!

secretstores/azure/keyvault/keyvault.go
cmendible: Yep casing might be an issue here, but other than that I think it's a good solution.

rynowak (approved):
Thanks!

rynowak:
@yaron2 will need someone with write access to approve, I'm not an approver for this repo 😆

yaron2 (approved):
lgtm

Issue #506, opened 2020-10-22
Dapr pub/sub publish is stripping Cloud Event fields


lynn-orrell:
If an already created Cloud Event is sent to Dapr to be published via the pub/sub api, Dapr is recreating the Cloud Event and ignoring (failing to propagate) existing Cloud Event properties.

https://github.com/dapr/components-contrib/blob/5e9140f1dbba55581b6fe5d0fe0a0530af30411c/pubsub/envelope.go#L72-L92

abserari:
What CloudEvent content is not good?

abserari:
https://github.com/dapr/components-contrib/pull/509 I try one here. Could give me some advice to promote this change. I think code change is enough and should represent it in the doc.

mchmarny:
Note, the CE extension names are unpredictable (`ce-*`) so this will have to include some form of parsing into map[string]interface{} and then adding one by one.

PR #602, opened 2021-01-14
Refactor and fix contenttype logic for pubsub.

artursouza:
# Description

Refactor and fix contenttype logic for pubsub.

## Issue reference

Please reference the issue this PR will close: #506 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/contenttype/utils.go
+++ b/contenttype/utils.go
@@ -1,0 +1,39 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+package contenttype
+
+import "strings"
+
+const (
+	// CloudEventContentType is the content type for cloud event.
+	CloudEventContentType = "application/cloudevents+json"
+	// JSONContentType is the content type for JSON.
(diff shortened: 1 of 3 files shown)


contenttype/utils.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func isContentType(contentType string, expected string) bool {
+	lowerContentType := strings.ToLower(contentType)
+	if lowerContentType == expected {
+		return true
+	}
+
+	semiColonPos := strings.Index(lowerContentType, ";")
+	if semiColonPos >= 0 {
```
yaron2, line 34: Is there a reason why this can't be changed to:

```go
return strings.Contains(lowerContentType, expected)
```

artursouza: To avoid false positives where one content type is a substring of another but not the same.

Example: `application/epub` is a substring of `application/epub+zip` but not the same.

yaron2: I would argue that in this case they are the same as far as content type matching is concerned, but I don't mind to keep this as is.

yaron2 (approved):
lgtm

Issue #596, opened 2021-01-13
Conformance tests - Setup and implement conformance test for Azure Service Bus PubSub


mukundansundar:
## Describe the proposal
Conformance tests - Setup and implement conformance test for Azure Service Bus
#404

PR #614, opened 2021-01-19
Azure service bus conf tests

mukundansundar:
# Description

## Issue reference

Please reference the issue this PR will close: #596 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -32,8 +32,8 @@ jobs:
         component:
         - output-binding.azure.blobstorage
         - output-binding.azure.storagequeues
-        - pubsub.redis
         - pubsub.azure.servicebus
+        - pubsub.redis
         - secretstores.localenv
         - secretstores.localfile
         - state.cosmosdb
@@ -111,5 +111,5 @@ jobs:
 
         echo "Running tests for Test${KIND_UPPER}Conformance/${KIND}/${NAME} ... "
 
-        go test -v -tags=conftests -count=1 ./tests/... \
+        go test -v -tags=conftests -count=1 ./tests/conformance \
           --run="Test${KIND_UPPER}Conformance/${KIND}/${NAME}"
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -31,25 +31,29 @@ jobs:
         # List here all the components that needs to be invoked in this workflow.
         component:
         - pubsub.redis
+        - pubsub.azure.servicebus
         - secretstores.localenv
+        - secretstores.localfile
         - state.cosmosdb
         - state.redis
         include:
+        # Unfortunately, Azure secrets can't have underscores in
+        # names, while environment variables with hyphens ('-') are
+        # troublesome.
+        #
+        # We work around here by leveraging the fact that
+        # environment variable names are case sensitive, so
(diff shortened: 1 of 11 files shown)


tests/config/state/tests.yml
```diff
@@ -3,5 +3,5 @@ components:
   - component: redis
     allOperations: true
     config:
-      maxInitDuration: 30
+      maxInitDuration: 100
```
mukundansundar, line 6: Because sometimes in local run it was taking some extra time.

artursouza (approved):

tests/conformance/common.go
```diff
@@ -74,10 +81,27 @@ func LoadComponents(componentPath string) ([]v1alpha1.Component, error) {
...
+
 func ConvertMetadataToProperties(items []v1alpha1.MetadataItem) map[string]string {
 	properties := map[string]string{}
 	for _, c := range items {
-		properties[c.Name] = c.Value.String()
+		val := c.Value.String()
+		if strings.HasPrefix(c.Value.String(), "${{") {
+			// look up env var with that name. remove ${{}} and space
+			k := strings.TrimSpace(strings.TrimSuffix(strings.TrimPrefix(val, "${{"), "}}"))
+			v := LookUpEnv(k)
```
mukundansundar, line 99: Fail test/ print error if env var value is empty.

tcnghia: Please don't fail parsing here using `log.Fatal`, since it will cause the other tests to fail as well.  Instead, can you please print a warning, or fail using `t.Error`?

mukundansundar: I cannot do a `t.Error` here, because if I did that it will not run the other tests in the configuration file. 
Instead I can continue to run other tests and pass an error back to the caller of `Run()` method which can then fail on behalf of this.

tcnghia (dismissed):

.github/workflows/conformance.yml
```diff
@@ -108,5 +111,5 @@ jobs:
 
         echo "Running tests for Test${KIND_UPPER}Conformance/${KIND}/${NAME} ... "
 
-        go test -v -tags=conftests -count=1 ./tests/... \
+        go test -v -tags=conftests -count=1 ./tests/conformance \
```
mukundansundar, line 114: this will only run the conformance tests and not the unit tests within `tests` directory.

tests/conformance/pubsub/pubsub.go
```diff
@@ -123,4 +120,10 @@ func ConformanceTests(t *testing.T, props map[string]string, pubusub pubsub.PubS
 			assert.LessOrEqual(t, config.messageCount, actualReadCount, "expected to read %v messages", config.messageCount)
 		})
 	}
+
+	t.Run(config.GetTestName("close pubsub"), func(t *testing.T) {
+		// Properly close connection to pubsub
+		// This is needed inside t.Run because of the way the individual tests are run based on name filtering in the conformance test workflow.
+		ps.Close()
```
mukundansundar, line 127: this will be removed in a later refactor.

artursouza (dismissed):

artursouza:
Please, resolve conflict.

artursouza (approved):

PR #615, opened 2021-01-20
mysql outbinding support

beiwei30:
# Description

A simple output binding for MySQL. Most of code are similar with Postgres's impl.

## Issue reference

Please reference the issue this PR will close: #616

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/pull/1102


yaron2:
Please open a corresponding docs PR to update the list of output bindings and add the schema

yaron2:
Thanks @beiwei30! have you tested this with a MySQL DB?

bindings/mysql/mysql.go
```diff
@@ -0,0 +1,281 @@
...
+		_ = rows.Close()
+		_ = rows.Err()
+	}()
+
+	columns, err := rows.Columns()
+	if err != nil {
+		return nil, errors.Wrapf(err, "error finding column names for %s", s)
+	}
+
+	var values [][]sql.RawBytes
```
pkedy, line 192: Hi @beiwei30!  How did you come up with using `sql.RawBytes` for this?  Its very concise but this documentation for it makes me wonder if it is safe to use like this.

```
// RawBytes is a byte slice that holds a reference to memory owned by
// the database itself. After a Scan into a RawBytes, the slice is only
// valid until the next call to Next, Scan, or Close.
```

There were other approaches that I found here. Could any of these work?

https://stackoverflow.com/questions/42774467/how-to-convert-sql-rows-to-typed-json-in-golang

beiwei30:
> Thanks @beiwei30! have you tested this with a MySQL DB?

Yes, it's tested.

bindings/mysql/mysql.go
beiwei30: I simply follow the sample code of 'RawBytes' from mysql's doc: https://github.com/go-sql-driver/mysql/wiki/Examples#rawbytes

In fact, RawBytes is simply a slice of byte:

```go
type RawBytes []byte
``` 

In my impl, I create sql.RawBytes for each iteration instead of reuse it so that the behavior described in the comments quoted above is not a concern.

beiwei30:
> Please open a corresponding docs PR to update the list of output bindings and add the schema

working on it.

beiwei30:
t

> > Please open a corresponding docs PR to update the list of output bindings and add the schema
> 
> working on it.

done, pls. check https://github.com/dapr/docs/pull/1102

bindings/mysql/mysql.go
pkedy: Thanks for the example link and agreed on the safety by creating a new row of sql.RawBytes for each iteration.

I'm also curious how numeric types are marshaled in the JSON...  does 1234 serialize to `1234` or `"1234"` (with quotes)?

beiwei30: I will try json marshal tomorrow and get back to you later.

beiwei30: @pkedy would you mind to take a look into https://github.com/dapr/components-contrib/pull/615/commits/dab941ece703d31c489c7b27c76b9878e8a22487#diff-6b618681e6b9a3154c568fe120e86733654e2f369edfea503df583ac5c3e8ef1R197

beiwei30: > Thanks for the example link and agreed on the safety by creating a new row of sql.RawBytes for each iteration.
> 
> I'm also curious how numeric types are marshaled in the JSON... does 1234 serialize to `1234` or `"1234"` (with quotes)?

pls. check https://github.com/dapr/components-contrib/pull/615/files#diff-dc91a6eb8044921dc5f73b0a7dd0e22fd2b03e619862931032b95fb2e9965cdfR48. For numeric types, it's marshalled into a number instead of a string.

pkedy: @beiwei30 I see that test :+1: and I ❤️ the changes you've made since I last looked at this. You might also consider adding an `assert.Contains` for the `TIME` value you have (for example does that serialize as a RFC 3339 string or integer in milliseconds) and maybe adding a `boolean` type to make sure that comes back as `true` or `false` and not `1` or `0`. 

What do you think about adding/calling your type tests from the integration tests also?  My thought is that `go-sqlmock` could return stuff differently than the MySQL client. Just a thought.

beiwei30: @pkedy pls. check  61c286d08a38df409479789415bea38781eeba87 and b6ce23db57e3df1d7eac131fac4caef36baef79c

beiwei30: with regarding to `boolean` type, in fact, mysql doesn't support `BOOL` at all.  I add a test to verify it is `TINYINT(1)` anyway. Pls. check this 62572d9e45a8de96b88ca7111f43f9eeab512736

yaron2 (approved):
lgtm

Issue #617, opened 2021-01-20
Conformance tests - Setup input bindings API tests 


mukundansundar:
## Describe the proposal
Conformance tests - Setup input bindings API tests 
Potential component Kafka
#404

mukundansundar:
@yaron2 please add this to 1.0.0 milestone project

mukundansundar:
closed with https://github.com/dapr/components-contrib/pull/637

PR #629, opened 2021-01-22
Add Kafka output binding and refactor bindings

halspang:
# Description
This commit adds Kafka output bindings tests and refactors the
general bindings tests to include input and output bindings in
the same set of tests.

## Note
Due to inconsistencies with Kafka, this just sets up the input bindings via a refactor. Kafka is currently only doing output binding tests. The input binding tests will be further flushed with another component.

## New Note
Now it includes an input binding as AzureStorageQueues worked very easily meaning that the test instability with Kafka was likely just an issue with Kafka/the binding.

## Issue reference
Partial: #617 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


tests/conformance/bindings/bindings.go
```diff
@@ -36,44 +39,67 @@ func (tc *TestConfig) createInvokeRequest() bindings.InvokeRequest {
...
 			for _, op := range ops {
 				assert.True(t, config.HasOperation(string(op)), fmt.Sprintf("Operation missing from conformance test config: %v", op))
 			}
 		})
 	}
 
-	// CreateOperations
+	inputBindingCall := 0
+	var wg sync.WaitGroup
+	if config.CommonConfig.HasOperation("read") {
```
mukundansundar, line 76:
```suggestion
	if config.HasOperation("read") {
```

can remove CommonConfig

tests/conformance/bindings/bindings.go
```diff
@@ -83,29 +109,59 @@ func ConformanceTests(t *testing.T, props map[string]string, binding bindings.Ou
...
+			go func() {
+				defer close(c)
+				wg.Wait()
+			}()
+
+			// To stop the test from hanging if there's no response, we can setup a simple timeout.
+			select {
+			case <-c:
+				assert.Greater(t, inputBindingCall, 0)
+			case <-time.After(60 * time.Second):
```
mukundansundar, line 153: make the timeout configurable.

mukundansundar (commented):
why is `wg.Add(1)` called for each operation? shouldn't it only be called for `create` operation?

tests/conformance/bindings/bindings.go
```diff
@@ -36,44 +39,67 @@ func (tc *TestConfig) createInvokeRequest() bindings.InvokeRequest {
 	}
 }
 
-func ConformanceTests(t *testing.T, props map[string]string, binding bindings.OutputBinding, config TestConfig) {
+func ConformanceTests(t *testing.T, props map[string]string, inputBinding bindings.InputBinding, outputBinding bindings.OutputBinding, config TestConfig) {
+	// We need to increase the number of processes as the input binding is its own process that kicks off multiple sub processes
+	runtime.GOMAXPROCS(20)
```
mukundansundar, line 44: @yaron2 Can you give your thoughts on this? Why is it that GOMAXPROCS needs an increase?

tests/conformance/bindings/bindings.go
halspang: Done.

tests/conformance/bindings/bindings.go
halspang: Done.

halspang:
> why is `wg.Add(1)` called for each operation? shouldn't it only be called for `create` operation?

Do all output binding actions not trigger the input binding?

yaron2:
> > why is `wg.Add(1)` called for each operation? shouldn't it only be called for `create` operation?
> 
> Do all output binding actions not trigger the input binding?

They are decoupled and have no direct relation to one another.

tests/conformance/bindings/bindings.go
yaron2: I'm not sure we need a change: `GOMAXPROCS` by default sets it's value to the number of CPU cores on a machine. It used to be `1` in older versions of Go.

halspang: Is there an alternate to adjusting it? The behavior I was seeing was (with debug statements):

Test starts
goroutine for consumption is kicked off
Test waits
Test completes
goroutine executes

It had the same behavior every time and increasing the max procs allowed the routine to run in parallel with the test.

yaron2: were you seeing this behavior on your local machine?

halspang: Yes, I haven't tried with the provisioned workflow.

yaron2: the test and the goroutine are running on the same process?

halspang: This is where I may not be well versed enough in how go does things under the hood. My gut instinct would say that a goroutine would just be a managed thread vs. go creating a new process. 

That being said, I can at least say that the test is what is kicking off these goroutines.

yaron2: Go will not be creating a new process, but a goroutine is "sort of" a managed thread. Goroutines are scheduled on top of OS system threads, which are set with GOMAXPROCS and default to the number of CPUs your box has. I find it very weird that the goroutine executed only after the test completes.

halspang: I agree, which is what lead me down the 'how many things does go run' path and we ended up here. Is this something that we really shouldn't be doing? Or is it just weird? If there's a better way I'd love to learn it cause I'm also not a huge fan of setting runtime like that in a test.

halspang: Circling back to this, I just got AzureStorageQueues to work. Unlike the Kafka binding, they only spin up a single goroutine. This did work without the max procs setting (though I left it in). It does seem to back that it was a goroutine limit though which is why I left it in in case other bindings had issues. Thoughts @yaron2? Given the currently implemented binding works without it I'm fine removing it, but I worry we'll just pass the buck down.

tests/config/bindings/tests.yml
```diff
@@ -1,13 +1,18 @@
...
     operations: ["create", "operations", "get"]
     config:
       blobName: $((uuid))
+      readBindingTimeout: 60
   - component: azure.storagequeues
+    operations: ["create", "operations", "read"]
+  - component: kafka
     operations: ["create", "operations"]
+    config:
+      readBindingTimeout: 60
```
mukundansundar, line 18: can you distinguish between `input_` metadata and `output_` metadata? This way all the `output_` metadata alone can be passed along to the InvokeRequest?

halspang: Done.

tests/conformance/bindings/bindings.go
```diff
@@ -0,0 +1,191 @@
...
+				req.Operation = bindings.GetOperation
+				resp, err := outputBinding.Invoke(&req)
+				assert.Nil(t, err, "expected no error invoking output binding")
+				assert.NotNil(t, resp)
+				assert.Nil(t, resp.Data)
+			}
+		})
+	}
+
+	if config.CommonConfig.HasOperation("read") {
```
mukundansundar, line 171: probably have `verify read` before `delete`?

tests/config/bindings/tests.yml
```diff
@@ -1,13 +1,18 @@
...
     config:
-      blobName: $((uuid))
+      output_blobName: $((uuid))
+      readBindingTimeout: 60
   - component: azure.storagequeues
+    operations: ["create", "operations", "read"]
+  - component: kafka
     operations: ["create", "operations"]
+    config:
+      readBindingTimeout: 60
```
mukundansundar, line 18: can remove this since kafka does not have read operation yet.

mukundansundar (commented):
Please fix the conformance.yml file. Also please add setup of local Kafka cluster probably using docker compose in the file itself. Similar to how a local redis or mongodb is setup.

tests/conformance/bindings/bindings.go
```diff
@@ -0,0 +1,191 @@
...
+		outputMetadata:     make(map[string]string),
+		readBindingTimeout: defaultTimeoutDuration,
+	}
+
+	if val, ok := config["readBindingTimeout"]; ok {
+		timeout, err := strconv.Atoi(val)
+		if err == nil {
+			testConfig.readBindingTimeout = time.Duration(timeout) * time.Second
+		}
+	}
```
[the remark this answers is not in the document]
halspang: Done.

tests/conformance/bindings/bindings.go
halspang: Makes sense, done.

tests/config/bindings/tests.yml
halspang: Done.

tests/config/bindings/tests.yml
```diff
@@ -1,13 +1,16 @@
...
+      output_key: $((uuid))
+      readBindingTimeout: 60
   - component: azure.blobstorage
     operations: ["create", "operations", "get"]
     config:
-      blobName: $((uuid))
+      output_blobName: $((uuid))
+      readBindingTimeout: 60
   - component: azure.storagequeues
+    operations: ["create", "operations", "read"]
```
mukundansundar, line 14: is the default readBindingTimeout sufficient for this?

tests/config/bindings/tests.yml
```diff
@@ -1,13 +1,16 @@
...
     config:
-      key: $((uuid))
+      output_key: $((uuid))
+      readBindingTimeout: 60
   - component: azure.blobstorage
     operations: ["create", "operations", "get"]
     config:
-      blobName: $((uuid))
+      output_blobName: $((uuid))
+      readBindingTimeout: 60
```
mukundansundar, line 12: can remove.

tests/config/bindings/tests.yml
```diff
@@ -1,13 +1,16 @@
-# the config map is passed in a metadata map to the Invoke request for output bindings
-componentType: output-binding
+componentType: bindings
 components:
   - component: redis
     operations: ["create", "operations"]
     config:
-      key: $((uuid))
+      output_key: $((uuid))
+      readBindingTimeout: 60
```
mukundansundar, line 7: can remove.

halspang: Done.

tests/config/bindings/tests.yml
halspang: Done.

tests/config/bindings/tests.yml
halspang: So far it's been sufficient, we could probably set it lower if we wanted to as it has always been pretty much instant for me.

halspang: It has been very quick for me, so we could even make it lower if we wanted too.

mukundansundar (approved):

halspang:
This commit was merged as part of https://github.com/dapr/components-contrib/pull/637. I'm leaving this PR open for a bit longer in case anyone wants to add more comments about this code which we can add as a new commit here. If there are no comments, I'll close it.

PR #631, opened 2021-01-23
Added binary data type detection

jigargandhi:
Added base64 encoding for binary data types

# Description

Added binary data type detection based on mime type.
Base 64 encoded binary data.

## Issue reference
This will not fix any issue but help in https://github.com/dapr/dapr/issues/2612

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/contenttype/utils.go
+++ b/contenttype/utils.go
@@ -26,7 +26,7 @@ func IsJSONContentType(contentType string) bool {
 
 // IsStringContentType determines if content type is string
 func IsStringContentType(contentType string) bool {
-	if strings.HasPrefix(contentType, "text/") {
+	if strings.HasPrefix(strings.ToLower(contentType), "text/") {
 		return true
 	}
 
--- a/contenttype/utils.go
+++ b/contenttype/utils.go
@@ -24,6 +24,15 @@ func IsJSONContentType(contentType string) bool {
 	return isContentType(contentType, JSONContentType)
 }
 
+// IsStringContentType determines if content type is string
+func IsStringContentType(contentType string) bool {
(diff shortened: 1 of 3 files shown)


contenttype/utils.go
```diff
@@ -24,6 +24,15 @@ func IsJSONContentType(contentType string) bool {
 	return isContentType(contentType, JSONContentType)
 }
 
+// IsStringContentType determines if content type is string
+func IsStringContentType(contentType string) bool {
```
artursouza, line 28: This is OK for now. I wonder if we could eventually have a richer MimeType comparison logic based on hierarchy. Similar to this project: https://github.com/gabriel-vasile/mimetype

artursouza: I created an issue to track this discussion: https://github.com/dapr/components-contrib/issues/635

contenttype/utils.go
```diff
@@ -24,6 +24,15 @@ func IsJSONContentType(contentType string) bool {
...
+	if strings.HasPrefix(contentType, "text/") {
```
artursouza, line 29: isContentType does ToLower(). Please, use ToLower() here too.

artursouza (changes requested):

artursouza:
Thanks for the contribution. I am OK with this PR if it can be case insensitive.

contenttype/utils.go
artursouza: I will add ToLower() here directly given the pressure on time for the release.

artursouza (approved):

PR #632, opened 2021-01-25
Fix get bulk secret for hashicorp vault.

artursouza:
# Description

Fix get bulk secret for hashicorp vault.

## Issue reference

Please reference the issue this PR will close: None

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/secretstores/hashicorp/vault/vault.go
+++ b/secretstores/hashicorp/vault/vault.go
@@ -209,7 +209,7 @@ func (v *vaultSecretStore) BulkGetSecret(req secretstores.BulkGetSecretRequest) 
 	}
 
 	// Create list secrets url
-	vaultSecretsPathAddr := fmt.Sprintf("%s/v1/secret/data/%s", v.vaultAddress, v.vaultKVPrefix)
+	vaultSecretsPathAddr := fmt.Sprintf("%s/v1/secret/metadata/%s", v.vaultAddress, v.vaultKVPrefix)
 
 	httpReq, err := http.NewRequestWithContext(context.Background(), "LIST", vaultSecretsPathAddr, nil)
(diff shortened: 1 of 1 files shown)


artursouza:
I was surprised when I saw that:

1. The list call to KV was not even working.
2. It was not including the secret's key in the dictionary and building it in a funny way.

Let's all take a moment to thank Java SDK's ITs for saving our a** one more time.

mukundansundar (approved):
lgtm

yaron2 (approved):

secretstores/hashicorp/vault/vault.go
```diff
@@ -209,7 +209,7 @@ func (v *vaultSecretStore) BulkGetSecret(req secretstores.BulkGetSecretRequest)
 	}
 
 	// Create list secrets url
-	vaultSecretsPathAddr := fmt.Sprintf("%s/v1/secret/data/%s", v.vaultAddress, v.vaultKVPrefix)
+	vaultSecretsPathAddr := fmt.Sprintf("%s/v1/secret/metadata/%s", v.vaultAddress, v.vaultKVPrefix)
```
yaron2, line 212: why did this change? there's another occurance for this in `getSecret`

artursouza: Mukundan made the same question. This URL here is used for the LIST API, which (as per hashicorp documentation) is a metadata operation.

https://www.vaultproject.io/api/secret/kv/kv-v2#list-secrets

In the end, I actually ran this as part of my ITs in Java SDK the old way does not work at all.

yaron2 (approved):
LGTM

PR #650, opened 2021-02-01
Added documentation on versioning Dapr components

### LINKED ISSUE #526: Document when to release a new version of a component
Put developer documentation in place that describes when to create a new package and version of a given component.
The guideline here is that every `metadata` field change, addition or removal constitutes a major version bump.

pkedy:
This PR adds documentation for when and how to create Version 2 of a component and beyond.

Resolves #526

PR for adding component versioning support is dapr/dapr#2757

### DIFF
--- a/docs/developing-component.md
+++ b/docs/developing-component.md
@@ -98,7 +98,7 @@ API versioning of Dapr components follows the same approach as [Go modules](http
 * Renaming or removing a `metadata` field that the component is currently using
 * Adding a required `metadata` field
 * Adding an optional field that does not have a backward compatible default value
-* Changing the component's logic such that it deviates from previously expected behavior
+* Making significant changes to the component's behavior that would adversely affect existing users
 
 In most cases, breaking changes can be avoided by using backward compatible `metadata` fields. When breaking changes cannot be avoided, here are the steps for creating the next major version of a component:
 
--- a/docs/developing-component.md
+++ b/docs/developing-component.md
@@ -53,7 +53,7 @@ make lint
 
 ## Validating with Dapr core
 
-1. Make sure you clone dapr and component-contrib repos under $GOPATH/src/github.com/dapr
+1. Make sure you clone Dapr and component-contrib repos under $GOPATH/src/github.com/dapr
 2. Replace github.com/dapr/components-contrib reference to the local component-contrib
 ```bash
 go mod edit -replace github.com/dapr/components-contrib=../components-contrib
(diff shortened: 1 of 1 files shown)


docs/developing-component.md
```diff
@@ -87,6 +87,23 @@ cp ./dist/darwin_amd64/debug/daprd ~/.dapr/bin
...
+5. Import your component to Dapr [main.go](https://github.com/dapr/dapr/blob/b3e1fe848de3ea7b297c712d188136919d314887/cmd/daprd/main.go#L20-L115)
+6. Register your component in Dapr [main.go](https://github.com/dapr/dapr/blob/b3e1fe848de3ea7b297c712d188136919d314887/cmd/daprd/main.go#L256-L385)
 7. Create a pullrequest in [Dapr](https://github.com/dapr/dapr/pulls)
+
+## Version 2 and beyond of a component
+
+API versioning of Dapr components follows the same approach as [Go modules](https://blog.golang.org/v2-go-modules) where the unstable version (v0) and first stable version (v1) are contained in the root directory of the component package.  For example v1 of the Redis binding component is located in `bindings/redis`. Code changes may continue in this package provided there are no breaking changes. Breaking changes include:
+
+* Renaming or removing a `metadata` field that the component is currently using
+* Adding a required `metadata` field or optional field that does not have a backward-compatible default value
```
artursouza, line 99: Split into two lines, it reads better.

docs/developing-component.md
```diff
@@ -87,6 +87,23 @@ cp ./dist/darwin_amd64/debug/daprd ~/.dapr/bin
...
+* Changing the component's logic such that it deviates from previously expected behavior
```
artursouza, line 100: This is very subjective. Can it be clearer? This can be interpreted as "no code change allowed".

artursouza (commented):
Small changes.

docs/developing-component.md
pkedy: How about something like

```
Making significant changes to the component's behavior that would adversely affect existing users
```

Then it seems like the subjectivity could come from the word "adversely". But is it more clear? Open to suggestions.

artursouza: There will always be subjectivity. It is better now because not every change will cause that. People can still abuse this statement but maintainers exist for a reason.

artursouza (approved):

pkedy:
Is this ready for approval @artursouza @yaron2 ?

artursouza (approved):

Issue #619, opened 2021-01-20
Conformance tests - Setup Azure Service Bus Queue bindings conf tests 


mukundansundar:
## Describe the proposal
Conformance tests - Setup Azure Service Bus Queue bindings conf tests 
#404

mukundansundar:
The queue has been setup ... once https://github.com/dapr/components-contrib/pull/608 and input bindings PRs are merged can complete this setup. 
@yaron2  Please add this to v1.0.0 project.

PR #651, opened 2021-02-01
Add Azure Service Bus Queues binding to tests

halspang:
# Description

Add Azure Service Bus Queues binding to tests

## Issue reference

Closes: #619 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/tests/config/bindings/azure/servicebusqueues/bindings.yml
+++ b/tests/config/bindings/azure/servicebusqueues/bindings.yml
@@ -1,0 +1,15 @@
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: servicebusqueue-binding
+  namespace: default
+spec:
+  type: bindings.azure.servicebusqueues
+  version: v1
+  metadata:
+  - name: connectionString
+    value: ${{AzureServiceBusConnectionString}}
+  - name: queueName
+    value: conf-tests
+  - name: ttlInSeconds
+    value: 60
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -33,6 +33,7 @@ jobs:
         # List here all the components that needs to be invoked in this workflow.
         component:
         - bindings.azure.blobstorage
+        - bindings.azure.servicebusqueues
         - bindings.azure.storagequeues
         - bindings.http
         - bindings.kafka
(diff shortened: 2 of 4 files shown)


tests/config/bindings/azure/servicebusqueues/bindings.yml
```diff
@@ -0,0 +1,15 @@
...
+  name: servicebusqueue-binding
+  namespace: default
+spec:
+  type: bindings.azure.servicebusqueues
+  version: v1
+  metadata:
+  - name: connectionString
+    value: ${{AzureServiceBusConnectionString}}
+  - name: queueName
+    value: ${{AzureServiceBusQueue}}
```
tcnghia, line 13: This secret needs to be part of required_secrets in conformance.yaml (line 69)

tests/config/bindings/azure/storagequeues/bindings.yml
```diff
@@ -12,6 +12,6 @@ spec:
   - name: storageAccessKey
     value: ${{AzureBlobStorageAccessKey}}
   - name: queue
-    value: ${{AzureBlobStorageQueue}}
+    value: conf-tests
```
tcnghia, line 15: Is this an intentional change? I think it's technically isn't a secret so probably fine hard-coded, but just making sure that this isn't a leftover line to running tests locally.

tests/config/bindings/azure/servicebusqueues/bindings.yml
halspang: The first one is, the 2nd one I meant to remove but accidentally removed the wrong config... I've made it hard coded as it doesn't feel like it needs to be a secret to me. Will change if you think it should be.

tests/config/bindings/azure/storagequeues/bindings.yml
halspang: Yeah that was a miss-click. service queue vs storage queue, changed it back.

tests/config/bindings/azure/servicebusqueues/bindings.yml
tcnghia: I think it should be fine hard-coded here. If we want to we can extract this out as an environment variable in the workflow later.

tcnghia (approved):

mukundansundar (approved):

artursouza (approved):

Issue #647, opened 2021-01-31
Allow DynamoDB state store to resolve credentials using AWS SDK credentials chain


cuongvo:
## Remove the required credential fields from the DDB state store definition

Currently in the DynamoDB state store we [require](https://github.com/dapr/components-contrib/blob/master/state/aws/dynamodb/dynamodb.go#L207-L209) metadata defining an access key, secret key and session token. We should consider making these fields optional to allow the AWS Go SDK to [fallback to its credentials provider chain](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html#specifying-credentials) to retrieve credentials. 

The credentials used are AWS temporary credentials, which expire and need to be rotated. In production, to avoid this complexity and ensure better security practices, users of AWS resources use tools like [kube2iam](https://github.com/jtblin/kube2iam) or if they're using AWS's managed Kubernetes offering EKS, the built in [feature](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-technical-overview.html).

AWS SDK's have default behavior that know how to retrieve these credentials as provided by the mentioned tools. When we provided the credentials directly (as currently we're required to do), this behavior is disabled.

Dapr details these points in the [Authenticating to AWS](https://docs.dapr.io/developing-applications/integrations/authenticating/authenticating-aws/) docs.

trondhindenes:
oh wow, that looks absolutely right. We need to adjust the dynamodb stuff so that it's "in line" with the shared aws auth code we added a while back. I can attempt to get this fixed, should be an easy fix.

yaron2:
Yep. @trondhindenes did an amazing job a while back consolidating aws authentication for components and enabling credentials support from the environment, and so it really should be an easy fix.

yaron2:
@cuongvo this is now merged and will come out with the 1.0 release.

Thanks @trondhindenes for picking this up.

PR #652, opened 2021-02-02
State:Dynamodb - Dont require access key/secret

trondhindenes:
# Description
Dont fail if aws access key/secret vars are empty. This is part of the "implicit aws auth flow" which we changed a way back. Dynamodb had some hard-coded checks that this PR removes.

Looks like we simply don't have any documentation for the dymamodb state store yet (https://v1-rc2.docs.dapr.io/operations/components/setup-state-store/supported-state-stores/) so I can try and get that in after this is merged (marking docs as "done" in the list below, since there's no doc updates blocking this PR)

See also https://github.com/dapr/dapr/issues/2469

## Issue reference
Please reference the issue this PR will close: #647

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/state/aws/dynamodb/dynamodb.go
+++ b/state/aws/dynamodb/dynamodb.go
@@ -203,7 +203,6 @@ func (d *StateStore) getDynamoDBMetadata(metadata state.Metadata) (*dynamoDBMeta
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


yaron2 (approved):
LGTM

Issue #604, opened 2021-01-15
Conformance Tests - Setup and implement Azure Key vault for tests


mukundansundar:
## Describe the proposal
Conformance Tests - Setup and implement Azure Key vault for tests
#404

mukundansundar:
Setup Azure Key vault in the resource group, add the conf tests secrets to it and add that to the conf test suite for secret stores.

PR #654, opened 2021-02-02
Add conformance tests for Azure Keyvault Secret Store

tcnghia:
# Description

Add conformance tests for Azure Keyvault Secret Store.

## Issue reference

Please reference the issue this PR will close: #604 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -41,6 +41,7 @@ jobs:
         - bindings.redis
         - pubsub.azure.servicebus
         - pubsub.redis
+        - secretstores.azure.keyvault
         - secretstores.localenv
         - secretstores.localfile
         - state.cosmosdb
@@ -72,6 +73,9 @@ jobs:
           required-secrets: AzureServiceBusConnectionString
         - component: bindings.azure.storagequeues
           required-secrets: AzureBlobStorageAccessKey,AzureBlobStorageAccount,AzureBlobStorageQueue
+        - component: secretstores.azure.keyvault
+          required-secrets: AzureKeyVaultSecretStoreTenantId,AzureKeyVaultSecretStoreClientId
+          required-certs: AzureKeyVaultSecretStoreCert
     steps:
     - name: Check out code onto GOPATH
       uses: actions/checkout@v2
@@ -92,6 +96,7 @@ jobs:
       if: matrix.required-secrets != ''
 
     - name: Start ngrok
+      if: contains(matrix.component, 'azure.eventgrid')
       run: |
         wget https://bin.equinox.io/c/4VmDzA7iaHb/ngrok-stable-linux-amd64.zip
         unzip -qq ngrok-stable-linux-amd64.zip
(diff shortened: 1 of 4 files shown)


tcnghia:
A successful run is here https://github.com/tcnghia/components-contrib/actions/runs/532275420

tcnghia:
BulkGet currently isn't enable due to https://github.com/dapr/components-contrib/issues/655

halspang (commented):
Looks good to me. I noticed that the component mentions a password for the cert. Is there value in creating a password protected cert and ensuring that works as well?

https://github.com/dapr/components-contrib/blob/master/secretstores/azure/keyvault/keyvault.go#L23

tcnghia:
> Looks good to me. I noticed that the component mentions a password for the cert. Is there value in creating a password protected cert and ensuring that works as well?
> 
> https://github.com/dapr/components-contrib/blob/master/secretstores/azure/keyvault/keyvault.go#L23

Since our cert is downloaded from KeyVault and cleaned up right away after done, I think we don't need to use a password like one that gets stored for a while (like on a dev laptop). If we add a password, it will need to be stored in the secret store anyway, which doesn't offer us more protection.

halspang:
> > Looks good to me. I noticed that the component mentions a password for the cert. Is there value in creating a password protected cert and ensuring that works as well?
> > https://github.com/dapr/components-contrib/blob/master/secretstores/azure/keyvault/keyvault.go#L23
> 
> Since our cert is downloaded from KeyVault and cleaned up right away after done, I think we don't need to use a password like one that gets stored for a while (like on a dev laptop). If we add a password, it will need to be stored in the secret store anyway, which doesn't offer us more protection.

I'm not worried about the protection so much as if it works or not as a path for the component. Then again, we're getting dangerously close to testing the Azure sdk instead of the component, so I can go either way. Thoughts?

tcnghia:
> > > Looks good to me. I noticed that the component mentions a password for the cert. Is there value in creating a password protected cert and ensuring that works as well?
> > > https://github.com/dapr/components-contrib/blob/master/secretstores/azure/keyvault/keyvault.go#L23
> > 
> > 
> > Since our cert is downloaded from KeyVault and cleaned up right away after done, I think we don't need to use a password like one that gets stored for a while (like on a dev laptop). If we add a password, it will need to be stored in the secret store anyway, which doesn't offer us more protection.
> 
> I'm not worried about the protection so much as if it works or not as a path for the component. Then again, we're getting dangerously close to testing the Azure sdk instead of the component, so I can go either way. Thoughts?

I see. You're correct. Since the password stuff is handled client side in the component, probably it's good to exercise that code path. I'll update the PR

tcnghia:
> > > > Looks good to me. I noticed that the component mentions a password for the cert. Is there value in creating a password protected cert and ensuring that works as well?
> > > > https://github.com/dapr/components-contrib/blob/master/secretstores/azure/keyvault/keyvault.go#L23
> > > 
> > > 
> > > Since our cert is downloaded from KeyVault and cleaned up right away after done, I think we don't need to use a password like one that gets stored for a while (like on a dev laptop). If we add a password, it will need to be stored in the secret store anyway, which doesn't offer us more protection.
> > 
> > 
> > I'm not worried about the protection so much as if it works or not as a path for the component. Then again, we're getting dangerously close to testing the Azure sdk instead of the component, so I can go either way. Thoughts?
> 
> I see. You're correct. Since the password stuff is handled client side in the component, probably it's good to exercise that code path. I'll update the PR

So... the documented usecase does not use a password, since the certificate when provisioned using `az` tool doesn't allow specifying a password. I think the password option was for the cases where the users created their Service Principal and upload a self-signed cert that they generated themselves.

Also, looking at the code, we just pass the password (even if empty) to the library. So as far as our code is concerned, the same code path is exercised.

Given the two reasons above I think it is probably best to test the documented usecase, which is passworld-less cert.

@halspang WDYT

halspang:
@tcnghia I think we should test the documented use case as well, so I'm fine with using the passwordless cert.

halspang (approved):

mukundansundar (commented):
Are the secrets already manually created in the keyvault? Can you add the bulk get once #657 is closed?

tcnghia:
> Are the secrets already manually created in the keyvault? Can you add the bulk get once #657 is closed?

Yes. Added BulkGet.

See successful run https://github.com/tcnghia/components-contrib/runs/1841070549?check_suite_focus=true

tcnghia:
@mukundansundar it's ready for another look

tcnghia:
Rebased. Successful run is here https://github.com/tcnghia/components-contrib/actions/runs/542048981 

@artursouza can you please help review? thanks

mukundansundar (approved):

tcnghia:
@mukundansundar  can you please help merge? thanks

mukundansundar:
> @mukundansundar can you please help merge? thanks

@tcnghia I do not have permissions.

cc @artursouza @yaron2

artursouza (approved):

Issue #655, opened 2021-02-03
Azure Keyvault Secretstore's Bulk GET is broken


tcnghia:
## Expected Behavior

Azure KeyVault secret store BulkGet works as intended.

## Actual Behavior

BulkGet does not work because after listing the secrets, our code expected the `secretItem.ID` to have the format `"{vaultURI}/{secretName}"`, while in fact they currently have the format `"{vaultURI}/secrets/{secretName}"`.

See code https://github.com/dapr/components-contrib/blob/master/secretstores/azure/keyvault/keyvault.go#L106

## Steps to Reproduce the Problem

Just use Azure KeyVault secret store BulkGet.

## Release Note

RELEASE NOTE: **FIX** Azure KeyVault secret store BulkGet.

cc @cmendible @yaron2

PR #657, opened 2021-02-03
Fix Azure KeyVault secretstore BulkGet URI parsing of the GetSecretsComplete response

tcnghia:
# Description
Fix Azure KeyVault secretstore BulkGet URI parsing of the GetSecretsComplete response to make it works with URI of form `{vaultURI}/secrets/{itemID}` in the GetSecretsComplete response.

## Issue reference

Please reference the issue this PR will close: #655 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/secretstores/azure/keyvault/keyvault.go
+++ b/secretstores/azure/keyvault/keyvault.go
@@ -102,9 +102,11 @@ func (k *keyvaultSecretStore) BulkGetSecret(req secretstores.BulkGetSecretReques
 		Data: map[string]map[string]string{},
 	}
 
+	secretIDPrefix := vaultURI + secretItemIDPrefix
+
 	for secretsResp.NotDone() {
 		secretItem := secretsResp.Value()
-		secretName := strings.TrimPrefix(strings.TrimPrefix(*secretItem.ID, vaultURI), secretItemIDPrefix)
+		secretName := strings.TrimPrefix(*secretItem.ID, secretIDPrefix)
 
 		secretResp, err := k.vaultClient.GetSecret(context.Background(), vaultURI, secretName, "")
 		if err != nil {
--- a/secretstores/azure/keyvault/keyvault.go
+++ b/secretstores/azure/keyvault/keyvault.go
@@ -25,6 +25,7 @@ const (
 	componentSPNTenantID            = "spnTenantId"
 	componentVaultName              = "vaultName"
 	VersionID                       = "version_id"
+	secretItemIdPrefix              = "/secrets/"
 )
 
 type keyvaultSecretStore struct {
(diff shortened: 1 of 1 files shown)


tcnghia:
/assign @mukundansundar @cmendible 

Please do not merge until code freeze is lifted. We will get this in after Milestone 3.

tcnghia:
failing test seems unrelated, probably flakey

cmendible (changes requested):
Instead of doing the double trim you can add in line 89 (before the loop) something like:

``` go
secretIDPrefix := fmt.Sprintf("%s%s", vaultURI, secretItemIDPrefix)
```

and then just:

``` go
secretName := strings.TrimPrefix(*secretItem.ID, secretIDPrefix)
```

tcnghia:
> Instead of doing the double trim you can add in line 89 (before the loop) something like:
> 
> ```go
> secretIDPrefix := fmt.Sprintf("%s%s", vaultURI, secretItemIDPrefix)
> ```
> 
> and then just:
> 
> ```go
> secretName := strings.TrimPrefix(*secretItem.ID, secretIDPrefix)
> ```

I don't know enough about the history of Azure KeyVault API to know if this URL format has been always like this. If previously the BulkGet works, then probably at some point the URL format was different.

Do you know if there is official documentation regarding the current URL format? Otherwise, I think we may be best backward compatible with the previous format as well, which the doubly `TrimPrefix` will offer.

cmendible:
> > Instead of doing the double trim you can add in line 89 (before the loop) something like:
> > 
> > ```go
> > secretIDPrefix := fmt.Sprintf("%s%s", vaultURI, secretItemIDPrefix)
> > ```
> > 
> > and then just:
> > 
> > ```go
> > secretName := strings.TrimPrefix(*secretItem.ID, secretIDPrefix)
> > ```
> 
> I don't know enough about the history of Azure KeyVault API to know if this URL format has been always like this. If previously the BulkGet works, then probably at some point the URL format was different.
> 
> Do you know if there is official documentation regarding the current URL format? Otherwise, I think we may be best backward compatible with the previous format as well, which the doubly `TrimPrefix` will offer.
> 
> 
> 
> 

The sufix is specified here: https://docs.microsoft.com/en-us/azure/key-vault/general/about-keys-secrets-certificates#object-types

I made a mistake with a stash on my side that never made it to the repo so I'm positive that the current implementation didn't work 😟. The API hasn't changed.

tcnghia:
> > > Instead of doing the double trim you can add in line 89 (before the loop) something like:
> > > ```go
> > > secretIDPrefix := fmt.Sprintf("%s%s", vaultURI, secretItemIDPrefix)
> > > ```
> > > 
> > > 
> > > and then just:
> > > ```go
> > > secretName := strings.TrimPrefix(*secretItem.ID, secretIDPrefix)
> > > ```
> > 
> > 
> > I don't know enough about the history of Azure KeyVault API to know if this URL format has been always like this. If previously the BulkGet works, then probably at some point the URL format was different.
> > Do you know if there is official documentation regarding the current URL format? Otherwise, I think we may be best backward compatible with the previous format as well, which the doubly `TrimPrefix` will offer.
> 
> The sufix is specified here: https://docs.microsoft.com/en-us/azure/key-vault/general/about-keys-secrets-certificates#object-types
> 
> I made a mistake with a stash on my side that never made it to the repo so I'm positive that the current implementation didn't work 😟. The API hasn't changed.

@cmendible Thanks. Addressed the feedback as suggested.

cmendible (approved):
LGTM

mukundansundar (approved):
lgtm

artursouza (approved):

artursouza (approved):

PR #661, opened 2021-02-06
Improving the versatility of the HTTP output binding

pkedy:
# Description

Improving the versatility of the HTTP output binding. Removed the `Read` method since it is not an input binding.

## Issue reference

Related to dapr/docs#153. Will update the docs for this binding as a follow up.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#153

### DIFF
--- a/bindings/http/http.go
+++ b/bindings/http/http.go
@@ -87,6 +87,7 @@ func (h *HTTPSource) Invoke(req *bindings.InvokeRequest) (*bindings.InvokeRespon
 		body = bytes.NewBuffer(req.Data)
 	}
 
+	// nolint: noctx
 	request, err := http.NewRequest(method, u, body)
 	if err != nil {
 		return nil, err
@@ -97,7 +98,6 @@ func (h *HTTPSource) Invoke(req *bindings.InvokeRequest) (*bindings.InvokeRespon
 	}
 	request.Header.Set("Accept", "application/json; charset=utf-8")
 
-	// nolint: noctx
 	resp, err := h.client.Do(request)
 	if err != nil {
 		return nil, err
--- a/bindings/http/http.go
+++ b/bindings/http/http.go
@@ -16,7 +16,6 @@ import (
 	"time"
 
 	"github.com/mitchellh/mapstructure"
-	"github.com/pkg/errors"
 
 	"github.com/dapr/components-contrib/bindings"
 	"github.com/dapr/dapr/pkg/logger"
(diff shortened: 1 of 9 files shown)


pkedy:
Please review @yaron2 @artursouza

mukundansundar (commented):
@pkedy If Read is being removed ... please change to conformance test config too ....

mukundansundar (commented):
@pkedy 
This seems like a breaking change ... it removes HTTP as an input binding.  cc @yaron2 @artursouza 

This also breaks the current conformance tests defined for HTTP binding.

bindings/http/http.go
```diff
@@ -36,69 +43,126 @@ func NewHTTP(logger logger.Logger) *HTTPSource {
...
+		body = bytes.NewBuffer(req.Data)
+	case "GET", "HEAD", "DELETE", "OPTIONS", "TRACE":
+	default:
+		return nil, fmt.Errorf("invalid operation: %s", req.Operation)
 	}
 
-	return b, nil
-}
-
-func (h *HTTPSource) Read(handler func(*bindings.ReadResponse) error) error {
```
artursouza, line 75: Why removing Read?

artursouza (changes requested):

yaron2:
> @pkedy
> This seems like a breaking change ... it removes HTTP as an input binding. cc @yaron2 @artursouza
> 
> This also breaks the current conformance tests defined for HTTP binding.

@mukundansundar this actually fixes the HTTP binding as it was never meant to be an input binding. the test must change accordingly. /cc @pkedy

artursouza:
> > @pkedy
> > This seems like a breaking change ... it removes HTTP as an input binding. cc @yaron2 @artursouza
> > This also breaks the current conformance tests defined for HTTP binding.
> 
> @mukundansundar this actually fixes the HTTP binding as it was never meant to be an input binding. the test must change accordingly. /cc @pkedy

Why not?

yaron2:
> > > @pkedy
> > > This seems like a breaking change ... it removes HTTP as an input binding. cc @yaron2 @artursouza
> > > This also breaks the current conformance tests defined for HTTP binding.
> > 
> > 
> > @mukundansundar this actually fixes the HTTP binding as it was never meant to be an input binding. the test must change accordingly. /cc @pkedy
> 
> Why not?

In the older implementation, it was an input binding, and input bindings are blocking as they continuously get events. So for the HTTP binding, in practice this meant that the URL defined in the Component was only invoked *once* during startup and never again.

The HTTP binding was really meant to be a bi-directional binding where a user can invoke an external system over HTTP and get the data back.

bindings/http/http.go
yaron2: See here: https://github.com/dapr/components-contrib/pull/661#issuecomment-775525544

artursouza:
It can be input binding where it gets dat

> > > > @pkedy
> > > > This seems like a breaking change ... it removes HTTP as an input binding. cc @yaron2 @artursouza
> > > > This also breaks the current conformance tests defined for HTTP binding.
> > > 
> > > 
> > > @mukundansundar this actually fixes the HTTP binding as it was never meant to be an input binding. the test must change accordingly. /cc @pkedy
> > 
> > 
> > Why not?
> 
> In the older implementation, it was an input binding, and input bindings are blocking as they continuously get events. So for the HTTP binding, in practice this meant that the URL defined in the Component was only invoked _once_ during startup and never again.
> 
> The HTTP binding was really meant to be a bi-directional binding where a user can invoke an external system over HTTP and get the data back.

The input binding could trigger an infinite loop getting from the URL every second or so. I am fine removing it for now.

artursouza (approved):

PR #664, opened 2021-02-06
Handle []byte in cosmosdb state txn

rynowak:
# Description

The issue here is that we had missed the case of pre-marshaled bytes in
the state transaction code path, which would result in the encoding/json
package doing it's normal thing and base64 encoding the data as a JSON
string.

The fix is to use json.RawMessage to avoid that. Turns out this code can
be simplified a bit.

## Issue reference

Fixes: dapr/dotnet-sdk#579

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/state/azure/cosmosdb/cosmosdb.go
+++ b/state/azure/cosmosdb/cosmosdb.go
@@ -6,7 +6,6 @@
 package cosmosdb
 
 import (
-	"bytes"
 	"encoding/json"
 	"fmt"
 	"strings"
@@ -41,16 +40,6 @@ type CosmosItem struct {
 	ID           string      `json:"id"`
 	Value        interface{} `json:"value"`
 	PartitionKey string      `json:"partitionKey"`
-}
-
-// CosmosItemWithRawMessage is a version of CosmosItem with a Value of RawMessage so this field
-// is not marshalled.  If it is marshalled it will end up in a form that
-// cannot be unmarshalled.  This type is used when SetRequest.Value arrives as bytes.
-type CosmosItemWithRawMessage struct {
-	documentdb.Document
-	ID           string              `json:"id"`
-	Value        jsoniter.RawMessage `json:"value"`
-	PartitionKey string              `json:"partitionKey"`
 }
 
 type storedProcedureDefinition struct {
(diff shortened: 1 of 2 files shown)


state/azure/cosmosdb/cosmosdb.go
```diff
@@ -43,16 +42,6 @@ type CosmosItem struct {
 	PartitionKey string      `json:"partitionKey"`
 }
 
-// CosmosItemWithRawMessage is a version of CosmosItem with a Value of RawMessage so this field
-// is not marshalled.  If it is marshalled it will end up in a form that
-// cannot be unmarshalled.  This type is used when SetRequest.Value arrives as bytes.
-type CosmosItemWithRawMessage struct {
```
rynowak, line 49: Using a separate type would complicated the code that handles state transactions. Turns out we don't need a separate type for this case, because `CosmosItem.Value` is already an `interface{}`

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -214,20 +203,8 @@ func (c *StateStore) Set(req *state.SetRequest) error {
 		options = append(options, documentdb.ConsistencyLevel(documentdb.Eventual))
 	}
 
-	b, ok := req.Value.([]uint8)
-	if ok {
-		// data arrived in bytes and already json.  Don't marshal the Value field again.
-		item := CosmosItemWithRawMessage{ID: req.Key, Value: b, PartitionKey: partitionKey}
-		var marshalled []byte
-		marshalled, err = convertToJSONWithoutEscapes(item)
```
rynowak, line 222: This code was relying on the fact that the cosmosdb APIs handle `[]byte` and special case it. This approach isn't really needed, since we can just use `CosmosItem` for both cases.

state/azure/cosmosdb/cosmosdb_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/dapr/components-contrib/state"
+	"github.com/stretchr/testify/assert"
+)
+
+type widget struct {
+	Color string `json:"color"`
+}
+
+func TestCreateCosmosItem(t *testing.T) {
```
rynowak, line 20: Tests basically verify we can convert `state.SetRequest` -> `CosmosItem` and that it produces the right JSON regardless of how the value was passed in.

state/azure/cosmosdb/cosmosdb_test.go
```diff
@@ -0,0 +1,78 @@
...
+		req := state.SetRequest{
+			Key:   "testKey",
+			Value: bytes,
+		}
+
+		item := createUpsertItem(req, partitionKey)
+		assert.Equal(t, partitionKey, item.PartitionKey)
+		assert.Equal(t, "testKey", item.ID)
+
+		// items need to be marshallable to JSON with encoding/json
```
rynowak, line 64: I know we prefer jsoniter, but encoding/json is used inside the cosmosdb client - so we should test that.

state/azure/cosmosdb/cosmosdb.go
artursouza: Is this a backwards compatible change in the sense that data stored with the old code can be read with this version?

yaron2 (approved):
LGTM but please address this:
https://github.com/dapr/components-contrib/pull/664/files#r571746713

state/azure/cosmosdb/cosmosdb.go
rynowak: So what happens with the bug is that your content is stored as a base64 encoded JSON string. So if the scenario you're talking about is using the SDK you can't round-trip that data because deserialization would fail.

If someone used the generated gRPC client to retrieve the data and manually deserialize it from the base64 string, then yes, this would back that use case.

🌈 Every change is a breaking change to someone 🌈

artursouza: OK. LGTM.

Issue #635, opened 2021-01-26
Include hierarchical mimetype check


artursouza:
## Describe the proposal

We do mime type checks in the code (especially in PubSub). The logic now is naive, it would be more future proof to have a hierarchical comparison solution. This library has hierarchy of mime types, for example: https://github.com/gabriel-vasile/mimetype

Example of logic we have currently for mimetype: https://github.com/dapr/components-contrib/pull/631/files

Ideally, we would reuse a library and not implement our own hierarchy.

Note: this issue does not include mimetype detection, this is just for the comparison logic.
Note2: this issue is NOT proposing to use content type detection but to *REUSE* the mimetype hierarchy from an existing library.

rynowak:
> The logic now is naive, it would be more future proof to have a hierarchical comparison solution, similar to what is done here

I would argue that this specific library would be shift in the wrong direction. What's going on in this library is **content type detection based on content**. Since we provide an explicit property for content type as part of the API, 'sniffing' the content type based on the content is not appropriate. This is something we'd likely have to security review as well, as tricking clients into misinterpreting content types and text encodings is a significant attack vector for browsers (as well as possibly other systems). 

Instead, I'd suggest we adopt a library that does content type parsing since we always expect the caller to specify the type. That would replace naive checks like `strings.HasPrefix`. We already have bugs in this area that no one has reported yet .. for example we have special handling for JSON, but it's based on `application/json` and wouldn't accept a content type like `application/vnd.widget+json` which is also valid. We also assume that that the content is UTF8 instead of looking for a `charset` parameter.

artursouza:
@rynowak This issue is *NOT* proposing to use content type detection but the hierarchy that this library has.

rynowak:
Cool, it sounds like we agree then. The library has this in its description (emphasis mine): 

> A fast golang library for MIME type and file extension **detection, based on magic numbers**

PR #674, opened 2021-02-09
Handles binary data in cloud event.

artursouza:
# Description

Handles binary data in cloud event.

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/2612

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/contenttype/utils.go
+++ b/contenttype/utils.go
@@ -33,6 +33,11 @@ func IsStringContentType(contentType string) bool {
 	return isContentType(contentType, "application/xml")
 }
 
+// IsBinaryContentType determines if content type is byte[]
+func IsBinaryContentType(contentType string) bool {
+	return isContentType(contentType, "application/octet-stream")
+}
+
 func isContentType(contentType string, expected string) bool {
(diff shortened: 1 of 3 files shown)


contenttype/utils.go
```diff
@@ -33,6 +33,11 @@ func IsStringContentType(contentType string) bool {
 	return isContentType(contentType, "application/xml")
 }
 
+// IsBinaryContentType determines if content type is byte[]
+func IsBinaryContentType(contentType string) bool {
+	return isContentType(contentType, "application/octet-stream")
```
jigargandhi, line 38: This is fine for now, later we will add proper content type detection

jigargandhi (approved):
Thank you for adding this

contenttype/utils.go
artursouza: Correct. We should use a richer detection using MimeType hierarchy. This issue is about that: https://github.com/dapr/components-contrib/issues/635

yaron2 (approved):
lgtm

Issue #407, opened 2020-07-23
Conformance Tests - Implement API Tests for StateStore


mukundansundar:
## Describe the proposal
Implement a sample test suite for Redis StateStore. #404 
Test the statestore interface completely.
Generate a report on test end. 
The redis setup is external to this for now. 
List and create issues for other tests that need to implemented as needed.
#404 

- [x] basic APIs init, set, get, delete, bulkset, bulkdelete
- [ ] transaction APIs

mukundansundar:
The basic set is convered in the #586

mukundansundar:
@artursouza Can you implement the tests for the transactions API once #586 is merged?

artursouza:
@mukundansundar Yes, I can.

mukundansundar:
@artursouza @yaron2 Please add a `breaking-change` label to this issue. 

The current label available is `breaking-changes`

PR #680, opened 2021-02-10
Conf state trx + Change mongodb get result for key not found  + Fix byte[] in cosmosdb

artursouza:
# Description

Validation in this run:
https://github.com/dapr/components-contrib/runs/1868788838?check_suite_focus=true

## Issue reference

Please reference the issue this PR will close: #407 
Will also close: #684 
Will also close: #685

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/state/mongodb/mongodb.go
+++ b/state/mongodb/mongodb.go
@@ -172,6 +172,7 @@ func (m *MongoDB) Get(req *state.GetRequest) (*state.GetResponse, error) {
 			// To behave the same as other state stores in conf tests.
 			return &state.GetResponse{}, nil
 		}
+
 		return &state.GetResponse{}, err
 	}
 
--- a/state/mongodb/mongodb.go
+++ b/state/mongodb/mongodb.go
@@ -167,6 +167,11 @@ func (m *MongoDB) Get(req *state.GetRequest) (*state.GetResponse, error) {
 	filter := bson.M{id: req.Key}
 	err := m.collection.FindOne(ctx, filter).Decode(&result)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


state/mongodb/mongodb.go
```diff
@@ -167,6 +167,11 @@ func (m *MongoDB) Get(req *state.GetRequest) (*state.GetResponse, error) {
 	filter := bson.M{id: req.Key}
 	err := m.collection.FindOne(ctx, filter).Decode(&result)
 	if err != nil {
+		if err.Error() == "mongo: no documents in result" {
```
mukundansundar, line 170: @yaron2 @artursouza Isn't this a breaking change? 
My question is do we want to change it this way now? 
Should the code be changed specifically for conf-tests? Or should it be handled as a special case in conf-tests?

yaron2: If this is a breaking change for the component, this is all the more reason to do this now.

artursouza: We better break it now then later.

state/redis/redis.go
```diff
@@ -374,3 +372,17 @@ func (r *StateStore) parseETag(req *state.SetRequest) (int, error) {
...
+	if value == nil {
+		return nil
+	}
+
+	s, ok := value.(string)
+	if ok {
+		return []byte(s)
+	}
+
+	bt, _ := utils.Marshal(value, r.json.Marshal)
```
mukundansundar, line 386: propagate err from here ...

yaron2 (approved):

Issue #695, opened 2021-02-17
Conformance Tests - PubSub: Pulsar


pkedy:
Add Pulsar to the pubsub conformance tests.

- Add docker-compose file
- Add `pubsub.pulsar` to `PR_COMPONENTS` in `conformance.yml`
- Add step to start Pulsar in `conformance.yml`
- Add component config to `/tests/config/pubsub`
- Add the creation of the component in `loadPubSub` of `/tests/conformance/common.go`
- Make any changes needed to the Pulsar component to make it pass

mukundansundar:
closed by #702

Issue #578, opened 2021-01-07
Not all pubsub implementations respect concurrency settings


trondhindenes:
## In what area(s)?
/area runtime

## What version of Dapr?
1.0.0-rc.1

## Expected Behavior
When configuring dapr with a rabbitmq pubsub and setting the concurrency to a value higher than 1, I would expect multiple requests to be sent to the app in paralell - new messages should not have to wait for current messages to be finished processing

## Actual Behavior
At least for rabbitmq pubsub, concurrency has no effect. The dapr will never handle more than one message at a time.

## Steps to Reproduce the Problem
configure a web api and issue multuple requests in paralell. Configre the endpoint sleep a few secs to make it easy to see that only one request will be processing at any given time, regardless of the concurrency setting

I would argue that this is a critical bug, especially since different pubsub implementations seem to have different behavior. Redis seems to respect concurrensy settings, while rabbitmq does not.

Issue #555, opened 2020-12-14
Redis pubsub messages are not properly handled


trondhindenes:
## Expected Behavior
I'm not an expert in redis streams, but we're working on seeing what kind of metrics we can get out of it in order to see if it's viable to implement dapr-pubsub using redis streams.

The expected behavior is that when dapr has successfully delivered a message from sender to receiver the message should be handled according to the specification of the underlying pubsub driver (redis streams in this case). 

This means calling the `XACK` command on the message(s) to remove them from the list of "pending messages" (this is documented at https://redis.io/commands/xack)

## Actual Behavior
It looks to me as if dapr never XACK's a message, which casues it to stay in redis in a pending state. Over time, this buildup is dangerous and it also causes unecessary re-delivery of messages.

This buildup of pending messages can be seen by issuing a `XPENDING` command, or by using the redis_exporter docker image (this way there's no need to know the names of streams/consumer groups before hand, so it's a little easier):
```bash
docker run -d -e REDIS_EXPORTER_CHECK_STREAMS="*" --net=host docker.io/oliver006/redis_exporter
curl http://localhost:9121/metrics | grep redis_stream
```

The cause seems to be this part of the redis implementation:
https://github.com/dapr/components-contrib/blob/e881e4ba28e34fdb26a89533a8ee881fb5d87b79/pubsub/redis/redis.go#L150-L153

Messages are acked if the `err` object is empty, but by implementing a bit of extra logging:
```
				err := handler(&msg)
				if err != nil {
					r.logger.Infof("error: %s", err)
				}
				if err == nil {
					r.logger.Infof("acking message %s in stream %s and group %s", message.ID, stream, consumerID)
					r.client.XAck(stream, consumerID, message.ID).Result()
				}
```
At least on my machine, every message causes an error:
```
INFO[0000] error: unknown status returned from app while processing pub/sub event 1753be78-2a1c-46c0-b6e8-a06362a06c14: ok  app_id=dummy-plugin-dev instance=trond-x1 scope=dapr.contrib type=log ver=edge
INFO[0000] error: unknown status returned from app while processing pub/sub event 833f4d53-898e-4c12-a957-025c4b2b8af6: ok  app_id=dummy-plugin-dev instance=trond-x1 scope=dapr.contrib type=log ver=edge
[...]

PR #702, opened 2021-02-18
Improve conformance tests and add pulsar

### LINKED ISSUE #696: Conformance Tests - Add more checks to PubSub
Add the following checks:

- Detect out of order processing when the messaging broker supports in-order delivery
- Ignore messages that may be be sent by a previous test run (helps local development)
- Increase the default timeout but stop waiting when all the expected messages are processed
- Fail every X message in addition to the first 2 to prevent false positives

pkedy:
Summary of changes:

- Adds all the additional PubSub conformance tests listed by #696
- Adds Pulsar to the conformance test per #695 and fixes to conform with the tests
- Fixes to NATS Streaming to conform with the tests
- Both Pulsar and NATS Streaming now use the [same backoff/retry mechanism](https://github.com/cenkalti/backoff) in the Kafka component as a first step to supporting in-order message processing
- Implemented a common retry/notify/recovery helper in `pubsub/retry.go` which could eventually include a function for parsing backoff/retry settings from metadata settings

Resolves #696, #695, #710 (Redis refactoring required to pass these improved checks)

The most important thing to note in this PR is that all of the pubsub components included in the conformance tests are using a shared retry mechanism as a first step in supporting in-order message processing. I will be submitting a follow up issue+PR to add backoff/retry settings which would include a `backoffMaxRetries` setting so that if the user wants to skips a message that the application cannot handle, it can.

Longer term there could be other settings to allow for parallelism, redelivery (Nack'ing), and possibly dead letter queues. As @trondhindenes points out in #578, ultimately the user needs to be able to easily reason about these settings across all of the pubsub components. Dapr has a huge opportunity to demystify streaming and pubsub. I will be typing up more thoughts on this in a separate issue for discussion.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


pubsub/natsstreaming/natsstreaming.go
```diff
@@ -290,6 +317,7 @@ func genRandomString(n int) string {
 }
 
 func (n *natsStreamingPubSub) Close() error {
+	n.cancel()
```
mukundansundar, line 320: similar to this, close background context for Kafka pubusb?

pubsub/pulsar/pulsar.go
```diff
@@ -96,38 +110,61 @@ func (p *Pulsar) Subscribe(req pubsub.SubscribeRequest, handler func(msg *pubsub
...
-	for cm := range msgs.Chan() {
-		p.HandleMessage(cm, topic, handler)
+func (p *Pulsar) ListenMessage(consumer pulsar.Consumer, topic string, handler func(msg *pubsub.NewMessage) error) {
+	defer consumer.Close()
+
+	for {
+		select {
+		case msg := <-consumer.Chan():
+			if err := p.HandleMessage(msg, topic, handler); err != nil {
+				p.logger.Errorf("Error processing message")
```
mukundansundar, line 134: There is no detail in this log. And it will print for every error message, should this be a debug log?

tests/conformance/pubsub/pubsub.go
```diff
@@ -137,14 +179,17 @@ func ConformanceTests(t *testing.T, props map[string]string, ps pubsub.PubSub, c
...
+			for k := 1; k <= config.messageCount; k++ {
+				data := []byte(fmt.Sprintf("%s%d", dataPrefix, k))
 				err := ps.Publish(&pubsub.PublishRequest{
 					Data:       data,
 					PubsubName: config.pubsubName,
 					Topic:      config.testTopicName,
 					Metadata:   config.publishMetadata,
 				})
+				if err == nil {
+					awaitingMessages[string(data)] = struct{}{}
```
mukundansundar, line 191: use sets instead of this?

pkedy: In Go, `map[string]struct{}` is a set.

pubsub/natsstreaming/natsstreaming.go
pkedy: In the Kafka pubsub, `closeSubscripionResources` handles calling cancel because it creates a new context/cancel in each call to `Subscribe`. It does feel like maybe the context should live for the duration of the component and not for the subscription. The feels large enough for its own issue and PR but I :100: agree.

pubsub/pulsar/pulsar.go
pkedy: Good catch.  I added more context on the message here and also filtering out `context.Cancelled`. More work will be done here in the follow up PR regard backoff configuration I mentioned in the description.

tests/conformance/pubsub/pubsub.go
pkedy: I'd like to stick with simple Go maps as sets instead of using `k8s.io/apimachinery/pkg/util/sets` as this use case is simple.

artursouza:
Let me take a look at this in the weekend. This is great PR.

pkedy:
I disabled in-order checking and simulating errors for Redis because right now the Redis pubsub component does not have retries and its also firing message processing off in go routines so order is not predictable. The Redis component will be enhanced in a future PR to address this and #555. The disabled flags will be removed then.

pubsub/kafka/kafka.go
```diff
@@ -60,35 +61,29 @@ type consumer struct {
...
 		}
-
-		var warningLogged bool
-		err := backoff.RetryNotify(func() error {
-			err := consumer.callback(&pubsub.NewMessage{
-				Topic: claim.Topic(),
-				Data:  message.Value,
-			})
+		if err := pubsub.RetryNotifyRecover(func() error {
+			consumer.logger.Debugf("Processing Kafka message: %s/%d/%d [key=%s]", message.Topic, message.Partition, message.Offset, asBase64String(message.Key))
```
artursouza, line 75: This means that every message would have the key base64 encoded. Please, remove this part for this log.

pubsub/kafka/kafka.go
```diff
@@ -368,3 +367,9 @@ func (k *Kafka) Close() error {
 func (k *Kafka) Features() []pubsub.Feature {
 	return nil
 }
+
+type asBase64String []byte
```
artursouza, line 371: This is overkill. Just invoke the one-liner directly: `base64.StdEncoding.EncodeToString(s)`

tests/config/pubsub/tests.yml
```diff
@@ -15,8 +17,12 @@ components:
       testTopicName: dapr-conf-test
   - component: redis
     allOperations: true
+    config:
+      checkInOrderProcessing: false
+      simulateErrors: false
```
artursouza, line 22: Why redis cannot simulate errors?

tests/conformance/pubsub/pubsub.go
```diff
@@ -6,35 +6,42 @@
...
-	defaultMessageCount          = 10
-	defaultMaxReadDuration       = 20 * time.Second
-	defaultWaitDurationToPublish = 5 * time.Second
+	defaultPubsubName             = "pubusub"
+	defaultTopicName              = "testTopic"
+	defaultMessageCount           = 10
+	defaultMaxReadDuration        = 60 * time.Second
+	defaultWaitDurationToPublish  = 5 * time.Second
+	defaultCheckInOrderProcessing = true
+	defaultSimulateErrors         = true
```
artursouza, line 31: Why simulate errors is configurable?

artursouza (changes requested):

pubsub/kafka/kafka.go
pkedy: That's actually not happening here. The reason for the `asBase64String` alias type (that implements [Stringer](https://golang.org/pkg/fmt/#Stringer)) is that it postpones the base64 encoding operation until the log is actually written. Since this is logged at Debug level, base64 is never invoked in the happy path. IMO since the key typically correlates to an entity ID in the user's overall architecture, it provides valuable context in the log.

pubsub/kafka/kafka.go
pkedy: See above

tests/config/pubsub/tests.yml
pkedy: Currently Redis does not have the retry logic that Kafka and Pulsar have. It also fires off each message in a separate gorouting which breaks in-order processing. I didn't want to make this PR too big by making significant changes to the Redis component without careful consideration (as an example maybe there be a setting for parallel processing/more than 1 in-flight message). These changes are likely non-trivial and IMO should be review independently.

My thought here was to add the simulate errors flag to allow the conformance test to pass for now. I think once we get all of the pubsub components added and conforming to the rules we envision, this setting can be removed. I'll add a comment to that affect. This is a baby step in thinking about the behaviors and settings of the pubsub components more holistically. Sorry, This was a late finding and I should have commented on the rationale as opposed to just relying on the commit message.

tests/conformance/pubsub/pubsub.go
pkedy: See above. I'm putting a comment here to consider removing later.

tests/config/pubsub/tests.yml
artursouza: So, the Redis implementation cannot handle errors? That is surprising. I can see why it does not support oder in case of error of a single message but cannot recover from an error is another statement.

pubsub/kafka/kafka.go
artursouza: OK. Please, add comment about this to avoid someone reverting this later.

tests/config/pubsub/tests.yml
pkedy: I didn't say it couldn't handle errors, just that we did't have retries there yet. :)
Currently, in case of an error, the message is not Ack'ed and I believe Redis attempts to redeliver it in the future. From what I observed, it is not always redelivered within the timeframe of our conformance test (which is bumped up to 60s in this PR). Allowing for redelivery of course also breaks in-order processing but that will be made configurable in an upcoming issue+PR.

pubsub/kafka/kafka.go
pkedy: Good call. I'll put the comment above `asBase64String`

tests/config/pubsub/tests.yml
artursouza: Please, keep redis in the test for errors and remove this configuration. It was passing consistently before this change. It might have to do with the extra errors added here. Making this configurable means that components might choose not handle message delivery and still be conformant.

pkedy: Redis was passing because the conformance test did not have the granular message tracking check that this PR introduced so it passing before was a false positive. I went ahead and enhanced the Redis component to support redelivery (reclaiming messages) which makes the test pass without the need of the temporary flag (which I also removed).

artursouza: Outstanding. I will take a look at this again soon.

pubsub/redis/metadata.go
```diff
@@ -5,9 +5,17 @@
...
-	host       string
-	password   string
-	consumerID string
-	enableTLS  bool
+	host              string
+	password          string
+	consumerID        string
+	enableTLS         bool
+	processingTimeout time.Duration
+	reclaimInterval   time.Duration
```
artursouza, line 18: Please, create/update an issue mentioning this new fields, so they can be in the release notes.

pubsub/redis/redis.go
```diff
@@ -37,7 +57,13 @@ func NewRedisStreams(logger logger.Logger) pubsub.PubSub {
 }
 
 func parseRedisMetadata(meta pubsub.Metadata) (metadata, error) {
-	m := metadata{}
+	// Default values
+	m := metadata{
+		processingTimeout: 5 * time.Second,
```
artursouza, line 62: Create a docs issue in `dapr/docs` to update docs for the new metadata and default values.

pubsub/redis/redis.go
```diff
@@ -113,68 +180,215 @@ func (r *redisStreams) Publish(req *pubsub.PublishRequest) error {
...
-	res, err := r.client.XReadGroup(&redis.XReadGroupArgs{
-		Group:    consumerID,
-		Consumer: consumerID,
-		Streams:  []string{stream, start},
-		Block:    0,
-	}).Result()
-	if err != nil {
-		return nil, err
-	}
+func (r *redisStreams) enqueue(stream, consumerID string, handler func(msg *pubsub.NewMessage) error, msgs []redis.XMessage) {
```
artursouza, line 196: Please, call out this refactoring as a separate issue with release notes.

pubsub/redis/redis.go
```diff
@@ -113,68 +180,215 @@ func (r *redisStreams) Publish(req *pubsub.PublishRequest) error {
...
 		start = ">"
+
+		// Return on cancelation
+		if r.ctx.Err() != nil {
+			return
+		}
+	}
+}
+
+func (r *redisStreams) reclaimMessagesLoop(stream string, handler func(msg *pubsub.NewMessage) error) {
```
artursouza, line 281: I need some explanation to understand this.

pubsub/redis/redis_test.go
```diff
@@ -103,17 +109,20 @@ func TestProcessStreams(t *testing.T) {
...
 	testRedisStream := &redisStreams{logger: logger.NewLogger("test")}
-	testRedisStream.processStreams(fakeConsumerID, generateRedisStreamTestData(2, 3, expectedData), fakeHandler)
+	testRedisStream.ctx, testRedisStream.cancel = context.WithCancel(context.Background())
+	testRedisStream.queue = make(chan redisMessage, 10)
+	go testRedisStream.worker()
+	testRedisStream.enqueue(fakeConsumerID, fakeConsumerID, fakeHandler, generateRedisStreamTestData(2, 3, expectedData))
 
-	// sleep for 10ms to give time to finish processing
-	time.Sleep(time.Millisecond * 10)
+	// Wait for the handler tp finish processing
```
artursouza, line 117: s/tp/to

artursouza (commented):
My only concern now is the complexity of the new redis pubsub implementation. Can we sync 1-1 on this?

artursouza:
FYI, please add any issues being closed by this PR to the v1.1 milestone.

pubsub/redis/redis.go
```diff
@@ -89,20 +95,20 @@ func parseRedisMetadata(meta pubsub.Metadata) (metadata, error) {
...
-		if reclaimIntervalMS, err := strconv.ParseUint(val, 10, 64); err == nil {
-			m.reclaimInterval = time.Duration(reclaimIntervalMS) * time.Millisecond
+	if val, ok := meta.Properties[redeliverInterval]; ok && val != "" {
+		if redeliverIntervalMs, err := strconv.ParseUint(val, 10, 64); err == nil {
+			m.redeliverInterval = time.Duration(redeliverIntervalMs) * time.Millisecond
 		} else if d, err := time.ParseDuration(val); err == nil {
-			m.reclaimInterval = d
+			m.redeliverInterval = d
 		} else {
 			return m, fmt.Errorf("redis streams error: can't parse processingTimeout field: %s", err)
```
mukundansundar, line 112:
```suggestion
			return m, fmt.Errorf("redis streams error: can't parse redeliverInterval field: %s", err)
```

redeliverInterval field

pubsub/redis/redis.go
pkedy: The comments I added should help

pubsub/redis/redis.go
```diff
@@ -367,21 +408,20 @@ func (r *redisStreams) handlePendingWithoutMessages(stream string, messageIDs ma
 			MinIdle:  r.metadata.processingTimeout,
 			Messages: []string{pendingID},
 		}).Result()
-		if err != nil && err != redis.Nil {
+		if err != nil && !errors.Is(err, redis.Nil) {
 			r.logger.Errorf("error claiming pending Redis message %s: %v", pendingID, err)
 
 			return
```
mukundansundar, line 413: should this be continue instead of return? That is try to claim other pending Messages instead of returning on the first one that is not a redis.Nil error ?

pkedy: Yes!  Good catch!

pubsub/redis/redis.go
pkedy: Created https://github.com/dapr/docs/issues/1255

pubsub/redis/redis.go
pkedy: Created https://github.com/dapr/components-contrib/issues/710

pubsub/redis/metadata.go
pkedy: Will add to https://github.com/dapr/components-contrib/issues/710

artursouza (approved):

Issue #681, opened 2021-02-10
Support blob deletion for the Azure Blob Storage binding


akkie:
The blob storage binding currently has no option to delete a blob. It would be great if this could be supported.

https://docs.microsoft.com/en-us/rest/api/storageservices/delete-blob

PR #705, opened 2021-02-19
add delete to blob storage binding

fabistb:
# Description
Adds the delete functionality to the azure blob binding.

## Issue reference
Please reference the issue this PR will close: #681

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/issues/1277


mukundansundar (commented):
add a doc update for this in dapr/docs repo and the possible values for deleteSnapshotOptions. 

Add the delete test to the conformance test for azure bindings.

fabistb:
@mukundansundar 
Created the issue in the docs repo and I will work on it.
https://github.com/dapr/docs/issues/1277

While enhancing the conformance test I faced a small problem.
In `.../tests/conformance/bindings.go` at the _DeleteOperation_ section it will invoke a GET after the DELETE if a CREATE has been performed and GET is configured. Unfortunately the GET operation of the azure blob binding will always respond with an error if a blob doesn't exist which will cause the test to fail.

A suggestion from your side how to deal with that?

mukundansundar:
> @mukundansundar
> Created the issue in the docs repo and I will work on it.
> [dapr/docs#1277](https://github.com/dapr/docs/issues/1277)
> 
> While enhancing the conformance test I faced a small problem.
> In `.../tests/conformance/bindings.go` at the _DeleteOperation_ section it will invoke a GET after the DELETE if a CREATE has been performed and GET is configured. Unfortunately the GET operation of the azure blob binding will always respond with an error if a blob doesn't exist which will cause the test to fail.
> 
> A suggestion from your side how to deal with that?

@artursouza @pkedy 
For this part I am thinking the test for get after delete for bindings should be configurable and not really a necessity for conformance. What are your thoughts on this?

yaron2 (approved):
lgtm

artursouza:
> > @mukundansundar
> > Created the issue in the docs repo and I will work on it.
> > [dapr/docs#1277](https://github.com/dapr/docs/issues/1277)
> > While enhancing the conformance test I faced a small problem.
> > In `.../tests/conformance/bindings.go` at the _DeleteOperation_ section it will invoke a GET after the DELETE if a CREATE has been performed and GET is configured. Unfortunately the GET operation of the azure blob binding will always respond with an error if a blob doesn't exist which will cause the test to fail.
> > A suggestion from your side how to deal with that?
> 
> @artursouza @pkedy
> For this part I am thinking the test for get after delete for bindings should be configurable and not really a necessity for conformance. What are your thoughts on this?

I think that having consistent behavior in the APIs is important. In this case, we must define what is the behavior to invoke GET when a resource does not exist: error or return nil? We have made changes to MongoDB state store for its behavior to be consistent with others in a similar scenario. Now, MongoDB will return nil when fetching an item that does not exist.

Long story short, change what needs to be changes to make components consistent and not relax the test.

fabistb:
> > > @mukundansundar
> > > Created the issue in the docs repo and I will work on it.
> > > [dapr/docs#1277](https://github.com/dapr/docs/issues/1277)
> > > While enhancing the conformance test I faced a small problem.
> > > In `.../tests/conformance/bindings.go` at the _DeleteOperation_ section it will invoke a GET after the DELETE if a CREATE has been performed and GET is configured. Unfortunately the GET operation of the azure blob binding will always respond with an error if a blob doesn't exist which will cause the test to fail.
> > > A suggestion from your side how to deal with that?
> > 
> > 
> > @artursouza @pkedy
> > For this part I am thinking the test for get after delete for bindings should be configurable and not really a necessity for conformance. What are your thoughts on this?
> 
> I think that having consistent behavior in the APIs is important. In this case, we must define what is the behavior to invoke GET when a resource does not exist: error or return nil? We have made changes to MongoDB state store for its behavior to be consistent with others in a similar scenario. Now, MongoDB will return nil when fetching an item that does not exist.
> 
> Long story short, change what needs to be changed to make components consistent and not relax the test.

Thanks for the insights.
Just for my understanding. Would such a change qualify for an version of the binding?

yaron2:
> > > > @mukundansundar
> > > > Created the issue in the docs repo and I will work on it.
> > > > [dapr/docs#1277](https://github.com/dapr/docs/issues/1277)
> > > > While enhancing the conformance test I faced a small problem.
> > > > In `.../tests/conformance/bindings.go` at the _DeleteOperation_ section it will invoke a GET after the DELETE if a CREATE has been performed and GET is configured. Unfortunately the GET operation of the azure blob binding will always respond with an error if a blob doesn't exist which will cause the test to fail.
> > > > A suggestion from your side how to deal with that?
> > > 
> > > 
> > > @artursouza @pkedy
> > > For this part I am thinking the test for get after delete for bindings should be configurable and not really a necessity for conformance. What are your thoughts on this?
> > 
> > 
> > I think that having consistent behavior in the APIs is important. In this case, we must define what is the behavior to invoke GET when a resource does not exist: error or return nil? We have made changes to MongoDB state store for its behavior to be consistent with others in a similar scenario. Now, MongoDB will return nil when fetching an item that does not exist.
> > Long story short, change what needs to be changed to make components consistent and not relax the test.
> 
> Thanks for the insights.
> Just for my understanding. Would such a change qualify for an version of the binding?

Just to be clear, binding operations are a per-binding concept and can have different return values. They're not adhering to a strict API and the operation should return whatever makes most sense for the binding first, and take consistency with similar operations for other bindings (if any) after. So all in all, conformance tests should not enforce consistency on *bindings operations*.

PR #716, opened 2021-02-24
Add ETag operation to statestore conformance test

halspang:
# Description
ETags are used to enable concurrency guarantees and are part of the
statestore specification. Not all stores need to use them, however
if the store chooses to, it should be part of the certification
process. This commit adds a simple ETag test scenario that verifies
the basic behaviors of ETags.

https://github.com/dapr/components-contrib/issues/711

## Issue reference
Closes: #711 

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #733, opened 2021-03-02
Make state.GetResponse ETag to a *string

### LINKED ISSUE #731: GetResponse does not return nil eTag in state store
## Expected Behavior
GetResponse should return nil eTag for item not found in state store, so the ETag value can be consumed and passed in a SetRequest after the get operation.

## Actual Behavior

GetResponse does not return nil eTag in state store. Right now, code doing set after get with etag, must check if resulting eTag is empty and transform it to nil on the set request.

Do a GetState and then follow by SetState for an item that does not exist.

See Response objects: https://github.com/dapr/components-contrib/blob/master/state/responses.go
See Request objects: https://github.com/dapr/components-contrib/blob/master/state/requests.go

## Release Note

RELEASE NOTE:

pkedy:
# Description

Per #731, this PR allows the ETag in `state.GetResponse` and `state.BulkGetResponse` to be returned as a `nil`. This is to make these structs consistent with `state.SetRequest` and `state.DeleteRequest` and to be able to copy the `ETag` field as-is when going between read and write operations.

Since there are not e2e tests for each of the changed state components, there is no way to verify these changes without a significant time investment to test test manually. These changes are fairly simple so I did not go that route but wanted to mention the risk and recommend a high level of scrutiny when reviewing the code.

**If this PR is merged, an issue in `dapr/dapr` should be created to adjust to the `ETag` field being `*string` instead of `string`.**

## Issue reference

Resolves #731

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/state/aerospike/aerospike.go
+++ b/state/aerospike/aerospike.go
@@ -14,9 +14,11 @@ import (
 
 	as "github.com/aerospike/aerospike-client-go"
 	"github.com/aerospike/aerospike-client-go/types"
+	"github.com/agrea/ptr"
+	jsoniter "github.com/json-iterator/go"
+
 	"github.com/dapr/components-contrib/state"
 	"github.com/dapr/dapr/pkg/logger"
-	jsoniter "github.com/json-iterator/go"
 )
 
 // metadata values
@@ -171,7 +173,7 @@ func (aspike *Aerospike) Get(req *state.GetRequest) (*state.GetResponse, error) 
 
 	return &state.GetResponse{
 		Data: value,
-		ETag: fmt.Sprintf("%d", record.Generation),
+		ETag: ptr.String(strconv.FormatUint(uint64(record.Generation), 10)),
 	}, nil
 }
 
--- a/state/couchbase/couchbase.go
+++ b/state/couchbase/couchbase.go
@@ -10,11 +10,13 @@ import (
 	"fmt"
 	"strconv"
 
+	"github.com/agrea/ptr"
+	jsoniter "github.com/json-iterator/go"
+	"gopkg.in/couchbase/gocb.v1"
+
 	"github.com/dapr/components-contrib/state"
 	"github.com/dapr/components-contrib/state/utils"
 	"github.com/dapr/dapr/pkg/logger"
-	jsoniter "github.com/json-iterator/go"
-	"gopkg.in/couchbase/gocb.v1"
 )
 
 const (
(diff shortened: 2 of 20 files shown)


state/aerospike/aerospike.go
```diff
@@ -171,7 +173,7 @@ func (aspike *Aerospike) Get(req *state.GetRequest) (*state.GetResponse, error)
 
 	return &state.GetResponse{
 		Data: value,
-		ETag: fmt.Sprintf("%d", record.Generation),
+		ETag: ptr.String(strconv.FormatUint(uint64(record.Generation), 10)),
```
artursouza, line 176: Why this is better than sprintf?

state/couchbase/couchbase.go
```diff
@@ -184,7 +186,7 @@ func (cbs *Couchbase) Get(req *state.GetRequest) (*state.GetResponse, error) {
 
 	return &state.GetResponse{
 		Data: data.([]byte),
-		ETag: fmt.Sprintf("%d", cas),
+		ETag: ptr.String(strconv.FormatUint(uint64(cas), 10)),
```
artursouza, line 189: same

artursouza (changes requested):
I am sure there is a reason for not using sprintf here, just need to know why. Thanks.

artursouza:
I believe conformance tests is what can give us assurance of this change. For components that don't have it, it means they are not in GA, so there is even less risk of causing big issues. Anyway, the change is simple enough.

state/aerospike/aerospike.go
pkedy: It's not a matter of being better, just that this simple case does not require it. Unless you need the the formatting options that Sprintf offers (i.e. padding, left justification, multiple value placeholders), it is inefficient compared to using the `strconv` package. IMO linters should pick this up as it is very similar to [using `fmt.Sprintf("%s", str)` unnecessarily](https://staticcheck.io/docs/checks#S1025) or [unnecessary use of fmt.Sprint](https://staticcheck.io/docs/checks#S1039).

For comparison:
[Go's formatting code](https://github.com/golang/go/blob/2b50ab2aee75d3c361fcd1eb39e830e2e73056b6/src/fmt/print.go#L974-L1147)
[Go's integer conversion code](https://github.com/golang/go/blob/2b50ab2aee75d3c361fcd1eb39e830e2e73056b6/src/strconv/itoa.go#L105-L152)

state/couchbase/couchbase.go
pkedy: See above

mukundansundar:
> I believe conformance tests is what can give us assurance of this change. For components that don't have it, it means they are not in GA, so there is even less risk of causing big issues. Anyway, the change is simple enough.

should this be merged after #716 then?

artursouza (approved):

PR #735, opened 2021-03-03
fix duration for event grid conf test

mukundansundar:
# Description

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/tests/config/bindings/tests.yml
+++ b/tests/config/bindings/tests.yml
@@ -19,7 +19,7 @@ components:
   - component: azure.eventgrid
     operations: ["create", "operations", "read"]
     config:
-      readBindingTimeout: 240
+      readBindingTimeout: 240s
   - component: azure.storagequeues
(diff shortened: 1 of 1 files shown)


artursouza (approved):

pkedy (approved):
I missed this one from the decode PR.

JFYI: decoding duration first tries to decode the string as an integer. If that succeeds the integer is treated as milliseconds. Then it falls back on trying to parse it using `time.ParseDuration`.

artursouza:
> I missed this one from the decode PR.
> 
> JFYI: decoding duration first tries to decode the string as an integer. If that succeeds the integer is treated as milliseconds. Then it falls back on trying to parse it using `time.ParseDuration`.

Maybe it should simply fail if cannot ParseDuration then and we avoid these confusion going forward.

pkedy:
I would agree, but I put in the integer parsing for backward compatibility.

PR #739, opened 2021-03-04
Adding RabbitMQ to the conformance tests

### LINKED ISSUE #694: Conformance Tests - PubSub: RabbitMQ
Add RabbitMQ to the pubsub conformance tests.

- Add docker-compose file
- Add `pubsub.rabbitmq` to `PR_COMPONENTS` in `conformance.yml`
- Add step to start RabbitMQ in `conformance.yml`
- Add component config to `/tests/config/pubsub`
- Add the creation of the component in `loadPubSub` of `/tests/conformance/common.go`
- Make any changes needed to the RabbitMQ component to make it pass

pkedy:
# Description

This PR adds RabbitMQ to the conformance tests.

I also made two small tweaks to the component itself:

* Using the Delivery API to either Ack or Nack a message
* Made consumerID only required when `Subscribe` is called so that an app can publish only

I will be adding retries support in a separate PR.

## Issue reference

Resolves #694

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/tests/config/pubsub/rabbitmq/pubsub.yml
+++ b/tests/config/pubsub/rabbitmq/pubsub.yml
@@ -1,0 +1,28 @@
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: pubsub
+spec:
+  type: pubsub.rabbitmq
+  version: v1
+  metadata:
+  - name: host
+    value: "amqp://localhost:5672"
+  - name: consumerID
+    value: "testConsumer"
+  - name: durable
+    value: "false"
+  - name: deletedWhenUnused
+    value: "false"
+  - name: autoAck
+    value: "false"
+  - name: deliveryMode
+    value: "0"
+  - name: requeueInFailure
+    value: "true"
+  - name: prefetchCount
+    value: "0"
+  - name: reconnectWait
+    value: "0"
+  - name: concurrencyMode
+    value: single
--- a/.github/infrastructure/docker-compose-rabbitmq.yml
+++ b/.github/infrastructure/docker-compose-rabbitmq.yml
@@ -1,0 +1,7 @@
+version: '2'
+services:
+  rabbitmq:
+    image: rabbitmq:3-management
+    ports:
+      - 5672:5672
+      - 15672:15672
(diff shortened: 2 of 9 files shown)


tests/config/pubsub/rabbitmq/pubsub.yml
```diff
@@ -0,0 +1,28 @@
...
+    value: "testConsumer"
+  - name: durable
+    value: "false"
+  - name: deletedWhenUnused
+    value: "false"
+  - name: autoAck
+    value: "false"
+  - name: deliveryMode
+    value: "0"
+  - name: requeueInFailure
```
mukundansundar, line 21: should this be `requeueOnFailure`? should the metadata field name be changed?

mukundansundar: @artursouza that would be a breaking change ... 

or the other scenario ... add an alias and at a later point when we are making v2 of the component phase the old one out

artursouza: I don't think that changing this is enough to justify a new version of the component. If we want to fix this now, we should add a new metadata `requestOnFailure` and use that too. If user provides both, it should be a fatal error. Then, when we actually get to a v2 of this component, we can remove the old one.

How does that sound?

artursouza (approved):
I approve this change. If we want to fix the metadata name, it should be in a new PR IMO. See my comment about *how* I think we should do it.

tests/config/pubsub/rabbitmq/pubsub.yml
mukundansundar: Exactly my thoughts on breaking change. I think an alias can be provided and yes I agree that a fatal error can be logged if both are provided.

mukundansundar (approved):
lgtm

tests/config/pubsub/rabbitmq/pubsub.yml
artursouza: OK. Let's do it as a separate PR. @pkedy Please, track this as a separate issue. I don't think is a priority for v1.1 to be honest.

PR #743, opened 2021-03-05
Add DNS name resolver. Add unit tests.

abogdanov37:
# Description

Add new name resolver based on DNS. It needs to implement custom deployment in Kubernetes. 

## Issue reference

dapr/dapr#2865

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] **Documentation update not needed.** Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_


nameresolution/dns/dns_test.go
```diff
@@ -0,0 +1,25 @@
...
+	"testing"
+
+	"github.com/dapr/components-contrib/nameresolution"
+	"github.com/dapr/dapr/pkg/logger"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestResolve(t *testing.T) {
+	resolver := NewResolver(logger.NewLogger("test"))
+	request := nameresolution.ResolveRequest{ID: "myid", Namespace: "abc", Port: 1234}
```
artursouza, line 18: So, does it mean namespace is ignored?

artursouza (changes requested):

nameresolution/dns/dns_test.go
abogdanov37: Yep. You are right.

artursouza: I should have been clear about my question. But why can't the namespace be part of the dns name? :)

abogdanov37: Its solution has two reasons
1. When we add a NAMESPACE environment variable Dapr decide that we run in Kubernetes mode and we must set control plane address to proceed
2. This mode (dns) universal in our vision and could be used with another orchestrator (Docker Swarm) or without an orchestrator on different hosts. Question about run Dapr with docker compose on several hosts was two days ago in the general channel in Discord.
Sorry for previous answer.

nameresolution/dns/dns.go
```diff
@@ -0,0 +1,33 @@
...
+func NewResolver(logger logger.Logger) nameresolution.Resolver {
+	return &resolver{logger: logger}
+}
+
+// Init initializes Kubernetes name resolver.
+func (k *resolver) Init(metadata nameresolution.Metadata) error {
+	return nil
+}
+
+// ResolveID resolves name to address in Kubernetes.
```
artursouza, line 29: Please, change the comments about Kubernetes.

nameresolution/dns/dns_test.go
artursouza: /cc @yaron2  Can you comment on this?

yaron2: @abogdanov37 namespaces are supported for Dapr in self-hosted mode as well. Dapr knows it runs in Kubernetes if the Kubernetes mode flag is specified for the `daprd` process.

abogdanov37: We had following problem. When we set namespace env variable sidecar with `daprd` can't get configuration from file. Because of that we cant proceed work

abogdanov37: How can I use namespace in self-hosted mode? I have no ideas(

msfussell: @abogdanov37 - Namespaces in self hosted mode are described here https://docs.dapr.io/operations/components/component-scopes/#namespaces. You set NAMEPSPACE env var.

abogdanov37: Thank you very much! It changes the World! Based on this I think I can add namespace support. I'll update PR a bit later.

abogdanov37: But after a bit of thinking, I see a not clear concept about namespaces in a standalone mode Dapr use namespace to separate components but not to finding services. Because of this, I don't want to use namespaces to find services. To find components I can set relevant Kubernetes service name as redis.storage.svc.cluster.local. All concept of my mode based on a single namespace.

artursouza: Dapr uses namespace to find services too - it allows cross namespace service invocation. Namespace is just a logical separation. Can that be helpful in your scenario?

yaron2:
@abogdanov37 are you making progress with this PR?

abogdanov37:
> @abogdanov37 are you making progress with this PR?

Yes. After discussion in PR https://github.com/dapr/docs/pull/1310 I think that we can close this PR and for DNS mode I can use Kubernetes mode. What do think about it?

abogdanov37:
Hi folks. I made changes two days ago. Please give me feedback. This PR blocks https://github.com/dapr/dapr/pull/2903.

artursouza (dismissed):

abogdanov37:
Hi folks! @artursouza thank you for approve. Who can merge PR?

AaronCrawfis:
> Hi folks! @artursouza thank you for approve. Who can merge PR?

@artursouza @yaron2

abogdanov37:
Hi folks! What can I do to force the merging of this PR?

artursouza:
@abogdanov37 We only force merge in some rare cases. We wait for build to pass before merging. We had a change in where logger package is imported from, so I fixed that in your PR. Let's see.

artursouza (approved):

abogdanov37:
@artursouza  Thank you very much.

artursouza:
@abogdanov37 Is the issue referred in this PR the correct one?

abogdanov37:
> @abogdanov37 Is the issue referred in this PR the correct one?

Yes

Issue #749, opened 2021-03-09
NameResolution: Consul


a-elsheikh:
This is a proposal to add a new nameresolution component for service discovery using consul

PR #750, opened 2021-03-09
Consul nameresolution component

a-elsheikh:
# Description

This PR will add a new name resolution component to dapr using Consul. This was driven by the need for an alternative to mdns for dapr service discovery in self-hosted mode. This will allow for dapr side cars to both register (optional) and resolve other dapr services by connecting to a Consul agent. It is fully configurable and shouldn't be too opinionated

## Issue reference

#749
dapr/docs#1408 (docs)

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


yaron2:
@a-elsheikh is this ready for review?

a-elsheikh:
> @a-elsheikh is this ready for review?

@yaron2 yes please, CLA reviewing it still pending but I should hopefully be able to sign soon. I've looked at this primarily in the context of self-hosted mode, it would be nice if someone could validate the behaviour in kubernetes

nameresolution/consul/consul.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+
+	var appID string
+	var appPort string
+	var host string
+	var httpPort string
+	var ok bool
+
+	if appID, ok = props[nr.AppID]; !ok {
+		return nil, fmt.Errorf("nr metadata property missing: %s", nr.AppID)
```
pkedy, line 238: `nr` in the error message may not make sense to the user. Could be `consul name resolution: ` which is the pattern I've seen in other components but there is not 100% consistency.

nameresolution/consul/consul.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+
+	if appPort, ok = props[nr.AppPort]; !ok {
+		return nil, fmt.Errorf("nr metadata property missing: %s", nr.AppPort)
```
pkedy, line 242: same

nameresolution/consul/consul.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+
+	if host, ok = props[nr.HostAddress]; !ok {
+		return nil, fmt.Errorf("nr metadata property missing: %s", nr.HostAddress)
```
pkedy, line 246: same

nameresolution/consul/consul.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+
+	if httpPort, ok = props[nr.DaprHTTPPort]; !ok {
+		return nil, fmt.Errorf("nr metadata property missing: %s", nr.DaprHTTPPort)
```
pkedy, line 250: same

nameresolution/consul/consul.go
```diff
@@ -0,0 +1,327 @@
...
+
+	if appPort, ok = props[nr.AppPort]; !ok {
+		return nil, fmt.Errorf("nr metadata property missing: %s", nr.AppPort)
+	}
+
+	if host, ok = props[nr.HostAddress]; !ok {
+		return nil, fmt.Errorf("nr metadata property missing: %s", nr.HostAddress)
+	}
+
+	if httpPort, ok = props[nr.DaprHTTPPort]; !ok {
```
pkedy, line 249: You might consider attempting to parse this as an unsigned integer for some level of validation.

nameresolution/consul/consul.go
```diff
@@ -0,0 +1,327 @@
...
+	return config, nil
+}
+
+// helper function for transforming interface{} into json serializable
+func convertGenericConfig(i interface{}) interface{} {
+	switch x := i.(type) {
+	case map[interface{}]interface{}:
+		m2 := map[string]interface{}{}
+		for k, v := range x {
+			m2[k.(string)] = convertGenericConfig(v)
```
pkedy, line 316: In the unlikely scenario that this is not a `string` it will panic.

nameresolution/consul/consul.go
```diff
@@ -0,0 +1,327 @@
...
+
+	data, err := json.Marshal(rawConfig)
+	if err != nil {
+		return config, fmt.Errorf("error serializing to json: %w", err)
+	}
+
+	decoder := json.NewDecoder(bytes.NewReader(data))
+	decoder.DisallowUnknownFields()
+
+	if err := decoder.Decode(&config); err != nil {
```
pkedy, line 303: It feels a little strange that this is being decoded into structs like `consul.QueryOptions` that have none fields like `ctx context.Context` which is not JSON friendly. My sense is to have separate structs the read in the config and then initialize the consul structs from there. WDYT? Is this JSON approach done elsewhere or in Consul itself?

nameresolution/consul/consul_test.go
```diff
@@ -0,0 +1,754 @@
...
+
+	nr "github.com/dapr/components-contrib/nameresolution"
+	"github.com/dapr/dapr/pkg/logger"
+	consul "github.com/hashicorp/consul/api"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+)
+
+type mockConsulResolver struct {
+	mock.Mock
```
pkedy, line 16: This is just my personal opinion. When possible I try to avoid using `github.com/stretchr/testify/mock` because _usually_ mocks can be accomplished more easily using Go interfaces. I created a Gist for you to see it applied in this file.

https://gist.github.com/pkedy/672d58762ccba9c2e100e9fa7ddf79a6

nameresolution/consul/consul.go
```diff
@@ -0,0 +1,327 @@
...
+		return config, fmt.Errorf("error deserializing to configSpec: %w", err)
+	}
+
+	return config, nil
+}
+
+// helper function for transforming interface{} into json serializable
+func convertGenericConfig(i interface{}) interface{} {
```
pkedy, line 311: Just curious: What in the contents of `i` is not serializable by `json.Marshal`?

pkedy (commented):
This is a first pass at making comments and asking questions. I still need to test this in K8s with Consul installed. @a-elsheikh Do you happen to have some quick instructions on how you tested this locally. Might be helpful to add a `README.md` in the directory for this component. I'd love to eventually see automated e2e tests for this.

nameresolution/consul/Readme.md
```diff
@@ -0,0 +1,102 @@
...
+        - "v1"
+        - "${OTHER_ENV_VARIABLE}"
+      meta:
+        DAPR_METRICS_PORT: "${DAPR_METRICS_PORT}"
+        DAPR_PROFILE_PORT: "${DAPR_PROFILE_PORT}"
+      daprPortMetaKey: "DAPR_PORT"        
+      queryOptions:
+        useCache: true
+        filter: "Checks.ServiceTags contains dapr"
+}
```
pkedy, line 70: Is this a typo?

a-elsheikh:
@pkedy thanks for picking this up, I'll begin to address your comments

nameresolution/consul/Readme.md
a-elsheikh: yes

nameresolution/consul/consul.go
a-elsheikh: I get what you mean by strange, this did feel like a shortcut. The question here is, is it worth adding a mapping layer of almost identical types to avoid the unfriendly fields? My initial concern with this was more, if we upgraded the consul library version in the future then changes in the library could potentially break user configuration, do you agree?

nameresolution/consul/consul.go
a-elsheikh: These errors are actually all raised in the context of the consul name resolution component, so perhaps we should remove nr all together and leave it as

`return nil, fmt.Errorf("metadata property missing: %s", nr.AppID)`

then in the log it will look as so like this

`msg="failed to initialize name resolution resolver consul: metadata property missing:...`

instead of this

`msg="failed to initialize name resolution resolver consul: consul name resolution: metadata property missing:...`

what do you think?

nameresolution/consul/consul.go
a-elsheikh: when the yaml is unmarshalled it yields nested maps with a key of type interface{} which is not supported by json

nameresolution/consul/consul.go
a-elsheikh: done

nameresolution/consul/consul.go
a-elsheikh: done

nameresolution/consul/consul.go
```diff
@@ -0,0 +1,327 @@
...
+func newConsulResolver(logger logger.Logger, consul consulResolverInterface, config resolverConfig) nr.Resolver {
+	return &resolver{
+		logger: logger,
+		consul: consul,
+		config: config,
+	}
+}
+
+type consulResolverInterface interface {
+	InitClient(config *consul.Config) error
```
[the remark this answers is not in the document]
a-elsheikh: that was also my preference but struggled with this one given the nested dependencies (Agent and Health). I've refactored the di to get closer to the real dependency, have a look and please suggest a better alternative if there is, golang is all new to me

nameresolution/consul/consul.go
a-elsheikh: done - made the change let me know if you think otherwise

nameresolution/consul/consul.go
a-elsheikh: done

nameresolution/consul/consul.go
a-elsheikh: done

a-elsheikh:
> This is a first pass at making comments and asking questions. I still need to test this in K8s with Consul installed. @a-elsheikh Do you happen to have some quick instructions on how you tested this locally. Might be helpful to add a `README.md` in the directory for this component. I'd love to eventually see automated e2e tests for this.

Regarding the readme, there is already one there. I've updated it in an attempt to simplify the instructions, let me know if it is lacking. Also, from what I understand the automated e2e tests are in the runtime repo no? I'm not sure if they run on an integration environment or are relying on packages that expose "test" servers, any guidance?

pkedy:
@a-elsheikh Sorry for the delay. I'll be taking another pass on this after our 1.1 release. You are correct in that the e2e tests are done in dapr/dapr. I should have comment about that in dapr/dapr#2904. Since this is a completely new nameresolution component, I'm not entirely sure how to incorporate it in the e2e test suite either but I'll take a look soon.

nameresolution/consul/consul.go
pkedy: > My initial concern with this was more, if we upgraded the consul library version in the future then changes in the library could potentially break user configuration, do you agree?

Yes! Yes good point! I'll take compile time errors over ambiguous runtime bugs any day :)

yaron2:
@pkedy can you review this again? we'd like to get this in for 1.2.

nameresolution/consul/consul.go
a-elsheikh: okay, this should be addressed in the latest commit

nameresolution/consul/consul.go
```diff
@@ -0,0 +1,273 @@
...
+		return fmt.Errorf("failed check on consul agent: %w", err)
+	}
+
+	return nil
+}
+
+// ResolveID resolves name to address via consul
+func (r *resolver) ResolveID(req nr.ResolveRequest) (string, error) {
+	cfg := r.config
+	services, _, err := r.client.Health().Service(req.ID, "", true, cfg.QueryOptions)
```
pkedy, line 111: What is the performance implication of calling this each time? Is the Consul API invoked on every request? I'm hoping there is some sort of caching involved. My years ago, I implemented service discovery w/ Consul using a Watch to keep everything in sync in memory. I'm not sure if that is a better approach but I at least wanted to understand how this would perform at scale.

pkedy (commented):
@a-elsheikh I've tested this locally with the distributed calendar quickstart and it all seems to work great. I put some logging statements in to make sure `ResolveID` was getting called and what it was returning. I noticed it was getting called on every invocation which led me to wonder if the Consul API was being invoked each time as well. See my comment. Thoughts?

nameresolution/consul/consul.go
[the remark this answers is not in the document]
pkedy: Looks like you made tweaks to the mocking approach to bring your unit test coverage over 80%.  Great job and thank you!

nameresolution/consul/consul.go
pkedy: Makes sense to me.  Please create an issue for tracking if you don't think you'll get to it as a fast follow to this PR.

a-elsheikh: sorry there might be a bit of a misunderstanding - regarding what you said above _"My sense is to have separate structs the read in the config and then initialize the consul structs from there."_ this is the approach I've ultimately taken, so configuration is no longer tightly coupled to the consul api structs

nameresolution/consul/consul.go
a-elsheikh: Yes, it will make a call to the local agent every time and that is where the caching sits. I've not done any benchmarking but a simple loop with 10k requests gives me an average of 1ms per round trip. No doubt, if we implement the watcher then that latency should shrink but I'll need a bit more time to get to that

yaron2: I think it's fine to leave this as is now and iterate on that with caching later on.

pkedy: I agree. Since it's to the local agent it's not that bad.

nameresolution/consul/Readme.md
```diff
@@ -0,0 +1,128 @@
...
+The consul name resolution component gives the ability to register and resolve other "daprized" services registered on a consul estate. It is flexible in that it allows for complex to minimal configurations driving the behaviour on init and resolution.
+
+## How To Use
+
+### Dapr Registration
+
+If you are using the dapr sidecar to register your service to consul then you will need the following configuration:
+
+```yaml
+spec:
```
pkedy, line 12: Having tested this on Kubernetes, I think we should have the following at the top of the configuration.

```yaml
apiVersion: dapr.io/v1alpha1
kind: Configuration
metadata:
  name: appconfig
```

Likewise, if the application deployments need to refer to the configuration name per https://docs.dapr.io/operations/configuration/configuration-overview/#kubernetes-sidecar

I also think an issues needs to be created in dapr/docs to create content about name resolution and this component.

nameresolution/consul/Readme.md
```diff
@@ -0,0 +1,128 @@
...
+    configuration:
+      selfRegister: true
+```
+
+### External Registration
+
+If consul service registration is managed externally from dapr you need to ensure that the dapr-to-dapr internal grpc port is added to the service metadata under `DAPR_PORT` (this key is configurable) and that the consul service Id matches the dapr app Id. The config should then look as follows:
+
+```yaml
+spec:
```
pkedy, line 24: See comment above

nameresolution/consul/Readme.md
```diff
@@ -0,0 +1,128 @@
...
+| AdvancedRegistration | [*api.AgentServiceRegistration](https://pkg.go.dev/github.com/hashicorp/consul/api@v1.3.0#AgentServiceRegistration) | Gives full control of service registration through configuration. If configured the component will ignore any configuration of Checks, Tags, Meta and SelfRegister. |
+
+## Samples Configurations
+
+### Basic
+
+The minimum configuration needed is the following:
+
+```yaml
+spec:
```
pkedy, line 59: See comment above

nameresolution/consul/Readme.md
```diff
@@ -0,0 +1,128 @@
...
+    component: "consul"
+}
+```
+
+### Registration with additional customizations
+
+Enabling `SelfRegister` it is then possible to customize the checks, tags and meta
+
+```yaml
+spec:
```
pkedy, line 70: See comment above

nameresolution/consul/Readme.md
```diff
@@ -0,0 +1,128 @@
...
+        useCache: true
+        filter: "Checks.ServiceTags contains dapr"
+```
+
+### Advanced registration
+
+Configuring the advanced registration gives you full control over all the properties possible when registering.
+
+```yaml
+spec:
```
pkedy, line 104: See comment above

pkedy:
@a-elsheikh These changes LGTM but I'm waiting to merge until the patch is applied to your branch in dapr/dapr. Once that happens then I think we can move quickly to merge both PRs (of course after updating the commit for dapr/components-contrib in dapr/dapr's `go.mod`).

nameresolution/consul/Readme.md
a-elsheikh: sounds good, I didn't add it originally because I figured it was implied, in any case that's there now and I've added the docs issue

nameresolution/consul/consul.go
a-elsheikh: @pkedy @yaron2 just an fyi I'm currently brewing this one and it's in testing, I'll PR it separately once it's ready

a-elsheikh:
@pkedy @yaron2 seems like merging master has failed the checks - could it be a flaky conformance test or some transient issue? pubsub.mqtt-vernemq is the one in question

yaron2 (approved):

Issue #752, opened 2021-03-10
Local Storage Binding


lynn-orrell:
A binding that allows reading / writing from the local file system would be nice, especially in deployments in which cloud based storage isn't available.

mukundansundar:
@artursouza  closed by #753. Leaving it open till it is added to dapr/dapr (Endgame process).

PR #753, opened 2021-03-10
Localstorage

lynn-orrell:
# Description

Introduced a localstorage binding that allows you to read/write from the local file system. Currently this only works in self-hosted mode as there _currently_ isn't a way to mount a volume in the dapr sidecar.

## Issue reference

#752 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/bindings/localstorage/localstorage.go
+++ b/bindings/localstorage/localstorage.go
@@ -15,7 +15,6 @@ import (
 	"path/filepath"
 	"strconv"
 
-	"github.com/Azure/azure-storage-blob-go/azblob"
 	securejoin "github.com/cyphar/filepath-securejoin"
 	"github.com/dapr/components-contrib/bindings"
 	"github.com/dapr/dapr/pkg/logger"
@@ -28,10 +27,8 @@ const (
 
 // LocalStorage allows saving files to disk
 type LocalStorage struct {
-	metadata     *Metadata
-	containerURL azblob.ContainerURL
-
-	logger logger.Logger
+	metadata *Metadata
+	logger   logger.Logger
 }
 
 // Metadata defines the metadata
--- a/bindings/localstorage/localstorage.go
+++ b/bindings/localstorage/localstorage.go
@@ -22,7 +22,7 @@ import (
 )
 
 const (
-	fileName = "fileName"
+	fileNameMetadataKey = "fileName"
 )
 
 // LocalStorage allows saving files to disk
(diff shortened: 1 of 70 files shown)


lynn-orrell:
@mukundansundar, can you have a look at this failing conformance test? Not sure what's going on here... Thanks!

artursouza:
@lynn-orrell Mounting a volume is outside the scope of Dapr. But, there is nothing stopping it from being mounted by the app operator. So, I think this can still be used in K8s.

bindings/localstorage/localstorage.go
```diff
@@ -0,0 +1,240 @@
...
+	"strconv"
+
+	securejoin "github.com/cyphar/filepath-securejoin"
+	"github.com/dapr/components-contrib/bindings"
+	"github.com/dapr/dapr/pkg/logger"
+	"github.com/google/uuid"
+)
+
+const (
+	fileName = "fileName"
```
artursouza, line 25: This name is very similar to a parameter. Can you rename this constant to something like `filenameMetadataKey`?

bindings/localstorage/localstorage.go
```diff
@@ -0,0 +1,240 @@
...
+
+// Init performs metadata parsing
+func (ls *LocalStorage) Init(metadata bindings.Metadata) error {
+	m, err := ls.parseMetadata(metadata)
+	if err != nil {
+		return err
+	}
+	ls.metadata = m
+
+	err = os.MkdirAll(ls.metadata.RootPath, 0666)
```
artursouza, line 52: Check if it already exists first. Also, why 0666 and not simply 0600?

bindings/localstorage/localstorage.go
```diff
@@ -0,0 +1,240 @@
...
+	if err == nil {
+		req.Data = decoded
+	}
+
+	absPath, _, err := getSecureAbsRelPath(ls.metadata.RootPath, filename)
+	if err != nil {
+		return nil, err
+	}
+
+	err = os.MkdirAll(filepath.Dir(absPath), 0666)
```
artursouza, line 102: Can filemask come from metadata too with a 0600 as default?

artursouza (changes requested):
Please, add to conformance tests.

bindings/localstorage/localstorage.go
lynn-orrell: Done.

bindings/localstorage/localstorage.go
lynn-orrell: After further investigation, I've set the perms to 0777 by default since the umask will be applied.

bindings/localstorage/localstorage.go
lynn-orrell: After further investigation, I've set the perms to 0777 by default since the umask will be applied.

artursouza (approved):

mukundansundar:
@lynn-orrell Can you add docs for this Component ?

PR #757, opened 2021-03-14
add Ping operation to`store` interface.

moonchanyong:
# Description

I have added `Ping` to the `Store` interface for health check endpoint.

## Issue reference

dapr/dapr#2167

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/state/azure/cosmosdb/cosmosdb.go
+++ b/state/azure/cosmosdb/cosmosdb.go
@@ -359,6 +359,10 @@ func (c *StateStore) Multi(request *state.TransactionalStateRequest) error {
 	return nil
 }
 
+func (c *StateStore) Ping() error {
+	return nil
+}
+
 func createUpsertItem(contentType string, req state.SetRequest, partitionKey string) CosmosItem {
 	byteArray, isBinary := req.Value.([]uint8)
 	if isBinary {
--- a/state/azure/cosmosdb/cosmosdb.go
+++ b/state/azure/cosmosdb/cosmosdb.go
@@ -28,6 +28,7 @@ type StateStore struct {
 	collection  *documentdb.Collection
 	db          *documentdb.Database
 	sp          *documentdb.Sproc
+	metadata    metadata
 	contentType string
 
 	features []state.Feature
@@ -139,6 +140,7 @@ func (c *StateStore) Init(meta state.Metadata) error {
 		return fmt.Errorf("collection %s for CosmosDB state store not found.  This must be created before Dapr uses it", m.Collection)
 	}
 
+	c.metadata = m
 	c.collection = &colls[0]
 	c.client = client
 	c.contentType = m.ContentType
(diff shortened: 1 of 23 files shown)


moonchanyong:
@yaron2  @artursouza I have added Ping operation for adding health check endpoint(dapr/dapr#2167). PTAL.

state/redis/redis.go
```diff
@@ -127,6 +127,10 @@ func parseRedisMetadata(meta state.Metadata) (metadata, error) {
 	return m, nil
 }
 
+func (r *StateStore) Ping() error {
+	return errors.New("not implemented")
```
artursouza, line 131: Default implementation should return nil. I think it would make it easier to adopt this feature. Otherwise, daprd would report unhealthy by default.

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -335,6 +336,10 @@ func (c *StateStore) Multi(request *state.TransactionalStateRequest) error {
 	return nil
 }
 
+func (c *StateStore) Ping() error {
+	return errors.New("not implemented")
```
artursouza, line 340: Probably, state stores that are in GA should have this feature implemented and added to conformance tests.

artursouza (changes requested):

state/azure/cosmosdb/cosmosdb.go
artursouza: I mean, to also add "ping()" validation to conformance tests.

state/redis/redis.go
mukundansundar: @artursouza Should this be a `Feature` ? If the Feature is present call Ping()? Instead of default nil implementation.

artursouza: I think returning nil by default is fine. We can make this a feature once there is specific behavior in a method that can impact decisions above. I don't think this is the case (yet).

moonchanyong: I have changed the default implementation of `Ping` operation to return nil.

state/azure/cosmosdb/cosmosdb.go
moonchanyong: @artursouza 
I try to add a conformance test. PTAL.

artursouza:
@yaron2  Are you OK with this feature? If so, the next step for this PR to be merged would be to implement the feature in every GA state store.

yaron2:
I'm good with this, it's important functionality that can later be expanded to other components.

moonchanyong:
@artursouza 

is there any doing for the next step?

artursouza:
@moonchanyong  For this PR, the next step is to implement Ping() for each of the state stores in GA as per https://docs.dapr.io/operations/components/setup-state-store/supported-state-stores/

After this PR is merged, the next step is to repeat the same for the other building blocks.

Thanks a lot for working on this. This is very appreciated.

moonchanyong:
@artursouza  
I have added implementations of `Ping` for each GA component.
but not added yet at blobstorage.
does have any idea to test the connection of blob storage?

yaron2:
@moonchanyong please resolve the conflict. Once done, we can merge this.

artursouza:
@moonchanyong What about calling `GetProperties()`?

moonchanyong:
sorry for the late
@yaron2  @artursouza 
Can you approve to run workflow??

artursouza:
@moonchanyong Please, add Ping() to the conformance tests.

moonchanyong:
@artursouza 

I have added conformance test at `tests/conformance/state/state.go`. 
Does it need another test ??

artursouza:
> @artursouza
> 
> I have added conformance test at `tests/conformance/state/state.go`.
> Does it need another test?


Ah. Thanks. I missed that. LGTM.

artursouza:
@moonchanyong Please, fix the lint errors. We are almost there.

moonchanyong:
@artursouza 

can you rerun the workflow for the test?

artursouza (approved):

artursouza:
Next, is to integrate this to the runtime.

PR #760, opened 2021-03-15
refactor GCP pubsub

trondhindenes:
# Description
- Make GCP pubsub support implicit credentials, in addition to explicit
- Clean up GCP pubsub metadata so that it lines up with other dapr pubsubs (separate struct for the gcp auth json stuff, it seemed a bit cleaner)
- make project id (project where subscription/topic should live) separate from "identity project id" (where the service account lives)

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#1308

### DIFF
--- a/pubsub/gcp/pubsub/metadata.go
+++ b/pubsub/gcp/pubsub/metadata.go
@@ -1,7 +1,5 @@
 package pubsub
 
-
-
 // GCPPubSubMetaData pubsub metadata
 type metadata struct {
 	consumerID              string
@@ -18,5 +16,3 @@ type metadata struct {
 	AuthProviderCertURL     string
 	ClientCertURL           string
 }
-
-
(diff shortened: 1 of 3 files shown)


yaron2:
```Go
var pubsubClient *gcppubsub.Client
var err error
```

trondhindenes:
thanks @yaron2  - i keep struggling do initalize the "defaults" with the correct type. Appreciate the help!

I think this PR should be in a semi-good state now - at least good enough to discuss if the direction is correct

yaron2:
There's a code lint issue

trondhindenes:
I'm not sure what the "misses" and "partials" are about - I added more test coverage but maybe that didn't help?

yaron2 (approved):
LGTM

Did you get to test this by any chance? just to make sure there are no regressions.

trondhindenes:
Thanks! Yep, we're using a test build internally already. Feeling fairly safe that we're good, but since we're changing the parameters there to be more "dapr-friendly" there _are_ breaking changes. Feel free to merge if you're okay with that, or let me know if we should discuss further.

PR #763, opened 2021-03-17
add some redis client config that may be commonly used

Taction:
# Description

add some config that can be set from metadata for redis client.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#1334

RELEASE NOTES: Added commonly used Redis client settings

### DIFF
--- a/pubsub/redis/metadata.go
+++ b/pubsub/redis/metadata.go
@@ -14,6 +14,17 @@ type metadata struct {
 	host string
 	// The Redis password
 	password string
+	// Dial timeout for establishing new connections.
+	dialTimeoutInSec int
+	// Timeout for socket reads. If reached, commands will fail
+	// with a timeout instead of blocking. Use value -1 for no timeout and 0 for default.
+	readTimeoutInSec int
+	// Timeout for socket writes. If reached, commands will fail
+	writeTimeoutInSec int
+	// Maximum number of socket connections.
+	poolSize int
+	// Connection age at which client retires (closes) the connection.
+	maxConnAgeInSec int
 	// The consumer identifier
 	consumerID string
 	// A flag to enables TLS by setting InsecureSkipVerify to true
--- a/pubsub/redis/metadata.go
+++ b/pubsub/redis/metadata.go
@@ -19,13 +19,13 @@ type metadata struct {
 
 	// Maximum number of retries before giving up.
 	// Default is to not retry failed commands.
-	maxRetries int
+	backoffMaxRetries int
 	// Minimum backoff between each retry.
 	// Default is 8 milliseconds; -1 disables backoff.
-	minRetryBackoff time.Duration
+	backOffMinInterval time.Duration
 	// Maximum backoff between each retry.
 	// Default is 512 milliseconds; -1 disables backoff.
-	maxRetryBackoff time.Duration
+	backOffMaxInterval time.Duration
 
 	// Dial timeout for establishing new connections.
 	dialTimeout time.Duration
--- a/pubsub/redis/metadata.go
+++ b/pubsub/redis/metadata.go
@@ -14,6 +14,19 @@ type metadata struct {
 	host string
 	// The Redis password
 	password string
+	// Database to be selected after connecting to the server.
+	db int
+
+	// Maximum number of retries before giving up.
+	// Default is to not retry failed commands.
+	maxRetries int
+	// Minimum backoff between each retry.
+	// Default is 8 milliseconds; -1 disables backoff.
+	minRetryBackoff time.Duration
+	// Maximum backoff between each retry.
+	// Default is 512 milliseconds; -1 disables backoff.
+	maxRetryBackoff time.Duration
+
 	// Dial timeout for establishing new connections.
 	dialTimeout time.Duration
 	// Timeout for socket reads. If reached, commands will fail
@@ -23,8 +36,26 @@ type metadata struct {
 	writeTimeout time.Duration
 	// Maximum number of socket connections.
 	poolSize int
+	// Minimum number of idle connections which is useful when establishing
+	// new connection is slow.
+	minIdleConns int
 	// Connection age at which client retires (closes) the connection.
+	// Default is to not close aged connections.
 	maxConnAge time.Duration
+	// Amount of time client waits for connection if all connections
+	// are busy before returning an error.
+	// Default is ReadTimeout + 1 second.
+	poolTimeout time.Duration
+	// Amount of time after which client closes idle connections.
+	// Should be less than server's timeout.
+	// Default is 5 minutes. -1 disables idle timeout check.
+	idleTimeout time.Duration
+	// Frequency of idle checks made by idle connections reaper.
+	// Default is 1 minute. -1 disables idle connections reaper,
+	// but idle connections are still discarded by the client
+	// if IdleTimeout is set.
+	idleCheckFrequency time.Duration
+
 	// The consumer identifier
 	consumerID string
 	// A flag to enables TLS by setting InsecureSkipVerify to true
--- a/pubsub/redis/metadata.go
+++ b/pubsub/redis/metadata.go
@@ -15,16 +15,16 @@ type metadata struct {
 	// The Redis password
 	password string
 	// Dial timeout for establishing new connections.
-	dialTimeoutInSec int
+	dialTimeout time.Duration
 	// Timeout for socket reads. If reached, commands will fail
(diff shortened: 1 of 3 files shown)


pubsub/redis/metadata.go
```diff
@@ -14,6 +14,17 @@ type metadata struct {
 	host string
 	// The Redis password
 	password string
+	// Dial timeout for establishing new connections.
+	dialTimeoutInSec int
```
pkedy, line 18: For consistency, this should be parsed as a `time.Duration` like `redeliverInterval` and `processingTimeout` are and the `InSec` suffix removed.

pubsub/redis/metadata.go
```diff
@@ -14,6 +14,17 @@ type metadata struct {
...
+	// Timeout for socket reads. If reached, commands will fail
+	// with a timeout instead of blocking. Use value -1 for no timeout and 0 for default.
+	readTimeoutInSec int
```
pkedy, line 21: Same

pubsub/redis/metadata.go
```diff
@@ -14,6 +14,17 @@ type metadata struct {
...
+	// Timeout for socket writes. If reached, commands will fail
+	writeTimeoutInSec int
```
pkedy, line 23: Same

pubsub/redis/metadata.go
```diff
@@ -14,6 +14,17 @@ type metadata struct {
...
+	// Maximum number of socket connections.
+	poolSize int
+	// Connection age at which client retires (closes) the connection.
+	maxConnAgeInSec int
```
pkedy, line 27: Same

pubsub/redis/redis.go
```diff
@@ -129,6 +134,46 @@ func parseRedisMetadata(meta pubsub.Metadata) (metadata, error) {
 		m.concurrency = uint(concurrency)
 	}
 
+	if val, ok := meta.Properties[dialTimeoutInSec]; ok && val != "" {
```
pkedy, line 137: Note how `redeliverInterval` and `processingTimeout` are parsed. First as uint milliseconds (e.g. `5000`), then as a duration (e.g. `5s`).

pkedy (changes requested):
I love seeing more connection settings coming in this PR! I'd like just a few changes for consistency with the duration-based settings.

pubsub/redis/metadata.go
Taction: I will add more  connection settings, and keep their format identical to definition in redis client config as much as possible.

pubsub/redis/redis.go
Taction: Thanks for advice, I will Parse duration-based settings like how redeliverInterval and processingTimeout are parsed.

pubsub/redis/redis.go
```diff
@@ -20,14 +20,27 @@ import (
...
-	consumerID        = "consumerID"
-	enableTLS         = "enableTLS"
-	processingTimeout = "processingTimeout"
-	redeliverInterval = "redeliverInterval"
-	queueDepth        = "queueDepth"
-	concurrency       = "concurrency"
+	host               = "redisHost"
+	password           = "redisPassword"
+	db                 = "redisDB"
+	maxRetries         = "maxRetries"
```
pkedy, line 26: My apologies, I didn't catch this before. In other components this field is called `backoffMaxRetries`, mainly because the [github.com/cenkalti/backoff](https://github.com/cenkalti/backoff) library is used.
Please rename this to `backoffMaxRetries`. It's just a consistency thing across components even though the Redis setting is differently named.

pubsub/redis/redis.go
```diff
@@ -20,14 +20,27 @@ import (
...
+	minRetryBackoff    = "minRetryBackoff"
```
pkedy, line 27: This one would be `backOffMinInterval`.

pubsub/redis/redis.go
```diff
@@ -20,14 +20,27 @@ import (
...
+	maxRetryBackoff    = "maxRetryBackoff"
```
pkedy, line 28: This one would be `backOffMaxInterval`.

pkedy (changes requested):

pkedy:
@Taction Thanks for all your work on this. I know there have been a few iterations now. Two things:

1) Before this can be merged, there should be an open PR in dapr/docs to update [this page](https://docs.dapr.io/operations/components/setup-pubsub/supported-pubsub/setup-redis-pubsub/) with the new settings. **Make sure branch from and target the `v1.1` branch in the PR. Its easy to forget.**

2) I think I have steered you in the wrong direction with the settings naming but wanted to ask your opinion.

*Context*: In `v1.2` we'll likely be adding more backoff settings to some of the pubsub components. The backoff/retry library we are using handles retries between the sidecar and the application.

*Where I steered you wrong*: All the settings I had you prefix with `backOff` are actually between the sidecar and Redis itself. Correct? If that is the case a different prefix should be used. Maybe `read`? I'm not actually sure where exactly these settings come into play in the client module but I'm assuming its for any CMD call. WDYT?

Taction:
> @Taction Thanks for all your work on this. I know there have been a few iterations now. Two things:
> 
> 1. Before this can be merged, there should be an open PR in dapr/docs to update [this page](https://docs.dapr.io/operations/components/setup-pubsub/supported-pubsub/setup-redis-pubsub/) with the new settings. **Make sure branch from and target the `v1.1` branch in the PR. Its easy to forget.**
> 2. I think I have steered you in the wrong direction with the settings naming but wanted to ask your opinion.
> 
> _Context_: In `v1.2` we'll likely be adding more backoff settings to some of the pubsub components. The backoff/retry library we are using handles retries between the sidecar and the application.
> 
> _Where I steered you wrong_: All the settings I had you prefix with `backOff` are actually between the sidecar and Redis itself. Correct? If that is the case a different prefix should be used. Maybe `read`? I'm not actually sure where exactly these settings come into play in the client module but I'm assuming its for any CMD call. WDYT?

1. For the second thing, I know what you actually want to tell me. I've seen the use of backoff options in mqtt component which is used for handle retries between the sidecar and the application.But there are lots components do not have these options so  I didn't catch the point at that time. And I should have confirmed with you about it.

As for the prefix, is `redis` a better choice（such as `redisMaxRetries`）? As far as I'm concerned it may have 2 potential advantage. One is that it may be more easily for users to tell these options apart from `backOff` options, as the name may shows that it is for redis configuration. Two is that `read` is a more general expression, if some other components have similar option between client and server `read` may be a better choice(I am not very familiar with other components now, so i assume that there is not much components whit these option).

2. For the first thing,I will work on this but it will take me some time to finish it. And It is very kind of you to stress out what is easy to forget.

What's more, in my opinion both `read` and `redis` is fine for the prefix. Which one do you think is more suitable?

pkedy:
@Taction I think we are in agreement! `redis` or `read` is fine as a prefix. As the author of this PR, I'll let you decide between the two. :) I just wanted to avoid `backOff` in this case because its for a different point in the call chain. Again-my mistake for steering you that way to begin with and thanks for your understanding. Once you make that change, we can finally get this merged unless @artursouza has other comments.

artursouza:
> @Taction
> 
> For 1, the `backOff*` settings are slowly being added to the components starting with the ones that really really need it (e.g. Kafka). Also, `backOffMaxRetries` is only setting for now but [more settings are one their way maybe in 1.2](https://github.com/dapr/components-contrib/compare/master...pkedy:using_config_package_part_1).
> 
> For 2, I think we are in agreement! `redis` or `read` is fine as a prefix. As the author of this PR, I'll let you decide between the two. :) I just wanted to avoid `backOff` in this case because its for a different point in the call chain. Again-my mistake for steering you that way to begin with and thanks for your understanding. Once you make that change, we can finally get this merged unless @artursouza has other comments.
> 
> Also, please mention the docs PR in this PR or visa versa. Thanks!


LGTM

pkedy:
@Taction To @mukundansundar 's point in your docs PR, we need an issue created as this is a pretty good sized enhancement. We do this for tracking things int the RELEASE NOTES.

pkedy:
@Taction I added RELEASE NOTES to the description of this PR which our release process also picks up. So in the interest of moving this along, no need to create an issue. Just know for the future that large changes need an issue.

pkedy (approved):

PR #764, opened 2021-03-17
Add app response to input binding event

yaron2:
This PR adds the ability for an app to response with a payload in response to an input binding event.

Part of https://github.com/dapr/dapr/issues/2628

### DIFF
--- a/bindings/Readme.md
+++ b/bindings/Readme.md
@@ -18,7 +18,7 @@ Input binding:
 ```go
 type InputBinding interface {
 	Init(metadata Metadata) error
-	Read(handler func(*ReadResponse) error) error
+	Read(handler func(*ReadResponse) ([]byte, error) error
 }
 ```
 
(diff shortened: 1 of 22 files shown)


pkedy (approved):
LGTM

Was surprised there are not existing input bindings that would return a payload.

Dare I ask if the `[]byte` should have a content type or metadata and be encapsulated in a `ReadResponseResponse` struct...  No. No. I won't ask that. :)

yaron2:
> LGTM
> 
> Was surprised there are not existing input bindings that would return a payload.
> 
> Dare I ask if the `[]byte` should have a content type or metadata and be encapsulated in a `ReadResponseResponse` struct... No. No. I won't ask that. :)

Since each input binding can be prescriptive about the type and format and data it expects, we can keep this as a byte array and let each binding validate the format. In the future we can change it if we feel the need.

PR #769, opened 2021-03-19
fix pubsub redis close bug

Taction:
# Description

Redis component close function called  `r.client.Close()` .
The pollNewMessagesLoop loop run in a goroutine use client to read from redis which will lead to an error: `error retrieving pending Redis messages: redis: client is closed`. 
When err occured this loop will continue. The context closed check will never have chance to run.
```
if err != nil {
	r.logger.Errorf("redis streams: error reading from stream %s: %s", stream, err)
	continue
}
// below will have no chance to run if client is closed
```
We need move the check to the top of this loop.

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


pkedy (approved):
Yep!  Totally makes sense!

PR #771, opened 2021-03-22
feat: change binding mqtt options to the same as pubsub

Hugome:
# Description

This PR improve the MQTT binding to use the same options/code as the MQTT pubsub (TLS, qos, ...)

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Extended the documentation / dapr/docs#1328

### DIFF
--- a/bindings/mqtt/mqtt.go
+++ b/bindings/mqtt/mqtt.go
@@ -298,7 +298,7 @@ func (m *MQTT) createClientOptions(uri *url.URL, clientID string) *mqtt.ClientOp
 	switch scheme {
 	case "mqtt":
 		scheme = "tcp"
-	case "mqtts":
+	case "mqtts", "tcps":
 		scheme = "ssl"
 	}
 	opts.AddBroker(scheme + "://" + uri.Host)
--- a/bindings/mqtt/mqtt.go
+++ b/bindings/mqtt/mqtt.go
@@ -6,33 +6,58 @@
 package mqtt
 
 import (
-	"encoding/json"
-	"errors"
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/pem"
 	"fmt"
 	"net/url"
 	"os"
 	"os/signal"
+	"strconv"
 	"syscall"
 	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/dapr/components-contrib/bindings"
+	"github.com/dapr/components-contrib/pubsub"
 	"github.com/dapr/dapr/pkg/logger"
 	mqtt "github.com/eclipse/paho.mqtt.golang"
-	"github.com/google/uuid"
+)
+
+const (
+	// Keys
+	mqttURL               = "url"
+	mqttTopic             = "topic"
+	mqttQOS               = "qos"
+	mqttRetain            = "retain"
+	mqttClientID          = "consumerID"
+	mqttCleanSession      = "cleanSession"
+	mqttCACert            = "caCert"
+	mqttClientCert        = "clientCert"
+	mqttClientKey         = "clientKey"
+	mqttBackOffMaxRetries = "backOffMaxRetries"
+
+	// errors
+	errorMsgPrefix = "mqtt binding error:"
+
+	// Defaults
+	defaultQOS          = 0
+	defaultRetain       = false
+	defaultWait         = 3 * time.Second
+	defaultCleanSession = true
 )
 
 // MQTT allows sending and receiving data to/from an MQTT broker
 type MQTT struct {
-	metadata *mqttMetadata
-	client   mqtt.Client
+	producer mqtt.Client
+	consumer mqtt.Client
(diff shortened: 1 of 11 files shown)


pkedy:
Just a heads up @Hugome - I'm going to try to add the config for conformance tests to your branch.

bindings/mqtt/mqtt.go
```diff
@@ -6,114 +6,258 @@
...
+		return nil, fmt.Errorf("mqtt error from publish: %v", token.Error())
+	}
 
 	return nil, nil
 }
 
 func (m *MQTT) Read(handler func(*bindings.ReadResponse) ([]byte, error)) error {
-	c := make(chan os.Signal, 1)
-	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
+	sigterm := make(chan os.Signal, 1)
```
pkedy, line 199: I can see why you did this. The binding interfaces to not contain a `Close() error` method or embed `io.Closer`. I'm just making a comment here for tracking purposes. I opened issue #779 to make this cleaner.

pkedy (dismissed):

pkedy:
@mukundansundar @artursouza pls do a quick review as I added code for the conformance test.

bindings/mqtt/mqtt.go
```diff
@@ -122,13 +266,39 @@ func (m *MQTT) connect(clientID string, uri *url.URL) (mqtt.Client, error) {
...
+
+	return tlsConfig
+}
+
+func (m *MQTT) createClientOptions(uri *url.URL, clientID string) *mqtt.ClientOptions {
 	opts := mqtt.NewClientOptions()
-	opts.AddBroker(fmt.Sprintf("tcp://%s", uri.Host))
+	opts.SetClientID(clientID)
+	opts.SetCleanSession(m.metadata.cleanSession)
+	opts.AddBroker(uri.Scheme + "://" + uri.Host)
```
pkedy, line 296: @Hugome is this a breaking change? The current docs/old code have the schema as `mqtt[s]` and no matter what it is forced to `tcp`. Would it make sense to replace `mqtts` with `tcps` and `mqtt` with `tcp`? If we do something like that we can preserve backward compatability. WDYT?

pkedy: Perhaps based on this it should be `ssl` instead of `tcps` and we have it wrong in the pubsub version...  Not sure.

https://pkg.go.dev/github.com/eclipse/paho.mqtt.golang#ClientOptions.AddBroker

Hugome: Oh yes you're right this is weird.
Maybe just the pubsub docs has issue ? (+1 for `tcp://` and `tcps://`)

Hugome: (Updated on the docs for `ssl://` https://github.com/dapr/docs/pull/1328/commits/cbc74878fa85e3b4e15aa82fe9f65e71efec1da3)

bindings/mqtt/mqtt.go
```diff
@@ -6,114 +6,258 @@
...
+	// mqtt broker allows only one connection at a given time from a clientID.
+	consumerClientID := fmt.Sprintf("%s-consumer", m.metadata.clientID)
+	c, err := m.connect(consumerClientID)
+	if err != nil {
+		return err
+	}
+	m.consumer = c
+
+	m.logger.Debugf("mqtt subscribing to topic %s", m.metadata.topic)
+	token := m.consumer.Subscribe(m.metadata.topic, m.metadata.qos, func(client mqtt.Client, mqttMsg mqtt.Message) {
```
artursouza, line 218: Is this coming from the pubsub implementation?

bindings/mqtt/mqtt.go
artursouza: So, is it a breaking change or not?

artursouza:
See my comments above.

bindings/mqtt/mqtt.go
pkedy: Currently yes.  What I think we (or @Hugome) needs to add is a schema coalesce to convert the previous schemes to what the MQTT client is expecting (ex. `mqtt://` -> `tcp://`, `mqtts` -> `ssl://`, etc.)

pkedy: To move this along, I added the following code to coalesce the scheme value
```golang
// URL scheme backward compatibility
scheme := uri.Scheme
switch scheme {
case "mqtt":
	scheme = "tcp"
case "mqtts", "tcps":
	scheme = "ssl"
}
opts.AddBroker(scheme + "://" + uri.Host)
```

artursouza (approved):

bindings/mqtt/mqtt.go
Hugome: Yes it's coming from the MQTT pubsub component

pkedy (approved):

Issue #792, opened 2021-04-01
customize cluster domain of k8s nameresolution


zhongfox:
kubernetes cluster domain is `cluster.local`, but it's configurable, currently we hard code in kubernetes nameresolution.
we'd better expose this option to user.

PR #793, opened 2021-04-01
add option to customize cluster domain of kubernetes nameresolution

zhongfox:
# Description

kubernetes cluster domain is cluster.local, but it's configurable, currently we hard code in kubernetes nameresolution.
we'd better expose this option to user.

## Issue reference

https://github.com/dapr/components-contrib/issues/792

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


pkedy (dismissed):

pkedy:
@zhongfox This looks good to me.  Do you have a PR in dapr/dapr that passes the Data element you added?

https://github.com/dapr/dapr/blob/master/pkg/messaging/direct_messaging.go#L226

pkedy:
@zhongfox You also need to open an issue to add this setting to dapr/docs.

zhongfox:
hi @pkedy, thanks for the comment.
after thinking twice， I prefer using `Metadata.Configuration`  of resolver to control the domain, rather than use data in each `resolveReuqest`, I think the domain is resolver level configuration.

since the `Metadata.Configuration`  is an interface, so the dapr/dapr runtime need no change.

please check the code, if its ok, I will add some doc to dapr/docs.

nameresolution/kubernetes/kubernetes.go
```diff
@@ -28,12 +29,19 @@ func NewResolver(logger logger.Logger) nameresolution.Resolver {
 
 // Init initializes Kubernetes name resolver.
 func (k *resolver) Init(metadata nameresolution.Metadata) error {
+	if config, ok := metadata.Configuration.(map[string]string); ok {
+		clusterDomain := config[ClusterDomainKey]
```
pkedy, line 33: If you set `k.clusterDomain = DefaultClusterDomain` here, then you don'y need the `if clusterDomain == "" {` in `ResolveID` and makes the code a little easier to follow (not that this change is complicated to begin with)

zhongfox: good idea, why I want check `if clusterDomain == ""` in `ResolveID` again, is because I'm afraid user may omit call `resolver.Init`, like the previous test case  https://github.com/dapr/components-contrib/blob/b404517388bd4b036f693b96dd73d1b0a1aff642/nameresolution/kubernetes/kubernetes_test.go#L16

this is what I did now:
1.  `k.clusterDomain = DefaultClusterDomain`  when NewResolver()
2.  remove `if clusterDomain == "" {` in `ResolveID`

please check again: https://github.com/dapr/components-contrib/pull/793/commits/8e42c2fca3085731081800a14a6cad63fdb898d7

pkedy:
@zhongfox This looks good. I added a call to convert the metadata.Configuration from `map[interface{}]interface{}` parsed from YAML to `map[string]interface{}`. I refactored a func to do this from the Consul nr component.

pkedy (approved):

nameresolution/configuration.go
```diff
@@ -0,0 +1,48 @@
...
+
+package nameresolution
+
+import (
+	"fmt"
+)
+
+// ConvertConfig converts map[interface{}]interface{} to map[string]interface{} to normalize
+// for JSON and usage in component initialization.
+func ConvertConfig(i interface{}) (interface{}, error) {
```
artursouza, line 14: Can this be moved into dapr/kit? Even better, is there any other existing utility that does that?

artursouza (changes requested):

nameresolution/configuration.go
pkedy: @artursouza I added this, and thought about putting it there. Do we know of anywhere else this is useful other than name resolution? In general, I like the idea of putting it in kit anyway. Can we proceed with merging this PR and I'll take an action item to move it to kit along with some other config stuff I am working on?

pkedy: There are definitely projects out there that deal with this.  [Here is a thread with some history](https://github.com/go-yaml/yaml/issues/139). In general, its probably not worth pulling in a new module just for this 1 function.

I'm thinking we should eventually move this to daprd so that the config comes pre converted/cleaned up. Again baby steps. I'll take this on as I move more config related stuff to kit.

artursouza: OK.

artursouza (approved):

Issue #779, opened 2021-03-26
Add `Close() error` to bindings


pkedy:
Currently, the PubSub interface contains a `Closer() error` function so that the component can cleanly shutdown when `daprd` is stopped. This does not exist in bindings and that has resulted in binding implementations are having to use `signal.Notify` in order to stop a subscription, etc.

Instead, PubSub and Binding interfaces could embed `io.Closer`. Many of the pubsub components have moved to using an internal content.Context for backoff/timeout purposes. The `Close` method should call the `cancel` func returned by `context.WithCancel` in order to idiomatically shutdown goroutines, etc.

This likely involves a change to `dapr/dapr` (most likely in `runtime.go`) to invoke `Close` on shutdown. It's also possible for components to optionally implement `io.Closer` so long as daprd checks for that and invokes accordingly. Our documentation on how to build a component would just need to make that clear.

pkedy:
This was resolved in dapr/dapr.  Opening new issues for each component as I run into "interrupt signal" based cleanup logic.

PR #804, opened 2021-04-10
Upgrade redis client to v8 and use context in redis pubsub component to cancel request

Taction:
# Description

Upgrade redis client to v8 and use context in redis pubsub component to cancel request.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


yaron2:
Please resolve conflicts.

bindings/redis/redis.go
```diff
@@ -66,7 +66,7 @@ func (r *Redis) Init(meta bindings.Metadata) error {
 	}
 
 	r.client = redis.NewClient(opts)
-	_, err = r.client.Ping().Result()
+	_, err = r.client.Ping(context.Background()).Result()
```
pkedy, line 69: I think it makes sense to add a `ctx` context variable in the `Redis` struct to handle cancelling the child contexts when closing the component. `r.ctx` would replace anywhere `context.Background()` is used. See the pubsub component.

pubsub/redis/redis.go
```diff
@@ -352,7 +352,7 @@ func (r *redisStreams) Init(metadata pubsub.Metadata) error {
 		r.client = redis.NewClient(options)
 	}
 
-	if _, err = r.client.Ping().Result(); err != nil {
+	if _, err = r.client.Ping(context.Background()).Result(); err != nil {
```
pkedy, line 355: If you move this below the context.WithCancel, you can use `r.ctx` instead of `context.Background()`

state/redis/redis.go
```diff
@@ -141,7 +141,7 @@ func (r *StateStore) Init(metadata state.Metadata) error {
 		r.client = r.newClient(m)
 	}
 
-	if _, err = r.client.Ping().Result(); err != nil {
+	if _, err = r.client.Ping(context.Background()).Result(); err != nil {
```
pkedy, line 144: I think it makes sense to add a `ctx` context variable in the `Redis` struct to handle cancelling the child contexts when closing the component. `r.ctx` would replace anywhere `context.Background()` is used. See the pubsub component.

pkedy (changes requested):
Thanks for this PR! We need to improve our usage of contexts/cancellation and this is related.

For the components that don't have a component-level `ctx context.Context`, we should add one and cancel it from `Close()`. If there is not currently a `Close()` func, please add one. There is this issue #779 to add it to bindings so it will be called on shutdown in the future.

bindings/redis/redis.go
Taction: I've add ctx and cancel the same as the pubsub component.Also implement `Close() error` function.

state/redis/redis.go
Taction: I've add ctx and cancel the same as the pubsub component.Also implement `Close() error` function.

pubsub/redis/redis.go
Taction: I've finish that.

pkedy (dismissed):
LGTM - Some of the `context.Background()` calls in the tests could be `ss.ctx` but since they are tests it really doesn't matter. Thanks for making the contexts work as they should!

artursouza (approved):

artursouza:
Please, run `go mod tidy`

Taction:
Finish resolve go mod. My go version is 1.15 before, which lead to `go mod tidy` have different behavior with 1.16.

wcs1only (approved):

artursouza (approved):

PR #807, opened 2021-04-10
Removing circular dependency on dapr/dapr

pkedy:
# Description

* Removed dependency on dapr/dapr in favor of dapr/kit
* Copied in `standalone_loader.go` from dapr/dapr for conformance tests
* Fixed go mod dependencies afterwards (`go mod tidy`)
* Go 1.16 in go.mod (which is I think why there are so many more entries in `go.mod` now)
* Removed accidental `fortio.org/fortio/log` dependency (probably meant to use `log` from the standard library)

## Issue reference

Part 1/2 for resolving dapr/dapr#2820

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo


pubsub/nats/nats.go
```diff
@@ -11,8 +11,8 @@ import (
 	"fmt"
 
 	"github.com/dapr/components-contrib/pubsub"
-	"github.com/dapr/dapr/pkg/logger"
-	nats "github.com/nats-io/go-nats"
+	"github.com/dapr/kit/logger"
+	nats "github.com/nats-io/nats.go"
```
artursouza, line 15: Why is this different?

artursouza (changes requested):
LGTM. Just one place where the change seems unrelated.

pubsub/nats/nats.go
pkedy: IIRC, I ran into errors when running `go mod tidy` that pointed back to using the old NATS client. Switching to the new client resolved the issue.

artursouza (approved):

PR #823, opened 2021-04-15
Add redis cluter mode for pubsub

Taction:
# Description

Add redis cluter mode for pubsub, change `*redis.Client` to a `redis.UniversalClient` interface which can be a Client or ClusterClient.Add a `redisType` metadata to define redis type as node or cluster.
There are two differences between redis cluster and node.
1. redis cluster's addr option is a string array, which is a string in node. Alough cluster client can find all redis cluster nodes by one redis node, but in case the node is breakdown that the cluster can get cluster nodes info by other nodes. It is better to define addr in cluster mode can include more nodes adderss separated by comma.
2. redis cluster can not define db option for it only use default db0.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#1431


pkedy (approved):

Taction:
I've finish related docs modify and create an issue in doc project

pubsub/redis/redis.go
```diff
@@ -92,6 +99,13 @@ func parseRedisMetadata(meta pubsub.Metadata) (metadata, error) {
 		m.password = val
 	}
 
+	if val, ok := meta.Properties[redisType]; ok && val != "" {
```
yaron2, line 102: This is a breaking change to Redis pub/sub. Can you make it so that `redisType` defaults to "node"?

yaron2 (changes requested):

pubsub/redis/redis.go
yaron2: Scratch that, missed an else statement :)

yaron2 (approved):

PR #827, opened 2021-04-19
Improve pulsar publisher performance.

xqianwang:
# Description

For pulsar connector component, it initiates a publisher for every single message publishing to a topic.  e.g. Publish 100 messages to persistent://apache/pulsar/test topic, daprd will create a publisher before sending message. This will affect performance a lot.  

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


pubsub/pulsar/pulsar.go
```diff
@@ -82,22 +83,28 @@ func (p *Pulsar) Init(metadata pubsub.Metadata) error {
 }
 
 func (p *Pulsar) Publish(req *pubsub.PublishRequest) error {
-	producer, err := p.client.CreateProducer(pulsar.ProducerOptions{
-		Topic: req.Topic,
-	})
-	if err != nil {
-		return err
+	if p.producer == nil || p.producer.Topic() != req.Topic {
```
artursouza, line 86: This will be helpful but I think we should take it a step further and have a multiplexer that will know which producer to use based on the topic and close them all in the end.

This could also be achieved with some sort of LRU cache, as long as the producers are closed() before being deleted from cache.

artursouza (changes requested):

pubsub/pulsar/pulsar.go
xqianwang: Updated PR for this, please review.

xqianwang:
Did a simple test on my local machine for performance comparation
Using code in this PR of pulsar component for publishing 100 messages: Used 1229 ms for sending messages
Using old code of pulsar component for publishing 100 messages: Used 4461 ms for sending messages
Almost 3-4 times faster in terms of performance

pubsub/pulsar/pulsar.go
```diff
@@ -69,6 +72,19 @@ func (p *Pulsar) Init(metadata pubsub.Metadata) error {
 	}
 	defer client.Close()
 
+	// initialize lru cache with size 100
+	c, err := lru.NewWithEvict(100, func(k interface{}, v interface{}) {
```
artursouza, line 76: Please, make this number a constant and use `10` instead. I don't think `100` is a reasonable default for number of topics being used.

Ideally, this constant should just be a default for a configuration from metadata - add a TODO for that.

artursouza (changes requested):

pubsub/pulsar/pulsar.go
xqianwang: Fixed

yaron2 (approved):

PR #832, opened 2021-04-21
Add timeout to cancel request for it would be redelivered

Taction:
# Description

For redis PubSub component, `processingTimeout` and `redeliverInterval` control message redelivering. But the old request  may not canceled when redelivered.

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


pkedy (approved):

pubsub/redis/redis.go
```diff
@@ -456,7 +456,13 @@ func (r *redisStreams) worker() {
 // by `reclaimPendingMessagesLoop`.
 func (r *redisStreams) processMessage(msg redisMessageWrapper) error {
 	r.logger.Debugf("Processing Redis message %s", msg.messageID)
-	if err := msg.handler(r.ctx, &msg.message); err != nil {
+	ctx := r.ctx
+	var cancel context.CancelFunc
+	if r.metadata.processingTimeout != 0 && r.metadata.redeliverInterval != 0 {
+		ctx, cancel = context.WithTimeout(ctx, r.metadata.processingTimeout)
+		defer cancel()
```
mukundansundar, line 463: @pkedy There is a r.cancel() in the Close() method. How does this defer affect the context and the call in r.cancel() ?

pkedy: @mukundansundar This `cancel()` only cancels the context created on line 462 and down the call stack.

mukundansundar: This cancels the derived context with timeout that was created in 462. In line 462, the context was created with the parent being r.ctx line 459. 
With the ctx being cancel here, will that affect the parent context and its usages elsewhere?

Taction: @mukundansundar This cancel will remove itself (the context created on line 462) from its parent (r.ctx), so it has no effect no its parent.And if r.cancel() is called first, the context created on line 462 will also be canceled, and this cancel() will do nothing.

artursouza (approved):

PR #836, opened 2021-04-22
Removing the Download Go Dependencies step from the build workflow

pkedy:
# Description

Lots of go lint errors are showing up in job runs since the `Download Go dependencies` step was added to `.github/workflows/components-contrib.yml`. Hoping removing it resolves the errors.

Also moved the timer in the pubsub conformance tests to a single variable used over each message iteration. Not 100% confident it will resolve the periodic conformance test timeouts but it is probably how the timer should work anyway.


artursouza (approved):

wcs1only (approved):
LGTM

Issue #841, opened 2021-04-26
Support version and engine path in vault secret store


Taction:
## Background information
Vault secret store component is based on the Vault KV secrets engine running in versioned mode. 
[Vault KV Secrets Engine - Version 2 (API)](https://www.vaultproject.io/api/secret/kv/kv-v2)

## Bugs

1. Only support secrets engine is enabled at the /secret path.
As described in the [API document](https://www.vaultproject.io/api/secret/kv/kv-v2):`This documentation assumes the kv secrets engine is enabled at the /secret path in Vault and that versioning has been enabled. Since it is possible to enable secrets engines at any location, please update your API calls accordingly.` But it is [hard coded](https://github.com/dapr/components-contrib/blob/master/secretstores/hashicorp/vault/vault.go#L149)

2. Do not support json format in value.
If the secret value is `{"json_data":{"mysql_db":"mysql_db"}}`, will lead to error `failed getting secrets from secret store vault: couldn't decode response body: json: cannot unmarshall object into Go struct field .data.data of type string`.

## Expected Behavior

1. Secrets can be defined in any path, not only /secret.
2. Secret value can be defined in json format. Add a `parseAsMap` option to define if the value need to be parsed into `map[string]string`, defaults to `true `to keep the same with previous, false means  handle the secret value as string, do not parse it as `map[string]string`. 

## Actual Behavior

1. Engine path /secret is hard coded
2. Parse secret value as `map[string]string`. If the secret value is `{"json_data":{"mysql_db":"mysql_db"}}`, will lead to error `failed getting secrets from secret store vault: couldn't decode response body: json: cannot unmarshall object into Go struct field .data.data of type string`.

## Steps to Reproduce the Problem

1. Define A secret key under /kv engine with dapr/hello path, it will be get from url `http://host:8200/v1/kv/data/dapr/hello`. And the path now is `vaultSecretPathAddr := fmt.Sprintf("%s/v1/secret/data/%s/%s?version=0", v.vaultAddress, v.vaultKVPrefix, secret)` will never generate a path like that.
2. Set the secret value to `{"json_data":{"mysql_db":"mysql_db"}}`. And get this secret.

## Release Note

RELEASE NOTE: Support version and engine path in vault secret store. Add parseAsMap option.

PR #842, opened 2021-04-26
Support version and engine path in vault secret store, add vaultValueType option.

Taction:
## Description

Support version and engine path in vault secret store, add parseAsMap option to define wether handle secret value as string or not. All of default behavior is  the same as previous.

1. Add engine path metadata for vault secret store component.
2. Add parseAsMap option. Define the behavior of parsing secret value as string or parsing it as `map[string]string`(defaults to parsing it as `map[string]string` keeping the same behavior with previous).
3. Support version for vault secret store.

## Issue reference

Please reference the issue this PR will close :  #841

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


Taction:
Now it is ready to be reviewed.

pkedy:
@Taction would you mind resolving the conflicts and check with `gofumpt`?

pkedy:
@Taction - more conflicts :-(

Would you mind resolving?

Taction:
@pkedy If it is better to split the different contents of this pr into different prs? In this way I think it might make code review easier.

secretstores/hashicorp/vault/vault.go
```diff
@@ -40,6 +43,12 @@ const (
 	defaultVaultKVPrefix         string = "dapr"
 	vaultHTTPHeader              string = "X-Vault-Token"
 	vaultHTTPRequestHeader       string = "X-Vault-Request"
+	vaultEnginePath              string = "enginePath"
+	parseAsMap                   string = "parseAsMap"
```
artursouza, line 47: Can these constants have a better name? Just naming it as the content will satisfy linter but does not really improve readability.

secretstores/hashicorp/vault/vault.go
```diff
@@ -96,6 +110,16 @@ func (v *vaultSecretStore) Init(metadata secretstores.Metadata) error {
 
 	v.vaultAddress = address
 
+	v.vaultEnginePath = defaultVaultEnginePath
+	if val, ok := props[vaultEnginePath]; ok && val != "" {
+		v.vaultEnginePath = val
+	}
+
+	v.parseAsMap = true
+	if val, ok := props[parseAsMap]; ok && strings.EqualFold(val, FalseStr) {
```
artursouza, line 119: Can you parse string to bool instead of equal?

secretstores/hashicorp/vault/vault.go
```diff
@@ -96,6 +110,16 @@ func (v *vaultSecretStore) Init(metadata secretstores.Metadata) error {
 
 	v.vaultAddress = address
 
+	v.vaultEnginePath = defaultVaultEnginePath
+	if val, ok := props[vaultEnginePath]; ok && val != "" {
+		v.vaultEnginePath = val
+	}
+
+	v.parseAsMap = true
```
artursouza, line 118: instead of parseAsMap what about the opposite and set something like "valueAsString". Another way, would be more extensible: "valueType" and then accept enum values like "map" and "text".

artursouza (changes requested):

secretstores/hashicorp/vault/vault.go
Taction: Good! Fixing all above by refactoring parseAsMap to vaultValueType, which accept two values map and text.

secretstores/hashicorp/vault/vault.go
```diff
@@ -197,55 +236,96 @@ func (v *vaultSecretStore) getSecret(secret string) (*vaultKVResponse, error) {
...
+	}
+
+	keys, err := v.listKeysUnderPath("")
+	if err != nil {
+		return secretstores.BulkGetSecretResponse{}, err
+	}
+
+	for _, key := range keys {
+		keyValues := map[string]string{}
+		secrets, err := v.getSecret(key, version)
```
artursouza, line 293: Can version as be passed in the Bulk Get API? it seems to always default to 0.

artursouza (changes requested):
LGTM, just a small change.

secretstores/hashicorp/vault/vault.go
Taction: The reason why I didn't add the version in the Bulk Get API is because that, for one secret if the version is specified but does not exist, none secret information will be returned. Therefore, when obtaining secrets in batches, the version is hard coded to 0 which means the latest version.

If it is better to do like following:
Using version passed in the Bulk Get API, those secrets ​​that do not have the specified version will not appear. If the version is not specified it will defaults to 0.  And write this into the document, to let users know how to use it better.

Taction:
I think this may be useful and should not be closed.
/cc @artursouza @pkedy

beiwei30 (dismissed):
This change looks good to me.  @artursouza your opinion on it?

LaurenceLiZhixin (approved):

sduo:
Hi Teams

I found this pr could fix some problems for me.
1st, the config file could defind "defaultVaultEnginePath" value, but in version 1.5.1 of dapr ,it is a raw code "secret";
2nd, rewrite the function getSecret, it looks like will be work fine;

so, please help me

yaron2:
LGTM. @artursouza are you ok with these changes?

yaron2 (approved):
lgtm

yaron2:
@Taction does this need a docs update?

Taction:
@yaron2 docs pr https://github.com/dapr/docs/pull/2174

secretstores/hashicorp/vault/vault.go
```diff
@@ -49,15 +52,37 @@ const (
 	defaultVaultKVPrefix         string = "dapr"
 	vaultHTTPHeader              string = "X-Vault-Token"
 	vaultHTTPRequestHeader       string = "X-Vault-Request"
+	vaultEnginePath              string = "enginePath"
+	vaultValueType               string = "vaultValueType"
+	versionID                    string = "version_id"
```
artursouza, line 57: This is not following the same naming convention. Please, update.

artursouza (changes requested):
Did you assess if there is a breaking change?

secretstores/hashicorp/vault/vault.go
Taction: There is no breaking change, the default value keeps it backward compatible.

`enginePath` and `vaultValueType` are used in component metadata, these fields are in Camel-Case.

`version_id` is used in query parameters, and other secret stores which support the version are using this key. the doc link: https://docs.dapr.io/reference/api/secrets_api/#query-parameters

artursouza (approved):

PR #844, opened 2021-04-26
Add GraphQL Binding

amirmm11:
# Description

Added GraphQL binding to connect to Hasura server

## Issue reference

This PR will close: #902


yaron2:
@mukundansundar can you please review?

halspang:
> @mukundansundar @yaron2
> 
> I think we will have to adjust the lint checker for Graphql/Hasura Operations which are `query` and `mutation` differ from other usual CRUD bindings and does not have `CreateOperation, DeleteOperation, GetOperation, ListOperation`
> 
> Let us know where to make that adjustment will be happy to do so. Thank you.

Wouldn't the desired behavior here be to have a case for all of those and return an error saying it's not a supported operation? You can see how the Azure storage binding does it as it only supports 3 out of 4 operations:

https://github.com/dapr/components-contrib/blob/master/bindings/azure/blobstorage/blobstorage.go#L204

alvinhenrick:
> > @mukundansundar @yaron2
> > I think we will have to adjust the lint checker for Graphql/Hasura Operations which are `query` and `mutation` differ from other usual CRUD bindings and does not have `CreateOperation, DeleteOperation, GetOperation, ListOperation`
> > Let us know where to make that adjustment will be happy to do so. Thank you.
> 
> Wouldn't the desired behavior here be to have a case for all of those and return an error saying it's not a supported operation? You can see how the Azure storage binding does it as it only supports 3 out of 4 operations:
> 
> https://github.com/dapr/components-contrib/blob/master/bindings/azure/blobstorage/blobstorage.go#L204

@halspang  Sounds like the right way to do it but it will have side effect to update other components to account for new operation `query` and `mutation` . 

Two new entries will have to be made and other component will have to `fallthrough` the two new operation. Linter might fail for other two cases missing in older bindings. 

https://github.com/dapr/components-contrib/blob/master/bindings/requests.go#L19-L24

```
const (
	GetOperation    OperationKind = "get"
	CreateOperation OperationKind = "create"
	DeleteOperation OperationKind = "delete"
	ListOperation   OperationKind = "list"
	QueryOperation	OperationKind = "query"
	MutationOperation OperationKind = "mutation"
)
```
We can also avoid updating the `requests.go` the way @amirmm11 has implemented it and apply the lint per binding because each binding can have its own kind of operation.

Let us know what would be a good idea. If adding two new operation and old binding will not complain we should be good.

Thank you.

halspang:
> > > @mukundansundar @yaron2
> > > I think we will have to adjust the lint checker for Graphql/Hasura Operations which are `query` and `mutation` differ from other usual CRUD bindings and does not have `CreateOperation, DeleteOperation, GetOperation, ListOperation`
> > > Let us know where to make that adjustment will be happy to do so. Thank you.
> > 
> > 
> > Wouldn't the desired behavior here be to have a case for all of those and return an error saying it's not a supported operation? You can see how the Azure storage binding does it as it only supports 3 out of 4 operations:
> > https://github.com/dapr/components-contrib/blob/master/bindings/azure/blobstorage/blobstorage.go#L204
> 
> @halspang Sounds like the right way to do it but it will have side effect to update other components to account for new operation `query` and `mutation` .
> 
> Two new entries will have to be made and other component will have to `fallthrough` the two new operation. Linter might fail for other two cases missing in older bindings.
> 
> https://github.com/dapr/components-contrib/blob/master/bindings/requests.go#L19-L24
> 
> ```
> const (
> 	GetOperation    OperationKind = "get"
> 	CreateOperation OperationKind = "create"
> 	DeleteOperation OperationKind = "delete"
> 	ListOperation   OperationKind = "list"
> 	QueryOperation	OperationKind = "query"
> 	MutationOperation OperationKind = "mutation"
> )
> ```
> 
> We can also avoid updating the `requests.go` the way @amirmm11 has implemented it and apply the lint per binding because each binding can have its own kind of operation.
> 
> Let us know what would be a good idea. If adding two new operation and old binding will not complain we should be good.
> 
> Thank you.

I don't think we should make old bindings fail if they don't have the new operation. I would be fine with letting `default` be considered covering for exhaustive. This is at least how I've handled things in the past in languages like Java. There's a setting for it in the linter but I'll leave it up to @mukundansundar or @pkedy to decide if they're OK with that.

Here's the setting you'd put in `.golangci.yaml`:

```yaml
  exhaustive:
    # check switch statements in generated files also
    check-generated: false
    # indicates that switch statements are to be considered exhaustive if a
    # 'default' case is present, even if all enum members aren't listed in the
    # switch
    default-signifies-exhaustive: true
```
I haven't tried it, but I think that'd work.

pkedy (commented):
Hi @amirmm11.  Is there any way this component could be generalized to a "GraphQL" component by putting headers like `x-hasura-access-key` in the component config?

yaron2:
> > > > @mukundansundar @yaron2
> > > > I think we will have to adjust the lint checker for Graphql/Hasura Operations which are `query` and `mutation` differ from other usual CRUD bindings and does not have `CreateOperation, DeleteOperation, GetOperation, ListOperation`
> > > > Let us know where to make that adjustment will be happy to do so. Thank you.
> > > 
> > > 
> > > Wouldn't the desired behavior here be to have a case for all of those and return an error saying it's not a supported operation? You can see how the Azure storage binding does it as it only supports 3 out of 4 operations:
> > > https://github.com/dapr/components-contrib/blob/master/bindings/azure/blobstorage/blobstorage.go#L204
> > 
> > 
> > @halspang Sounds like the right way to do it but it will have side effect to update other components to account for new operation `query` and `mutation` .
> > Two new entries will have to be made and other component will have to `fallthrough` the two new operation. Linter might fail for other two cases missing in older bindings.
> > https://github.com/dapr/components-contrib/blob/master/bindings/requests.go#L19-L24
> > ```
> > const (
> > 	GetOperation    OperationKind = "get"
> > 	CreateOperation OperationKind = "create"
> > 	DeleteOperation OperationKind = "delete"
> > 	ListOperation   OperationKind = "list"
> > 	QueryOperation	OperationKind = "query"
> > 	MutationOperation OperationKind = "mutation"
> > )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We can also avoid updating the `requests.go` the way @amirmm11 has implemented it and apply the lint per binding because each binding can have its own kind of operation.
> > Let us know what would be a good idea. If adding two new operation and old binding will not complain we should be good.
> > Thank you.
> 
> I don't think we should make old bindings fail if they don't have the new operation. I would be fine with letting `default` be considered covering for exhaustive. This is at least how I've handled things in the past in languages like Java. There's a setting for it in the linter but I'll leave it up to @mukundansundar or @pkedy to decide if they're OK with that.
> 
> Here's the setting you'd put in `.golangci.yaml`:
> 
> ```yaml
>   exhaustive:
>     # check switch statements in generated files also
>     check-generated: false
>     # indicates that switch statements are to be considered exhaustive if a
>     # 'default' case is present, even if all enum members aren't listed in the
>     # switch
>     default-signifies-exhaustive: true
> ```
> 
> I haven't tried it, but I think that'd work.

Components can have component specific operations. Even if you add those to `requests.go` it shouldn't affect other components.

artursouza:
> @amirmm11 That is a very good question. Yes, it can be done with `Metadata` but I don't think it will be pretty because the 1) the properties are `map[string]string` and 2) secrets are involved.
> 
> There could be some prefix based way to do this (But again, not pretty)
> 
> ```yaml
>   metadata:
>   - name: headerKeys
>     value: headerAccessKey
>   - name: headerAccessKeyName
>     value: "x-hasura-access-key"
>   - name: headerAccessKeyValue
>     secretKeyRef:
>     	name: hasura-secret
>         key:  hasura-password
> ```
> 
> /cc @artursouza @mukundansundar @yaron2 Thoughts on how to configure a nested `map[string]string` like header values?


With the current design, it would require some inner structure in the string like as URL query params encoding: "X=1&Y=2"

pkedy:
@artursouza Don't forget the use case here involves secrets (specifically an access key). I didn't think you could embed secrets using a single `headers` property with query params encoding. Is that incorrect? In other words can you do something like:

```
x-hasura-access-key=${secret(hasura-secret,hasura-password)}&X=1&Y=2
```

If so, it is not documented here: https://docs.dapr.io/operations/components/component-secrets/

artursouza:
> @artursouza Don't forget the use case here involves secrets (specifically an access key). I didn't think you could embed secrets using a single `headers` property with query params encoding. Is that incorrect? In other words can you do something like:
> 
> ```
> x-hasura-access-key=${secret(hasura-secret,hasura-password)}&X=1&Y=2
> ```
> 
> If so, it is not documented here: https://docs.dapr.io/operations/components/component-secrets/


I see - then we should not wait for this inlining feature for this PR (please, create an issue to track that) we already have `{uuid}` today and we add more values to be resolved in metadata.

I am OK with the experience suggested by Phil. Once we have inline values for secrets, we can support both ways.

amirmm11:
@yaron2 

I updated the code to use a map as a header to connect any GraphQL connection and fixed the linting issues.

```golang
// GraphQL represents GraphQL output bindings
type GraphQL struct {
	client *graphql.Client
	header map[string]string
	logger logger.Logger
}
```
and the Yaml sidecar should like this to connect to Hasura, for example:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: example.bindings.graphql
spec:
  type: bindings.graphql
  version: v1
  metadata:
    - name: endpoint
      value: localhost:8080
    - name: x-hasura-access-key
      value: localhost:1234
 ```

**Comment by pkedy** [comment]
@amirmm11 I like this simpler approach more than the what I recommended. But there still needs to be a way to separate normal config settings from headers. What if you used a prefix on the names like this?

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: example.bindings.graphql
spec:
  type: bindings.graphql
  version: v1
  metadata:
    - name: endpoint
      value: localhost:8080
    - name: header:x-hasura-access-key
      secretKeyRef:
    	name: hasura-secret
        key:  hasura-password
```

alvinhenrick:
> @amirmm11 I like this simpler approach more than the what I recommended. But there still needs to be a way to separate normal config settings from headers. What if you used a prefix on the names like this?
> 
> ```yaml
> apiVersion: dapr.io/v1alpha1
> kind: Component
> metadata:
>   name: example.bindings.graphql
> spec:
>   type: bindings.graphql
>   version: v1
>   metadata:
>     - name: endpoint
>       value: localhost:8080
>     - name: header:x-hasura-access-key
>       secretKeyRef:
>     	name: hasura-secret
>         key:  hasura-password
> ```

@pkedy would it be sufficient and safe to assume that header will start with `x-zzzzz` ... Anything not following such convention can be safely assumed as not a header ? Following the HTTP conventional style.

amirmm11:
> > @amirmm11 I like this simpler approach more than the what I recommended. But there still needs to be a way to separate normal config settings from headers. What if you used a prefix on the names like this?
> > ```yaml
> > apiVersion: dapr.io/v1alpha1
> > kind: Component
> > metadata:
> >   name: example.bindings.graphql
> > spec:
> >   type: bindings.graphql
> >   version: v1
> >   metadata:
> >     - name: endpoint
> >       value: localhost:8080
> >     - name: header:x-hasura-access-key
> >       secretKeyRef:
> >     	name: hasura-secret
> >         key:  hasura-password
> > ```
> 
> @pkedy would it be sufficient and safe to assume that header will start with `x-zzzzz` ... Anything not following such convention can be safely assumed as not a header ? Following the HTTP conventional style.
> 
> In this case `endpoint` is clearly not a header while `x-hasura-access-key` by following http style header convention.

@alvinhenrick @pkedy I assumed that everything except endpoint is a header. I just wanted not to add many complexities to the sidecar.

pkedy:
@amirmm11 I think we want config to be explicit so that if the user makes a mistake they receive and error instead of treating is as a header. I think the prefix is succinct, clear, and doesn't add too much complexity.

@alvinhenrick You are correct in that custom headers should have a `x-` prefix, but I'm concerned about a case where the backend violates that convention or you need to specify a standard header (e.g. `Authorization`).

IMO this is a case where I think [clear is better than clever](https://dave.cheney.net/2019/07/09/clear-is-better-than-clever).

amirmm11:
> @amirmm11 I think we want config to be explicit so that if the user makes a mistake they receive and error instead of treating is as a header. I think the prefix is succinct, clear, and doesn't add too much complexity.
> 
> @alvinhenrick You are correct in that custom headers should have a `x-` prefix, but I'm concerned about a case where the backend violates that convention or you need to specify a standard header (e.g. `Authorization`).
> 
> IMO this is a case where I think [clear is better than clever](https://dave.cheney.net/2019/07/09/clear-is-better-than-clever).

@pkedy Done as you recommended.

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: example.bindings.graphql
spec:
  type: bindings.graphql
  version: v1
  metadata:
    - name: endpoint
      secretKeyRef:
        name: dev/RDS/hasura_url
        key: dev/RDS/hasura_url
    - name: header:x-hasura-access-key
      secretKeyRef:
        name: dev/RDS/hasura_admin_secret
        key: dev/RDS/hasura_admin_secret
auth:
  secretStore: example.aws.secretmanager
  ```

**Comment by alvinhenrick** [comment]
> @amirmm11 I think we want config to be explicit so that if the user makes a mistake they receive and error instead of treating is as a header. I think the prefix is succinct, clear, and doesn't add too much complexity.
> 
> @alvinhenrick You are correct in that custom headers should have a `x-` prefix, but I'm concerned about a case where the backend violates that convention or you need to specify a standard header (e.g. `Authorization`).
> 
> IMO this is a case where I think [clear is better than clever](https://dave.cheney.net/2019/07/09/clear-is-better-than-clever).

Sounds good.

**Comment by pkedy** [comment]
@amirmm11 I was trying to help resolve the linter issues and refactored the code a bit to resolve the cyclomatic complexity. I found that the query and mutation commands were basically the same logic. I've attached a diff for you to see.

Given that the query and mutation commands are basically the same, would it make sense to verify that queries command strings are queries and mutations command strings are mutations? To do this, is it as simple as checking the first keyword?

```golang
	requestString = strings.TrimSpace(requestString)
	re := regexp.MustCompile(`(?m)` + requestKey + `\b`)
	matches := re.FindAllStringIndex(requestString, 1)
	if len(matches) != 1 || matches[0][0] != 0 {
		return fmt.Errorf("GraphQL Error: command is not a %s", requestKey)
	}
```

[patch.diff.zip](https://github.com/dapr/components-contrib/files/6664893/patch.diff.zip)

amirmm11:
Thank you @pkedy 

I added your changes.

pkedy:
@amirmm11 Looks like a gofumpt issue.  Pls change the loop at line 68 to

```golang
	for k, v := range p {
		if strings.HasPrefix(k, "header:") {
			gql.header[strings.TrimPrefix(k, "header:")] = v
		}
	}
```

This removes and extra newline and also that error check which is not necessary.

bindings/graphql/graphql.go
```diff
@@ -0,0 +1,172 @@
...
+	client *graphql.Client
+	header map[string]string
+	logger logger.Logger
+}
+
+var _ = bindings.OutputBinding(&GraphQL{})
+
+// NewGraphQL returns a new GraphQL binding instance
+func NewGraphQL(logger logger.Logger) *GraphQL {
+	var gql GraphQL
```
pkedy, line 47: Simplification:

```golang
return &GraphQL{
    logger: logger
}
```

bindings/requests.go
```diff
@@ -17,8 +17,10 @@ type OperationKind string
...
 const (
-	GetOperation    OperationKind = "get"
-	CreateOperation OperationKind = "create"
-	DeleteOperation OperationKind = "delete"
-	ListOperation   OperationKind = "list"
+	GetOperation      OperationKind = "get"
+	CreateOperation   OperationKind = "create"
+	DeleteOperation   OperationKind = "delete"
+	ListOperation     OperationKind = "list"
+	QueryOperation    OperationKind = "query"
```
pkedy, line 24: Late catch: Component specific operation constants should not go here. Please move them to `grapnel.go`.

bindings/graphql/graphql.go
amirmm11: @pkedy I created that way to avoid getting linting warning `graphql.go:50:10  exhaustivestruct  client, header are missing in GraphQL'

amirmm11:
@pkedy Done. Thanks

pkedy (approved):
@amirmm11 Thanks for hanging in there! LGTM!

/cc @artursouza ready for review

bindings/graphql/graphql.go
```diff
@@ -0,0 +1,169 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/dapr/components-contrib/bindings"
+	"github.com/dapr/kit/logger"
+	graphql "github.com/machinebox/graphql"
```
artursouza, line 18: This library is not as active as this other one: https://github.com/graphql-go/graphql

Any reason why this one was chosen?

artursouza (changes requested):

bindings/graphql/graphql.go
amirmm11: @artursouza That is a very light library that is sufficient for building a binding.

pkedy: I like this choice because this library focuses on being a very simple GraphQL client and does not provide all the extra server-side support.  Also, main contributor is a well-known Gopher ;-)

If there are issues in the future w/r/t maintenance we can swap it out.

alvinhenrick: @artursouza @yaron2 Can we look into this ? I think we would like this component in next release if possible.. We can always release the updated version if the library causes any issues ..We have been using it for a while and it is stable....

alvinhenrick: @artursouza @yaron2 Can we look into this ? I think we would like this component in next release if possible.. We can always release the updated version if the library causes any issues ..We have been using it for a while and it is stable....

artursouza: OK. LGTM.

artursouza (approved):

PR #854, opened 2021-05-03
Configurable retry

### LINKED ISSUE #853: Make the retry package available to use across all components
## Describe the proposal

Currently, there is a retry "helper" available in the pubsub package. This should be enhanced as appropriate and moved to `internal` for general use inside components contrib.

pkedy:
# Description

Moves the retry helper from `pubsub` to the `internal` package for general use and adds the ability to decode retry configuration (constant vs. exponential, etc) from an `interface{}` (e.g. metadata/settings).

## Issue reference

Resolves #853

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


internal/retry/retry.go
```diff
@@ -0,0 +1,158 @@
...
+
+	"github.com/dapr/components-contrib/internal/config"
+)
+
+// PolicyType denotes if the back off delay should be constant or exponential.
+type PolicyType int
+
+const (
+	// PolicyConstant is a backoff policy that always returns the same backoff delay.
+	PolicyConstant PolicyType = iota
```
artursouza, line 19: What is this `itoa` trick?

internal/retry/retry.go
```diff
@@ -0,0 +1,158 @@
...
+func DecodeConfigWithPrefix(c *Config, input interface{}, prefix string) error {
+	input, err := config.PrefixedBy(input, prefix)
+	if err != nil {
+		return err
+	}
+
+	return DecodeConfig(c, input)
+}
+
+// NewBackOff returns a BackOff instance for use with `RetryNotifyRecover`
```
artursouza, line 79: nit: NotifyRecover

artursouza (commented):
Some small comments. Overall, LGTM.

pkedy:
@artursouza Thanks for the review.  I'm going to convert this to a Draft so I can double check that we have not regressed on default num retries.

internal/retry/retry.go
pkedy: https://yourbasic.org/golang/iota/
https://golang.org/ref/spec#Iota

artursouza: LGTM! I learned something new.

pkedy:
@artursouza OK so we have not regressed in that all of the components using retry will default to constant backoff event 5 seconds indefinitely. For Kafka that is essential.  For the other brokers that have retries baked into the broker itself, we might want to consider a capped default maxRetries.

So I think this PR is ready to merge but I may submit another issue/PR that recommends other defaults based on the broker. Open to ideas here. This is what is nasty about retries :)

artursouza (approved):

Issue #852, opened 2021-05-03
Redis subscriber stops receiving messages after TCP reset.


wcs1only:
## Expected Behavior

If a TCP connection goes away without the redis subscriber receiving a TCP FIN or RST, the subscriber should timeout and reconnect.

## Actual Behavior

There is an issue with our E2E tests on Windows that is causing TCP connections to be dropped with a RST whenever a public IP address is added to a service. We have yet to root cause why this is happening, but this is definitely a case that Dapr should handle gracefully.

Here is how the Dapr redis subscriber component operates today:

|redis|  message direction|                           Dapr|
| -- | -- |--|
|                                    |<-| xreadgroup block:0 (forever)|
|waits for message | | |
|new message | -> | process message |
|                                  | <- | xack|

If some intermediary (like a stateful firewall or LB) looses context on the connection while Dapr is blocked, Dapr may hang indefinitely not realizing that the connection has gone away. 

|redis|  message direction|                           Dapr|
| -- | -- |--|
|                                    |<-| xreadgroup block:0 (forever)|
|waits for message | | |
|new message | -> <- X RST | Does not get message or RST |
| connection closed| | Connection still open, waiting for xreadgroup response |

Eventually, TCP keepalive will come along and detect that connection is actually gone vs just waiting for data and surface a connection close to Dapr. After which Dapr will recover. This can take several minutes or even hours depending on OS behavior.

## Steps to Reproduce the Problem

This can be reproduced using any simple pubsub setup (like the pubsub quickstart). All you need to do is trigger a connection
RST temporarily. 

Assuming your redis server is running on Linux, all you have to do is drop outbound traffic with a RST for 5s or so:
```
sudo iptables -I OUTPUT -p tcp --sport 6379 -j REJECT --reject-with tcp-reset && sleep 5 && sudo iptables -D OUTPUT -p tcp --sport 6379 -j REJECT --reject-with tcp-reset
```

## Release Note

RELEASE NOTE: **FIX** Redis subscriber should not block infinitely on XReadGroup

rtrindvg:
Are you sure this bug is fixed? I am using 1.6.0 and as reported on issue https://github.com/dapr/dapr/issues/4302, the problem still seems to exist.

Since my redis lives inside the same k8s cluster as daprd, it may be recreated with another IP when restarted. I am not sure if this changes anything related to this fix.

yaron2:
[...]

PR #855, opened 2021-05-04
Do not block indefinitely on XGroupRead for redis subscriber

wcs1only:
# Description

Allow XGroupRead operation to timeout periodically so we will know when a connection has disappeared. There is an existing redis parameter to read timeout, we can use it for our XGroupRead interval.

## Issue reference

Please reference the issue this PR will close: #852

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/pubsub/redis/redis.go
+++ b/pubsub/redis/redis.go
@@ -502,10 +502,12 @@ func (r *redisStreams) pollNewMessagesLoop(stream string, handler pubsub.Handler
 			Consumer: r.metadata.consumerID,
 			Streams:  []string{stream, ">"},
 			Count:    int64(r.metadata.queueDepth),
-			Block:    0,
+			Block:    r.metadata.readTimeout,
 		}).Result()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pubsub/redis/redis.go
```diff
@@ -502,10 +502,12 @@ func (r *redisStreams) pollNewMessagesLoop(stream string, handler pubsub.Handler
 			Consumer: r.metadata.consumerID,
 			Streams:  []string{stream, ">"},
 			Count:    int64(r.metadata.queueDepth),
-			Block:    0,
+			Block:    r.metadata.idleCheckFrequency,
```
wcs1only, line 505: I went back and forth about using readTimeout here instead. What do we think?

artursouza: Well. Frequency gives the impression that this is a recurring task. Timeout seems to be more generic.

I think readTimeOut might be a better fit here.

wcs1only: 👍
I'll change it.

artursouza (approved):

PR #858, opened 2021-05-05
Fix DNS name resolver to work without namespace

abogdanov37:
# Description

Add new name resolver based on DNS. It needs to implement custom deployment in Kubernetes. 

## Issue reference

dapr/dapr#2865

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] **Documentation update not needed.** Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_


nameresolution/dns/dns.go
```diff
@@ -26,7 +27,11 @@ func (k *resolver) Init(metadata nameresolution.Metadata) error {
...
 
-// ResolveID resolves name to address in orchestrator.
+// ResolveID resolves name to address use DNS
 func (k *resolver) ResolveID(req nameresolution.ResolveRequest) (string, error) {
-	return fmt.Sprintf("%s-dapr.%s.svc:%d", req.ID, req.Namespace, req.Port), nil
+	if strings.TrimSpace(req.Namespace) != "" {
+		return fmt.Sprintf("%s-dapr.%s.svc:%d", req.ID, req.Namespace, req.Port), nil
+	}
+
+	return fmt.Sprintf("%s-dapr:%d", req.ID, req.Port), nil
```
pkedy, line 36: Hi, @abogdanav37. I'm trying to understand the rationale for this change. For service to service invocation, Dapr apps are not aware of the namespace. Can you clarify the need for this DNS pattern? Thanks!

pkedy: @abogdanov37 just a friendly ping.

abogdanov37: Hi! Sorry for the long response time. The main idea is running in different environments. When we run in Kubernetes we can use namespace but when we run in docker-compose it won't work.

abogdanov37: We can use mDNS mode for this case but it works only on a single machine

artursouza: @pkedy @yaron2 Can this PR be merged?

Issue #857, opened 2021-05-05
Move The common Redis code to internal


Taction:
## Describe the proposal

Redis is used in pubsub、binding and state store. The parsing of redis metadata 、creating redis instance and supporting redis cluster will be repeated. So move the redis client 、 metadata and the parsing metadata function to internal (maybe internal/component/redis), and embed it in pubsub、binding and state store.

This will also fix #851.

PR #859, opened 2021-05-05
Move the common Redis code to internal

Taction:
# Description

Move Redis client and its metadata to internal.

## Issue reference

Please reference the issue this PR will close: #857 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


Taction:
I will refactor it to setting pattern.

pkedy:
@Taction I love that you will use the settings pattern. Thank for doing it! Just be careful that we don't have a regression. That will be the challenge getting it applied to all the existing components.

Taction:
@pkedy Now it is ready to be reviewed. I try to modify the existing content as little as possible to facilitate code review. I extracted the part of parsing redis client and settings form properties.And I've tested the redis cluser and node mode in my local enviroment.

One thing needs extra attention, I changed the behavior of parsing `time.Duration` in file `internal/config/decode.go`. If it is a negative value, keep the original value instead of resolving to milliseconds(because -1 is a useful vaule in some redis duration option) .

pkedy:
@Taction This is great! I'll try to review this tomorrow. Got my 2nd vaccine shot today so if not tomorrow then early next week.

The change to the config package makes sense to me. Thanks again for incorporating it! You are a true hero :)

Taction:
[Redis state store](https://docs.dapr.io/reference/components-reference/supported-state-stores/setup-redis/#spec-metadata-fields) already have `maxRetries ` and `maxRetryBackoff` metadata, for backward compatibility I make this change.

internal/config/decode.go
```diff
@@ -90,9 +90,12 @@ func decodeString(
 
 	switch t {
 	case typeDuration:
-		// Check for simple integer values and treat them
+		// Check for simple integer value and if the value is positive treat it
 		// as milliseconds
 		if val, err := strconv.Atoi(dataString); err == nil {
+			if val < 0 {
```
pkedy, line 96: Now that I look at this, since this is Redis specific, I think it makes more sense to use a type alias for this in `internal/component/redis/settings.go`. It would just move the logic you have to the `DecodeString` function.

```golang
type Duration time.Duration

func (r *Duration) DecodeString(value string) error {
	if val, err := strconv.Atoi(value); err == nil {
		if val < 0 {
			*r = Duration(val)
			return nil
		}
		*r = Duration(time.Duration(val) * time.Millisecond)
		return nil
	}

	// Convert it by parsing
	d, err := time.ParseDuration(value)
	if err != nil {
		return err
	}

	*r = Duration(d)
	return nil
}
```

That of course creates a bunch of places where you need to cast/convert to `time.Duration` so I went ahead and attached a patch file.  Please review.

[patch.diff.zip](https://github.com/dapr/components-contrib/files/6542508/patch.diff.zip)

pkedy (commented):
This looks good. Please use an alias Duration type for handling the `-1` case since it is Redis specific. I don't know when a negative duration would make sense but I don't want to take chances :)

internal/config/decode.go
Taction: @pkedy  LGTM! I've reviewed the changes, and as you said although `we don't know when a negative duration would make sense`, but processing the code in this way is more robust！

pkedy: I'm not sure which way you are (alias type vs globally) thinking is more robust. I'm going to assume it is OK to apply the patch to this branch. If not, we can revert.

pkedy (approved):
LGTM but I also want @mukundansundar to take a look to make sure we didn't have a regression with the settings refactoring.

pkedy:
@artursouza Let's update `go.mod` in `dapr/dapr` as soon as this is merged so we can have it proactively go through e2e testing.

internal/config/decode.go
Taction: Sorry for didn't make myself clear. Briefly I prefer alias type too.

CodeMonkeyLeet:
@mukundansundar Have you had the chance to take a look a this?

bindings/redis/redis.go
```diff
@@ -48,32 +45,21 @@ func NewRedis(logger logger.Logger) *Redis {
 
 // Init performs metadata parsing and connection creation
 func (r *Redis) Init(meta bindings.Metadata) error {
-	m, err := r.parseMetadata(meta)
+	_, err := r.parseMetadata(meta)
```
artursouza, line 48: Why is this parsing still needed if the output is dropped?

internal/component/redis/settings.go
```diff
@@ -0,0 +1,103 @@
...
+	"fmt"
+	"strconv"
+	"time"
+
+	"github.com/dapr/components-contrib/internal/config"
+)
+
+type Settings struct {
+	// The Redis host
+	Host string `mapstructure:"redisHost"`
```
artursouza, line 18: Is this structure backwards compatible in the sense that existing components from apps will continue to work as-is?

bindings/redis/redis.go
Taction: Because of there are two config `maxRetries` and `maxRetryBackoff` are parsed but not used. And I think just delete these code may not be a good idea, at least it shouldn’t be in this pr. Maybe a separate issue should be opened to discuss this.

internal/component/redis/settings.go
Taction: Yes, benefiting from the configuration of these components is relatively small, and most of the naming rules are consistent. There are also two configs have different metadata name for the same configuration, but I've handled this by parsing the old metadata first.

Issue #861, opened 2021-05-07
Add backoff for rabbitmq PubSub component


Taction:
## Describe the proposal

Add backoff for rabbitmq PubSub component.

1. Add `backOffEnable` option to define wether backoff is enabled or not.
2. Use retry helper to support backoff.

If this is acceptable, please let me know. I will fix the doc asap.

PR #862, opened 2021-05-07
Add backoff for rabbitmq PubSub component

Taction:
# Description

Add backoff for rabbitmq PubSub component.

## Issue reference

Please reference the issue this PR will close: #861 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


pubsub/rabbitmq/metadata.go
```diff
@@ -14,6 +14,7 @@ type metadata struct {
 	deleteWhenUnused bool
 	autoAck          bool
 	requeueInFailure bool
+	backOffEnable    bool  // Enable backoff or not
```
pkedy, line 17: Would it make more sense to have backOff enabled by default and disabled by setting `backOffMaxRetries = 0`?  I'm thinking this setting may not be necessary. Thoughts?

Taction: Yes, that's makes sense. I will fix it.

A small additional points: the `hazelcast` and `mqtt` component's `backOffMaxRetries` option is defaults to `"0"` which means the component will not retry processing the message. And for this component its default value is `-1` which means retry indefinitely. If there is any necessary to keep it consistent across all components.

Taction:
It seems there are some changes need to be done in the tests, I'll check and fix it later.

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -27,6 +28,7 @@ const (
 	metadataDeliveryModeKey      = "deliveryMode"
 	metadataRequeueInFailureKey  = "requeueInFailure"
 	metadataReconnectWaitSeconds = "reconnectWaitSeconds"
+	metadataBackOffEnable        = "backOffEnable"
```
pkedy, line 31: There is a constant here from before.

pubsub/rabbitmq/metadata.go
pkedy: Yeah, that's a great point. Here is my thought:

For brokers that retry at a per-message level, the default can be `0` and dapr-level retries are "opt in" (because the broker handles retries on its own).
For brokers that are offset/position based (like Kafka), the indefinite retries are essential or messages are not potentially dropped/not processed. Once pubsub has something like dead letter/retry queues, a `0` default value would be "safe".

pubsub/rabbitmq/rabbitmq.go
Taction: fixed

pubsub/rabbitmq/metadata.go
Taction: Added another default config `DefaultConfigWithNoRetry` in internal/retry, which set the value of `MaxRetries` to 0. For brokers that retry at a per-message level, can use this config as its default config.

pkedy:
@Taction Thanks for the fixes. Will look at this again soon. We are in the middle of our v1.2 release.

internal/retry/retry.go
```diff
@@ -55,6 +55,15 @@ func DefaultConfig() Config {
 	}
 }
 
+// DefaultConfigWithNoRetry represents the default configuration with `MaxRetries` set to 0.
+// This may be useful for those brokers which can handles retries on its own.
+func DefaultConfigWithNoRetry() Config {
```
pkedy, line 60: Great idea!

pkedy (approved):
LGTM

artursouza (approved):

PR #864, opened 2021-05-08
Add function to generate CloudEvent for raw subscribed events.

artursouza:
# Description

Add function to generate CloudEvent for raw subscribed events.

## Issue reference

Please reference the issue this PR will partially address: https://github.com/dapr/dapr/issues/2308

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/metadata/utils.go
+++ b/metadata/utils.go
@@ -18,6 +18,9 @@ const (
 	// TTLMetadataKey defines the metadata key for setting a time to live (in seconds)
 	TTLMetadataKey = "ttlInSeconds"
 
+	// RawPayloadKey defines the metadata key for forcing raw payload in pubsub
+	RawPayloadKey = "rawPayload"
+
 	// PriorityMetadataKey defines the metadata key for setting a priority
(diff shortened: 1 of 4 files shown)


metadata/utils.go
```diff
@@ -66,3 +69,17 @@ func TryGetPriority(props map[string]string) (uint8, bool, error) {
 
 	return 0, false, nil
 }
+
+// TryIsRawPayload determines if payload should be used as-is.
+func TryIsRawPayload(props map[string]string) (bool, error) {
```
yaron2, line 74: nit: Can you rename to `IsRawPayload` or even `RawPayload`? The `Try` is redundant here as.

yaron2 (changes requested):

metadata/utils.go
artursouza: I used the Try prefix to be consistent with the other methods since it can fail to parse it and return error.

yaron2 (approved):
lgtm

Issue #865, opened 2021-05-06
Dapr CLI commands take ~5 seconds to execute


jkotalik:
I'm trying to run dapr commands locally, and every command (even `dapr --version` and `dapr`) take 5 seconds to execute. 

https://user-images.githubusercontent.com/8302101/117346362-823a1180-ae5c-11eb-8bf6-825f221d54c7.mp4

## Steps to Reproduce the Problem

I can't repro this on WSL2, only on Windows. I tried uninstalling and reinstalling dapr which didn't help.

## Versions

CLI version: 1.1.0
Runtime version: 1.1.2

wcs1only:
With aerospike v2.7.0:
```
$ time dapr --version
CLI version: edge
Runtime version: 1.1.2

real    0m8.827s
user    0m0.000s
sys     0m0.015s
```

With aerospike v4.5.0:
```
$ time dapr --version
CLI version: edge
Runtime version: edge

real    0m0.531s
user    0m0.000s
sys     0m0.046s

```

PR #866, opened 2021-05-10
Update Aerospike to v4.5.0

wcs1only:
# Description

The Aerospike library was using microsecond precision sleep in a loop in its module init() function for random number generation. Windows, as it turns out, doesn't have microsecond precision sleep, so it was sleeping for for a total of nearly 8s. This is happening any time the daprd binary is invoked on windows. This has been fixed upstream, so the solution is to migrate to a newer version of this library.

With aerospike v2.7.0:
```
$ time dapr --version
CLI version: edge
Runtime version: 1.1.2

real    0m8.827s
user    0m0.000s
sys     0m0.015s
```

With aerospike v4.5.0:
```
$ time dapr --version
CLI version: edge
Runtime version: edge

real    0m0.531s
user    0m0.000s
sys     0m0.046s

```

## Issue reference

Please reference the issue this PR will close: #865

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/state/aerospike/aerospike.go
+++ b/state/aerospike/aerospike.go
@@ -162,9 +162,11 @@ func (aspike *Aerospike) Get(req *state.GetRequest) (*state.GetResponse, error) 
 
 	policy := &as.BasePolicy{}
 	if req.Options.Consistency == state.Strong {
-		policy.ConsistencyLevel = as.CONSISTENCY_ALL
+		policy.ReadModeAP = as.ReadModeAPAll
+		policy.ReadModeSC = as.ReadModeSCLinearize
 	} else {
-		policy.ConsistencyLevel = as.CONSISTENCY_ONE
+		policy.ReadModeAP = as.ReadModeAPOne
+		policy.ReadModeSC = as.ReadModeSCSession
 	}
 	record, err := aspike.client.Get(policy, asKey)
 	if err != nil {
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/DATA-DOG/go-sqlmock v1.5.0
 	github.com/Shopify/sarama v1.23.1
 	github.com/a8m/documentdb v1.2.1-0.20190920062420-efdd52fe0905
-	github.com/aerospike/aerospike-client-go v2.7.0+incompatible
+	github.com/aerospike/aerospike-client-go v4.5.0+incompatible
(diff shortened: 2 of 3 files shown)


state/aerospike/aerospike.go
```diff
@@ -162,9 +162,9 @@ func (aspike *Aerospike) Get(req *state.GetRequest) (*state.GetResponse, error)
 
 	policy := &as.BasePolicy{}
 	if req.Options.Consistency == state.Strong {
-		policy.ConsistencyLevel = as.CONSISTENCY_ALL
+		policy.ReadModeAP = as.ReadModeAPAll
```
mukundansundar, line 165: From the code: https://github.com/aerospike/aerospike-client-go/blob/v4.5.0/policy.go#L53
It seems there are two parameters to consider? Is that the scenario?
```
// ReadModeAP indicates read policy for AP (availability) namespaces.
ReadModeAP ReadModeAP //= ONE

// ReadModeSC indicates read policy for SC (strong consistency) namespaces.
ReadModeSC ReadModeSC //= SESSION;
```
The same file in v2.7.0 https://github.com/aerospike/aerospike-client-go/blob/v2.7.0/policy.go#L47
```
// How replicas should be consulted in a read operation to provide the desired
// consistency guarantee. Default to allowing one replica to be used in the
// read operation.
ConsistencyLevel ConsistencyLevel //= CONSISTENCY_ONE
```

wcs1only: I saw that, and I read:
https://github.com/aerospike/aerospike-client-go/blob/master/read_mode_sc.go

My interpretation was that SESSION is an ok default, since it means the clients only read from the master server in this case? If someone with more familiarity with the upstream service wants to make the case that this should be ReadModeSCLinearize, I can change it... This seems to me like it would only matter if this resource was being shared with non-dapr clients?

wcs1only: I did some more digging:
https://aerospike.com/blog/developers-understanding-aerospike-transactions/

I'm not sure we have the same definition of strong consistency that they do? For us strong consistency means that all replicas have committed a change before a write operation completes. For them, it seems like consistency is enforced on the read side? But the more I read, the more I think we probably want ReadModeSCLinearize, here.

Good call out, I will change this.

wcs1only: I wonder if @abhirockzz has any knowledge of this.

Issue #871, opened 2021-05-15
Add Feishu Webhook Binding


ipfans:
## Describe the proposal

A binding that allows reading / writing from [Feishu](https://www.feishu.cn) webhook. Feishu is a team collaborate toolkit developed by ByteDance.

PR #872, opened 2021-05-15
Add Feishu Webhook Binding

ipfans:
# Description

A binding that allows reading / writing from [Feishu](https://www.feishu.cn) webhook. Feishu is a team collaborate toolkit developed by ByteDance.

## Issue reference

Please reference the issue this PR will close: #871

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


bindings/feishu/webhook/webhook.go
```diff
@@ -0,0 +1,224 @@
...
+func (t *FeishuWebhook) Read(handler func(*bindings.ReadResponse) ([]byte, error)) error {
+	t.logger.Debugf("Feishu webhook: start read input binding")
+
+	webhooks.Lock()
+	defer webhooks.Unlock()
+	_, loaded := webhooks.m[t.settings.ID]
+	if loaded {
+		return fmt.Errorf("feishu webhook error: duplicate id %s", t.settings.ID)
+	}
+	webhooks.m[t.settings.ID] = &outgoingWebhook{handler: handler}
```
pkedy, line 96: It looks like `Read` (Input Binding method) must be called for a webhook to be created so that `Invoke` with `GetOperation` (Output Binding method) can be called to retrieve a message . I don't think this is how this is intended to work. Is there a reason why you set it up this way? Maybe we can recommend a different approach that fits into the model.

/cc: @artursouza @mukundansundar Thoughts?

artursouza: Any reason why this cannot go to Init()?

pkedy: Looks like this was borrowed from the DingTalk component and the same questions apply to that component.
https://github.com/dapr/components-contrib/blob/master/bindings/alicloud/dingtalk/webhook/webhook.go#L88-L101

ipfans: @pkedy @artursouza yes, it copied from DingTalk component, I will update this later

yaron2: @ipfans can you please update?

ipfans: @yaron2 @pkedy Sorry, I had almost forgot it.Here is the determination if there are duplicate Webhook ID, is there any alternative solution to achieve a similar goal?

yaron2:
ping @ipfans

PR #880, opened 2021-05-23
optimize(outbindings): OperationKind provide Equal method

1046102779:
reference pr: https://github.com/dapr/dapr/pull/3211


bindings/requests.go
```diff
@@ -22,3 +24,9 @@ const (
 	DeleteOperation OperationKind = "delete"
 	ListOperation   OperationKind = "list"
 )
+
+// Equal compare OperationKind
+func (o OperationKind) Equal(target OperationKind) bool {
```
mukundansundar, line 29: rename this to be EqualFold ?

1046102779: done.

pkedy: This is a very simple change, but I'm trying to understand the utility/impact of adding this function. It seems like many (if not most) of the components treat operation with case-sensitivity. If it is generally acceptable that operations to be case-sensitive (lowercase), I personally would rely on Go's `==` syntax instead of introducing a function like this. This makes that distinction unclear and we need to make a choice and provide guidance.

@1046102779 Are you just trying to add a function for only the scenarios where case-insensitivity is needed? If so, can you provide some examples. If you/we think that operations should be case-insensitive, then all of the existing components should be updated. Regardless, however and whenever this function would be used should be described in the doc comment.

pkedy: @1046102779 ping. Did we agree in another thread that it makes sense to keep the existing case sensitive approach? Just wondering if we can close this one.

PR #885, opened 2021-05-25
Added new output binding for Cosmos Graph DB - Gremlin

hokandil:
# Description
I added a new output binding to support graph DB in Azure Cosmos via the Gremlin API.
unfortunately there no issue reference for this PR , although I found it useful to share this with you.

* [ *] Code compiles correctly
* [ *] Created/updated tests


yaron2:
@beiwei30 can you please review?

hokandil:
@beiwei30 I fixed the issue causing the golangci-lint to fail.

bindings/azure/cosmosgraphdb/cosmosgraphdb.go
```diff
@@ -0,0 +1,130 @@
...
+	client   *gremcos.Cosmos
+	logger   logger.Logger
+}
+
+type cosmosGraphDBCredentials struct {
+	URL       string `json:"url"`
+	MasterKey string `json:"masterKey"`
+	Username  string `json:"username"`
+	// NumMaxActiveConnections string `json:"NumMaxActiveConnections"`
+	// ConnectionIdleTimeout   string `json:"ConnectionIdleTimeout"`
```
halspang, line 44: Do we not want to use these and give defaults if they aren't provided?

bindings/azure/cosmosgraphdb/cosmosgraphdb.go
```diff
@@ -0,0 +1,130 @@
...
+	if req == nil {
+		return nil, errors.New("invoke request required")
+	}
+
+	if req.Metadata == nil {
+		return nil, errors.New("metadata required")
+	}
+	// m.logger.Debugf("operation: %v", req.Operation)
+
+	gq, ok := req.Metadata[commandGremlinKey]
```
halspang, line 107: Just for my understanding, the goal here is to pass the entire Gremlin query through the metadata? Would it not be better to pass that through the request's body?

halspang (commented):
Couple sections of commented code, let's remove it if we don't need it or put it back if we do.

bindings/azure/cosmosgraphdb/cosmosgraphdb.go
hokandil: yes, @halspang you are right , some comments need to be removed along with these values (NumMaxActiveConnections, ConnectionIdleTimeout)  , I felt like it is over complicating the component YAML file , so I decided to remove it . any way the gremcos package assign there values to a default values which is more than enough for now.

bindings/azure/cosmosgraphdb/cosmosgraphdb.go
hokandil: Yes, the goal here is to pass the entire Gremlin query through the metadata. I saw the same with the MySQL binding so I decided to follow to be consistent. please let me knew if I need to change this to request body.

bindings/azure/cosmosgraphdb/cosmosgraphdb.go
hokandil: All unwanted comment removed.

yaron2:
@hokandil please resolve the conflicts and we'll assign a reviewer.

hokandil:
@yaron2 Done, conflicts resolved , thank you

yaron2:
@halspang can you please review this?

bindings/azure/cosmosgraphdb/cosmosgraphdb.go
```diff
@@ -0,0 +1,120 @@
...
+// NewCosmosGraphDB returns a new CosmosGraphDB instance
+func NewCosmosGraphDB(logger logger.Logger) *CosmosGraphDB {
+	return &CosmosGraphDB{logger: logger}
+}
+
+// Init performs CosmosDB connection parsing and connecting
+func (c *CosmosGraphDB) Init(metadata bindings.Metadata) error {
+	m, err := c.parseMetadata(metadata)
+	if err != nil {
+		return errors.New("failed to create the Cosmos Graph DB connector - metadata error")
```
halspang, line 54: Should we be eating the errors here? Will this make debugging more difficult?

bindings/azure/cosmosgraphdb/cosmosgraphdb.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	c.metadata = m
+	client, err := gremcos.New(c.metadata.URL,
+		gremcos.WithAuth(c.metadata.Username, c.metadata.MasterKey),
+	)
+	if err != nil {
+		return errors.New("failed to create the Cosmos Graph DB connector")
```
halspang, line 61: Same comment.

bindings/azure/cosmosgraphdb/cosmosgraphdb_test.go
```diff
@@ -0,0 +1,25 @@
...
+	"testing"
+
+	"github.com/dapr/components-contrib/bindings"
+	"github.com/dapr/kit/logger"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseMetadata(t *testing.T) {
+	m := bindings.Metadata{}
+	m.Properties = map[string]string{"Url": "a", "masterKey": "a", "username": "a"}
```
halspang, line 18: Nit: I'd prefer it if we had different values for each section that's parsed.

halspang (commented):
Changes look good to me, just some small last things.

@yaron2 is passing the query though metadata the standard? If so, what's the reasoning behind it? I'm fine with it if that's the case, just curious.

bindings/azure/cosmosgraphdb/cosmosgraphdb.go
hokandil: @halspang You are right , I resolved it.

bindings/azure/cosmosgraphdb/cosmosgraphdb.go
hokandil: @halspang  I am not sure, I just trying to be consistent with the other bindings code by following the same patterns. I saw this on the cosmosDB binding code. please let me knew if I need to remove.

yaron2: This is fine

bindings/azure/cosmosgraphdb/cosmosgraphdb.go
```diff
@@ -0,0 +1,122 @@
...
+			respOpKey:        string(req.Operation),
+			respGremlinKey:   gq,
+			respStartTimeKey: startTime.Format(time.RFC3339Nano),
+		},
+	}
+	d, err := c.client.Execute(gq)
+	if err != nil {
+		return nil, errors.New("error excuting gremlin")
+	}
+	resp.Data = d[0].Result.Data
```
yaron2, line 115: Can you check if there's an item 0 here? if there isn't, this will panic.

bindings/azure/cosmosgraphdb/cosmosgraphdb.go
yaron2: @hokandil please change this to not pass as metadata, but rather as a string from `req.Data`.

yaron2 (changes requested):

yaron2:
@hokandil I left two comments. fix those, and resolve the linter issue:

```
Check failure on line 53 in bindings/azure/cosmosgraphdb/cosmosgraphdb.go

GitHub Actions
/ Build linux_amd64 binaries
bindings/azure/cosmosgraphdb/cosmosgraphdb.go#L53
File is not `gofmt`-ed with `-s` (gofmt)
```

Then we can merge :)

hokandil:
@yaron2 I am on testing phase, I will commit shortly , maybe tomorrow .

hokandil:
@yaron2 Done

bindings/azure/cosmosgraphdb/cosmosgraphdb.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (c *CosmosGraphDB) Invoke(req *bindings.InvokeRequest) (*bindings.InvokeResponse, error) {
+	var jsonPoint map[string]interface{}
+	err := json.Unmarshal(req.Data, &jsonPoint)
+	if err != nil {
+		return nil, errors.New("CosmosGraphDB Error: Cannot convert request data")
+	}
+
+	gq := fmt.Sprintf("%s", jsonPoint[commandGremlinKey])
```
pkedy, line 98: Nit: It might more sense to unmarshal into an inline struct instead of `map[string]interface{}` since you know the field you expect.

```golang
type gremlinData struct {
  Gremlin string `json:"gremlin"` // Note sure if this is always a string, use `interface{}` otherwise.
}
var jsonPoint gremlinData
```

Also, using `%v` might make more since for an `interface{}` going in. But again, if the value is always going to be a `string` then the `fmt.Sprintf` is unnecessary.

```golang
gq := fmt.Sprintf("%v", jsonPoint[commandGremlinKey])
```

pkedy (commented):
Just pointing out a few potential improvements.

yaron2 (approved):

Issue #900, opened 2021-05-27
MQTT binding: Disconnect clients inside a Close method


pkedy:
## Expected Behavior

This component should implement a `Close() error` function to clean up the client instance (by calling `Disconnect`).

## Actual Behavior

The client is not closed.

## Release Note

RELEASE NOTE: MQTT binding: Disconnect added for closing the underlying client

PR #905, opened 2021-06-01
mqtt: implement Close

daixiang0:
Signed-off-by: Long Dai <long0dai@foxmail.com>

# Description

## Issue reference

Please reference the issue this PR will close: #900
## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/bindings/mqtt/mqtt.go
+++ b/bindings/mqtt/mqtt.go
@@ -313,7 +313,9 @@ func (m *MQTT) createClientOptions(uri *url.URL, clientID string) *mqtt.ClientOp
 }
 
 func (m *MQTT) Close() error {
-	m.consumer.Disconnect(1)
+	if m.consumer != nil {
+		m.consumer.Disconnect(1)
+	}
 	m.producer.Disconnect(1)
 
 	return nil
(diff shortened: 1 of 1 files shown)


bindings/mqtt/mqtt.go
```diff
@@ -311,3 +311,10 @@ func (m *MQTT) createClientOptions(uri *url.URL, clientID string) *mqtt.ClientOp
 
 	return opts
 }
+
+func (m *MQTT) Close() error {
+	m.consumer.Disconnect(1)
```
pkedy, line 316: I believe it is possible for `m.consumer` to be `nil` because it is initialized in `Read` and not `Init`. Because of that you will want an `if m.consumer != nil { m.consumer.Disconnect(1) }`.

pkedy (changes requested):
Thanks for picking this up.  I think there is a small tweak needed here.  After that it should be ready to approve.

bindings/mqtt/mqtt.go
daixiang0: Done.

yaron2 (approved):

Issue #923, opened 2021-06-07
Mysql should support more data types.


cvictory:
## Expected Behavior
In mysql, there are types as following:
```
// From : github.com/go-sql-driver/mysql@v1.5.0/fields.go
case fieldTypeBit:
		return "BIT"
	case fieldTypeBLOB:
		if mf.charSet != collations[binaryCollation] {
			return "TEXT"
		}
		return "BLOB"
	case fieldTypeDate:
		return "DATE"
	case fieldTypeDateTime:
		return "DATETIME"
	case fieldTypeDecimal:
		return "DECIMAL"
	case fieldTypeDouble:
		return "DOUBLE"
	case fieldTypeEnum:
		return "ENUM"
	case fieldTypeFloat:
		return "FLOAT"
	case fieldTypeGeometry:
		return "GEOMETRY"
	case fieldTypeInt24:
		return "MEDIUMINT"
	case fieldTypeJSON:
		return "JSON"
	case fieldTypeLong:
		return "INT"
	case fieldTypeLongBLOB:
		if mf.charSet != collations[binaryCollation] {
			return "LONGTEXT"
		}
		return "LONGBLOB"
	case fieldTypeLongLong:
		return "BIGINT"
	case fieldTypeMediumBLOB:
		if mf.charSet != collations[binaryCollation] {
			return "MEDIUMTEXT"
		}
		return "MEDIUMBLOB"
	case fieldTypeNewDate:
		return "DATE"
	case fieldTypeNewDecimal:
		return "DECIMAL"
	case fieldTypeNULL:
		return "NULL"
	case fieldTypeSet:
		return "SET"
	case fieldTypeShort:
		return "SMALLINT"
	case fieldTypeString:
		if mf.charSet == collations[binaryCollation] {
			return "BINARY"
		}
		return "CHAR"
	case fieldTypeTime:
		return "TIME"
	case fieldTypeTimestamp:
		return "TIMESTAMP"
	case fieldTypeTiny:
		return "TINYINT"
	case fieldTypeTinyBLOB:
		if mf.charSet != collations[binaryCollation] {
			return "TINYTEXT"
		}
		return "TINYBLOB"
	case fieldTypeVarChar:
		if mf.charSet == collations[binaryCollation] {
			return "VARBINARY"
		}
		return "VARCHAR"
	case fieldTypeVarString:
		if mf.charSet == collations[binaryCollation] {
			return "VARBINARY"
		}
		return "VARCHAR"
[...]

PR #926, opened 2021-06-07
fix: Mysql should support more data types. #923

beiwei30:
# Description

fix: Mysql should support more data types. #923

## Issue reference

Please reference the issue this PR will close: #923

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


bindings/mysql/mysql.go
```diff
@@ -285,84 +287,54 @@ func jsonify(rows *sql.Rows) ([]byte, error) {
...
+func convert(columnTypes []*sql.ColumnType, values []interface{}) map[string]interface{} {
+	r := map[string]interface{}{}
 
-			continue
+	for i, ct := range columnTypes {
+		value := values[i]
+
+		switch v := values[i].(type) {
+		case driver.Valuer:
+			if vv, err := v.Value(); err == nil {
```
pkedy, line 325: Should `err` be handled?   Feels like it should.

bindings/mysql/mysql.go
```diff
@@ -285,84 +287,54 @@ func jsonify(rows *sql.Rows) ([]byte, error) {
...
+				value = vv
```
pkedy, line 326: I'm unsure of this but I'm wondering if `driver.Value` should be converted back to `interface{}`. It's a minor detail but could matter for reflection.

```golang
value = interface{}(vv)
```

bindings/mysql/mysql.go
```diff
@@ -285,84 +287,54 @@ func jsonify(rows *sql.Rows) ([]byte, error) {
...
-		if s, ok := (scanArgs[i]).(*sql.NullTime); ok {
-			r[v.Name()] = s.Time
+func convert(columnTypes []*sql.ColumnType, values []interface{}) map[string]interface{} {
+	r := map[string]interface{}{}
 
-			continue
+	for i, ct := range columnTypes {
+		value := values[i]
+
+		switch v := values[i].(type) {
```
pkedy, line 323: How are null values handled? Is there a case needed for those to explicitly set the key to `nil`?

pkedy (commented):
Couple questions

bindings/mysql/mysql.go
beiwei30: fixed.

bindings/mysql/mysql.go
beiwei30: fixed.

bindings/mysql/mysql.go
beiwei30: I revised it to 

```go
		if value != nil {
			r[ct.Name()] = value
		}
```

beiwei30:
@pkedy Phil, would you pls. to review this change again?

beiwei30:
@pkedy any update on this?

yaron2 (approved):
lgtm

Issue #928, opened 2021-06-08
Why is the value of ConsumerID assigned to consumerGroup?


iamazy:
Hi, there

https://github.com/dapr/components-contrib/blob/464d53d94b259bd0ac2fc6342816c4dea5f3210f/pubsub/kafka/kafka.go#L50
https://github.com/dapr/components-contrib/blob/464d53d94b259bd0ac2fc6342816c4dea5f3210f/pubsub/kafka/kafka.go#L126

Why is the value of `kafkaMetadata`'s `ConsumerID` assigned to `Kafka`'s `consumerGroup`, I think they are not the same thing! It's better to change it to consumerGroup, i.e.

```diff
type kafkaMetadata struct {
	Brokers         []string `json:"brokers"`
--	ConsumerID      string   `json:"consumerID"`
++      ConsumerGroup   string   `json:"consumerGroup"`
	AuthRequired    bool     `json:"authRequired"`
	SaslUsername    string   `json:"saslUsername"`
	SaslPassword    string   `json:"saslPassword"`
	MaxMessageBytes int      `json:"maxMessageBytes"`
}
```

iamazy:
@khous

CodeMonkeyLeet:
@pkedy This seems confusing alright, though it's API breaking to change the name. Given that the Kafka pubsub component is in Beta, we should fix this sooner rather than later. 

On a related note, the field is not described in Dapr docs today, though it's _de facto_ a part of the public metadata. Given that the Kafka binding does [document the equivalent field](https://docs.dapr.io/reference/components-reference/supported-bindings/kafka/) as `consumerGroup`, we should probably standardize that.

pkedy:
@iamazy @CodeMonkeyLeet If I understand the history correctly, `consumerID` is used consistently across the other pubsub components in order to try make the concept generic. If more of the components are currently using `consumerID` (which I think they are), we probably want to standardize on that. Basically favoring consistency in the settings over the component level details (which is sort of what Dapr is all about).

Also another docs issue... `consumerID` is not included in the Kafka pubsub docs page
https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-apache-kafka/
The docs should mention that `consumerID` maps to Kafka's `consumerGroup`.

iamazy:
[...]

PR #965, opened 2021-06-18
rename ConsumerID to ConsumerGroup

iamazy:
# Description

In kafka pubsub component, rename kafkaMetadata.`ConsumerID` to kafkaMetadata.`ConsumerGroup`

## Issue reference

#928 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


pkedy:
@iamazy You need to update the config in the conformance test for the Kafka component. I also think it might make sense to make both `consumerID` and `consumerGroup` work, since Kafka is a prevalent pubs broker. If we can make a small change to prevent breaking users, we should IMO.

iamazy:
> I also think it might make sense to make both `consumerID` and `consumerGroup` work, since Kafka is a prevalent pubsub broker. If we can make a small change to prevent breaking users, we should IMO.

I am agree with you. But actually sarama don't expose place to set `ConsumerID`(or `MemberID` in this library), It only expose Member.UserData in `sarama.Config`.

iamazy:
Whether clientID should be exposed too？@pkedy

pkedy:
@iamazy Yes, I agree - otherwise it defaults to `sarama`. It could at least default to the AppID but I'm not sure that value/field is available to the component.

pkedy:
> I am agree with you. But actually sarama don't expose place to set ConsumerID(or MemberID in this library), It only expose Member.UserData in sarama.Config.

To clarify, I'm recommending `consumerID` be an alias to `consumerGroup` so current users do not need to worry about a config change.

iamazy:
@pkedy hi, I have set ConsumerID as alias of ConsumerGroup and add ClientID to support the other kafka compatible middleware, please check it. Thanks

pubsub/kafka/kafka.go
```diff
@@ -277,8 +290,20 @@ func (k *Kafka) Subscribe(req pubsub.SubscribeRequest, handler pubsub.Handler) e
 func (k *Kafka) getKafkaMetadata(metadata pubsub.Metadata) (*kafkaMetadata, error) {
 	meta := kafkaMetadata{}
 	// use the runtimeConfig.ID as the consumer group so that each dapr runtime creates its own consumergroup
-	meta.ConsumerID = metadata.Properties["consumerID"]
-	k.logger.Debugf("Using %s as ConsumerGroup name", meta.ConsumerID)
+	if val, ok := metadata.Properties["consumerID"]; ok && val != "" {
+		meta.ConsumerID = val
```
pkedy, line 294: Could you assign this to `meta.ConsumerGroup` instead and remove the need for `meta.ConsumerID`?

tests/config/pubsub/kafka/pubsub.yml
```diff
@@ -10,5 +10,7 @@ spec:
     value: localhost:9092
   - name: consumerID
```
pkedy, line 11: We can remove this meta property now. Correct?

pubsub/kafka/kafka.go
```diff
@@ -123,7 +127,12 @@ func (k *Kafka) Init(metadata pubsub.Metadata) error {
 
 	k.brokers = meta.Brokers
 	k.producer = p
-	k.consumerGroup = meta.ConsumerID
+	if meta.ConsumerID != "" {
```
pkedy, line 130: It feels like you could handle this case in `getKafkaMetadata` and could remove this given my comment below.

pkedy (changes requested):
@iamazy Thanks for doing all this and adding the alias. Just a few questions. We can get this approved quickly after this.

pubsub/kafka/kafka.go
iamazy: I misunderstood your original statement and I correct it now

tests/config/pubsub/kafka/pubsub.yml
iamazy: yes

pubsub/kafka/kafka.go
iamazy: done

pkedy (approved):
LGTM! Thanks for the contribution, @iamazy!

pkedy:
Ready for review @artursouza

pubsub/kafka/kafka.go
```diff
@@ -47,7 +47,8 @@ type Kafka struct {
 
 type kafkaMetadata struct {
 	Brokers         []string `json:"brokers"`
-	ConsumerID      string   `json:"consumerID"`
+	ConsumerGroup   string   `json:"consumerGroup"`
```
artursouza, line 50: Please, keep support for consumerID too to avoid breaking people that are using this. This component is not GA but Kafka is somewhat popular and backwards compatibility here is a small cost.

artursouza (changes requested):

artursouza (approved):
Ignore my comment above. I see it is backwards compatible. LGTM

pkedy (approved):

PR #967, opened 2021-06-21
fix: Dapr runtime panic when handle Pub/Sub (dapr#3281)

beiwei30:
# Description

fix issue dapr/dapr#3281 (Dapr runtime panic when handle Pub/Sub without CloudEvents status retry or drop #3281)

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/3281

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


pubsub/envelope.go
```diff
@@ -103,6 +103,27 @@ func FromCloudEvent(cloudEvent []byte, topic, pubsub, traceID string) (map[strin
 	m[TopicField] = topic
 	m[PubsubField] = pubsub
 
+	// specify default value if it's unspecified from the original CloudEvent
+	if m[IDField] == nil {
```
pkedy, line 107: dapr/dapr#3356 fixes the issue in the runtime. I would still have this change, but I think we should remove the `""` defaults for ID and Type and keep the rest. IMO `nil`/not set is better than an empty string when the field has really has no value.

pkedy (changes requested):
I like the defaults, but not empty string for ID and Type.

pubsub/envelope.go
beiwei30: @pkedy pls. take a look again.

beiwei30:
@pkedy any update on this?

yaron2:
Thanks for this @beiwei30

Issue #968, opened 2021-06-21
Added confluent kafka support for alibaba cloud kafka service


cshi0:
## Describe the proposal
The current kafka component is implemented by using sarama. However, it does not support the legacy versions used by alibaba cloud kafka service, and it is not tested against alibaba cloud's kafka service. Therefore, a new component is implemented by using confluent kafka sdk to support the kafka service on alibaba cloud.

Ref: dapr/docs#1582

cshi0:
@pkedy The version I was testing on is 2.2, and I cannot connect to brokers with ssl enabled. I have contacted the kafka team at Alibaba cloud, and they told me that sarama has compatibility problem with kafka 2.2. 

Or maybe the configuration of my client is incorrect, so I attached my testing codes here. Please tell me how I can fix the configuration to enable ssl. The error given is ` client has run out of available brokers to talk to (Is your cluster reachable?) `. It I can connect to the brokers with confluent kafka sdk with the same settings. Thank you!

```golang
package main

import (
	"crypto/tls"
	"crypto/x509"
	"fmt"
	"io/ioutil"
	"os"
	"strings"

	"github.com/Shopify/sarama"
)

type KafkaConsumerConfig struct {
	Brokers  []string
	Topic    string
	consumer sarama.Consumer
	client   sarama.Client
}

func main() {
	config := KafkaConsumerConfig{}
	config.Brokers = strings.Split(os.Getenv("KAFKA_ENDPOINTS"), ",")
	config.Topic = "dapr_test"

	consumerConfig := sarama.NewConfig()
	consumerConfig.Version = sarama.V2_2_0_0

	consumerConfig.Net.SASL.User = os.Getenv("KAFKA_SASL_USERNAME")
	consumerConfig.Net.SASL.Password = os.Getenv("KAFKA_SASL_PASSWORD")
	consumerConfig.Net.SASL.Handshake = true
	consumerConfig.Net.SASL.Enable = true
	consumerConfig.Net.SASL.Mechanism = sarama.SASLTypePlaintext

	consumerConfig.Net.TLS.Enable = true

	caCertPool := x509.NewCertPool()

	b, _ := ioutil.ReadFile(os.Getenv("KAFKA_CA_LOCATION"))
	caCertPool.AppendCertsFromPEM(b)

	tlsConfig := &tls.Config{
		RootCAs:    caCertPool,
		ClientAuth: tls.RequestClientCert,
	}
[...]

Issue #961, opened 2021-06-17
Add Alicloud MNS support for pubsub


cshi0:
## Describe the proposal
A pubsub component support Alibaba Cloud MNS service.

PR #969, opened 2021-06-21
added support for confluent kafka (on alibaba cloud)

cshi0:
# Description

Added a new kafka pub/sub component using confluent kafka sdk to support kafka service on alibaba cloud.

## Issue reference

Please reference the issue this PR will close: #968 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#1582


pkedy (dismissed):
Just putting a "Request changes" on this until you can try Sarama's `Version` per my comment here.

https://github.com/dapr/components-contrib/issues/968#issuecomment-865324505

pubsub/alicloud/confluentkafka/confluentkafka.go
```diff
@@ -0,0 +1,266 @@
...
+			return err
+		}
+		msg.TopicPartition.Offset = kafka.Offset(offset)
+	}
+
+	if keyStr, ok := md["key"]; ok {
+		msg.Key = []byte(keyStr)
+	}
+
+	// if timestampStr, ok := md["timestamp"]; ok {
```
pkedy, line 119: Remove dead/commented code

pubsub/alicloud/confluentkafka/confluentkafka.go
```diff
@@ -0,0 +1,266 @@
...
+	return &Kafka{ //nolint:exhaustivestruct
+		logger: logger,
+	}
+}
+
+func (k *Kafka) parseMetadata(props map[string]string) error {
+	k.ConfigMap = &kafka.ConfigMap{}
+
+	var metadata map[string]interface{}
+	if configJson, ok := props["configJson"]; ok {
```
pkedy, line 49: Is there a way to convey this using metadata? It might not be possible since metadata is a `map[string]string`. But if the configuration is flat, handle this differently.

pubsub/alicloud/confluentkafka/confluentkafka.go
```diff
@@ -0,0 +1,266 @@
...
+		err := json.Unmarshal([]byte(configJson), &metadata)
+		if err != nil {
+			return err
+		}
+
+		for key, val := range metadata{
+			if reflect.ValueOf(val).Kind() == reflect.Float64{
```
pkedy, line 56: Instead of using the `reflect` package, this would be simpler:

```golang
if v, ok := val.(float64); ok && v == math.Trunc(v) {
  val = int(v)
}
k.ConfigMap.SetKey(key, v))
```

pubsub/alicloud/confluentkafka/confluentkafka.go
```diff
@@ -0,0 +1,266 @@
...
+	if err != nil {
+		return err
+	}
+
+	k.producer, err = kafka.NewProducer(k.ConfigMap)
+	if err != nil {
+		return err
+	}
+
+	k.consumer, err = kafka.NewConsumer(k.ConfigMap)
```
pkedy, line 86: Should initializing the consumer be moved to `Subscribe`? (Perhaps surrounded by a `sync.Once`)?

pubsub/alicloud/confluentkafka/confluentkafka.go
```diff
@@ -0,0 +1,266 @@
...
+		partition, err := strconv.ParseInt(partitionStr, 10, 32)
+		if err != nil {
+			return err
+		}
+		msg.TopicPartition.Partition = int32(partition)
+	} else {
+		msg.TopicPartition.Partition = kafka.PartitionAny
+	}
+
+	if offsetStr, ok := md["offset"]; ok {
```
pkedy, line 107: I didn't think you could specify the offset when publishing a message...  Am I incorrect?

pubsub/alicloud/confluentkafka/confluentkafka.go
```diff
@@ -0,0 +1,266 @@
...
+	// 		if err != nil {
+	// 			return err
+	// 		}
+	// 		msg.TimestampType = kafka.TimestampType(timestampType)
+	// 	} else {
+	// 		return fmt.Errorf("a value associated with key timestampType must be provided if timestamp is provided")
+	// 	}
+	// }
+
+	if headerJson, ok := md["headers"]; ok {
```
pkedy, line 141: IMO encoding the messages as JSON seems like an unnecessary step...  wondering if it should any metadata fields that are not handled above (key, partition) or ones that have a name with a prefix of `header:`.

pkedy (commented):
A few questions.

pkedy:
@cshi0 I also think `alibaba` should be removed from the package/path because this is not really specific to Alibaba's service.

pubsub/alicloud/confluentkafka/confluentkafka.go
cshi0: deleted

pubsub/alicloud/confluentkafka/confluentkafka.go
cshi0: The values can have different types like integer and boolean, and the number of items in the config is really large. So, I cannot define a struct for it.

pubsub/alicloud/confluentkafka/confluentkafka.go
cshi0: modified

pubsub/alicloud/confluentkafka/confluentkafka.go
cshi0: When sync.Once is used for initializing consumer in Subscribe, the error might be missed if Subscribe is called multiple times, and it will be very confusing.

pubsub/alicloud/confluentkafka/confluentkafka.go
cshi0: You are correct. I have deleted this field.

pubsub/alicloud/confluentkafka/confluentkafka.go
cshi0: Changed to the prefix way

cshi0:
@pkedy I have moved it out of the alicloud folder to pubsub.

yaron2:
@pkedy please review this again and approve it ok.

yaron2:
@cshi0 please fix linter issues: https://github.com/dapr/components-contrib/pull/969/checks?check_run_id=3049654728

pkedy:
@cshi0 Still looks like there are some linter issues. After those are fixed, this can be approved quickly.

cshi0:
> @cshi0 Still looks like there are some linter issues. After those are fixed, this can be approved quickly.

I found the version of my linter is different from the one used in the ci, so I changed the version of my linter, and it should be ok now. Sorry about that.

yaron2:
@cshi0 tests are failing.

cshi0:
> @cshi0 tests are failing.

The tests are updated. Please run the workflow again, thanks.

cshi0:
@yaron2 Could you please rerun the workflow? Please also rerun the workflow for this one #961 please. Thanks!

pkedy:
The Windows tests need to work with the Kafka library before this can be merged.  :(

daixiang0:
ping @cshi0

cshi0:
> ping @cshi0

Thanks for reminding! I will fix it asap.

cshi0:
@yaron2 @daixiang0 

I just found out that confluent kafka go client does not support windows. I wonder whether it means it cannot be merged?

image

daixiang0:
I am not sure whether `// +build` help, you can try to build for windows expect this packge.

cshi0:
> I am not sure whether `// +build` help, you can try to build for windows expect this packge.

@daixiang0 I have tried `// +build !windows` and `// +build linux darwin`, but none of them works. I wonder what should be the correct tag if I want to exclude both building and testing on windows?

daixiang0:
See https://stackoverflow.com/questions/24030059/skip-some-tests-with-go-test.

cshi0:
> See https://stackoverflow.com/questions/24030059/skip-some-tests-with-go-test.

@daixiang0 Thank you! The build should be skipped for windows now.

pkedy:
Awesome, @cshi0.  Thanks for getting Windows to skip!

pkedy (dismissed):
LGTM

`skip_confluent_kafka` still feels strange but if we find a cleaner option, we can update how this works.  I think this will require devs to use `make test` instead of `go test ./...`. Perhaps not a big deal though.

artursouza:
@cshi0 Please, look at the latest changes to Makefile, it now has CGO disabled by default to enable Windows development. In this case, confluence kafka should probably be disabled based on that flag as well.

daixiang0:
ping @cshi0

cshi0:
> @cshi0 Please, look at the latest changes to Makefile, it now has CGO disabled by default to enable Windows development. In this case, confluence kafka should probably be disabled based on that flag as well.

@artursouza I'm really sorry for the late reply. If CGO is disabled, confluent kafka cannot be compile on other platform like linux.

cvictory:
@cshi0  Did you discussed with Alicloud developers who work on the Kafka? 

@artursouza If alicloud should use confluent-kafka, the cgo should be enable.

```
Prebuilt librdkafka binaries are included with the Go client and librdkafka does not need to be installed separately on the build or target system. The following platforms are supported by the prebuilt librdkafka binaries:

Mac OSX x64
glibc-based Linux x64 (e.g., RedHat, Debian, CentOS, Ubuntu, etc) - without GSSAPI/Kerberos support
musl-based Linux 64 (Alpine) - without GSSAPI/Kerberos support
When building your application for Alpine Linux (musl libc) you must pass -tags musl to go get, go build, etc.

CGO_ENABLED must NOT be set to 0 since the Go client is based on the C library librdkafka.

If GSSAPI/Kerberos authentication support is required you will need to install librdkafka separately, see the Installing librdkafka chapter below, and then build your Go application with -tags dynamic.
```
refer from : https://github.com/confluentinc/confluent-kafka-go/

pubsub/confluentkafka/confluentkafka.go
```diff
@@ -0,0 +1,233 @@
...
+	isNewTopic := true
+	for _, t := range k.topics {
+		if t == req.Topic {
+			isNewTopic = false
+
+			break
+		}
+	}
+	if isNewTopic {
+		k.topics = append(k.topics, req.Topic)
```
LaurenceLiZhixin, line 200: The changes to the topic list may not be  concurrent safe？

pubsub/confluentkafka/confluentkafka.go
```diff
@@ -0,0 +1,233 @@
...
+	} else {
+		msg.TopicPartition.Partition = kafka.PartitionAny
+	}
+
+	if keyStr, ok := md["key"]; ok {
+		msg.Key = []byte(keyStr)
+	}
+
+	for k, v := range md {
+		if strings.HasPrefix(k, "headers.") {
```
LaurenceLiZhixin, line 106: You can define a new constant named headerPrefix
```go
const (
    headerPrefix = "header." 
)

LaurenceLiZhixin (changes requested):

beiwei30:
@cshi0 would you mind to fix failures in CI?

beiwei30:
@artursouza What particular reason do you know why CGO is disabled by default? If CGO is disabled by default, then this component will not be disabled at all unless user configures to enable CGO and this component explicitly in order to use it.

beiwei30:
@yaron2 @CodeMonkeyLeet @artursouza Your comments on this?

yaron2:
ping @cshi0

PR #972, opened 2021-06-23
Updated to use common Azure auth logic

### LINKED ISSUE #970: Client Credentials authN & authZ is supported by all Dapr components having an Azure implementation
At the minimum, all dapr components having an implementation as an Azure service need to support authentication and authorization to Azure via the Client Credentials flow (not on behalf of, please) which is supported for most Azure services (i.e. external services from Dapr's point of view).   This feature may also apply to dapr components that implement services provided by other non-Azure providers via the Client Credential flow, but I use only Azure herein so I am not familiar with these.

Azure Client Credential flow involves the developer registering a daprized service (which needs to use some Azure service via a dapr component) with an Azure Active Directory tenant.  This gives the service a Service Principal which has a unique ClientId, plus either a ClientSecret or the key to a Certificate.  Then the Service Principal is registered by the developer with each Azure service that the daprized service needs to access. Plus, fine grained authorization permissions are also specified at this point for the daprized service as well.  This allows the running service to obtain a Client Credential access token from Azure Active Directory, and use that token (via the dapr component) to access the external Azure resource and perform the operations on the resource that the previously set fine grained authorization allows, all via the dapr feature being requested here.

Please note that **the above scenario applies to both Managed Identities** (system assigned and user assigned) **and non-managed identities as well**.  _Managed Identities_ are Azure located services, whose AAD identities are either assigned  by Azure (system assigned) or by a developer (user assigned). A _non-managed identity_ is some service that may or may not be located on Azure which has a Service Principle (as described above).  All these types of identities have Service Principals and thus the ability to get fine grained access authorization to many different Azure resources and services.  A main difference between Managed Identities and unmanaged identities is that Managed Identities always have a certificate which is managed and renewed automatically by Azure.  Unmanaged identities require the developer to furnish either a ClientSecret or certificate, plus renew it manually as well.

Long-story-short -- If dapr can support Managed Identities to allow a daprized service to access "external" Azure services, then (probably with little or no extra work) it can also support unmanaged identities!  

Here is the business value:  

1. Unmanaged identities are important to support since they allow services located on-prem or on-edge to use Azure AD as an identity provider and thus fully participate in the first class authentication Azure AD provides, adding another level to the recommended defense-in-depth strategy.  Note that this allows the use of Dapr to propagate far beyond the cloud and Kubernetes clusters, increasing the breadth of adoption of Dapr.

2. The Client Credentials flow via Azure Active Directory also allows a service to utilize the fine grained access authorization provided by most Azure  services.  Such fine grained access permissions often offer a radically smaller attack surface, compared to the previous generation of Shared Access Signatures. 

3.  Microsoft recommends that Shared Access Signatures now be used only if one wants permissions to automatically expire after a fairly short period of time.    For most services Microsoft now recommends that they use Service Principles to authenticate instead of Shared Access Signature, and also to use fine grained authorization that is available via Azure Active Directory for each Azure Service.  **Therefore, the key business value here is having the capability for daprized services to use Azure Best Practices for security, thereby reaping the great increase in protection they provide.**

For example, a daprized service needs to read from an Azure Service Bus Topic then save the result in an Azure Storage Blob.  In this case the daprized service needs a Service Principal obtained from registering the service with Azure Active Directory, or by getting it from a Managed Identity.  Then the developer needs to register the Service Principal with the Azure Service Bus Topic and Azure Storage Blob they intend to use so as to specify that the Service Principal has only permission to read from the Topic and write to the Blob.  

Then the developer will provide the information required to obtain a Client Credentials flow access token from Azure AD while the service is running to the Dapr Pubsub and State components it will use.  Perhaps this can be done via something like a SecretRef, call it an AccessTokenRef.  In other words, put the info required to obtain an access token in a AccessToken component that is referenced by both the pubsub and state components.  This is like a Secret in concept.  Note that an AccessToken component info (some of which is highly confidential) should also be stored in a SecretStore (most likely a Key Vault).

Once all the information is in the proper places the daprized service can run and access the Topic and Blob via dapr components as necessary without having to do anything in the developer-written-code.  However, please note that caching access tokens is absolutely necessary to provide excellent performance.  The Azure SDK provides "automatic" (or at least in a number of secnarios) access token caching.  Please see Azure Identity client library for .NET - Version 1.4.0 at https://docs.microsoft.com/en-us/dotnet/api/overview/azure/identity-readme for details on the most recent capabilities of AAD client SDK.  The dapr components need to implement token caching so as to make it transparent to the developer.

Note that another acceptable implementation would be to have a Dapr Access Token Utility service (app) that runs with its own Dapr sidecar, and which is utilized by all other dapr instances in a Dapr Mesh to retrieve the security access tokens their services (apps) need.  This further increases the separation of concerns by not mixing security token code with dapr component code.  Such a token utility might, in concept, be somewhat like the existing Actor Placement Service or Zipkin.  

This approach could also greatly leverage the use of existing cloud service provider SDKs in developing the code for the Access Token Utility and shortening development time.  This approach could also easily lead to using existing Security Token service code that is now in use in legacy scenarios being used in daprized services, thus providing the ability to bring some legacy services into the Dapr Mesh.  In short, such an approach would allow developer written Access Token Utility services to be written.  Such things may need a separate utility-contrib repository.

Thanks.  Please let me know if you have any  questions.

For background on this, plus links to related Microsoft documentation please see https://github.com/dapr/components-contrib/issues/938.

@artursouza -- This is the result of #938. Thanks for your assistance

## Release Note

RELEASE NOTE:

ItalyPaleAle:
# TLDR

This PR fixes #970 and it generally tries to improve the authentication logic for (eventually) all Azure services, making it consistent and allowing Azure AD + RBAC wherever possible. Using Azure AD allows greater security and simplified management, such as via MSI. In the process, this PR also adds support for sovereign Azure clouds (Gov, China, etc).

# Description

Currently, Dapr authenticates with Azure services primarily via "shared keys". For example, for Azure Storage it uses the "account key", which is not recommended as it's harder to manage the access token securely and it's less auditable. Quoting from the Azure Storage [documentation](https://docs.microsoft.com/en-us/azure/storage/common/storage-auth):

> Microsoft recommends using Azure AD when possible for maximum security and ease of use

This PR adds support for authenticating with Azure services through Azure AD (eventually, supporting all those who support this). This enables authentication via Azure AD "service principals", including client credentials (using a client secret), client certificate credentials (using a client certificate), and MSI. It then enables authorization via Azure RBAC, which can be fine-tuned by admins for better access control.

While I was implementing this, I also enabled the usage of other Azure clouds where possible (Azure China, Azure Gov, etc). For Azure Storage, the goal is to be able to support emulators too for simpler local development (like [Azurite](https://github.com/Azure/Azurite)).

The code is partly based on the authorizer file that already existed for Azure Key Vault; I have expanded it to support Azure Storage and other services, and I have made it possible to support sovereign clouds.

# Example

Example of using `state.azure.blobstorage` using client credentials:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: azblob
spec:
  type: state.azure.blobstorage
  version: v1
  metadata:
    - name: accountName
      value: xx
    - name: spnClientId
      value: xxx
    - name: spnClientSecret
      value: xxx
    - name: spnTenantId
      value: xxx
    - name: containerName
      value: xxx
```

# Naming consistency

The new authorizer uses key names `spn*` to be consistent with what the Azure Key Vault authorizer was doing before.

Within Dapr, other modules (such as `bindings/azure/eventgrid`) use `tenantId`, `clientId`, and `clientSecret`. We should likely standardize on one style (I personally have a preference for the Event Grid's one, omitting `spn` because that's an Azure-specific name that is unclear in the OAuth world). For backwards-compatibility, the "deprecated" names can be kept as aliases (such as `spnClientId` -> `clientId`)

# Work status

The PR is **not complete** yet. I am opening it as a draft for a review of the underlying authorization logic, and I will complete the support for the remaining modules then (unless someone wants to help :) )

Here's the list of work that's been done and still to do:

- [ ] `bindings/azure/blobstorage` - Uses keys
   - [ ] Documentation update
- [ ] `state/azure/cosmosdb`
   - 🟡 Made a PR to upstream dependency to add Azure AD auth support, which has been merged. Waiting for a new release of the SDK now. https://github.com/a8m/documentdb/pull/33
   - [ ] Documentation update
- [ ] `bindings/azure/eventgrid` - Currently supports client credentials auth but with its own logic
   - [ ] Documentation update
- [ ] `bindings/azure/eventhubs` - Uses keys
   - [ ] Documentation update
- [ ] `bindings/azure/servicebusqueues` - Uses "connection strings"
   - [ ] Documentation update
- [ ] `bindings/azure/signalr` - Uses "connection strings"
   - [ ] Documentation update
- [ ] `bindings/azure/storagequeues` - Uses keys
   - [ ] Documentation update
- [ ] `pubsub/azure/eventhubs` - Uses keys
   - [ ] Documentation update
- [ ] `pubsub/azure/servicebus` - Uses "connection strings"
   - [ ] Documentation update
- [X] `secretstores/azure/keyvault`
  - [ ] Documentation update
- [X] `state/azure/blobstorage`
   - [ ] Documentation update
- [ ] `state/azure/cosmosdb`
   - 🟡 Made a PR to upstream dependency to add Azure AD auth support, which has been merged. Waiting for a new release of the SDK now. https://github.com/a8m/documentdb/pull/33
   - [ ] Documentation update
- [ ] `state/azure/tablestorage` - Uses keys
   - 🛑 Blocked because Azure SDK doesn't support Azure AD authentication for Azure Storage Tables. The current SDK is deprecated and it's unlikely the feature will be added. A new SDK is under development. Tracked on https://github.com/Azure/azure-sdk-for-go/issues/7278
   - [ ] Documentation update


artursouza:
@msfussell @yaron2 FYI

secretstores/azure/keyvault/keyvault.go
```diff
@@ -56,6 +56,7 @@ func (k *keyvaultSecretStore) Init(metadata secretstores.Metadata) error {
 	}
 
 	k.vaultName = settings.Values[componentVaultName]
+	k.vaultDNSSuffix = settings.AzureEnvironment.KeyVaultDNSSuffix
```
artursouza, line 59: Please, keep the default behavior so it is backwards compatible.

state/azure/blobstorage/blobstorage.go
```diff
@@ -79,15 +79,25 @@ func (r *StateStore) Init(metadata state.Metadata) error {
...
+	var containerURL azblob.ContainerURL
+	customEndpoint, ok := metadata.Properties[endpointKey]
+	if ok && customEndpoint != "" {
+		URL, err := url.Parse(fmt.Sprintf("%s/%s/%s", customEndpoint, meta.accountName, meta.containerName))
+		if err != nil {
+			return err
+		}
+		containerURL = azblob.NewContainerURL(*URL, p)
+	} else {
+		URL, _ := url.Parse(fmt.Sprintf("https://%s.blob.%s/%s", meta.accountName, env.StorageEndpointSuffix, meta.containerName))
```
artursouza, line 98: Backwards compatibility.

state/azure/blobstorage/blobstorage.go
```diff
@@ -169,16 +179,10 @@ func getBlobStorageMetadata(metadata map[string]string) (*blobStorageMetadata, e
 		return nil, fmt.Errorf("missing or empty %s field from metadata", accountNameKey)
 	}
 
-	if val, ok := metadata[accountKeyKey]; ok && val != "" {
```
artursouza, line 172: This authentication method should still be supported.

artursouza (commented):
Overall is on a good path. I would ask to keep backwards compatibility and also keep support for the existing (although non-recommended) authentication via secret key and connection strings. This can be handy for local development and also to give existing apps some room to migrate.

state/azure/blobstorage/blobstorage.go
ItalyPaleAle: This is backwards-compatible because the default environment is for Azure public cloud, so the suffix is `core.windows.net`.

secretstores/azure/keyvault/keyvault.go
ItalyPaleAle: Yes this will be backwards-compatible. Unless someone manually opts in to use a different Azure cloud, this will match the same URL as before.

state/azure/blobstorage/blobstorage.go
ItalyPaleAle: Yes, i will ensure this is added back

ItalyPaleAle:
Thanks for the first round of reviews. I'll continue iterating on this.

Understood on maintaining support for the "legacy" out, agree it should be preserved.

daixiang0:
Could move task list to issue, and separate PR as small as possible, it would help track and review.

ItalyPaleAle:
@daixiang0 I think this would make sense. I could trim down this PR to only what's implemented today (a few things need to be updated/fixed, I'll get to that soon - just need to find some time :) ).

The other component updates could be moved to a separate issue, and we could open a PR for each.

daixiang0:
thanks for your contributions! please fix the conflict.

berndverst:
@ItalyPaleAle I think this contribution is great and exactly what we need.

I like that you centralized and standardized the Azure Auth. Also, I approve of the KV related changes (I was working on those independently myself).

state/azure/blobstorage/blobstorage.go
berndverst: It wasn't removed -- just moved to `authentication/azure/storage`

berndverst:
@artursouza feel free to review this now. I think this is in a state where we can merge it. Right now this only updates behaviors for State Store from Azure Storage and Secret Store with Keyvault. It does however generally put the common Azure Auth logic in place. We can in subsequent PRs switch the other Azure components to use this logic.

This change here is also fully backwards compatible.

I introduced new Environment variables as alternatives to the old `spn` prefixes ones. Both will work. This is to ensure consistency with the official Azure Go SDK (and other Azure auth libraries).

yaron2:
> @artursouza feel free to review this now. I think this is in a state where we can merge it. Right now this only updates behaviors for State Store from Azure Storage and Secret Store with Keyvault. It does however generally put the common Azure Auth logic in place. We can in subsequent PRs switch the other Azure components to use this logic.
> 
> This change here is also fully backwards compatible.
> 
> I introduced new Environment variables as alternatives to the old `spn` prefixes ones. Both will work. This is to ensure consistency with the official Azure Go SDK (and other Azure auth libraries).

It's important to make sure this is backward compatible. I'm happy with this change.

yaron2:
Workflow is running now.

ItalyPaleAle:
@yaron2 I'm taking one more pass to review this PR and also the following changes. Working synchronously with @berndverst as I'm writing this. We should be done soon.

berndverst:
Given that the various settings are specified as metadata in the component YAMLs we decided to name them in a more YAML standard way with camelcase. The naming is still derived from the naming in the other Azure SDKs, but is camelcase with an `azure` prefix to disambiguate.

ItalyPaleAle:
@yaron2  we're done. The last changes include:

- Cleaning up the name of the keys that can be used in the metadata. As per my "Naming consistency" point above, some components were using naming like "spnClientId", some were using "azureClientId", and some "clientId". This now supports all three, although the preferred one should be "azureClientId" (and similar).
- MSI should be fixed now.

I think this is ready for review at this point. Some notes:

1. Backwards compatibility should have been maintained at all times, for all components. Everything that works today will work with this new PR. if that's not the case, please let me know as it should be considered a bug.
2. As discussed, this was scoped down for now and it only updates `secretstores/azure/keyvault` and `state/azure/blobstorage`. The common auth layer is now generic enough that we should be able to leverage it for every other component (excluding those that use Table Storage because of lack of upstream support), but that work will need to happen in separate PRs.

authentication/azure/storage.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	// Fallback to using Azure AD
+	spt, err := settings.GetServicePrincipalToken()
+	if err != nil {
+		return nil, nil, err
+	}
+	var tokenRefresher azblob.TokenRefresher = func(credential azblob.TokenCredential) time.Duration {
+		log.Debug("Refreshing Azure Storage auth token")
+		fmt.Println("Refreshing Azure Storage auth token")
```
ItalyPaleAle, line 43: I just noticed there are 2 log statements here. Perhaps only one of the two should be kept (or neither). Same on lines 53-54. Not sure what the standard for logging in Dapr is, maybe you could decide how to fix this please?

yaron2: We shouldn't use fmt.Println

berndverst: I already removed it before I saw your comment @yaron2 :)

berndverst:
@artursouza @yaron2 can you please take a look and review? :)

state/azure/blobstorage/blobstorage_test.go
```diff
@@ -50,13 +50,11 @@ func TestGetBlobStorageMetaData(t *testing.T) {
 	t.Run("All parameters passed and parsed", func(t *testing.T) {
 		m := make(map[string]string)
 		m["accountName"] = "acc"
-		m["accountKey"] = "key"
```
artursouza, line 53: Why `key` is not parsed anymore?

artursouza (changes requested):
Just a comment on a test that might affect backwards compatibility.

state/azure/blobstorage/blobstorage_test.go
ItalyPaleAle: The `accountKey` property doesn't exist anymore in the object (see line 59 deleted in the test below).

Instead, the shared key is read in the method that returns the `azblob.Credential` object: https://github.com/dapr/components-contrib/blob/b41101713896855d113db2425d4b5163d633e8c5/authentication/azure/storage.go#L26

artursouza (approved):

Issue #306, opened 2020-03-23
Add optional support for TTL (Time to Live) for state stores


wanton7:
## In what area(s)?

> /area runtime

## Describe the feature

If I have example a Redis state store that I want to use as cache, having support for TTL when writing a state would be very beneficial. State stores that don't support TTL could just ignore TTL.

denis111:
+1. Would be very useful for storing session tokens (like allow (for logged in) or deny (for logged-out) users) even for things like stateless jwt tokens for additional security.

yaron2:
State API calls take in an optional metadata per request.

Does something like the following seem reasonable?

```
/v1.0/state/redis

[
  {
     "key": "key1",
     "value: "myVal",
     "metadata": {
         "ttl": "1d"
     }
  }
]
```

denis111:
@yaron2 I guess it would be fine.
Right now we use Spring Boot's redis client and it has @TimeToLive annotation wich you put on long property in the class you want to save in redis, maybe something similar could be made in dapr's java sdk, if not then it's not a problem setting metadata property manually.

berndverst:
We'll use `ttlInSeconds` in metadata, just like for PubSub

yaron2:
Changing to P2.

berndverst:
Just a quick note:
In addition to Redis and CosmosDB I'll be adding MongoDB and Cassandra. Maybe Couchbase if desired or I will let someone else contribute that change :)

UmutComlekci:
With references of your Redis & Cosmos DB TTL developments, I wrote Memcached TTL support  :)

berndverst:
Looking into this more I am not sure it makes sense to support TTL for MongoDB here. This would require a custom field (typically a date field) and an index on the entire collection specifying this field to be used for expiration.

I'd argue that collection / container / bucket level TTLs are out of scope for this change. (Perhaps they could be supported at a later time via runtime config settings.)

PR #990, opened 2021-07-02
Adds support for TTL in Redis State Store

berndverst:
# Description

Adds support for TTL in Redis State Store

## Issue reference

#306 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


berndverst:
@pkedy could you do an initial pass on this PR please?

berndverst:
References:
https://redis.io/commands/expire
https://redis.io/commands/persist

state/redis/redis.go
```diff
@@ -241,6 +243,19 @@ func (r *StateStore) setValue(req *state.SetRequest) error {
 
 		return fmt.Errorf("failed to set key %s: %s", req.Key, err)
 	}
+	switch {
+	case ttl == -1:
+		_, err = r.client.Do(r.ctx, "PERSIST", req.Key).Result()
```
artursouza, line 248: Do we even need to do anything when ttl is not set?

state/redis/redis.go
```diff
@@ -318,6 +340,19 @@ func (r *StateStore) parseETag(req *state.SetRequest) (int, error) {
 	return ver, nil
 }
 
+func (r *StateStore) parseTTL(req *state.SetRequest) int {
+	if val, ok := req.Metadata[ttlInSeconds]; ok && val != "" {
+		parsedVal, err := strconv.ParseInt(val, defaultBase, defaultBitSize)
+		if err != nil {
+			return 0
```
artursouza, line 347: Please, return error and fail the request. Swallowing an error and returning a valid number can hide a problem for in the application code.

artursouza (changes requested):

state/redis/redis.go
```diff
@@ -318,6 +346,19 @@ func (r *StateStore) parseETag(req *state.SetRequest) (int, error) {
 	return ver, nil
 }
 
+func (r *StateStore) parseTTL(req *state.SetRequest) (int, error) {
```
artursouza, line 349: Please, send a new PR later centralizing the parseTTL code since it is duplicated in many state stores.

state/redis/redis.go
artursouza: Please, make sure this is in the documentation.

artursouza (dismissed):

state/redis/redis.go
berndverst: Sure!

berndverst:
Tests pass locally. The cycle complexity shouldn't trigger as `.golanci.yml` specifies a minimum of 10, yet it seems that our Github workflow doesn't read this config file and alerts on anything.

artursouza (approved):

artursouza:
@berndverst I made the adjustment for lint to pass. Let me know how you feel about these.

berndverst:
The irony is that these changes are computationally slower than what I had due to more if statements that need to be evaluated.

Of course it's fine for this PR but I think we should change the linger configuration. I like the original code better.

Issue #994, opened 2021-07-05
Dapr state store component is not handling FIRST_WRITE concurrency option correctly.


prtk-g:
## Expected Behavior

If there are two concurrent requests to save state for the same key, with concurrency option as `FIRST_WRITE`, and etag omitted or null, only one of the requests should succeed and the other should fail with an exception. 

## Actual Behavior

One of the requests overrides the other and there is a lost update.

## Steps to Reproduce the Problem

1. Start a Redis database server
```bash
docker run -d -p 8001:6379 redis
```
2. Create a dapr state component file
```yml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: redis
spec:
  type: state.redis
  version: v1
  metadata:
  - name: redisHost
    value: localhost:8001
  - name: redisPassword
    value: ""
```

3.  Run Dapr with state component file in the current directory
```bash
dapr run  --components-path . --dapr-http-port 3500
```
4. Send 1st request to save value for a key, with concurrency option as `FIRST_WRITE`
```bash
curl --location --request POST 'http://localhost:3500/v1.0/state/redis' \
--header 'Content-Type: application/json' \
--data-raw '[
    {
        "key": "1",
        "value": "value-1",
        "options": {
            "concurrency": "first-write",
            "consistency": "strong"
        }
    }
]'
```
5. Send 2nd request to save a different value for the same key, again with concurrency option as `FIRST_WRITE`. 
```bash
curl --location --request POST 'http://localhost:3500/v1.0/state/redis' \
--header 'Content-Type: application/json' \
--data-raw '[
    {
        "key": "1",
        "value": "value-2",
        "options": {
            "concurrency": "first-write",
            "consistency": "strong"
        }
    }
]'
```
6. Send request to get the state for saved key 
```bash
curl --location --request GET 'http://localhost:3500/v1.0/state/redis/1' \
--data-raw ''
```
Result: 
```bash
"value-2"
```
## Note
As per discussion on [this](https://discord.com/channels/778680217417809931/778749797242765342/860418694773735444) post on discord, if the etag value is null or is omitted, dapr handles the request in last write wins pattern. But since this may result in some complexity for saving state for a non-existing key, `FIRST_WRITE` option should be given priority over etag's null value, for handling the concurrency for the request.

## Release Note

RELEASE NOTE: **FIX** 

RELEASE NOTE:

artursouza:
Please, add conformance test to validate/fix the same in other state stores.

[...]

PR #995, opened 2021-07-06
state: correct FIRST_WRITE

daixiang0:
Signed-off-by: Long Dai <long0dai@foxmail.com>

# Description

As per discussion, if the etag value is null or is omitted, dapr handles the request in last write wins pattern. But since this may result in some complexity for saving state for a non-existing key, FIRST_WRITE option should be given priority over etag's null value, for handling the concurrency for the request.

## Issue reference

Please reference the issue this PR will close: #994

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


daixiang0:
ping @artursouza @yaron2

daixiang0:
@pkedy @artursouza code finish, please help check CI fail, not sure it is env issue since no changes about it, thanks.

state/mongodb/mongodb.go
```diff
@@ -179,6 +179,8 @@ func (m *MongoDB) setInternal(ctx context.Context, req *state.SetRequest) error
 	filter := bson.M{id: req.Key}
 	if req.ETag != nil {
 		filter[etag] = *req.ETag
+	} else if req.Options.Concurrency == state.FirstWrite {
+		filter[etag] = uuid.NewString()
```
artursouza, line 183: Was this tested? Why matching eTag on a random UUID fixes the firstWrite wins?

tests/conformance/state/state.go
```diff
@@ -474,4 +474,72 @@ func ConformanceTests(t *testing.T, props map[string]string, statestore state.St
 		features := statestore.Features()
 		assert.False(t, state.FeatureETag.IsPresent(features))
 	}
+
+	if config.HasOperation("first-write") {
```
artursouza, line 478: This is great! Thanks. BTW, can you add a scenario with first-write and an etag to update an item? Something like:

+ Write an item
+ Read item and get eTag (eTag1)
+ Set with FIRST_WRITE and eTag as eTag1
+ Read item again and check that eTag (eTag2 != eTag1) and value changes
+ Set again with FIRST_WRITE and eTag as eTag1
+ Check the set above failed.

artursouza (changes requested):

state/mongodb/mongodb.go
daixiang0: Tested, if set etag when firstWrite, then the second write would raise a error.

tests/conformance/state/state.go
daixiang0: Done.

daixiang0:
ping @artursouza

artursouza (approved):

PR #996, opened 2021-07-06
Adds support for TTL in Cassandra State Store

berndverst:
# Description

Adds support for TTL in Cassandra State Store

## Issue reference

#306 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


berndverst:
@artursouza please take a look.

state/cassandra/cassandra.go
```diff
@@ -286,7 +287,12 @@ func (c *Cassandra) Set(req *state.SetRequest) error {
 		session = sess
 	}
 
-	return session.Query("INSERT INTO ? (key, value) VALUES (?, ?)", c.table, req.Key, bt).Exec()
+	ttl, err := parseTTL(req.Metadata)
+	if err != nil {
+		return fmt.Errorf("error parsing TTL from Metadata: %s", err)
+	}
+
+	return session.Query("INSERT INTO ? (key, value) VALUES (?, ?) USING TTL ?", c.table, req.Key, bt, ttl).Exec()
```
artursouza, line 295: `0` is the value parsed if ttl is not present - default behavior. So, what is the behavior for `0` TTL in cassandra? Also, is it also in seconds?

artursouza (changes requested):

state/cassandra/cassandra.go
berndverst: Yes it is seconds. And Cassandra says that 0 or any negative number means to persist the record. https://docs.datastax.com/en/cql-oss/3.3/cql/cql_using/useExpireExample.html

berndverst: Thanks for catching this. This was only correct when a TTL could not be set at the table level. I updated it now. Someone wishing to explicitly to persist an item needs to specify 0 or any negative number. It's completely passthrough now.

state/cassandra/cassandra.go
```diff
@@ -303,3 +313,19 @@ func (c *Cassandra) createSession(consistency gocql.Consistency) (*gocql.Session
 
 	return session, nil
 }
+
+func parseTTL(requestMetadata map[string]string) (*int, error) {
+	if val, found := requestMetadata[metadataTTLKey]; found && val != "" {
+		parsedVal, err := strconv.ParseInt(val, 10, 0)
+		if err != nil {
+			zero := 0
```
artursouza, line 321: You can just return nil, err since the value is meaningless for error.

artursouza (changes requested):
Just a minor comment.

state/cassandra/cassandra.go
artursouza: I fixed this small thing.

artursouza (approved):

artursouza:
@berndverst I made the change to return nil instead of 0 for the case of an error. Let me know what you think.

berndverst:
Makes sense. I think that was a remnant of an earlier version I had. Thanks for updating it.

Issue #683, opened 2021-02-10
Support the retrieval of blob metadata for the Azure Blob Storage binding


akkie:
The blob storage binding currently has no option to get the metadata of a blob. The binding currently has the option to store metadata, but there is no way to retrieve it. It would be great if this could be supported.

https://docs.microsoft.com/en-us/rest/api/storageservices/get-blob-metadata

Issue #682, opened 2021-02-10
Support blob listing for the Azure Blob Storage binding


akkie:
The blob storage binding currently has no option to list blobs in a container. It would be great if this could be supported.

https://docs.microsoft.com/en-us/rest/api/storageservices/list-blobs

PR #999, opened 2021-07-06
Return blob metadata and add list operation

akkie:
# Description

This PR adds the possibility to return the user defined blob metadata and it returns the option to list blobs from a container.

I've started first to implement one PR for each feature, but during development of the second feature it turned out that code is shared between the both features. It therefore made sense to implement both features into one PR. I hope that is ok?

Fixes: #683
Fixes: #682

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#1623


.devcontainer/devcontainer.json
```diff
@@ -12,7 +12,7 @@
 		// "BIND_LOCALHOST_DOCKER": "true"
 	},
 	"extensions": [
-		"ms-vscode.go",
+		"golang.go",
```
artursouza, line 15: Why this change?

bindings/azure/blobstorage/blobstorage.go
```diff
@@ -201,26 +231,113 @@ func (a *AzureBlobStorage) delete(blobURL azblob.BlockBlobURL, req *bindings.Inv
...
 	default:
 		return nil, fmt.Errorf("unsupported operation %s", req.Operation)
 	}
 }
+
+func (a *AzureBlobStorage) getBlobURL(metadata map[string]string) azblob.BlockBlobURL {
+	name := ""
+	if val, ok := metadata[blobName]; ok && val != "" {
+		name = val
+		delete(metadata, blobName)
```
artursouza, line 335: It is odd that this method modifies the map. Any reason?

artursouza (changes requested):

artursouza:
> @pkedy @artursouza There is some inconsistency in the blob storage binding:
> 
> * The form of the metadata keys is inconsistent. For the `blobName` the camel case form is used and the others use pascal case. When I look at the other bindings, camel case is mainly used. Is there a common rule?
camel case should be used. As of now, this component is in Alpha and not GA, which means it can have breaking change. I would add support to camel case for the ones using pascal case to give consistency but keep backwards compatibility (alpha does not require backwards compatibility but it is nice to keep IMO). Then, add a TODO to remove the pascal case support when the component moves to GA. Adding a new version would be overkill now IMO.
image


> * Is there a rule when metadata and when data should be used? For the create option it makes sense to send the `blobName` in the metadata, because the content is used for the data. But for all other options, the data could be used instead of the metadata. I have the list option also implemented with metadata, but I think it makes more sense to send the data as JSON in the data field.
> I like blobName in the metadata for all method for consistency.



> What do you think about that? Maybe it makes sense to create a new version of the binding and clean the things up?
Because this is not in GA yet, it can have breaking changes. Like I said, I would try to bring consistency without breaking changes as much as possible and add a TODO to remove the "old way" when moving it to GA.

.devcontainer/devcontainer.json
akkie: The ms-vscode extension doesn't exists anymore. The ms-vscode repo (https://github.com/microsoft/vscode-go) was archived a while ago with a note that https://github.com/golang/vscode-go is now the new home.

bindings/azure/blobstorage/blobstorage.go
akkie: I have moved that part to the create function because this is the only method which forwards the metadata to Azure to store it with the blob

akkie:
I followed your suggestion to make it backward compatible and use metadata where the blob name is involved. This means that the create, get and delete options will use metadata tu submit data and the list option will use the data property. I need to update the documentation, but the PR can be reviewed.

bindings/azure/blobstorage/blobstorage.go
```diff
@@ -22,16 +23,32 @@ import (
...
 const (
 	blobName                 = "blobName"
-	contentType              = "ContentType"
-	contentMD5               = "ContentMD5"
-	contentEncoding          = "ContentEncoding"
-	contentLanguage          = "ContentLanguage"
-	contentDisposition       = "ContentDisposition"
-	cacheControl             = "CacheControl"
-	deleteSnapshotOptions    = "DeleteSnapshotOptions"
+	marker                   = "marker"
```
artursouza, line 26: a constant that just repeats the content is not so useful. Please, add more context in the name for improved readability. This feedback is for the other constants too. I know it was already like that but this is an opportunity for us to raise the bar here.

artursouza (changes requested):
LGTM. Just a small readability improvement.

bindings/azure/blobstorage/blobstorage.go
akkie: Do you have an example?

.devcontainer/devcontainer.json
daixiang0: I think this change should better place in other PR, the choice about recommended extensions need more discuss.

akkie: @daixiang0 I have rolled back the change and will create a new pull request

bindings/azure/blobstorage/blobstorage.go
akkie: I have added a prefix `metadataKey` for each constant. Hope this is OK?

daixiang0: I think it would be great to add comment where define them and those meanings.

akkie: @daixiang0 Do you mean for each or only for the group?

akkie: @daixiang0 Just to be sure. You suggest to add a comment for each constant where the meaning is described?

```go
const (
	// Defines the name of the blob
	metadataKeyBlobName = "blobName"
	// Defines a string value that identifies the portion of the list to be returned with the next list operation.
	// The operation returns a marker value within the response body if the list returned was not complete. The marker
	// value may then be used in a subsequent call to request the next set of list items.
	metadataKeyMarker             = "marker"
        ...
)
```

daixiang0: Is there any doc for those?

>  Defines the name of the blob
Comments like this make no sense.

akkie: I have also updated the doc: https://github.com/dapr/docs/pull/1623/files

Please don't get me wrong. But could you please be more precise with the changes you suggest? I think otherwise we are going around in circles here

daixiang0: Is there a doc for "blobName", "marker" and so on? Add link here. If not, add comment for each if necessary.

akkie: Is this OK now?

yaron2:
@akkie please run `go mod tidy` and push the changes. we can then merge.

yaron2 (approved):

Issue #1017, opened 2021-07-18
Add priority option in SMTP binding. It will allow select email priority


fjvela:
## Describe the feature
Add priority option in SMTP binding in order to select the email priority (values: highest[1] to lowest[5])

https://www.chilkatsoft.com/p/p_471.asp

## Release Note

RELEASE NOTE: **ADD** SMTP binding added priority option

PR #1018, opened 2021-07-18
Add priority option in SMTP binding. It will allow select email priority

fjvela:
# Description

Add priority option in SMTP binding in order to select the email priority (values: highest[1] to lowest[5])

https://www.chilkatsoft.com/p/p_471.asp

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/components-contrib/issues/1017
## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_1665_

image


bindings/smtp/smtp.go
```diff
@@ -132,12 +141,13 @@ func (s *Mailer) parseMetadata(meta bindings.Metadata) (Metadata, error) {
 	smtpMeta.EmailBCC = meta.Properties["emailBCC"]
 	smtpMeta.EmailFrom = meta.Properties["emailFrom"]
 	smtpMeta.Subject = meta.Properties["subject"]
+	smtpMeta.Priority = defaultPriority
```
artursouza, line 144: Please, add a priority metadata support too, so it can be set at the configuration level as well. If not present, then default to this. Thanks for your contribution.

artursouza (commented):
LGTM to be overall.

bindings/smtp/smtp.go
```diff
@@ -16,6 +16,10 @@ import (
 	"gopkg.in/gomail.v2"
 )
 
+const (
+	defaultPriority = 3
```
daixiang0, line 20: Should we mark all priority as const?

bindings/smtp/smtp.go
```diff
@@ -160,5 +174,29 @@ func (metadata Metadata) mergeWithRequestMetadata(req *bindings.InvokeRequest) M
...
+
+func (metadata *Metadata) parsePriority(req string) error {
+	if req == "" {
+		metadata.Priority = defaultPriority
+	} else {
+		priority, err := strconv.Atoi(req)
+		if err != nil {
+			return err
+		}
+		if priority < 1 || priority > 5 {
```
daixiang0, line 195: could add comment for priority?

bindings/smtp/smtp.go
fjvela: I think it isn't needed.  I added lowest and highest priority as const in order to be more clear.

bindings/smtp/smtp.go
fjvela: I added lowest and highest priority as const in order to be more clear.

daixiang0: I am not familiar with SMTP priority, please share a doc here or comment in the code.

fjvela: https://www.chilkatsoft.com/p/p_471.asp

artursouza (approved):

PR #1025, opened 2021-07-20
Add invoke middleware that support invoking other service by middleware in middleware call chain

li1234yun:
# Description

In some case, we want to invoke other service to do something before calling real invoked service, we can use this middleware to achieve this goal.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#1680


daixiang0:
Is there any proposal for this? I do not find how users can use it.

li1234yun:
> Is there any proposal for this? I do not find how users can use it.



> Is there any proposal for this? I do not find how users can use it.

Hi,

I don't know the middleware need the proposal. If need, I will doc it. I don't provide the middleware doc, because I just want to know if the middleware is acceptable. Maybe you can view it quickly, if you think it is fine, I will doc the related doc and push it.

li1234yun:
This middleware is invoke other service middleware. When we want to invoke middleware method in middleware call chain, we can use it. A common scenario is that invoke authorization service to authorize the permission before we invoke real service method.

msfussell:
> This middleware is invoke other service middleware. When we want to invoke other service method in middleware call chain, we can use it. A common scenario is that invoke authorization service to authorize the current invoke permission before we invoke real service method.

Can you explain your scenario in some more detail? We have not really seen this sort of ask before, so it would be interesting to understand the details of the scenario you are trying to achieve.

li1234yun:
Hello, @msfussell 

> Can you explain your scenario in some more detail? We have not really seen this sort of ask before, so it would be interesting to understand the details of the scenario you are trying to achieve.

I will illustrate my current scenario, maybe you can provide better solution, because I think my solution is not good. My solution just use the service invocation concept in the middleware layer, maybe it solved in dapr runtime will be better.

image

As shown in the figure above, I want to invoke authorization service to judge current invoke permission before I invoke target service's method. If pass, I will continue to invoke the target service's method, otherwise the request will be denied.

artursouza:
What are examples of such rules? I do believe this would work better in Dapr runtime because we would make it generic for HTTP and gRPC.

li1234yun:
> What are examples of such rules? I do believe this would work better in Dapr runtime because we would make it generic for HTTP and gRPC.

Yes, I agree with you!

Example:
- I want to invoke target service method, but I need to verify the user permission firstly. I can just transfer my invoke to authorization service to verify user permission.If invoke pass, I will continue to invoke target service method. Otherwise the invoke will be denied and will not invoke the target service method.

li1234yun:
Hello!

I want to know that how can I do now? @artursouza I think your consideration is correct! Will you achieve at the dapr runtime?

artursouza:
@li1234yun There are many ways to achieve this. A design proposal would be ideal to initiate the discussion. Here are some ideas:

1. Create a "Intercept" or "Chain" CRD that can be applied to an API for pre processing or post processing - basically expanding the notion of middleware beyond just HTTP.
2. Add it as a built-in feature in service invocation, where the authorization endpoint can be passed as a metadata and dapr runtime interprets it and makes the service invocation.

/cc @pkedy has some ideas related to item no. 1.

li1234yun:
> @li1234yun There are many ways to achieve this. A design proposal would be ideal to initiate the discussion. Here are some ideas:
> 
> 1. Create a "Intercept" or "Chain" CRD that can be applied to an API for pre processing or post processing - basically expanding the notion of middleware beyond just HTTP.
> 2. Add it as a built-in feature in service invocation, where the authorization endpoint can be passed as a metadata and dapr runtime interprets it and makes the service invocation.
> 
> /cc @pkedy has some ideas related to item no. 1.

Hello! @artursouza @yaron2 

I'm sorry for reply you now! I just forgot the pull request. I will wait the above achieve to avoid introducing new redundant code. Thank you!

greenie-msft:
@artursouza, can you provide an update on this PR? There is an open docs PR for this as well.

artursouza:
I don't think we should have this PR merged until there is a proposal and we can compare alternatives. This feature starts to get into workflows and I think it is better to think about that holistically.

/cc @yaron2 @pkedy @greenie-msft

greenie-msft:
@artursouza, let's close this PR.

yaron2:
Closing for now, but we can re-open if this is discussed and agreed on in a GitHub issue.

Issue #960, opened 2021-06-17
Certification tests for Azure Event Hubs binding


artursouza:
## Describe the feature
Certify Azure Event Hubs as GA in input/output binding

## Release Note

RELEASE NOTE:

artursouza:
The IoT hub from "integration test" needs to move to conformance test because the "integration tests" are not officially supported in a GitHub pipelines and it overlaps in objective with our certification tests.

Instead of retry, let's validate reconnection. Add a scenario where the connection to EvenHubs is stopped for a few seconds and validate that publish and subscribe continues to work after connection is recovered.

For AAD authentication, validate one of the authentication options is fine but verify that the code is using the "shared" code for Azure auth.

tanvigour:
@artursouza  Added a scenario covering reconnection, auth testing and IoT migration from integration. Does this look good?

berndverst:
@artursouza FYI -- this component cannot be certified until pending open issues are resolved, especially https://github.com/dapr/components-contrib/issues/1598

It would be very helpful to have a certification test that verifies this very situation. There may be required updates to this component as a result.

Issue #951, opened 2021-06-17
Rewrite Azure Event Hubs Pubsub component


artursouza:
## Describe the feature
Rewrite Azure Events Hub by mapping pubsub component to Azure EventHub namespaces and Azure EventHub to a topic. Automatic provisioning of an EventHub (topic) can be enabled via metadata, similar to how is done for the Azure Service Bus component.
Certify Azure Events Hub as GA in pubsub.

## Release Note

RELEASE NOTE:  **Redesigned** Azure EventHub PubSub component

yaron2:
This component needs to be removed completely, it cannot handle topics the way Dapr promises to consumers.

artursouza:
@yaron2 @berndverst Can there be a new implementation of this component that handles that?

berndverst:
@artursouza new implementation is different from GA -- as such it probably should start out as being Alpha or Beta?

I successfully ran the existing Azure Event Hubs Pub Sub conformance tests. If this doesn't meet our PubSub guarantees we need to overhaul our conformance tests too.

berndverst:
Both @CodeMonkeyLeet  and I put significant work into investigating how to address the behavior of Event Hubs and how to GA this component. Unfortunately it just isn't impossible the way Event Hubs are designed and given the paradigm Dapr promises. Any workaround that could be implemented would be so complex (for example requiring lots of additional infra resources) such that we cannot GA the component.

We need to deprecate EventHubs for Release 1.4.

For reference I'm adding one PR that highlights several of the issues:
https://github.com/dapr/components-contrib/pull/1055

berndverst:
@artursouza please update the issue title to the following text:

**Deprecate Azure Event Hubs Pubsub component.**

We should keep this as part of the 1.4 milestone as this requires docs changes.

yaron2:
Again, I agree with the assessment @berndverst laid out. The namespace/topics design in Event Hubs makes this impossible.

berndverst:
We will announce the deprecation in the docs as soon as possible and then beginning in the 1.4 release notes. The component will be removed from 1.6

https://github.com/dapr/components-contrib/issues/1073

TalalMahmoodChaudhry:
Hi @yaron2 one question to this:

You state here that recommended use for dapr and Event Hub is binding componenet:
https://discord.com/channels/778680217417809931/778680217417809934/874326568825126932

However bindings do not support tracing IDs. Will this be supported for Event Hub binding in the future?
[...]

Issue #964, opened 2021-06-17
Azure EventHubs pubsub component doesn't handle topics correctly


ecaisse:
## Expected Behavior
With Azure EventHubs as a pubsub component, when publishing an event with Dapr to a specific topic, the event should be sent to the endpoint configured for that topic as configured via the /dapr/subscribe endpoint.

## Actual Behavior
Events can be sent to the wrong endpoint, resulting in the wrong handler being called. When combined with the .NET SDK, the wrong controller is called without raising errors despite the failed deserialization.

Tested with Dapr standalone on Windows 10.

## Steps to Reproduce the Problem
Code sample: [DaprEventHub.zip](https://github.com/dapr/components-contrib/files/6669801/DaprEventHub.zip)

1. In the Azure Portal, create an Event hub and storage container as per the [instructions in the Dapr docs](https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-azure-eventhubs/#create-an-azure-event-hub)
    - Create the Event Hubs namespace in a resource group
    - In the Event Hubs namespace, create a new Event Hub
      - In my tests, I was using 4 partitions with 1 day of retention. Capture was off.
    - In the Event Hub _overview_, add a consumer group with the ID of the app supplied to the `dapr run` command
    - In the Event Hub _shared access policies_, add a new policy with Send and Listen for claims.
2. Create a [pubsub component yaml file](https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-azure-eventhubs/#component-format) in the components folder (%UserProfile%\.dapr\components\ in Windows) with the connection string from the Event Hub and storage account information
3. Compile the code sample (.NET) and start the application with Dapr
    - `dapr run -a my-app --app-port 5000 --dapr-http-port 55502 --dapr-grpc-port 55501 --log-level debug dotnet run`
4. In the browser (or using curl/iwr), access http://localhost:5000/api/topic1 around 10 times to generate events
    - All events using that endpoint will be generated with a "Name" property of "Topic1"
5. In the console, you should see the event handlers being triggered. Note that they should all be handled by Topic1Controller, but some of them are not.
    - This issue is intermittent, there is a chance that all events will be properly handled by Topic1Controller. If that is the case, in the components yaml file, change the storageContainerName to another name and restart the application. This will replay all events generated at step 4.

Sample output: 
![image](https://user-images.githubusercontent.com/1854279/122391584-2d9cb280-cf41-11eb-8dcf-d049a5208484.png)

[...]

Issue #939, opened 2021-06-11
Daprd integration testing


artursouza:
## Describe the proposal

Components today are only tested via conformance tests, which validates only the common behaviors. In addition, we need daprd-level integration testing where daprd is build with the latest code for components-contrib and the components are tested with their specific behavior, which means it would require test code per component.

Note: this is not a substitute for conformance tests since those keep the components with compatible behavior.

daixiang0:
Would it bring duplicate tests since compatible behavior?

artursouza:
@daixiang0 There might be some overlap in the test - in other words, there might eventually be compatibility bugs that can also reflect in the integration test results.

daixiang0:
Does it mean just change install way as first step to enable integration test?

akkie:
Does [this type of tests](https://github.com/dapr/components-contrib/pull/973/files) also fall under this category?

artursouza:
> Does this type of tests (#973) also fall under this category?


Yes.

berndverst:
This is done by our certification tests today. Closing this issue as completed - thanks for laying the groundwork @pkedy

PR #1040, opened 2021-07-29
Add Event Hubs pubsub and bindings conformance tests

CodeMonkeyLeet:
# Description

- Update automation for setting up Event Hubs resources in Azure for conformance testing
- Add conformance tests for Event Hubs pubsub and bindings components
- Add retry & backoff handling on subscriber handling error to Event Hubs PubSub component for conformance tests.
- Add cancellation context to Azure EventHubs and update Close() to invoke cancel prior to closing the hub, which cleans up both senders and receivers.

> Note that this does not yet add Event Hubs tests to the github conformance.yml workflow as that requires the new Event Hubs resource to be set up in the Dapr production test environment first.

## Issue reference

Fulfills the conformance testing requirement necessary for GA certification of Event Hubs components as tracked by #951 and #960 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/.github/infrastructure/conformance/azure/conf-test-azure-eventhubs.bicep
+++ b/.github/infrastructure/conformance/azure/conf-test-azure-eventhubs.bicep
@@ -1,0 +1,40 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation and Dapr Contributors.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+param eventHubsNamespaceName string
+param rgLocation string = resourceGroup().location
+param confTestTags object = {}
+
+var eventHubName = '${eventHubsNamespaceName}-topic'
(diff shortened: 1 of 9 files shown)


berndverst:
@CodeMonkeyLeet I thought EventHubs doesn't behave the way PubSub is supposed to work with Dapr. Should the conformance test capture this and therefore fail?

CodeMonkeyLeet:
@berndverst the problem with the conformance tests is that they intended to be generic across components of the same type and do not handle component-specific behaviors, hence the epic #939; the conformance tests feel like a low bar and fail to capture significant issues such as #964, which would need to be added as integration tests according to that epic.

I'd spin up a separate issue on improving the conformance tests (like checking topic routing) but that's going to require a separate PR to address and verify for all existing components.

artursouza (approved):

PR #1044, opened 2021-07-31
Upload conformance test logs as GitHub workflow artifacts

CodeMonkeyLeet:
# Description

The conformance tests do not currently upload their test results as GitHub workflow artifacts that analytics can be run on. This PR aligns test log reporting with dapr/dapr E2E tests so that maintainers can start applying the same analytics tools to the conformance test results as well.

## Issue reference

N/A

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -129,6 +129,12 @@ jobs:
       with:
         path: ./src/github.com/dapr/components-contrib
 
+    - name: Setup test output
+      shell: bash
+      run: |
+        export TEST_OUTPUT_FILE_PREFIX=$GITHUB_WORKSPACE/test_report
+        echo "TEST_OUTPUT_FILE_PREFIX=$TEST_OUTPUT_FILE_PREFIX" >> $GITHUB_ENV
+
     - uses: Azure/login@v1
       with:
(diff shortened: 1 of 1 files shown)


CodeMonkeyLeet:
@wcs1only Can you take a look at this? https://github.com/CodeMonkeyLeet/components-contrib/actions/runs/1083954907 has a sample run from my fork and the resulting artifacts.

.github/workflows/conformance.yml
```diff
@@ -229,10 +235,13 @@ jobs:
 
     - name: Set up Go
       uses: actions/setup-go@v2
+      with:
+        go-version: '^1.16.6'
```
artursouza, line 239: Why do you need to set the go version and not work with the latest?

CodeMonkeyLeet: This surprised me, but the actions/setup-go@v2 isn't pulling down the latest version, it is pulling down go1.15.14 linux/amd64 when left unspecified, which caused the `go install` call later to fail. 

The `^1.16.6` syntax should get us any minor version greater than that, and is not fixed to the exact version.

artursouza (approved):

PR #1052, opened 2021-08-03
Add Close() function to Azure ServiceBusQueues binding

### LINKED ISSUE #895: Azure ServiceBusQueues binding: Add a `Close()` function to clean up client
## Expected Behavior

This component should implement a `Close() error` function to clean up the client instance.

## Actual Behavior

The client is not closed.

## Release Note

RELEASE NOTE: Azure ServiceBusQueues binding: Close added for closing the underlying client

CodeMonkeyLeet:
# Description

- Add Close() function to Azure ServiceBusQueues binding to clean up client 
- Add Close() invocation to bindings.go conformance tests

## Issue reference

Resolves #895

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


CodeMonkeyLeet:
@pkedy Could you take a look at this?

tests/conformance/bindings/bindings.go
```diff
@@ -152,7 +154,7 @@ func ConformanceTests(t *testing.T, props map[string]string, inputBinding bindin
 
 					return nil, nil
 				})
-				assert.NoError(t, err, "input binding read returned an error")
+				assert.ErrorIsf(t, err, context.Canceled, "expected Read canceled on Close")
```
pkedy, line 157: I'm wondering if it's better for components to return `nil` if they are cancelled. Or perhaps it is OK to return `nil` _or_ `context.Canceled`. In other words, dapr/dapr might consider both cases as "happy path".

CodeMonkeyLeet: In my limited go experience, it seems idiomatic that errors are returned when a context is canceled (it is reflected in `ctx.Err()` after all). Rather than attempting to police that contributed components always suppress the error, it's probably easier to have daprd runtime.go ignore `context.Canceled` centrally than to expect components to always return `nil`.

We could say that either `nil` or `context.Canceled` is acceptable, but from the testing perspective, it's hard to tell when we have `nil` vs. the assertion was never hit. This was the case before, as Close on teardown of the process would not be during the lifetime of the test. Once the explicit Close invocation was added during the test, pretty much all of the components have to cancel the blocking Read and they all apparently return `context.Canceled` today (or this check for a specific err should be failing in cases where no err is returned).

pkedy: > it's probably easier to have daprd runtime.go ignore context.Canceled centrally than to expect components to always return nil.

100% agree. We are on the same page.

> it's hard to tell when we have nil vs. the assertion was never hit.

We can make the test pass on `nil` or `content.Canceled`. You might just do something like:

```golang
assert.True(t, err == nil || errors.Is(context.Canceled))
```

CodeMonkeyLeet: I added the change, but part of why I wrote that check specifically to only accept `context.Cancelled` is to figure out what the _de facto_ behavior is and when anyone is introducing a new non-normative behavior. It's a matter of preference to have stricter tests and relax them as needed, vs. being loose up front.

artursouza (approved):

artursouza:
@pkedy Can we merge this?

PR #1047, opened 2021-08-02
Add Event Hubs conformance tests to GitHub workflow

CodeMonkeyLeet:
# Description

- Add the EventHubs bindings and pubsub conformance tests to the automated conformance.yml workflow
- Update conformance test setup for EventHubs to allow the EventHubs pubsub and bindings tests to run concurrently in the workflow by separating the targeted EventHubs and checkpoint storage containers.

> ⚠ This PR requires updating the Dapr Azure test environment before it can be merged. 

##  Issue reference

This PR will also serve as a case study for addressing #1001 

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

Issue #773, opened 2021-03-23
pubsub using eventhub with multiple consumer groups


msjiaxie:
## Ask your question here
We are using EventHub as a message broker, we created a EventHub instance(topic). We have two subscribers with different dapr app id, so we created two consumer groups using two app id. Then we published 10 messages to the EventHub instance, but one consumer group A received 4 messages and the other one consumer group B received 6 messages. From our understanding, consumer group A and B should receive 10 messages separately.

[Setup Azure Eventhubs](https://docs.dapr.io/operations/components/setup-pubsub/supported-pubsub/setup-azure-eventhubs/)
```
Create consumer groups for each subscriber
For every Dapr app that wants to subscribe to events, create an Event Hubs consumer group with the name of the dapr id. For example, a Dapr app running on Kubernetes with dapr.io/app-id: "myapp" will need an Event Hubs consumer group named myapp.
```

The official document says that each consumer group should maintain its own meta data. 

[Consumer Groups](https://docs.microsoft.com/en-us/azure/event-hubs/event-hubs-features)
```
The publish/subscribe mechanism of Event Hubs is enabled through consumer groups. A consumer group is a view (state, position, or offset) of an entire event hub. Consumer groups enable multiple consuming applications to each have a separate view of the event stream, and to read the stream independently at their own pace and with their own offsets.
```

Do I misunderstand the usage of pubsub eventhub? Anyone could help?

CodeMonkeyLeet:
@msjiaxie This does seem to be unintuitive behavior, let me take a look and get back to you.

CodeMonkeyLeet:
@msjiaxie it seems like there are a couple of overlapping issues here, and the gist is that the alpha EventHub pubsub component has Dapr semantics which are at odds with the Event Hubs semantics. Dapr pubsub in general provides a [competing consumers pattern](https://docs.microsoft.com/en-us/previous-versions/msp-n-p/dn568101(v=pandp.10)) for consuming events, and the behavior you observe is consistent with that pattern.

However, as you note, EventHubs was designed explicitly for throughput using the partitioned consumers pattern, and that is not properly handled through the Dapr pubsub interface. The EventHubs _bindings_ component does let you use the partitioned consumer pattern, though I've not tested its behavior with respect to consumer groups.

[...]

PR #1055, opened 2021-08-04
Support multiple topics in Event Hubs

berndverst:
# Description

Updates the Event Hubs component to treat each PubSub Topic as an Event Hub of the same name. These must be manually created as discussed with @artursouza.

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests


berndverst:
@CodeMonkeyLeet would love to get your thoughts on this.

EventHub Namespace -> multiple Event Hubs (one per topic)

Credentials would be SAS token with Send/Receive permission for the entire EventHub Namespace.

CodeMonkeyLeet:
@berndverst This was my reaction when I first learned about #964 too, but there are several issues:
- The component relies on the EventHubs Go SDK Hub to manage checkpointing so that it can produce the competing consumers subscription semantics for Dapr pubsub. 
   - Checkpointing in EventHubs depends on Azure Blob Storage containers, with a file recording checkpoints for each partition. 
   - The partition files are not qualified by the EventHub, so if you point two Hub objects consuming from two different Event Hubs but use the same storage container for checkpointing, it breaks delivery promises.
   - The existing implementation takes as config metadata a single storage container to use, so for this PR to work, it would also need to have access to a blob storage account with separate storage containers for every "topic" event hub, and we've just doubled the manual resource management needed to get Dapr to work with Event Hubs pubsub.
   
    - Side note: There was also a potential issue with how the implementation currently works in that multiple consumers targeting the same Event Hub should each take ownership of a partition and consume from that, but there may have been a bug in the deprecated version of the SDK dapr was using where only the first client would get all the messages (see #773). This may have been fixed when I upgraded the dependency to the currently supported version, but I haven't checked.
   
- Mechanical issues aside, it brings us to the general concern that it doesn't address manual resource management of "topic" Event Hubs, which crosses the streams between devops resource deployment and app business logic in the definition of topics. 

   - I looked briefly at a variant where the component would use the EventHubs namespace management APIs to generate "topic" Event Hubs as needed (requiring credentials with more privileges) but there is no good way of managing the lifetime of these in the existing pubsub model (not safe to delete). Fundamentally, this aspect of Event Hubs is at odds with Dapr's attempts to separate app logic from infrastructural concerns: In pretty much all other pubsub components, topics are logical and not resource abstractions. 

Given @yaron2's https://github.com/dapr/components-contrib/pull/1047#issuecomment-892312743 I wouldn't invest more time in this until we hear otherwise.

berndverst:
There are too many issues to resolve here and the complexity and infra requirements are simply too much to ever certify any possible approach as GA.

This component needs to be deprecated.

PR #1058, opened 2021-08-06
Supports bot assignment via /assign.

artursouza:
# Description

Supports bot assignment via /assign.

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/3245

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/.github/workflows/dapr-bot.yml
+++ b/.github/workflows/dapr-bot.yml
@@ -1,0 +1,32 @@
+name: dapr-bot
+
+on:
+  issue_comment: {types: created}
+
+jobs:
+  daprbot:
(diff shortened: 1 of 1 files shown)


yaron2 (approved):

CodeMonkeyLeet:
Changes look good ... if the expectation is that anyone can use the `/assign` comment, it would be useful to have the bot createComment in reply to when the action fails because the item already has an assignee, otherwise there's not clear feedback loop to the user as to whether it is or is not intended to be a command and under what conditions it works.

PR #1059, opened 2021-08-07
Adds Default TTL Option for Redis State Store

### LINKED ISSUE #1060: Default TTL support for Redis State Store
## Describe the feature
Redis does not allow setting a default TTL at the database / table level. As such to use TTL with Dapr a TTL needs to be explicitly sent with every State Store request.

We should add a default TTL option in the Redis component definition to use whenever no TTL is provided with individual state store requests.

## Release Note

RELEASE NOTE: **ADD** support for default TTL in the Redis State Store component

berndverst:
# Description

Redis unlike other databases that support TTL does not have an option to set a default TTL at the database or table / collection level.

This addition to the Redis State Store component allows defining a default TTL to be used whenever an individual State Store request does not specify a TTL.

Fixes #1060 

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/state/redis/metadata.go
+++ b/state/redis/metadata.go
@@ -10,4 +10,5 @@ import "time"
 type metadata struct {
 	maxRetries      int
 	maxRetryBackoff time.Duration
+	ttlInSeconds    *int
 }
--- a/state/redis/redis.go
+++ b/state/redis/redis.go
@@ -87,7 +87,7 @@ func parseRedisMetadata(meta state.Metadata) (metadata, error) {
 		m.maxRetryBackoff = time.Duration(parsedVal)
 	}
 
-	if val, ok := meta.Properties[maxRetryBackoff]; ok && val != "" {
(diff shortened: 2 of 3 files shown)


berndverst:
With this PR the component YAML will include a `ttlInSeconds` option (completely optional) to set a default TTL.

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: <NAME>
  namespace: <NAMESPACE>
spec:
  type: state.redis
  version: v1
  metadata:
  - name: redisHost
    value: <HOST>
  - name: redisPassword
    value: <PASSWORD>
  - name: enableTLS
    value: <bool> # Optional. Allowed: true, false.
  - name: failover
    value: <bool> # Optional. Allowed: true, false.
  - name: sentinelMasterName
    value: <string> # Optional
  - name: maxRetries
    value: # Optional
  - name: maxRetryBackoff
    value: # Optional
  - name: ttlInSeconds
    value: <int> # Optional
```

artursouza (approved):

Issue #1065, opened 2021-08-09
pubsub/aws/snssqs component to support sqs dead-letters queue


amimimor:
## Describe the feature
The feature might be considered as related to dapr/dapr issue:  https://github.com/dapr/dapr/issues/2217 .
However, the configurable AWS SQS dead-letters builtin feature makes the above redundant as no special behavior needs to be implemented to make use of the removal of inflight messages, upon receive failure, from their origin queue to the dead-letters queue (termed "DLQ" hereafter).

The suggested feature entails the addition of 2 pubsub metadata fields, namely:
```
spec:
  type: pubsub.snssqs
  version: v1
  metadata:
    - name: sqsDeadLettersQueueName
      value: "dl-queue-name"
    - name: deadLettersMaxReceives
      value: "10"

```
The feature enablement is conditional of the existence `sqsDeadLettersQueueName` field (mandatory to enable DLQ) with optional `deadLettersMaxReceives` set to either the user-defined value, or to the current `messageRetryLimit`. In the later case where `deadLettersMaxReceives` is implicitly set to the `messageRetryLimit`, the message would be automatically moved to DLQ and the non-DLQ behavior for deleting the message from the queue would not be called. In case the user specified a `deadLettersMaxReceives` greater than `messageRetryLimit`, the component would fail to load and error out the reason.

 

In many production systems, deletion of messages from pubsub/queues is unreasonable (even if the defined consumer fails to receive), as this may void the system correctness and prohibit the system's engineers from being able to reason about the system's integrity. In some extreme cases, deletion of messages after retries might mean actual loss of money or revenue or other business critical operations. 

Currently that is exactly the default pubsub component behavior.

Therefore it would be advantageous to protect those system with safeguards against message loss by using mechanisms such as DLQ.

On AWS SQS, this is quite easily achievable upon queue creation, by passing in an additional setup that directs SQS to automatically move "un-ACKed" messages to DLQ. 
 
## Release Note

RELEASE NOTE:

PR #1066, opened 2021-08-09
pubsub/aws/snssqs to support SQS dead-letter queue

amimimor:
# Description

I've added 2 metadata fields which define:
* the dead-letters queue (DLQ) name (mandatory)
* and the max number of times a message is not acknowledged before it is moved to DLQ by SQS (optiona)

Once those fields are set  the deletion of messages after the `messageRetryLimit` is not performed and the component assumes that the message is automatically moved by SQS to the defined DLQ.

Additionally, I've added an integration test that runs either with AWS directly or locally using `localstack` (sns,sqs,sts mockups) and tests 2 scenarios:
* current default behavior (w/o DLQ)
* new behavior with DLQ

## Issue reference
https://github.com/dapr/components-contrib/issues/1065

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


artursouza:
@halspang Please, check this PR with Matt.

amimimor:
hi @artursouza . My integration tests over AWS are failing with 'missing region' which indicates that other AWS related envvars are indeed present in the build context (somehow)(i've set t.Skip in cases where the credentials related envvars are missing). Am I right? If so, I need another AWS envvar, namely `AWS_REGION` to be set to a valid value in the test context

amimimor:
> hi @artursouza . My integration tests over AWS are failing with 'missing region' which indicates that other AWS related envvars are indeed present in the build context (somehow)(i've set t.Skip in cases where the credentials related envvars are missing). Am I right? If so, I need another AWS envvar, namely `AWS_REGION` to be set to a valid value in the test context

well, the tests seem to pass ok now (maybe I saw a cached result). thanks anyways and sorry for the bother

pubsub/aws/snssqs/snssqs.go
```diff
@@ -54,15 +55,19 @@ type snsSqsMetadata struct {
 	messageVisibilityTimeout int64
 	// number of times to resend a message after processing of that message fails before removing that message from the queue. Default: 10
 	messageRetryLimit int64
+	// if sqsDeadLettersQueueName is set to a value, then the deadLettersMaxReceives defines the number of times a message is received
+	// before it is moved to the dead-letters queue. This value must be smaller than messageRetryLimit
+	deadLettersMaxReceives int64
```
artursouza, line 60: Can this be renamed to messageReceiveLimit?

pubsub/aws/snssqs/snssqs.go
```diff
@@ -175,6 +180,33 @@ func (s *snsSqs) getSnsSqsMetatdata(metadata pubsub.Metadata) (*snsSqsMetadata,
 		md.messageRetryLimit = retryLimit
 	}
 
+	if val, ok := getAliasedProperty([]string{"sqsDeadLettersQueueName"}, metadata); ok {
+		md.sqsDeadLettersQueueName = val
+	}
+
+	if val, ok := getAliasedProperty([]string{"deadLettersMaxReceives"}, metadata); !ok {
+		md.deadLettersMaxReceives = md.messageRetryLimit
```
artursouza, line 188: Why not make Max Receive mandatory? Mixing these can be confusing.

pubsub/aws/snssqs/snssqs_integ_test.go
```diff
@@ -0,0 +1,310 @@
+// ------------------------------------------------------------
```
artursouza, line 1: Move these to the e2e folder until we have a standardized way to write integration tests. See this as example: https://github.com/dapr/components-contrib/tree/master/tests/e2e/bindings/zeebe

artursouza (changes requested):

pubsub/aws/snssqs/snssqs.go
amimimor: sure. I borrowed it from the AWS official param. but your suggestion makes sense

pubsub/aws/snssqs/snssqs_integ_test.go
amimimor: this is WIP and lots of work. So I think I'll open another PR for that. @artursouza wdyt? I've implemented the suggested changes (thanks, good ones!). I'm going to trigger a new build as it seem to glitch. please re-review

artursouza: I am OK for the E2E test to be in a separate PR. Once we have integration tests, it can land there too.

pubsub/aws/snssqs/snssqs.go
```diff
@@ -54,17 +55,21 @@ type snsSqsMetadata struct {
...
+	messageReceiveLimit int64
 	// amount of time to await receipt of a message before making another request. Default: 1
 	messageWaitTimeSeconds int64
 	// maximum number of messages to receive from the queue at a time. Default: 10, Maximum: 10
 	messageMaxNumber int64
 }
 
 const (
-	awsSqsQueueNameKey = "dapr-queue-name"
+	awsSqsQueueNameKey = "dapr-worker-name"
```
artursouza, line 68: This change is a breaking change. This component is alpha but I would avoid breaking changes still.

artursouza (changes requested):
Just a small change. Plus, don't forget to remove the integ_test file.

pubsub/aws/snssqs/snssqs.go
amimimor: sure. you're right

artursouza (approved):

Issue #1063, opened 2021-08-09
Enable setting Message Properties for Service Bus Component


TalalMahmoodChaudhry:
## Describe the feature
Service Bus Message have default properties. The list of properties can be found [here](https://docs.microsoft.com/en-us/rest/api/servicebus/message-headers-and-properties#message-headers). Currently only the generally supported [TTL property is set in code](https://github.com/dapr/components-contrib/blob/6c28306b10dbea1bf6f8a5773123ec4f7669106e/pubsub/azure/servicebus/servicebus.go#L299-L305).

It would be great if other properties are also supported, especially the **Label** property. This is very useful when using filtering for different subscriptions of a service bus topic. 

## Release Note

 RELEASE NOTE: **ADD** New feature in Dapr Service Bus Component. Now it is possible to set message property *Label*

Issue #668, opened 2021-02-08
Azure ServiceBus allow messageid overwrite via metadata


bgelens:
## Describe the proposal

At this time it doesn't seem possible to overwrite the message id through passing it via metadata. Only TTL seems supported. For deduplication purposes, the message id needs to be provided instead of generated.

Please expose the overwrite capability for message id

bgelens:
@jjcollinge Sorry to tag you if not the correct person. I'd like to un-stale this issue (feature request). It would still be very useful to be able to provide our own messageid with service bus to leverage dedup functionality

jjcollinge:
/assign

@bgelens, np - happy to look into implementing this when I get time. Sorry I can't guarantee when that'll be, but should be within the next couple of weeks.

PR #1071, opened 2021-08-11
[WIP] Enable ASB Message Metadata

jjcollinge:
# Description

Added ability to set Azure Service Bus headers when publishing via the Azure Service Bus pubsub component.
See https://docs.microsoft.com/en-us/rest/api/servicebus/message-headers-and-properties#message-headers for reference.

## Issue reference

Please reference the issue this PR will close: #1063 , #668 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/pubsub/azure/servicebus/message.go
+++ b/pubsub/azure/servicebus/message.go
@@ -1,0 +1,164 @@
+package servicebus
+
+import (
+	"fmt"
+
+	azservicebus "github.com/Azure/azure-service-bus-go"
+	contrib_metadata "github.com/dapr/components-contrib/metadata"
+	"github.com/dapr/components-contrib/pubsub"
+)
+
+const (
+	// MessageIdMetadataKey defines the metadata key for the message id.
+	MessageIdMetadataKey = "MessageId"
+
+	// CorrelationIdMetadataKey defines the metadata key for the correlation id.
+	CorrelationIdMetadataKey = "CorrelationId"
+
+	// SessionIdMetadataKey defines the metadata key for the session id.
+	SessionIdMetadataKey = "SessionId"
+
+	// LabelMetadataKey defines the metadata key for the label.
+	LabelMetadataKey = "Label"
+
+	// ReplyToMetadataKey defines the metadata key for the reply to value.
+	ReplyToMetadataKey = "ReplyTo"
+
+	// ToMetadataKey defines the metadata key for the to value.
+	ToMetadataKey = "To"
+
+	// PartitionKeyMetadataKey defines the metadata key for the partition key.
+	PartitionKeyMetadataKey = "PartitionKey"
+
+	// ContentTypeMetadataKey defines the metadata key for the content type.
+	ContentTypeMetadataKey = "Content-Type"
+)
+
+// NewMessageFromRequest builds a new Azure Service Bus message from a PublishRequest.
+func NewMessageFromRequest(req *pubsub.PublishRequest) (*azservicebus.Message, error) {
+	msg := azservicebus.NewMessage(req.Data)
+
+	// Common properties.
+	ttl, hasTTL, _ := contrib_metadata.TryGetTTL(req.Metadata)
+	if hasTTL {
+		msg.TTL = &ttl
+	}
+
+	// Azure Service Bus specific properties.
+	// reference: https://docs.microsoft.com/en-us/rest/api/servicebus/message-headers-and-properties#message-headers
+	msgId, hasMsgId, _ := tryGetMessageId(req.Metadata)
+	if hasMsgId {
+		msg.ID = msgId
+	}
+
+	correlationId, hasCorrelationId, _ := tryGetCorrelationId(req.Metadata)
+	if hasCorrelationId {
+		msg.CorrelationID = correlationId
+	}
+
+	sessionId, hasSessionId, _ := tryGetSessionId(req.Metadata)
+	if hasSessionId {
+		msg.SessionID = &sessionId
+	}
+
+	label, hasLabel, _ := tryGetLabel(req.Metadata)
+	if hasLabel {
+		msg.Label = label
+	}
+
+	replyTo, hasReplyTo, _ := tryGetReplyTo(req.Metadata)
+	if hasReplyTo {
+		msg.ReplyTo = replyTo
+	}
+
+	to, hasTo, _ := tryGetTo(req.Metadata)
+	if hasTo {
+		msg.To = to
+	}
+
+	partitionKey, hasPartitionKey, _ := tryGetPartitionKey(req.Metadata)
+	if hasPartitionKey {
+		if hasSessionId {
+			if partitionKey != sessionId {
+				return nil, fmt.Errorf("session id %s and partition key %s should be equal when both present", sessionId, partitionKey)
+			}
+		}
+
+		if msg.SystemProperties == nil {
+			msg.SystemProperties = &azservicebus.SystemProperties{}
+		}
+
+		msg.SystemProperties.PartitionKey = &partitionKey
+	}
+
+	contentType, hasContentType, _ := tryGetContentType(req.Metadata)
+	if hasContentType {
+		msg.ContentType = contentType
+	}
+
+	return msg, nil
+}
+
+func tryGetMessageId(props map[string]string) (string, bool, error) {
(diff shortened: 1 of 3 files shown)


jjcollinge:
@artursouza I'm trying to resolve the linked issues by using Azure Service Bus (ASB) metadata to enrich the pubsub message. I have handled this on the publisher side ok. However, on the receiver side, if I map server set metadata from the ASB message onto the pubsub message it is not then added to the request that is sent on to the app.

As you can see in the screenshot below - the `pubsubSubscribedMessage` passed into the [publishMessageHTTP](https://github.com/dapr/dapr/blob/a1957445eaff25b44f543b9a577bf237fcf9ae1e/pkg/runtime/runtime.go#L1216) function does contain the server set metadata. However, this isn't added to the `InvokeMethodRequest`. Is this by design or just a missing feature?
image

jjcollinge:
I think implementing this might require a few Dapr changes but let me just walk through my current understanding. 

__What is the goal?__
Enrich Dapr pubsub messages with implementation specific metadata.

__What is the problem?__
A publishing app can add metadata to a pubsub request using query params. However, we currently only support [`ttl`](https://docs.dapr.io/developing-applications/building-blocks/pubsub/pubsub-message-ttl/#introduction) as an attribute that will be added to the CloudEvent envelope. The senders metadata will be passed into the PubSub component and the component can use that to decorate their native message (but __not__ the CE data as this is already serialized before the call into the PubSub component - unless we deserialize, edit and reserialize again) before sending it to the server. Some of that decorated data might be used by the server for routing but some might be intended for the recipient i.e. `replyTo` in the case of ASB. On the receiving side, the PubSub component receives a new native message from the server, containing a data payload and some server provided attributes (i.e. `enqueuedTime`, etc.). The PubSub component builds a `pubsub.NewMessage` from the received native message and calls the Dapr handler which unmarshals the data field into a CloudEvent and then calls either `publishMessageHTTP` or `publishMessageGRPC` which both ignore any additional metadata outside the CE on the `pubsub.NewMessage`. So the problems are that on the publishing side the PubSub component cannot add any custom attributes to the CE envelope (if we want to support this?). And on the receiving side, the PubSub component cannot add any custom attributes to the CE envelope as this comes directly from the native message data (again, unless we unmarshal and reserialize) or add any metadata to the request sent to the app.

__What are our options?__
I think the simplest solution would be to not let publishers add custom attributes to the CE envelope and let the PubSub component use the data directly from metadata and apply it to their native message (I have done this in this PR).
On the receiving side, we can modify `publishMessageHTTP` and `publishMessageGRPC` to __not__ ignore Metadata and add it to the app request headers/metadata. However, this would mean we explicitly make the decision _not_ to use CE custom extensions and instead use a Dapr and transport specific metadata format for PubSub specific metadata. This would also mean the envelope itself does not encapsulate all the data the receiving app might need and metadata would be in 2 places.

An alternative if we wanted to use CE attributes might be to add some hooks to the PubSub interface so that it can add metadata to the CE before it is serialized (similar to TTL but PubSub specific) and also set metadata on the CE on the receiving side after it has been unmarshalled.

This might result in the receiving app getting a CE looking something like:
```
{
	"data": {
		"messageType": 'A',
		"message": 'Message on A'
	},
	"specversion": '1.0',
	"datacontenttype": 'application/json',
	"source": 'react-form',
	"pubsubname": 'pubsub',
	"traceid": '00-509cb4d370acebae0b4496b1bc910aed-adcb6c6f71af1f27-01',
	"id": 'e113a376-d0af-4150-8802-c4e98e7e4673',
	"type": 'com.dapr.event.sent',
	"topic": 'A',
	"asbsessionid": "123",  // Must conform to naming convention
	"asbcorrelationid": "456",
	"asbreplyto": "http://localhost:3000",
	"asbdeliverycount": 3
}
```
[CloudEvents Naming Conventions](https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#attribute-naming-convention)

I'm still fairly new to CE so this might not make sense but I'd welcome some discussion to help us support custom message metadata for PubSub. cc: @yaron2 @artursouza

artursouza:
@jjcollinge Thanks for the detailed and relevant discussion on this.

Metadata handling is an important topic and, like you noticed, it can be handled in the runtime level or at the component level. `ttl` is an example where it is handled in both, depending on the component's capabilities. I do think that it is time for us to let components to decorate the CE object on publish or on delivery (subscribe). Let me summarize what I think we can do based on your feedback and how I see this working in the future.

I like the idea to use metadata to trigger custom behavior in the pubsub component - it is aligned to other APIs in Dapr while the CE payload is more focused in the generic pubsub abstraction. In that way, to keep symmetry, I agree that the custom metadata from the component should be delivered outside of the CE payload too and subscribers can handle that via gRPC metadata or HTTP header. To avoid conflict with standard HTTP headers, these metadata keys must not conflict with other HTTP headers.

/cc @pkedy @yaron2 @CodeMonkeyLeet  WDYT?

jjcollinge:
Ok, so this change along with the changes to the runtime [here](https://github.com/dapr/dapr/pull/3604) result in the following metadata being set by the pubsub component.
For HTTP:
image
For gRPC:
image
__NOTE:__ For grpc we are using the [metadata](https://pkg.go.dev/google.golang.org/grpc/metadata) package so additional properties are just added to a map rather than as independent `context` values which is not scalable.

Currently on publish the TTL is extracted from the metadata and set as the CE expiration field - I've also allowed it to set the TTL property in the ASB metadata incase a consumer wants to bind to the original ASB metadata rather than the CE property.

What do you think we should do regarding the message id? This PR would allow a publisher to set the ASB message ID property by using metadata but the CE will still have a [unique ID generated by Dapr](https://github.com/dapr/dapr/blob/a7f847d686b63c91904e5a8c1821cc1fefd088d2/pkg/runtime/pubsub/cloudevents.go#L30). This isn't necessarily a problem - just the consumer would need to be aware how to handle each property independently for things like deduplication.

n.b. it appears we don't have good coverage of the publish metadata property in our SDKs.

@artursouza if this PR is in line with your thinking, let me know and I'll clean up the code and add some tests. If not, lets discuss.

jjcollinge:
Worth pointing out that this is a different approach to that taken in [#2735](https://github.com/dapr/dapr/pull/2735/files) which adds the metadata as a field directly onto the gRPC message. In the case of gRPC, the `TopicEventRequest` is the CloudEvent so adding this wouldn't align with the approach discussed above.

artursouza:
@jjcollinge `TopicEventRequest` having metadata is orthogonal to this change. I am actually thinking if we need metadata in `TopicEventRequest`.

BTW, this change LGTM.

jjcollinge:
Ok thanks. I'm on holiday until the 13th so I'll fix up the PR upon my return.

jjcollinge:
> @jjcollinge `TopicEventRequest` having metadata is orthogonal to this change. I am actually thinking if we need metadata in `TopicEventRequest`.
> 
> BTW, this change LGTM.

@artursouza - I'll hopefully get some time to get this finalized on Sunday. I know you mentioned that [PR 3593](https://github.com/dapr/dapr/pull/3593) is an orthogonal change but I'd appreciate some clarity there. In that PR, for gRPC, the same data (`msg.metadata`) is being added directly to the CE payload as I am adding to the gRPC context metadata. If you compare [PR 3593](https://github.com/dapr/dapr/pull/3593) and my Dapr PR [PR 3604](https://github.com/dapr/dapr/pull/3604). I'd think we should be clear on where we want this data to reside. cc: @pkedy

jjcollinge:
Does this require a docs update?

artursouza:
> Does this require a docs update?

Yes, please.

jjcollinge:
So just one thing to note on this PR that might be slightly confusing for users is the fact that Dapr [will always autogen](https://github.com/dapr/dapr/blob/88bb0764ab72d0443d211345dda57052056bd6f3/pkg/runtime/pubsub/cloudevents.go#L30) a new ID for the cloud event. So people can set `metadata.MessageId` and read `metadata.MessageId` using this change but this isn't a special field and the `id` in the cloud event will be different. I don't think it's a problem as long as users are aware of this behavior. I'll make a note in the docs PR.

jjcollinge:
docs PR merged for v1.5 https://github.com/dapr/docs/pull/1802

artursouza:
Approved.

Issue #949, opened 2021-06-17
Certify Azure SQL Server as GA in state store


artursouza:
## Describe the feature
Certify Azure SQL Server as GA in state store

## Release Note

RELEASE NOTE:

daixiang0:
I would like to know what is the standard for GA, also I see #913, is it an enhancement?

daixiang0:
Hi, @artursouza I would like to know what is the standard for GA, also I see #913, is it an enhancement?

CodeMonkeyLeet:
@daixiang0 https://docs.dapr.io/operations/components/certification-lifecycle/ describes the stated standards for GA, and minimally it looks like a conformance test would need to be added for the sqlserver state component ... If you're interested, I'm actually looking into making the conformance tests for Azure services easier to set up. 

I've also looked into #913, and in discussion with @paulyuk, it's an enhancement and not a prerequisite for Azure SQL GA certification.

daixiang0:
Is there a doc to comment maintainers for GA components?

artursouza:
@daixiang0 I don't think we have an official list of maintainer per GA component. @msfussell thoughts on this?

daixiang0:
As https://docs.dapr.io/operations/components/certification-lifecycle describes, we need a list for it, or we may comment it.

artursouza:
@msfussell @CodeMonkeyLeet I think we need to have a special file on each component folder to mention who are the maintainers for each component and when there is a PR that affects the file, Dapr Bot automatically adds them to the review. In addition, the same list must be documented in the docs website for each component.

CodeMonkeyLeet:
I agree that starting to track in specific component maintainers in the repo is going to be necessary for the GA certifications, and the suggestion aligns with how [Prow](https://github.com/kubernetes/test-infra/tree/master/prow) handles it. 

One thing to note is that you can already do this with the GitHub [CODEOWNERS](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-code-owners#codeowners-syntax) file without having to add a bot for this. Since the component maintainers should have PR approval privileges, making them the subfolder owners in CODEOWNERS not only automatically adds them to component reviews, but also gives them codeowner approval rights. If we want to add additional pre-merge checks, then those are easier to incrementally add as Dapr Bot commands.

berndverst:
## Component test plan

- [ ] [Integration] Test default vs custom SQLServer `schema` name handling
[...]

PR #1078, opened 2021-08-14
Conformance tests & bugfixes for SQLServer

### LINKED ISSUE #1036: Sidecar crashing when state.sqlserver included
## Expected Behavior

When I attempted to include a sql server state component in my Dapr app configuration within a local Kind Kubernetes cluster the Dapr sidecar for my service should initiate the state store and the Dapr sidecar should spin up and start servicing requests for the associated service.

## Actual Behavior
The sidecar starts to initialise the state component, but a panic is thrown and the sidecar spins into a spiral of restarts.

The following logs are generated..
[dapr-sidecar.log](https://github.com/dapr/components-contrib/files/6896587/dapr-sidecar.log)

After discussion within Discord it was suggested that the sql dependency may require updating.. a quick check shows that this looks like the most likely issue.

Screenshot_2021-07-28_at_22 15 27

Install the latest version of Dapr (v1.3.0 at point of raising this issue), ensure a sidecar is being injected into a service pod and then add a sql state component - I had the following state config..

```
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: idea-state
  namespace: default
spec:
  type: state.sqlserver
  version: v1
  metadata:
    - name: connectionString
      value: Server=xxxxxxx.database.windows.net;Database=xxxxx-xxxx;User ID=uxxx;Password=xxxx;
    - name: tableName
      value: state # Required.
    - name: schema
      value: "idea" # Optional. defaults to "dbo"
```

As above I guess :/ 

## Release Note

RELEASE NOTE: **UPDATE** Bumped go-mssqldb dependency version

berndverst:
# Description

Adds conformance tests to and fixes bugs in the SqlServer state store component.

The new conformance test suite has been added to the GitHub workflow and has already successfully run as part of this PR.

Conformance tests can manually be run like so:

```bash
docker run --name sqlserver -e "ACCEPT_EULA=Y" -e "SA_PASSWORD=Pass@Word1" -p 1433:1433 -d mcr.microsoft.com/mssql/server:2019-GA-ubuntu-16.04
go test -v -tags=conftests -count=1 ./tests/conformance -run="TestStateConformance/sqlserver"
```

Integration tests can then be run via the following which checks for the same variable to be set
```bash
export DAPR_TEST_SQL_CONNSTRING="server=localhost;user id=sa;password=Pass@Word1;port=1433;"
go test -v ./state/sqlserver
```

Note that I did not add the `DAPR_TEST_SQL_CONNSTRING` variable. This has been in the code for a long time.

## Issue reference

#949
Fixes #1036 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

### DIFF
--- a/.github/infrastructure/conformance/azure/setup-azure-conf-test.sh
+++ b/.github/infrastructure/conformance/azure/setup-azure-conf-test.sh
@@ -415,3 +415,9 @@ echo "INFO: setup-azure-conf-test completed."
 echo "INFO: Remember to \`source ${ENV_CONFIG_FILENAME}\` before running local conformance tests."
 echo "INFO: ${AZURE_CREDENTIALS_FILENAME} contains the repository secret to set to run the GitHub conformance test workflow."
 echo "INFO: To teardown the conformance test resources, run ${TEARDOWN_SCRIPT_NAME}."
+
+# ----------------------------------
+# Populate SQLServer test settings
+# ----------------------------------
+echo "Configuring SQL Server test settings ..."
+echo export DAPR_TEST_SQL_CONNSTRING="server=localhost;user id=sa;password=Pass@Word1;port=1433;" >> "${ENV_CONFIG_FILENAME}"
--- a/.github/infrastructure/conformance/azure/setup-azure-conf-test.sh
+++ b/.github/infrastructure/conformance/azure/setup-azure-conf-test.sh
@@ -415,9 +415,3 @@ echo "INFO: setup-azure-conf-test completed."
 echo "INFO: Remember to \`source ${ENV_CONFIG_FILENAME}\` before running local conformance tests."
 echo "INFO: ${AZURE_CREDENTIALS_FILENAME} contains the repository secret to set to run the GitHub conformance test workflow."
 echo "INFO: To teardown the conformance test resources, run ${TEARDOWN_SCRIPT_NAME}."
-
-# ----------------------------------
-# Populate SQLServer test settings
-# ----------------------------------
-echo "Configuring SQL Server test settings ..."
-echo export DAPR_TEST_SQL_CONNSTRING="server=localhost;user id=sa;password=Pass@Word1;port=1433;" >> "${ENV_CONFIG_FILENAME}"
--- a/go.mod
+++ b/go.mod
@@ -44,7 +44,7 @@ require (
 	github.com/dancannon/gorethink v4.0.0+incompatible
 	github.com/dapr/kit v0.0.2-0.20210614175626-b9074b64d233
 	github.com/deepmap/oapi-codegen v1.8.1 // indirect
-	github.com/denisenkom/go-mssqldb v0.0.0-20191128021309-1d7a30a10f73
+	github.com/denisenkom/go-mssqldb v0.0.0-20210411162248-d9abbec934ba
 	github.com/dghubble/go-twitter v0.0.0-20190719072343-39e5462e111f
 	github.com/dghubble/oauth1 v0.6.0
 	github.com/didip/tollbooth v4.0.2+incompatible
--- a/go.mod
+++ b/go.mod
@@ -24,7 +24,7 @@ require (
 	github.com/aerospike/aerospike-client-go v4.5.0+incompatible
 	github.com/agrea/ptr v0.0.0-20180711073057-77a518d99b7b
 	github.com/ajg/form v1.5.1 // indirect
-	github.com/alibaba/sentinel-golang v1.0.2-0.20210728053800-194d4be01dfe
+	github.com/alibaba/sentinel-golang v1.0.3
 	github.com/alicebob/miniredis/v2 v2.13.3
 	github.com/aliyun/aliyun-oss-go-sdk v2.0.7+incompatible
 	github.com/aliyun/aliyun-tablestore-go-sdk v1.6.0
--- a/state/sqlserver/migration.go
+++ b/state/sqlserver/migration.go
@@ -126,8 +126,7 @@ func (m *migration) executeMigrations() (migrationResult, error) {
 }
 
 func runCommand(tsql string, db *sql.DB) error {
-	_, err := db.Exec(tsql)
-	if err != nil {
+	if _, err := db.Exec(tsql); err != nil {
 		return err
 	}
 
@@ -275,32 +274,61 @@ func (m *migration) ensureUpsertStoredProcedureExists(db *sql.DB, mr migrationRe
 		CREATE PROCEDURE %s (
 			@Key 			%s,
 			@Data 			NVARCHAR(MAX),
-			@RowVersion 	BINARY(8))
+			@RowVersion 	BINARY(8),
+			@FirstWrite		BIT)
 		AS
 			IF (@RowVersion IS NOT NULL)
 			BEGIN
 				UPDATE [%s]
 				SET [Data]=@Data, UpdateDate=GETDATE()
 				WHERE [Key]=@Key AND RowVersion = @RowVersion
-
 				RETURN
 			END
 
-			BEGIN TRY
-				INSERT INTO [%s] ([Key], [Data]) VALUES (@Key, @Data);
-			END TRY
+			IF (@FirstWrite=1)
+			BEGIN
+				BEGIN TRANSACTION;
+				IF EXISTS (SELECT * FROM [%s] WHERE [KEY]=@KEY)
+				BEGIN
+					THROW 2714, ''FIRST-WRITE: COMPETING RECORD ALREADY WRITTEN.'', 1
+					RETURN
+				END
+				BEGIN
+					BEGIN TRY
+						INSERT INTO [%s] ([Key], [Data]) VALUES (@Key, @Data);
+					END TRY
+		
+					BEGIN CATCH
+						IF ERROR_NUMBER() IN (2601, 2627)
+						UPDATE [%s]
+						SET [Data]=@Data, UpdateDate=GETDATE()
+						WHERE [Key]=@Key AND RowVersion = ISNULL(@RowVersion, RowVersion)
+					END CATCH
+				END
+				COMMIT;
+			END
+			ELSE
+			BEGIN
+				BEGIN TRY
+					INSERT INTO [%s] ([Key], [Data]) VALUES (@Key, @Data);
+				END TRY
 
-			BEGIN CATCH
-				IF ERROR_NUMBER() IN (2601, 2627)
-				UPDATE [%s]
-				SET [Data]=@Data, UpdateDate=GETDATE()
-				WHERE [Key]=@Key AND RowVersion = ISNULL(@RowVersion, RowVersion)
-			END CATCH`,
+				BEGIN CATCH
+					IF ERROR_NUMBER() IN (2601, 2627)
+					UPDATE [%s]
+					SET [Data]=@Data, UpdateDate=GETDATE()
+					WHERE [Key]=@Key AND RowVersion = ISNULL(@RowVersion, RowVersion)
+				END CATCH
+			END`,
 		mr.upsertProcFullName,
 		mr.pkColumnType,
 		m.store.tableName,
 		m.store.tableName,
-		m.store.tableName)
+		m.store.tableName,
+		m.store.tableName,
+		m.store.tableName,
+		m.store.tableName,
+	)
 
 	return m.createStoredProcedureIfNotExists(db, mr.upsertProcName, tsql)
 }
--- a/state/sqlserver/migration.go
+++ b/state/sqlserver/migration.go
@@ -271,57 +271,78 @@ func (m *migration) createStoredProcedureIfNotExists(db *sql.DB, name string, es
 /* #nosec */
 func (m *migration) ensureUpsertStoredProcedureExists(db *sql.DB, mr migrationResult) error {
 	tsql := fmt.Sprintf(`
-		CREATE PROCEDURE %s (
-			@Key 			%s,
-			@Data 			NVARCHAR(MAX),
-			@RowVersion 	BINARY(8),
-			@FirstWrite		BIT)
-		AS
-			IF (@RowVersion IS NOT NULL)
-			BEGIN
-				UPDATE [%s]
-				SET [Data]=@Data, UpdateDate=GETDATE()
-				WHERE [Key]=@Key AND RowVersion = @RowVersion
-				RETURN
-			END
-
-			IF (@FirstWrite=1)
-			BEGIN
-				BEGIN TRANSACTION;
-				IF EXISTS (SELECT * FROM [%s] WHERE [KEY]=@KEY)
-				BEGIN
-					THROW 2714, ''FIRST-WRITE: COMPETING RECORD ALREADY WRITTEN.'', 1
-					RETURN
-				END
-				BEGIN
-					BEGIN TRY
-						INSERT INTO [%s] ([Key], [Data]) VALUES (@Key, @Data);
-					END TRY
-		
-					BEGIN CATCH
-						IF ERROR_NUMBER() IN (2601, 2627)
-						UPDATE [%s]
-						SET [Data]=@Data, UpdateDate=GETDATE()
-						WHERE [Key]=@Key AND RowVersion = ISNULL(@RowVersion, RowVersion)
-					END CATCH
-				END
-				COMMIT;
-			END
-			ELSE
-			BEGIN
-				BEGIN TRY
-					INSERT INTO [%s] ([Key], [Data]) VALUES (@Key, @Data);
-				END TRY
-
-				BEGIN CATCH
-					IF ERROR_NUMBER() IN (2601, 2627)
-					UPDATE [%s]
-					SET [Data]=@Data, UpdateDate=GETDATE()
-					WHERE [Key]=@Key AND RowVersion = ISNULL(@RowVersion, RowVersion)
-				END CATCH
-			END`,
+			CREATE PROCEDURE %s (
+				@Key 			%s,
+				@Data 			NVARCHAR(MAX),
+				@RowVersion		BINARY(8),
+				@FirstWrite		BIT)
+			AS
+				IF (@FirstWrite=1)
+					BEGIN
+						IF (@RowVersion IS NOT NULL)
+							BEGIN
+								BEGIN TRANSACTION;
+								IF NOT EXISTS (SELECT * FROM [%s] WHERE [KEY]=@KEY AND RowVersion = @RowVersion)
+									BEGIN
+										THROW 2601, ''FIRST-WRITE: COMPETING RECORD ALREADY WRITTEN.'', 1
+									END
+								BEGIN
+									UPDATE [%s]
+									SET [Data]=@Data, UpdateDate=GETDATE()
+									WHERE [Key]=@Key AND RowVersion = @RowVersion
+								END
+								COMMIT;
+							END
+						ELSE
+							BEGIN
+								BEGIN TRANSACTION;
+								IF EXISTS (SELECT * FROM [%s] WHERE [KEY]=@KEY)
+									BEGIN
+										THROW 2601, ''FIRST-WRITE: COMPETING RECORD ALREADY WRITTEN.'', 1
+									END
+								BEGIN
+									BEGIN TRY
+										INSERT INTO [%s] ([Key], [Data]) VALUES (@Key, @Data);
+									END TRY
+						
+									BEGIN CATCH
+										IF ERROR_NUMBER() IN (2601, 2627)
+											UPDATE [%s]
+											SET [Data]=@Data, UpdateDate=GETDATE()
+											WHERE [Key]=@Key AND RowVersion = ISNULL(@RowVersion, RowVersion)
+									END CATCH
+								END
+								COMMIT;	
+							END
+					END
+				ELSE
+					BEGIN
+						IF (@RowVersion IS NOT NULL)
+							BEGIN
+								UPDATE [%s]
+								SET [Data]=@Data, UpdateDate=GETDATE()
+								WHERE [Key]=@Key AND RowVersion = @RowVersion
+								RETURN
+							END
+						ELSE
+							BEGIN
+								BEGIN TRY
+									INSERT INTO [%s] ([Key], [Data]) VALUES (@Key, @Data);
+								END TRY
+					
+								BEGIN CATCH
+									IF ERROR_NUMBER() IN (2601, 2627)
+										UPDATE [%s]
+										SET [Data]=@Data, UpdateDate=GETDATE()
+										WHERE [Key]=@Key AND RowVersion = ISNULL(@RowVersion, RowVersion)
+								END CATCH
+							END
+					END
+	`,
 		mr.upsertProcFullName,
 		mr.pkColumnType,
+		m.store.tableName,
+		m.store.tableName,
 		m.store.tableName,
 		m.store.tableName,
 		m.store.tableName,
--- a/state/sqlserver/sqlserver.go
+++ b/state/sqlserver/sqlserver.go
@@ -442,10 +442,6 @@ func (s *SQLServer) Delete(req *state.DeleteRequest) error {
 
 	// err represents errors thrown by the stored procedure or the database itself
 	if err != nil {
-		if req.ETag != nil {
-			return state.NewETagError(state.ETagMismatch, err)
-		}
-
 		return err
 	}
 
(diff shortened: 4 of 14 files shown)


.github/infrastructure/conformance/azure/setup-azure-conf-test.sh
```diff
@@ -415,3 +415,9 @@ echo "INFO: setup-azure-conf-test completed."
 echo "INFO: Remember to \`source ${ENV_CONFIG_FILENAME}\` before running local conformance tests."
 echo "INFO: ${AZURE_CREDENTIALS_FILENAME} contains the repository secret to set to run the GitHub conformance test workflow."
 echo "INFO: To teardown the conformance test resources, run ${TEARDOWN_SCRIPT_NAME}."
+
+# ----------------------------------
+# Populate SQLServer test settings
+# ----------------------------------
+echo "Configuring SQL Server test settings ..."
+echo export DAPR_TEST_SQL_CONNSTRING="server=localhost;user id=sa;password=Pass@Word1;port=1433;" >> "${ENV_CONFIG_FILENAME}"
```
daixiang0, line 423: Is there any places to control those hardcode value? I prefer to set and use them as ENV.

go.mod
```diff
@@ -44,7 +44,7 @@ require (
 	github.com/dancannon/gorethink v4.0.0+incompatible
 	github.com/dapr/kit v0.0.2-0.20210614175626-b9074b64d233
 	github.com/deepmap/oapi-codegen v1.8.1 // indirect
-	github.com/denisenkom/go-mssqldb v0.0.0-20191128021309-1d7a30a10f73
+	github.com/denisenkom/go-mssqldb v0.0.0-20210411162248-d9abbec934ba
```
daixiang0, line 47: Is it necessary?

state/sqlserver/migration.go
```diff
@@ -275,32 +274,61 @@ func (m *migration) ensureUpsertStoredProcedureExists(db *sql.DB, mr migrationRe
...
-				UPDATE [%s]
-				SET [Data]=@Data, UpdateDate=GETDATE()
-				WHERE [Key]=@Key AND RowVersion = ISNULL(@RowVersion, RowVersion)
-			END CATCH`,
+			IF (@FirstWrite=1)
+			BEGIN
+				BEGIN TRANSACTION;
+				IF EXISTS (SELECT * FROM [%s] WHERE [KEY]=@KEY)
+				BEGIN
+					THROW 2714, ''FIRST-WRITE: COMPETING RECORD ALREADY WRITTEN.'', 1
```
CodeMonkeyLeet, line 293: It's unusual that you're returning an error 2714 which is reserved for objects (i.e. tables, databases) which rows are not generally considered in SQL. 

I assume SQL would return 2601/2627 since that is also what this proc checks for to switch from insert to update, so either of those are probably more idiomatic errors to return. Given that this is a stored proc, it may also be more useful to `RAISERROR` with a specific error message that describes the Dapr-enforced first-write wins constraint here instead.

state/sqlserver/sqlserver.go
```diff
@@ -453,8 +453,8 @@ func (s *SQLServer) Delete(req *state.DeleteRequest) error {
 		return err
 	}
 
-	if rows != 1 {
-		return fmt.Errorf("items was not updated")
+	if req.ETag != nil && rows != 1 {
+		return state.NewETagError(state.ETagMismatch, nil)
```
CodeMonkeyLeet, line 457: The existing error check is still valid independent of ETags, so it may be useful to preserve the separation of errors. There is also a question of whether empty string ETags are considered mismatches in this case, since elsewhere in the code, it always collapses empty strings into the nil case.

```suggestion
	rows != 1 {
		if req.ETag != nil && *req.ETag != "" {
			return state.NewETagError(state.ETagMismatch, nil)
		} else {
			return fmt.Errorf("items was not updated")
		}
	}
```

state/sqlserver/sqlserver.go
```diff
@@ -578,15 +578,22 @@ func (s *SQLServer) executeSet(db dbExecutor, req *state.SetRequest) error {
...
 	etag := sql.Named(rowVersionColumnName, nil)
-	if req.ETag != nil {
+	if req.ETag != nil && *req.ETag != "" {
 		var b []byte
 		b, err = hex.DecodeString(*req.ETag)
 		if err != nil {
 			return state.NewETagError(state.ETagInvalid, err)
 		}
-		etag.Value = b
+		etag = sql.Named(rowVersionColumnName, b)
```
CodeMonkeyLeet, line 587: Nice catch!

.github/infrastructure/conformance/azure/setup-azure-conf-test.sh
berndverst: This line isn't necessary anymore for the conformance test because I hardcoded it to the Docker container connection string anyway.

But note that as of 2 years ago or so this environment variable is used to run SQL Server integration tests (take a look at `state/sqlserver/sqlserver_integration_test.go`-- I just used whatever was already being used there.

Of course ultimately you can always configure conformance tests differently by editing values in `tests/conformance/config`. The environment variables exposed there exist for our CI / GitHub actions purposes. So if you want to use a different environment variable yourself, `tests/conformance/config` is the place to do that.

go.mod
berndverst: Yes it absolutely is necessary. Please see the linked bug and discussion in Discord. Today the SQL Server component is broken with Go 1.16 and updating the module to this particular version (this is the commit corresponding to the latest tag) fixes it.

state/sqlserver/migration.go
```diff
@@ -275,32 +274,61 @@ func (m *migration) ensureUpsertStoredProcedureExists(db *sql.DB, mr migrationRe
...
+					RETURN
+				END
+				BEGIN
+					BEGIN TRY
+						INSERT INTO [%s] ([Key], [Data]) VALUES (@Key, @Data);
+					END TRY
+		
+					BEGIN CATCH
+						IF ERROR_NUMBER() IN (2601, 2627)
```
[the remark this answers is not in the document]
berndverst: I have no idea what they are. I am just maintaining this part of the existing stored procedure code to add a new special case for WRITE FIRST wins.

go.mod
berndverst: https://github.com/denisenkom/go-mssqldb/issues/688

state/sqlserver/migration.go
```diff
@@ -275,32 +274,61 @@ func (m *migration) ensureUpsertStoredProcedureExists(db *sql.DB, mr migrationRe
 		CREATE PROCEDURE %s (
 			@Key 			%s,
 			@Data 			NVARCHAR(MAX),
-			@RowVersion 	BINARY(8))
+			@RowVersion 	BINARY(8),
+			@FirstWrite		BIT)
 		AS
 			IF (@RowVersion IS NOT NULL)
```
CodeMonkeyLeet, line 280: This initial update if any rowversion is provided seems wrong. It basically short circuits everything else to try an update (which I believe can fail if the row doesn't already exist) requiring that the rowversion matches. Looking at the overall logical flow of the process, it seems like it should:

- Try to insert without using the etag
	- If it succeeds, key did not previously exist, and we correctly ignored an irrelevant etag if provided 
		- In the future, this should fail if not also first-write, but that is a breaking change incoming after https://github.com/dapr/dapr/issues/2739, option 1c.
	- Else, key exists and operation should be an update.
		- If first-write and Etag does not match, fail the operation
		- Else update the row with the optional Etag (via the ISNULL construction)

CodeMonkeyLeet (commented):
Thanks for all the hard work digging into this! The stored proc logic still seems iffy to me though, might want to check with the maintainers as to the expectations here.

state/sqlserver/migration.go
artursouza: What if FirstWrite=1 and RowVersion is NULL? It feels that FirstWrite check has higher priority over RowVersion.

state/sqlserver/migration.go
```diff
@@ -275,32 +274,61 @@ func (m *migration) ensureUpsertStoredProcedureExists(db *sql.DB, mr migrationRe
...
+						SET [Data]=@Data, UpdateDate=GETDATE()
+						WHERE [Key]=@Key AND RowVersion = ISNULL(@RowVersion, RowVersion)
+					END CATCH
+				END
+				COMMIT;
+			END
+			ELSE
+			BEGIN
+				BEGIN TRY
+					INSERT INTO [%s] ([Key], [Data]) VALUES (@Key, @Data);
```
artursouza, line 313: In this case, FirstWrite=0 but RowVersion can be NULL or not, so it must be handled here too.

state/sqlserver/sqlserver.go
artursouza: I agree, if an updated was expected without eTag but nothing happened, it is also an error.

artursouza (changes requested):
Great progress. It is almost there.

tests/config/state/sqlserver/statestore.yml
```diff
@@ -0,0 +1,12 @@
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: statestore
+spec:
+  type: state.sqlserver
+  metadata:
+  - name: connectionString
+    value: "server=localhost;user id=sa;password=Pass@Word1;port=1433;"
+    # use ${{DAPR_TEST_SQL_CONNSTRING}} when not using Docker
```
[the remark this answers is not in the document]
berndverst: already done before you commented

state/sqlserver/sqlserver_integration_test.go
```diff
@@ -242,12 +242,12 @@ func testSingleOperations(t *testing.T) {
 			assert.NotNil(t, err)
 			_, etag := assertLoadedUserIsEqual(t, store, johnV3.ID, johnV3)
 
-			// 9. Delete with invalid ETAG should fail
+			// 9. Delete with invalid ETAG
```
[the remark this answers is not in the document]
berndverst: that was already reverted before you commented

.github/infrastructure/conformance/azure/setup-azure-conf-test.sh
```diff
@@ -415,3 +415,9 @@ echo "INFO: setup-azure-conf-test completed."
 echo "INFO: Remember to \`source ${ENV_CONFIG_FILENAME}\` before running local conformance tests."
 echo "INFO: ${AZURE_CREDENTIALS_FILENAME} contains the repository secret to set to run the GitHub conformance test workflow."
 echo "INFO: To teardown the conformance test resources, run ${TEARDOWN_SCRIPT_NAME}."
+
+# ----------------------------------
+# Populate SQLServer test settings
+# ----------------------------------
+echo "Configuring SQL Server test settings ..."
```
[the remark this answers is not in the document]
berndverst: yup and again that was reverted before you commented here

state/sqlserver/migration.go
```diff
@@ -275,32 +274,61 @@ func (m *migration) ensureUpsertStoredProcedureExists(db *sql.DB, mr migrationRe
 		CREATE PROCEDURE %s (
 			@Key 			%s,
 			@Data 			NVARCHAR(MAX),
-			@RowVersion 	BINARY(8))
+			@RowVersion 	BINARY(8),
+			@FirstWrite		BIT)
```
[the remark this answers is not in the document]
berndverst: SQL Server prefers using Bit to represent Boolean from my research!

state/sqlserver/migration.go
berndverst: RAISERROR did not work at all the way the stored procedure is created here. It did work in a SQL editor I was using but over 2 hours I could not work around that. THROW works fine -- I can change to any error code that is desirable.

state/sqlserver/migration.go
berndverst: @CodeMonkeyLeet I cannot manually set an Etag. The Etag is added by the stored procedure only. Any attempt to manually set an etag miserable failed in that the upsert was completely ignored -- I spent one full day just on that.

Does that change what you are suggesting here?

CodeMonkeyLeet: Interpret ETag as RowVersion in my comment, it's just an input into the method. I'm not suggesting adding an ETag where one was not provided in the request.

state/sqlserver/migration.go
berndverst: The problem is that the stored procedure and DB has for 2 years always set a Rowversion / Etag. So Rowversion cannot actually be null in the DB, only in the request.

What I observed happening with the existing stored procedure was that any attempt to update a Key without also specifying the correct Rowversion failed (was ignored as NO-OP), unless the Rowversion was NULL. When the existing ROWVERSION was provided the record was updated and the ROWVERSION was incremeneted.

state/sqlserver/sqlserver.go
berndverst: I specifically asked you in a meeting about this behavior @artursouza - whether a deletion of a record which doesn't exist can be a no-op.

The reason being that the conformance tests attempt to delete a bunch of records that were not actually inserted but expect those deletions to succeed (no op I guess)

berndverst: The Delete test depends on the Set test, and the set test inserts a different set then is being deleted in the delete test.

https://github.com/dapr/components-contrib/blob/master/tests/conformance/state/state.go#L207

https://github.com/dapr/components-contrib/blob/master/tests/conformance/state/state.go#L237

Clearly the other components are not throwing an error when a record is deleted which does not exist!

berndverst: I see two choices -- either those conformance tests need to be fixed to expect failures of deletions which do not exist (I would prefer this), or this integration needs to be updated not to expect a failure when a deletion is made with an invalid ETAG: https://github.com/dapr/components-contrib/blob/master/state/sqlserver/sqlserver_integration_test.go#L245

berndverst: Any preferences here @artursouza @CodeMonkeyLeet ?

I wonder what the fallout is if we correct the conformance test.

state/sqlserver/migration.go
```diff
@@ -275,32 +274,61 @@ func (m *migration) ensureUpsertStoredProcedureExists(db *sql.DB, mr migrationRe
 		CREATE PROCEDURE %s (
 			@Key 			%s,
 			@Data 			NVARCHAR(MAX),
-			@RowVersion 	BINARY(8))
+			@RowVersion 	BINARY(8),
```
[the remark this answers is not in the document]
berndverst: Good idea

state/sqlserver/sqlserver.go
CodeMonkeyLeet: Changing the test suggests a breaking change in the error contract with the app (conformance tests are supposed to check adherence to it after all).

The problem in the existing SQL server implementation specifically is that it doesn't differentiate between the delete failing because:
- the item did not exist (no-op) and 
- the item exists but delete targeted the wrong version (dapr wants to error on the ETag mismatch).

If the code only checks the presence of ETag on failed delete, it could be a false negative (failure could be because the item did not exist, ETag or otherwise). Instead of checking for the deleted rows, is there a try-catch mechanism available to determine the cause of failure?

state/sqlserver/sqlserver_integration_test.go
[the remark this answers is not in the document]
CodeMonkeyLeet: Did you push your changes? They're still showing up.

state/sqlserver/migration.go
berndverst: I agree with you both now @CodeMonkeyLeet @artursouza. Sorry for my confusion.

**I incorrectly thought the First Write concept only applied to database records that do not exist in the database.**

state/sqlserver/migration.go
berndverst: Ignore my comment - I was confused about the First Write behavior. Will update.

state/sqlserver/sqlserver.go
berndverst: Thanks for taking the time to have a call with me @CodeMonkeyLeet.

I'll summarize the conversation / these tests again here:

The first think we do is throw any actual database level error returned. Now it's important to note that SQLServer doesn't throw an error when a key to be deleted doesn't exist.

Where the key doesn't exist we receive a success but with rows Affected = 0. The conformance tests expect a Delete Request where an item did not exist to succeed (return nil for the error). So in the case where we don't have an ETag we can say that rows = 0 or rows = 1 are acceptable states and therefore we return `nil`. Due to the primary key constraint we can't possibly have more rows affected, so therefore we can always return `nil` if there was no database error and no ETag provided.

In the case where an ETag is provided the question is whether a deletion which did not actually remove a row (because the key didn't exist) is an error. I believe yes -- and an existing integration test (`state/sqlserver/sqlserver_integration_test.go`) specifically tests for this error. For this particular reason I'm throwing an ETag error in the case where the tag was provided but no record delete (but again no failure from the database itself).

I believe this is the correct behavior - and the conformance tests and the existing integration test would agree.

state/sqlserver/sqlserver_integration_test.go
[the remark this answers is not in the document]
berndverst: I was dumb and added the `should fail` to test #10 when I meant to add it to test #9. Adding it back now.

berndverst:
@artursouza sorry for all the confusion here. I believe I was able to address all concerns and was also able to simplify some error handling.

Please note that the code changed can't meaningfully be tested via unit tests -- there are no similar unit tests today. There are however integration tests at `state/sqlserver/sqlserver_integration_test.go` which can be run by setting the environment variable I mentioned. Unfortunately CODE COV has no awareness of these tests so the drop in test coverage isn't accurate.

state/sqlserver/migration.go
berndverst: I tried RAISEERROR again but by design this doesn't exit the SQL execution flow. SQL Server engineers recommend using THROW.

state/sqlserver/migration.go
berndverst: Please take a look at my latest change on this. I think I am handling it correctly now -- and conformance tests and integration tests pass.

artursouza (dismissed):
LGTM.  Thanks for revisiting the logic in the stored procedure. Please, wait for approval from @CodeMonkeyLeet  too.

state/sqlserver/migration.go
```diff
@@ -272,35 +271,85 @@ func (m *migration) createStoredProcedureIfNotExists(db *sql.DB, name string, es
...
+			AS
+				IF (@FirstWrite=1)
+					BEGIN
+						IF (@RowVersion IS NOT NULL)
+							BEGIN
+								BEGIN TRANSACTION;
+								IF NOT EXISTS (SELECT * FROM [%s] WHERE [KEY]=@KEY AND RowVersion = @RowVersion)
+									BEGIN
+										THROW 2601, ''FIRST-WRITE: COMPETING RECORD ALREADY WRITTEN.'', 1
+									END
```
CodeMonkeyLeet, line 288: This is not always the case though: this case can also be because the row was never written before and the key does not exist.

state/sqlserver/sqlserver.go
```diff
@@ -440,23 +440,23 @@ func (s *SQLServer) Delete(req *state.DeleteRequest) error {
...
+	// err represents errors thrown by the stored procedure or the database itself
 	if err != nil {
-		if req.ETag != nil {
-			return state.NewETagError(state.ETagMismatch, err)
-		}
-
 		return err
 	}
 
+	// if the key is not found, then the stored procedure returns 0 rows affected
```
CodeMonkeyLeet, line 448:
```suggestion
	// if the row with matching key (and ETag if specified) is not found, then the stored procedure returns 0 rows affected
```

state/sqlserver/sqlserver.go
```diff
@@ -440,23 +440,23 @@ func (s *SQLServer) Delete(req *state.DeleteRequest) error {
...
 	rows, err := res.RowsAffected()
 	if err != nil {
 		return err
 	}
 
-	if rows != 1 {
-		return fmt.Errorf("items was not updated")
+	// When an ETAG is specified, a row must have been deleted or else we return an ETag mismatch error
+	if req.ETag != nil && rows != 1 {
```
CodeMonkeyLeet, line 455: Should this also treat empty string req.ETag as nil ETag for consistency with executeSet?
```suggestion 
	if rows != 1 && req.ETag != nil && *req.ETag != "" {
```

state/sqlserver/migration.go
```diff
@@ -272,35 +271,85 @@ func (m *migration) createStoredProcedureIfNotExists(db *sql.DB, name string, es
...
-				INSERT INTO [%s] ([Key], [Data]) VALUES (@Key, @Data);
-			END TRY
-
-			BEGIN CATCH
-				IF ERROR_NUMBER() IN (2601, 2627)
-				UPDATE [%s]
-				SET [Data]=@Data, UpdateDate=GETDATE()
-				WHERE [Key]=@Key AND RowVersion = ISNULL(@RowVersion, RowVersion)
-			END CATCH`,
+			CREATE PROCEDURE %s (
```
CodeMonkeyLeet, line 274: Overall, I think the logic here is still problematic. I understand the desire not to churn existing code, but it looks like there are still issues relative to the expected behaviors expressed in #2739 1c. Condensed here for reference:

StateOptions.Concurrency | Req.Etag | Expected Behavior in State Store
-- | -- | --
CONCURRENCY_FIRST_WRITE | not nil | Write if etag matches
CONCURRENCY_FIRST_WRITE | nil | Write if key does not exist
CONCURRENCY_LAST_WRITE or unspecified | not nil | Return error (Upcoming breaking change. Today, state components write regardless of etag)
CONCURRENCY_LAST_WRITE or unspecified | nil | Write regardless of etag

"Write" in this case refers to an upsert (i.e. insert key if it does not already exist, otherwise update). The original code ignores the top half of that table, but it also looks like it already did not implement the bottom half correctly. Please correct any readings of the code here: 

- If an ETag is provided and the key does not exist, it attempts to update, which returns 0 rows modified but does not error and does not insert the value. It is silently lost.
- If an ETag is provided and the key does exist, it will attempt to update using the ETag. If the ETag does not match, it again does not error and does not update the value, which has FIRST_WRITE semantics instead of LAST_WRITE (default) semantics. It should ignore the ETag completely in this case.

Preserving the original logic propagates these issues forwards. The new logic for handling FIRST_WRITE works, although it raises some questions about the specification for behavior:

- If ETag provided and key does not exist, ETag can never match. Should the method raise an error or silently ignore the write? 
    - This implementation returns an error
- If ETag provided and key exists, but ETag does not match, should the method raise an error or silently ignore the write?
    - This implementation silently ignores the update. (The update on row 290 finds no matches).  

That seems potentially inconsistent as an error contract for Dapr.

@artursouza To keep this moving forwards, since you've already approved the PR and the new code is okay, I suggest we file a separate issue to track the logical inconsistencies in the existing code. In continuing the discussion of Dapr ETag/concurrency state handling, I'll also follow up on the issue with my questions on the expected error handling.

CodeMonkeyLeet (commented):
See extended discussion on existing questions on error state handling. I'm okay with taking the PR, modulo tracking ongoing issues in the concurrency handling contract in Dapr.

state/sqlserver/migration.go
berndverst: That's an excellent point. What do you suggest? Changing the error message?

If the record does not exist it's not actually possible to specify an ETAG (Row Version) to be written -- the stored procedure / DB won't allow the use of ETAGs other than to update an existing record.

state/sqlserver/sqlserver.go
berndverst: Yes probably.

state/sqlserver/migration.go
artursouza: The issue for LAST_WRITE and eTag not null is behavior change beyond the scope of this change because it is a broader concern of the expectations for the state API behavior. I am OK with the current logic in this component.

The error handling for silent etag mistmatch in the stored procedure is handled in code by checking number of rows affected.

@CodeMonkeyLeet Let me know if you have any other concerns on this PR.

state/sqlserver/migration.go
CodeMonkeyLeet: Yeah, I would just change the error message to something more general

state/sqlserver/migration.go
CodeMonkeyLeet: Nope, I think it's definitely a step forward and we should take it, thanks for reminding me about the downstream handling of returning errors from the zero rows updated check.

artursouza (approved):

Issue #1080, opened 2021-08-16
Update S3 binding


fjvela:
## Describe the feature
Update S3 binding:

- update create operation: support upload files in base64 (https://docs.dapr.io/reference/components-reference/supported-bindings/s3/#upload-a-file doesn't work), return location and version id 
- add get operation
- add delete operation
- add list operation

## Release Note

RELEASE NOTE:
- **UPDATE** S3 binding. Create operation: support upload files in base64, return location and version id 
- **ADD**  S3 binding: get operation
- **ADD**  S3 binding: delete operation
- **ADD**  S3 binding: list operation

CodeMonkeyLeet:
@tanvigour @artursouza Target of opportunity for v1.4 release?

tanvigour:
> @tanvigour @artursouza Target of opportunity for v1.4 release?

looks like there's already a PR for this, so yeah we can add it to 1.4

tanvigour:
> > @tanvigour @artursouza Target of opportunity for v1.4 release?
> 
> looks like there's already a PR for this, so yeah we can add it to 1.4

I don't have rights to can @artursouza add it to 1.4?

artursouza:
I made some small comments in the PR. For community contributions, we add to the milestone when it is about to be merged because we cannot predict when the PR will be merged.

PR #1081, opened 2021-08-16
Update S3 binding

fjvela:
# Description

Update S3 binding:

- update create operation: support upload files in base64  (https://docs.dapr.io/reference/components-reference/supported-bindings/s3/#upload-a-file doesn't work), return location and version id 
- add get operation
- add delete operation
- add list operation

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/components-contrib/issues/1080

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: https://github.com/dapr/docs/issues/1722


bindings/aws/s3/s3.go
```diff
@@ -44,39 +71,168 @@ func (s *AWSS3) Init(metadata bindings.Metadata) error {
 	if err != nil {
 		return err
 	}
-	uploader, err := s.getClient(m)
+	session, err := s.getSession(m)
```
artursouza, line 74: Please, also implement the Close() method.

bindings/aws/s3/s3.go
```diff
@@ -44,39 +71,168 @@ func (s *AWSS3) Init(metadata bindings.Metadata) error {
...
+	case bindings.CreateOperation:
+		return s.create(req)
+	case bindings.GetOperation:
+		return s.get(req)
+	case bindings.DeleteOperation:
+		return s.delete(req)
+	case bindings.ListOperation:
+		return s.list(req)
+	default:
+		return nil, fmt.Errorf("s3 binding error. unsupported operation %s", req.Operation)
```
artursouza, line 232: Please, default to CreateOperation for backwards compatibility.

bindings/aws/s3/s3.go
```diff
@@ -44,39 +71,168 @@ func (s *AWSS3) Init(metadata bindings.Metadata) error {
...
+	_, err := s.downloader.Download(buff,
+		&s3.GetObjectInput{
+			Bucket: aws.String(s.metadata.Bucket),
+			Key:    aws.String(key),
+		})
+	if err != nil {
+		return nil, fmt.Errorf("s3 binding error: error downloading S3 object: %w", err)
+	}
+
+	return &bindings.InvokeResponse{
```
artursouza, line 164: Shouldn't GET also have an option to base64 encode the response?

artursouza (changes requested):
LGTM overall. Just some small comments.

daixiang0:
@fjvela please sync master, thanks.

yaron2:
@fjvela please look at the comments made by @artursouza

bindings/aws/s3/s3.go
fjvela: Added, but it's an empty method. I didn't found any method to close AWS session

bindings/aws/s3/s3.go
fjvela: I think it isn’t necessary.

When you invoke the binding, operation info is required. If you invoke the binding and the operation isn’t defined in bindings.OperationKind method you will get an error.

If you add a new operation in bindings.OperationKind method and you don’t implement it in Invoke method I think it’s better to get an error instead of execute Create operation

artursouza: This component is still in Alpha, which means breaking changes are allowed. Making it backwards compatible is not necessary but we still avoid breaking existing users unnecessarily.

To favor existing customer experience, I think we should make this backwards compatible. It is not a big change.

/cc @pkedy @CodeMonkeyLeet thoughts?

fjvela: Yes, it's an easy change but not needed.  The changes are compatible and we aren't going to break any application:


All request must include operation field if you don't include it you will get an error also you will get an error if you try invoke and operation not defined in bindings.OperationKind method.

image

artursouza: OK. I got it. It is being enforced prior to the component code.

yaron2:
@fjvela there's a test failing.

bindings/aws/s3/s3_test.go
```diff
@@ -9,21 +9,135 @@ import (
...
+		_, err := s3.get(&r)
+		assert.Error(t, err)
+	})
+}
+
+func TestDeleteOption(t *testing.T) {
+	s3 := NewAWSS3(logger.NewLogger("s3"))
+
+	t.Run("return error if key is missing", func(t *testing.T) {
+		r := bindings.InvokeRequest{}
```
yaron2, line 139: Metadata fix also needed here :)

yaron2 (changes requested):

yaron2:
@artursouza I'm including this in `1.4`.

artursouza (approved):

Issue #894, opened 2021-05-27
Azure EventHubs binding: Remove "close by interrupt signal" logic


pkedy:
## Expected Behavior

This component has a client to cleanup.  The Dapr runtime now calls `Close() error` on any component that implements `io.Closer`. Logic to cleanup should be moved there.

## Actual Behavior

Currently it uses a signal channel to wait to close itself.

RELEASE NOTE: **FIX** Azure EventHubs binding implements `Close()` for cleanup

CodeMonkeyLeet:
Closing this issue as the addition and the move of logic into `Close() error` is handled by #1093.

Using https://github.com/dapr/dapr/issues/3555 to track the broader question of replacing wait on SIGTERM with appropriate cancellation contexts.

PR #1093, opened 2021-08-18
Add Close() method to EventHubs bindings

CodeMonkeyLeet:
# Description

Add the Close() method that can trigger the cleanup of the Hub object.

## Issue reference

Addresses part of #894

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* ~~[ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_~~

### DIFF
--- a/bindings/azure/eventhubs/eventhubs.go
+++ b/bindings/azure/eventhubs/eventhubs.go
@@ -238,7 +238,7 @@ func (a *AzureEventHubs) Read(handler func(*bindings.ReadResponse) ([]byte, erro
 	signal.Notify(exitChan, os.Interrupt, syscall.SIGTERM)
 	<-exitChan
 
-	a.hub.Close(context.Background())
+	a.Close()
 
 	return nil
 }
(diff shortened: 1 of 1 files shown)


CodeMonkeyLeet:
@pkedy For discussion:

This change only does the simple part of adding a Close() method that daprd expects to be able to invoke, and will free the `Hub` object. 

For the desire expressed in the title of #894 to remove the "close by interrupt signal logic", this is a pattern that's used fairly commonly across a number of other bindings (rocketmq, aws kinesis, storage queues kafka, kubernetes, mqtt, twitter, zeebee) to make the Read() call blocking when the underlying component SDK is not. This is more generally described in https://github.com/dapr/dapr/issues/3555.

To address it for EventHubs would be to create a component owned context that can be cancelled in Close() and to wait on that instead of the sigterm. It's a deeper change though and we might want to do all the bindings at once if that is the case; specifically, the existing conformance tests don't really validate this cancellation semantic and so that feels like it should be a prerequisite to this work. 

Thoughts?

PR #1102, opened 2021-08-27
ci: standard linter config

daixiang0:
Signed-off-by: Long <long0dai@foxmail.com>

# Description

Ref https://github.com/dapr/dapr/issues/3537

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


.golangci.yml
```diff
@@ -1,231 +1,247 @@
...
-    - lll
-    - scopelint
-    - unparam
-    - wsl
-    - gomnd
-    - godot
-    - testpackage
-    - goerr113
-    - gci
+# options for analysis running
```
CodeMonkeyLeet, line 1: Did you change all the line endings in the file as well? For future reference, it's helpful to separate widespread formatting changes from code changes into separate commits so that the diff tool can highlight the more important code changes in one of the commits.

CodeMonkeyLeet (commented):
LGTM, basically applying the dapr/dapr/.golangci.yml and then fixing all the static analysis checks, correct? 

We probably want to try and coordinate the merge for this since it touches so many files and any new PR merge has the potential to require new lint fixes. @artursouza Do we have a window after codefreeze for v1.4 to push this set of PRs?

.golangci.yml
daixiang0: I use `dos2unix` :)

CodeMonkeyLeet: Good to know! So the advice is that next time, your should do a commit after running `dos2unix` on the file, then do your code changes in a separate commit so that a reviewer can inspect the code changes independent from the formatting changes in the GitHub PR view.

daixiang0: Sure.

bindings/azure/storagequeues/storagequeues_test.go
```diff
@@ -221,7 +222,7 @@ func TestReadQueueDecode(t *testing.T) {
 
 	time.Sleep(30 * time.Second)
 
-}
+}.
```
artursouza, line 225: Really? a dot here? Is there a way to skip lint for this case?

artursouza (changes requested):
When the comment is a command, JSON or XML, then we want the content to be as-is and not have a period. Is there a way to disable lint for those?

bindings/azure/storagequeues/storagequeues_test.go
daixiang0: Fixed.

daixiang0:
> When the comment is a command, JSON or XML, then we want the content to be as-is and not have a period. Is there a way to disable lint for those?

Add `//nolint`, details see https://golangci-lint.run/usage/false-positives/#nolint.

artursouza (approved):

PR #1106, opened 2021-08-30
feat(pubsub): add in-memory pub/sub

riezebosch:
# Description

Added an adapter for a hand-picked in-memory service bus implementation to support integration testing. In [this example](https://github.com/riezebosch/dapr-tdd/blob/main/DaprDemo.Api.IntegrationTests/DemoControllerTests.cs#L48), I use RabbitMQ as the underlying message broker but that makes it hard(er) to run integration tests in CI since you need to first spin it up from the pipeline.

Next thing I'll probably create is a C# wrapper to start and stop the self hosted sidecar from the integration test (and call it Dapr.Wrapr 🤟🏻 ).

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/issues/1761

### DIFF
--- a/pubsub/in-memory/in-memory.go
+++ b/pubsub/in-memory/in-memory.go
@@ -1,0 +1,56 @@
+package inmemory
+
+import (
+	"context"
+
+	"github.com/asaskevich/EventBus"
+	"github.com/dapr/components-contrib/pubsub"
+	"github.com/dapr/kit/logger"
+)
+
+type bus struct {
+	bus EventBus.Bus
+	ctx context.Context
+	log logger.Logger
+}
+
+func New(logger logger.Logger) pubsub.PubSub {
+	return &bus{
+		log: logger,
+	}
+}
+
+func (a *bus) Close() error {
+	return nil
+}
+
+func (a *bus) Features() []pubsub.Feature {
+	return nil
+}
+
+func (a *bus) Init(metadata pubsub.Metadata) error {
+	a.bus = EventBus.New()
+	a.ctx = context.Background()
+
+	return nil
+}
+
+func (a *bus) Publish(req *pubsub.PublishRequest) error {
+	a.bus.Publish(req.Topic, a.ctx, req.Data)
+
+	return nil
+}
+
+func (a *bus) Subscribe(req pubsub.SubscribeRequest, handler pubsub.Handler) error {
+	return a.bus.Subscribe(req.Topic, func(ctx context.Context, data []byte) {
+		for i := 0; i < 10; i++ {
+			if err := handler(ctx, &pubsub.NewMessage{Data: data, Topic: req.Topic, Metadata: req.Metadata}); err != nil {
+				a.log.Error(err)
+
+				continue
+			}
+
+			return
+		}
+	})
+}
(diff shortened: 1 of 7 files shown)


yaron2:
That's a very good addition that might help lots of folks to get started locally without the Redis docker container.

pubsub/in-memory/in-memory.go
```diff
@@ -0,0 +1,44 @@
...
+	return nil
+}
+
+func (a *InMemoryBus) Publish(req *pubsub.PublishRequest) error {
+	a.bus.Publish(req.Topic, a.ctx, req.Data)
+	return nil
+}
+
+func (a *InMemoryBus) Subscribe(req pubsub.SubscribeRequest, handler pubsub.Handler) error {
+	a.bus.SubscribeAsync(req.Topic, func(ctx context.Context, data []byte) {
```
CodeMonkeyLeet, line 40: `SubscribeAsync` does return an error, so it's probably useful to check for that and handle it (even if you expect as an invariant that the call should never error today)

CodeMonkeyLeet (commented):
Thanks for this contribution, it looks good!

Could you also open an issue in dapr/docs for tracking documentation of this component, per the PR template? 

Also, while it's not an alpha component requirement, I generally encourage folks to add new components to the [conformance tests](https://github.com/dapr/components-contrib/blob/master/tests/conformance/README.md). It should be really easy in this case since there's no backing service to configure, and you can try it locally by updating the `tests/conformance/common.go` and `tests/config/pubsub/tests.yml` files to include this, and add the trivial `tests/config/pubsub/in-memory/pubsub.yml` config for it.

pubsub/in-memory/in-memory.go
riezebosch: I'm not really sure I need the `Async` version of the method since both work fine when I load the component into dapr, but the other variant is blocking the tests and I couldn't figure it out using goroutines.

riezebosch:
> Thanks for this contribution, it looks good!

Thanks!

> Could you also open an issue in dapr/docs for tracking documentation of this component, per the PR template?

✅ 

> Also, while it's not an alpha component requirement, I generally encourage folks to add new components to the [conformance tests](https://github.com/dapr/components-contrib/blob/master/tests/conformance/README.md). It should be really easy in this case since there's no backing service to configure, and you can try it locally by updating the `tests/conformance/common.go` and `tests/config/pubsub/tests.yml` files to include this, and add the trivial `tests/config/pubsub/in-memory/pubsub.yml` config for it.

✅

CodeMonkeyLeet:
Thanks for adding the component to conformance tests as well, the changes LGTM.

Automation is catching linter errors though, could you address them? The linter errors will show up inline under the Files changed tab in this PR.

riezebosch:
Done, plus updated from upstream.

CodeMonkeyLeet (approved):
LGTM! Thanks again!

riezebosch:
Great! I accidentally removed the maintaoners from the reviewers 🙈 How will it now be closed?

pkedy (approved):

artursouza (approved):

yaron2:
@artursouza do we want to add this to 1.4?

artursouza:
Yes, this should be on 1.4.

PR #1107, opened 2021-08-31
feature/pubsub: add delay queue params for pulsar

1046102779:


pubsub/pulsar/pulsar.go
```diff
@@ -123,16 +123,30 @@ func (p *Pulsar) Publish(req *pubsub.PublishRequest) error {
...
 	return nil
 }
 
+// parsePublishMetadata parse publish metadata
+func parsePublishMetadata(req *pubsub.PublishRequest) (msg *pulsar.ProducerMessage) {
+	msg = &pulsar.ProducerMessage{
+		Payload: req.Data,
+	}
+	if val, ok := req.Metadata["deliver_at"]; ok {
+		msg.DeliverAt, _ = time.Parse("2006-01-02 15:04:05", val)
```
CodeMonkeyLeet, line 140: Is there a reason why this needs a custom time format? If not, I would recommend standardizing on the `time.RFC3339` format as most existing components use.

Also, do not ignore the parse error, the method should fail with an error on invalid inputs.

pubsub/pulsar/pulsar.go
```diff
@@ -123,16 +123,30 @@ func (p *Pulsar) Publish(req *pubsub.PublishRequest) error {
...
 
 	return nil
 }
 
+// parsePublishMetadata parse publish metadata
+func parsePublishMetadata(req *pubsub.PublishRequest) (msg *pulsar.ProducerMessage) {
+	msg = &pulsar.ProducerMessage{
+		Payload: req.Data,
+	}
+	if val, ok := req.Metadata["deliver_at"]; ok {
```
CodeMonkeyLeet, line 139: I would follow the metadata naming convention in Dapr for `camelCased` metadata property names, so `deliverAt` and `deliverAfter`.

pubsub/pulsar/pulsar.go
```diff
@@ -123,16 +123,30 @@ func (p *Pulsar) Publish(req *pubsub.PublishRequest) error {
...
+		msg.DeliverAt, _ = time.Parse("2006-01-02 15:04:05", val)
+	}
+	if val, ok := req.Metadata["deliver_after"]; ok {
+		tmp, _ := strconv.Atoi(val)
+		msg.DeliverAfter = time.Duration(tmp)
```
CodeMonkeyLeet, line 144: Why are we not using `time.ParseDuration(val)` here instead? It seems very unfriendly to the dev to require that they calculate the integrate an integer duration instead of Go-standard duration strings like "3h" used elsewhere in Dapr.

CodeMonkeyLeet (commented):
Thanks for submitting this extra support for Pulsar pubsub! 

In addition to the code comments, can you also make sure to open an issue in dapr/docs to track adding documentation for how to use the new publish metadata fields?

CodeMonkeyLeet:
Closing as superseded by #1112.

PR #1112, opened 2021-09-01
feature/pubsub: add delay queue params for pulsar

1046102779:
[PR 1107](https://github.com/dapr/components-contrib/pull/1107)


1046102779:
[docs PR1759](https://github.com/dapr/docs/pull/1759)

1046102779:
@CodeMonkeyLeet help me check this pr, I recently need to use this feature. thks

pubsub/pulsar/pulsar.go
```diff
@@ -123,16 +125,29 @@ func (p *Pulsar) Publish(req *pubsub.PublishRequest) error {
...
 	return nil
 }
 
+// parsePublishMetadata parse publish metadata
+func parsePublishMetadata(req *pubsub.PublishRequest) (msg *pulsar.ProducerMessage) {
+	msg = &pulsar.ProducerMessage{
+		Payload: req.Data,
+	}
+	if val, ok := req.Metadata[deliverAt]; ok {
+		msg.DeliverAt, _ = time.Parse("2006-01-02 15:04:05", val)
```
CodeMonkeyLeet, line 142: The https://github.com/dapr/components-contrib/pull/1107#discussion_r699544119 still applies:

- Is there a reason to use a custom time format instead of `time.RFC3339`? 
- Do not ignore the errors returned from `time.Parse` and `time.ParseDuration`. The function signature should be updated to also return the error, which can be checked and returned by the calling `Publish()` function.

```go
func parsePublishMetadata(req *pubsub.PublishRequest) (*pulsar.ProducerMessage, error) 
```

CodeMonkeyLeet (changes requested):
This is a direct replacement for #1107, correct? I will close that PR if that is the case.

1046102779:
> This is a direct replacement for #1107, correct? I will close that PR if that is the case.

Yes, you can close that PR

pubsub/pulsar/pulsar.go
1046102779: updated.

pubsub/pulsar/pulsar.go
```diff
@@ -123,16 +126,39 @@ func (p *Pulsar) Publish(req *pubsub.PublishRequest) error {
...
+// parsePublishMetadata parse publish metadata
+func parsePublishMetadata(req *pubsub.PublishRequest) (
+	msg *pulsar.ProducerMessage, err error) {
+	msg = &pulsar.ProducerMessage{
+		Payload: req.Data,
+	}
+	if val, ok := req.Metadata[deliverAt]; ok {
+		msg.DeliverAt, err = time.Parse(time.RFC3339, val)
+		if err != nil {
+			return
```
CodeMonkeyLeet, line 149: The standard practice in go is to not return incomplete objects on error, so in this case the methods should `return nil, err`.

CodeMonkeyLeet (approved):
LGTM, thanks!

pubsub/pulsar/pulsar_test.go
```diff
@@ -17,6 +18,21 @@ func TestParsePulsarMetadata(t *testing.T) {
...
+func TestParsePublishMetadata(t *testing.T) {
+	m := &pubsub.PublishRequest{}
+	m.Metadata = map[string]string{
+		"deliverAt":    "2021-08-31T11:45:02Z",
+		"deliverAfter": "60s",
+	}
+	msg, err := parsePublishMetadata(m)
+	assert.Nil(t, err)
+
+	val, _ := time.ParseDuration("60s")
```
daixiang0, line 30: Could we use things like `60 * time.minutes`?

1046102779: Yes, 60m

daixiang0: seconds, sorry.

artursouza (approved):

Issue #1117, opened 2021-09-02
add state conformance test for postgresql


beiwei30:
## Expected Behavior

add state conformance test for postgresql (dapr/dapr#3533)

## Release Note

RELEASE NOTE:

PR #1118, opened 2021-09-02
add state conformance test for postgresql #1117

beiwei30:
# Description

add state conformance test for postgresql #1117 (dapr/dapr#3533)

## Issue reference

Please reference the issue this PR will close: #1117

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/.github/infrastructure/docker-compose-postgresql.yml
+++ b/.github/infrastructure/docker-compose-postgresql.yml
@@ -1,0 +1,11 @@
+version: '2'
+services:
+  db:
+    image: postgres
+    restart: always
+    ports:
+      - "5432:5432"
+    environment:
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: example
+      POSTGRES_DB: dapr_test
--- a/tests/config/state/tests.yml
+++ b/tests/config/state/tests.yml
@@ -9,3 +9,5 @@ components:
     allOperations: true
   - component: sqlserver
     allOperations: true
+  - component: postgresql
+    allOperations: true
--- a/tests/conformance/common.go
+++ b/tests/conformance/common.go
@@ -51,6 +51,7 @@ import (
 	ss_local_file "github.com/dapr/components-contrib/secretstores/local/file"
 	s_cosmosdb "github.com/dapr/components-contrib/state/azure/cosmosdb"
 	s_mongodb "github.com/dapr/components-contrib/state/mongodb"
+	s_postgresql "github.com/dapr/components-contrib/state/postgresql"
 	s_redis "github.com/dapr/components-contrib/state/redis"
 	s_sqlserver "github.com/dapr/components-contrib/state/sqlserver"
 	conf_bindings "github.com/dapr/components-contrib/tests/conformance/bindings"
@@ -369,6 +370,8 @@ func loadStateStore(tc TestComponent) state.Store {
 		store = s_mongodb.NewMongoDB(testLogger)
 	case "sqlserver":
 		store = s_sqlserver.NewSQLServerStateStore(testLogger)
+	case "postgresql":
+		store = s_postgresql.NewPostgreSQLStateStore(testLogger)
 	default:
 		return nil
 	}
--- a/tests/config/state/postgresql/statestore.yml
+++ b/tests/config/state/postgresql/statestore.yml
@@ -1,0 +1,11 @@
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: statestore
+spec:
+  type: state.postgresql
+  metadata:
+    - name: connectionString
+      value: "host=localhost user=postgres password=example port=5432 connect_timeout=10 database=dapr_test"
+    - name: actorStateStore
+      value: "true"


.github/infrastructure/docker-compose-postgresql.yml
```diff
@@ -0,0 +1,11 @@
...
+services:
+  db:
+    image: postgres
+    restart: always
+    ports:
+      - "5432:5432"
+    environment:
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: example
+      POSTGRES_DB: dapr_test
```
CodeMonkeyLeet, line 11: Nit: add newline at end of file for Linux text files.

.github/infrastructure/docker-compose-postgresql.yml
```diff
@@ -0,0 +1,11 @@
+version: '2'
+services:
+  db:
+    image: postgres
+    restart: always
+    ports:
+      - "5432:5432"
+    environment:
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: example
```
CodeMonkeyLeet, line 10: It's not really to big a deal since this is just for test, but I don't know if there's a security guideline for not including passwords in committed config files. @artursouza, do you have an opinion?

If you want an example for dynamically populating this values with a temporary secret, https://github.com/dapr/components-contrib/pull/1115 also has an example on how to do that.

CodeMonkeyLeet (commented):
This looks good, thanks for adding the test as well!

Could you also include it into the conformance test workflow? See https://github.com/dapr/components-contrib/pull/1115/files file .github/workflows/conformance.yml for an example. This will allow the PR automation to test your change as well.

beiwei30:
I also try to fix postgresql's state to pass the conformance test. Now postgresql can pass all operations except for 'first-write'.  Would you mind to review it too?

Besides, I believe mysql's state cannot pass the test either. I will follow up by filing a new issue later.

.github/infrastructure/docker-compose-postgresql.yml
beiwei30: done.

.github/infrastructure/docker-compose-postgresql.yml
beiwei30: I think it's fine since it's for test only. The hardcoded plain-text password also appears in other docker compose file.

beiwei30:
> This looks good, thanks for adding the test as well!
> 
> Could you also include it into the conformance test workflow? See https://github.com/dapr/components-contrib/pull/1115/files file .github/workflows/conformance.yml for an example. This will allow the PR automation to test your change as well.

done.

tests/config/state/tests.yml
```diff
@@ -9,3 +9,6 @@ components:
     allOperations: true
   - component: sqlserver
     allOperations: true
+  - component: postgresql
+    allOperations: false
+    operations: [ "set", "get", "delete", "bulkset", "bulkdelete", "transaction", "etag" ]
```
CodeMonkeyLeet, line 14: Nit: add newline at end of file for valid Linux text file

state/postgresql/postgresdbaccess.go
```diff
@@ -274,7 +270,7 @@ func (p *postgresDBAccess) ensureStateTable(stateTableName string) error {
 		p.logger.Info("Creating PostgreSQL state table")
 		createTable := fmt.Sprintf(`CREATE TABLE %s (
 									key text NOT NULL PRIMARY KEY,
-									value json NOT NULL,
+									value text NOT NULL,
```
CodeMonkeyLeet, line 273: Thanks for fixing postgresql to pass the conformance tests ... can you explain why the type for the column needs to be changed to `text` though? It looks like elsewhere in the `setValue` the code explicitly marshals to `json`. What was the problem that required this type change?

beiwei30: It looks like both postgresql and mysql cannot accept the following payload as a json type.


https://github.com/dapr/components-contrib/blob/master/tests/conformance/state/state.go#L70

```go
		{
			key:                  fmt.Sprintf("%s-trx-bytes", key),
			value:                []byte{0x1},
			expectedReadResponse: []byte{0x1},
			transactionOnly:      true,
			transactionGroup:     1,
		},
```

otherwise, the following error occurs:

```
Expected nil, but got: &pgconn.PgError{Severity:"ERROR", Code:"22P02", Message:"invalid input syntax for type json", Detail:"Token "testValue2" is invalid.", Hint:"", Position:0, InternalPosition:0, InternalQuery:"", Where:"JSON data, line 1: testValue2", SchemaName:"", TableName:"", ColumnName:"", DataTypeName:"", ConstraintName:"", File:"jsonfuncs.c", Line:621, Routine:"json_ereport_error"
```

tests/config/state/tests.yml
beiwei30: fixed.

state/postgresql/postgresdbaccess.go
CodeMonkeyLeet: I see, thanks for explaining the problem. Given the expectation that the component expects the raw byte blob to be returned as-is on the Get call, I think change of the field to `text` is reasonable, and there is no expectation in the postgres state component itself that the value column will be manipulated via JSON inspection, so it doesn't seem to matter. 

I have other concerns about the attempt to marshal as JSON if the intent of the conformance tests is that the state stores are binary-preserving:
https://github.com/dapr/components-contrib/blob/418fc726f5501239bcf034ae09a4a29e040ba758/state/postgresql/postgresdbaccess.go#L101
but that should be tracked as part of the broader discussion in https://github.com/dapr/dapr/issues/3533#issuecomment-911789176 and not here.

CodeMonkeyLeet (approved):
LGTM, thanks for all the work on this!

beiwei30:
> I have other concerns about the attempt to marshal as JSON if the intent of the conformance tests is that the state stores are binary-preserving:

Agree, I went through all state implementations, and it looks like json format is expected by state spec:

```go
	var v string
	b, ok := req.Value.([]byte)
	if ok {
		v = string(b)
	} else {
		v, _ = jsoniter.MarshalToString(req.Value)
	}
```

CodeMonkeyLeet:
@beiwei30 The merge of the mysql conformance tests conflicts with this PR, can you take a look?

beiwei30:
> @beiwei30 The merge of the mysql conformance tests conflicts with this PR, can you take a look?

@CodeMonkeyLeet I have resolved the conflict, pls. take a look again.

tests/conformance/common.go
```diff
@@ -53,6 +53,7 @@ import (
 	ss_local_file "github.com/dapr/components-contrib/secretstores/local/file"
 	s_cosmosdb "github.com/dapr/components-contrib/state/azure/cosmosdb"
 	s_mongodb "github.com/dapr/components-contrib/state/mongodb"
+	s_postgresql "github.com/dapr/components-contrib/state/postgresql"
```
CodeMonkeyLeet, line 56: @beiwei30 can you fix the linter issue, it will block a merge? (Run `gofmt` on this) I think this may be because the entry is not in alphabetical order.

beiwei30: done. @CodeMonkeyLeet pls. review again.

CodeMonkeyLeet (approved):
LGTM, thanks for adding the tests!

I've noticed that both the postgres and mysql conformance tests don't run the `first-write` tests today, probably because they fail, but from a correctness perspective, any component that is supporting the `etag` tests should be checking the `first-write` interaction as well. Instead of holding up this PR, I'll file some follow up issues to track adding those to the workflow you've established.

beiwei30:
> LGTM, thanks for adding the tests!
> 
> I've noticed that both the postgres and mysql conformance tests don't run the `first-write` tests today, probably because they fail, but from a correctness perspective, any component that is supporting the `etag` tests should be checking the `first-write` interaction as well. Instead of holding up this PR, I'll file some follow up issues to track adding those to the workflow you've established.

There's one issue (https://github.com/dapr/components-contrib/issues/1123) to address it. Feel free to close it if it's DUP.

state/postgresql/postgresdbaccess.go
artursouza: Instead of converting to JSON, is there a way to handle the special case for ""? There are advantages to keep the structure as JSON since it can allow queries in the near future. BTW, JSONB seems to be the right choice here: http://www.silota.com/docs/recipes/sql-postgres-json-data-types.html

Regarding "", it might be better off to store it as null - so making this column nullable could allow that.

artursouza: Also, byte[] should also be handled in a special way. In cosmosDB, there is a new property (would be a new column here) to specify that the value is binary and should be base64 decoded on read.

artursouza (changes requested):
Please, consider using JSONB to enable querying in the near future.

/cc @dmitsh

berndverst:
To make #1123 work you probably need to update the stored procedure. If you modify the stored procedure, please version it / change the name so everybody upgrading Dapr will receive the correct version. See #1175

beiwei30:
> Instead of converting to JSON, is there a way to handle the special case for ""? 

@artursouza I have updated the pull request to handle "", byte array, and jsonb. It is worth to mention that I simply disallow the empty string ("") in set state operation. Would you mind to review the whole change again?

PR #1131, opened 2021-09-10
Restored backwards compatibility with 1.3

ItalyPaleAle:
#972 accidentally introduced a backwards-incompatible change with a feature added in 1.3. Before, it was possible to specify an Azure environment for the AKV secret store by passing a FQDN as "vaultName" property that included the suffix for the Azure environment.
#972 introduced a better way to handle this (using the "azureEnvironment" metadata property), but accidentally broke the behavior added in 1.3
This patch restores full compatibility with 1.3. Although that behavior should be considered deprecated and thus discouraged (and it will be removed from docs), it will still be supported.

This should be considered a hotfix for the upcoming 1.4

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [ ] N/A Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

### DIFF
--- a/secretstores/azure/keyvault/keyvault.go
+++ b/secretstores/azure/keyvault/keyvault.go
@@ -62,6 +62,7 @@ func (k *keyvaultSecretStore) Init(metadata secretstores.Metadata) error {
 					vaultName = strings.TrimPrefix(vaultName, "https://")
 				}
 				metadata.Properties[componentVaultName] = vaultName
+
 				break
 			}
 		}
--- a/secretstores/azure/keyvault/keyvault.go
+++ b/secretstores/azure/keyvault/keyvault.go
@@ -45,6 +45,29 @@ func NewAzureKeyvaultSecretStore(logger logger.Logger) secretstores.SecretStore 
 
 // Init creates a Azure Key Vault client
 func (k *keyvaultSecretStore) Init(metadata secretstores.Metadata) error {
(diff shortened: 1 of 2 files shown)


secretstores/azure/keyvault/keyvault.go
```diff
@@ -45,6 +45,29 @@ func NewAzureKeyvaultSecretStore(logger logger.Logger) secretstores.SecretStore
 
 // Init creates a Azure Key Vault client
 func (k *keyvaultSecretStore) Init(metadata secretstores.Metadata) error {
+	// Fix for maintaining backwards compatibility with a change introduced in 1.3 that allowed specifying an Azure environment by setting a FQDN for vault name
+	// This should be considered deprecated and users should rely the "azureEnvironment" metadata instead, but it's maintained here for backwards-compatibility
+	if vaultName, ok := metadata.Properties[componentVaultName]; ok {
+		keyVaultSuffixToEnvironment := map[string]string{
+			".vault.azure.net":         "AZUREPUBLICCLOUD",
```
ItalyPaleAle, line 52: Note that the list of suffixes is hardcoded here because it was in 1.3. This will support any (and all) the ones that were supported by 1.3

yaron2 (dismissed):

artursouza (approved):

yaron2:
@ItalyPaleAle there's a small lint issue

yaron2 (approved):

PR #1133, opened 2021-09-10
Cherry-pick: Restored backwards compatibility with 1.3 (#1131)

artursouza:
* Restored backwards compatibility with 1.3
#972 accidentally introduced a backwards-incompatible change with a feature added in 1.3. Before, it was possible to specify an Azure environment for the AKV secret store by passing a FQDN as "vaultName" property that included the suffix for the Azure environment.
#972 introduced a better way to handle this (using the "azureEnvironment" metadata property), but accidentally broke the behavior added in 1.3
This patch restores full compatibility with 1.3. Although that behavior should be considered deprecated and thus discouraged (and it will be removed from docs), it will still be supported.

* Lint

# Description

Cherry pick of #1131 into release

## Issue reference

Please reference the issue this PR will close: #972 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


CodeMonkeyLeet (approved):
LGTM

berndverst:
LGTM

PR #1136, opened 2021-09-12
fix http opa middleware status code parse

Taction:
# Description

Metadata properties is in map[string]string form. But `DefaultStatus` here is an int value, so it can not just be parsed by `json.Unmarshal`.

## Issue reference

Please reference the issue this PR will close: dapr/dapr#3216 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


middleware/http/opa/middleware.go
```diff
@@ -15,7 +16,7 @@ import (
 
 type middlewareMetadata struct {
 	Rego            string `json:"rego"`
-	DefaultStatus   int    `json:"defaultStatus,omitempty"`
+	DefaultStatus   int    `json:"-"`
```
CodeMonkeyLeet, line 19: Instead of splitting the marshalling between annotation and the special handling in `getNativeMetadata` consider keeping this annotation and replacing the special handling in `getNativeMetadata` with a custom `UnmarshalJSON` handler for clarity:

```go

func (mm *middlewareMetadata) UnmarshalJSON(b []byte) error {
	// For brevity, assume we only unmarshal from metadata string map.
	// If middlewareMetadata is itself roundtripped through JSON, need 
	// to handle both string and int mappings to DefaultStatue.
	var v map[string]string
	if err := json.Unmarshal(b, &v); err != nil {
		return err
	}
	mm.Rego = v["rego"]
	mm.IncludedHeaders = v["includedHeaders"]

	if ds, ok := v["defaultStatus"]; ok {
		if intDS, intErr := strconv.Atoi(ds); intErr != nil {
			return intErr
		} else {
			mm.DefaultStatus = intDS
		}
	}
	return nil
}
```

middleware/http/opa/middleware_test.go
```diff
@@ -193,6 +193,17 @@ func TestOpaPolicy(t *testing.T) {
...
+			meta: middleware.Metadata{
+				Properties: map[string]string{
+					"rego": `
+						package http
+						allow = false`,
+					"defaultStatus": "500",
+				},
+			},
+			status: 500,
+		},
```
CodeMonkeyLeet, line 206: Thanks for adding this test in to catch the regression! 

Suggestion: also add a test for the interaction with the `defaultStatus` field as well:

```go
		"rego priority over defaultStatus metadata": {
			meta: middleware.Metadata{
				Properties: map[string]string{
					"rego": `
						package http
						allow = {
							"allow": false,
							"status_code": 301
						}`,
					"defaultStatus": "500",
				},
			},
			status: 301,
		},
```

CodeMonkeyLeet (commented):
Thanks so much for investigating and submitting this fix along with https://github.com/dapr/dapr/pull/3657! See comments inline ...

middleware/http/opa/middleware.go
Taction: Good! Done.

middleware/http/opa/middleware_test.go
Taction: Done.

middleware/http/opa/middleware.go
Taction: I defined an alias type of int for defaultStatus. And define with a custom UnmarshalJSON handler for it.

middleware/http/opa/middleware.go
```diff
@@ -43,6 +46,26 @@ var (
 	errOpaInvalidResultType = errors.New("got an invalid type back from repo policy. Only a boolean or map is valid")
 )
 
+func (s *Status) UnmarshalJSON(b []byte) error {
+	if len(b) == 0 {
+		return nil
+	}
+	if b[0] == '"' {
```
CodeMonkeyLeet, line 53: If you want only unmarshal the `DefaultStatus` as a custom type, I would suggest using the typesafe handling of go instead of attempting to infer the type by walking the array (for example, this code doesn't account for a byte array with leading `"` but not ending `"`).

```go
func (s *Status) UnmarshalJSON(b []byte) error {
	var v interface{}
	if err := json.Unmarshal(b, &v); err != nil {
		return err
	}
	switch t := v.(type) {
	case string:
		if i, err := strconv.Atoi(t); err != nil {
			return err
		} else {
			*s = Status(i)
		}
	case int:
		*s = Status(t)
	default:
		return fmt.Errorf("invalid defaultStatus value: %v", t)
	}
	return nil
}
```

middleware/http/opa/middleware.go
```diff
@@ -43,6 +46,26 @@ var (
 	errOpaInvalidResultType = errors.New("got an invalid type back from repo policy. Only a boolean or map is valid")
 )
 
+func (s *Status) UnmarshalJSON(b []byte) error {
+	if len(b) == 0 {
```
CodeMonkeyLeet, line 50: Given the `omitempty` annotation, I'm okay with rejecting JSON that attempts to define `defaultStatus` without a value ... it simplifies things to just use the default `UnmarshalJSON` to an `interface{}` type to handle edge cases. Does this check handle `"defaultStatus": null` or just `"defaultStatus": []`?

In either case, if you're explicitly handling special cases, I would add the appropriate tests for them in middleware_test.go

Taction: use json unmarshal to `interface{}` to handle edge cases. Also add some tests.

middleware/http/opa/middleware.go
```diff
@@ -43,6 +48,32 @@ var (
...
+		return nil
+	}
+	var v interface{}
+	if err := json.Unmarshal(b, &v); err != nil {
+		return err
+	}
+	switch value := v.(type) {
+	case float64:
```
CodeMonkeyLeet, line 60: I'm guessing that this has to be `float64` because that's what `json.Unmarshal` converts string numbers to and not `int` ... thank you so much for adding the test that catches these cases!

Because this is a float value now though, we should trap illegal inputs here instead of silently truncating, so add a check for that:
``` go
if value != math.Trunc(value) {
	return fmt.Errorf("invalid float value %f parse to status(int)", value)
}
```

middleware/http/opa/middleware.go
```diff
@@ -43,6 +48,32 @@ var (
...
+
+		return nil
+	case string:
+		intVal, err := strconv.Atoi(value)
+		if err != nil {
+			return err
+		}
+		*s = Status(intVal)
+
+		return nil
```
CodeMonkeyLeet, line 71: Optional: you've done a lot of work on this already (thanks!), so this doesn't have to go in this PR, but it occurred to me that since `Status` is now a custom type, we can also enforce that it is in the correct range for RFC 2616 status codes (100-599), so instead of returning immediately from each case statement, add a range check after the switch-case before returning `nil` error.

middleware/http/opa/middleware_test.go
```diff
@@ -232,3 +258,55 @@ func TestOpaPolicy(t *testing.T) {
...
+		jsonBytes   []byte
+		expectValue Status
+		expectError bool
+	}{
+		"int value": {
+			jsonBytes:   []byte(`{"value":2}`),
+			expectValue: Status(2),
+			expectError: false,
+		},
+		"string value": {
```
CodeMonkeyLeet, line 276: Per comment in the code, I'd also add a test case to check that we reject non-integer values
```go
		"invalid float value": {
			jsonBytes:   []byte(`{"value":2.9}`),
			expectError: true,
		},
```

middleware/http/opa/middleware.go
Taction: Done.

middleware/http/opa/middleware_test.go
Taction: Done.

middleware/http/opa/middleware.go
Taction: Done with tests.

CodeMonkeyLeet (approved):
LGTM, and thanks once again for the effort put into this, it is greatly appreciated!

artursouza (approved):

Issue #1014, opened 2021-07-16
For better traceability of Azure Components, consider adding headers that identify the origin of the request as coming from Dapr 


greenie-msft:
## Describe the feature
For better traceability with Azure components, consider adding a header that identifies the origin of the request as coming from an application using Dapr. 

List of Azure Components :
- Azure Blob Storage
- Azure CosmosDB 
- Azure SQL Server 
- Azure Table Storage
- Azure Events Hub 
- Azure Service Bus 
- Azure Key Vault

<br class="Apple-interchange-newline">

## Release Note

RELEASE NOTE:

daixiang0:
Why limit Azure? Do other components have same issues?

artursouza:
@daixiang0 If other cloud providers want to have the same, they should be able to create an issue here with instructions in how to add the agent id to the request.

berndverst:
To my knowledge there is no standard way to identify the client calling the Azure service. I spent some time looking to Azure SQL for example and could not identify any way to log information about the client (such as SDK version or type or any user agent etc).
An additional problem is that we are dealing with a variety of protocols used by the various SDK clients.

A possible solution:
1. Call application insights for an account managed by Microsoft. This does not seem ideal as it feels like unnecessary tracking. It may erode trust in Dapr Azure components.
2. We will need to research each individual product's support for this. Some products (such as SQL Server according to my research) do not have support for this.
3. _Something else..._

berndverst:
#1152 should be in good shape now to address this.

User agents of the form `dapr-$version`, e.g. `dapr-1.3.5`, `dapr-1.4.2-rc.6` and `dapr-edge` are sent for the various services. Depending on the service this is the entire user agent / metadata or just a substring of the user agent.

PR #1152, opened 2021-09-16
Add dapr user agent for Azure Components

berndverst:
# Description

Set custom user agent for `dapr` to be used with various Azure services.

This change prepends `dapr-$version` to the user Agent, or sets this as the user Agent where no other value exists.

## Issue reference

#1014 

## Checklist

* [X] Code compiles correctly

The upstream components already have tests for setting custom user agents. No additional tests needed here.

### DIFF
--- a/pubsub/azure/eventhubs/eventhubs.go
+++ b/pubsub/azure/eventhubs/eventhubs.go
@@ -27,7 +27,6 @@ const (
 	// metadata.
 	connectionString = "connectionString"
 	consumerID       = "consumerID" // passed by dapr runtime
-	userAgent        = "dapr"
 
 	// required by subscriber.
 	storageAccountName   = "storageAccountName"
@@ -158,6 +157,7 @@ func (aeh *AzureEventHubs) Init(metadata pubsub.Metadata) error {
 	if err != nil {
 		return err
 	}
+	userAgent := "dapr-" + logger.DaprVersion
 	aeh.metadata = m
 	hub, err := eventhub.NewHubFromConnectionString(aeh.metadata.connectionString,
 		eventhub.HubWithUserAgent(userAgent))
--- a/pubsub/azure/eventhubs/eventhubs.go
+++ b/pubsub/azure/eventhubs/eventhubs.go
@@ -27,6 +27,7 @@ const (
 	// metadata.
 	connectionString = "connectionString"
 	consumerID       = "consumerID" // passed by dapr runtime
+	userAgent        = "dapr"
 
 	// required by subscriber.
 	storageAccountName   = "storageAccountName"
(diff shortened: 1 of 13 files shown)


berndverst:
FYI @greenie-msft

CodeMonkeyLeet (commented):
Code changes LGTM, not clear if it addresses #1014 as intended though? @greenie-msft, what's the expectation here?

berndverst:
@CodeMonkeyLeet it does address that issue for Service Bus and Event Hubs, but not other services.

We have an email thread with that service team where we discussed this in depth:
 
TL;DR -- Service Bus and Event Hubs teams receive a product, version, user agent and a few other identifiers. An example they provided was the Python SDK.
 
@greenie-msft and I looked at the Python SDK in depth to see just how this is sent. It is transmitted via the AMQP message metadata properties -- this itself is set globally by the sender client constructor -- so the implementation is very specific to Service Bus and also Event Hubs.
 
The GO SDK for these services (just like the Python one) has the ability to prepend to the user agent that is transmitted. This change allows us to add a dapr specific identifier.

Now the much harder problem -- completely out of scope of this PR or anything on our (Dapr) team -- getting the logs from the Microsoft internal Service Bus / Event Hubs distributing tracing log store (Geneva). The team said they likely need to implement something to directly filter and forward events on their end. Something they need for business reasons anyway and this is on their backlog.

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -157,7 +158,8 @@ func (aeh *AzureEventHubs) Init(metadata pubsub.Metadata) error {
 		return err
 	}
 	aeh.metadata = m
-	hub, err := eventhub.NewHubFromConnectionString(aeh.metadata.connectionString)
+	hub, err := eventhub.NewHubFromConnectionString(aeh.metadata.connectionString,
```
artursouza, line 161: Do we want to include version too? Maybe follow the `<product>/<version>` pattern, similar as in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent

artursouza (changes requested):

pubsub/azure/eventhubs/eventhubs.go
berndverst: Yes we do -- but how can I get the version? :D This is in components contrib after all.

berndverst: @artursouza also this isn't a HTTP header, so other user agents don't apply. The way these SDKs are implemented what is sent will be `$ourstring/$theirstring`

artursouza: We don't need to match the HTTP header but it is just as inspiration. The version is important. We already set the version in the dapr build, we can add the version to another public property too, this time in components contrib. Take a look at the Makefile in dapr/dapr repo.

artursouza (approved):

PR #1153, opened 2021-09-16
Add StateStore query API

dmitsh:
# Description

Added StateStore query API and its implementation for MongoDB

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/3662

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


state/azure/cosmosdb/cosmosdb_query.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/dapr/components-contrib/state/query"
+)
+
+type Query struct {
+	query documentdb.Query
+	limit int
+	token string
+}
+
+func (q *Query) VisitEQ(f *query.EQ) (string, error) {
```
artursouza, line 27: In addition to the query text, can this also contain the map of parameters as a response? I think we should support prepared statement natively, otherwise, user code would still be vulnerable to SQL injection if the value is hand crafted for that.

state/mongodb/mongodb_query.go
```diff
@@ -0,0 +1,176 @@
...
+	str := fmt.Sprintf(`{ %q: { "$in": [ %q`, f.Key, f.Vals[0])
+	for _, v := range f.Vals[1:] {
+		str += fmt.Sprintf(", %q", v)
+	}
+	str += " ] } }"
+
+	return str, nil
+}
+
+func (q *Query) visitFilters(op string, filters []query.Filter) (string, error) {
```
artursouza, line 49: Also, instead of returning string, it could return interface{}, so for databases like MongoDb, it would already be the struct that the filter expects.

artursouza (changes requested):

state/mongodb/mongodb_query.go
dmitsh: This function returns string because it generates the query, which is string

state/azure/cosmosdb/cosmosdb_query.go
dmitsh: The function argument `query.EQ` already contains the exact key/value. What is the purpose and content of the suggested map of parameters?

state/azure/cosmosdb/cosmosdb_query.go
```diff
@@ -0,0 +1,192 @@
...
+func (q *Query) execute(client *documentdb.DocumentDB, collection *documentdb.Collection) ([]state.QueryResult, string, error) {
+	opts := []documentdb.CallOption{documentdb.CrossPartition()}
+	if q.limit != 0 {
+		opts = append(opts, documentdb.Limit(q.limit))
+	}
+	if len(q.token) != 0 {
+		opts = append(opts, documentdb.Continuation(q.token))
+	}
+	items := []CosmosItem{}
+	resp, err := client.QueryDocuments(collection.Self, &q.query, &items, opts...)
```
artursouza, line 140: https://docs.microsoft.com/en-us/azure/cosmos-db/sql/sql-query-parameterized-queries#examples

state/mongodb/mongodb_query.go
```diff
@@ -0,0 +1,176 @@
...
+
+type Query struct {
+	query  string
+	filter interface{}
+	opts   *options.FindOptions
+}
+
+func (q *Query) VisitEQ(f *query.EQ) (string, error) {
+	// { <key>: <val> }
+	return fmt.Sprintf("{ %q: %q }", f.Key, f.Val), nil
```
artursouza, line 32: `key` or `value` could have a content that could turn this string into an invalid JSON. Instead, build the query object directly in the visitor methods.

PR #1156, opened 2021-09-20
azure storage queue fix double quotation

fjvela:
# Description

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/bindings/azure/storagequeues/storagequeues.go
+++ b/bindings/azure/storagequeues/storagequeues.go
@@ -13,7 +13,7 @@ import (
 	"net/url"
 	"os"
 	"os/signal"
-	"strings"
+	"strconv"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


fjvela:
hello @CodeMonkeyLeet!

I need some help, I checked the changes on my machine. Now when you create a  new msg  it doesn't create double quotation on it.

I see that the binding is input/output  (https://docs.dapr.io/reference/components-reference/supported-bindings/#microsoft-azure) and also it implement a read method but I don't see how can I execute it. Do you know how can I execute it?

CodeMonkeyLeet:
@fjvela The `Read()` is exercised by daprd much like how daprd handles subscriptions in the pubsub components; you can see an example of how it is consumed by the app in https://docs.dapr.io/reference/api/bindings_api/#invoking-service-code-through-input-bindings.

bindings/azure/storagequeues/storagequeues.go
```diff
@@ -70,13 +70,17 @@ func (d *AzureQueueHelper) Init(accountName string, accountKey string, queueName
 func (d *AzureQueueHelper) Write(data []byte, ttl *time.Duration) error {
 	ctx := context.TODO()
 	messagesURL := d.queueURL.NewMessagesURL()
-	s := string(data)
```
artursouza, line 73: Two questions:
1. Is this a breaking change?
2. Can the handling of byte[] be done via a Content-Type metadata instead?

artursouza (changes requested):

bindings/azure/storagequeues/storagequeues.go
fjvela: 1. It could be a break change if you are removing the double quotation and you don't check if there aren't any quotation to remove. To avoid this, may be we can add a new option in metadata to indicate if the component must remove the double quotation, indicate this break change in the next release and then change the behaviour to remove always the double quotation

2. About Content-Type, I don't know how to do it. Can you give me more information?

yaron2: /cc @artursouza

fjvela: @artursouza can you check my comments?

berndverst:
This change broke support for UTF8 character sets like Chinese. We need to revert this or change it.

Issue #950, opened 2021-06-17
Certify Azure Table Storage as Stable in state store


artursouza:
## Describe the feature
Certify Azure Table Storage as GA in state store

## Release Note

RELEASE NOTE:

tanvigour:
## Test Plan Draft

### Existing tests (Conformance test)
- Test for mixed data types
- Test for init
- Test for ping
- Test for operations  :  get , set delete, bulkdete, transaction, etag, first-write

###  Proposed tests 
-  Test for existing  table name vs new table handling (functional)
-  Test for authentication using Azure auth layer : https://github.com/dapr/components-contrib/issues/1103
-  Test for consistency in behavior of Partition and Row key with non-Dapr consumers (conformance)
-  Test for data persistence : write with dapr and read natively should have consistent data (data validation) , vice versa?

berndverst:
@tanvigour @artursouza important update:

This component should switch to the new / current Track 2 SDK for Table Storage.

This SDK supports BOTH the Azure Table Storage (as part of a Storage Container) as well as the Cosmos DB Table API.

https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/data/aztables

IMO we should not GA this component using the existing outdated SDK

berndverst:
From recent Discord activity is clear that customers expect this component to connect to **both** Azure Blobstorage Tables as well as Cosmos DB Tables API. That can be facilitated by using the new Track 2 SDK and exposing the relevant options as component metadata.

Tracking the Track 2 migration work separately at https://github.com/dapr/components-contrib/issues/1528

tanvigour:
> From recent Discord activity is clear that customers expect this component to connect to **both** Azure Blobstorage Tables as well as Cosmos DB Tables API. That can be facilitated by using the new Track 2 SDK and exposing the relevant options as component metadata.
> 
> Tracking the Track 2 migration work separately at #1528

Do we have an estimate  on migration of table storage state store to track 2 sdk?

tanvigour:
## Test Plan Draft

### Proposed tests (Certification)

- Test handing the  "TableName" for existing vs new table 
   -  Create a table "ExistingTable" and specify that in the metadata tablename field
   -  Confirm "ExistingTable" is being used as table for azure table storage
   -  Don't create a table before hand and specify name "NewTable" in metadata field
   - Confirm new table named "NewTable" is created and used by Azure Table storage

- Test for data persistence
   - Write data using Dapr
   - Read data natively
[...]

PR #1159, opened 2021-09-20
Conformance Tests for Azure Table Storage

berndverst:
# Description

Progress towards #950 : Adds conformance tests for Azure Table Storage Pass.

Fixes incorrect ETag Deletion behavior.

* [X] Code compiles correctly
* [X] Created/updated tests


tests/config/state/tests.yml
```diff
@@ -5,10 +5,11 @@ components:
...
-  - component: cosmosdb
+  - component: azure.cosmosdb
     allOperations: true
   - component: sqlserver
     allOperations: true
   - component: mysql
-    allOperations: false
     operations: [ "set", "get", "delete", "bulkset", "bulkdelete", "transaction", "etag" ]
+  - component: azure.tablestorage
+    operations: ["set", "get", "delete", "etag", "bulkset", "bulkdelete"]
```
CodeMonkeyLeet, line 15: Is `first-write` not supported by azure.tablestorage or is it that just it fails the first-write tests? Unlike `transaction` there's not a separate function implementation necessary for that test. 

(It looks like mysql may have the same issue, where the test was dropped because it could not pass?)

tests/config/state/azure/tablestorage/statestore.yml
```diff
@@ -0,0 +1,14 @@
...
+spec:
+  type: state.azure.tablestorage
+  version: v1
+  metadata:
+  - name: accountName
+    value: ${{AzureBlobStorageAccount}}
+  - name: accountKey
+    value: ${{AzureBlobStorageAccessKey}}
+  - name: tableName
+    value: TestTable
```
CodeMonkeyLeet, line 14: Nit: newline at EOF for linux text files.

state/azure/tablestorage/tablestorage.go
```diff
@@ -227,11 +230,11 @@ func (r *StateStore) deleteRow(req *state.DeleteRequest) error {
...
-	var etag string
 	if req.ETag != nil {
-		etag = *req.ETag
+		entity.OdataEtag = *req.ETag
+
+		return entity.Delete(false, nil)
 	}
-	entity.OdataEtag = etag
 
 	return entity.Delete(true, nil)
```
CodeMonkeyLeet, line 239: For my understanding, is there a reason for forcing the delete? It seems like the bool arg only relates to whether the ETag is ignored for the deletion anyway, so it seems like both returns could be consolidated into just the non-forced delete call. Might be useful to add a comment on its necessity.

CodeMonkeyLeet (commented):
Nice, thanks for adding this and fixing up the component!

state/azure/tablestorage/tablestorage.go
berndverst: We discussed this via chat. I've added code comments that hopefully help explain this. This is also documented in the tablestorage documentation.

tests/config/state/tests.yml
berndverst: Forgot to add this one. I fixed the first-write behavior now.

berndverst:
@CodeMonkeyLeet thanks for the review. Take another look when you got a chance.

state/azure/tablestorage/tablestorage.go
```diff
@@ -195,20 +193,25 @@ func (r *StateStore) writeRow(req *state.SetRequest) error {
...
-	err := entity.Update(false, nil)
+	err := entity.Insert(storage.FullMetadata, nil)
 	if err != nil {
-		if isNotFoundError(err) {
-			// When entity is not found (set state first time) create it
-			entity.OdataEtag = ""
-
-			return entity.Insert(storage.FullMetadata, nil)
+		if etag == "" {
+			return state.NewETagError(state.ETagMismatch, err)
```
CodeMonkeyLeet, line 202: Doesn't this depend on whether the concurrency state is FirstWrite? (i.e. it should be a valid request to upsert with default last write wins semantics and no ETag specified, the function should have called Update instead).

It doesn't look like the component checks for req.Options.Concurrency anywhere in relation to its ETag handling, which seems wrong.

berndverst: Thanks for reminding me of that table. I now covered the combination of etag set / not set and concurrency options.

state/azure/tablestorage/tablestorage.go
```diff
@@ -195,20 +193,29 @@ func (r *StateStore) writeRow(req *state.SetRequest) error {
...
+			if isNotFoundError(err) {
+				return state.NewETagError(state.ETagMismatch, err)
+			}
+		}
+
+		return err
 	}
 
-	return err
+	return nil
```
CodeMonkeyLeet, line 218: The logic LGTM, good job working it out. I would suggest heavily commenting it because the _why_ it works took me a while to unravel that from the structure of the code and the quirks of the conformance test expectations. 

I would also suggest explicitly checking for why Insert fails and handling expected cases with Update only.

```go
err := entity.Insert(storage.FullMetadata, nil)
if err != nil {
	// If Insert failed because item already exists, try to Update instead per Upsert semantics
	if isEntityAlreadyExistsError(err) {
		// Always Update using the etag when provided even if Concurrency != FirstWrite.
		// This behavior is enforced by test state.go:440 today, where we treat the presence of etag as trumping Concurrency.
		// In the future, we should error if an etag is provided when not using FirstWrite (see #2739)
		if etag != "" {
			uerr := entity.Update(false, nil)
			if uerr != nil {
				if isNotFoundError(uerr) {
					return state.NewETagError(state.ETagMismatch, uerr)
				}
				return uerr
			}
		} else if req.Options.Concurrency == state.FirstWrite {
			// Otherwise, if FirstWrite was set, but no etag was provided for an Update operation
			// explicitly flag it as an error. This is enforced by test state.go:541.
			// entity.Update itself does not flag the test case as a mismatch as it does not distinguish
			// between nil and "" etags, the initial etag will always be "", which would match on update.
			return state.NewETagError(state.ETagMismatch, errors.New("update with Concurrency.FirstWrite without ETag"))
		} else {
			// Finally, last write semantics without ETag should always perform a force update.
			return entity.Update(true, nil)
		}
	} else {
		// Any other unexpected error on Insert is propagated to the caller
		return err
	}
}

return nil
```

where 

```go
func isEntityAlreadyExistsError(err error) bool {
	azureError, ok := err.(storage.AzureStorageServiceError)

	return ok && azureError.Code == "EntityAlreadyExists"
}
```

CodeMonkeyLeet (approved):
LGTM!

PR #1009, opened 2021-07-13
Return SystemProperties on events from EventHubs bindings and pubsub

CodeMonkeyLeet:
# Description

- Update azure-event-hubs-go dependency to currently supported v3 with IoT Hub SystemProperties.
- Write `Event.SystemProperties` to `ReadResponse.Metadata` for EventHubs input binding.
- Write `Event.SystemProperties` to `NewMessage.Metadata` for EventHubs pubsub.

## Issue reference

Fixes: #227

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [ ] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: https://github.com/dapr/docs/issues/1639

Issue #227, opened 2020-02-16
Event-Hub input binding should propagate event meta-data


chleit:
Currently the Event-Hub input binding only propagates the data payload of an event.
AFAIK I could see this is done in the following code-snippet:

```
                        handler(&bindings.ReadResponse{
				Data: event.Data,
			})
```

I would change this to propagate the full event including all SystemProperties like "iothub-connection-device-id".

matthieuklotz:
Hi,
We have the same use case: we need to access to metadatas of event produced by IoT Devices. But currently it is not possible.
Is someone have some news about this issue?

Regards,

CodeMonkeyLeet:
The IoTHub SystemProperties only got added to the Azure EventHubs golang SDK in [version 3.2.0](https://github.com/Azure/azure-event-hubs-go/commit/bc2987d6d5deed48b84abd63d3bb4ab4e315c3d5), but the Dapr EventHubs pubsub and binding components currently are on the deprecated v1.3.1, so minimally Dapr needs to:

1. update the version dependency to `github.com/Azure/azure-event-hubs-go/v3`
2. expose the SystemProperties in the bindings.ReadResponse to the caller.

PR #1174, opened 2021-09-29
Support setting up Azure IoT Hub for integration testing with EventHubs components

CodeMonkeyLeet:
# Description

- Add setup of Azure IoT Hub to setup-azure-conf-tests.sh for integration testing with EventHubs bindings/pubsub.
- Backfill functional tests for #1009

## Issue reference

Test follow-up to #227 
Test infrastructure to support #951 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/bindings/azure/eventhubs/eventhubs_integration_test.go
+++ b/bindings/azure/eventhubs/eventhubs_integration_test.go
@@ -1,0 +1,107 @@
+//go:build integration_test
+// +build integration_test
+
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation and Dapr Contributors.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+package eventhubs
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/dapr/components-contrib/bindings"
+	"github.com/dapr/kit/logger"
+)
+
+const (
+	// Note: Reuse the environment variables names from the conformance tests where possible to support reuse of setup workflows
+
+	// iotHubConnectionStringEnvKey defines the key containing the integration test connection string
+	// For the default EventHub endpoint for an Azure IoT Hub, it will resemble:
+	// Endpoint=sb://<iotHubGeneratedNamespace>.servicebus.windows.net/;SharedAccessKeyName=service;SharedAccessKey=<key>;EntityPath=<iotHubGeneratedPath>
+	iotHubConnectionStringEnvKey = "AzureIotHubEventHubConnectionString"
+	iotHubConsumerGroupEnvKey    = "AzureIotHubBindingsConsumerGroup"
+	iotHubNameEnvKey             = "AzureIotHubName"
+	storageAccountNameEnvKey     = "AzureBlobStorageAccount"
+	storageAccountKeyEnvKey      = "AzureBlobStorageAccessKey"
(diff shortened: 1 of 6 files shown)


CodeMonkeyLeet:
FYI @tanvigour. This can be sliced and diced as needed to fit with the new integration framework depending on your testing priorities for the EventHubs work.

bindings/azure/eventhubs/eventhubs_integration_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+func testReadIotHubEvents(t *testing.T) {
+	logger := logger.NewLogger("bindings.azure.eventhubs.integration.test")
+	eh := NewAzureEventHubs(logger)
+	err := eh.Init(createIotHubBindingsMetadata())
+	assert.Nil(t, err)
+
+	// Invoke az CLI via bash script to send test IoT device events
+	// Requires the AZURE_CREDENTIALS environment variable to be already set (output of `az ad sp create-for-rbac`)
+	cmd := exec.Command("/bin/bash", "../../../tests/scripts/send-iot-device-events.sh")
```
artursouza, line 61: Is there a way to make this work based on the repo "root" instead of this folder?

pubsub/azure/eventhubs/eventhubs_integration_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+func testReadIotHubEvents(t *testing.T) {
+	logger := logger.NewLogger("pubsub.azure.eventhubs.integration.test")
+	eh := NewAzureEventHubs(logger)
+	err := eh.Init(createIotHubPubsubMetadata())
+	assert.Nil(t, err)
+
+	// Invoke az CLI via bash script to send test IoT device events
+	// Requires the AZURE_CREDENTIALS environment variable to be already set (output of `az ad sp create-for-rbac`)
+	cmd := exec.Command("/bin/bash", "../../../tests/scripts/send-iot-device-events.sh")
```
artursouza, line 63: same.

artursouza (commented):
LGTM overall. Why mixing this with conformance infra? Is there a way to separate infra that is only applicable to integration test?

bindings/azure/eventhubs/eventhubs_integration_test.go
CodeMonkeyLeet: I'm not keen on relative paths here, but I couldn't think of a good, simple way to reference a file elsewhere in the git tree. Alternatives considered:
- shell out and run `git rev-parse --show-toplevel` and then build the the path from the git root. 
- require the caller provide the script path as an env variable
- copy the script to each test folder.

I didn't think it was worth spending too much time on, particularly if this would be rewritten to a different framework (and likely relocated) at that point.

CodeMonkeyLeet:
> LGTM overall. Why mixing this with conformance infra? Is there a way to separate infra that is only applicable to integration test?

Operationally, I would suggest there be a single system for deploying pre-provisioned cloud resources, and then add knobs to the control plane for it as necessary for how the maintainer workflow for requires it. Right now, the test Azure resources are provisioned once into a single pool, so this PR just combines it into the existing deployment script; as the need arises, the script could parameterized to deploy select resources as necessary to different resource groups, and this may be along groupings of different test types.

Given that there is a new test framework coming up that may drive different requirements for resource deployment as part of the test execution, I would think of this as documenting the resource deployment configuration needed in a usable way in the interim, and see what the requirements are for per-test granularity of resource deployment for cloud infra.

artursouza (approved):

PR #1180, opened 2021-10-01
Cosmos DB Conformance Tests

berndverst:
Makes Cosmos DB Conformance Tests Pass


berndverst:
@daixiang0 @CodeMonkeyLeet made some changes now

By removing the upstream identification hydrator I was able to avoid the struct requirement that forced a particular Cosmos DB entity structure. Now we have complete flexibility again.

Do note that the partition path specified upon table (collection) creation means this field / attribute must exist in every document sent to this output binding.

PR #1182, opened 2021-10-01
Add state.sqlserver conformance test against Azure SQL

CodeMonkeyLeet:
# Description

- Add support in setup-azure-conf-test.sh for deploying Azure SQL server the state.sqlserver conformance/integration testing.
   - Avoid pre-creating the database as Azure SQL DBs charge for hosting that does not scale to zero.
   - Provision the test runner Service Principal as Contributor so it can configure the firewall and create DBs as needed during testing.
- Add conformance test state.azure.sql for exercising state.sqlserver component against Azure SQL instance.
   - Add special handling in conformance.yml to configure Azure SQL firewall rules to permit test running IP to connect, and to clean up firewall rule and test DB afterwards.
   - Existing sqlserver_integration_test.go can reuse the deployed Azure SQL instance by following similar steps to configure the SQL firewall, and mapping `export DAPR_TEST_SQL_CONNSTRING="$AzureSqlServerConnectionString"`, where `AzureSqlServerConnectionString` can be sourced from the setup-azure-conf-test.sh generated conf-test-config.rc file.

## Issue reference

Rounding out conformance testing specifically against Azure SQL for https://github.com/dapr/components-contrib/issues/949

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/tests/config/state/azure/sql/statestore.yaml
+++ b/tests/config/state/azure/sql/statestore.yaml
@@ -1,0 +1,17 @@
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: statestore
+spec:
+  type: state.sqlserver
+  metadata:
+  - name: connectionString
+    value: ${{AzureSqlServerConnectionString}}
+  - name: tableName
+    value: dapr_conf_test
+  - name: keyType
+    value: string
+  - name: keyLength
+    value: 120
+  - name: schema
+    value: proto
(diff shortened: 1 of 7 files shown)


tests/config/state/azure/sql/statestore.yaml
```diff
@@ -0,0 +1,17 @@
...
+kind: Component
+metadata:
+  name: statestore
+spec:
+  type: state.sqlserver
+  metadata:
+  - name: connectionString
+    value: ${{AzureSqlServerConnectionString}}
+  - name: tableName
+    value: dapr_conf_test
```
artursouza, line 11: Should the tableName also be dynamic to avoid run conflicts?

CodeMonkeyLeet: It could be, but in general all the conformance tests are not parallelizable today with duplicate instances of themselves. 
The use of the env var just makes the framework look up the value in AKV for a value that is fixed at resource creation time and non-secret.

artursouza (approved):

Issue #1185, opened 2021-10-05
Misconfigured Azure Service Bus Binding Inconsistent Behavior


halspang:
## Expected Behavior

A misconfigured component should have consistent behavior. Either consistently handling it or throwing an exception to help resolve the issue.


## Actual Behavior

An Azure Service Bus binding with a misconfigured queue name (`testQueue`) will successfully create a queue with a different name (`testqueue`). After the first pass subsequent passes fail as it tries to recreate the existing queue. 

First pass:
```
INFO[0000] Azure Service Bus - Put                       app_id=Slothlinen-Pegasus instance=Hals-MacBook-Pro.local scope=dapr.contrib type=log ver=edge
INFO[0001] successful init for output binding sb-binding (bindings.azure.servicebusqueues/v1)  app_id=Slothlinen-Pegasus instance=Hals-MacBook-Pro.local scope=dapr.runtime type=log ver=edge
INFO[0002] Azure Service Bus - Comparing desired queue testQueue to found queue testqueue  app_id=Slothlinen-Pegasus instance=Hals-MacBook-Pro.local scope=dapr.contrib type=log ver=edge
INFO[0002] Azure Service Bus - Put                       app_id=Slothlinen-Pegasus instance=Hals-MacBook-Pro.local scope=dapr.contrib type=log ver=edge
ERRO[0002] failed to init input binding sb-binding (bindings.azure.servicebusqueues/v1): error code: 409, Details: SubCode=40900. Conflict. You're requesting an operation that isn't allowed in the resource's current state. To know more visit https://aka.ms/sbResourceMgrExceptions. . TrackingId:372ab722-5261-4e5a-ae5e-7b22d24d3024_G61, SystemTracker:halspangsb.servicebus.windows.net:testQueue2, Timestamp:2021-10-05T23:07:33  app_id=Slothlinen-Pegasus instance=Hals-MacBook-Pro.local scope=dapr.runtime type=log ver=edge
[...]

PR #1186, opened 2021-10-06
Force lowercase name for Azure Service Bus Queue

halspang:
# Description

Azure Service Bus Queue names are case-insensitive. When one is
created it will always be forced to lowercase. If a queue with a
non-lowercase is provided, it will be created but can never be
matched causing the next Init to fail.

## Issue reference

Please reference the issue this PR will close: #1185 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/bindings/azure/servicebusqueues/servicebusqueues.go
+++ b/bindings/azure/servicebusqueues/servicebusqueues.go
@@ -8,6 +8,7 @@ package servicebusqueues
 import (
 	"context"
 	"encoding/json"
+	"strings"
 	"time"
 
 	servicebus "github.com/Azure/azure-service-bus-go"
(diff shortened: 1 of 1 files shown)


bindings/azure/servicebusqueues/servicebusqueues.go
```diff
@@ -130,6 +131,8 @@ func (a *AzureServiceBusQueues) parseMetadata(metadata bindings.Metadata) (*serv
 
 	m.ttl = ttl
 
+	m.QueueName = strings.ToLower(m.QueueName)
```
CodeMonkeyLeet, line 134: Consider commenting why this case conversion is happening for future maintainers.

CodeMonkeyLeet (commented):
Change LGTM. Per the issue template, can you also open a docs issue to describe the handling of the `queueName` field in https://docs.dapr.io/reference/components-reference/supported-bindings/servicebusqueues/ as being case-insensitive with a forced conversion to lowercase on queue creation?

bindings/azure/servicebusqueues/servicebusqueues.go
halspang: Good call, added a comment.

Doc issue: https://github.com/dapr/docs/issues/1839

CodeMonkeyLeet: The recent force push doesn't look like it contains code changes?

artursouza (approved):

Issue #1103, opened 2021-08-28
Update all other Azure components to use the common auth layer (and support Azure AD auth)


ItalyPaleAle:
This is a "spin-off" from #972 given how it was merged with only 2 components updated.

Goal of this issue is to track what Azure components still need to be updated to support authentication via the "common auth layer" (implemented in #972). This enables components to authenticate with Azure using Azure AD, including support for service principals and MSI.

While updating components to use the common Azure auth layer, we have been making sure (where appropriate/relevant) to also support for other Azure clouds (Azure China, Azure Germany, Azure Gov, and also Azure Stack) and even emulators when appropriate.

Here's the list of components that use Azure and their status:

- [x] `bindings/azure/blobstorage` - Uses keys
  - [x] Update for track2 SDK: https://github.com/dapr/components-contrib/issues/1530
  - [x] 🛑 Waiting until new version of track2 SDK
- [x] `bindings/azure/cosmosdb`
  - #1104
  - [x] Update for track 2 SDK: https://github.com/dapr/components-contrib/pull/1876
- [x] `bindings/azure/eventgrid` - Currently supports client credentials auth but with its own logic
- [x] `bindings/azure/eventhubs` - Uses keys
  - 🛑 Update for track 2 SDK: Blocked because the SDK is not yet available
- [x] `bindings/azure/servicebusqueues`
  - #1201
  - [x] Update for track 2 SDK: https://github.com/dapr/components-contrib/issues/1531
- [X] `bindings/azure/signalr` - Uses "connection strings"
  - #1843
  - Does not use a SDK
- [x] `bindings/azure/storagequeues` - Uses keys
  - #1842
  - 🛑  Update for track 2 SDK: Blocked because the SDK is not yet available
- [x] `pubsub/azure/eventhubs` - Uses keys
  - #1292
  - 🛑  Update for track 2 SDK: Blocked because the SDK is not yet available
- [x] `pubsub/azure/servicebus`
  - #1201
  - [x] Update for track 2 SDK: https://github.com/dapr/components-contrib/issues/1532
- [X] `secretstores/azure/keyvault`
  - #972
  - [x] Update for track 2 SDK: #1290
- [X] `state/azure/blobstorage`
  - #972
  - [x] Update fo track2 SDK: https://github.com/dapr/components-contrib/issues/1529
- [x] `state/azure/cosmosdb`
  - #1104
  - [x] Update for track 2 SDK: https://github.com/dapr/components-contrib/issues/1526
[...]

PR #1201, opened 2021-10-15
Move Service Bus Pubsub/Binding to common auth

halspang:
# Description

Both the pubsub and input/output binding for Azure Service Bus were
connecting via a connection string. This is still supported but will
now fallback to using AAD from the common auth library. This is also
the recommended auth pattern going forward.

## Issue reference

Partial: #1103 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: https://github.com/dapr/docs/issues/1867


berndverst:
FYI, @halspang and I successfully verified this works with managed identity now

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: azure-servicebus
spec:
  type: pubsub.azure.servicebus
  version: v1
  metadata:
  - name: namespaceName
    value: "dapr2-conf-test-servicebus"
  - name: consumerID
    value: "helloworld"
 ```

We ran the conformance test in an environment with managed identity where the identity was given "Azure Service Bus Data Owner" role for the particular Service Bus Namespace

File: bindings/azure/servicebusqueues/servicebusqueues.go [hunk 1]
```diff
@@ -61,10 +64,36 @@ func (a *AzureServiceBusQueues) Init(metadata bindings.Metadata) error {
 	userAgent := "dapr-" + logger.DaprVersion
 	a.metadata = meta
 
-	ns, err := servicebus.NewNamespace(servicebus.NamespaceWithConnectionString(a.metadata.ConnectionString),
-		servicebus.NamespaceWithUserAgent(userAgent))
-	if err != nil {
-		return err
+	var ns *servicebus.Namespace
+	if a.metadata.ConnectionString != "" {
```
CodeMonkeyLeet, line 68: Did you mean `if a.metadata.ConnectionString == ""` here? You want to use the Azure auth if not explicitly provided the ConnectionString, right?

Also, it might be clearer to just say `a.metadata.NamespaceName != ""` since `parseMetadata` now enforces that the ConnectionString should not be set if NamespaceName is also set.

CodeMonkeyLeet (approved):
LGTM, thanks for the fixes!

authentication/azure/auth.go
```diff
@@ -41,6 +41,8 @@ func NewEnvironmentSettings(resourceName string, values map[string]string) (Envi
 	case "cosmosdb":
 		// Azure Cosmos DB (data plane)
 		es.Resource = "https://" + azureEnv.CosmosDBDNSSuffix
+	case "serviceBus":
```
ItalyPaleAle, line 44: Nit: this should probably be all lowercased for consistency with the other identifiers

ItalyPaleAle:
FYI there's a new "track 2" SDK for Azure Service Bus: https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/messaging/azservicebus

Eventually the component will need to be updated to that (and pending merge of #1290 for auth support)

authentication/azure/auth.go
artursouza: done

artursouza (approved):

PR #1203, opened 2021-10-18
WIP: Migrate mDNS to use Sync.Map

jjcollinge:
# Description
The current implementation of the mDNS name resolver uses a native Go `map` and a `mutex` to protect access. This PR migrates the name resolver to use the `Sync.Map` struct instead - which should be more efficient for this read heavy cache.
This PR also attempts to clean up the code a bit to make it more accessible to others and remove duplication.

- Debug log on cache miss rather than hit to reduce noise
- Try to remove abbreviated var names
- Align log format with lowercase
- Remove duplication across IPv4 and IPV6 functions by consolidating
- Delete app entries from cache when no addresses left

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


jjcollinge:
I'm going to close this as it's a premature optimization and probably isn't warranted.

jjcollinge:
There were probably some valid changes in the PR other than using Sync.Map so I'll likely open a new PR with those.

PR #1206, opened 2021-10-19
Integration test setup for AKV

berndverst:
In order to test the various authentication methods for KeyVault several credentials and component configurations need to be configured.

This PR makes the necessary changes.


berndverst:
@CodeMonkeyLeet you previously suggested creating different configs for the different authentication mechanisms. Thoughts on this PR?

I am planning to call the new authentication methods as conformance tests via the new integration test framework (I will need to spin up the managed identity enabled environment from there), though I could also enable serviceprincipal auth directly via the test matrix here.

berndverst:
Removed the Managed Identity configuration from the conformance tests since it can't be run there. Will instead consider creating a sample / quickstart.

Issue #1084, opened 2021-08-16
SASL OAUTHBEARER support in Kafka binding


sthussey:
## Describe the feature
In enterprise settings, authentication - particularly federated authentication - is a requirement for shared platforms like a message queue or event stream such as Kafka. Dapr is a great tool-set for easing access to EDA, but is lacking in some enterprise-grade features like better support of federated authentication. I'm looking for interesting in adding support for federated authentication to Kafka specifically via an open-standard like OAuth2/OIDC via the SASL OAUTHBEARER protocol.

Currently the only path I see to using federated identity between Dapr and Kafka is either a custom listener in Kafka (which rules out using something like Strimzi easily) or using commercial Kafka that has LDAP for using with the SASL username/password. Since the underlying Sarama client supports OAUTHBEARER, the only addition would be generating a valid token in Dapr to pass along and correctly respond to token expiration by generating a new token. 

I could see two methods for this:
 - Make Dapr a OAuth2 client directly 
 - Allow the application to host a Dapr callback that would generate the token. This has the effect (benefit?) of allowing universal binding configuration, but allowing each client service to bring-their-own-identity.

## Release Note

RELEASE NOTE: **ADD** Support OAuth2 authentication with Kafka for pub/sub binding

CodeMonkeyLeet:
@yaron2, any thoughts on this? Dapr has concessions for OAuth handling for service invocation via middleware, but doesn't extend to bindings. At the same time, there have been asks for more direct access to sarama supported auth scenarios (e.g. https://github.com/dapr/components-contrib/issues/349 for SSL, not OAuth) which may point to this being easier to manage on a per component basis.

sthussey:
Just wanted to come back to this - I'd be happy to put together a PR with something in place if there is interest in merging it.

yaron2:
@sthussey it'd be great if you can add this to the existing components. Re: general purpose, lets open a separate issue for that :)

sthussey:
I'm working on the PR for this. First draft of code is finished, just need to take time to setup an integration environment to test it.

https://github.com/sthussey/components-contrib/tree/feat/kafka-oidc-support

sthussey:
PR #1229 is still in progress.

yaron2:
Reopened

PR #1229, opened 2021-10-24
Add OAuth2/OIDC support to Kafka pubsub

sthussey:
# Description

This change leverages Sarama's support for bearer tokens to support OAuth2/OIDC authentication
to access a Kafka broker. It utilizes the client_credentials flow as the most appropriate for machine-to-machine
connections.

Additionally, I found the sarama logs were useful in debugging so I've added a bridge from the sarama logger to
the dapr debug logger. This can be dropped since it isn't part of OIDC support.

## Issue reference

Please reference the issue this PR will close: #1084

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#1882


sthussey:
I've updated the E2E test to include both mTLS and OAuth2 authentication. I think the linters should also pass now, but will make any additional updates required after the gate runs are approved.

pubsub/kafka/kafka.go
```diff
@@ -147,17 +156,30 @@ func (k *Kafka) Init(metadata pubsub.Metadata) error {
...
-		k.saslUsername = meta.SaslUsername
-		k.saslPassword = meta.SaslPassword
-		updateAuthInfo(config, k.saslUsername, k.saslPassword)
-	}
 	err = updateTLSConfig(config, meta)
 	if err != nil {
 		return err
 	}
 
+	switch k.authType {
```
daixiang0, line 164: Does it mean force auth?

pubsub/kafka/kafka.go
```diff
@@ -344,65 +393,96 @@ func (k *Kafka) getKafkaMetadata(metadata pubsub.Metadata) (*kafkaMetadata, erro
...
-		return nil, errors.New("kafka error: invalid value for 'authRequired' attribute")
+		return nil, errors.New("kafka error: 'authType' attribute was empty")
 	}
-	meta.AuthRequired = validAuthRequired
 
-	// ignore SASL properties if authRequired is false
-	if meta.AuthRequired {
-		if val, ok := metadata.Properties["saslUsername"]; ok && val != "" {
+	switch strings.ToLower(val) {
+	case "password":
```
daixiang0, line 405: Better to make authType as const.

pubsub/kafka/sarama_log_bridge.go
```diff
@@ -0,0 +1,21 @@
+package kafka
```
daixiang0, line 1: Why need to package logger here?

sthussey: This allows someone to run Dapr in debug mode and get the Sarama logs for troubleshooting. It isn't required for the core work in this PR, but I needed the functionality to troubleshoot the development and thought it would be useful to leave it for end-users.

pubsub/kafka/kafka.go
```diff
@@ -308,6 +335,33 @@ func (k *Kafka) Subscribe(req pubsub.SubscribeRequest, handler pubsub.Handler) e
 	return nil
 }
 
+// upgradeMetadata updates metadata properties based on deprecated usage.
+func (k *Kafka) upgradeMetadata(metadata pubsub.Metadata) (pubsub.Metadata, error) {
+	// If authType is not set, derive it from authRequired.
+	if val, ok := metadata.Properties["authType"]; !ok || val == "" {
```
daixiang0, line 341: How about get those value first then check to reduce `if` block?

sthussey: I've refactored this to reduce the nested levels.

daixiang0:
Please `git rebase` master branch.

sthussey:
I ran the full certification suite locally and golangci-lint, so I think the changes are ready for the CI workflows. Would like to try and get this in for 1.6 so we can use it downstream.

@daixiang0

daixiang0:
@sthussey LGTM overall, let's wait the test result.

sthussey:
I looked at the certification CI failure. It is likely caused by the Kafka brokers losing their data on container stop/start. I refactored them to use volumes rather than writing to the container root. The CI agent hosts must be configured differently than my local docker.

tests/certification/pubsub/kafka/kafka_test.go
```diff
@@ -60,10 +62,14 @@ const (
...
 
-var brokers = []string{"localhost:19092", "localhost:29092", "localhost:39092"}
+var (
+	brokers          = []string{"localhost:19092", "localhost:29092", "localhost:39092"}
+	oauthClientQuery = "https://localhost:4444/clients/dapr"
+)
 
 func TestKafka(t *testing.T) {
 	log := logger.NewLogger("dapr.components")
+	// log.SetOutputLevel(logger.DebugLevel)
```
daixiang0, line 72: Remove this line.

tests/certification/pubsub/kafka/strimzi-kafka1-config/server.config
```diff
@@ -0,0 +1,104 @@
+##############################
```
daixiang0, line 1: Could we only keep custom config for those?

sthussey:
I'll respond to the comments above. I'm a bit perplexed about the CI failure for the certification test since the same commit passes running against my fork. I'll see if I can find something that would cause it to fail here.

https://github.com/sthussey/components-contrib/runs/4520833263?check_suite_focus=true

yaron2:
> I'll respond to the comments above. I'm a bit perplexed about the CI failure for the certification test since the same commit passes running against my fork. I'll see if I can find something that would cause it to fail here.
> 
> https://github.com/sthussey/components-contrib/runs/4520833263?check_suite_focus=true

This might be a flaky test. I re-ran the checks.

yaron2:
@pkedy can you do a quick pass here? you're the de-facto Dapr Kafka expert :)

tests/certification/pubsub/kafka/kafka_test.go
sthussey: Done

tests/certification/pubsub/kafka/strimzi-kafka1-config/server.config
sthussey: Most of this is needed unless I tried to build templating into flow and dynamically create each broker config from a single template. Is the added complexity worth it? These files should mostly be static.

sthussey:
I adjusted a bit of the Kafka config back to defaults. I ran it a few times on my fork and CI passed each time. Could we try a re-run here to see if CI is passing now?

daixiang0 (approved):
LGTM

yaron2:
Thanks for the great contribution @sthussey.

sthussey:
Thanks for all the support getting this done. Dapr is a great project we look forward to utilizing downstream. I'll work on getting the docs PR finished.

PR #1248, opened 2021-10-29
feature/pulsar: add producer option "disableBatching"

1046102779:
The pulsar MQ strategy is **at least once**, it does not guarantee that the message is not repeated, and the DisableBatching in the producer Option is false by default. The impact of this default strategy is:
When the pendingBatching in a pulsar client is greater than 1, the batch is submitted to the server for processing, for example: 10 messages, one of which fails to process, and the other 9 messages are processed successfully, then the pulsar server will return an error message that the server reports an error , So that the entire connection will be actively closed, and then resend all the messages on this pendingBatching around 100ms repeatly,  then 9 of them will be resent and will be consumed by consumers, and the wrong message will never be lost. As a result, the messages on the pulsar server queue are piled up in a short time, and the repeated messages are repeatedly processed by the consumer. Therefore, the disableBatching parameter must be added, and it is recommended that DisableBatching=true is turned on by default in the Options parameter of CreateProducer.

wecom-temp-7d2e72f72c62507e8e7c945b545afaba

after 1 hour:

wecom-temp-d7cc7c57b98e8c15f8f2017e48e4b303


daixiang0:
Please update related docs to add this option.

CodeMonkeyLeet:
A couple of observations:

- In general, when introducing new functionality, the preferred approach is to make the new functionality opt-in (in this case `disableBatching` remains false by default) to minimize surprise for developers upgrading. We could say that should be true, but I don't think the case for doing so is obvious based on your description of the problem (in general, defaults optimize for the common case, which should not be the expectation that batches are failing all the time and require resending)

- `disableBatching` is a fairly all-or-nothing approach to the problem, and pulsar offers parameters for tuning the batching behavior, including `batchingMaxPublishDelay` and `batchingMaxMessages` to control the batching granularity and amortize the size of potential resends on batch failure. My impression is that the Dapr component should not be opinionated on how the pulsar component is configured (i.e. disabling batching for all users, which is the opposite of a normal pulsar config) and could address some of the concerns expressed here by passing through the parameters for managing this behavior to the dev. 

@pkedy @artursouza that said, based on a previous discussion of batching in pubsub. we had identified that component-controlled batching on pub independent of Dapr runtime was problematic because of the possibility of catastrophic loss (i.e. call to dapr pub API returns success to client, but the underlying component has not yet sent message in a batch, so a dapr sidecar crash loses all messages). It might be the case that this is not even an option exposed to the end-user, and is always internally set to false for pubsub.pulsar.

1046102779:
> A couple of observations:
> 
> * In general, when introducing new functionality, the preferred approach is to make the new functionality opt-in (in this case `disableBatching` remains false by default) to minimize surprise for developers upgrading. We could say that should be true, but I don't think the case for doing so is obvious based on your description of the problem (in general, defaults optimize for the common case, which should not be the expectation that batches are failing all the time and require resending)
> * `disableBatching` is a fairly all-or-nothing approach to the problem, and pulsar offers parameters for tuning the batching behavior, including `batchingMaxPublishDelay` and `batchingMaxMessages` to control the batching granularity and amortize the size of potential resends on batch failure. My impression is that the Dapr component should not be opinionated on how the pulsar component is configured (i.e. disabling batching for all users, which is the opposite of a normal pulsar config) and could address some of the concerns expressed here by passing through the parameters for managing this behavior to the dev.
> 
> @pkedy @artursouza that said, based on a previous discussion of batching in pubsub. we had identified that component-controlled batching on pub independent of Dapr runtime was problematic because of the possibility of catastrophic loss (i.e. call to dapr pub API returns success to client, but the underlying component has not yet sent message in a batch, so a dapr sidecar crash loses all messages). It might be the case that this is not even an option exposed to the end-user, and is always internally set to false for pubsub.pulsar.

message queue has different service semantics, such as:
1. At most once ;
2. At least once
3. Exactly once

For pubsub, when we switch message queues, we cannot guarantee that the semantics are consistent. For pulsar, producer option `dispatchBatching` defaultly is closed, at least pulsar server will not produce duplicate messages that have been correctly processed

CodeMonkeyLeet:
> message queue has different service semantics, such as:
> 
> 1. At most once ;
> 2. At least once
> 3. Exactly once
> 
> For pubsub, when we switch message queues, we cannot guarantee that the semantics are consistent. For pulsar, producer option `dispatchBatching` defaultly is closed, at least pulsar server will not produce duplicate messages that have been correctly processed

I think I see the issue you're pointing out; I was referring to the fact that the existing implementation already [enables batching by default](https://github.com/apache/pulsar-client-go/blob/171ef578e91a661c35523dccca35618b8c560b93/pulsar/producer.go#L123) so changing that default should be documented as a break with the existing behavior. For the reason you state, and issues with component controlled batching I mentioned, I agree that it's probably something we want to disable; my question for @artursouza is whether we should even expose the option for users to enable this batching.

pubsub/pulsar/pulsar.go
```diff
@@ -54,6 +55,15 @@ func parsePulsarMetadata(meta pubsub.Metadata) (*pulsarMetadata, error) {
 		}
 		m.EnableTLS = tls
 	}
+	// DisableBatching is defaultly disable batching.
+	m.DisableBatching = true
+	if val, ok := meta.Properties[disableBatching]; ok {
+		enableBatching, err := strconv.ParseBool(val)
```
mukundansundar, line 61: please rename variable to `disableBatch` or something similar to `disableBatching`. The complete opposite of that does create some confusion.

daixiang0:
@1046102779 please fix conflicts.

Issue #1253, opened 2021-10-30
Body of email empty in SMTP binding when invoking the binding over gRPC


wmeints:
## Expected Behavior

The data sent to the SMTP binding is correctly translated to e-mail content.

## Actual Behavior

The body of the e-mail is empty when you invoke the SMTP binding over gRPC.

## Steps to Reproduce the Problem

I noticed this behavior when using the SMTP output binding in Python. When executing the following code:

```python
with DaprClient() as client:
    client.invoke_binding(
        binding_name="sendmail",
        operation="create",
        data=message_body,
        binding_metadata=message_properties
    )
```

The client sends the metadata to the correct binding with the correct operation. However, when I check the generated email, the body of the email is empty. After some more debugging I found out that when I send a POST request to the same binding it works. I've checked this with the following code:

```python
message_data = {
    "data": message_body,
    "operation": "create",
    "metadata": {
          "subject": "Fine for exceeding the speed limit.",
          "emailTo": vehicle.ownerEmail,
          "emailFrom": "test@domain.org"
    }
 }

 response = requests.post("http://localhost:3601/v1.0/bindings/sendmail", json=message_data)
```

This leads me to believe that the DaprClient isn't sending the data to the binding.

## Release Note

RELEASE NOTE: 

**FIX** The body of the e-mail sent by the SMTP binding is no longer empty when you call it over gRPC

PR #1254, opened 2021-10-30
Fix body content parsing for SMTP binding

wmeints:
# Description

This fixes a problem with the body of e-mail messages sent by the
SMTP binding.

When the SMTP binding was invoked over GRPC it would not translate the
data of the request to the right format in the e-mail. Instead it would
send an empty body. When the SMTP binding was invoked over HTTP it would
correctly unquote the data of the request.

Now when you call the SMTP binding, it will try to unquote the body content to handle the data coming in over HTTP. When the data can't be unquoted, we use the incoming data as-is as it comes over GRPC and is already in a usable format.

## Issue reference

Fixes https://github.com/dapr/components-contrib/issues/1253

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


yaron2 (approved):
lgtm

Issue #1255, opened 2021-10-30
Warning generated by SMTP binding when username and password are both set.


wmeints:
## Expected Behavior

The SMTP binding generates a warning when no username and password are provided.

## Actual Behavior

The SMTP binding generates a warning when username AND password are set.

## Steps to Reproduce the Problem

1. Create a binding configuration for SMTP:

```
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: sendmail
spec:
  type: bindings.smtp
  version: v1
  metadata:
  - name: host
    value: localhost
  - name: port
    value: 4025
  - name: user
    secretKeyRef:
      name: smtp.user
      key: smtp.user
  - name: password
    secretKeyRef:
      name: smtp.password
      key: smtp.password
  - name: skipTLSVerify
    value: true
auth:
  secretStore: trafficcontrol-secrets
```

2. Start an app with `dapr run`.

## Release Note

RELEASE NOTE:

CodeMonkeyLeet:
Fixed by #1256

PR #1256, opened 2021-10-30
Fix warning in smtp binding when user and password are set

wmeints:
# Description

Changed the behavior of the SMTP binding so it explicitly checks if both user and password are unset before logging a warning about the missing credentials. Right now, the binding raises a warning when they're set, which is incorrect.

## Issue reference

Fixes issue https://github.com/dapr/components-contrib/issues/1255

## Checklist

* [x] Code compiles correctly


Taction (approved):
LGTM.

yaron2 (dismissed):

bindings/smtp/smtp.go
```diff
@@ -138,8 +138,10 @@ func (s *Mailer) parseMetadata(meta bindings.Metadata) (Metadata, error) {
 	//nolint
```
daixiang0, line 138: remove this please.

daixiang0 (approved):

artursouza (approved):

Issue #1282, opened 2021-11-06
MQTT bindings conformance tests have a ~1% failure rate on Create test


CodeMonkeyLeet:
## Expected Behavior

MQTT bindings conformance tests should pass reliably.

## Actual Behavior

MQTT bindings conformance tests have a ~1% failure rate on Create test with the following failure signature:

```
=== RUN   TestBindingsConformance/mqtt-emqx/create
time="2021-11-03T23:35:17.345828789Z" level=info msg="Create test running ..." instance=fv-az206-587 scope=bindingsTest type=log ver=unknown
time="2021-11-03T23:35:17.347180905Z" level=debug msg="mqtt publishing topic testBindingTopic with data: [91 123 34 101 118 101 110 116 84 121 112 101 34 58 34 116 101 115 116 34 44 34 101 118 101 110 116 84 105 109 101 34 58 32 34 50 48 49 56 45 48 49 45 50 53 84 50 50 58 49 50 58 49 57 46 52 53 53 54 56 49 49 90 34 44 34 115 117 98 106 101 99 116 34 58 34 100 97 112 114 45 99 111 110 102 45 116 101 115 116 115 34 44 34 105 100 34 58 34 65 50 51 52 45 49 50 51 52 45 49 50 51 52 34 44 34 100 97 116 97 34 58 34 114 111 111 116 47 62 34 125 93]" instance=fv-az206-587 scope=testLogger type=log ver=unknown
    bindings.go:191: 
        	Error Trace:	bindings.go:191
        	Error:      	Received unexpected error:
        	            	mqtt error from publish: write tcp [::1]:32810->[::1]:1884: use of closed network connection
        	Test:       	TestBindingsConformance/mqtt-emqx/create
        	Messages:   	expected no error invoking output binding
time="2021-11-03T23:35:17.348703823Z" level=info msg="Create test done." instance=fv-az206-587 scope=bindingsTest type=log ver=unknown
        --- FAIL: TestBindingsConformance/mqtt-emqx/create (0.00s)
```

Tracing the MQTT code, this seems internal to the paho.mqtt.golang lib, and may be related to https://github.com/eclipse/paho.mqtt.golang/issues/410. Dapr uses the default `AutoReconnect` value (`true`) for the MQTT client configuration and from the traces it looks like MQTT is constantly reconnecting in the background. 

In the failure cases, it looks like there is a race condition between the disconnect and when `publish` is called to queue a new outbound message. The most minimal example I've seen is:

<pre>
[...]

PR #1289, opened 2021-11-10
Add internal retry to MQTT bindings Invoke()

CodeMonkeyLeet:
# Description

There is an internal race condition in MQTT client Publish method when configured with autoreconnect (default) where the call can fail if the connection is broken right after the outgoing event is queued but before the dead connection is blocked.

This change mitigates that on the Dapr side with a limited retry loop 3 times at fixed 200ms interval as a patch before the Dapr resiliency feature is introduced that exposes more generalized retry controls to the dev.

## Issue reference

Patches #1282

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* ~Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]~


bindings/mqtt/mqtt.go
```diff
@@ -186,14 +186,27 @@ func (m *MQTT) Operations() []bindings.OperationKind {
...
 func (m *MQTT) Invoke(req *bindings.InvokeRequest) (*bindings.InvokeResponse, error) {
-	m.logger.Debugf("mqtt publishing topic %s with data: %v", m.metadata.topic, req.Data)
-
-	token := m.producer.Publish(m.metadata.topic, m.metadata.qos, m.metadata.retain, req.Data)
-	if !token.WaitTimeout(defaultWait) || token.Error() != nil {
-		return nil, fmt.Errorf("mqtt error from publish: %v", token.Error())
-	}
-
-	return nil, nil
+	// MQTT client Publish() has an internal race condition in the default autoreconnect config.
```
daixiang0, line 189: Need to comment delete time when resiliency feature introduced.

CodeMonkeyLeet: To clarify, is the recommendation to reword to:
> This is not configurable to keep this as an implementation detail for this component, as the additional public config metadata required **will** be replaced by the more general Dapr APIs for resiliency moving forwards.

mukundansundar (commented):
@CodeMonkeyLeet  should this change also be made in MQTT pubsub for now ?

CodeMonkeyLeet:
> @CodeMonkeyLeet should this change also be made in MQTT pubsub for now ?

Unclear. As noted in the bug, this has never shown up as a pubsub test failure; even when running repros with tracing enabled in my private fork, I've not come across a trace where a disconnect right after queuing and outbound message happens, so it may be that the temporal dynamics of the pubsub component makes it much less likely to hit the condition.

The patch fix is safe enough that we could apply it to pubsub as well, but there are two tradeoffs:
- it adds a 600ms latency to Invoke for unrecoverable publish failures
- it would probably have to be taken out of the pubsub component as well once the resiliency feature is introduced, which puts the responsibility on the dev for configuring retries and such.

bindings/mqtt/mqtt.go
berndverst: I'm fine with keeping this as is and for this components to be updated when resiliency features are out.

berndverst (approved):

artursouza (approved):

PR #1132, opened 2021-09-10
Bumping ASB version and adapt to latest go-autorest/adal library (MSI_ENDPOINT)

pkedy:
# Description

* Updates to `go.mod` for the Azure Service Bus SDK
* EventGrid API (Future -> FutureAPI)
* Setting `MSI_ENDPOINT` environment variable to unit test

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #1290, opened 2021-11-11
Azure SDK "track 2": authentication and secretstore/azure/keyvault

ItalyPaleAle:
# Description

This PR is yet another step into addressing #1103.

The Azure SDK team is rebuilding all services SDKs essentially from scratch, in a more streamlined way. Sadly, that also comes with a complete rewrite of the auth logic, and the previous auth logic implemented for the "common Azure auth layer" is not compatible with the "track 2" Azure SDKs.

This PR updates the secretstore/azure/keyvault component to use the new Azure Key Vault secret store SDK (https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/keyvault/azsecrets) that is part of "track 2".

Most importantly, however, this PR introduces support for the new Azure Identity client module (https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/azidentity) which all the updated SDKs rely on.

While the internals of the Azure Key Vault secret store have changed, the code should be fully backwards-compatible for Dapr users.

PS: I have tested this against a live Key Vault resource in my Azure subscription, but I do not have a way to test it against Azure China or other clouds. If anyone could help with that, it would be great!

## Issue reference

#1103 (doesn't close the issue, but it's part of that larger story)

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Extended the documentation: N/A


authentication/azure/auth.go
```diff
@@ -69,10 +72,58 @@ func (s EnvironmentSettings) GetAzureEnvironment() (*azure.Environment, error) {
...
+	}
+	return azidentity.NewChainedTokenCredential(creds, nil)
+}
+
 // GetAuthorizer creates an Authorizer retrieved from, in order:
 // 1. Client credentials
 // 2. Client certificate
-// 3. MSI.
+// 3. MSI
+// This is used by the older Azure SDKs
```
berndverst, line 126:
```suggestion
// This is used by the older Azure SDKs.
```

berndverst:
```suggestion
// This is used by the older Azure SDKs.
```

authentication/azure/auth.go
```diff
@@ -69,10 +72,58 @@ func (s EnvironmentSettings) GetAzureEnvironment() (*azure.Environment, error) {
 	return &env, err
 }
 
+// GetTokenCredential returns an azcore.TokenCredential retrieved from, in order:
+// 1. Client credentials
+// 2. Client certificate
+// 3. MSI
+// This is used by the newer ("track 2") Azure SDKs
```
berndverst, line 79:
```suggestion
// This is used by the newer ("track 2") Azure SDKs.
```

authentication/azure/auth.go
```diff
@@ -85,7 +136,8 @@ func (s EnvironmentSettings) GetAuthorizer() (autorest.Authorizer, error) {
 // GetServicePrincipalToken returns a Service Principal Token retrieved from, in order:
 // 1. Client credentials
 // 2. Client certificate
-// 3. MSI.
+// 3. MSI
+// This is used by the older Azure SDKs
```
berndverst, line 140:
```suggestion
// This is used by the older Azure SDKs.
```

berndverst:
Thanks for the PR @ItalyPaleAle. I will review!

ItalyPaleAle:
Two notes after a code review:

1. In track 2 SDKs, the APIs to control the telemetry data have changed. While this is specific to each component, it's not possible (or at least, not straightforward) to set the full User Agent anymore. Instead, the APIs expose an "ApplicationID" field that can be used to add a string of up to 24 characters. Currently I've set that to `""dapr-" + logger.DaprVersion` which is the value of the old User-Agent. However the SDKs now append their own string to that, so the User-Agent that is sent will be something like `dapr-1.4.0 azsecrets/v1.0.0` which includes the version of the SDK. If this is undesirable I can make the change to match the old behavior.
2. In the old version, there were no timeouts and the component was using `context.Background()` everywhere. This is probably not desirable: a blocked network request would now hang the goroutine forever. The better option would be to use a context with a timeout. Speaking with @berndverst, he recommended not making this change now, so I've used `context.TODO()` instead for now

berndverst:
Thanks @ItalyPaleAle. Dapr will have to revisit the proper handling and setting of context throughout components as part of resiliency feature plans.

berndverst (approved):
Thanks for this PR enabling the Track 2 auth / identity libraries in addition to the existing ones.

I verified these additions (and existing tests) work. I updated my KeyVault certification tests which validate all auth mechanisms to run against this PR. The tests pass.

berndverst:
I also manually verified that managed identity works in production with these changes!

berndverst:
@greenie-msft @paulyuk you will enjoy this PR as this adds support for the Track 2 Identity libraries / the AAD auth mechanisms required for all Track 2 SDKs.

berndverst:
cc @artursouza this is ready for your review

authentication/azure/auth_test.go
```diff
@@ -146,6 +146,8 @@ func TestGetMSI(t *testing.T) {
 }
 
 func TestFallbackToMSI(t *testing.T) {
+	os.Setenv("MSI_ENDPOINT", "test")
```
artursouza, line 149: Please, do not depend on Env variables in unit tests because it is global state that can make the tests flaky if ran in parallel.

artursouza (changes requested):

artursouza:
LGTM overall, just one concern about using env variables in unit testing.

authentication/azure/auth_test.go
berndverst: This is a partial revert of the revert of https://github.com/dapr/components-contrib/pull/1132

This was the easiest way @pkedy was able to get the MSI tests to pass after upgrading the auth libraries. This change was in master until we reverted it (for unrelated reasons) last week.

berndverst: Also keep in mind `MSI_ENDPOINT` or `IDENTITY_ENDPOINT` are special purposes environment variables that are set in Azure Managed Identity environments. See for example https://docs.microsoft.com/en-us/azure/app-service/overview-managed-identity

MSI itself in fact always relies on these special environment variables. As such this use here is fine!

berndverst: The alternative is to remove unit tests for MSI entirely which isn't great.

artursouza: We need to find a solution where tests that depend on the presence or absence of some environment variables are behind a mutex to make sure they don't run in parallel. I will approve this PR but this should be fixed afterwards since it adds flakiness to unit tests.

artursouza (approved):

PR #1292, opened 2021-11-11
rewrite EventHubs component

### LINKED ISSUE #964: Azure EventHubs pubsub component doesn't handle topics correctly

### LINKED ISSUE #1225: Support for partition key while publishing to Azure Event Hubs
## Describe the feature
Able to set partition key while publishing messages to Azure Event Hubs.
With this capability of assigning partition key while publishing - the user can maintain the order of the events under required scenarios

Similar to : https://github.com/dapr/components-contrib/issues/667

mukundansundar:
# Description

Rewrite of eventhubs component.
1. Added support for partition keys.
2. Added support for honoring incoming topic names in requests. 
3. Fixed bug on running multiple subscribers for a topic(leaser checkpointer prefix issue).
4. Added support for authentication via AAD.
5. Added support for creating eventHubs/consumer groups on the fly as long as a the event hubs namespace is created and with right permissions. 

Adding more resiliency and retries if needed on connection will be part of the certification task and is left off for this PR.

## Issue reference

part of #951 
closes #964 
closes dapr/dapr#3753
closes #1225 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: https://github.com/dapr/docs/issues/1941

### DIFF
--- a/pubsub/azure/eventhubs/eventhubs.go
+++ b/pubsub/azure/eventhubs/eventhubs.go
@@ -7,38 +7,48 @@ package eventhubs
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
+	"github.com/Azure/azure-amqp-common-go/v3/aad"
+	"github.com/Azure/azure-amqp-common-go/v3/conn"
 	eventhub "github.com/Azure/azure-event-hubs-go/v3"
 	"github.com/Azure/azure-event-hubs-go/v3/eph"
 	"github.com/Azure/azure-event-hubs-go/v3/storage"
+	mgmt "github.com/Azure/azure-sdk-for-go/services/eventhub/mgmt/2017-04-01/eventhub"
 	"github.com/Azure/azure-storage-blob-go/azblob"
 	"github.com/Azure/go-autorest/autorest/azure"
 
+	azauth "github.com/dapr/components-contrib/authentication/azure"
 	"github.com/dapr/components-contrib/pubsub"
 	"github.com/dapr/kit/logger"
 	"github.com/dapr/kit/retry"
 )
 
 const (
-	// metadata.
-	connectionString = "connectionString"
-	consumerID       = "consumerID" // passed by dapr runtime
 
-	// required by subscriber.
-	storageAccountName   = "storageAccountName"
-	storageAccountKey    = "storageAccountKey"
-	storageContainerName = "storageContainerName"
+	// connection string entity path key.
+	entityPathKey = "EntityPath"
+	// metadata partitionKey key.
+	partitionKeyMetadataKey = "partitionKey"
 
 	// errors.
-	missingConnectionStringErrorMsg     = "error: connectionString is a required attribute"
-	missingStorageAccountNameErrorMsg   = "error: storageAccountName is a required attribute"
-	missingStorageAccountKeyErrorMsg    = "error: storageAccountKey is a required attribute"
-	missingStorageContainerNameErrorMsg = "error: storageContainerName is a required attribute"
-	missingConsumerIDErrorMsg           = "error: missing consumerID attribute"
+	hubManagerCreationErrorMsg               = "error: creating eventHub manager client"
+	invalidConnectionStringErrorMsg          = "error: connectionString is invalid"
+	missingConnectionStringNamespaceErrorMsg = "error: connectionString or eventHubNamespace is required"
+	missingStorageAccountNameErrorMsg        = "error: storageAccountName is a required attribute for subscribe"
+	missingStorageAccountKeyErrorMsg         = "error: storageAccountKey is a required attribute for subscribe"
+	missingStorageContainerNameErrorMsg      = "error: storageContainerName is a required attribute for subscribe"
+	missingConsumerIDErrorMsg                = "error: missing consumerID attribute for subscribe"
+	bothConnectionStringNamespaceErrorMsg    = "error: both connectionString and eventHubNamespace are given, only one should be given"
+	missingResourceGroupNameMsg              = "error: missing resourceGroupName attribute required for entityManagement"
+	missingSubscriptionIDMsg                 = "error: missing subscriptionID attribute required for entityManagement"
+	entityManagementConnectionStrMsg         = "error: entity management support is not available with connectionString"
+	differentTopicConnectionStringErrorTmpl  = "error: specified topic %s does not match the event hub name in the provided connectionString"
 
 	// Event Hubs SystemProperties names for metadata passthrough.
 	sysPropSequenceNumber             = "x-opt-sequence-number"
@@ -52,6 +62,18 @@ const (
 	sysPropIotHubConnectionModuleID   = "iothub-connection-module-id"
 	sysPropIotHubEnqueuedTime         = "iothub-enqueuedtime"
 	sysPropMessageID                  = "message-id"
+
+	defaultMessageRetentionInDays = 1
+	defaultPartitionCount         = 1
+
+	resourceCheckMaxRetry                       = 5
+	resourceCheckMaxRetryInterval time.Duration = 5 * time.Minute
+	resourceCreationTimeout       time.Duration = 15 * time.Second
+	resourceGetTimeout            time.Duration = 5 * time.Second
+
+	// See https://docs.microsoft.com/en-us/azure/event-hubs/event-hubs-quotas for numbers.
+	maxMessageRetention = int32(90)
+	maxPartitionCount   = int32(1024)
 )
 
 func subscribeHandler(ctx context.Context, topic string, e *eventhub.Event, handler pubsub.Handler) error {
@@ -98,21 +120,32 @@ func subscribeHandler(ctx context.Context, topic string, e *eventhub.Event, hand
 
 // AzureEventHubs allows sending/receiving Azure Event Hubs events.
 type AzureEventHubs struct {
-	hub      *eventhub.Hub
-	metadata azureEventHubsMetadata
-
-	logger        logger.Logger
-	ctx           context.Context
-	cancel        context.CancelFunc
-	backOffConfig retry.Config
+	metadata           *azureEventHubsMetadata
+	logger             logger.Logger
+	ctx                context.Context
+	cancel             context.CancelFunc
+	backOffConfig      retry.Config
+	hubClients         map[string]*eventhub.Hub
+	eventProcessors    map[string]*eph.EventProcessorHost
+	hubManager         *eventhub.HubManager
+	eventHubSettings   azauth.EnvironmentSettings
+	managementSettings azauth.EnvironmentSettings
+	cgClient           *mgmt.ConsumerGroupsClient
+	tokenProvider      *aad.TokenProvider
 }
 
 type azureEventHubsMetadata struct {
-	connectionString     string
-	consumerGroup        string
-	storageAccountName   string
-	storageAccountKey    string
-	storageContainerName string
+	ConnectionString        string `json:"connectionString,omitempty"`
+	EventHubNamespace       string `json:"eventHubNamespace,omitempty"`
+	ConsumerGroup           string `json:"consumerID"`
+	StorageAccountName      string `json:"storageAccountName,omitempty"`
+	StorageAccountKey       string `json:"storageAccountKey,omitempty"`
+	StorageContainerName    string `json:"storageContainerName,omitempty"`
+	EnableEnitityManagement bool   `json:"enableEntityManagement,omitempty,string"`
+	MessageRetentionInDays  int32  `json:"messageRetentionInDays,omitempty,string"`
+	PartitionCount          int32  `json:"partitionCount,omitempty,string"`
+	SubscriptionID          string `json:"subscriptionID,omitempty"`
+	ResourceGroupName       string `json:"resourceGroupName,omitempty"`
 }
 
 // NewAzureEventHubs returns a new Azure Event hubs instance.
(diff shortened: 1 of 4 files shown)


authentication/azure/auth_amqp.go
```diff
@@ -0,0 +1,21 @@
...
+
+package azure
+
+import "github.com/Azure/azure-amqp-common-go/v3/aad"
+
+// GetTokenProvider creates a TokenProvider for AAD retrieved from, in order:
+// 1. Client credentials
+// 2. Client certificate
+// 3. MSI.
+func (s EnvironmentSettings) GetAADTokenProvider() (*aad.TokenProvider, error) {
```
mukundansundar, line 14: Same as the one in https://github.com/dapr/components-contrib/pull/1201

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -98,62 +117,304 @@ func subscribeHandler(ctx context.Context, topic string, e *eventhub.Event, hand
...
+	processor, err := eph.New(aeh.ctx, aeh.metadata.EventHubNamespace, topic, aeh.tokenProvider, leaserCheckpointer, leaserCheckpointer, eph.WithNoBanner(), eph.WithConsumerGroup(aeh.metadata.ConsumerGroup))
+	if err != nil {
+		return nil, err
+	}
+	aeh.logger.Debugf("processor initialized via AAD for topic %s", topic)
+
+	return processor, nil
+}
+
+func (aeh *AzureEventHubs) createHubManager() error {
```
mukundansundar, line 352: HubManager client from azure-event-hubs go sdk for creating/managing an event hub.

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -98,62 +117,304 @@ func subscribeHandler(ctx context.Context, topic string, e *eventhub.Event, hand
...
+		return err
 	}
+	_, err = aeh.getConsumerGroupsClient()
+	if err != nil {
+		return err
+	}
+	return aeh.createConsumerGroup(hubName)
+}
+
+func (aeh *AzureEventHubs) getConsumerGroupsClient() (*mgmt.ConsumerGroupsClient, error) {
```
mukundansundar, line 216: consumer group client from the event hubs management clients in azure-go-sdk for managing consumer groups.

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -7,38 +7,48 @@ package eventhubs
...
 	"errors"
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
+	"github.com/Azure/azure-amqp-common-go/v3/aad"
+	"github.com/Azure/azure-amqp-common-go/v3/conn"
 	eventhub "github.com/Azure/azure-event-hubs-go/v3"
 	"github.com/Azure/azure-event-hubs-go/v3/eph"
```
mukundansundar, line 20: use of event processor host will also affect batch API ... as right now there does not seem to be any option for batching within EPH.

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -196,30 +514,42 @@ func (aeh *AzureEventHubs) Publish(req *pubsub.PublishRequest) error {
...
+		}
+	}
+	cred, err := azblob.NewSharedKeyCredential(aeh.metadata.StorageAccountName, aeh.metadata.StorageAccountKey)
 	if err != nil {
 		return err
 	}
 
-	leaserCheckpointer, err := storage.NewStorageLeaserCheckpointer(cred, aeh.metadata.storageAccountName, aeh.metadata.storageContainerName, azure.PublicCloud)
+	// Set topic name, consumerID prefix for partition checkpoint lease blob path.
+	leaserPrefixOpt := storage.WithPrefixInBlobPath(aeh.getStoragePrefixString(req.Topic))
```
mukundansundar, line 532: This is needed to support multiple consumers with the same storage container option.

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -162,15 +423,62 @@ func (aeh *AzureEventHubs) Init(metadata pubsub.Metadata) error {
...
+		aeh.eventHubSettings = settings
+		tokenProvider, err := aeh.eventHubSettings.GetAADTokenProvider()
+		if err != nil {
+			return fmt.Errorf("%s %s", hubManagerCreationErrorMsg, err)
+		}
+		aeh.tokenProvider = tokenProvider
+		aeh.logger.Info("connecting to Azure EventHubs via AAD. connection established on first publish/subscribe")
+		aeh.logger.Debugf("req.Topic field in incoming requests honored")
+
+		if aeh.metadata.EnableEnitiyManagement {
```
mukundansundar, line 463: A couple of more points here: 
if entity Management is enabled, it does need permissions at least at the EventHub namespace level to be able to create event hubs and consumer groups. 
Also the other metadata, SubscriptionID and ResourceGroupName is a mandatory requirement when enabling entityManagement as for now, it assumes that the topic will be subscribed to. 

If needed resourceGroupName and SubscriptionID can be validated only as part of subscription metadata validation since it will only be required for a subscriber and not for a publisher alone.

mukundansundar: wrt comment https://github.com/dapr/components-contrib/pull/1055#issuecomment-892832206 from @CodeMonkeyLeet , entity management for EventHubs through Dapr does step into provisioning infrastructure resources.

CodeMonkeyLeet (commented):
I'm going to timebox my review on this for now because I've spent a couple of hours looking at this and I'm not done yet. I think this needs a design review first as I'm spending a chunk of time trying to infer if particular behaviors are intended. 

In particular, there are a lot of contingent and interacting configuration options as indicated in this [comment](https://github.com/dapr/components-contrib/pull/1292#discussion_r748098512 ) and I think it would be a good starting point to enumerate the expected valid configuration combinations and the resulting behaviors.

In general, it seems like we could scope down the scenarios we support, for example, supporting connection string auth to EventHubs namespace but without enableEntityManagement where the component supports multiple event hubs that must be predeployed already seems like a something we could just choose not to support, hence design review.

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -98,62 +117,304 @@ func subscribeHandler(ctx context.Context, topic string, e *eventhub.Event, hand
...
-	storageAccountKey    string
-	storageContainerName string
+	ConnectionString       string `json:"connectionString,omitempty"`
+	EventHubNamespace      string `json:"eventHubNamespace,omitempty"`
+	ConsumerGroup          string `json:"consumerID"`
+	StorageAccountName     string `json:"storageAccountName,omitempty"`
+	StorageAccountKey      string `json:"storageAccountKey,omitempty"`
+	StorageContainerName   string `json:"storageContainerName,omitempty"`
+	EnableEnitiyManagement bool   `json:"enableEntityManagement,omitempty,string"`
+	MessageRetentionInDays int32  `json:"messageRetentionInDays,omitempty,string"`
```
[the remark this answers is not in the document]
mukundansundar: Yes without the annotation as String, it fails when unmarshaling the JSON serialised config.

mukundansundar:
> I'm going to timebox my review on this for now because I've spent a couple of hours looking at this and I'm not done yet. I think this needs a design review first as I'm spending a chunk of time trying to infer if particular behaviors are intended.
> 
> In particular, there are a lot of contingent and interacting configuration options as indicated in this [comment](https://github.com/dapr/components-contrib/pull/1292#discussion_r748098512) and I think it would be a good starting point to enumerate the expected valid configuration combinations and the resulting behaviors.
> 
> In general, it seems like we could scope down the scenarios we support, for example, supporting connection string auth to EventHubs namespace but without enableEntityManagement where the component supports multiple event hubs that must be predeployed already seems like a something we could just choose not to support, hence design review.

@CodeMonkeyLeet 
I will shortly update the design for this in the rewrite Issue https://github.com/dapr/components-contrib/issues/951. 
Based on the decision there, I can modify this PR.

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -52,6 +62,15 @@ const (
 	sysPropIotHubConnectionModuleID   = "iothub-connection-module-id"
 	sysPropIotHubEnqueuedTime         = "iothub-enqueuedtime"
 	sysPropMessageID                  = "message-id"
+
+	defaultMessageRetentionInDays = 1
+	defaultPartitionCount         = 1
+
+	timeoutInSec time.Duration = 5 * time.Second //nolint
```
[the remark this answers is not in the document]
mukundansundar: This is for the name `timeoutInSec`, that is throwing a linter error.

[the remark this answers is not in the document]
artursouza: Please, use a new name and consider a bigger default as resource provisioning can be slow.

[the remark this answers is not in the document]
artursouza: If needed, split into multiple timeout configs for different scenarios.

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -98,62 +117,301 @@ func subscribeHandler(ctx context.Context, topic string, e *eventhub.Event, hand
...
+	defer cancel()
+	g, err := aeh.cgClient.Get(ctx, aeh.metadata.ResourceGroupName, aeh.metadata.EventHubNamespace, hubName, aeh.metadata.ConsumerGroup)
+	if err != nil {
+		if g.HasHTTPStatus(404) {
+			return true, nil
+		}
+		return false, err
+	}
+	if *g.Name == aeh.metadata.ConsumerGroup {
+		aeh.logger.Infof("consumer group %s esists for the requested topic/eventHub %s", aeh.metadata.ConsumerGroup, hubName)
```
artursouza, line 255: typo: exists

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -52,6 +62,15 @@ const (
 	sysPropIotHubConnectionModuleID   = "iothub-connection-module-id"
 	sysPropIotHubEnqueuedTime         = "iothub-enqueuedtime"
 	sysPropMessageID                  = "message-id"
+
+	defaultMessageRetentionInDays = 1
```
artursouza, line 66: Can the default be infinity?

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -52,6 +62,15 @@ const (
...
+	defaultPartitionCount         = 1
+
+	timeoutInSec time.Duration = 5 * time.Second //nolint
+
+	// See https://docs.microsoft.com/en-us/azure/event-hubs/event-hubs-quotas for numbers.
+	maxMessageRetention = int32(90)
```
artursouza, line 72: Is this max coming from event hubs or is it arbitrary?

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -98,62 +117,301 @@ func subscribeHandler(ctx context.Context, topic string, e *eventhub.Event, hand
...
 
 type azureEventHubsMetadata struct {
-	connectionString     string
-	consumerGroup        string
-	storageAccountName   string
-	storageAccountKey    string
-	storageContainerName string
+	ConnectionString        string `json:"connectionString,omitempty"`
+	EventHubNamespace       string `json:"eventHubNamespace,omitempty"`
+	ConsumerGroup           string `json:"consumerID"`
```
artursouza, line 137: Why JSON is consumerID and not ConsumerGroup? I think these should match, not?

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -98,62 +117,301 @@ func subscribeHandler(ctx context.Context, topic string, e *eventhub.Event, hand
...
+	if err != nil {
+		return nil, err
+	}
+	client.Authorizer = a
+	aeh.cgClient = &client
+	return aeh.cgClient, nil
+}
+
+func (aeh *AzureEventHubs) createConsumerGroup(hubName string) error {
+	create, err := aeh.shouldCreateConsumerGroup(hubName)
```
artursouza, line 229: Please, add a retry with exponential backoff in all cases where there can be a race condition between multiple sidecar instances. Have a Max retry of 5 or something. This is helpful because users prefer this over relying only on K8s restarts.

artursouza (changes requested):
Overall is good. Just some minor comments.

pubsub/azure/eventhubs/eventhubs.go
mukundansundar: no it cannot be. See https://docs.microsoft.com/en-us/azure/event-hubs/event-hubs-quotas for numbers.

pubsub/azure/eventhubs/eventhubs.go
mukundansundar: see comment above line.

pubsub/azure/eventhubs/eventhubs.go
mukundansundar: I think consumerID is a field coming from dapr ... It is across all pubsub ... not only eventhub ... And in every pubsub it is mapped to some form of a field called cosumer group or such ...

pubsub/azure/eventhubs/eventhubs.go
mukundansundar: resolved all three.

pubsub/azure/eventhubs/eventhubs.go
mukundansundar: done

pubsub/azure/eventhubs/eventhubs.go
artursouza: The metadata should make sense within the component itself. Someone that uses EventHubs expects ConsumerGroup as the terminology. There is no requirement that metadata should match cross components.

artursouza (changes requested):
LGTM. I just have a comment to use ConsumerGroup for consistency within EventHubs. That wins over having the same name cross components for metadata.

pubsub/azure/eventhubs/eventhubs.go
mukundansundar: Currently the metadata properties are marshalled into json and then un-marshalled into the Go struct here. `consumerID` is the incoming property name from `daprd` Here `consumerID` is automatically set by `daprd` and not an independent property of this component. 
It is called consumerGroup within this component.

artursouza: OK. I found the code that handles this logic in daprd. That is the logic that automatically sets the ConsumerID attributed based on the app ID. So, this attribute is the same for all components. LGTM.

artursouza (approved):

Issue #1250, opened 2021-10-29
GCP pub/sub subscription stops working after error


yordan-pavlov:
## Expected Behavior

GCP pub/sub subscription should be resilient to temporary service / network issues.

## Actual Behavior

When the Subscription::Receive method returns an error, the subscription is not reconnected automatically and only starts working again after the dapr process is restarted; see https://github.com/dapr/components-contrib/blob/master/pubsub/gcp/pubsub/pubsub.go#L238

An example error can be:

Subscription closed with error: 
rpc error: code = Unknown 
desc = closing transport due to: 
connection error: 
desc = \"error reading from server: EOF\", 
received prior goaway: 
code: NO_ERROR, 
debug data: server_shutting_down" 

## Steps to Reproduce the Problem

(1) Start dapr sidecar and application and subscribe to GCP pub/sub
(2) After some amount of time, when the subscription is closed due to an error, new messages will not be received any more
(3) Restart dapr to continue receiving messages

NOTE: since errors that have caused the subscription to close are not currently logged, there is no way to know when the issue has occurred; I was only able to get to the error above after adding the following code just before `return err` in the `handleSubscriptionMessages` method.
```
if err != nil {
	g.logger.Warnf("Subscription closed with error: %s", err)
}
```

I imagine that a reconnect loop could be implemented in the `handleSubscriptionMessages` method to fix this issue and make the dapr GCP pub/sub component more resilient, similar to the dapr Azure service bus component here: https://github.com/dapr/components-contrib/blob/master/pubsub/azure/servicebus/servicebus.go#L342

## Release Note

RELEASE NOTE:

yaron2:
Thanks for reporting this. Would you be interested in submitting a PR to fix this?

yordan-pavlov:
@yaron2 I would be happy to submit a PR for this.
Also, it appears that this is an issue with the GO GCP pub/sub library which has been reported here https://github.com/googleapis/google-cloud-go/issues/4257 and fixed in version 1.12.2.

In that case, I wonder if it would be enough to just upgrade to v 1.12.2 of the GCP pub/sub library? Or do you think it would still be useful to implement a reconnect loop in the dapr GCP pub/sub component as well (in addition to upgrading the GCP pub/sub library)?

I will be testing the fix at work over the next few days and will report here by the end of this week.

yaron2:
Perhaps upgrading the SDK version is enough. It would be great if you can report back and open a PR for this when possible.

[...]

PR #1285, opened 2021-11-09
Fix #1250 by upgrading cloud.google.com/go/pubsub to v1.12.2

yordan-pavlov:
# Description

Upgrade cloud.google.com/go/pubsub to v1.12.2 which includes a fix for https://github.com/googleapis/google-cloud-go/issues/4257 where the `Subscription::Receive` method exits with error but should instead automatically reconnect.

Prior to upgrading the GCP pub/sub library, the dapr sidecar would keep running (without logging any errors or warnings) but the subscription connection would stop working and no further messages would be received until the dapr sidecar is restarted. An example error that could trigger this behavior is:

rpc error: code = Unknown
desc = closing transport due to:
connection error:
desc = "error reading from server: EOF",
received prior goaway:
code: NO_ERROR,
debug data: server_shutting_down

After upgrading cloud.google.com/go/pubsub to v1.12.2, the issue hasn't been observed any more after several days of testing.

## Issue reference

Please reference the issue this PR will close: #1250

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [ ] not relevant to this change - Created/updated tests
* [ ] not relevant to this change - Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #1293, opened 2021-11-11
Add reconnect loop for GCP pub/sub subscriptions

yordan-pavlov:
# Description

Add a reconnection loop for GCP pub/sub subscriptions, very similar to the one that already exists for Azure Service Bus subscriptions here https://github.com/dapr/components-contrib/blob/master/pubsub/azure/servicebus/servicebus.go#L384

This is related to #1250 where the connection to a GCP pub/sub subscription could be lost due to a recoverable error. Even if the GCP pub/sub library does handle some recoverable errors internally, a reconnect loop can still be useful in a number of cases. For example:
* the GCP pub/sub library might not handle all recoverable errors (that's why https://github.com/dapr/components-contrib/pull/1285 exists to upgrade the GCP pub/sub library)
* a reconnect loop enables a scenario where the application has to wait for the subscription to be created
* a reconnect loop would also report errors such as a "not found" error thus making the investigation of issues with a GCP pub/sub subscription easier

## Issue reference

Please reference the issue this PR will close: #1250

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#1970


pubsub/gcp/pubsub/pubsub.go
```diff
@@ -16,7 +17,9 @@ import (
 )
 
 const (
-	errorMessagePrefix                 = "gcp pubsub error:"
+	errorMessagePrefix = "gcp pubsub error:"
+
+	// metadata keys
```
daixiang0, line 22: Please fix lint issues.

yordan-pavlov: @daixiang0 thank you for reviewing, I have just published a couple of commits to fix the lint issues and also add tests for the two new metadata properties.

pubsub/gcp/pubsub/metadata.go
```diff
@@ -16,4 +16,6 @@ type metadata struct {
 	ClientCertURL           string
 	DisableEntityManagement bool
 	EnableMessageOrdering   bool
+	MaxReconnectionAttempts int `json:"maxReconnectionAttempts"`
+	ConnectionRecoveryInSec int `json:"connectionRecoveryInSec"`
```
Taction, line 20: Are these json tags useful?

pubsub/gcp/pubsub/pubsub.go
```diff
@@ -235,20 +266,76 @@ func (g *GCPPubSub) Subscribe(req pubsub.SubscribeRequest, handler pubsub.Handle
...
+				}
+				g.logger.Debugf("Number of reconnect attempts remaining for subscription %s: %d", sub.ID(), attempts)
+			}
+		}
+	}()
+
+	// Reconnect loop.
+	var receiveErr error = nil
+	for {
+		receiveErr = sub.Receive(context.Background(), func(ctx context.Context, m *gcppubsub.Message) {
```
Taction, line 305: Please use `g.ctx` instead of `context.Background()`, so this will be canceled when `Close()` is called.

pubsub/gcp/pubsub/metadata.go
yordan-pavlov: good question - I put them in because I noticed they are defined in the Azure Service Bus implementation, but happy to remove if not useful

pubsub/gcp/pubsub/pubsub.go
yordan-pavlov: I was thinking about this, but left like that because the Azure Service Bus implementation uses `context.Background()` here https://github.com/dapr/components-contrib/blob/master/pubsub/azure/servicebus/servicebus.go#L411
but it doesn't mean it's correct - it might be better to use `g.ctx` as you suggest

pubsub/gcp/pubsub/metadata.go
Taction: There is no usage, you can remove them.

pubsub/gcp/pubsub/pubsub.go
Taction: I have taken a quick look at the code you mentioned. It creates a `ctx` based `context.Background()` and cancels it every time in the for loop. The key point is to make sure there is no goroutine leak in the for loop and when `Close()` is called. Both ways seem ok to me.

pubsub/gcp/pubsub/metadata.go
yordan-pavlov: done - removed json tags

pubsub/gcp/pubsub/pubsub.go
yordan-pavlov: done - used `g.ctx` instead of `context.Background()`

pubsub/gcp/pubsub/pubsub.go
```diff
@@ -235,20 +266,76 @@ func (g *GCPPubSub) Subscribe(req pubsub.SubscribeRequest, handler pubsub.Handle
...
+				}
+				g.logger.Debugf("Number of reconnect attempts remaining for subscription %s: %d", sub.ID(), attempts)
+			}
+		}
+	}()
+
+	// Reconnect loop.
+	var receiveErr error = nil
+	for {
+		receiveErr = sub.Receive(g.ctx, func(ctx context.Context, m *gcppubsub.Message) {
```
CodeMonkeyLeet, line 305: For my understanding, is the Subscription object reusable on err from `Receive` (i.e. should the Subscription object be recreated for the retry?)

pubsub/gcp/pubsub/pubsub.go
```diff
@@ -235,20 +266,76 @@ func (g *GCPPubSub) Subscribe(req pubsub.SubscribeRequest, handler pubsub.Handle
...
+		g.logger.Warnf("Connection to subscription %s closed with error: %s", sub.ID(), receiveErr)
+		reconnectionAttemptsRemaining := readReconnectAttemptsRemaining()
+		if reconnectionAttemptsRemaining == 0 {
+			g.logger.Errorf("Reconnection attempts exhausted for subscription %s, giving up after %d attempts.", sub.ID(), g.metadata.MaxReconnectionAttempts)
+			break
 		}
-	})
 
-	return err
+		g.logger.Infof("Sleping for %d seconds before attempting to reconnect to subscription %s ... [%d/%d]", g.metadata.ConnectionRecoveryInSec, sub.ID(), g.metadata.MaxReconnectionAttempts-reconnectionAttemptsRemaining, g.metadata.MaxReconnectionAttempts)
```
CodeMonkeyLeet, line 333: Typo:
```suggestion
               g.logger.Infof("Sleeping for %d seconds before attempting to reconnect to subscription %s ... [%d/%d]", g.metadata.ConnectionRecoveryInSec, sub.ID(), g.metadata.MaxReconnectionAttempts-reconnectionAttemptsRemaining, g.metadata.MaxReconnectionAttempts)
```

CodeMonkeyLeet (commented):
Generally LGTM; per the PR template, can you make sure that there's an issue filed in https://github.com/dapr/docs for the new config metadata properties? It helps to make sure that the added features are appropriately documented with each release.

pubsub/gcp/pubsub/pubsub.go
yordan-pavlov: good question - yes, I have tested that messages are still received after the subscription is reconnected (with the code as is, where it's reusing the Subscription object instead of creating a new one)

pubsub/gcp/pubsub/pubsub.go
yordan-pavlov: good spot, thank you

pubsub/gcp/pubsub/pubsub.go
yordan-pavlov: I have also added an issue to update the docs and linked in the description of this PR above

CodeMonkeyLeet (dismissed):
Thanks, LGTM!

CodeMonkeyLeet:
Spoke too soon: Linter is complaining the code needs a `gofmt`

yordan-pavlov:
yes, it must the automated commit for the log message typo, I will fix it tomorrow morning when I am back at work

CodeMonkeyLeet (approved):

pubsub/gcp/pubsub/pubsub.go
```diff
@@ -235,20 +266,76 @@ func (g *GCPPubSub) Subscribe(req pubsub.SubscribeRequest, handler pubsub.Handle
...
+	// Periodically refill the reconnect attempts channel to avoid
+	// exhausting all the refill attempts due to intermittent issues
+	// occurring over a longer period of time.
+	reconnCtx, reconnCancel := context.WithCancel(g.ctx)
+	defer reconnCancel()
+
+	go func() {
+		// Calculate refill interval relative to MaxReconnectionAttempts and ConnectionRecoveryInSec
+		// in order to allow reconnection attempts to be exhausted in case of a permanent issue (e.g. wrong subscription name).
+		refillInterval := time.Duration(2*g.metadata.MaxReconnectionAttempts*g.metadata.ConnectionRecoveryInSec) * time.Second
```
artursouza, line 286: Is this an implementation of a token bucket? If so, can we reuse an existing library that does that?

artursouza (changes requested):
@pkedy Can this use the retry logic from the kit library?

pkedy:
> @pkedy Can this use the retry logic from the kit library?

Yes. I think the probably makes sense here. No loop needed. I’ll paste some links to examples in other components. Usage is pretty simple and straightforward.

pubsub/gcp/pubsub/pubsub.go
yordan-pavlov: it might make sense to reuse existing dapr retry tools if they match the use case, but does that mean that the very similar retry loop implementation in the Azure Service Bus pub/sub component here https://github.com/dapr/components-contrib/blob/master/pubsub/azure/servicebus/servicebus.go#L384 also has to change?

pubsub/gcp/pubsub/pubsub.go
```diff
@@ -235,20 +266,76 @@ func (g *GCPPubSub) Subscribe(req pubsub.SubscribeRequest, handler pubsub.Handle
...
+				if attempts < g.metadata.MaxReconnectionAttempts {
+					reconnAttempts <- struct{}{}
+				}
+				g.logger.Debugf("Number of reconnect attempts remaining for subscription %s: %d", sub.ID(), attempts)
+			}
+		}
+	}()
+
+	// Reconnect loop.
+	var receiveErr error = nil
```
pkedy, line 303: It looks like reconnecting is a matter of calling `handler` again? It seems like you can replace this w/ the dapr/kit/retry package.

Initialization (perhaps consider a prefix like `reconnect` for this scenario)
https://github.com/dapr/components-contrib/blob/2006bcccb01e397e8987aa03fde249b4d5368b0c/pubsub/azure/eventhubs/eventhubs.go#L176-L182

Retry operation
https://github.com/dapr/components-contrib/blob/2006bcccb01e397e8987aa03fde249b4d5368b0c/pubsub/azure/eventhubs/eventhubs.go#L216-L226

pubsub/gcp/pubsub/pubsub.go
```diff
@@ -131,6 +142,24 @@ func createMetadata(pubSubMetadata pubsub.Metadata) (*metadata, error) {
 		}
 	}
 
+	result.MaxReconnectionAttempts = defaultMaxReconnectionAttempts
```
pkedy, line 145: This could be replaced by the initialization example from my other comment.

pkedy (commented):
Some suggestions for using the `dapr/kit/retry` package.

yordan-pavlov:
@pkedy 
> Yes. I think dapr/kit/retry probably makes sense here. No loop needed.

I don't think the behavior implemented with the retry loop in this PR can be achieved by simply using `dapr/kit/retry` without a loop - retry up to max retry **every time the subscription is disconnected** (not just one time).

Even if the back-off library is used within a loop, the behavior still won't be the same - retry attempts are used and slowly refilled back up over time - the longer the subscription connection lasts, the more retry attempts accumulate up to max retries. Only if retry attempts are used up in quick succession will the retry loop stop retrying.

pkedy:
@yordan-pavlov thanks for the explanation. I’ll take a closer look at this after a few days of Thanksgiving PTO.

pkedy:
@yordan-pavlov sorry for the delay. I'm preparing for a conference talk with week.

yordan-pavlov:
@yaron2 I have fixed the DCO issues and now all checks are passing

daixiang0 (approved):

berndverst:
@pkedy can you give this another look to decide whether `dapr/kit/retry` works here? If not, any thoughts on what should be improved?

Taction:
@yordan-pavlov can you please resolve conflicts?

yordan-pavlov:
@Taction done - conflict resolved

pubsub/gcp/pubsub/pubsub.go
yordan-pavlov: I have reviewed the code for 'dapr/kit/retry' available here https://github.com/dapr/kit/blob/main/retry/retry.go and the backoff library it uses (https://pkg.go.dev/github.com/cenkalti/backoff) and I don't think it's directly applicable since it has different behavior - it would NOT refill the number of reconnection attempts after a period of no errors, which means that it would be prone to stop reconnecting if the reconnect attempts are slowly exhausted over a long period of time. I am therefore resolving this without change.

berndverst (approved):

berndverst:
@yordan-pavlov I have merged this. Please make sure to complete the docs PR against the v1.7 branch of docs as soon as possible. https://github.com/dapr/docs/issues/1970

yordan-pavlov:
thanks @berndverst, docs pull request added here https://github.com/dapr/docs/pull/2262

PR #1298, opened 2021-11-12
Update to use latest Dev Container

berndverst:
# Description

Update to use latest Dev Container (introduced in https://github.com/dapr/dapr/pull/3887)


yaron2 (approved):
lgtm

berndverst:
This doesn't touch binaries. Test failure is golangci-lint timing out, but we don't need to run this test anyway.

Issue #1189, opened 2021-10-07
Support Azure Service Bus metadata handling in bindings component


CodeMonkeyLeet:
## Describe the feature
Per the implementation of #1071 it looks like the Azure Service Bus (ASB) queues bindings component should also be improved to support all [ASB message properties](https://docs.microsoft.com/en-us/rest/api/servicebus/message-headers-and-properties#message-headers). From the code it looks like only MessageID and CorrelationID are explicitly handled on Invoke, and Label is additionally handled on Read.

## Release Note
RELEASE NOTE: **ADD** Comprehensive Azure Service Bus message handling to Azure Service Bus Queues bindings.

berndverst:
FYI @artursouza

berndverst:
@ItalyPaleAle to verify. The share code of all ASB components probably already extracts and sets all ASB properties in Binding Invoke - but if it doesn't that can probably added easily.

ItalyPaleAle:
It doesn't look like it's happening to the binding: https://github.com/dapr/components-contrib/blob/e4fed1324f7e527d404c9189aa3ad6ffd5452406/bindings/azure/servicebusqueues/servicebusqueues.go#L176

I don't remember why it was done this way, probably to preserve the past behavior. But yes, it shouldn't be hard to add this with the shared code...

artursouza:
Needs to validate if this is still not done given the rewrites of the component. /cc @ItalyPaleAle

PR #1301, opened 2021-11-12
support ASB metadata handling in bindings

mukundansundar:
# Description

Support all ASB metadata handling in bindings component

## Issue reference

Please reference the issue this PR will close: #1189 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: https://github.com/dapr/docs/issues/1942


bindings/azure/servicebusqueues/servicebusqueues.go
```diff
@@ -156,31 +153,23 @@ func (a *AzureServiceBusQueues) Invoke(req *bindings.InvokeRequest) (*bindings.I
...
 	return nil, client.Send(ctx, msg)
 }
 
 func (a *AzureServiceBusQueues) Read(handler func(*bindings.ReadResponse) ([]byte, error)) error {
 	var sbHandler servicebus.HandlerFunc = func(ctx context.Context, msg *servicebus.Message) error {
-		_, err := handler(&bindings.ReadResponse{
-			Data:     msg.Data,
-			Metadata: map[string]string{id: msg.ID, correlationID: msg.CorrelationID, label: msg.Label},
+		mmd, err := asbmessage.NewMessageWithMetadataFromASBMessage(msg)
+		if err != nil {
```
CodeMonkeyLeet, line 167: The change in semantics here is tricky: functionally, nothing has changed in this implementation since `NewMessageWithMetadataFromASBMessage` never actually returns errors today, but you're suggesting that the Dapr _should_ enforce property mapping errors here. Some future dev may assume has always been the case and is well-handled when this could be a behavior breaking change when exercised for the first time.

My understanding of bindings is that they are generally dumb pipes as much as possible, with the responsibility on the app to handle any errors or issues, whereas pubsub is allowed to be more opinionated. As such, I would suggest that `NewMessageWithMetadataFromASBMessage` does not return errors, and any future metadata validation is the responsibility of the consuming component (i.e. pubsub component or bindings consumer).

CodeMonkeyLeet (commented):
Generally looks good to me, one of the maintainers might have to weigh in on the question of broker response validation in input bindings though.

bindings/azure/servicebusqueues/servicebusqueues.go
```diff
@@ -16,16 +16,13 @@ import (
 	"github.com/cenkalti/backoff/v4"
 
 	"github.com/dapr/components-contrib/bindings"
+	asbmessage "github.com/dapr/components-contrib/internal/component/azure/servicebus"
```
daixiang0, line 19: What does the `internal` mean?

bindings/azure/servicebusqueues/servicebusqueues.go
mukundansundar: Right now as you said, there is no error returned, I will remove that field completely and will also do it for the other function.

bindings/azure/servicebusqueues/servicebusqueues.go
mukundansundar: In general, we have used internal to notify that the module is to be used only within components contrib repo and not directly outside. It was used for factoring out common code in the case of retry and redis client connections .... I am using the same one for factoring out common code between bindings and pubsub ASB.

berndverst:
Please sync with @halspang on this PR because he is also making changes to the Azure Service Bus component, specifically for stable certification. This PR will almost certainly encounter major conflicts.

msfussell:
@mukundansundar - What is happening with this PR, since I believe this is the one that you showed in a community meeting, yes?

berndverst:
@mukundansundar is this something that you can complete for 1.8? If not maybe we should find a new owner for this or close this PR.

mukundansundar:
@berndverst I can complete it for 1.8 ...

artursouza:
@mukundansundar  any updates on this?

berndverst:
@mukundansundar any update?

PR #1306, opened 2021-11-15
Implements a binding -> pubsub adapter

### LINKED ISSUE #1305: [Proposal] Create an adapter to use PubSub components as Bindings
Currently, there are several bindings that have a pubsub counterpart (E.g. Kafka, Rabbit, Pulsar, NATS, Redis). In many cases the code in the PubSub component is far more robust. Given the similarity between bindings (Data + Metadata) and pubsub (Topic + Data + Metadata), an adapter could be used that specifies the read/invoke topics to subscribe and publish to. The primary motivation is that adapter reuses code and users can expect the same behavior between the binding and pubsub.

Here are some issues that would possibly be resolved for Kafka:

#1277
#1084
#955
#1006
#168

The downside is that if there are current inconsistencies in the component settings, then migrating to this approach would break existing versions. Since none of these components are marked as stable (yet), now is the time to do this if at all.

pkedy:
# Description

Implements an adapter that allows a PubSub component to be loaded as an Input/Output Binding.

A sample is implemented for Kafka and instantiated like this.

```golang
import (
	"github.com/dapr/components-contrib/bindings/kafka"
)

func example() {
	myBinding := kafka.NewKafkaAdapter(log)
}
```

This PR does not remove the current Kafka binding implementation (instantiated by `kafka.NewKafka(log)`).

If this approach were used, the docs pages _could_ point to their pubsub counterparts for referencing component settings.

## Issue reference

Closes #1305

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


bindings/pubsubadapter/pubsubadapter.go
```diff
@@ -0,0 +1,150 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/dapr/components-contrib/bindings"
+	"github.com/dapr/components-contrib/pubsub"
+)
+
+const (
+	defaultReadTopicAttribute   = "topics"
```
CodeMonkeyLeet, line 17: This is weirdly specific to the Kafka binding currently (i.e. there's no equivalent in MQTT, RabbitMQ etc. for taking a list of topics to listen on) and it doesn't look like it's actually used in the adapter, can it be dropped to simplify? It looks like the intention is for this to be breaking from the existing Kafka binding implementation anyway.

bindings/pubsubadapter/pubsubadapter.go
```diff
@@ -0,0 +1,150 @@
...
+	a.name = metadata.Name
+
+	return a.broker.Init(pubsub.Metadata{ //nolint:wrapcheck
+		Properties: metadata.Properties,
+	})
+}
+
+// Read subscribes to the configured topic and invokes `handler`
+// when a message is received.
+func (a *Adapter) Read(handler func(*bindings.ReadResponse) ([]byte, error)) error {
```
CodeMonkeyLeet, line 109: Bindings have `Read()` as a blocking function where `Subscribe()` is not. That probably needs to be preserved.

bindings/pubsubadapter/pubsubadapter.go
```diff
@@ -0,0 +1,150 @@
...
+		Metadata:   req.Metadata,
+	})
+
+	return nil, err //nolint:wrapcheck
+}
+
+// Operations returns "create" and "publish" as an operations
+// but operation is ignored by `Invoke`.
+func (a *Adapter) Operations() []bindings.OperationKind {
+	return []bindings.OperationKind{bindings.CreateOperation, bindings.OperationKind("publish")}
```
CodeMonkeyLeet, line 144: This suggests a broader problem that should be standardized across bindings: implementations should generally provide a broader range of consistent actions (i.e. other pubsub-like bindings should also advertise `publish`), but I suspect that most of them use `create` as publish to fit with how the conformance test behaves, which is an undesirable restriction.

CodeMonkeyLeet (commented):
In general, I think a better adapter strategy might be to create an adapter to expose a pubsub interface over a binding implementation, since pubsub tends to be a more opinionated abstraction and bindings may have to expose more lower-level knobs. 

Side note: PR title should probably be clarified, since this is creating a binding component given a pubsub implementation and the `->` is ambiguous in the relationship it implies.

bindings/pubsubadapter/pubsubadapter.go
pkedy: Great point

bindings/pubsubadapter/pubsubadapter.go
pkedy: Agreed.  For now, I'm trying not to break the existing bindings, however since non are "stable", we could make Invoke inspect the value of the operation.

pkedy:
> In general, I think a better adapter strategy might be to create an adapter to expose a pubsub interface over a binding implementation, since pubsub tends to be a more opinionated abstraction and bindings may have to expose more lower-level knobs.

I understand this sentiment. Bindings are free form so in theory could go from generic to more specific. I am hesitant about that approach because bindings basically have no contract. It would not be ideal to base a strong contract on something without compile time checks. It could certainly work though.

That aside, I think bindings for PubSub brokers is a special case. In this scenario, the bindings are a more opinionated version of the pubsub (specifically hiding the notion of a topic). The abstraction goes from PubSub (topic + data + metadata) to Binding (data + metadata). The goal is to eliminate redundant and not as well maintained code of some of the existing bindings and be able to certify them (and keep them certified) more easily.

bindings/pubsubadapter/pubsubadapter.go
pkedy: I think this is OK because the runtime starts a goroutine to handle calling `Read()`. In the adapter scenario, `Read()` would return immediately (no error) but the Subscribe goroutine would be running and log any errors. Same net affect as far as I can tell.

https://github.com/dapr/dapr/blob/42a901540f0f5a28a618721471ad208dfbf04a19/pkg/runtime/runtime.go#L2217-L2227

PR #1308, opened 2021-11-16
Add `make modtidy-all` action

CodeMonkeyLeet:
# Description

With the addition of the certification tests, changes in the go.mod definition for component-contrib will also need to be propagated to the go.mod for each of the tests under tests/certification.

- Add a `make modtidy-all` action to support updating all go.mod files.
- Update developing-component.md to include use of `make modtidy-all`.

## Issue reference

N/A

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


mukundansundar (commented):
looks good, I think the components_contrib.yaml workflow SHOULD be modified to check for diffs across all the go.mod files and if so error out instead of only looking at the root go.mod file as below.
https://github.com/dapr/components-contrib/blob/master/.github/workflows/components-contrib.yml#L65

CodeMonkeyLeet:
> looks good, I think the components_contrib.yaml workflow SHOULD be modified to check for diffs across all the go.mod files and if so error out instead of only looking at the root go.mod file as below. https://github.com/dapr/components-contrib/blob/master/.github/workflows/components-contrib.yml#L65

I'm open to that, although the cert test go.mod files are already checked as part of the [certification test workflow](https://github.com/dapr/components-contrib/blob/3eafb8bcd45def59fc42f96618354df673597a37/.github/workflows/certification.yml#L176), so coverage exists today. The question is whether they should show up as a build failure of components-contrib, which seems a little strong given that most folks aren't paying too much attention to cert tests and their failures yet.

artursouza (approved):

PR #1330, opened 2021-11-24
update some ratelimit_middleware.go comments

xj524598:
when i read ratelimit_middleware.go ,i found some comments are not right ,so i update some ratelimit_middleware.go comments


daixiang0 (approved):
@xj524598 thanks for your contribution!

Issue #1335, opened 2021-11-25
AWS SQS/SNS Wrong queue access policy (broken in last version (1.5.0))


mhozhda-agiliway:
## Expected Behavior

When some service is subscribed to many topics it's queue should have an access policy than allows all subscribed topics to publish events to the queue.

## Actual Behavior

Currently the generated queue access policy has a condition that only allows one of the topics to publish events to it.
Tested that it is not always the same topic (I assume the last subscribed topic rewrites policy and restricts access to itself only)

Generated access policy example:
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "Service": "sns.amazonaws.com"
      },
      "Action": "sqs:SendMessage",
      "Resource": "arn:aws:sqs:region:XXXXXXXXX:your-service-name",
      "Condition": {
        "ArnEquals": {
          "aws:SourceArn": "arn:aws:sns:region:XXXXXXXXX:lasttopicname"
        }
      }
    }
  ]
}

In the previous version of dapr 'Condition' is not specified at all.

Seems to be broken with this commit: 
https://github.com/dapr/components-contrib/commit/e9deaf37812131bca5d2c610572e290e92bb34ce
restrictQueuePublishPolicyToOnlySNS() method should add new sources to condition but not override existing

## Steps to Reproduce the Problem

Create a test app that subscribes to few (>1) topics. Run the app with AWS SQS/SNS component configured for pubsub.
Check that instead of allowing all subscribed topics to publish to created queue it only allows it to one of them
in generated queue access policy.

## Release Note

RELEASE NOTE:
Fixed bug with SQS/SNS wrong queue access policy

mhozhda-agiliway:
@amimimor Sorry for tagging you directly, but probably you can implement a quick fix for it

yaron2:
@amimimor please comment here with /assign.

amimimor:
@yaron2 we need to re-open this issue as I found and fixed another issue related to the fix (`localstack` testing appear to be inadequate in terms of resource access policies)

PR #1337, opened 2021-11-25
#1335 bugfix - multi SNS topics publishing policy to SQS

amimimor:
# Description

Bugfix - SNS access policy statement to SQS will not be overridden as a result of this bugfix and would not be restricted to just one topic in cases where 2 topics are used by a single dapr app (1 SQS queue)
Bugfix - minor behavioral error when handling any received message. all messages were logged as having an error (only visible at debug level though)
Gofmt - godot issues on comments

## Issue reference

Please reference the issue this PR will close: #1335 

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests


pubsub/aws/snssqs/snssqs.go
```diff
@@ -65,6 +65,27 @@ type snsSqsMetadata struct {
 	messageMaxNumber int64
 }
 
+type ArnEquals struct {
```
yaron2, line 68: Can this be made a non exported field? (ie. `arnEquals instead of ArnEquals`).

This comment is also true for `Condition` and `Statement`.

yaron2 (changes requested):

pubsub/aws/snssqs/snssqs.go
amimimor: I don't think it is technically possible as the JSON unmarshalling needs the fields to be public. but maybe I'm wrong. see this: https://stackoverflow.com/a/11129474

yaron2: The fields can be exported, that's fine - I'm referring to the struct itself.

amimimor: 👍🏿

yaron2 (approved):
lgtm

PR #1340, opened 2021-11-26
created name resolution local component

jigargandhi:
# Description

Created a local name resolution component that does parsing of command line arguments to daprd/daprd.exe to identify different dapr instances.

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/3256

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


beiwei30 (changes requested):
This change looks good to me. @jigargandhi would you mind to run `go mod tidy` for both pubsub.kafka certification test and pubsub.rabbitmq certification test?

jigargandhi:
@beiwei30 ran go mod tidy and fixed issues

beiwei30 (dismissed):
LGTM.

daixiang0 (dismissed):

nameresolution/local/local.go
```diff
@@ -0,0 +1,85 @@
...
+		logger: logger,
+	}
+}
+
+func (resolver *resolver) Init(metadata nameresolution.Metadata) error {
+	return nil
+}
+
+func (resolver *resolver) ResolveID(req nameresolution.ResolveRequest) (string, error) {
+	processes, err := ps.Processes()
```
artursouza, line 36: Can this be cached? This seem to be a super expensive operation per Invoke() call. Doing it for a cache miss is kind of OK but once cached, we should be able to reuse. Also, invalidating the cache is super important. Maybe we change this logic to be running in the background updating a lookup table and then the ResolveID() method just does a lookup.

artursouza (changes requested):

nameresolution/local/local.go
jigargandhi: The value can be cached and we can run a go routine which invalidates the cache every 30 seconds.
We will then need to store process id also in cache, and validate that the process id is actually for dapr process only (it may be that the process id is reassigned)
If the process id does not match then it will start the cache invalidation process again. 
Also, the cache invalidation process will also run only once 30 seconds.

Taction:
@jigargandhi Thanks for this PR! Would you please fix DCO issue? You can refer to [here](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md)  about how to fix it.

jigargandhi:
@Taction Sure, I will be adding one more commit as per a review comment from @artursouza , will sign off that commit.

nameresolution/local/local.go
artursouza: Correct. We need to keep track of the PIDs too.

beiwei30:
@jigargandhi Jigar, at the same time, pls. fix [go mod tidy check](https://github.com/dapr/components-contrib/runs/4748225282?check_suite_focus=true) failure.

jigargandhi:
please review this https://github.com/dapr/components-contrib/pull/1455 PR, closing this as messed up sign-off

PR #1342, opened 2021-11-29
File based naming resolver

beiwei30:
# Description

Provide a file based naming resolver to replace mDNS so that dapr can address each other either on the same host or via shared file system, which would be more stable and friendly for development.

The default directory to keep naming informations is: `$HOME/.dapr/naming`, and it can be customized in Dapr's config.yaml:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Configuration
metadata:
  name: appconfig
spec:
  nameResolution:
    component: "file"
    configuration:
      dir: "/var/tmp/dapr-naming"
```

Pro and con of this solution:
* Pro: Since the name resolving is file based, it means that it can not only work in slim mode, but also in standalone mode, since the directory to keep the naming informations can be anywhere as long as it's accessible to Dapr's instances.
* Con: Since there has no lifecycle mechanism for components, this solution doesn't have chance to remove (unregister) its naming info from the file when Dapr instance goes away. It may not a big deal in development stage, but in the worst case user needs to know where the directory locates and delete the files manually.

## Issue reference

This PR will close: dapr/dapr#3256

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


yaron2:
There's also this PR: https://github.com/dapr/components-contrib/pull/1340.

Both close the same issue, however I do not see a problem accepting both. Thoughts?

beiwei30:
will continue to update dapr/dapr and dapr/docs once this solution is accepted.

beiwei30:
pls. use https://github.com/dapr/components-contrib/pull/1380 for further review.

Issue #1322, opened 2021-11-22
Support path-style access for S3 bindings


ga0x1ang:
## Describe the feature
While AWS S3 is deprecating path-styled access, a lot S3 API compatible products are still supporting it.
For example, MinIO is using path-style by default.
With this feature, S3 binding can be used by more S3 compatible products.

S3 SDK supports path-style access by passing **S3ForcePathStyle** = true to client.
S3 Binding can achieve it through existing **mergeWithRequestMetadata** function

## Release Note

RELEASE NOTE: Add S3ForcePathStyle metadata key to S3 binding

rainfd:
I aggre with it. I will commit the code later.

yaron2:
keep alive

yaron2:
Closed via #1360

PR #1360, opened 2021-12-04
add S3ForcePathStyle support for s3 binding

rainfd:
# Description

Add S3ForcePathStyle support for S3 API compatible products, like Minio.

## Issue reference

#1322 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#2022

### DIFF
--- a/bindings/aws/s3/s3.go
+++ b/bindings/aws/s3/s3.go
@@ -42,14 +42,15 @@ type AWSS3 struct {
 }
 
 type s3Metadata struct {
-	Region       string `json:"region"`
-	Endpoint     string `json:"endpoint"`
-	AccessKey    string `json:"accessKey"`
-	SecretKey    string `json:"secretKey"`
-	SessionToken string `json:"sessionToken"`
-	Bucket       string `json:"bucket"`
-	DecodeBase64 bool   `json:"decodeBase64,string"`
-	EncodeBase64 bool   `json:"encodeBase64,string"`
+	Region         string `json:"region"`
+	Endpoint       string `json:"endpoint"`
+	AccessKey      string `json:"accessKey"`
+	SecretKey      string `json:"secretKey"`
+	SessionToken   string `json:"sessionToken"`
+	Bucket         string `json:"bucket"`
+	DecodeBase64   bool   `json:"decodeBase64,string"`
+	EncodeBase64   bool   `json:"encodeBase64,string"`
+	ForcePathStyle bool   `json:"forcePathStyle,string"`
 }
 
 type createResponse struct {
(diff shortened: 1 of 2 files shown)


bindings/aws/s3/s3.go
```diff
@@ -79,10 +80,13 @@ func (s *AWSS3) Init(metadata bindings.Metadata) error {
...
+
+	cfg := aws.NewConfig().WithS3ForcePathStyle(m.ForcePathStyle)
+
 	s.metadata = m
-	s.s3Client = s3.New(session)
-	s.downloader = s3manager.NewDownloader(session)
-	s.uploader = s3manager.NewUploader(session)
+	s.s3Client = s3.New(session, cfg)
+	s.downloader = s3manager.NewDownloaderWithClient(s.s3Client)
+	s.uploader = s3manager.NewUploaderWithClient(s.s3Client)
```
mukundansundar, line 89: is there a reason to change to `NewDownloaderWithClient` from `NewDownloader` ?

rainfd: `S3ForcePathStyle` only can be set with s3 client or session. I couldn't find a way to set it up in `NewDownloader`.

yaron2:
@mukundansundar please see comment and re-review.

yaron2:
@rainfd we now need to sign off our commits, please see here: https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md#how-to-add-sign-offs-retroactively.

bindings/aws/s3/s3.go
yaron2: @mukundansundar

rainfd:
@yaron2 comment has been modified.

bindings/aws/s3/s3.go
mukundansundar: > `S3ForcePathStyle` only can be set with s3 client or session. I couldn't find a way to set it up in `NewDownloader`.

@rainfd if it can be set in the session why not set it there, since session is already being used?

rainfd: > @rainfd if it can be set in the session why not set it there, since session is already being used?

`func GetClient(accessKey string, secretKey string, sessionToken string, region string, endpoint string) (*session.Session, error)`
session get by `GetClient` in authentication/aws/aws.go,  setting ForcePathStyle here is also ok, like
`func GetClient(accessKey string, secretKey string, sessionToken string, region string, endpoint string, forcePathStyle bool) (*session.Session, error) {` 
 but `GetClient` is used by other aws componment. Should I add modify here?

mukundansundar: Thanks fir explaining that ... And no no need to make change in the common method. 
Can you add a docs PR for this and also unit tests to make sure that the metadata is parsed correctly?

rainfd: @mukundansundar test and docs have already added.
dapr/docs#2054

mukundansundar (approved):
lgtm

yaron2:
@rainfd can you please open a docs PR that updates the metadata?

rainfd:
> @rainfd can you please open a docs PR that updates the metadata?

dapr/docs#2054  What's missing?

yaron2:
> > @rainfd can you please open a docs PR that updates the metadata?
> 
> [dapr/docs#2054](https://github.com/dapr/docs/pull/2054) What's missing?

Oh didn't see that, all good.

PR #1361, opened 2021-12-04
Tidy & Pin certification tests to latest Dapr/Dapr Master Commits

berndverst:
Go mod tidy
Pin certification tests to latest dapr/dapr in Master


daixiang0 (commented):
I perfer to pin a RC version rather than master as develop should always happen based on a stable branch.

/cc @artursouza @yaron2

berndverst:
@daixiang0 we will want to know if anything introduced into dapr/dapr master or components-contrib master breaks our certification tests. The idea is that we fix master immediately.

In the Release branches like `release-1.5` we keep the versions pinned.

Also, the certification test framework is still a work in progress and new features may be added. This should not have to wait for a release as this is internal details - not something externally facing.

artursouza:
Master branch should keep track of dapr/dapr master IMO. We can tag to an RC once we cut the release branch. This is important for us to catch regressions during development and not only after we start release activities.

artursouza (approved):

PR #1364, opened 2021-12-06
Merge release1.5 (hotfix 1.5.1) into master

berndverst:
Merge release1.5 (hotfix 1.5.1) into master

It is expected that this PR does not contain any code changes because release 1.5.1 only had cherry-picks from master.

We deliberately keep all go.mod / go.sum versions from master.


daixiang0:
@berndverst seems no changes?

PR #1369, opened 2021-12-09
Support CloudEvent optional and extension attributes: first step

CrazyHZM:
# Description
https://github.com/dapr/dapr/issues/3991 First step：
Modify the Envelope method to support attributes. The time, dataschema, and subject in the CloudEvent spec are all placed in attributes.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


pubsub/envelope.go
```diff
@@ -84,14 +87,11 @@ func NewCloudEventsEnvelope(id, source, eventType, subject string, topic string,
 		PubsubField:          pubsubName,
 		TraceIDField:         traceID,
 		TraceStateField:      traceState,
+		AttributesField:      attributes,
 	}
 
 	ce[ceDataField] = ceData
 
-	if subject != "" {
```
mukundansundar, line 91: I think this is a breaking change.
Right now if we are not following the complete CE spec, I would much rather keep this field as well as add one in the attributes field if possible. Announce deprecation of the outer subject field and in 2 releases remove it. 
/cc @yaron2 @artursouza @pkedy

pubsub/rocketmq/rocketmq.go
```diff
@@ -249,8 +249,14 @@ func (r *rocketMQ) adaptCallback(topic, consumerGroup, mqType, mqExpr string, ha
 	return func(ctx context.Context, msgs ...*primitive.MessageExt) (mqc.ConsumeResult, error) {
 		success := true
 		for _, v := range msgs {
+			attributes := map[string]interface{}{
```
mukundansundar, line 252: is this the only place where subject is added?

CrazyHZM: Time and dataschema can also be added, but these are optional, including subject

pubsub/envelope.go
CrazyHZM: Agree to keep this field temporarily, wait for more people to express their opinions, I am making changes.

pkedy: It does make sense to keep the current `NewCloudEventsEnvelope` method and store the optional, [`subject` attribute](https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#subject) in the attributes internally.

One question I have is why `attributes` is a `[]byte` instead of a `map[string]interface{}` to support the possible values from the CE spec?

CrazyHZM: @pkedy 
The reason for using byte is that in the publish event, the definition of PublishEventRequest needs to be expressed by byte. The advantage of using []byte instead of map[string]interface{} is that dapr will not need to re-parse the attributes.
In dapr/dapr#3991 I gave the second step.

greghoover:
Did this or other related steps make it into the 1.6 release?

CrazyHZM:
> Did this or other related steps make it into the 1.6 release?

@greghoover 
There is no merge related content to 1.6 for the time being, and a pr is resubmitted: https://github.com/dapr/components-contrib/pull/1460
Mainly to fix the DCO problem.

PR #1299, opened 2021-11-12
Added content-type to SetRequest and blob storage to test

jigargandhi:
# Description

Added Content-Type to SetRequest.

## Issue reference
https://github.com/dapr/dapr/issues/2892

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #_[issue number]_

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #1376, opened 2021-12-13
Added ContentType to pubsub/binding/state request-response

jigargandhi:
# Description

Added Content-Type to SetRequest.

Created a new PR for https://github.com/dapr/components-contrib/pull/1299

## Issue reference
https://github.com/dapr/dapr/issues/2892

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


jigargandhi:
@artursouza please review this.. messed up previous one due to DCO

bindings/responses.go
```diff
@@ -11,8 +11,9 @@ import (
 
 // ReadResponse is the return object from an dapr input binding.
 type ReadResponse struct {
-	Data     []byte            `json:"data"`
-	Metadata map[string]string `json:"metadata"`
+	Data        []byte            `json:"data"`
+	Metadata    map[string]string `json:"metadata"`
+	ContentType *string           `json:"contentType,omitempty"`
```
pkedy, line 16: Nit: You can use `string` instead of `*string` here. The zero value of a string in Go works well here. Same goes for all the other occurrences below.

artursouza: This is something I asked for to keep it similar to how eTag is done.

artursouza (approved):

artursouza:
@jigargandhi This change will also need to have issues in all SDKs, Docs and quickstarts too. Since it is optional, we don't need to change all docs but add something about it.

bindings/responses.go
jigargandhi: resolving the comment as per above comment

artursouza:
@yaron2  I am leaning towards moving this to 1.7 so we have time to do all the changes in SDKs. Thoughts?

yaron2:
> @yaron2 I am leaning towards moving this to 1.7 so we have time to do all the changes in SDKs. Thoughts?

Supportive.

berndverst (approved):
Hadn't seen the PR before but this looks good to me. (I recently used the Blob Storage SDK extensively to prepare the Output Binding for stable release.)

PR #1380, opened 2021-12-14
feat: file based naming resolver

beiwei30:
# Description

Provide a file based naming resolver to replace mDNS so that dapr can address each other either on the same host or via shared file system, which would be more stable and friendly for development.

The default directory to keep naming informations is: `$HOME/.dapr/naming`, and it can be customized in Dapr's config.yaml:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Configuration
metadata:
  name: appconfig
spec:
  nameResolution:
    component: "file"
    configuration:
      dir: "/var/tmp/dapr-naming"
```

Pro and con of this solution:
* Pro: Since the name resolving is file based, it means that it can not only work in slim mode, but also in standalone mode, since the directory to keep the naming informations can be anywhere as long as it's accessible to Dapr's instances.
* Con: Since there has no lifecycle mechanism for components, this solution doesn't have chance to remove (unregister) its naming info from the file when Dapr instance goes away. It may not a big deal in development stage, but in the worst case user needs to know where the directory locates and delete the files manually.

## Issue reference

This PR will close: dapr/dapr#3256

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


yaron2:
@daixiang0 can you please review this?

daixiang0:
@yaron2 sure.

nameresolution/file/file.go
```diff
@@ -0,0 +1,190 @@
...
+
+	"github.com/dapr/components-contrib/nameresolution"
+	"github.com/dapr/kit/config"
+	"github.com/dapr/kit/logger"
+)
+
+const (
+	dirKey           = "dir"
+	defaultDaprDir   = ".dapr"
+	defaultNamingDir = "naming"
```
daixiang0, line 23: How about put it into `$HOME/.dapr/components` as it is a component too?

nameresolution/file/file.go
```diff
@@ -0,0 +1,190 @@
...
+	f := filepath.Join(r.dir, info.AppID)
+	namingInfos, err := loadNamingInfo(f)
+	if err != nil {
+		r.logger.Errorf("fail to load naming info from file: %s", f)
+		return err
+	}
+
+	index := -1
+	for i, ni := range namingInfos {
+		if ni.AppID == info.AppID &&
```
daixiang0, line 92: How about converting check into a function, then any changes can happen in that?

nameresolution/file/file.go
```diff
@@ -0,0 +1,190 @@
...
+	if err = os.MkdirAll(r.dir, os.ModePerm); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// ResolveID resolves name to address via file.
+func (r *resolver) ResolveID(req nameresolution.ResolveRequest) (string, error) {
+	lock := r.newFileLock(req.ID)
```
daixiang0, line 162: Seems we need to read files when a request comes, how about reading one time then handle all route info, only re-read when the file changes?

yaron2:
ping @beiwei30

nameresolution/file/file.go
artursouza: Can this be cached too? Maybe the file update happens in the background instead and the ResolveID() method will only do a lookup on a in-memory hash table.

beiwei30:
> ping @beiwei30

sorry for late response, will follow up the review comments today.

nameresolution/file/file.go
beiwei30: refactor it into separated method, pls. check.

nameresolution/file/file.go
beiwei30: introduce filewatcher to load naming info only when necessary.

beiwei30:
@daixiang0 @artursouza I have updated this PR, pls. take a look again.

nameresolution/file/file.go
beiwei30: Instead of putting it in components dir, it should be configured with $HOME/.dapr/config.yaml (see https://github.com/dapr/components-contrib/pull/1380#issue-1079414460), what do you think, @daixiang0 ?

Taction:
@beiwei30 Please resolve the conflicts.

yaron2:
keepalive

beiwei30:
@yaron2 would you mind to review it again? I resolved the conflict today.

nameresolution/file/file.go
daixiang0: I think so.

daixiang0 (dismissed):
LGTM.

yaron2:
@beiwei30 can you please run `go mod tidy` and push the changes? Thanks!

berndverst:
> @beiwei30 - When the app becomes unhealthy, there is lot of manual activity involved -
> 
> 1. User should manually delete the name resolution file and lock file.
> 2. All the app instances should be restarted to create a new file with updated name-resolution information for all app instances.[Since we have a common name resolution information for all apps with same id]
> 
> Also having 777 file permissions allows anyone to write to the nameresolution file which is a security issue, since the consuming app can be misguided to connect to any endpoint.
> 
> To complete this solution, there should be a cleanup of the nameresolution information. when file is chosen as name resolution - dapr cli can periodically check if all the apps as running and clean up the nameresolution information to be accurate - thoughts? @mukundansundar @berndverst

I think we need someone to create a design document for the overall design / experience including the CLI, and the various edge cases before we should proceed here.

This feature is useful, but there seem to be plenty of gaps.

ItalyPaleAle:
@berndverst and I synced yesterday, and I agree a file-based nameresolver should happen in 1.13, but IMHO this should be based on SQLIte.

Using a single SQLite database will already solve problems such as concurrent access by multiple sidecars (locking, etc). A single file is also easier to handle and to garbage-collect. Lastly, SQLite is also generally [faster than the filesystem](https://www.sqlite.org/fasterthanfs.html) for these kinds of things.

Issue #1371, opened 2021-12-09
RabbitMQ: channel not initialized, even if channel exists


WhoAteDaCake:
## Expected Behavior
Dapr to send messages, since channel exists

## Actual Behavior

Dapr complains about. 
error in subscriber for scraper-scraper in ensureSubscription: channel not initialized"

## Steps to Reproduce the Problem

I'm not quite sure? It started happening since I've updated to 1.5.1

![logs](https://user-images.githubusercontent.com/16735813/145417059-c83b7213-0df3-4248-a2e9-d38b96582417.png)
![queue](https://user-images.githubusercontent.com/16735813/145417066-d6ebef6e-800a-4f0f-90fa-ee6646746bec.png)
![exchange](https://user-images.githubusercontent.com/16735813/145417074-ab674ea1-7253-47ee-a7db-63f3560d336c.png)

## Release Note

RELEASE NOTE:

dmitsh:
@WhoAteDaCake , could you provide instructions on how to reproduce this issue?

WhoAteDaCake:
Sorry, but I couldn't catch the exact scenario. I'll try to replicate over the weekend. It started happening as soon as I updated from 1.3.1 -> 1.5.1. We reverted back now and the issue has stopped

dmitsh:
@WhoAteDaCake , I have a fix for #1372.
Could you try it with your test case by changing `daprd` image from `daprio/daprd:1.5.1` to `dmitsh/daprd:dev-rmq-linux-amd64` ?

WhoAteDaCake:
It seems to be working as expected now. Thank you for your work @dmitsh

andreatosato:
This not resolve my problem.

I create this branch repository: https://github.com/andreatosato/dapr-cloudchampion/tree/rabbitmq-error
```
time="2021-12-14T21:22:11.8985009Z" level=error msg="rabbitmq pub/sub: error in subscriber for cloudchampion-web-orderstatus in ensureSubscription: channel not initialized" app_id=cloudchampion-web instance=c821022bbc95 scope=dapr.contrib type=log ver=edge

time="2021-12-14T21:22:11.8985535Z" level=warning msg="rabbitmq pub/sub: subscriber is reconnecting in 3s ..." app_id=cloudchampion-web instance=c821022bbc95 scope=dapr.contrib type=log ver=edge

[...]

PR #1382, opened 2021-12-16
pub/sub rabbitmq: add publishing retries

dmitsh:
# Description

add publishing retries in RabbitMQ pib/sub

## Issue reference

Please reference the issue this PR will close: #1371

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


pubsub/rabbitmq/rabbitmq.go
```diff
@@ -205,18 +207,26 @@ func (r *rabbitMQ) publishSync(req *pubsub.PublishRequest) (rabbitMQChannelBroke
 func (r *rabbitMQ) Publish(req *pubsub.PublishRequest) error {
 	r.logger.Debugf("%s publishing message to %s", logMessagePrefix, req.Topic)
 
-	channel, connectionCount, err := r.publishSync(req)
-	if err != nil {
+	attempt := 0
```
artursouza, line 210: I am debating if we should have a flag or not to enable/disable retries in component metadata. The upcoming resiliency work, this flag would be deprecated soon after.

Another option is not to merge this and wait for the resiliency.

dmitsh: We promised this fix in 1.6. I would suggest adding `publishMaxRetries` flag to RabbitMQ metadata and setting it by default to `3`.
Once resiliency is merged, we will deprecate it.
/cc @pkedy @yaron2

yaron2: This work will probably conflict with resiliency.

Things like connection establishment for the first time *could* be retried inside the component because these are unlikely to be covered by the resiliency feature, but I agree with @artursouza that IF there's overlap, I'm inclined to keep this open for now and wait.

@dmitsh we could do as you suggest, but adding a flag for several weeks just to deprecate it after one release in a beta/stable level component seems unnecessary to me.

dmitsh: So, if I can summarize, there are 3 options:
 (1) do nothing now, and integrate with the resiliency feature later. In the meantime, advise users to implement retries in their application.
 (2) add hard-coded retry logic (implemented in this PR), which will be later replaced with the resiliency feature
 (3) add temporary configuration for publishing retries, which will be later deprecated once the resiliency feature is merged.
 
I agree that the third option is not desired.

I would personally tend to the second option, because it unblocks current users facing this issue, and could be seamlessly replaced with the resilience feature (regardless of when it will be delivered).

yaron2: > So, if I can summarize, there are 3 options: (1) do nothing now, and integrate with the resiliency feature later. In the meantime, advise users to implement retries in their application. (2) add hard-coded retry logic (implemented in this PR), which will be later replaced with the resiliency feature (3) add temporary configuration for publishing retries, which will be later deprecated once the resiliency feature is merged.
> 
> I agree that the third option is not desired.
> 
> I would personally tend towards the second option, because it unblocks current users facing this issue, and could be seamlessly replaced with the resilience feature (regardless of when it will be delivered).

Thanks for the summary. I actually vote for (2), but without removing it when resiliency is in, and here's why:

Resiliency is an opt-in feature. If we put the retries now, users will get a working component for 1.6, then in 1.7 they will experience a regression to 1.5 behavior UNLESS they specify a resiliency policy. users will not be able to possibly know they should be doing that, because they have no idea we removed *default* retries in place of optional ones. Also, the resiliency configuration to match the deprecated component retries are nearly impossible to reason about from a user perspective.

This is a perfectly valid example of keeping component retries to deal with a *known* issue and provide component level retries to deal with that.

Then, resiliency policies naturally fit in with whatever behavior the component is exhibiting with the component level retries included.
In other words, the component has default retries to guarantee an acceptable level of performance. Then, developers may want to put resiliency features to account for extra retries because their RabbitMQ environment needs it (based on configuration, hardware, resource constraints, etc.)

cc @artursouza

dmitsh: > Thanks for the summary. I actually vote for (2) as well

Do you want to approve this PR?

yaron2: > > Thanks for the summary. I actually vote for (2) as well
> 
> Do you want to approve this PR?

I'd like to get @artursouza's take on this approach too.

artursouza: LGTM

artursouza (approved):

PR #1404, opened 2022-01-04
fixing tests

dmitsh:
# Description

Fixing test failures


.github/workflows/certification.yml
```diff
@@ -182,7 +182,7 @@ jobs:
     - name: Check that go mod tidy is up-to-date
       working-directory: ${{ env.TEST_PATH }}
       run: |
-        go mod tidy
+        go mod tidy -compat=1.17
```
daixiang0, line 185: Now we pin Go as 1.17?

PR #1417, opened 2022-01-06
Update all certification tests to use latest Dapr/Dapr

berndverst:

# Description

#1416 broke something, this should hopefully resolve that :) FYI @dmitsh


dmitsh (approved):

Taction (approved):

daixiang0:
I think we should always add this to release plan as well.

PR #1418, opened 2022-01-07
hotfix: bindings.cron Input & Output communicates through a signal queue

1046102779:
`InputBinding`  and `OutputBinding` are two different instances, it's `stopChan` variable is two different scope instances, these cannot complete the communication with each other. So bindings.cron need to global signal channel in bindings package.

The test result in multi bindings.cron instances as follows:
image


1046102779:
https://github.com/dapr/components-contrib/pull/1405

yaron2 (approved):
lgtm

Issue #1426, opened 2022-01-09
Pubsub AWS SNS/SQS - assets operations timeouts, retrying, backoffs


amimimor:
## Describe the feature
The feature provides the following capabilities:
* allow the user to define a timeout for AWS assets related operations, instead of relying on `net/http` timeouts. Assets operations being, but not limited to: SNS topics creation/verification, SQS queues creation/verification, Subscriptions, STS caller identity calls, etc.
* use `context` and `cancel` in all AWS go sdk API calls
* component `Close` method should call the  `cancel` function of the parent context and cleanly cancel child operations
* utilize exponential backoff* when errors in pulling SQS messages occur (such as in network interruptions). This would lower the log flooding with error messages upon retries of the component
* publish messages with user configurable `retry` implementation
* introduce new metadata config field: `disableDeleteOnRetryLimit` the prevent the component from deleting the message (effectively ack'ing it) when the number of retries passes the limit (current behavior)

 *internally, AWS go sdk has a default retry with backoff mechanism but it is used anew for each poll interval, within an endless loop, therefore a Dapr custom backoff should be used per polling cycle, as would be clearer in the following PR

## Release Note

RELEASE NOTE:

PR #1433, opened 2022-01-11
Pubsub AWS SNS/SQS - adding context, cancellation, timeouts, retrying \w backoff & disable delete of messages on failure

amimimor:
# Description

* allow the user to define a timeout for AWS assets related operations, instead of relying on `net/http` timeouts. Assets operations being, but not limited to: SNS topics creation/verification, SQS queues creation/verification, Subscriptions, STS caller identity calls, etc.
* use `context` and `cancel` in all AWS go sdk API calls
* component `Close` method should call the  `cancel` function of the parent context and cleanly cancel child operations
* utilize exponential backoff* when errors in pulling SQS messages occur (such as in network interruptions). This would lower the log flooding with error messages upon retries of the component
* publish messages with user configurable `retry` implementation
* introduce new metadata config field: `disableDeleteOnRetryLimit` the prevent the component from deleting the message (effectively ack'ing it) when the number of retries passes the limit (current behavior)
* introduce new metadata config field: `assetsManagementTimeoutSeconds` that sets a timeout for all AWS API calls that are not related to either publish message nor pull message, but administrative in nature, such as topic/subscription/queue creation and other queries
* refactoring of all metadata related ops into a separate file
* refactoring of the former `handleMessage` method, for better readability, into smaller methods called directly from the method pulling SQS messages

 *internally, AWS go sdk has a default retry with backoff mechanism but it is used anew for each poll interval, within an endless loop, therefore a Dapr custom backoff should be used per polling cycle, as would be clearer in the following PR

## Issue reference

Please reference the issue this PR will close: #1426 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#2092

### DIFF
--- a/pubsub/aws/snssqs/metadata.go
+++ b/pubsub/aws/snssqs/metadata.go
@@ -1,0 +1,252 @@
+package snssqs
+
+import (
+	"errors"
+	"fmt"
+	"strconv"
+
+	"github.com/dapr/components-contrib/pubsub"
+)
+
+type snsSqsMetadata struct {
+	// aws endpoint for the component to use.
+	Endpoint string
+	// access key to use for accessing sqs/sns.
+	AccessKey string
+	// secret key to use for accessing sqs/sns.
+	SecretKey string
+	// aws session token to use.
+	SessionToken string
+	// aws region in which SNS/SQS should create resources.
+	Region string
+	// name of the queue for this application. The is provided by the runtime as "consumerID".
+	sqsQueueName string
+	// name of the dead letter queue for this application.
+	sqsDeadLettersQueueName string
+	// flag to SNS and SQS FIFO.
+	fifo bool
+	// a namespace for SNS SQS FIFO to order messages within that group. limits consumer concurrency if set but guarantees that all
+	// published messages would be ordered by their arrival time to SQS.
+	// see: https://aws.amazon.com/blogs/compute/solving-complex-ordering-challenges-with-amazon-sqs-fifo-queues/
+	fifoMessageGroupID string
+	// amount of time in seconds that a message is hidden from receive requests after it is sent to a subscriber. Default: 10.
+	messageVisibilityTimeout int64
+	// number of times to resend a message after processing of that message fails before removing that message from the queue. Default: 10.
+	messageRetryLimit int64
+	// upon reaching the messageRetryLimit, disables the default deletion behaviour of the message from the SQS queue, and resetting the message visibilty on SQS
+	// so that other consumers can try consuming that message.
+	disableDeleteOnRetryLimit bool
+	// if sqsDeadLettersQueueName is set to a value, then the messageReceiveLimit defines the number of times a message is received
+	// before it is moved to the dead-letters queue. This value must be smaller than messageRetryLimit.
+	messageReceiveLimit int64
+	// amount of time to await receipt of a message before making another request. Default: 1.
+	messageWaitTimeSeconds int64
+	// maximum number of messages to receive from the queue at a time. Default: 10, Maximum: 10.
+	messageMaxNumber int64
+	// disable resource provisioning of SNS and SQS.
+	disableEntityManagement bool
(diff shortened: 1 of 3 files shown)


pubsub/aws/snssqs/metadata.go
```diff
@@ -0,0 +1,252 @@
...
+
+//nolint:cyclop
+func (s *snsSqs) getSnsSqsMetatdata(metadata pubsub.Metadata) (*snsSqsMetadata, error) {
+	md := snsSqsMetadata{}
+	props := metadata.Properties
+	md.sqsQueueName = metadata.Properties["consumerID"]
+	s.logger.Debugf("Setting queue name to %s", md.sqsQueueName)
+
+	if val, ok := getAliasedProperty([]string{"Endpoint", "endpoint"}, metadata); ok {
+		s.logger.Debugf("endpoint: %s", val)
```
daixiang0, line 98: A debug log for `snsSqsMetadata` in the end of func is enough I think.

pubsub/aws/snssqs/metadata.go
```diff
@@ -0,0 +1,252 @@
...
+
+		if timeout < 1 {
+			return nil, errors.New("messageVisibilityTimeout must be greater than 0")
+		}
+
+		md.messageVisibilityTimeout = timeout
+	}
+
+	if val, ok := props["messageRetryLimit"]; !ok {
+		md.messageRetryLimit = 10
```
daixiang0, line 136: Does it comment in doc?

pubsub/aws/snssqs/metadata.go
```diff
@@ -0,0 +1,252 @@
...
+func parseFloat64(input string, propertyName string) (float64, error) {
+	val, err := strconv.ParseFloat(input, 64)
+	if err != nil {
+		return 0, fmt.Errorf("parsing %s failed with: %w", propertyName, err)
+	}
+	return val, nil
+}
+
+//nolint:cyclop
+func (s *snsSqs) getSnsSqsMetatdata(metadata pubsub.Metadata) (*snsSqsMetadata, error) {
```
daixiang0, line 91: This function is really long as I do not think of a good idea to solve it.

/cc @artursouza do you have a good idea for this?

pubsub/aws/snssqs/snssqs.go
```diff
@@ -329,31 +146,56 @@ func (s *snsSqs) Init(metadata pubsub.Metadata) error {
...
 	s.metadata.accountID = *callerIDOutput.Account
-
-	s.snsClient = sns.New(sess)
-	s.sqsClient = sqs.New(sess)
-
 	return nil
 }
 
 func (s *snsSqs) buildARN(serviceName, entityName string) string {
-	// arn:aws:sns:us-east-1:302212680347:aws-controltower-SecurityNotifications
```
daixiang0, line 356: Keep an example or a doc link is useful.

daixiang0 (commented):
@amimimor Thanks for your great contribution! Please add more tests to cover more cases.

pubsub/aws/snssqs/metadata.go
amimimor: done: moved to the end of refactored method

pubsub/aws/snssqs/metadata.go
amimimor: yes

pubsub/aws/snssqs/metadata.go
amimimor: done. I have broken it up

pubsub/aws/snssqs/snssqs.go
amimimor: this was in the source by mistake (with my private aws account id oh dear) and therefore removed, besides that it is internal and builds a standard AWS ARN so I don't think it needs documentation

Issue #1445, opened 2022-01-16
#1426 bugs - visibility timeout reset; disable entity management in DLQ 


amimimor:
## Expected Behavior

I've introduced 2 bugs in https://github.com/dapr/components-contrib/pull/1433 that need fixing (PR will follow this):
1. reset message visibility is reset the message timeout to a new default instead of zero
2. writing SQS settings when DLQ is used although we shouldn't if the user asked to `disableEntityManagement` 

## Actual Behavior

1. reset visibility to zero
2. don't write SQS attributes if the user asked to `disableEntityManagement`

## Release Note

RELEASE NOTE:

PR #1446, opened 2022-01-16
bugfixes: visibility timeout, dlq queue attributes

amimimor:
Signed-off-by: Amit Mor <amit.mor@hotmail.com>

# Description

bugfixes
1. reset visibility timeout to zero so other consumer can immediately consume retried but failed messages
2. not setting/rewritting SQS attributes when DLQ is set and also disableEntityManagement 

## Issue reference

Please reference the issue this PR will close: #1445 

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/pubsub/aws/snssqs/snssqs.go
+++ b/pubsub/aws/snssqs/snssqs.go
@@ -482,7 +482,7 @@ func (s *snsSqs) resetMessageVisibilityTimeout(queueURL string, receiptHandle *s
 	changeMessageVisibilityInput := &sqs.ChangeMessageVisibilityInput{
 		QueueUrl:          aws.String(queueURL),
 		ReceiptHandle:     receiptHandle,
-		VisibilityTimeout: aws.Int64(s.metadata.messageVisibilityTimeout),
+		VisibilityTimeout: aws.Int64(0),
 	}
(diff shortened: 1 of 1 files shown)


daixiang0:
Please add tests to ensure fix.

amimimor:
> Please add tests to ensure fix.

Thanks for your feedback.

Testing this component is currently done (semi-)manually over `localstack` and later over a live AWS environment. However, `localstack` is not as enforcing as a real AWS environment in many aspects and thus provides basic SNS/SQS functionality. 

Testing manually over a real environment takes time and some of the component features are complicated to test (like resetting a message visibility timeout so that other consumers can consume the messages that the consumer pulled and failed to process)

I think that by their nature, these bugs can be covered by heavily investing in mocking AWS services and tests in the spirit of certification tests but on a live AWS environment. Unit testing just doesn't cover these kind of bugs so until mocks and certification tests are in place, your suggestions, as valuable as they are, are currently unrealistic as my time is limited

daixiang0:
@amimimor got it, let's create an issue to track it :)

amimimor:
> @amimimor got it, let's create an issue to track it :)

Will do. Thanks!

berndverst (approved):

PR #1460, opened 2022-01-25
Support CloudEvent optional and extension attributes: first step

CrazyHZM:
# Description
https://github.com/dapr/dapr/issues/3991 First step：
Modify the Envelope method to support attributes. The time, dataschema, and subject in the CloudEvent spec are all placed in attributes.

Flollow https://github.com/dapr/components-contrib/pull/1369： Fix DCO

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


yaron2:
cc @artursouza

daixiang0:
@CrazyHZM ping

berndverst:
@CrazyHZM I am currently exploring this from a different perspective:

I want to maintain additional optional cloud event fields. The Cloud Events SDKs turn these optional fields into HTTP headers (or gRPC metadata).

The problem is that the Dapr PubSub implementations do not store these headers anywhere. So every Dapr Pubsub implementation needs to be updated to store this data somewhere.

We can either do this in the runtime:

```
type CompleteEvent struct {
  event cloudevent
  headers map[string]string
}
```

this could be serialized to bytes and then used as the `data` value that is stored in all the pubsub implementations.

An alternative approach is for each PubSub implementation to handle storing of the custom attributes / metadata separately.

In the subscribe function however we would need to take this event and separate the headers from the cloud event again (or retrieve the metadata attributes manually), then in the Dapr API we need to return the additional attributes via headers or gRPC metadata. For example the Cloud Events Python SDK wouldn't deserialize custom cloud events attributes natively and hence we must read this via the headers / gRPC metadata.

CrazyHZM:
> @CrazyHZM I am currently exploring this from a different perspective:
> 
> I want to maintain additional optional cloud event fields. The Cloud Events SDKs turn these optional fields into HTTP headers (or gRPC metadata).
> 
> The problem is that the Dapr PubSub implementations do not store these headers anywhere. So every Dapr Pubsub implementation needs to be updated to store this data somewhere.
> 
> We can either do this in the runtime:
> 
> ```
> type CompleteEvent struct {
>   event cloudevent
>   headers map[string]string
> }
> ```
> 
> this could be serialized to bytes and then used as the `data` value that is stored in all the pubsub implementations.
> 
> An alternative approach is for each PubSub implementation to handle storing of the custom attributes / metadata separately.
> 
> In the subscribe function however we would need to take this event and separate the headers from the cloud event again (or retrieve the metadata attributes manually), then in the Dapr API we need to return the additional attributes via headers or gRPC metadata. For example the Cloud Events Python SDK wouldn't deserialize custom cloud events attributes natively and hence we must read this via the headers / gRPC metadata.

I‘m so sorry for replying so late, I think the first way is good with limited headers. I've been busy with other things recently. If you want, you can lead this issue, and I can participate in the discussion from the side.

berndverst:
> > @CrazyHZM I am currently exploring this from a different perspective:
> > I want to maintain additional optional cloud event fields. The Cloud Events SDKs turn these optional fields into HTTP headers (or gRPC metadata).
> > The problem is that the Dapr PubSub implementations do not store these headers anywhere. So every Dapr Pubsub implementation needs to be updated to store this data somewhere.
> > We can either do this in the runtime:
> > ```
> > type CompleteEvent struct {
> >   event cloudevent
> >   headers map[string]string
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > this could be serialized to bytes and then used as the `data` value that is stored in all the pubsub implementations.
> > An alternative approach is for each PubSub implementation to handle storing of the custom attributes / metadata separately.
> > In the subscribe function however we would need to take this event and separate the headers from the cloud event again (or retrieve the metadata attributes manually), then in the Dapr API we need to return the additional attributes via headers or gRPC metadata. For example the Cloud Events Python SDK wouldn't deserialize custom cloud events attributes natively and hence we must read this via the headers / gRPC metadata.
> 
> I‘m so sorry for replying so late, I think the first way is good with limited headers. I've been busy with other things recently. If you want, you can lead this issue, and I can participate in the discussion from the side. @berndverst

Sounds good @CrazyHZM. I have a draft PR here -- https://github.com/dapr/dapr/pull/4873

Essentially all values that aren't well known cloud event attributes will be maintained as Attributes (which maps to Cloud Event Extensions). In the SDKs we need to map Attributes from the proto to the Extensions attribute of the CloudEvent SDK -- I have a draft for the Python SDK for this also. It's very straightforward.

I think this will be easiest way to handle this.

I tried to follow the Cloud Event spec as closely as possible.

berndverst:
I will close this PR for now because I think the work actually needs to be done in dapr/dapr anyway and my PR will likely accomplish this goal. Feel free to comment on that PR instead.

Issue #1464, opened 2022-01-26
Support JetStream wildcard subscription


Gehhilfe:
## Describe the feature

NATS JetStream allows wildcard subscription e.g. topic.* matching topic.a or topic.> matching topic.a.b.c.d . Therefore streams can have multiple different topics when wildcards are used.

Subscribing with wildcards is working, but received messages cant be routed back to application, because the subscribed topic is different to received topic of the message.

Therefore I propose following changes:

- Message received on a wildcard subscription should be send back with the wildcard topic, than the message can be route as usually
- The actual topic is moved into message metadata and can be used by the application

**This should not break existing application, because non wildcard subscription behavior doesn't change!**

## Release Note

RELEASE NOTE: **ADD** NATS JetStream PubSub supports now wildcard subscriptions

PR #1465, opened 2022-01-26
Add support for nats jetstream wildcard subscriptions

Gehhilfe:
Signed-off-by: Tim Burkert <burkert.tim@gmail.com>

# Description

Currently, messages received on jetstream wildcard subscriptions can't be routed back to the app, because the subscribed topic is not matching to the received topic.

Used the subscribed topic to return the message to dapr from nats component and moved to actual topic metadata. This should not break any applications, because it only adds a metadata field and doesn't change behavior for non wildcard subscriptions.

## Issue reference

Please reference the issue this PR will close: #1464 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


pubsub/jetstream/jetstream.go
```diff
@@ -132,8 +132,11 @@ func (js *jetstreamPubSub) Subscribe(req pubsub.SubscribeRequest, handler pubsub
 			js.l.Debugf("Processing JetStream message %s/%d", m.Subject,
 				jsm.Sequence)
 			opErr := handler(js.ctx, &pubsub.NewMessage{
-				Topic: m.Subject,
+				Topic: req.Topic,
```
mukundansundar, line 135: Why's this. being changed to `req.Topic` but the message field is used in the below line?
should it also be changed in the previous line?
Quick question is the message subject generally different from topic?

mukundansundar: nvm saw the issue description ...

yaron2 (approved):
LGTM

Thanks @Gehhilfe!

PR #1482, opened 2022-02-08
Handle errors from parallel mode

gdlcf88:
It seems that errors in the parallel mode cannot be caught:
https://github.com/dapr/components-contrib/blob/v1.6.0/pubsub/rabbitmq/rabbitmq.go#L423

Based on https://github.com/dapr/components-contrib/pull/576


daixiang0:
Please update tests to cover this case.

gdlcf88:
Hi @daixiang0, I have updated the tests. https://github.com/dapr/components-contrib/pull/1482/commits/8cb1052e6b19d2e6aab5ee5b22ca9976fae743b9

Taction:
@gdlcf88 Please fix linter issues.(You can run `gofumpt -e -d pubsub/rabbitmq/rabbitmq_test.go` to see details)

yaron2:
This is very useful PR @gdlcf88. Thanks!

@daixiang0 can you please review this?

yaron2:
ping @daixiang0

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -100,6 +100,27 @@ type rabbitMQConnectionBroker interface {
...
 
+type ErrorCollection struct {
+	errors []error
+	mux    sync.Mutex
+}
+
+func (c *ErrorCollection) Append(e error, stopCh chan struct{}) {
+	c.mux.Lock()
+	if len(c.errors) == 0 {
+		stopCh <- struct{}{}
```
Taction, line 111: Reading from a closed channel is safe and unblocked, so `stopCh <- struct{}{}` can be safely deleted.

Taction:
@gdlcf88 Thanks for your contribution! LGTM overall, just a nit change.

gdlcf88:
@Taction - Thanks for reviewing. 😊

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -409,32 +432,49 @@ func (r *rabbitMQ) subscribeForever(req pubsub.SubscribeRequest, queueName strin
...
+				return ec.GetErrors()
+			case d := <-msgs:
+				wg.Add(1)
+				go func(channel rabbitMQChannelBroker, d amqp.Delivery, topic string, handler pubsub.Handler) {
+					if e := r.handleMessage(channel, d, topic, handler); e != nil && mustReconnect(channel, []error{e}) {
+						ec.Append(e, stopCh)
+					}
+					wg.Done()
+				}(channel, d, topic, handler)
+			}
```
Taction, line 477: Also, I'd more like make `stopCh` as a field of `ErrorCollection` to make the code cleaner. And the param of `Append` can be an error only.

```
case pubsub.Parallel:
		ec := ErrorCollection{errNotify: make(chan struct{})}
		var wg sync.WaitGroup
		for {
			select {
			case <-ec.errNotify:
				wg.Wait()
				return ec.GetErrors()
			case d := <-msgs:
				wg.Add(1)
				go func(channel rabbitMQChannelBroker, d amqp.Delivery, topic string, handler pubsub.Handler) {
					if e := r.handleMessage(channel, d, topic, handler); e != nil && mustReconnect(channel, []error{e}) {
						ec.Append(e)
					}
					wg.Done()
				}(channel, d, topic, handler)
			}
		}
	}
```
And for ErrorCollection definition:
```
type ErrorCollection struct {
	errors    []error
	mux       sync.Mutex
	errNotify chan struct{}
}

func (c *ErrorCollection) Append(e error) {
	c.mux.Lock()
	if len(c.errors) == 0 {
		close(c.errNotify)
	}
	c.errors = append(c.errors, e)
	c.mux.Unlock()
}
```

gdlcf88: @Taction - Done. And I am trying to make the ErrorCollection shared since we may use it in other packages. But I am not sure whether the sharing way is a good practice in the Golang. https://github.com/gdlcf88/components-contrib/commit/2a64a565beae7d90b58e3af045505e7cc857f33b

Taction: This is a good idea, but since this is a small change and there have no other packages using it for now, I would rather refactor the code when anywhere else uses it.

Taction:
The pubsub.rabbitmq certification test is failed all the time, can you please look into this?

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -101,14 +101,22 @@ type rabbitMQConnectionBroker interface {
...
+	ErrNotify chan struct{}
 }
 
-func (c *ErrorCollection) Append(e error, stopCh chan struct{}) {
+func NewErrorCollection() ErrorCollection {
+	return ErrorCollection{
+		errors:    []error{},
+		ErrNotify: make(chan struct{}),
+	}
+}
```
Taction, line 114: You should return a `*ErrorCollection` here since your struct has a `sync.Mutex` field.

```
func NewErrorCollection() *ErrorCollection {
	return &ErrorCollection{
		errors:    []error{},
		ErrNotify: make(chan struct{}),
	}
}
```

gdlcf88:
> The pubsub.rabbitmq certification test is failed all the time, can you please look into this?

Log
```
2022-05-21T07:06:26.6486731Z [36;1m  echo "::error:: No certification test file was found for component pubsub.rabbitmq"[0m
```

certification.yml
```yaml
        # Fail the step if we found no test to run
        if grep -q "\[no test files\]" ${{ env.TEST_OUTPUT_FILE_PREFIX }}_certification.json ; then
          echo "::error:: No certification test file was found for component ${{ matrix.component }}"
          exit -1
        fi
```

@Taction - Sorry, I am not sure what's wrong with this PR. Could you please help?

yaron2:
@gdlcf88 please resolve conflicts.

PR #1483, opened 2022-02-08
Fix: unsubscribe by id

LaurenceLiZhixin:
Signed-off-by: LaurenceLiZhixin <382673304@qq.com>

# Description

I changed the unsubscribe logic from key level to subscription level, using unique subscribe-id to locate old subscripition.

Please reference the issue this PR will close: #3906

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


configuration/redis/redis.go
```diff
@@ -258,17 +259,14 @@ func (r *ConfigurationStore) Get(ctx context.Context, req *configuration.GetRequ
 	}, nil
 }
 
-func (r *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.SubscribeRequest, handler configuration.UpdateHandler) error {
+func (r *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.SubscribeRequest, handler configuration.UpdateHandler) (string, error) {
+	subscribeID, _ := uuid.GenerateUUID()
```
daixiang0, line 263: Could we ignore it safely?

go.mod
```diff
@@ -161,7 +161,10 @@ require (
 	github.com/sijms/go-ora/v2 v2.2.22
 )
 
-require github.com/nats-io/nkeys v0.3.0
+require (
+	github.com/hashicorp/go-uuid v1.0.1
```
daixiang0, line 165: Could we use `github.com/google/uuid` to reduce deps?

yaron2:
ping @LaurenceLiZhixin

go.mod
LaurenceLiZhixin: fine

yaron2:
@daixiang0 please re-review.

configuration/requests.go
```diff
@@ -33,12 +33,13 @@ type SubscribeRequest struct {
 	Metadata map[string]string `json:"metadata"`
 }
 
-// UnSubscribeRequest is the object describing a request to unsubscribe configuration.
-type UnSubscribeRequest struct {
-	Keys []string `json:"keys"`
```
daixiang0, line 38: Does it means we can not use one API to unsubscribe many config?

LaurenceLiZhixin: One unsubscribe call is only for stopping only on subscription created before. The previous subscription may contains many keys, which means that we can unsubscribe many config once. @daixiang0

daixiang0: So it is a breaking change for this API?

LaurenceLiZhixin: Yes, redis is the first impl of this API. @LaurenceLiZhixin

daixiang0 (approved):

Issue #1211, opened 2021-10-20
Add Query StateStore API implementation for Redis StateStore


dmitsh:
## Describe the feature
Per https://github.com/dapr/dapr/issues/3662, we have added Query StateStore API and implemented it for ModgoDB and CosmosDB (https://github.com/dapr/components-contrib/pull/1153)

We need to implement this API for Redis StateStore

## Release Note

RELEASE NOTE:

yaron2:
Reopened.

yaron2:
@dmitsh are you planning to work on this for 1.6?

dmitsh:
> @dmitsh are you planning to work on this for 1.6?

wasn't planning. @manojrupireddy, are you still interested in pursuing this task?

dmitsh:
Natively, Redis only offers primary key access. However, with the help of two modules, it is possible to store and query JSON documents:

- RedisJSON allows to store and manipulate values as JSON objects (https://oss.redis.com/redisjson/)
- RediSearch enables secondary indexing and searching capability (https://oss.redis.com/redisearch/).

The current Redis component implementation does not require any additional modules, and stores the values as raw data.

The use of the aforementioned modules allows querying the state, but it comes with a price:
1. RediSearch requires setting the JSON object schema prior to querying
2. RedisJSON stores data in a special fashion.

This will be incompatible with the current implementation. One way to address this issue is to develop an alternative Redis component that supports JSON structures. The new component will require presence of RedisJSON and RediSearch, and will allow setting the object schemas.

To make the new Redis component to support other formats, we would need to introduce optional content type to the State Store API. If the content type equals to `application/json`, the component will store the value as JSON object. Otherwise the value will be stored as raw data.

In the future, the component could support additional formats.

This is a very high-level description of the proposed design.
Please comment if you have any comments and suggestions.
After gathering the feedback, I'm intended to propose more detailed design.

artursouza:
This starts to feel like a different implementation of the redis state store component.

dmitsh:
> This starts to feel like a different implementation of the redis state store component.

it can be the same component, but there should be an agreement that if JSON content type is declared, the underlying Redis service should include RedisJSON and RediSearch modules.

pkedy:
[...]

PR #1488, opened 2022-02-08
redis statestore: implement state query API

dmitsh:
# Description

implement state query API for Redis statestore

## Issue reference

Please reference the issue this PR will close: #1211

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#2187

### DIFF
--- a/state/redis/redis.go
+++ b/state/redis/redis.go
@@ -15,7 +15,6 @@ package redis
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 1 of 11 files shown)


state/redis/redis.go
```diff
@@ -25,20 +24,69 @@ import (
...
+	  return error("failed to delete " .. KEYS[1])
+	end`
+	setJSONQuery = `
+	local etag = redis.pcall("JSON.GET", KEYS[1], ".version");
+	if type(etag) == "table" then
+	  redis.call("JSON.DEL", KEYS[1]);
+	end;
+	if not etag or type(etag)=="table" or etag == "" then
+	  etag = ARGV[1];
+	end;
```
mukundansundar, line 66: why is there any addition of these lines? Based on comparing this to the `setDefaultQuery`, both perform the same logic but slightly different conditions are there. Why is the `etag` being overwritten here?

state/redis/redis.go
```diff
@@ -25,20 +24,69 @@ import (
...
+	local fwr = redis.pcall("JSON.GET", KEYS[1], ".first-write");
+	if etag == ARGV[1] or ((not fwr or type(fwr) == "table") and ARGV[1] == "0") then
```
mukundansundar, line 68: Why is there a check of `type(fwr) == "table"` here but not in line 43 in the `setDefaultQuery`?

state/redis/redis.go
```diff
@@ -25,20 +24,69 @@ import (
...
+	  redis.call("JSON.SET", KEYS[1], "$", ARGV[2]);
```
mukundansundar, line 69: Why is the 3rd param `$` instead of `data`?

state/redis/redis.go
```diff
@@ -25,20 +24,69 @@ import (
...
+	  if ARGV[3] == "0" then
+	    redis.call("JSON.SET", KEYS[1], ".first-write", 0);
+	  end;
+	  return redis.call("JSON.SET", KEYS[1], ".version", (etag+1))
```
mukundansundar, line 73: Apparently we always start with version 1 in the setDefaultQuery, but here it is `ARGV[1] + 1`? Why is it different that way?

dmitsh: same as before,  json module does not support `HINCRBY`, so we need to mimic it.

state/redis/redis.go
dmitsh: because json module does not support `HINCRBY`, so we need to mimic it.

state/redis/redis.go
dmitsh: that's how json module works. `$` writes down the whole object. We cannot update a field without having an object.

tests/conformance/state/state.go
```diff
@@ -223,10 +229,14 @@ func ConformanceTests(t *testing.T, props map[string]string, statestore state.St
 			for _, scenario := range scenarios {
 				if !scenario.bulkOnly && !scenario.transactionOnly {
 					t.Logf("Setting value for %s", scenario.key)
-					err := statestore.Set(&state.SetRequest{
+					req := &state.SetRequest{
 						Key:   scenario.key,
 						Value: scenario.value,
-					})
+					}
+					if config.ComponentName == componentRedis && len(scenario.contentType) != 0 {
```
artursouza, line 236: Please, keep the if clause but remove the condition on component name. Since contentType is optional, it will just be ignored by other state stores. The goal is to keep the conformance tests as independent as possible to component specifics. In this case, passing an optional param on all is OK.

artursouza: We might need to ave certification tests for Redis now so we don't pollute more of the conf tests with component specific scenarios.

state/redis/redis.go
```diff
@@ -241,6 +311,53 @@ func (r *StateStore) Get(req *state.GetRequest) (*state.GetResponse, error) {
 	}, nil
 }
 
+func (r *StateStore) getJSON(req *state.GetRequest) (*state.GetResponse, error) {
+	res, err := r.client.Do(r.ctx, "JSON.GET", req.Key).Result()
+	if err != nil {
+		return nil, err
+	}
+
+	str, ok := res.(string)
```
mukundansundar, line 320: similar to line 295 should there be a nil check for `res`? 
I see that it is being typecasted to string, and that should return an empty string for nil values, but the return value from this function is different than what it is from the getDefault function ...

tests/conformance/state/state.go
dmitsh: I will create an issue to update certification tests

artursouza (approved):

Issue #1489, opened 2022-02-09
[DynamoDB] Support for time to live (state store)


wimvdc:
## Intro

The state store implementation for DynamoDB currently lacks time to live support.

## Proposal:

- As described in the spec, fetch `ttlInSeconds`from the metadata of a request (if present)
- Since DynamoDB expects the value to be "_[...] value must be a timestamp in [Unix epoch time format](https://en.wikipedia.org/wiki/Unix_time) in seconds [...]_", we add that value to the current epoch in seconds.
- When ttlInSeconds = -1, we don't set the value in DynamoDB 
- Hardcoded field name in DynamoDB: `expireAt` 

## References:
- DynamoDB TTL general - https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/TTL.html
- DynomoDB TTL details - https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/time-to-live-ttl-before-you-start.html

Note: I would like to pick up this proposal if it gets a _go_

yaron2:
Hi @wimvdc, this will be a most welcome contribution.

Assigned to you and triaged into the next milestone.

PR #1493, opened 2022-02-10
[DynamoDB] StateStore - Add TTL from metadata to Set and BulkSet

wimvdc:
# Description

Adding ttlInSeconds support to DynamoDB as state store (Set and SetBulk)
This is my first ever experience with Go, so please let me know how to improve the new code.

## Issue reference

#1489

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


state/aws/dynamodb/dynamodb.go
```diff
@@ -250,3 +293,24 @@ func (d *StateStore) marshalToString(v interface{}) (string, error) {
...
+		parsedVal, err := strconv.ParseInt(val, 10, 0)
+		if err != nil {
+			return nil, err
+		}
+		parsedInt := int64(parsedVal)
+		//To follow the spec
+		if parsedInt == -1 {
+			return nil, nil
+		}
+		//DynmoDB expects a epoch timestamp in seconds
```
berndverst, line 309:
```suggestion
		//DynamoDB expects an epoch timestamp in seconds
```

state/aws/dynamodb/dynamodb.go
```diff
@@ -250,3 +293,24 @@ func (d *StateStore) marshalToString(v interface{}) (string, error) {
...
+		experationTime := time.Now().Unix() + parsedInt
```
berndverst, line 310:
```suggestion
		expirationTime := time.Now().Unix() + parsedInt
```

state/aws/dynamodb/dynamodb.go
```diff
@@ -250,3 +293,24 @@ func (d *StateStore) marshalToString(v interface{}) (string, error) {
...
+
+		return &experationTime, nil
```
berndverst, line 312:
```suggestion
		return &expirationTime, nil
```

state/aws/dynamodb/dynamodb_test.go
```diff
@@ -232,6 +313,45 @@ func TestSet(t *testing.T) {
 		err := ss.Set(req)
 		assert.NotNil(t, err)
 	})
+	t.Run("Un-successfully set item with ttl (invalid value)", func(t *testing.T) {
```
berndverst, line 316: Can you please update every occurrence of `un-successfully` to be `unsuccessfully`?

berndverst (changes requested):

berndverst:
Also please fix linter errors!

state/aws/dynamodb/dynamodb.go
```diff
@@ -250,3 +293,24 @@ func (d *StateStore) marshalToString(v interface{}) (string, error) {
...
+//Parse and process ttlInSeconds
+func (d *StateStore) parseTTL(req *state.SetRequest) (*int64, error) {
+	if val, ok := req.Metadata["ttlInSeconds"]; ok && val != "" {
+		parsedVal, err := strconv.ParseInt(val, 10, 0)
+		if err != nil {
+			return nil, err
+		}
+		parsedInt := int64(parsedVal)
+		//To follow the spec
+		if parsedInt == -1 {
```
berndverst, line 306: `-1` was intended to be used to explicitly persist a record, for example in the case where the state store supports a default TTL that is applied to every new entity.

It does not seem that DynamoDB supports such a concept. Perhaps you should return for every negative value?

berndverst:
From my understanding of the Dynamo DB documentation the attribute used for TTL is configurable when TTL is enabled on a table. https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/time-to-live-ttl-before-you-start.html

However this PR here uses a hardcoded `expireAt`.

I think this should be configurable via the component Metadata, similar to how the table name is being specified today:
https://github.com/dapr/components-contrib/blob/master/state/aws/dynamodb/dynamodb.go#L42

Can you please update the PR accordingly? Also make sure to create a follow up PR against the Dapr Docs to document the TTL feature, and document this component metadata field to specify the TTL attribute (and that someone must manually enable the TTL feature and for the Dynamo DB table and specify the attribute name)

Thanks!

berndverst:
Another thought:
The Dynamo DB documentation suggests that entities that have expired are still returned for 48 hours or so.

Should the Get & Bulk Get operations examine the current expiryTime and filter out those entities if that expiryTime is in the past?

wimvdc:
> From my understanding of the Dynamo DB documentation the attribute used for TTL is configurable when TTL is enabled on a table. https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/time-to-live-ttl-before-you-start.html or https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/time-to-live-ttl-how-to.html
> 
> However this PR here uses a hardcoded `expireAt`. I think this should be configurable via the component Metadata, similar to how the table name is being specified today: https://github.com/dapr/components-contrib/blob/master/state/aws/dynamodb/dynamodb.go#L42
> 

I have changed it so that the component metadata needs to have `ttlAttributeName` specified in order to make use of TTL.

> And of course this component metadata field (as opposed to the state store request metadata) should be optional. If the field is not provided, then the TTL feature cannot function, so any state store request TTL values should be ignored.
> 
I added an extra if in the `parseTTL` function

> Can you please update the PR accordingly? Also make sure to create a follow up PR against the Dapr Docs to document the TTL feature, and document this component metadata field to specify the TTL attribute (and that someone must manually enable the TTL feature and for the Dynamo DB table and specify the attribute name)
> 
> This is the page that needs to be updated with the optional metadata field and more details: https://docs.dapr.io/reference/components-reference/supported-state-stores/setup-dynamodb/ And you'll want to update the following table to check the TTL column: https://docs.dapr.io/reference/components-reference/supported-state-stores/
> 
> Thanks!

I'm not sure what the order should be, wait until this PR has been merged? Or in parallel? 

> Another thought: The Dynamo DB documentation suggests that entities that have expired are still returned for 48 hours or so.
> 
> Should the Get & Bulk Get operations examine the current expiryTime and filter out those entities if that expiryTime is in the past?
That's a hard one. 

On the one hand you have chosen to use DynamoDB in favor of another technology/service as state store so I might be strange to temper with the functionalities that DynamoDB offers. 

On the other hand developers might not be aware of this behaviour, especially if they try to develop cloud agnostic. 

Personally I would not implement the extra checks in (bulk) Get. But it might be useful to return the TTL field in (bulk) Get so that a developer can choose to do something with it ?

state/aws/dynamodb/dynamodb.go
wimvdc: What I tried to archive;
Imagine you create the following item:
```
key: dummyKey
value: someValue
ttl: 90
```

And 20 seconds later you do:
```
key: dummyKey
value: someValue
```

Will this override the existing TTL field so that DynamoDB won't delete it after 70 seconds?

berndverst: That depends on how Dynamo DB handles this (which I personally don't know as I am not a user of DynamoDB). In general I think that should indeed reset the TTL to whatever the default is (no TTL?).

If there is a default TTL (which I think Dynamo DB doesn't have) you might need to specify explicitly that you want to ignore the default TTL and persist the item. That's what `-1` was intended to be for.

I think you don't want have to worry about that -- I can't seem to find anything about default TTLs / globally defined TTLs in Dynamo DB, so it probably doesn't apply here.

berndverst:
image

berndverst:
@wimvdc discussion here:
https://discord.com/channels/778680217417809931/781589820128493598/942871929306636428

@yaron2 and I agree that BulkGet and Get should filter out all records with expired TTL in the DynamoDB state store component.

Thanks!

wimvdc:
> > > Also please fix linter errors!
> > 
> > 
> > Should be fixed now
> > > From my understanding of the Dynamo DB documentation the attribute used for TTL is configurable when TTL is enabled on a table. https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/time-to-live-ttl-before-you-start.html or https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/time-to-live-ttl-how-to.html
> > > However this PR here uses a hardcoded `expireAt`. I think this should be configurable via the component Metadata, similar to how the table name is being specified today: https://github.com/dapr/components-contrib/blob/master/state/aws/dynamodb/dynamodb.go#L42
> > 
> > 
> > I have changed it so that the component metadata needs to have `ttlAttributeName` specified in order to make use of TTL.
> > > And of course this component metadata field (as opposed to the state store request metadata) should be optional. If the field is not provided, then the TTL feature cannot function, so any state store request TTL values should be ignored.
> > 
> > 
> > I added an extra if in the `parseTTL` function
> > > Can you please update the PR accordingly? Also make sure to create a follow up PR against the Dapr Docs to document the TTL feature, and document this component metadata field to specify the TTL attribute (and that someone must manually enable the TTL feature and for the Dynamo DB table and specify the attribute name)
> > > This is the page that needs to be updated with the optional metadata field and more details: https://docs.dapr.io/reference/components-reference/supported-state-stores/setup-dynamodb/ And you'll want to update the following table to check the TTL column: https://docs.dapr.io/reference/components-reference/supported-state-stores/
> > > Thanks!
> > 
> > 
> > I'm not sure what the order should be, wait until this PR has been merged? Or in parallel?
> > > Another thought: The Dynamo DB documentation suggests that entities that have expired are still returned for 48 hours or so.
> > > Should the Get & Bulk Get operations examine the current expiryTime and filter out those entities if that expiryTime is in the past?
> > > That's a hard one.
> > 
> > 
> > On the one hand you have chosen to use DynamoDB in favor of another technology/service as state store so I might be strange to temper with the functionalities that DynamoDB offers.
> > On the other hand developers might not be aware of this behaviour, especially if they try to develop cloud agnostic.
> > Personally I would not implement the extra checks in (bulk) Get. But it might be useful to return the TTL field in (bulk) Get so that a developer can choose to do something with it ?
> 
> @wimvdc in general when you look at Dapr components you'll find that the state store components are the fully managed components that hide a lot of details from you. Whenever you want more low level control you use the equivalent bindings.
> 
> My vote would be to filter out the expired keys and to perform these extra checks.
> 
> If you want this complete DynamoDB control I would ensure (in a separate PR and it doesn't have to be done by you) that the Dynamo DB binding instead has a way to return all entities including the TTL state. https://docs.dapr.io/reference/components-reference/supported-bindings/dynamodb/
> 
> What do you think?



> @wimvdc discussion here: https://discord.com/channels/778680217417809931/781589820128493598/942871929306636428
> 
> @yaron2 and I agree that BulkGet and Get should filter out all records with expired TTL in the DynamoDB state store component.
> 
> Thanks!

Sounds reasonable, will work on the filter

state/aws/dynamodb/dynamodb.go
wimvdc: Ok, I removed the IF statement and modified the tests a bit.

wimvdc:
I'm currently stuck with DCO-stuff.
I need to signoff some old commits but I can't use rebase because that would override merged commits.
Any help?

daixiang0:
I think you can rebase main branch and force push to fix DCO.

wimvdc:
I really messed up the git history.
Closing this in favor of a new PR with correct commits, sorry for my unorthodox way of working.
New PR #1512

Issue #1210, opened 2021-10-20
New conf test scenario for state stores: transaction order


artursouza:
## Describe the proposal

As per #1209, CosmosDb implementation does not respect the order of transactions. This was caught as we changed our E2E tests on AKS to use CosmosDB and one of the E2E tests covered this scenario while our conformance tests don't.

This proposal is to add the following scenario to conformance tests for state stores and fix the state stores that do not pass this test:

1. Invoke Transaction API with the following operations in order:
```yaml
  upsert key1 value1
  delete key1
  upsert key2 value2
  delete key3
  upsert key3 value3
```
2. Validate key1 is not present.
3. Validate key2 is present.
4. Validate key3 is present.

shubham1172:
Per the conformance tests run in https://github.com/dapr/components-contrib/runs/5177659685, the following state stores are failing too:
1. SQL Server - fails with the error message `delete affected only 0 rows, expected 1` because deletes are attempted before upserts in the code, no matter what.
2. Postgresql - deletion always happens before upserts and the transaction fail.
3. MySQL - same as Postgresql

shubham1172:
@artursouza, do you think the fixes should be a part of the same PR? Or should we track them separately?

Issue #1209, opened 2021-10-20
CosmosDB state stores changes order of operations in transaction


artursouza:
## Expected Behavior

When an app performs the following transaction in state store:
```txt
upsert key1 value1
delete key1
upsert key2 value2
```

The expected result in the database is that key1 is not present and key2 is present, respecting the order of the operations.

## Actual Behavior

The result in the database is that key1 and key2 are present, not respecting the order of the operations.

## Steps to Reproduce the Problem

Perform a transaction operation in CosmosDB state store.

## Release Note

RELEASE NOTE: **FIX** CosmosDb transaction API to respect order of operations

artursouza:
This requires the stored procedure to be rewritten.

dillenmeister:
1) If state operations with the same key appears more than once, isn't it always only the last operation that will matter? Would this be a solution instead of respecting the order, or are there use cases where you want to update the same state twice? The suggested conformance test (#1210) would pass if only the last operation per key was executed.

```
upsert key1 value1 // skip
delete key1
upsert key2 value2
delete key3 // skip
upsert key3 value3
```

2) Can two operations with the same key ever be in the same transaction if etags are included (first write concurrency mode)?
Wouldn't this fail on the second operation?  (note related issue #1244)
```
upsert key1 value1 etag1
delete key1 etag1 // fails because state with key1 has a new etag
upsert key2 value2 etag2
```

shubham1172:
I was able to reproduce the issue with the following snippet. On every alternate run, I see that both k1 and k2 are present (problematic), and the other times only k2 is present (as expected).

```js
/**
 * To reproduce #1209, perform the following in a transaction:
 * upsert k1, v1
 * delete k1
 * upsert k2, v2
 * 
 * Expected: k1 is not present, k2 is present with value v2
 * Actual: k1 and k2 are present with values v1 and v2
 * 
 * @param client DaprClient
 */
async function performOperations(client: DaprClient) {
    var operations = [
        {
            operation: "upsert",
            request: {
                key: "key1",
                value: {
                    part: "partition",
                    zipcode: 100
                }
            }
        },
        {
            operation: "delete",
            request: {
                key: "key1"
            }
        },
        {
            operation: "upsert",
            request: {
                key: "key2",
                value: {
                    part: "partition",
                    zipcode: 200
                }
            }
        }
    ];

    var metadata = {
        partitionKey: "partition"
    }

[...]

PR #1496, opened 2022-02-11
[WIP] Transaction order conformance test for state stores

shubham1172:
# Description

Introduces a new test that checks the order of transactions. Say a transaction that contains the following:
```bash
upsert k1 v1
delete k1
upsert k2 v2
```
The order of operations must be respected, and `k1` should be deleted, whilst `k2` should have a value of `v2`.

Note, an anomaly was found in Cosmos DB state store in #1209, and it was also found that the transaction run fails to maintain the order in alternate runs.

## Issue reference

Please reference the issue this PR will close: #1210

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_ (NA)

### DIFF
--- a/tests/conformance/state/state.go
+++ b/tests/conformance/state/state.go
@@ -433,6 +433,62 @@ func ConformanceTests(t *testing.T, props map[string]string, statestore state.St
 				}
 			}
 		})
+
+		t.Run("transaction-order", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


shubham1172:
Cosmos DB fails the new test:

image

PR #1498, opened 2022-02-11
Postgres state store: Fix `Set` with ETag

### LINKED ISSUE #1497: Postgres State Store: `cannot convert to pguint32` error when updating with ETag
## Expected Behavior

I should be able to update a key using ETag.

## Actual Behavior

A conversion error is returned...

```
cannot convert 956 to pguint32
```

where `956` is the ETag value.

Calling `Set` on an existing item with an ETag should reproduce this.

I discovered this by using the Postgres state store as my Actor state store (config below) and invoking CreateReminder. The first call works, but successive calls fail (and block the actor due to retry code in the actor system).

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: statestore
spec:
  type: state.postgresql
  version: v1
  metadata:
  - name: connectionString
    value: host=XXX user=XXX password=XXX dbname=XXX port=5432 sslmode=disable
  - name: actorStateStore
    value: 'true'
```

## Release Note

RELEASE NOTE: **FIX** Setting a value using ETag for the Postgres state store

pkedy:
Postgres state: Convert etag to uint32 instead of int to fix conversion error

# Description

I determined that Posgres (or the client driver used) requires the argument to be passed in as an `uint32` so that it can convert it to a `pguint32`/`XID`. This PR makes a simple parsing/casting tweak to pass the expected type.

## Issue reference

Closes #1497

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


state/postgresql/postgresdbaccess.go
```diff
@@ -131,12 +131,13 @@ func (p *postgresDBAccess) setValue(req *state.SetRequest) error {
 			ON CONFLICT (key) DO UPDATE SET value = $2, isbinary = $3, updatedate = NOW();`,
 			tableName), req.Key, value, isBinary)
 	} else {
-		// Convert req.ETag to integer for postgres compatibility
-		var etag int
-		etag, err = strconv.Atoi(*req.ETag)
+		// Convert req.ETag to uint32 for postgres XID compatibility
+		var etag64 uint64
```
halspang, line 135: You reference uint32 in the comment, but that's a uint64. Does it matter?

halspang (commented):
This seems like something that would get hit a lot if it was breaking the ETag. There is a conformance test for etags and postgres is included in the list of tested stores. Why didn't that test catch this? Can we make it?

state/postgresql/postgresdbaccess.go
pkedy: `uint64` is used because that is what is returned by `strconv.ParseUint` so the `64` suffix was to make that explicit.

pkedy:
I'm going to try updating the conformance test to reproduce this.

pkedy:
This issues stems from the fact that I was embedding Dapr and using a newer version of `github.com/jackc/pgx/v4` (specifically `v4.15.0`). Once I updated the version locally I began to see the error in the conformance test.

```
go test -v -tags=conftests -count=1 ./tests/conformance -run="TestStateConformance/postgresql"
```

```
    state.go:513:
        	Error Trace:	state.go:513
        	Error:      	Expected nil, but got: &errors.errorString{s:"cannot convert 1279 to pguint32"}
        	Test:       	TestStateConformance/postgresql/etag
--- FAIL: TestStateConformance (0.32s)
    --- FAIL: TestStateConformance/postgresql (0.21s)
        --- PASS: TestStateConformance/postgresql/init (0.03s)
        --- PASS: TestStateConformance/postgresql/ping (0.00s)
        --- PASS: TestStateConformance/postgresql/set (0.02s)
        --- PASS: TestStateConformance/postgresql/get (0.01s)
        --- PASS: TestStateConformance/postgresql/query (0.00s)
        --- PASS: TestStateConformance/postgresql/delete (0.01s)
        --- PASS: TestStateConformance/postgresql/bulkset (0.04s)
        --- PASS: TestStateConformance/postgresql/bulkdelete (0.01s)
        --- PASS: TestStateConformance/postgresql/transaction (0.05s)
        --- FAIL: TestStateConformance/postgresql/etag (0.01s)
```

I will update the version of `github.com/jackc/pgx/v4` as well.

halspang (approved):
Thanks for digging into why the conformance tests didn't catch this at first!

berndverst (approved):

PR #1506, opened 2022-02-17
Update Copyright year to 2022

Shoothzj:
@yaron2  Should this be 2022 ?


daixiang0:
I am not sure it is a good choice, changing it each year is grotesque.

Issue #1505, opened 2022-02-17
Azure SQL Server state store changes order of transaction


shubham1172:
Similar to #1209

## Expected Behavior
For the following transaction:
```
upsert key1 value1
delete key1
upsert key2 value2
```
Expectation is that `key1` should be deleted, `key2` should have a value of `value2`.

## Actual Behavior
`key1`:`value1` is present, along with `key2`: `value2`.

## Steps to Reproduce the Problem
Please use the JS snippet from the referenced issue (1209)

## Details
In the transaction snippet for this state store, the delete operations are always executed before upsert operations. This results in failure with message `delete affected only 0 rows, expected 1` whenever the above tx is attempted.

## Release Note

RELEASE NOTE: **FIX** Azure SQL Server state store's transaction API to respect order of operations.

PR #1516, opened 2022-02-21
Fix transaction API for Azure SQL server

shubham1172:

# Description

This PR introduces a logic for only keeping the latest operations per key in a transaction. This is similar to what was proposed in https://github.com/dapr/components-contrib/issues/1209#issuecomment-982731149. For example, say the original transaction is:
```bash
upsert k1 v1  # redundant, gets deleted later
delete k1
delete k2     # redundant, gets set to v2 later
upsert k2 v2
```

Since #1517 was a prerequisite, I have merged the changes with this one. It ignores row mismatch error while using bulk delete in transactions.

## Issue reference

Please reference the issue this PR will close: #1505, #1517

## Checklist

* [X] Code compiles correctly
* (NA) Created/updated tests - will be added as part of #1210

### DIFF
--- a/state/errors.go
+++ b/state/errors.go
@@ -62,3 +62,21 @@ func NewETagError(kind ETagErrorKind, err error) *ETagError {
 		kind: kind,
 	}
 }
+
+// BulkDeleteRowMismatchError represents mismatch in rowcount while deleting rows.
+type BulkDeleteRowMismatchError struct {
+	expected uint64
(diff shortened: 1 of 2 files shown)


shubham1172:
image

yaron2:
This is a needed fix, thanks for the PR.

daixiang0 (approved):

PR #1518, opened 2022-02-21
Fix transaction API for Azure Cosmos DB

shubham1172:

# Description

Similar to #1516, this introduces a logic to fix the ordering in a transaction. For a given key, only the latest operation is executed. The other operations are ignored. (From related PR: ) This is similar to what was proposed in https://github.com/dapr/components-contrib/issues/1209#issuecomment-982731149. For example, say the original transaction is:
```bash
upsert k1 v1  # redundant, gets deleted later
delete k1
delete k2     # redundant, gets set to v2 later
upsert k2 v2
```

## Issue reference

Please reference the issue this PR will close: #1209

## Checklist

* [X] Code compiles correctly
* (NA) Created/updated tests, will be fixed with #1210


shubham1172:
Test from #1496
image

On `master`:
image

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -376,12 +376,23 @@ func (c *StateStore) Multi(request *state.TransactionalStateRequest) error {
...
 
+	keyMap := make(map[string]struct{})
+
 	partitionKey := unknownPartitionKey
 
-	for _, o := range request.Operations {
+	// The order of unique key operations does not matter in an atomic transaction.
+	// Only the latest operation for any unique key is selected for execution.
+	// The other operations are redundant, and hence ignored.
+	for i := len(request.Operations) - 1; i >= 0; i-- {
```
artursouza, line 386: I think we need to create a new version of the stored procedure where it takes a single array of operations at the same order as the input. This was the main reason I decided to track this as an issue vs putting a small fix in.

As @berndverst  mentioned in the call, the new stored procedure version should be under a new name to avoid conflict with older versions.

artursouza (changes requested):

state/azure/cosmosdb/cosmosdb.go
shubham1172: @artursouza, @berndverst, do we need a stored procedure? I understand that we had some custom logic in place earlier, but now it's just a series of operations. Is this not something that can be directly done from the state store code?

shubham1172:
Ran the actors e2e test - **actor_features**, that covers state change scenarios, with changes from this PR.

image

artursouza (approved):

Issue #1521, opened 2022-02-22
MySQL state store changes order of transaction


shubham1172:
Similar to #1209 

## Expected Behavior

For the following transaction:
```
upsert key1 value1
delete key1
upsert key2 value2
```
Expectation is that `key1` should be deleted, `key2` should have a value of `value2`.

## Actual Behavior

Both `key1`:`value1` and `key2`:`value2` are present.

## Steps to Reproduce the Problem
Please use the JS snipped from the reference issue (1209)

## Details
Under the hood, MySQL store attempts all the deletions first, followed by all the upserts. This changes the actual ordering of operations within the transaction, giving wrong results.

## Release Note

RELEASE NOTE: **FIX** MySQL state store's transaction API to respect order of operations.

PR #1523, opened 2022-02-22
Fix ordering in transaction API for MySQL

shubham1172:

# Description

Similar fix to #1516. Only keep the latest operations for each key and ignore the other operations. This ensures that the ordering of the operations is intact.

## Issue reference

Please reference the issue this PR will close: #1521

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


state/mysql/mysql.go
```diff
@@ -523,6 +535,34 @@ func (m *MySQL) Multi(request *state.TransactionalStateRequest) error {
 	return nil
 }
 
+// Returns the set requests.
+func (m *MySQL) getSets(req state.TransactionalStateOperation) (state.SetRequest, error) {
```
daixiang0, line 539: Rename as `getSetRequest`

state/mysql/mysql.go
```diff
@@ -523,6 +535,34 @@ func (m *MySQL) Multi(request *state.TransactionalStateRequest) error {
...
+
+	if setReq.Key == "" {
+		return setReq, fmt.Errorf("missing key in upsert operation")
+	}
+
+	return setReq, nil
+}
+
+// Returns the delete requests.
+func (m *MySQL) getDeletes(req state.TransactionalStateOperation) (state.DeleteRequest, error) {
```
daixiang0, line 553: Rename as `getDeleteRequest`

state/mysql/mysql.go
shubham1172: This is the same convention we have been following for other state stores. Should we keep it for the sake of uniformity, or do you think it's better to update? Same is valid for the `getDeletes` comment.

For example https://github.com/dapr/components-contrib/blob/99df8741ea84ffcb973441efa858a26e8a46dda2/state/sqlserver/sqlserver.go#L357

daixiang0: OK.

state/mysql/mysql.go
```diff
@@ -341,7 +341,28 @@ func (m *MySQL) deleteValue(req *state.DeleteRequest) error {
...
+
+	tx, err := m.db.Begin()
+	if err != nil {
+		return err
+	}
+
+	if len(req) > 0 {
+		for _, d := range req {
+			da := d // Fix for goSec G601: Implicit memory aliasing in for loop.
+			err = m.Delete(&da)
```
daixiang0, line 354: Not sure whether we can use &req[i] here.

shubham1172: You mean, instead of `da := d`? Both of them can fix it IMO but I used this pattern for consistency with our codebase.

shubham1172: /cc @daixiang0

artursouza (approved):

PR #1533, opened 2022-02-24
Add retriable error conditions for Azure Service Bus PubSub message sending

### LINKED ISSUE #1495: Azure ServiceBus pub sub send message. Connection reset by peer
## Expected Behavior

Connection reset by peer result in retry backoff behavior similar to server busy

## Actual Behavior

Instead of re-opening the previous issue #1148 I open this new one as it is only related to publishing

cc: @jjcollinge 

Azure ServiceBus Pub Sub experiences some issues every now and then where the exception is "connection reset by peer"

I've investigated the issue a bit and found the code here
https://github.com/dapr/components-contrib/blob/1cb659c0b8130a64d557a879eecd7aad7d71f41e/pubsub/azure/servicebus/servicebus.go#L353

image

As I understand it what is happening is this:

1. sending the message via the go sdk results in an amqp error
2. the amqp error does not satisfy the condition evaluation as the amqpError.Condition is not server-busy but instead it is empty or internal-error (guessing here)
3. no retry is done

In our application logs we see the following:

```
{
  "Timestamp": "2022-02-11T00:55:02.8162561Z",
  "EventId": 13,
  "LogLevel": "Error",
  "Category": "Microsoft.AspNetCore.Server.Kestrel",
  "Message": "Connection id \u00220HMFCIO6OK4IJ\u0022, Request id \u00220HMFCIO6OK4IJ:00000002\u0022: An unhandled exception was thrown by the application.",
  "Exception": "Dapr.DaprException: Publish operation failed: the Dapr endpoint indicated a failure. See InnerException for details.  ---\u003E Grpc.Core.RpcException: Status(StatusCode=\u0022Internal\u0022, Detail=\u0022error when publish to topic mytopic in pubsub dapr-pubsub-component: read tcp 10.22.242.169:35876-\u003E10.22.2.10:5671: read: connection reset by peer\u0022)    at Dapr.Client.DaprClientGrpc.MakePublishRequest(String pubsubName, String topicName, ByteString content, Dictionary\u00602 metadata, CancellationToken cancellationToken)    --- End of inner exception stack trace ---    at Dapr.Client.DaprClientGrpc.MakePublishRequest(String pubsubName, String topicName, ByteString content, Dictionary\u00602 metadata, CancellationToken cancellationToken)    at IncidentManagementService.Business.Services.WorkitemService\u00601.Update(String Id, Dictionary\u00602 fieldUpdates, Boolean skipDataValidation) in .... abrev",
  "State": {
    "Message": "Connection id \u00220HMFCIO6OK4IJ\u0022, Request id \u00220HMFCIO6OK4IJ:00000002\u0022: An unhandled exception was thrown by the application.",
    "ConnectionId": "0HMFCIO6OK4IJ",
    "TraceIdentifier": "0HMFCIO6OK4IJ:00000002",
    "{OriginalFormat}": "Connection id \u0022{ConnectionId}\u0022, Request id \u0022{TraceIdentifier}\u0022: An unhandled exception was thrown by the application."
  }
}
```

I've not managed to capture the side car log but don't expect any entry as there is no logging line for permanent backoff. I've raised the logging level to debug so if we see something I'll update this issue with results.

Wait for servicebus to have this behavior. Sometimes it will happen a couple of times in a span of days, sometimes it takes weeks for it to happen. It always seems to be intermittent behavior on Azure side.

This environment is on dapr 1.5.1. We also have 1.6 rolled out in other clusters but I don't expect any difference as the `doPublish` code between the 2 versions is the same.

## Release Note

RELEASE NOTE:

**FIX** Azure Service Bus Pub/Sub Connection reset by peer retry

berndverst:

Add retriable error conditions for Azure Service Bus PubSub message sending

Utilizing conditions as listed here:
https://docs.microsoft.com/en-us/javascript/api/@azure/amqp-common/conditionstatusmapper?view=azure-node-preview

Fixes #1495

### DIFF
--- a/pubsub/azure/servicebus/servicebus.go
+++ b/pubsub/azure/servicebus/servicebus.go
@@ -70,6 +70,20 @@ const (
 	defaultPublishInitialRetryInternalInMs = 500
 )
 
+var retriableSendingErrors = map[amqp.ErrorCondition]struct{}{
+	"com.microsoft:server-busy'":             {},
+	amqp.ErrorResourceLimitExceeded:          {},
+	amqp.ErrorResourceLocked:                 {},
+	amqp.ErrorTransferLimitExceeded:          {},
+	amqp.ErrorInternalError:                  {},
+	amqp.ErrorIllegalState:                   {},
+	"com.microsoft:message-lock-lost":        {},
+	"com.microsoft:session-cannot-be-locked": {},
+	"com.microsoft:timeout":                  {},
+	"com.microsoft:session-lock-lost":        {},
+	"com.microsoft:store-lock-lost":          {},
+}
+
 type handle = struct{}
 
 type azureServiceBus struct {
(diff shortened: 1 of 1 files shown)


yaron2 (approved):

halspang (approved):
I think this is fine in general, but is there not a place where the sdk/anything else keeps track of this? I worry it'll get out of sync.

berndverst:
@halspang there is not. And this is the old SDK which we eventually need to replace anyway.

yaron2 (approved):

artursouza (approved):

PR #1524, opened 2022-02-22
Fix ordering in transaction API for PostgreSQL

shubham1172:
Signed-off-by: Shubham Sharma <shubhash@microsoft.com>

# Description

Similar fix to #1516. Only the latest operation for each key and ignore the other operations. This ensures that the ordering of the operations is intact.

## Issue reference

Please reference the issue this PR will close: #1522

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [X] Code compiles correctly
* [X] Created/updated tests
* (NA) Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #1550, opened 2022-03-03
Transaction order conformance test for state stores

shubham1172:

~~This PR is blocked until the following PRs are completed:~~
1. #1518
1. #1523
1. #1524
1. #1542

# Description

Introduces a new test that checks the order of transactions. Say a transaction that contains the following:

```bash
delete k1 # where k1 was already present in the store
upsert k2 v2 # where k2 was not present in the store
delete k2
upsert k3 v3 # where k3 was not present in the store
```
The order of operations must be respected, and `k1` and `k2` should be deleted, whilst `k3` should have a value of `v3`.

Note, an anomaly was found in Cosmos DB state store in #1209 and then discovered for 3 other state stores.

## Issue reference

Please reference the issue this PR will close: #1210

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


tests/conformance/state/state.go
```diff
@@ -453,6 +453,100 @@ func ConformanceTests(t *testing.T, props map[string]string, statestore state.St
 				}
 			}
 		})
+
+		t.Run("transaction-order", func(t *testing.T) {
```
surenderssm, line 457: Is there any sequence to this test? 

Is statestore expected to be initialized with some data set, i ask this because i see a `delete` on `key2` but not for `key1`

tests/conformance/state/state.go
```diff
@@ -453,6 +453,100 @@ func ConformanceTests(t *testing.T, props map[string]string, statestore state.St
...
+				{
+					Operation: state.Upsert,
+					Request: state.SetRequest{
+						Key:   thirdKey,
+						Value: thirdValue,
+					},
+				},
+			}
+
+			expected := map[string][]byte{
```
surenderssm, line 526: nit : How about reword this from `expected` to `expectedKeys` or `expectedKeysMap`, any thoughts?

tests/conformance/state/state.go
mukundansundar: Yes the tests are currently run in sequence ...

tests/conformance/state/state.go
mukundansundar: would suggest `expectedValues`

tests/conformance/state/state.go
shubham1172: This is a standalone test and the only prerequisite is `init`.

There is no expectation from any other tests to set value. That's why we have prerequisites to upsert and delete various keys as needed for the test.

artursouza (approved):

PR #1572, opened 2022-03-12
Adds Azure Active Directory support to the Azure EventHubs Binding

berndverst:
# Description

Now the Event Hubs Binding can connect to Event Hubs and the necessary Storage account using Azure Active Directory.

The required metadata (to be documented) for AAD auth are:

```
"consumerGroup":        os.Getenv(eventHubsBindingsConsumerGroupEnvKey),
"storageAccountName":   os.Getenv(azureBlobStorageAccountEnvKey),
"storageContainerName": os.Getenv(eventHubsBindingsContainerEnvKey),
"eventHub":           os.Getenv(eventHubBindingsHubEnvKey),
"eventHubNamespace":  os.Getenv(eventHubBindingsNamespaceEnvKey),

// plus the Azure specific credentials such as Service Principal (see Authentication with Azure docs)

"azureTenantId":      os.Getenv(azureTenantIdEnvKey),
"azureClientId":      os.Getenv(azureServicePrincipalClientIdEnvKey),
"azureClientSecret":  os.Getenv(azureServicePrincipalClientSecretEnvKey),
```


.github/infrastructure/conformance/azure/conf-test-azure.bicep
```diff
@@ -46,7 +46,7 @@ param certAuthSpId string
 param sqlServerAdminPassword string
 
 var confTestRgName = '${toLower(namePrefix)}-conf-test-rg'
-var acrName = '${toLower(namePrefix)}conftestregistry'
+// var acrName = '${toLower(namePrefix)}conftestregistry'
```
artursouza, line 49: Please, remove code if this is not needed. Leaving code commented out is not good for readability unless it is code example.

artursouza (changes requested):

.github/infrastructure/conformance/azure/conf-test-azure.bicep
berndverst: This file is a special case -- this infrastructure provisioning script is rarely run or modified.

@artursouza I want to keep this because it will be needed in the future. The explanation for that is further down in the file.

We will need this when we create tests on container apps. Today the certification test for KeyVault that relies on containers is commented out as well because it was creating too much flakiness. For container apps we almost certainly will want to publish to Azure Container Registry.

I opted to keep the code here because it is working correct code for the Azure Container Registry infrastructure provisioning which we will need again.

I am not sure yet whether to remove the Keyvault managed identity test entirely. If we decide to remove that code completely we can also clean up the provisioning of the ACR resource at that time.

artursouza (approved):

PR #1578, opened 2022-03-15
Support using AAD auth for Event Hubs Storage account

### LINKED ISSUE #1566: Azure EventHubs PubSub should use AAD for checkpointing storage
Currently EventHubs PubSub has support for AAD authentication, but still requires the storage account connection used for lease checkpointing to use a connection string instead of AAD auth.

This should be updated to also support AAD auth.

berndverst:

# Description

Ensures the Event Hubs binding can connect to the Storage Account for leaser / checkpointer using AAD. This requires at least Storage Blob Contributor role at the Storage Account level.

We are reusing the blob storage token from AAD credential auth logic that is well-tested and used elsewhere.

Fixes #1566


berndverst:
FYI @mukundansundar

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -140,20 +140,22 @@ type AzureEventHubs struct {
 	managementSettings azauth.EnvironmentSettings
 	cgClient           *mgmt.ConsumerGroupsClient
 	tokenProvider      *aad.TokenProvider
+	storageCredential  azblob.Credential
+	azureEnvironment   *azure.Environment
```
mukundansundar, line 144: seems this variable is not used ....

berndverst: one sec :)

berndverst: Done!

mukundansundar: still have the variable.

berndverst: Pushed the wrong branch. Try now.

surenderssm (commented):
Looks good to me !!

Out of context, I do not see notion of balancing strategy while listening to an event hub. I remember using options like balance and greedy in other SDKs.

daixiang0 (changes requested):
Could you add tests to cover changes?

berndverst:
> Could you add tests to cover changes?

This PR is actually a bugfix adding missing behavior to the component. Therefore I would appreciate it if you can approve this @daixiang0.

@mukundansundar tested the component locally previously with AAD authentication but did not update the integration test. He will be creating certification tests to make this component stable. At that point we will be adding the E2E tests.

The alternative IMO is to revert the existing PRs - but that is unnecessary.

daixiang0 (approved):

PR #1581, opened 2022-03-15
[WIP] marshal mongodb bson into relaxed json

mukundansundar:
Signed-off-by: Mukundan Sundararajan <msundar.ms@outlook.com>

# Description

Change to marshal to Ext JSON with Relaxed format instead of Canonical Format. 

Without the change to `canonical` representation as false, the added conformance test scenario fails 
```
state.go:257: Checking value presence for b92a399b3889432782b78babd1169ab8-struct-with-int
    state.go:675: 
        	Error Trace:	state.go:675
        	            				state.go:266
        	Error:      	unmarshal error
        	Test:       	TestStateConformance/mongodb/get
        	Messages:   	error: %!w(*json.UnmarshalTypeError=&{object 0x10721fe20 12 IntValueType message}), json: {"message":{"$numberInt":"42"}}
```

## Issue reference

Please reference the issue this PR will close: #1582 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/state/mongodb/mongodb.go
+++ b/state/mongodb/mongodb.go
@@ -225,7 +225,14 @@ func (m *MongoDB) Get(req *state.GetRequest) (*state.GetResponse, error) {
 	case string:
 		data = []byte(obj)
 	case primitive.D:
-		if data, err = bson.MarshalExtJSON(obj, true, true); err != nil {
+		// Setting canonical to `false`.
+		// See https://docs.mongodb.com/manual/reference/mongodb-extended-json/#bson-data-types-and-associated-representations
+		// Having bson marshalled into Relaxed JSON instead of canonical JSON, this way type preservation is lost but
+		// interoperability is preserved
+		// See https://mongodb.github.io/swift-bson/docs/current/SwiftBSON/json-interop.html
+		// A decimal value stored as BSON will be returned as {"d": 5.5} if canonical is set to false instead of
+		// {"d": {"$numberDouble": 5.5}} when canonical JSON is returned.
+		if data, err = bson.MarshalExtJSON(obj, false, true); err != nil {
 			return &state.GetResponse{}, err
 		}
 	default:
--- a/tests/conformance/state/state.go
+++ b/tests/conformance/state/state.go
@@ -33,7 +33,7 @@ type ValueType struct {
 	Message string `json:"message"`
 }
 
-type IntValueType struct {
+type intValueType struct {
 	Message int32 `json:"message"`
 }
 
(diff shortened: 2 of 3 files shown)


halspang (commented):
Will this change in encoding get to any user's code? Will they have to handle it differently? Or will this all go unnoticed by them?

state/mongodb/mongodb.go
```diff
@@ -225,7 +225,14 @@ func (m *MongoDB) Get(req *state.GetRequest) (*state.GetResponse, error) {
 	case string:
 		data = []byte(obj)
 	case primitive.D:
-		if data, err = bson.MarshalExtJSON(obj, true, true); err != nil {
+		// Setting canonical to `false`.
```
artursouza, line 228: This seems to be a breaking change. If this component is Alpha, it is OK.

mukundansundar: It is not alpha ... It is stable component since v1.0 ...

mukundansundar:
> Will this change in encoding get to any user's code? Will they have to handle it differently? Or will this all go unnoticed by them?

In general, when an ID field with decimal type is stored, the user expects the ID field to be returned as a decimal not as an object with type information. 

So storing {"ID": 1003.4} and getting {"ID":{"$numberDouble": "1003.4"}} is not expected. Any JSON parser when it tries to unmarshal the second object will fail. 

For a user who manually decodes the returned value into an object this will break ie. return string value of the response instead of object, then manually following MongoDB canonical form, convert returned string into object representation.

berndverst:
I agree with this change.

Currently what we have is:
Save State with Dapr API Data Format but Receive State with Mongo DB API Data format. That is super inconsistent and confusing.

Support for the pure Mongo syntax format belongs into a potential Binding, but definitely not into the State Store.

I support making this "breaking" change. I highly doubt someone would expect to receive the current response output when using the Dapr API.

state/mongodb/mongodb.go
dmitsh: > This seems to be a breaking change. If this component is Alpha, it is OK.

This is not a breaking change, but a bug fix.

dmitsh:
This PR should fix querying non-string values: https://github.com/dapr/components-contrib/pull/1584

tests/conformance/state/state.go
```diff
@@ -33,6 +33,10 @@ type ValueType struct {
 	Message string `json:"message"`
 }
 
+type IntValueType struct {
```
shubham1172, line 36: Can this be unexported?

mukundansundar:
@dapr/maintainers-components-contrib Any call on whether this breaking change can go into existing version of component or not?

tests/conformance/state/state.go
mukundansundar: done

mukundansundar:
@dapr/maintainers-components-contrib The conformance test  in PR #1585 as well validates the necessity of  this PR for fixing MongoDB. The call to take is: is it ok if it is a breaking change for few if it is a bug fix .... ?

artursouza:
OK. Let's merge this PR as a fix and mention that in the breaking changes section.

PR #1585, opened 2022-03-16
Adds new state store conformance test to validate Get/Set formats match

berndverst:

# Description

This PR adds a conformance test to verify that the data / body of the State Store Set request and the Data retrieved by the Get request are compatible.


mukundansundar:
to be merged after #1581

artursouza:
Conf test for mongodb is failing.

tests/conformance/state/state.go
```diff
@@ -298,6 +298,45 @@ func ConformanceTests(t *testing.T, props map[string]string, statestore state.St
...
+					t.Logf("Retrieving value for %s", scenario.key)
+					req := &state.GetRequest{
+						Key: scenario.key,
+					}
+					if len(scenario.contentType) != 0 {
+						req.Metadata = map[string]string{metadata.ContentType: scenario.contentType}
+					}
+					res, err := statestore.Get(req)
+					assert.Nil(t, err)
+					assertEquals(t, scenario.value, res)
```
mukundansundar, line 314: The `assertEquals` method actually properly unmarshals, the `res.Data` field and compares it to `scenario.value` ...

tests/conformance/state/state.go
```diff
@@ -298,6 +298,45 @@ func ConformanceTests(t *testing.T, props map[string]string, statestore state.St
...
+					t.Logf("Rewriting retrieved value for %s", scenario.key)
+					setReq := &state.SetRequest{
+						Key:   scenario.key,
+						Value: res,
```
mukundansundar, line 318: This is an issue, the initial `SetRequest` (line 241) uses `scenario.value` (proper Go type) as the `Value` (interface{}) field. But here the whole response is set as the `Value` field. That needs to change. 
Discounting that even if we set the `Value` field to `res.Data` it will not be the same as the previous set request, since the previous set request, the `scenario.value` variable was having proper Go type of the value and not `[]byte` which is the type for `res.Data`. So a proper type conversion needs to be done for this Set call to mimic the initial set so that it can be compared back to the `scenario.value` value. 


Again we need to use the type of `scenario.value` to unmarshal, `res.Data` properly before using that here. 
The below code snippet is essentially a copy of the `assertEquals` method defined at the end of file.
```go

        var d interface{}
	switch v := scenario.value.(type) {
	case intValueType:
		// Custom type requires case mapping
		if err := json.Unmarshal(res.Data, &v); err != nil {
			assert.Failf(t, "unmarshal error", "error: %w, json: %s", err, string(res.Data))
		}
		d = v
	case ValueType:
		// Custom type requires case mapping
		if err := json.Unmarshal(res.Data, &v); err != nil {
			assert.Failf(t, "unmarshal error", "error: %w, json: %s", err, string(res.Data))
		}
		d = v
	case int:
		// json.Unmarshal to float64 by default, case mapping to int coerces to int type
		if err := json.Unmarshal(res.Data, &v); err != nil {
			assert.Failf(t, "unmarshal error", "error: %w, json: %s", err, string(res.Data))
		}
		d = v
	case []byte:
		d = v
	default:
		// Other golang primitive types (string, bool ...)
		if err := json.Unmarshal(res.Data, &v); err != nil {
			assert.Failf(t, "unmarshal error", "error: %w, json: %s", err, string(res.Data))
		}
		d = v
	}
```

And use `d` as the `SetRequest.Value` field in line 318, since now `d` is properly typed and will mimic the `SetRequest` from line 241.

tests/conformance/state/state.go
```diff
@@ -298,6 +298,45 @@ func ConformanceTests(t *testing.T, props map[string]string, statestore state.St
...
+					t.Logf("Retrieving rewritten value for key %s", scenario.key)
+					req2 := &state.GetRequest{
+						Key: scenario.key,
+					}
+					if len(scenario.contentType) != 0 {
+						req.Metadata = map[string]string{metadata.ContentType: scenario.contentType}
+					}
+					res2, err := statestore.Get(req2)
+					assert.Nil(t, err)
+					assertEquals(t, res, res2)
```
mukundansundar, line 334: In `assertEquals` method, the first argument after `t`, here it is `res`, is of type `interface{}` in the method definition whose type is used to properly unmarshal the field `Data` from the later argument, here `res2` which is actually the GetResponse. Again the issue here is `res2.Data` is actual data that we are interested in, and `res` is another GetResponse object and not the actual data that is compared against. 
So this should be `assertEquals(t, scenario.value, res2)`.

mukundansundar (commented):
These are my observations on this conformance test. 
`SetRequest` uses `interface{}` for `Value` field, so proper Go typing data is passed to the callee. 
Whereas, `GetResponse` uses `[]byte` which needs to be properly unmarshaled before it can be compared with the initial set value or further used.

tests/conformance/state/state.go
mukundansundar: Or was it the actual intention to store the response data as such (along with key, etag and value)?

mukundansundar:
Further investigating this, I did try with saving and getting a plain struct with a []byte array field:

Value as seen in MongoDB: 
```
"value" : { "message" : BinData(0,"dGVzdCBieXRlIGFycmF5IHN0b3JhZ2U=") }
```

With Relaxed JSON representation 

```
    state.go:266: Checking value presence for 0b73a40e01a641c0a52011cc191b65be-struct-with-int

time="2022-03-18T19:09:37.193409+05:30" level=info msg="Decoded BSON result [{message 42}]" instance=test scope=testLogger type=log ver=unknown

time="2022-03-18T19:09:37.193421+05:30" level=info msg="BSON marshaled as EXT JSON string : {\"message\":42}" instance=test scope=testLogger type=log ver=unknown

-- Gets unmarshaled correctly. Passes assertions.---

    state.go:266: Checking value presence for 0b73a40e01a641c0a52011cc191b65be-struct-with-byte-array

time="2022-03-18T19:09:37.194188+05:30" level=info msg="Decoded BSON result [{message {0 [116 101 115 116 32 98 121 116 101 32 97 114 114 97 121 32 115 116 111 114 97 103 101]}}]" instance=test scope=testLogger type=log ver=unknown

time="2022-03-18T19:09:37.194201+05:30" level=info msg="BSON marshaled as EXT JSON string : {\"message\":{\"$binary\":{\"base64\":\"dGVzdCBieXRlIGFycmF5IHN0b3JhZ2U=\",\"subType\":\"00\"}}}" instance=test scope=testLogger type=log ver=unknown

--ERROR stack trace unmarshal error---
```

With Canonical JSON representation - before #1581 

```
state.go:266: Checking value presence for dcbdc14d4c804e589cd6bee148d42b75-struct-with-int
time="2022-03-18T19:12:31.463263+05:30" level=info msg="Decoded BSON result [{message 42}]" instance=test scope=testLogger type=log ver=unknown

time="2022-03-18T19:12:31.46328+05:30" level=info msg="BSON marshaled as EXT JSON string : {\"message\":{\"$numberInt\":\"42\"}}" instance=test scope=testLogger type=log ver=unknown

--ERROR stack trace unmarshal error---

state.go:266: Checking value presence for dcbdc14d4c804e589cd6bee148d42b75-struct-with-byte-array
time="2022-03-18T19:12:31.464091+05:30" level=info msg="Decoded BSON result [{message {0 [116 101 115 116 32 98 121 116 101 32 97 114 114 97 121 32 115 116 111 114 97 103 101]}}]" instance=test scope=testLogger type=log ver=unknown

time="2022-03-18T19:12:31.464106+05:30" level=info msg="BSON marshaled as EXT JSON string : {\"message\":{\"$binary\":{\"base64\":\"dGVzdCBieXRlIGFycmF5IHN0b3JhZ2U=\",\"subType\":\"00\"}}}" instance=test scope=testLogger type=log ver=unknown

--ERROR stack trace unmarshal error---
```

The struct with []byte array is stored as binary and the Canonical and Relaxed representations are same as said in the [doc](https://docs.mongodb.com/manual/reference/mongodb-extended-json/#mongodb-bsontype-Binary).

mukundansundar:
@artursouza @yaron2 thoughts on this ?

Issue #1472, opened 2022-02-03
Close method on Resolver interface


jjcollinge:
## Ask your question here
I'm not sure if there's already been discussions around this topic (I can't find any) - is there any reason we cannot add a `Close()` method to the [`resolver`](https://github.com/dapr/components-contrib/blob/8297fc929bc602156eb19daf76c2282c4bbf268a/nameresolution/nameresolution.go#L17) interface to allow any components that use stateful services to unregister? This would allow Consul and mDNS to shutdown correctly without leaving orphaned services. The change would be additive (non-breaking) so was wondering if anyone else had looked into this and found a good reason not to add this?

daixiang0:
Agree to add it.

mukundansundar:
Agree to this. 

@dapr/maintainers-components-contrib 
Labels to add: P1 size:xs good-first-issue area/runtime/nameresolution triaged/resolved

jjcollinge:
I'm happy to implement this @mukundansundar - could you please assign me so I can track it.

mukundansundar:
@jjcollinge please use `/assign` to assign yourself to the issue.

mukundansundar:
not-stale

berndverst:
Not marking this as "good first issue" because it also requires updating the dapr/dapr, and coordinating the changes is cumbersome.

ItalyPaleAle:
@berndverst We don't necessarily need to update the interface for this. It can be implemented as an optional `Close() error` method.

The runtime is already checking, upon termination, if components implement `io.Closer` (e.g. `Close() error`), and if they do, it invokes them:

https://github.com/dapr/dapr/blob/ccce9e451a4b96d1d21a7688196ee7aa89654d6f/pkg/runtime/runtime.go#L2652-L2660

berndverst:
> @berndverst We don't necessarily need to update the interface for this. It can be implemented as an optional `Close() error` method.
> 
> The runtime is already checking, upon termination, if components implement `io.Closer` (e.g. `Close() error`), and if they do, it invokes them:
> 
> https://github.com/dapr/dapr/blob/ccce9e451a4b96d1d21a7688196ee7aa89654d6f/pkg/runtime/runtime.go#L2652-L2660

That works.

ItalyPaleAle:
I think this issue can be closed.

mDNS already implements the Close() method that disposes of resources.

[...]

PR #1591, opened 2022-03-17
WIP: Sync mDNS first browse

jjcollinge:

# Description
As discussed in https://github.com/dapr/dapr/issues/4323 the mDNS name resolver can result in inconsistent performance. I added a concurrency test to understand better how it was behaving and it did indeed suffer poor performance when concurrently called. I believe this was because if the appID was not in the cache, all the concurrent requesters would perform a mDNS browse - to which only some would get a response - this meant others would periodically retry (every 4 seconds) until they got a response and eventually returned the correct address. I've updated this so that when the appID is not in the cache, the first browse is only done by 1 requester per appID and that requester pushes the result to the other waiters via a channel. This has allowed me to run the concurrency test with 1000 requesters and have them all resolve within ~3 seconds.
However, there is still some behaviour I need to understand better. If you perform the test with 1 requester, the test takes 0.254s, if you perform the test with 2 requesters then the test takes 3.2s, if you perform the test with 1000 requesters the test takes 3.2s. I've ran a pprof profile against the test but it hasn't given me much insights - I will continue to dig into this to try and understand better why this is the case.

## Issue reference

Please reference the issue this PR will close: [#dapr/dapr/4323](https://github.com/dapr/dapr/issues/4323)

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/nameresolution/mdns/mdns.go
+++ b/nameresolution/mdns/mdns.go
@@ -122,9 +122,52 @@ func (a *addressList) next() *string {
 	return &addr.ip
 }
 
+type WaiterPool struct {
+	First   *sync.Once
+	Waiters []Waiter
+}
+
+func NewWaiterPool(w Waiter) *WaiterPool {
+	return &WaiterPool{
+		First:   new(sync.Once),
+		Waiters: []Waiter{w},
+	}
+}
+
+func (p *WaiterPool) Add(w Waiter) {
+	p.Waiters = append(p.Waiters, w)
+}
+
+func (p *WaiterPool) Remove(w Waiter) {
+	for i, waiter := range p.Waiters {
+		if waiter == w {
+			p.Waiters = append(p.Waiters[:i], p.Waiters[i+1:]...)
+			break
+		}
+	}
+}
+
+type Waiter struct {
+	AddrChan chan string
+	ErrChan  chan error
+}
+
+func (w *Waiter) Close() {
+	close(w.AddrChan)
+	close(w.ErrChan)
+}
+
+func NewWaiter() Waiter {
+	return Waiter{
+		AddrChan: make(chan string),
+		ErrChan:  make(chan error),
+	}
+}
+
 // NewResolver creates the instance of mDNS name resolver.
 func NewResolver(logger logger.Logger) nameresolution.Resolver {
 	r := &resolver{
+		waiters:          make(map[string]*WaiterPool),
 		appAddressesIPv4: make(map[string]*addressList),
 		appAddressesIPv6: make(map[string]*addressList),
 		refreshChan:      make(chan string),
@@ -155,6 +198,8 @@ func NewResolver(logger logger.Logger) nameresolution.Resolver {
 }
 
 type resolver struct {
+	waiters          map[string]*WaiterPool
+	waitersMu        sync.RWMutex
 	ipv4Mu           sync.RWMutex
 	appAddressesIPv4 map[string]*addressList
 	ipv6Mu           sync.RWMutex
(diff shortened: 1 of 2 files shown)


jjcollinge:
It does appear that the 3s correlates to the refreshTimeout, for every unique appID you get (n-1) * 3s delay.

nameresolution/mdns/mdns.go
```diff
@@ -258,58 +321,168 @@ func (m *resolver) ResolveID(req nameresolution.ResolveRequest) (string, error)
...
-	// is no guarantee that this will ONLY be invoked on the
-	// first address. Ensure that multiple invocations of this
-	// function are safe.
-	onFirst := func(ip string) {
-		addr = ip
-		cancel() // cancel to stop browsing.
-	}
+		err := m.browse(ctx, appID, onFirst)
+		if err != nil {
+			m.pubErrToSubs(appID, err)
```
mukundansundar, line 427: @jjcollinge I think true should be set on the `done` channel after this line.

nameresolution/mdns/mdns.go
```diff
@@ -258,58 +321,168 @@ func (m *resolver) ResolveID(req nameresolution.ResolveRequest) (string, error)
...
+	pool, ok := m.subs[reqID]
+	if !ok {
+		// we would always expect atleast 1 subscriber for this reqID.
+		m.logger.Warnf("no subscribers found for app id %s", reqID)
+		return
 	}
+	for _, subscriber := range pool.Subscribers {
+		// ErrChan is a buffered channel so this is non blocking.
+		subscriber.ErrChan <- err
+		subscriber.Close()
```
mukundansundar, line 468: can this be called as part of the defer function in line 341?
As always a subscriber either receives error (even on cox deadline exceeded), or the address. In which case subscription channels can be closed once the subscriber consumes the address/error.

nameresolution/mdns/mdns.go
```diff
@@ -258,58 +321,168 @@ func (m *resolver) ResolveID(req nameresolution.ResolveRequest) (string, error)
...
+	pool, ok := m.subs[reqID]
+	if !ok {
+		// we would always expect atleast 1 subscriber for this reqID.
+		m.logger.Warnf("no subscribers found for app id %s", reqID)
+		return
+	}
+	for _, subscriber := range pool.Subscribers {
+		// AddrChan is a buffered channel so this is non blocking.
+		subscriber.AddrChan <- addr
+		subscriber.Close()
```
mukundansundar, line 484: can this be called as part of the defer function in line 341?
As always a subscriber either receives error (even on cox deadline exceeded), or the address. In which case subscription channels can be closed once the subscriber consumes the address/error.

nameresolution/mdns/mdns.go
jjcollinge: Yes good catch thanks.

nameresolution/mdns/mdns.go
jjcollinge: It's safer to do this here where we are writing so we know the channel is open - if we do it in the defer statement there are other code paths such as the second cache check that will return before the sender may have written to the channel. In other words, there is a chance the channel may never be read but we want to guarantee that we can at least write to it. In fact, moving this to the defer does result in a panic trying to write to a closed channel during the test.

nameresolution/mdns/mdns.go
jjcollinge: Yes good catch, thanks.

nameresolution/mdns/mdns.go
jjcollinge: ditto

jjcollinge:
@mukundansundar there tests were flaky as there is no `Close()` method on the Resolver interface yet (raised https://github.com/dapr/components-contrib/issues/1472) - I've decided to go ahead and implement the Close method on this resolver and use it in the tests to stop all the goroutine leaks etc. and make sure they run cleanly.

jjcollinge:
Ok I believe I fixed that issue - I can now safely remove the catch all and scale to 100,000s without issue. I will keep the catch all in place just in case but I do not expect it to be used.

mukundansundar:
@jjcollinge Is there some flakiness added to the test for mdns? I am seeing that it failed in Darwin alone and windows got cancelled ....

jjcollinge:
I'll investigate

jjcollinge:
@mukundansundar so this is a balancing act - the previous tests were "flakey" because they took too long to get the address and timed out - I think this was because they were trying to do 50,000 concurrent ResolveID requests and this suffers performance degradation due to lock contention. We cannot support infinite concurrent requesters in finite time so we'll need to decide what behaviour we want. We can keep the timeout at 2 seconds and at some scale this will start to return timeout errors or we remove the timeout and just let the resolver get slower and slower. It probably makes a lot more sense to use 1000 concurrent requesters in the test. WDYT?

mukundansundar:
> @mukundansundar so this is a balancing act - the previous tests were "flakey" because they took too long to get the address and timed out - I think this was because they were trying to do 50,000 concurrent ResolveID requests and this suffers performance degradation due to lock contention. We cannot support infinite concurrent requesters in finite time so we'll need to decide what behaviour we want. We can keep the timeout at 2 seconds and at some scale (hardware depending) this will start to return timeout errors (which is what failed the tests) or we remove the timeout and just let the resolver get slower and slower but not return errors. It probably makes a lot more sense to use 1000 concurrent requesters in the test as this is not a high perf runner and we probably don't expect more than 1000 concurrent calls to ResolveID before the cache is populated in an actual Dapr deployment. On my MBP I can run 300k without hitting the timeout. WDYT? Let's hold off merging this until I can confirm my theory.

As you rightly put it is a balancing act. 
Do you think 1000 concurrent requests might cause issues with low powered dev laptops? And would that cause a flaky test when developer runs the test locally?

jjcollinge:
@mukundansundar I wouldn't think it would causes issues - but I guess we want to be on the safe side here as the importance of this tests is really to exercise the concurrent code path and that can be done at much lower concurrency levels. I will reduce it from 1000 to 100 as that should be high enough to highlight any issues but low enough to not be flakey on any hardware.

jjcollinge:
@mukundansundar I'm fairly happy this is working as expected and I've reduced the concurrency levels to 100. I've just ran this test suite through a few hundred times on my machine (with no cache) and they all passed.

nameresolution/mdns/mdns.go
```diff
@@ -384,81 +673,89 @@ func (m *resolver) refreshAllApps(ctx context.Context) error {
 }
 
 // browse will perform a non-blocking mdns network browse for the provided app id.
-func (m *resolver) browse(ctx context.Context, appID string, onEach func(ip string)) error {
+func (m *Resolver) browse(ctx context.Context, appID string, onEach func(ip string)) error {
```
mukundansundar, line 676: quick question why is the callback named `onEach`? The only non nil value it gets is an `onFirst` method?

nameresolution/mdns/mdns.go
```diff
@@ -384,81 +673,89 @@ func (m *resolver) refreshAllApps(ctx context.Context) error {
...
-						break
-					}
+			// TODO: we currently only use the first IPv4 and IPv6 address.
+			// We should understand the cases in which additional addresses
+			// are returned and whether we need to support them.
+			if hasIPv4Address {
+				addr = fmt.Sprintf("%s:%d", entry.AddrIPv4[0].String(), port)
+				m.addAppAddressIPv4(appID, addr)
+			}
+			if hasIPv6Address {
```
mukundansundar, line 712: here I can see you are using the same logic as was before ... but is it fair to assume that ipv6 has preference over ipv4 if both are there for an app?

nameresolution/mdns/mdns.go
```diff
@@ -384,81 +673,89 @@ func (m *resolver) refreshAllApps(ctx context.Context) error {
...
+			select {
+			case entry := <-results:
+				if entry == nil {
+					break
+				}
+				handleEntry(entry)
+			case <-ctx.Done():
+				// drain the results before exiting.
+				for len(results) > 0 {
+					handleEntry(<-results)
```
mukundansundar, line 735: why do we need to drain the results here ?

nameresolution/mdns/mdns.go
```diff
@@ -122,49 +129,144 @@ func (a *addressList) next() *string {
...
+				return
 			}
 		}
 	}()
 
 	return r
 }
 
-type resolver struct {
+type Resolver struct {
```
mukundansundar, line 239: does this struct need to be exported if only NewResolver is used to access it ?

nameresolution/mdns/mdns.go
```diff
@@ -122,49 +129,144 @@ func (a *addressList) next() *string {
...
+		logger:                  logger,
 	}
 
 	// refresh app addresses on demand.
 	go func() {
-		for appID := range r.refreshChan {
-			if err := r.refreshApp(context.Background(), appID); err != nil {
-				r.logger.Warnf(err.Error())
+		for {
+			refreshCtx, cancel := context.WithCancel(context.Background())
```
mukundansundar, line 203: Should this line be moved to the switch case when an appID is available in the refreshChan?

nameresolution/mdns/mdns_test.go
```diff
@@ -162,6 +239,223 @@ func TestResolverMultipleInstances(t *testing.T) {
...
+	appC := nr.Metadata{Properties: map[string]string{
+		nr.MDNSInstanceName:    appCName,
+		nr.MDNSInstanceAddress: appCAddress,
+		nr.MDNSInstancePort:    appCPort,
+		nr.MDNSInstanceID:      appCID,
+	}}
+	err3 := resolver.Init(appC)
+	require.NoError(t, err3)
+
+	// act...
```
mukundansundar, line 375: In this test case, I can see that Init is called on the same resolver 3 times. Is this to facilitate the concurrency test? 
Because my understanding is when daprd starts, NewResolver is called once and Init is called when initNameResolution() is called from the runtime.go file.

jjcollinge: yes this is not the behaviour used by the runtime but just the tests.

nameresolution/mdns/mdns.go
jjcollinge: Not exporting it means NewResolver returns an unexported type or we return an interface - both which are linting issues as they hide the concrete type.

nameresolution/mdns/mdns.go
jjcollinge: it's a fair point so please open a new issue to track that.

nameresolution/mdns/mdns.go
jjcollinge: legacy reasons, this was originally a non nil value in refreshApp but is no longer used as browse updates the cache directly.

nameresolution/mdns/mdns.go
```diff
@@ -194,17 +296,51 @@ func (m *resolver) Init(metadata nameresolution.Metadata) error {
...
+	return nil
+}
+
+func (m *Resolver) registerMDNS(instanceID string, appID string, ips []string, port int) error {
 	started := make(chan bool, 1)
 	var err error
 
+	// Register the app id with the resolver.
+	done := make(chan struct{})
+	key := fmt.Sprintf("%s:%d", appID, port) // WARN: we do not support unique ips.
```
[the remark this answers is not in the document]
mukundansundar: Based on reply https://github.com/dapr/components-contrib/pull/1591#discussion_r831373152 is a map needed here ?

nameresolution/mdns/mdns.go
```diff
@@ -373,7 +659,10 @@ func (m *resolver) refreshAllApps(ctx context.Context) error {
 		go func(a string) {
 			defer wg.Done()
 
-			m.refreshApp(ctx, a)
+			err := m.refreshApp(ctx, a)
+			if err != nil {
+				m.logger.Warnf("error refreshing mDNS addresses for app id %s: %v", a, err)
```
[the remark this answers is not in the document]
jjcollinge: Good question, so assuming the reason we can't refresh the app causes every app ID refresh to fail. You'd log once per app ID every 30 seconds. Depending how you monitor this might be annoying behaviour but I don't think it'd flood the log. Maybe it should be an aggregated error using `uber.go/multierr` to avoid multiple log lines.

nameresolution/mdns/mdns.go
```diff
@@ -122,49 +129,144 @@ func (a *addressList) next() *string {
...
+}
+
+func (s *Subscriber) Close() {
+	close(s.AddrChan)
+	close(s.ErrChan)
+}
+
+func NewSubscriber() Subscriber {
+	return Subscriber{
+		// ID is assigned by the pool.
```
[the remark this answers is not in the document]
jjcollinge: Correct, this is a left over comment and should be removed.

nameresolution/mdns/mdns.go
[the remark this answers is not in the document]
jjcollinge: This map is dapr instance - so in reality we only expect 1 app to register with the current behaviour. The map exists to track registrations so the resolver can be closed properly and not leak goroutines in the tests.

nameresolution/mdns/mdns.go
```diff
@@ -122,49 +129,144 @@ func (a *addressList) next() *string {
...
+				return
 			}
 		}
 	}()
 
 	// refresh all app addresses periodically.
 	go func() {
 		for {
-			time.Sleep(refreshInterval)
+			refreshCtx, cancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
jjcollinge: Yes this could be moved above the `for`. Thanks.

nameresolution/mdns/mdns.go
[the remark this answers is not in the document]
mukundansundar: Can you add the comment that the map is only for tests and that we expect only 1 app registration?

nameresolution/mdns/mdns.go
jjcollinge: Yes ideally - good spot

nameresolution/mdns/mdns.go
mukundansundar: done https://github.com/dapr/components-contrib/issues/1606

jjcollinge: thanks

PR #1597, opened 2022-03-17
[WIP] Fix serialization issue with 64 bit data type

shubham1172:

# Description

This PR changes the serialization logic in pubsub's CloudEvent envelope to use `encoding/json` with `UseNumber` on its decoder, to prevent data loss with float64 conversion,

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/3837

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/pubsub/envelope.go
+++ b/pubsub/envelope.go
@@ -14,12 +14,13 @@ limitations under the License.
 package pubsub
 
 import (
+	"bytes"
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"time"
 
 	"github.com/google/uuid"
-	jsoniter "github.com/json-iterator/go"
 
 	contrib_contenttype "github.com/dapr/components-contrib/contenttype"
 	contrib_metadata "github.com/dapr/components-contrib/metadata"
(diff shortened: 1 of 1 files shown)


mukundansundar (commented):
@shubham1172  Can you add tests which use large numbers and verify the encoding and decoding is correct?

shubham1172:
> @shubham1172 Can you add tests which use large numbers and verify the encoding and decoding is correct?

@mukundansundar added test for the publisher side. Without `json.Number`, we would expect to see floating point (attached issue has an example), but now we get the entire number. The part on the subscriber is unchanged, and will work as expected.

daixiang0 (approved):

mukundansundar (approved):
lgtm. But @dapr/maintainers-components-contrib This is a breaking changing, though it is a bug fix. For smaller changes also, now the data is going to be serialised as a string.

mukundansundar:
Based on https://github.com/dapr/dapr/issues/3837#issuecomment-1074238941 ... removing breaking change label ...

PR #1602, opened 2022-03-18
fix/typo: Change traceid to traceparent

1046102779:
Because traceparent can transmit info={version, traceid, spanid, traceflag} to consumer. Only transmit traceid is not useful.

https://github.com/dapr/dapr/issues/4398

https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/extensions/distributed-tracing.md#traceparent


yaron2:
@1046102779 please make this change backwards compatibles, so keep traceid and add traceparent.

1046102779:
> @1046102779 please make this change backwards compatibles, so keep traceid and add traceparent.

ok. we may remove traceid after a few versions。

PR #1604, opened 2022-03-19
fix/typo: Change traceid to traceparent

1046102779:
https://github.com/dapr/components-contrib/pull/1602

fix DCO.


daixiang0 (approved):
/cc @yaron2

PR #1608, opened 2022-03-22
use releaxedJson for get behind a flag

mukundansundar:
Signed-off-by: Mukundan Sundararajan <msundar.ms@outlook.com>

# Description

Follow up to #1581 , have put the relaxedJson representation behind a flag in GetRequest alone. 

When this metadata is passed in GetRequest, then relaxedJson representation is used, else the v1.6.0 behavior of canonicalJson will be used. This will make the PR 1581 as not a breaking change immediately for a Stable component Get API. 

After 2 releases following normal Dapr deprecation policy, this flag can be removed. 

For now it will be deprecated and a Warning log will be printed for every Get Request which uses canonical json representation. 

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/state/mongodb/mongodb.go
+++ b/state/mongodb/mongodb.go
@@ -47,7 +47,7 @@ const (
 	writeConcern     = "writeConcern"
 	readConcern      = "readConcern"
 	operationTimeout = "operationTimeout"
-	relaxedJsonKey   = "relaxedJson"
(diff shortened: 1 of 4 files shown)


mukundansundar:
> I think it is better to keep the breaking change and not default to the warning behavior.

closing this PR.

PR #1613, opened 2022-03-24
Workaround for Cosmos DB state store unable to launch when using AAD auth

### LINKED ISSUE #1603: Cosmos DB: Auth with Azure AD (incl. MSI) failing because CRUD on stored procedures is a management plane operation
## Expected Behavior

Should be able to use Cosmos DB (e.g. state store) with Azure AD credentials, including MSI.

## Actual Behavior

Initializing a state store component that uses Cosmos DB fails with an error similar to:

```
process component mystatestore error: Forbidden, Request blocked by Auth *** : The given request [GET /dbs/***/colls/***/sprocs/] cannot be authorized by AAD token in data plane. Learn more: https://aka.ms/cosmos-native-rbac.\r\nActivityId: ***, Microsoft.Azure.Documents.Common/2.14.0" app_id=*** instance=*** scope=dapr.runtime type=log ver=edge
```

The issue is that the Cosmos DB components need to access stored procedures and CRUD operations on those are management plane operations that can't be performed with tokens for the data plane ([source](https://docs.microsoft.com/en-us/azure/cosmos-db/how-to-setup-rbac#permission-model)). Currently the SDK requests tokens for data plane operations only, while management plane operations require different tokens.

This needs to be fixed in the SDK first: https://github.com/a8m/documentdb

## Release Note

RELEASE NOTE: **FIX** Issues authenticating with Cosmos DB using Azure AD credentials

/assign

ItalyPaleAle:
Fixes #1603

# Description

As explained in #1603, wen using Azure AD auth (including MSI) we cannot perform CRUD operations on stored procedures on Cosmos DB, so the component failed to init.

This works around that issue by creating a secondary stored procedure that returns the version of the Dapr SP. If authenticating with AAD, that SP is invoked and if the version mismatches, we guide the user to the docs where they can learn how to set things up (docs update is on my TODO).

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_ **TODO**

### DIFF
--- a/state/azure/cosmosdb/cosmosdb.go
+++ b/state/azure/cosmosdb/cosmosdb.go
@@ -39,9 +39,6 @@ import (
 type StateStore struct {
 	state.DefaultBulkStore
 	client      *documentdb.DocumentDB
-	collection  *documentdb.Collection
-	db          *documentdb.Database
-	sp          *documentdb.Sproc
 	metadata    metadata
 	contentType string
 
@@ -87,10 +84,12 @@ type storedProcedureDefinition struct {
 
 const (
 	storedProcedureName   = "__dapr_v2__"
+	versionSpName         = "__daprver__"
 	metadataPartitionKey  = "partitionKey"
 	unknownPartitionKey   = "__UNKNOWN__"
 	metadataTTLKey        = "ttlInSeconds"
 	statusTooManyRequests = "429" // RFC 6585, 4
+	statusNotFound        = "NotFound"
 )
 
 // NewCosmosDBStateStore returns a new CosmosDB state store.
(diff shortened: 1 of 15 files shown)


ItalyPaleAle:
Upstream SDK was just updated (https://github.com/a8m/documentdb/pull/36) so I've pushed an update to go.mod too. This PR is good to review/merge.

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -156,79 +155,39 @@ func (c *StateStore) Init(meta state.Metadata) error {
...
-			return backoff.Permanent(err)
-		}
-
-		// get a link to the sp
-		for i := range sps {
-			if sps[i].Id == storedProcedureName {
-				c.sp = &sps[i]
-
-				break
+	err = backoff.RetryNotify(func() (innerErr error) {
```
ItalyPaleAle, line 166: I moved most of this lambda function to a separate function below to reduce its complexity

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -39,9 +39,6 @@ import (
 type StateStore struct {
 	state.DefaultBulkStore
 	client      *documentdb.DocumentDB
-	collection  *documentdb.Collection
```
ItalyPaleAle, line 42: Thanks to the latest updates in the upstream SDK, we can now refer to collections in Cosmos DB via their ID and not just the "self link". So there's no need to save these in the object anymore. The function `c.findCollection()` returns the link, generated every time.

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -456,23 +419,105 @@ func (c *StateStore) Query(req *state.QueryRequest) (*state.QueryResponse, error
...
+func (c *StateStore) getCollectionLink() string {
+	return fmt.Sprintf("dbs/%s/colls/%s/", c.metadata.Database, c.metadata.Collection)
+}
+
+// getSprocLink returns the link to the stored procedures.
+func (c *StateStore) getSprocLink(sprocName string) string {
+	return fmt.Sprintf("dbs/%s/colls/%s/sprocs/%s", c.metadata.Database, c.metadata.Collection, sprocName)
+}
+
+func (c *StateStore) checkStoredProcedures() error {
```
ItalyPaleAle, line 436: `checkStoredProcedures` is used for AAD auth

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -456,23 +419,105 @@ func (c *StateStore) Query(req *state.QueryRequest) (*state.QueryResponse, error
...
+	if err == nil {
+		c.logger.Debugf("Cosmos DB stored procedure version: %d", ver)
+	}
+	if err != nil || (err == nil && ver != spVersion) {
+		return fmt.Errorf("Dapr requires stored procedures created in Cosmos DB before it can be used as state store. Those stored procedures are currently not existing or are using a different version than expected. When you authenticate using Azure AD we cannot automatically create them for you: please start this state store with a Cosmos DB master key just once so we can create the stored procedures for you; otherwise, you can check our docs to learn how to create them yourself: https://aka.ms/dapr/cosmosdb-aad") // nolint:stylecheck
+	}
+	return nil
+}
+
+func (c *StateStore) ensureStoredProcedures() error {
```
ItalyPaleAle, line 448: `ensureStoredProcedures` is used for auth with master key

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -456,23 +419,105 @@ func (c *StateStore) Query(req *state.QueryRequest) (*state.QueryResponse, error
...
+
+	// check version
+	replace := false
+	if verSp != nil {
+		var ver int
+		err = c.client.ExecuteStoredProcedure(verSpLink, nil, &ver, documentdb.PartitionKey("1"))
+		if err == nil {
+			c.logger.Debugf("Cosmos DB stored procedure version: %d", ver)
+		}
+		if err != nil || (err == nil && ver != spVersion) {
```
ItalyPaleAle, line 470: Note that this checks the actual version and if it's the same as what we were expecting. If the version is different (lower/higher), it triggers the commands to replace the stored procedures.

This should address all cases, including when customers upgrade Dapr, or even when they downgrade it.

state/azure/cosmosdb/storedprocedure.go
```diff
@@ -116,3 +118,8 @@ function dapr_multi_v2(operations) {
         }
     }
 }`
+
+const spVersionDefinition = `function daprSpVersion(prefix) {
```
ItalyPaleAle, line 122: This is the new sproc. It just returns a constant with the version of the stored procedure

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -156,79 +155,39 @@ func (c *StateStore) Init(meta state.Metadata) error {
...
-		}
-
-		// get a link to the sp
-		for i := range sps {
-			if sps[i].Id == storedProcedureName {
-				c.sp = &sps[i]
-
-				break
+	err = backoff.RetryNotify(func() (innerErr error) {
+		_, innerErr = c.findCollection()
```
ItalyPaleAle, line 167: Because of the latest update in the upstream SDK that allows us to address collections and sprocs via their ID and not just a self link, I've simplified this to a single call to `ReadCollection` rather than 2 queries before. This is also a simpler query that runs faster and uses less RUs :)

berndverst:
@ItalyPaleAle why do we need the new stored procedure? That part doesn't make sense to me.

We could just check whether the existing stored procedure constant exists in the list of stored procedures - this is similar to what you are doing in `ensureStoredProcedure` and if it doesn't provide instructions for creating it. Shouldn't that suffice?

If AAD auth can't modify stored procedures I'd assume you have to manually create the stored procedure whenever the stored procedure is updated in Dapr anyway. FYI - whenever we update it in Dapr we choose a new name constant. Therefore the check for the existence of the stored procedure would rightly identify that the stored procedure must be created manually.

Or are you saying with AAD auth you can actually update a stored procedure but not create a new one and therefore?

Please clarify for more context.

ItalyPaleAle:
With AAD you cannot perform any CRUD operation on stored procedures. You can't even check if a SP exists. You can only invoke a SP. This is why I thought of creating a second SP that just returns the version.

But open to suggestions if you have better options!

berndverst:
Ah, the fact that you can't check whether it exists explains your approach.

Sounds good - I'll take another look at this PR.

berndverst (dismissed):
LGTM

berndverst:
@ItalyPaleAle please run "make modtidy-all" then push a commit for all the go.mod / go.sum changes.

berndverst (approved):

PR #1634, opened 2022-04-04
Use go:embed for Cosmos DB stored procedures

ItalyPaleAle:
Sorry is this is so last-minute.

While writing the docs for manually creating the stored procedures for Cosmos DB (see: https://github.com/dapr/docs/issues/2325 ), I realized that I was going to have to include source code (the JS of the stored procedures) inside the docs. That was not ideal.

A better solution would be to just tell users to `curl` the JS source code of the stored procedures from a tag in this repository. However, currently the source of the stored procedures was included in a variable in a Go file, so it was not possible to fetch that code only. Using Go embeds allows maintaining the stored procedures' source code in a separate file and have it embedded at build time.

TL;DR: After this PR is merged and we have the source code of the stored procedures in a tag, users will be able to just do something like `curl -LO "https://raw.githubusercontent.com/dapr/components-contrib/v1.7.0/state/azure/cosmosdb/storedprocedures/__dapr_v2__.js"` to get the source code of the stored procedure. This would significantly improve the quality of life of users following our docs.

### DIFF
--- a/state/azure/cosmosdb/cosmosdb.go
+++ b/state/azure/cosmosdb/cosmosdb.go
@@ -14,6 +14,7 @@ limitations under the License.
 package cosmosdb
 
 import (
+	_ "embed"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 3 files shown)


ItalyPaleAle:
If this can't be merged into `release-1.7`, I think for this release we could accept merging into master only. The command would then be `curl -LO "https://raw.githubusercontent.com/dapr/components-contrib/master/state/azure/cosmosdb/storedprocedures/__dapr_v2__.js"` . This is not as good because it's untagged, so the source code could change, but it's better than nothing (we can then add a tag starting with 1.8)

ItalyPaleAle:
Closing in favor of #1635 (merged into release-1.7) and #1636 (targeting master)

PR #1636, opened 2022-04-05
Use go:embed for Cosmos DB stored procedures

ItalyPaleAle:
This is the second spin-off from #1634 and it targets the master branch (not for 1.7)

With #1635 the code of the Cosmos DB stored procedures was copied into JS files to be included in the source tree, however those files were not used (that is, they did not have an impact on the final build artifact).

This completes the work by actually making Dapr refer to those files at build time, avoiding duplication. Using Go Embed, at build-time the content of the JS files is included into the binary. (Bonus: it's now possible to edit the JS code as JS and not as part of a Go string!)

See #1634 for context.


berndverst:
Linter failure will be expected and will be resolved once release-1.7 is merged into master.

ItalyPaleAle:
@berndverst this should be ready for review now, targeting 1.8

berndverst (approved):

PR #1642, opened 2022-04-05
Merge release 1.7 rc3 into master

berndverst:
# Description

Merge release 1.7 rc3 into master

We do this now instead of after v1.7.0 release because we need some test fixes that are only in the release branch.


berndverst:
The irony is that this fails because of the Golinter workflow fix which is in this very PR itself :)

artursouza (approved):

PR #1643, opened 2022-04-05
Yet another Cosmos DB auth fix

ItalyPaleAle:
This is fixed by upgrading the upstream SDK
Also added a missing handling of throttling in the Delete method


berndverst (approved):
Verified on my local machine the conformance and certification tests pass with this change.

PR #1644, opened 2022-04-05
Cosmos DB: Delete does not need to query for the doc first 

ItalyPaleAle:
This, which is possible now because of improvements in the upstream SDK, should speed up delete operations significantly (and reducing RU usage).

Also, this should fix potential concurrency issues.

(Note this is built on top of #1643)


berndverst:
Will merge this AFTER release-1.7 is merged into master.

berndverst (approved):

PR #1646, opened 2022-04-05
Update azure service bus go to 0.11.5

yaron2:
This PR updates the Azure Service Bus Go SDK version to 0.11.5. There have been many reliability fixes since 0.10.10.


berndverst:
Some method signatures changed -- you cannot simply update the SDK

PR #1654, opened 2022-04-09
WIP add sqlite state store (pending for gRPC component)

lotuc:
# Description

Implemented a sqlite state store.

The implementation is largely a copy of the oracle state store. And since it uses [github.com/mattn/go-sqlite3](github.com/mattn/go-sqlite3) as sqlite driver, `CGO` needs to enabled.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/state/sqlite/sqlitesetvalue.go
+++ b/state/sqlite/sqlitesetvalue.go
@@ -16,8 +16,8 @@ import (
 
 // Parsed Set Request.
 type setRequest struct {
-	dbAccess sqliteDBAccess
-	tx       *sql.Tx
+	tx        *sql.Tx
+	tableName string
 
 	key         string
 	value       string
@@ -60,8 +60,9 @@ func prepareSetRequest(a *sqliteDBAccess, tx *sql.Tx, req *state.SetRequest) (*s
 	value := string(bt)
 
 	return &setRequest{
-		dbAccess:    *a,
-		tx:          tx,
+		tx:        tx,
+		tableName: a.tableName,
+
 		key:         req.Key,
 		value:       value,
 		concurrency: &req.Options.Concurrency,
@@ -74,7 +75,7 @@ func prepareSetRequest(a *sqliteDBAccess, tx *sql.Tx, req *state.SetRequest) (*s
 func (req *setRequest) setValue() (bool, error) {
 	newEtag := uuid.New().String()
 	tx := req.tx
-	tableName := req.dbAccess.tableName
+	tableName := req.tableName
 
 	// Only check for etag if FirstWrite specified (ref oracledatabaseaccess)
 	var res sql.Result
--- a/state/sqlite/sqlitesetvalue.go
+++ b/state/sqlite/sqlitesetvalue.go
@@ -22,7 +22,7 @@ type setRequest struct {
 	key         string
 	value       string
 	isBinary    bool
-	ttlSeconds  *int
+	ttlSeconds  *int64
 	concurrency *string
 	etag        *string
 }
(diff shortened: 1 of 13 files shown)


lotuc:
Some lint problem related to this PR is found. Please ignore my latest push. I'll fix it soon.

state/sqlite/sqlitesetvalue.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	requestValue := req.Value
+	byteArray, isBinary := req.Value.([]uint8)
+	if isBinary {
+		requestValue = base64.StdEncoding.EncodeToString(byteArray)
+	}
+
+	// Convert to json string.
+	bt, _ := utils.Marshal(requestValue, json.Marshal)
```
daixiang0, line 56: Please handle error.

state/sqlite/sqlitesetvalue.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// Returns nil or non-negative value, nil means never expire.
+func parseTTL(requestMetadata map[string]string, logger logger.Logger) (*int, error) {
+	if val, found := requestMetadata[metadataTTLKey]; found && val != "" {
+		parsedVal, err := strconv.ParseInt(val, 10, 0)
+		if err != nil {
+			return nil, fmt.Errorf("error in parsing ttl metadata : %w", err)
+		}
+		parsedInt := int(parsedVal)
```
daixiang0, line 144: Still need to convert after `strconv.ParseInt`?

lotuc:
@daixiang0 Hi, thanks for reviewing. I just fixed some problems you mentioned.

And `make lint | grep -i sqlite` now returns nothing and `make test` is also fine. 

But `golangci-lint run --max-issues-per-linter 2000 --max-same-issues 2000 | grep -i sqlite` returns some lint errors like 

```
...
state/sqlite/sqlite_integration_test.go:793:6            thelper           test helper function should start from t.Helper()
...
state/sqlite/sqlitesetvalue.go:76:2                      varnamelen        variable name 'tx' is too short for the scope of its usage
```

Should I fix them all? Some of them seems harmless like `tx` var name for a `sql.Tx`.

ps. I run the lint command in the dev container.

lotuc:
And also, I'm not rebasing when fixing subsequent problems. Will do some rebasing & cleanup when it's finally accepted, is that ok?

lotuc:
@daixiang0 Hi, I just passed conformation test locally in the dev container.
```
CGO_ENABLED=1 dlv test --build-flags '-v -tags=conftests' ./tests/conformance -- -test.run "TestStateConformance/sqlite"

make CGO=1 test
```

And also, `make lint | grep -i sqlite` returns nothing yet. Would you mind give it a review? Thanks a lot.

berndverst:
@lotuc if CGO needs to be enabled then we cannot accept this. Our builds of Dapr disable CGO by default -- and this is intentional.

https://github.com/dapr/dapr/blob/master/Makefile#L25

FYI @yaron2 @artursouza

artursouza:
Is there a way to have SQLite without CGO?

yaron2:
@lotuc We really appreciate this PR, but enabling CGO compilation in daprd is problematic and has broader, less desirable implications on the project.

We are working on enabling gRPC enabled components for 1.8, which would allow this component to run as it's own CGO process without affecting the daprd process.

So I see the following options:

1. Change this implementation to not depend on CGO. If this is doable, we can merge to daprd right away.
2. Keep this PR open as-is and wait until gRPC components are enabled.

lotuc:
@yaron2 Ok, I'll keep this open until gRPC components is ready.

ItalyPaleAle:
@lotuc I've taken your code, reviewed it, and implemented it in a pluggable component that should work with Dapr 1.9! Take a look here: https://github.com/ItalyPaleAle/dapr-sqlite-statestore

I'm asking for that repo to be accepted into the [dapr-sandbox org](https://github.com/dapr/community/issues/231).

Issue #1601, opened 2022-03-18
API  adds context in building blocks.


1046102779:
Whether the building blocks standard API should add context to transmit timeout mechanism, trace and other information? I think it is very necessary.

<img width="1596" alt="image" src="https://user-images.githubusercontent.com/20457624/159002341-7f6ea561-1735-4e31-bca4-0327960c31b7.png">

I think the way of handling is unfriendly. It is very necessary to add context to the network transmission. At the same time, we can pass the context information to the downstream, so that the entire trace link can be formed with the downstream. a trace link should not be only exist in dapr cluster.

But `component.type= configuration` exists context, API definitions are not consistent.
```golang
// Store is an interface to perform operations on store.
type Store interface {
	// Init configuration store.
	Init(metadata Metadata) error

	// Get configuration.
	Get(ctx context.Context, req *GetRequest) (*GetResponse, error)

	// Subscribe configuration by update event.
	Subscribe(ctx context.Context, req *SubscribeRequest, handler UpdateHandler) (string, error)

	// Unsubscribe configuration with keys
	Unsubscribe(ctx context.Context, req *UnsubscribeRequest) error
}

```

1046102779:
1. Binding: Change from OutputBinding interface
```golang

// OutputBinding is the interface for an output binding, allowing users to invoke remote systems with optional payloads.
type OutputBinding interface {
	Init(metadata Metadata) error
	Invoke(req *InvokeRequest) (*InvokeResponse, error)
	Operations() []OperationKind
}
```

to 

```golang
// OutputBinding is the interface for an output binding, allowing users to invoke remote systems with optional payloads.
type OutputBinding interface {
	Init(metadata Metadata) error
	Invoke(ctx context.Context, req *InvokeRequest) (*InvokeResponse, error)
	Operations() []OperationKind
}
```

2. State: Change from Store And BulkStore interface
```golang
// Store is an interface to perform operations on store.
type Store interface {
	BulkStore
	Init(metadata Metadata) error
	Features() []Feature
	Delete(req *DeleteRequest) error
	Get(req *GetRequest) (*GetResponse, error)
	Set(req *SetRequest) error
	Ping() error
}

// BulkStore is an interface to perform bulk operations on store.
type BulkStore interface {
	BulkDelete(req []DeleteRequest) error
	BulkGet(req []GetRequest) (bool, []BulkGetResponse, error)
	BulkSet(req []SetRequest) error
}
```

to 

```golang
[...]

PR #1681, opened 2022-04-24
Add context in bindings interface

pigletfly:
Signed-off-by: pigletfly <wangbing.adam@gmail.com>

# Description

add context.Context in bindings interface.

## Issue reference
https://github.com/dapr/dapr/issues/2716
https://github.com/dapr/components-contrib/issues/1601

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/bindings/alicloud/dingtalk/webhook/webhook.go
+++ b/bindings/alicloud/dingtalk/webhook/webhook.go
@@ -51,7 +51,7 @@ type webhookResult struct {
 }
 
 type outgoingWebhook struct {
-	handler func(*bindings.ReadResponse) ([]byte, error)
+	handler func(context.Context, *bindings.ReadResponse) ([]byte, error)
 }
 
 var webhooks = struct { // nolint: gochecknoglobals
@@ -94,7 +94,7 @@ func (t *DingTalkWebhook) Init(metadata bindings.Metadata) error {
 }
 
 // Read triggers the outgoing webhook, not yet production ready.
-func (t *DingTalkWebhook) Read(handler func(*bindings.ReadResponse) ([]byte, error)) error {
+func (t *DingTalkWebhook) Read(handler func(context.Context, *bindings.ReadResponse) ([]byte, error)) error {
 	t.logger.Debugf("dingtalk webhook: start read input binding")
 
 	webhooks.Lock()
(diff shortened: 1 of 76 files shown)


berndverst (changes requested):
@pigletfly Can you please undo the reordering of the imports? The existing order was much easier to read and understand.

A few things to note:
You also need to open a PR against `dapr/dapr` to actually provide the context to these binding methods.

It needs to be added to `binding.Read` and `binding.Invoke`

https://github.com/dapr/dapr/blob/master/pkg/runtime/runtime.go#L844
https://github.com/dapr/dapr/blob/master/pkg/runtime/runtime.go#L1019

berndverst:
This PR looks good, but we shouldn't reorder the imports like this :)

berndverst:
Also as you can see the stable components certification tests for bindings will fail because of this PR. You will need to update the runtime to set the context, and then you will need to update those tests as I mentioned before.

pigletfly:
ok, I will update this PR step by step

pigletfly:
@daixiang0 could you help to review this PR and trigger the CI workflow ?

berndverst:
@pigletfly the CI workflow already ran when you pushed to it. There is still a linter failure -- you need to `gofumpt` the files you modified.

There is no need to also get a review from @daixiang0 since I am a maintainer, but feel free. Once you fix the linter issues (by making `components-contrib / Build linux_amd64` pass) I can approve and merge this.

In the meantime, could you please start making the `dapr/dapr` PR? You can later update that PR by adding a commit with the `go.mod` and `go.sum` update to import the latest contrib via `go get github.com/dapr/components-contrib@master`, we really want to be quick about merging all of these things so that our tests will not be failing for a longer period of time.

berndverst:
Thanks for already updating the certification tests @pigletfly. By the way it is expected that these will fail at this time.

That's why it is important we get the `dapr/dapr` created and merged soon, and then another PR for contrib for the submodule `test/certification` to pin the latest `dapr/dapr` there.

pigletfly:
@berndverst I pinged @daixiang0 because we are in the same time zone. PR in `dapr/dapr` https://github.com/dapr/dapr/pull/4550

berndverst (approved):

berndverst:
@pigletfly unfortunately even as a maintainer I cannot merge this because of the breaking certification tests -- I will need to work with a steering committee member to force merge this. Maintainers do not have this power unfortunately.

Maybe @yaron2 or @artursouza can help out here.

Certification test failures are expected as discussed. To resolve those failures we must merge https://github.com/dapr/dapr/pull/4550 (and bring in the latest contrib with this very PR here), then we need to update the certification tests to pin the latest dapr/dapr.

pigletfly:
sounds like there is a circular dependency between dapr/dapr and dapr/components-contrib ?

berndverst:
@pigletfly sort of -- the main issue is that the stable certification tests are a submodule within contrib. That has some advantages (e.g. you can immediately run certification tests when you modify a component due to the local replace directive) but it also has the disadvantage of this complicated process whenever you change the Dapr runtime.

pigletfly:
thanks for your detailed explanations. @berndverst

artursouza (approved):

PR #1693, opened 2022-04-27
Fix 4529: Ignore Subscribe/Get wrong redis configuration type keys.

LaurenceLiZhixin:
# Description

## Issue reference

fix: https://github.com/dapr/dapr/issues/4529

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


LaurenceLiZhixin:
@daixiang0  cc @beiwei30

daixiang0 (dismissed):

configuration/redis/redis.go
```diff
@@ -242,6 +243,10 @@ func (r *ConfigurationStore) Get(ctx context.Context, req *configuration.GetRequ
 
 		redisValue, err := r.client.Get(ctx, redisKey).Result()
 		if err != nil {
+			if strings.Contains(err.Error(), redisWrongTypeIdentifyStr) {
```
shubham1172, line 246: @LaurenceLiZhixin should we add a unit test for this scenario?

LaurenceLiZhixin: Yes I added it at https://github.com/dapr/components-contrib/pull/1693/files#diff-8d2a56ddd092485603a1d00e47936c985c2baebae58884df2f0e51ea408d75b9R263
User can have any type of value in redis when using configuraiton api.
@shubham1172

shubham1172: Should it be a part of `setupMiniRedis`? Or should we add a new test scenario: `get does not throw error for wrong type`?

berndverst (changes requested):
@LaurenceLiZhixin please add the additional test scenario mentioned by @shubham1172 

A test that inserts an incompatible data type and then calls GetConfiguration. The test needs to verify no error was returned and the invalid data type was indeed skipped.

configuration/redis/redis.go
LaurenceLiZhixin: OK, I would fix it.

shubham1172: Thanks!

LaurenceLiZhixin:
Thanks for your advice, I have updated the code
@berndverst @shubham1172

configuration/redis/redis_test.go
```diff
@@ -118,9 +120,43 @@ func TestConfigurationStore_Get(t *testing.T) {
 			},
 			wantErr: true,
 		},
+		{
+			name: "test does not throw error for wrong type during get all",
```
shubham1172, line 124: The behavior introduced in this PR is - "Whenever a key type is unsupported, we log a warning and skip returning it". Is it being tested here? It looks like we are trying to get all the values and all of them are returned.

configuration/redis/redis_test.go
```diff
@@ -118,9 +120,43 @@ func TestConfigurationStore_Get(t *testing.T) {
...
+			prepare: func(client *redis.Client) {
+				client.HSet(context.Background(), "notSupportedType", []string{"key1", "value1", "key2", "value2"})
```
shubham1172, line 126: What does this line do?

LaurenceLiZhixin: This line writes an unsupported value to redis. After that, the test case try to get all values. and skip returning this unsupported value.

shubham1172: Got it. Since string slice is an unsupported type, should we change it to this to avoid confusion with keys and values?
```suggestion
				client.HSet(context.Background(), "testKeyUnsupported", []string{"value1", "value2", "value3"})
```

shubham1172: https://redis.io/docs/manual/data-types/ although it looks like lists are supported by Redis.

daixiang0 (approved):

beiwei30 (approved):
LGTM too.

PR #1695, opened 2022-04-29
Refactory kafka pubsub components for code reuse

skyao:
# Description

Refactory kafka pubsub components, extract the kafka common code from pubsub component and laster we can reuse them in kafka binding component.

This is just move code from kafka pubsub to internal/kafka and split the common kafka logic and pubsub component api, it should be safe from the view of functionality.

## Issue reference

https://github.com/dapr/components-contrib/issues/1305

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #1696, opened 2022-04-29
Refactory kafka binding to reuse the kafka common code extracting from kafka pubsub component

skyao:
# Description

As discussed in https://github.com/dapr/components-contrib/issues/1305, kafka binding should reuse the kafka common code extracting from kafka pubsub component, which will be updated in PR https://github.com/dapr/components-contrib/pull/1695.

This PR will replaced the implementation of kafka binding component to base on the code base of kafka pubsub component, so we need careful to handle the difference between the legacy kafka binding component  and kafka pubsub component.

## Issue reference

https://github.com/dapr/components-contrib/issues/1305

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


internal/component/kafka/metadata_test.go
```diff
@@ -0,0 +1,308 @@
+package kafka
```
artursouza, line 1: license headers.

bindings/kafka/kafka.go
```diff
@@ -15,321 +15,86 @@ package kafka
 
 import (
```
artursouza, line 16: I propose we split into v1 (old code) and v2 (this version). And we invest in making v2 stable and give users time to migrate from v1 to v2 before deprecating.

Can you change one of the apps in E2E tests to explain the breaking changes as per https://github.com/dapr/dapr/pull/4565 ?

artursouza (changes requested):

bindings/kafka/kafka.go
yaron2: @skyao if there any breaking changes to the spec, please point them out clearly and answer whether or not we can change this to not have breaking changes.

If these breaking changes (in either .spec.metadata scheme or request/response) cannot be made non-breaking, this must be a v2.

internal/component/kafka/metadata_test.go
skyao: OK, added.

skyao:
## kafka metadata

I list the metadata in pubsub component and binding component in following table:

|                     | pubsub metadata                          | comes from                                        | binding metadata                                | comes from                  | Compatibility         |
| ------------------- | ---------------------------------------- | ------------------------------------------------- | ----------------------------------------------- | --------------------------- | --------------------- |
|                     | ClientID             string              | metadata["clientID"]                              | none                                            |                             | NEW                   |
| topics to subscribe | none                                     | passed in SubscribeRequest.Topic                  | Topics          []string `json:"topics"`        | metadata["topics"]          | OK, binding extension |
| topic to publish    | none                                     | passed in PublishRequest.Topic                    | PublishTopic    string   `json:"publishTopic"`  | metadata["publishTopic"]    | OK, binding extension |
|                     |                                          |                                                   |                                                 |                             |                       |
|                     |                                          |                                                   |                                                 |                             |                       |
|                     | Brokers              []string            | metadata["brokers"]                               | Brokers         []string `json:"brokers"`       | metadata["brokers"]         | OK                    |
|                     | ConsumerGroup        string              | metadata["consumerID"], metadata["consumerGroup"] | ConsumerGroup   string   `json:"consumerGroup"` | metadata["consumerGroup"]   | OK                    |
|                     | InitialOffset        int64               | metadata["initialOffset"]                         | InitialOffset   int64    `json:"initialOffset"` | metadata["initialOffset"]   | OK                    |
|                     | MaxMessageBytes      int                 | metadata["maxMessageBytes"]                       | MaxMessageBytes int                             | metadata["maxMessageBytes"] | OK                    |
|                     | Version              sarama.KafkaVersion | metadata["version"]                               | Version         sarama.KafkaVersion             | metadata["version"]         | OK                    |
|                     | ConsumeRetryInterval time.Duration       | metadata["consumeRetryInterval"]                  | none                                            |                             | NEW                   |
|                     |                                          |                                                   |                                                 |                             |                       |
|                     |                                          |                                                   |                                                 |                             |                       |
|                     | deprecated                               | metadata["authRequired"]                          | AuthRequired    bool     `json:"authRequired"`  | metadata["authRequired"]    | OK, deprecated        |
|                     | AuthType             string              | metadata["authType"]                              | none                                            |                             | NEW                   |
| password            | SaslUsername         string              | metadata["saslUsername"]                          | SaslUsername    string   `json:"saslUsername"`  |                             | OK                    |
| password            | SaslPassword         string              | metadata["saslPassword"]                          | SaslPassword    string   `json:"saslPassword"`  |                             | OK                    |
| oidc                | OidcTokenEndpoint    string              | metadata["oidcTokenEndpoint"]                     | none                                            |                             | NEW                   |
| oidc                | OidcClientID         string              | metadata["oidcClientID"]                          | none                                            |                             | NEW                   |
| oidc                | OidcClientSecret     string              | metadata["oidcClientSecret"]                      | none                                            |                             | NEW                   |
| oidc                | OidcScopes           []string            | metadata["oidcScopes"]                            | none                                            |                             | NEW                   |
| mtls                | TLSDisable           bool                | metadata["disableTls"]                            | none                                            |                             | NEW                   |
| oidc                | TLSSkipVerify           bool             | metadata["skipVerify"]                            | none                                            |                             | NEW                   |
| oidc                | TLSCaCert            string              | metadata["caCert"]                                | none                                            |                             | NEW                   |
| mtls                | TLSClientCert        string              | metadata["clientCert"]                            | none                                            |                             | NEW                   |
| mtls                | TLSClientKey         string              | metadata["clientKey"]                             | none                                            |                             | NEW                   |

## Change to pubsub component

No change to pubsub component.

## Change to binding component

### ClientID

This metadata only exists in pubsub component.  This metadata is used to set the  ClientID field of sarama config:

```go
config := sarama.NewConfig()	
if meta.ClientID != "" {
    config.ClientID = meta.ClientID
}
```

It should be safe to add in binding component as new feature.

### topics

For the kafka topics to publish (output binding) or subscibe (input binding), in pubsub component they are coming from the request (SubscribeRequest.Topic and PublishRequest.Topic), in binding component they are coming from metadata (metadata["topics"] and metadata["publishTopic"]).

In this refactory, the common kafka  coding in `internal/component/kafka` will accept the topics as parameter of method:

```go
func (k *Kafka) Publish(topic string, data []byte, metadata map[string]string) error {}

func (k *Kafka) Subscribe(topics []string, _ map[string]string, handler EventHandler) error {}
```

In new (after refactory) pubsub and binding components, they will set these parameters from the request(pubsub component) or from the metadata (binding component). 

So we can keep these metadata in binding component and there is no change in new (after refactory)  binding component.

### consumeRetryInterval

In pubsub component, it supports retry and add new metadata "consumeRetryInterval". 

This can be considered as new feature in new (after refactory)  binding component.

### authentication

kafka binding support only password authentication

```go
	k.authRequired = meta.AuthRequired

	// ignore SASL properties if authRequired is false
	if meta.AuthRequired {
		k.saslUsername = meta.SaslUsername
		k.saslPassword = meta.SaslPassword
	}
```

kafka pubsub support 3 auth type:  

- oidc:  SASL OAuth2/OIDC authentication
- mtls: mTLS authentcation
- password: SASL Password authentication

```go
k.authType = meta.AuthType
switch k.authType {
	case oidcAuthType:
		err = updateOidcAuthInfo(config, meta)
	case passwordAuthType:
		k.saslUsername = meta.SaslUsername
		k.saslPassword = meta.SaslPassword
		updatePasswordAuthInfo(config, k.saslUsername, k.saslPassword)
	case mtlsAuthType:
		err = updateMTLSAuthInfo(config, meta)
	}
```

And in pubsub component, the upgradeMetadata() handles the "authRequired" metadata as deprecated: 

```go
// upgradeMetadata updates metadata properties based on deprecated usage.
func (k *Kafka) upgradeMetadata(metadata map[string]string) (map[string]string, error) {
	authTypeVal, authTypePres := metadata["authType"]
	authReqVal, authReqPres := metadata["authRequired"]
	saslPassVal, saslPassPres := metadata["saslPassword"]

	// If authType is not set, derive it from authRequired.
	if (!authTypePres || authTypeVal == "") && authReqPres && authReqVal != "" {
		k.logger.Warn("AuthRequired is deprecated, use AuthType instead.")
		validAuthRequired, err := strconv.ParseBool(authReqVal)
		if err == nil {
			if validAuthRequired {
				// If legacy authRequired was used, either SASL username or mtls is the method.
				if saslPassPres && saslPassVal != "" {
					// User has specified saslPassword, so intend for password auth.
					metadata["authType"] = passwordAuthType
				} else {
					metadata["authType"] = mtlsAuthType
				}
			} else {
				metadata["authType"] = noAuthType
			}
		} else {
			return metadata, errors.New("kafka error: invalid value for 'authRequired' attribute")
		}
	}

	return metadata, nil
}
```

So, for these authentication metadata from binding components, they are compatible with these from pubsub component, except the  "authRequired" metadata will be marked as deprecated.

## Conclusion

From the point of view of metadta, There is no broking changes to binding component in this refactory. 

Yet there are some compatible changes :

- new feature: add support for ClientID
- new feature: add support for consume retry
- new feature: add support for OAuth2/OIDC authentication and mTLS authentcation
- metadata "authRequired" is deprecated, use  "authType" instead.

skyao:
I will continue to check why  e2e tests failed.

skyao:
The first broking change found! 

In kafka binding component, if we don't want to do authentication for kafka, then we need to set authRequired to false:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: binding-topic
  namespace: default
spec:
  type: bindings.kafka
  version: v1
  metadata:
  - name: topics # Input binding topic
    value: binding-topic
  - name: brokers
    value: localhost:9092,localhost:9093
  - name: consumerGroup
    value: group1
  - name: publishTopic # Output binding topic
    value: binding-topic
  - name: authRequired
    value: "false"
  - name: initialOffset
    value: oldest
```

In kafka pubsub component, if we don't want to do authentication for kafka, then we need to set authRequired to false and set disableTls to true: 

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: pubsub
spec:
  type: pubsub.kafka
  version: v1
  metadata:
  - name: brokers
    value: localhost:9092
  - name: consumerGroup
    value: pubsubgroup1
  - name: authRequired
    value: "false"
  - name: initialOffset
    value: oldest
  - name: disableTls
    value: true
```

In kafka pubsub component, there is a updateTLSConfig() method which will check  metadata disableTls:

```go
func updateTLSConfig(config *sarama.Config, metadata *kafkaMetadata) error {
	if metadata.TLSDisable {
		config.Net.TLS.Enable = false
		return nil
	}
	config.Net.TLS.Enable = true
......
```

If we don't set disableTls to true, then dapr will fail to connect to kafka.

yaron2:
@skyao that's a great find!

I think what we need to do is actually remove the need for pub/sub to specify BOTH `authRequired: false` and `disableTLS: true`.

If `authRequired` is `false`, then `disableTLS` should be treated as `true` by the component. Setting `authRequired` as `true` and `disableTLS` as `false` is an invalid configuration.

What do you think? also cc @pkedy to keep my Kafka observations honest.

skyao:
Update:

Today I tried to fix the e2e test case fail. Now 226 tests passed, and 3 failures. 

It seems that there is no big problem for this refactory.

bindings/kafka/kafka.go
skyao: From the difference  found, it seems that it's possible to complete this refactory without introduce v2 api. 

Of cause we need to resolve these problems first:

1. pub/sub specify BOTH authRequired: false and disableTLS: true
2. risk of retry mechanism and app return fail for specified message. The API is not changed but the behavior after retry is not compatible.

skyao:
All e2e test passed, now I can summary this refactory:

### Compatible changes for kafka binding component 

- new feature: add support to set ClientID
- new feature: add support for consume retry
- new feature: add support for OAuth2/OIDC authentication and mTLS authentication
- metadata "authRequired" is deprecated, use "authType" instead.

### Broken changes for kafka binding component :

- To disable authentication, MUST add disableTLS = true, otherwise dapr will fail to connect to kafka. In fact, this change caused all old kafka binding test cases failed for kafka connection.

- Retry mechanism: if the app returns fail for specified message content, then dapr will work well without retry mechanism and run into endless loop with retry mechanism.

skyao:
@yaron2 @artursouza need your review to make decision.

artursouza:
> > To disable authentication, MUST add disableTLS = true, otherwise dapr will fail to connect to kafka. In fact, this change caused all old kafka binding test cases failed for kafka connection.
> 
> See my comment [here](https://github.com/dapr/components-contrib/pull/1696#issuecomment-1118914195).
> 
> Re: the retry issue, I believe we should _remove_ the component specific retries in the pub/sub component since:
> 
> 1. These are not needed for any reason related to bug workaround, SDK issue etc. It's just an extra resiliency feature that was baked into the component code.
> 2. We want to remove as many component specific retries as we can where possible
> 3. This doesn't constitute a breaking change

The removal of retries should be configurable and keep the current behavior. This way, we don't break people that actually use this feature.

Regarding the other breaking changes, I propose the following:
+ Keep support for "authRequired", so it does not break existing customers.
+ Parse "disableTLS" in a smart way where it is disabled automatically if auth is disabled.

yaron2:
Lets summarize then:

1. Keep `authRequired` and if set to `false`, do not require `disableTLS` to be set to `true`.
2. Refactor the common code to only include retries if used by a pub/sub component.

cc @skyao

skyao:
As suggested by @yaron2  and @artursouza :

- keep support for "authRequired" (deprecated, use "authType" instead)
- Parse "disableTLS" in a smart way where it is disabled automatically if auth is disabled.
- Keep `authRequired` and if set to `false`, do not require `disableTLS` to be set to `true`.
- Refactor the common code to only include retries if used by a pub/sub component.

I have updated the code of this refactory:

- if authentication is disabled, allways set config.Net.TLS.Enable to false

  So that we don't need to set  `disableTLS` to be set to `true` to disable authentication. This will remove the broken change in kafka binding component, and improve kafka pubsub component.

- update the common kafka comsume code, add support for disable consume retry

​	   "consumeRetryEnabled" property will used to enable consume retry, it will default to true for kafka pubsub component and default to false to kafka binding component, so that we will keep the same behavior after this refactory: retry enabled in  kafka pubsub component and no retry in kafka binding component.

​	  And introduced a new metadata named "consumeRetryEnabled" to overwrite this default behavior. The user can set "consumeRetryEnabled" metadata to false to disable retry in pubsub component, and set "consumeRetryEnabled" metadata to true to enable retry in binding component.

All e2e test passed, now I can summary this refactory again:

### Compatible changes for kafka pubsub component

- new feature: add support to disable comsume retry by set metadata "consumeRetryEnabled" to false
- To disable authentication,  do not require `disableTLS` to be set to `true`.

### Compatible changes for kafka binding component

- new feature: add support to set kafka ClientID
- new feature: add support for OAuth2/OIDC authentication and mTLS authentication
- metadata "authRequired" is deprecated, use "authType" instead.
- new feature: add support to enable comsume retry by set metadata "consumeRetryEnabled" to true

It is a good news that now after this adjustment, there is no broken changes after refactory, thus we don't need to introduce v2 of kafka binding component.

artursouza:
> * new feature: add support to enable comsume retry by set metadata "consumeRetryEnabled" to true

Great, please, create an issue in dapr/docs to document these changes in the components.

artursouza (approved):

PR #1685, opened 2022-04-25
Use revive instead of golint

pigletfly:
Signed-off-by: pigletfly <wangbing.adam@gmail.com>

# Description

The linter 'golint' is deprecated (since v1.41.0).ref: https://github.com/golangci/golangci-lint/issues/1892

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #_[issue number]_

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #1697, opened 2022-04-29
Updated to Go 1.18

ItalyPaleAle:
This PR updates Go to 1.18 in all places, including setting the minimum version in `go.mod` files and in all the CI's and containers.

A number of dependencies, most notably the newest Azure SDKs, require Go 1.18 (because of generics) and we will need to update the version of Go used by Dapr to be able to rely on them.


berndverst (changes requested):
What is causing the current linter failures in this PR? Is there a PR that needs to be merged to address this?

ItalyPaleAle:
Yes this one: https://github.com/dapr/components-contrib/pull/1685

Note that you should not merge this PR (in components-contrib) until the matching PR in dapr/dapr is merged (which is blocked by changing the linter there too). Otherwise dapr/dapr will fail to build if this package requires a newer version of Go.

ItalyPaleAle:
I've also added `go.work` and the respective `go.work.sum` files. New with go 1.18 is that gopls can use the `go.work` file to provide language server capabilities when there are more than one `go.mod` files in a workspace. This fixes a longstanding issue in VS Code where (and other editors that use gopls) where IntelliSense is not working in test apps and other folders that have their own `go.mod`.

tests/e2e/pubsub/jetstream/docker-compose.yaml
```diff
@@ -9,7 +9,7 @@ services:
       - ./jwt:/jwt
 
   dev:
-    image: golang:1.17.6
+    image: golang:1.18
```
daixiang0, line 12: Should pin a specific version?

ItalyPaleAle: This pins to 1.18 but allows patch updates.

berndverst: @daixiang0 this should be ok -- patch version changes are fine

berndverst (approved):

PR #1700, opened 2022-05-03
WIP - Azure and AWS Bindings with Azure Auth

acartcat:
# Description

Testing the waters here to see how to get some changes in:
1.  The change to Azure Auth allows the Default Credential to move through the chain of potential creds including the Azure CLI that is useful from the developer desktop
2.  For AWS SQS queue it is possible to have a cred that allows reading of the queue but not use of the method GetQueueURL on the AWS Client.  The change enables the use of the full URL in the component spec of the binding.
3.  For Azure Service Bus binding the send and receive will look to see if the message has populated user properties in addition to the standard system properties.

## Issue reference

1.  Looks like this is in flux with the move to the new Azure SDK for GO
2.  Addresses an issue I have in a live scenario
3.  same as 2

I have (2) and (3) running in an AKS cluster with a custom build of Dapr and seems to be working well

## Checklist
Test changes are a bit limited and no document changes at this time.
Happy to discuss

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/authentication/azure/auth.go
+++ b/authentication/azure/auth.go
@@ -129,6 +129,23 @@ func (s EnvironmentSettings) GetTokenCredential() (azcore.TokenCredential, error
 		}
 	}
 
+	// 4. MSI using Default Credential
+	// DefaultAzureCredential is a default credential chain for applications that will be deployed to Azure.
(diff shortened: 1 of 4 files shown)


acartcat:
Wrong account :-(

PR #1701, opened 2022-05-03
WIP - Azure and AWS Bindings - Azure Auth MSI

HarryAcker:
# Description

Testing the waters here to see how to get some changes in:

The change to Azure Auth allows the Default Credential to move through the chain of potential creds including the Azure CLI that is useful from the developer desktop
For AWS SQS queue it is possible to have a cred that allows reading of the queue but not use of the method GetQueueURL on the AWS Client. The change enables the use of the full URL in the component spec of the binding.
For Azure Service Bus binding the send and receive will look to see if the message has populated user properties in addition to the standard system properties.

## Issue reference

1. Looks like this is in flux with the move to the new Azure SDK for GO
2. Addresses an issue I have in a live scenario
3. same as 2

I have (2) and (3) running in an AKS cluster with a custom build of Dapr and seems to be working well

## Checklist

Test changes are a bit limited and no document changes at this time.
Happy to discuss

* [X] Code compiles correctly
* [X] Created/updated tests


bindings/aws/sqs/sqs.go
```diff
@@ -60,15 +61,19 @@ func (a *AWSSQS) Init(metadata bindings.Metadata) error {
...
 	}
 
-	queueName := m.QueueName
-	resultURL, err := client.GetQueueUrl(&sqs.GetQueueUrlInput{
-		QueueName: aws.String(queueName),
-	})
-	if err != nil {
-		return err
-	}
+	if m.QueueURL != "" {
```
yaron2, line 64: @amimimor can you please review the SQS portion regarding the addition of `queueURL`? Thanks!

authentication/azure/auth.go
```diff
@@ -129,6 +129,23 @@ func (s EnvironmentSettings) GetTokenCredential() (azcore.TokenCredential, error
...
+	// 4. MSI using Default Credential
+	// DefaultAzureCredential is a default credential chain for applications that will be deployed to Azure.
+	// It combines credentials suitable for deployed applications with credentials suitable in local development.
+	// It attempts to authenticate with each of these credential types, in the following order:
+	//	- EnvironmentCredential
+	//	- ManagedIdentityCredential
+	//	- AzureCLICredential
+	// Consult the documentation for these credential types for more information on how they authenticate.
+	{
+		cred, err := azidentity.NewDefaultAzureCredential(nil)
```
yaron2, line 141: @berndverst can you review this `auth` part please?

yaron2:
Thanks for this contribution @HarryAcker! I pinged a coupe of folks to start reviewing this.

authentication/azure/auth.go
yaron2: or @ItalyPaleAle

ItalyPaleAle:
Hi @HarryAcker  thanks a lot for this contribution.

I have two thoughts on this:

**First**

WRT to the thing @yaron2 specifically @ me for, I agree that supporting Azure CLI credentials would be really helpful especially during development. There are however a few issues with this PR:

1. The `authentication/azure` package currently must provide authentication tokens for 2 different versions of the Azure SDKs. `GetTokenCredential()`, the method you modified, provides the credentials for the "track2" SDKs only (these are the new SDKs which are not yet available for all Azure services). Whatever solution is implemented there, needs to be implemented in `GetAuthorizer()` too, which is for the "track1" (older) SDKs.
2. Even in the case of the "track2" auth code, I would prefer not to rely on `NewDefaultAzureCredential` because that has its own pipeline which has some duplication (it again reads from the env and from MSI).  Instead, I'd add to the auth pipeline just `NewAzureCLICredential` to support the missing CLI credentials.
3. What I'm not 100% sure about is whether the credentials from the Azure CLI have a wide enough scope however. Those will definitely work for ARM calls (which I don't believe we make any), but will not work for things like Cosmos DB, because in that case you need a principal which is explicitly authorized to access your database. I am not sure they will work with Key Vault too.

**Second**

This PR is quite big in scope, and it would be great if you could please split it in 3:

- The work on AWS binding
- The work on the Azure Service Bus binding (which note that I'm rewriting literally these days to upgrade to the track2 SDK, so I can incorporate that change but it would need to be reviewed and merged separately!)
- The part around authenticating with Azure

This would make it much easier to review the changes since they're quite different from each other :)

bindings/aws/sqs/sqs.go
amimimor: Sorry to spoil a little here:
I don't see a lot of value passing in the full queue URL as it can be built based on the region+accountID+queueName so the call to AWS SQS might be expensive here and could be replaced by some sorts of string builder (see: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-queue-message-identifiers.html)
(as with pubsub sns/sqs) it needs to be documented that using accesskey and secretkey need not to be used if using AWS node/pod roles
regarding the above @yaron2 waiting for your input

@yaron2 I think there is a lot to be done here in terms of using contexts, configured retries and durations and in general I would have felt more comfortable seeing that this component somehow aligns (I know it depends on community contribution...) with pubsub sns/sqs

HarryAcker:
All good.  Firstly I'll move these out to three separate PRs.
AWS Binding - My problem with what is currently implemented is the call to client.GetQueueUrl requires a specific level of auth which the credentials may not allow.  I'll stick with my custom build and take a look at the pub/sub implementation.
Azure Binding - Sounds like the move to the new Azure SDK will carry some significant change so will take a look at a track 2 implementation.
Azure Auth - I'll switch to adding NewAzureCLICredential and validate it works for me in my dev and test environment.
In general appreciate the quick feedback and must say this being my first serious attempt at contributing to an OSS project.. it's going to be fun!

authentication/azure/auth.go
berndverst: This does not look right at all. I understand what the DefaultAzureCredential does but you just duplicated a bunch of things the code already supports today.

Whether we should the Azure CLI credential here needs to be discussed, but if so that should be the only change -- adding support for the Azure CLI credential.

@ItalyPaleAle knows this code much better so he should review this.

berndverst (changes requested):

berndverst:
Try to keep your PRs as self contained as possible. Multiple PRs are fine - but only change one component at a time if at all possible :)

onionhammer:
This may not be the way to do it, but is anyone going to propose an alternative that resolves the underlying need for local development?

berndverst:
Please file a new issue @onionhammer or discuss in our discord. It's unclear what need you are speaking of, and a closed PR isn't the place to discuss.

PR #1648, opened 2022-04-06
WIP: Move Azure Service Bus to track2 sdk

halspang:
# Description

Move Azure Service Bus to track2 SDK. The original SDK that we're using is deprecated, we should move to the track2 SDK to get the latest and greatest in features/durability/security.

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #1532 #1530 

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #1702, opened 2022-05-04
Update Azure Service Bus components to track2 SDK

### LINKED ISSUE #1531: [Azure][Track 2 SDK] Migrate Azure Service Bus Queues Binding
Migrate Azure Service Bus Queues Binding to SDK github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus

### LINKED ISSUE #1532: [Azure][Track 2 SDK] Migrate Azure Service Bus PubSub
Migrate Azure Service Bus PubSub to SDK github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus

ItalyPaleAle:
# Description

Move the Azure Service Bus components (binding and pubsub) to the new "track2" SDKs. Thanks a lot to @halspang for the initial work in #1648!

Thanks to the update and some refactoring work, two metadata keys for Azure Service Bus Pub/Sub are not needed anymore:

- `prefetchCount`
- `maxActiveMessagesRecoveryInSec`

Notes:

1. Because the Service Bus SDKs depended on the newer version of azcore and azidentity, I had to update the other Azure SDKs that were already in "track2". Some of the updates contained breaking changes in the APIs, which I've fixed.
1. ~~This requires Go 1.18 so it depends on #1697 (which depends on #1685 itself)~~

## Issue reference

Please reference the issue this PR will:
closes #1532 
closes #1531
closes #1701


pubsub/azure/servicebus/servicebus.go
```diff
@@ -489,7 +487,6 @@ func (a *azureServiceBus) Subscribe(req pubsub.SubscribeRequest, handler pubsub.
 			attempts := readAttemptsStale()
 			if attempts == 0 {
 				a.logger.Errorf("Subscription to topic %s lost connection, unable to recover after %d attempts", sub.topic, a.metadata.MaxReconnectionAttempts)
```
ItalyPaleAle, line 489: This is unchanged, but wondering if we should make the app crash here (`Fatalf`)? Otherwise, the error would get logged and the app continues to work but without a connection to Service Bus.

pubsub/azure/servicebus/message_test.go
```diff
@@ -121,108 +115,17 @@ func TestNewASBMessageFromPubsubRequest(t *testing.T) {
...
-				assert.Equal(t, tc.expectedAzServiceBusMessage.Label, msg.Label)
-				assert.NotNil(t, msg.SystemProperties)
-				assert.Equal(t, tc.expectedAzServiceBusMessage.SystemProperties.PartitionKey, msg.SystemProperties.PartitionKey)
-				assert.Equal(t, tc.expectedAzServiceBusMessage.SystemProperties.ScheduledEnqueueTime.Unix(), msg.SystemProperties.ScheduledEnqueueTime.Unix())
-			}
-		})
-	}
-}
-
-func TestNewPubsubMessageFromAzServiceBusMessage(t *testing.T) {
```
ItalyPaleAle, line 141: This test can't be implemented with track2 SDKs. The method to create these objects is un-exported.

Can be re-added if https://github.com/Azure/azure-sdk-for-go/pull/17848 is merged

berndverst:
@ItalyPaleAle please fix linter issues. Also use `/ok-to-test` to trigger actions for conformance and certification tests (separately from this PR) to run the Service Bus Queues binding certification test.

berndverst:
At the moment this breaks EventHubs Binding, PubSub and Service Bus Queues Binding.

Please take a look by running the certification and conformance tests via `/ok-to-test`

ItalyPaleAle:
@berndverst This is ready for review IMHO.

- All tests are passing (waiting for the conformance tests to complete)
- I have confirmed that both the binding and pubsub are able to recover successfully in case of failures (both pubsub and input/output bindings)

Once this is merged, I'll need to update the docs because of metadata changes, as well as our E2E tests if needed.
-

berndverst:
Eventhubs Binding conformance test is failing
Eventhubs Binding certification test is failing
Service Bus Queues Binding certification test is failing

(you can see in the results of your /ok-to-test run)

ItalyPaleAle:
I'll look into the Service Bus binding failure tomorrow.

For Event Hubs, that's very odd because I did not touch that module besides for a minimal change that was required. However, it looked like the SDK got upgraded (which was un-intentional). I've reverted the change to the SDK and let's see if that fixes the tests.

ItalyPaleAle:
@berndverst  I have fixed all tests that were failing because of changes in my code.

After fixing dependencies, there were 2 tests that fail.

- Service Bus Queue binding: when a message fails to be processed by a handler, it's inserted back into the queue. This means that it's re-delivered later, so it may not come in order. However the conf tests were expecting messages to be in order. I think this was a problem with the tests as the re-enqueueing of the message is arguably the desired behavior
- Event Hubs binding: this is still failing and it looks like it was failing before 100% of the times when running on master. See: https://github.com/dapr/components-contrib/actions/workflows/certification.yml The error in this PR is the same. The test is broken and needs to be fixed, but it should not block this PR.

(On why the Event Hubs test is failing: the test `TestAzureServiceBusQueueRetriesOnError` is creating 2 Event Hubs listeners with the same consumer group, which seems to be disallowed)

berndverst (approved):

Issue #1671, opened 2022-04-15
[pubsub-pulsar] Add metadata property to configure BatchingMaxSize in Pulsar


saber-wang:
Publishing failed
```
time="2022-04-15T01:52:07Z" level=error msg="[unable to add message to batch]" producerID=8 producer_name=pulsar-cluster-zk-3-2988 properties="map[]" size=270636 topic="persistent://public/default/testEvent"
time="2022-04-15T01:52:07.670125177Z" level=debug msg="rpc error: code = Internal desc = error when publish to topic testEvent in pubsub pubsub: message add to batch failed: AddToBatchFailed" app_id=test instance=test-1.2.65-5f89f79776-bkpxf scope=dapr.runtime.grpc.api type=log ver=1.7.0
```

berndverst:
Seems related to https://github.com/dapr/components-contrib/issues/1659 which is fixed by PR https://github.com/dapr/components-contrib/pull/1661/files

We will be releasing Dapr 1.7.1 shortly for an unrelated reason and will include this fix there.

saber-wang:
@berndverst  https://github.com/apache/pulsar-client-go/blob/7773d27562cec30687879dc7ce240e0a02ec7a42/pulsar/producer.go#L146

BatchingMaxSize specifies the maximum number of bytes permitted in a batch. (default 128 KB) ,the message is greater than 128, so sending failed. should add parameters to support setting this value

shubham1172:
@saber-wang can you please also update the issue title to reflect the same, e.g. "Add metadata property to configure BatchingMaxSize in Pulsar". Please feel free to take a stab at this issue.

PR #1707, opened 2022-05-05
Add metadata property to configure Batching in Pulsar

saber-wang:
## Issue reference

Please reference the issue this PR will close: #1671

## Checklist

* [√] Code compiles correctly
* [√] Created/updated tests
* [√] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#2426


pubsub/pulsar/pulsar.go
```diff
@@ -30,15 +30,17 @@ import (
...
-	tenant          = "tenant"
-	namespace       = "namespace"
-	persistent      = "persistent"
+	host                = "host"
+	consumerID          = "consumerID"
+	enableTLS           = "enableTLS"
+	deliverAt           = "deliverAt"
+	deliverAfter        = "deliverAfter"
+	disableBatching     = "disableBatching"
+	batchingMaxSize     = "batchingMaxSize"
```
shubham1172, line 39: nit: order can be same in this struct and metadata.go

shubham1172 (dismissed):
LGTM!

shubham1172 (approved):

pubsub/pulsar/pulsar_test.go
```diff
@@ -37,6 +39,8 @@ func TestParsePulsarMetadata(t *testing.T) {
 	assert.Equal(t, true, meta.DisableBatching)
 	assert.Equal(t, defaultTenant, meta.Tenant)
 	assert.Equal(t, defaultNamespace, meta.Namespace)
+	assert.Equal(t, 100, meta.BatchingMaxSize)
```
shubham1172, line 42: The tests are failing, the parsing logic does not seem to be working. Can you please take a look @saber-wang ?

shubham1172 (changes requested):
Unit tests failing.

pubsub/pulsar/pulsar_test.go
saber-wang: @shubham1172  Execute make test and throw "make: no rule to make target 'test'. Stop" exception. How should I handle it?

shubham1172: That is strange, the Makefile has the `test` command. What environment are on?

saber-wang: @shubham1172 
 environment : win11 64bit, PowerShell 7.2.2,Make 4.3
I passed the test with "go test - V"

shubham1172 (dismissed):

shubham1172:
Thank you for the contribution, @saber-wang 🥳!

.gitignore
```diff
@@ -3,4 +3,5 @@
 .vscode
 /vendor
 go.work
-.dccache
\ No newline at end of file
+.dccache
+/.vs
```
daixiang0, line 7: Theme unrelated changes, please remove.

pubsub/pulsar/pulsar.go
```diff
@@ -92,7 +94,22 @@ func parsePulsarMetadata(meta pubsub.Metadata) (*pulsarMetadata, error) {
 		}
 		m.DisableBatching = disableBatching
 	}
-
+	m.BatchingMaxMessages = 1000
```
daixiang0, line 97: Please add TODO and document those hardcode.

saber-wang: I looked at the pulasr client's API and can remove these hard coded code

.history/pubsub/pulsar/pulsar_20220505155033.go
```diff
@@ -0,0 +1,340 @@
+/*
```
daixiang0, line 1: Please remove changes about `.history`.

.gitignore
```diff
@@ -3,4 +3,3 @@
 .vscode
 /vendor
 go.work
-.dccache
```
shubham1172, line 6: Changes still there.

shubham1172 (approved):

pubsub/pulsar/pulsar.go
```diff
@@ -92,7 +94,20 @@ func parsePulsarMetadata(meta pubsub.Metadata) (*pulsarMetadata, error) {
 		}
 		m.DisableBatching = disableBatching
 	}
-
+	if val, ok := meta.Properties[batchingMaxMessages]; ok {
```
shubham1172, line 97: We don't have a default value here, and we rely on the upstream for it. So say if the default value of `bachingMaxMessages` get changed in the upstream, this will start using that on its own. 

However, if this value changes, our docs will be stale, since we are defining the default values there - https://github.com/dapr/docs/pull/2426/

Should we also hard code the defaults here to avoid that? Or instead of mentioning the default value in the docs, should we point to the upstream documentation? (I would prefer the former because it'll be less confusing for users). 

@saber-wang @daixiang0

saber-wang: Hi, how can I format something like "5S" as "time. Duration"?

daixiang0: Let's hardcode here and document it clearly.

shubham1172: > And how can I format something like "5s" as "time. Duration"?

Sorry I did not understand this, can you explain more on what are you trying to do?

saber-wang: @shubham1172  I have found `time.ParseDuration(val)` 😂.

shubham1172 (changes requested):
As discussed, waiting for settings to be hardcoded.

pubsub/pulsar/pulsar.go
```diff
@@ -317,3 +352,14 @@ func (p *Pulsar) formatTopic(topic string) string {
 	}
 	return fmt.Sprintf(topicFormat, persist, p.metadata.Tenant, p.metadata.Namespace, topic)
 }
+
+func formatDuration(durationString string) (time.Duration, error) {
```
saber-wang, line 356: @shubham1172 Is there a better way?

pubsub/pulsar/pulsar.go
```diff
@@ -50,6 +53,12 @@ const (
 	topicFormat      = "%s://%s/%s/%s"
 	persistentStr    = "persistent"
 	nonPersistentStr = "non-persistent"
+	// defaultBatchingMaxPublishDelay init default for maximum delay to batch messages
```
shubham1172, line 56: Please fix linting issues.

pubsub/pulsar/pulsar.go
```diff
@@ -50,6 +53,12 @@ const (
...
+	defaultBatchingMaxPublishDelay = 10 * time.Millisecond
```
shubham1172, line 57: 1ms, same as upstream: https://pulsar.apache.org/docs/en/client-libraries-go/

pubsub/pulsar/pulsar.go
shubham1172: We can make it simpler by always expecting a number. 

The metadata can be `batchingMaxPublishDelayMs` to clearly indicate that it should be in milliseconds.

pubsub/pulsar/metadata.go
```diff
@@ -13,13 +13,18 @@ limitations under the License.
...
-	Token           string `json:"token"`
+	Host                    string        `json:"host"`
+	ConsumerID              string        `json:"consumerID"`
+	EnableTLS               bool          `json:"enableTLS"`
+	DisableBatching         bool          `json:"disableBatching"`
+	Tenant                  string        `json:"tenant"`
+	Namespace               string        `json:"namespace"`
+	Persistent              bool          `json:"persistent"`
+	Token                   string        `json:"token"`
+	BatchingMaxPublishDelay time.Duration `json:"batchingMaxPublishDelay"`
```
shubham1172, line 27: Let's add the batching metadata right next to disable batching (since they are inter related). Same ordering for constants in `pulsar.go`.

pubsub/pulsar/pulsar.go
saber-wang: @shubham1172  Actually 10ms https://github.com/apache/pulsar-client-go/blob/5ee63303d43e8ee05f2c957a32c5526f29f54a36/pulsar/producer_impl.go#L38

berndverst (approved):

pubsub/pulsar/pulsar.go
shubham1172: @berndverst this comment was still pending.

saber-wang: @shubham1172 Need to change to such a design?

shubham1172: That would have been cleaner and simpler but Bernd has already approved/merged this PR. If we have enough consensus, we can create another PR on top to change that I guess.

saber-wang: @shubham1172  I don't have a better idea. I can create a new rp to modify it

shubham1172: Thanks @saber-wang. Let's see what Bernd/@daixiang0 have to say.

PR #1708, opened 2022-05-06
Add context in state component

pigletfly:
Signed-off-by: pigletfly <wangbing.adam@gmail.com>

# Description

add context in state interface:

```
type Store interface {
	BulkStore
	Init(metadata Metadata) error
	Features() []Feature
	Delete(ctx context.Context, req *DeleteRequest) error
	Get(ctx context.Context, req *GetRequest) (*GetResponse, error)
	Set(ctx context.Context, req *SetRequest) error
	Ping(ctx context.Context) error
}
```

## Issue reference

https://github.com/dapr/dapr/issues/2716
https://github.com/dapr/components-contrib/issues/1601

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/state/alicloud/tablestore/tablestore_test.go
+++ b/state/alicloud/tablestore/tablestore_test.go
@@ -14,6 +14,7 @@ limitations under the License.
 package tablestore
 
 import (
+	"context"
 	"testing"
 
 	"github.com/agrea/ptr"
@@ -63,7 +64,7 @@ func TestReadAndWrite(t *testing.T) {
 			Value: "value of key",
 			ETag:  ptr.String("the etag"),
 		}
-		err := store.Set(setReq)
+		err := store.Set(context.Background(), setReq)
 		assert.Nil(t, err)
 	})
 
@@ -71,7 +72,7 @@ func TestReadAndWrite(t *testing.T) {
 		getReq := &state.GetRequest{
 			Key: "theFirstKey",
 		}
-		resp, err := store.Get(getReq)
+		resp, err := store.Get(context.Background(), getReq)
 		assert.Nil(t, err)
 		assert.NotNil(t, resp)
 		assert.Equal(t, "value of key", string(resp.Data))
@@ -83,7 +84,7 @@ func TestReadAndWrite(t *testing.T) {
 			Value: "1234",
 			ETag:  ptr.String("the etag"),
 		}
-		err := store.Set(setReq)
+		err := store.Set(context.Background(), setReq)
 		assert.Nil(t, err)
 	})
 
@@ -91,14 +92,14 @@ func TestReadAndWrite(t *testing.T) {
 		getReq := &state.GetRequest{
 			Key: "theSecondKey",
 		}
-		resp, err := store.Get(getReq)
+		resp, err := store.Get(context.Background(), getReq)
 		assert.Nil(t, err)
 		assert.NotNil(t, resp)
 		assert.Equal(t, "1234", string(resp.Data))
 	})
 
 	t.Run("test BulkSet", func(t *testing.T) {
-		err := store.BulkSet([]state.SetRequest{{
+		err := store.BulkSet(context.Background(), []state.SetRequest{{
 			Key:   "theFirstKey",
 			Value: "666",
 		}, {
@@ -110,7 +111,7 @@ func TestReadAndWrite(t *testing.T) {
 	})
 
 	t.Run("test BulkGet", func(t *testing.T) {
-		_, resp, err := store.BulkGet([]state.GetRequest{{
+		_, resp, err := store.BulkGet(context.Background(), []state.GetRequest{{
 			Key: "theFirstKey",
 		}, {
 			Key: "theSecondKey",
@@ -126,12 +127,12 @@ func TestReadAndWrite(t *testing.T) {
 		req := &state.DeleteRequest{
 			Key: "theFirstKey",
 		}
-		err := store.Delete(req)
+		err := store.Delete(context.Background(), req)
 		assert.Nil(t, err)
 	})
 
 	t.Run("test BulkGet2", func(t *testing.T) {
-		_, resp, err := store.BulkGet([]state.GetRequest{{
+		_, resp, err := store.BulkGet(context.Background(), []state.GetRequest{{
 			Key: "theFirstKey",
 		}, {
 			Key: "theSecondKey",
--- a/state/aerospike/aerospike.go
+++ b/state/aerospike/aerospike.go
@@ -14,6 +14,7 @@ limitations under the License.
 package aerospike
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -110,7 +111,7 @@ func (aspike *Aerospike) Features() []state.Feature {
 }
 
 // Set stores value for a key to Aerospike. It honors ETag (for concurrency) and consistency settings.
-func (aspike *Aerospike) Set(req *state.SetRequest) error {
+func (aspike *Aerospike) Set(ctx context.Context, req *state.SetRequest) error {
 	err := state.CheckRequestOptions(req.Options)
 	if err != nil {
 		return err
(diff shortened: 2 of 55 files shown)


berndverst (changes requested):
Please address linter failures. You may want to run the linter locally to speed up addressing linter issues.

pigletfly:
sure,can we merge https://github.com/dapr/components-contrib/pull/1685 first?

state/alicloud/tablestore/tablestore_test.go
```diff
@@ -63,15 +64,15 @@ func TestReadAndWrite(t *testing.T) {
 			Value: "value of key",
 			ETag:  ptr.String("the etag"),
 		}
-		err := store.Set(setReq)
+		err := store.Set(context.TODO(), setReq)
```
ItalyPaleAle, line 67: In all tests, I'd use `context.Background()` as not setting a context is intentional.

The idea of `context.TODO()` is that eventually needs to be replaced (and when you perform a search for `TODO` in a repo, they appear).

ItalyPaleAle (commented):
I think this is a great contribution, using contexts greatly increases the reliability of code.

My concern is that I see that in many, if not most, components the context is passed but not actually used. If your method accepts a context, it's sort of expected that it will rely on it (ie. if the context is canceled, the method should return).

pigletfly:
> I think this is a great contribution, using contexts greatly increases the reliability of code.
> 
> My concern is that I see that in many, if not most, components the context is passed but not actually used. If your method accepts a context, it's sort of expected that it will rely on it (ie. if the context is canceled, the method should return).

yes, it depends on third-party implements, if the SDK doesn't accept a `context`, then `context` is useless actually.

ItalyPaleAle:
Well, it just needs to be implemented in a SDK-by-SDK way.

For example, for aerospike (just because it's the first one in alphabetic order), you can see a way to halt the request on a context cancellation here: https://github.com/aerospike/aerospike-client-go/issues/255#issue-416695859

For AWS Dynamo DB, the methods need to be changed to add the `WithContext` variant. For example `GetItem` -> `GetItemWIthContext`; `PutItem` -> `PutItemWithContext`

pigletfly:
> Well, it just needs to be implemented in a SDK-by-SDK way.
> 
> For example, for aerospike (just because it's the first one in alphabetic order), you can see a way to halt the request on a context cancellation here: [aerospike/aerospike-client-go#255 (comment)](https://github.com/aerospike/aerospike-client-go/issues/255#issue-416695859) The same technique could be used in other SDKs that don't implement context support at all.
> 
> For AWS Dynamo DB, the methods need to be changed to add the `WithContext` variant. For example `GetItem` -> `GetItemWIthContext`; `PutItem` -> `PutItemWithContext`

now if the third-party SDK has used `context` for cancellation, I just added the `context` in the calling method, if not, we should wait for the SDK support, is that right ? @ItalyPaleAle

ItalyPaleAle:
@pigletfly

Note that some SDKs have context support but the method is named different. For example, for AWS Dynamo DB, you have to add `WithContext`... For example, `GetItem` becomes `GetItemWithContext`. This may be the case for other SDKs.

Also please see my comment in the code. In tests (files ending in `_tests.go`) I think we should NOT use `context.TODO()` but rather `context.Background()` because it's not a "TODO" to fix in the future, but an actual choice.

yaron2:
@pigletfly please fix the failing tests.

yaron2:
ping @pigletfly

pigletfly:
I don't know why  some of the conformance tests failed with no logs.

pigletfly:
The left failing tests are caused by the intermediate state.After this PR is merged, I will create a separate PR in dapr/dapr as well. @yaron2 @berndverst

ItalyPaleAle:
@pigletfly can you please address my review's comments too?

yaron2:
ping @pigletfly

pigletfly:
> @pigletfly with "address" I am asking that you please fix the issues, not just "resolve" the comments :)
> 
> 1. Please make sure that the context is actually used in all components. I have pointed out various examples of SDKs where method names need to be changed so they use a context. Without that work, this PR is incomplete and is actually giving maintainers more work in the future
> 2. Please do not use context.TODO() in tests but use context.Background() instead. "TODO" is used when you are saying that in the future that needs to be changed

should we send PR to corresponding sdk repo?Take aerospike for an example, https://github.com/dapr/components-contrib/blob/1af102f6f3dd3c96a0594e78e196e1917546904d/state/aerospike/aerospike.go#L165-L180 , we add `ctx` in Get method, but aspike.client.Get doesn't accept ctx. @ItalyPaleAle

berndverst:
@pigletfly just try your best to use ctx where available, and maybe we can fix upstream components in subsequent changes.

To avoid breaking tests, can you please add temporary methods that keep the old signature?

You added:
```
func (s *AliCloudTableStore) BulkGet(ctx context.Context, reqs []state.GetRequest) (bool, []state.BulkGetResponse, error)
```

but we also should add
```
func (s *AliCloudTableStore) BulkGet(reqs []state.GetRequest) (bool, []state.BulkGetResponse, error) {
  return s.BulkGet(ctx.Background(), reqs)

}
```

This way we can avoid breaking certification tests. Please do something like this for every state store component where you added context.

pigletfly:
> @pigletfly just try your best to use ctx where available, and maybe we can fix upstream components in subsequent changes.
> 
> To avoid breaking tests, can you please add temporary methods that keep the old signature?
> 
> You added:
> 
> ```
> func (s *AliCloudTableStore) BulkGet(ctx context.Context, reqs []state.GetRequest) (bool, []state.BulkGetResponse, error)
> ```
> 
> Maybe can name this `BulkGetWithContext` for now?
> 
> and then we also should add
> 
> ```
> func (s *AliCloudTableStore) BulkGet(reqs []state.GetRequest) (bool, []state.BulkGetResponse, error) {
>   return s.BulkGetWithContext(ctx.Background(), reqs)
> 
> }
> ```
> 
> This way we can avoid breaking certification tests. Please do something like this for every state store component where you added context.
> 
> Eventually we can then remove the old method and make the new method the default.
> 
> The best recommendation will be that we rename all methods to `*WithContext` and keep the old methods to call `*WithContext` setting the ctx to Background. Then we can import the latest contrib into dapr/dapr and update the interface to use the `*WithBackground` methods instead. Once that is merged we can update the certification tests in `contrib` and remove the old methods (the ones without `*WithContext`) at that time.
> 
> This is the only way to avoid introducing test failures - which we do not want to do anymore.

Good idea !!!

berndverst:
> > @pigletfly just try your best to use ctx where available, and maybe we can fix upstream components in subsequent changes.
> > To avoid breaking tests, can you please add temporary methods that keep the old signature?
> > You added:
> > ```
> > func (s *AliCloudTableStore) BulkGet(ctx context.Context, reqs []state.GetRequest) (bool, []state.BulkGetResponse, error)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Maybe can name this `BulkGetWithContext` for now?
> > and then we also should add
> > ```
> > func (s *AliCloudTableStore) BulkGet(reqs []state.GetRequest) (bool, []state.BulkGetResponse, error) {
> >   return s.BulkGetWithContext(ctx.Background(), reqs)
> > 
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This way we can avoid breaking certification tests. Please do something like this for every state store component where you added context.
> > Eventually we can then remove the old method and make the new method the default.
> > The best recommendation will be that we rename all methods to `*WithContext` and keep the old methods to call `*WithContext` setting the ctx to Background. Then we can import the latest contrib into dapr/dapr and update the interface to use the `*WithBackground` methods instead. Once that is merged we can update the certification tests in `contrib` and remove the old methods (the ones without `*WithContext`) at that time.
> > This is the only way to avoid introducing test failures - which we do not want to do anymore.
> 
> Good idea !!!

@pigletfly could you give this a try after we complete the 1.8 release? At this point we won't include it in this release.

berndverst:
@pigletfly do you have time to work on this?

Please update the various function to accept a context parameter. Then also create another function with the old signature so you don't break certification tests.

Example:
Update `func (aspike *Aerospike) Set(req *state.SetRequest) error` to `func (aspike *Aerospike) Set(ctx context.Context, req *state.SetRequest) error`

Then also add a new function with the old signature which calls the new function with context but passes in `context.Todo()` -- in this case Todo is fine because we will be deleting that function eventually.

```
func (aspike *Aerospike) Set(req *state.SetRequest) error {
    return aspike.Set(context.Todo(), req)
}
```

PR #1715, opened 2022-05-10
Initial support for CloudEvents binary mode

vangheem:
# Description

The purpose of this PR is to provide support for binary mode CloudEvents as described in this issue: https://github.com/dapr/dapr/issues/4588

Right now, this is mostly a POC. I would like more feedback and would like to make sure the project is interested in supporting this feature and okay with this approach.

## Issue reference

Should I create a separate issue for this repository?

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


yaron2:
This is a great improvement @vangheem. Please resolve the conflicts and we can get started on a review.

yaron2:
@vangheem please resolve the conflict.

yaron2:
@vangheem please see the minor linter issue here: https://github.com/dapr/components-contrib/pull/1715/files.

Also, can you please add tests for the new methods introduced here?

vangheem:
Yup, sorry, I've been slammed at work. I'll get it done this weekend! Thank you :)

vangheem:
How do I get these builds to run?

I also thought that I'd wait to update docs until after we finish the implementation.

I'll continue working on this and the implementation to make sure it can be supported correctly with more than just kafka.

pubsub/envelope.go
```diff
@@ -121,6 +122,44 @@ func NewCloudEventsEnvelope(id, source, eventType, subject string, topic string,
...
+	}
+	if eventType == "" {
+		eventType = DefaultCloudEventType
+	}
+	if dataContentType == "" {
+		dataContentType = DefaultCloudEventDataContentType
+	}
+
+	ce := map[string]interface{}{
+		"ce-" + IDField:              id,
```
gunniwho, line 144: I think these should all be `ce_` according to the [kafka protocol binding spec](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/bindings/kafka-protocol-binding.md#3231-property-names)

vangheem: Alright, I updated; however, I think this PR needs some work/adjustment. The pubsub implementation will need to manage properly serializing/deserializing

Each binding has a different way to place the data:

- http: `ce-` prefixed headers 
- amqp: `cloudEvents:` prefixed properties
- kafka: `ce_` prefixed headers

gunniwho: Yikes, I didn't realize there were different prefixes for different things...

gunniwho:
@vangheem have you thought about the scenario where a topic is replayed using the updated dapr? Would it be backwards compatible?

vangheem:
@gunniwho If it didn't already have the headers, the replayed messages would not have them either right? There shouldn't be any incompatibilities but you won't get the tracing connections either. Unless you had something else in mind?

gunniwho:
@vangheem I was thinking about the scenario where I've employed dapr pubsub for a while and then I upgrade it to use binary mode and it starts publishing events in a different format. What if I then need to replay the topic? The first events in the topic are in the old format and the latest in the new format. Would the dapr consumer be able to handle that switch and the client (user app) not notice at all?

vangheem:
@gunniwho hmmm, yes, that would be a problem. When using binaryMode, it's changing the format of the stored messages. It's similar to rawPayload. I wouldn't mix/match any of these on a single topic.

PR #1716, opened 2022-05-10
Complete Eventhubs bindings certification test

tanvigour:
Signed-off-by: tanvigour <tanvi.gour@gmail.com>

# Description

Contains remaining "IoThub and service principal authentication" test cases
## Issue reference

Please reference the issue this PR will close: #960 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


tanvigour:
Turns out the issue was network being intercepted in multiple test points. It does seem like the cleanup post network interruption isn't happening properly. Since reconnection doesn't need to be tested for all test points, we can get away by keeping it only in first one and get clean test runs. This fixes these runs locally. Although for github, I do still see timeouts and it seems like that could be because same test is using multiple components for example consumer2/eventhub_input.yml and consumer2/eventhubs_output.yml, using same secrets which probably isn't generating different components for each of them. 

Question for @berndverst , do the auto generated components get created once per test run, which means if you use multiple component files with secrets like I do per testpoint, basically is sharing resources? If so, how do I force it to generate new component for each yml per test? do I need to create new env vars through bicep ?

.github/infrastructure/conformance/azure/conf-test-azure-iothub.bicep
```diff
@@ -18,6 +18,7 @@ param confTestTags object = {}
 var iotHubBindingsConsumerGroupName = '${iotHubName}/events/bindings-cg'
 var iotHubPubsubConsumerGroupName = '${iotHubName}/events/pubsub-cg'
 var iotHubPubsubEhCertificationConsumerGroupName = '${iotHubName}/events/pubsub-eh-certification-cg'
+var iotHubBindingsEhCertificationConsumerGroupName = '${iotHubName}/events/bindings-consumergroup'
```
berndverst, line 21: When did you add this? Anything added here won't be applied to the infrastructure automatically and must be applied to me. I was not aware that anything was missing from the infra today?

tanvigour: Thanks for catching this. It's not needed anymore, rolling back

tests/certification/bindings/azure/eventhubs/eventhubs_test.go
```diff
@@ -369,7 +379,7 @@ func TestEventhubBindingMultiplePartition(t *testing.T) {
 			outputmsg2[i] = fmt.Sprintf("output binding: Message %03d", i)
 		}
 		consumerGroup2.ExpectStrings(outputmsg2...)
-		time.Sleep(40 * time.Second)
+		time.Sleep(120 * time.Second)
```
berndverst, line 382: As discussed offline, in this test, please send 50 events to partition 0 (via the component default) and 50 events to partition 1 via the event metadata partitionkey override. Then verify only the desired events went to the appropriate partition and nowhere else.

berndverst (changes requested):

tanvigour:
@berndverst : Addressed your feedback and test runs clean now, please review. Thanks!

tests/certification/bindings/azure/eventhubs/eventhubs_test.go
```diff
@@ -334,61 +347,73 @@ func TestEventhubBindingMultiplePartition(t *testing.T) {
...
+						Metadata:  metadata0,
+					})
+				require.NoError(ctx, err, "error publishing message")
+			}
 
-		// Assert the observed messages
-		consumerGroup2.Assert(ctx, time.Minute)
-		return nil
-	}
+			// Assert the observed messages
```
berndverst, line 384: Please immediately send the events to the other binding here first before deleting or asserting anything!

The point is we want to see that there is potential cross-pollination so to say. There should not be any and that's what we must verify.

berndverst (changes requested):
Also, on Line 424 and 433 please add:

consumerGroup1.FailIfNotExpected(t, string(in.Data))

consumerGroup2.FailIfNotExpected(t, string(in.Data))

respectively.

This is very important. Right now you don't have anything that actually checks whether an event was sent to the wrong consumerGroup. Your test would presently still pass even if an event were sent to both consumerGroups at once.

Hence all these changes are required.

berndverst (approved):

PR #1725, opened 2022-05-13
Fix Azure deploy script for users with a first.last email

ItalyPaleAle:
The existing script was not working for Microsoft employees with a first.last email address configured


berndverst (approved):

berndverst:
This change is already included in #1702

Issue #1709, opened 2022-05-06
Multiple subscriptions for MQTT component are routed to one API endpoint 


rkrawiel:
## Expected Behavior

When having multiple subscriptions to different topics should route them to configured API endpoints.

## Actual Behavior

When having more than one subscription all messages are routed to last registered API endpoint (handler)

## Steps to Reproduce the Problem
Having two subscriptions with configuration like this:
Config
```
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: testpubsub
spec:
  type: pubsub.mqtt
  version: v1
  metadata:
  - name: url
    value: "xxx"
  - name: qos
    value: 1
  - name: retain
    value: "false"
  - name: cleanSession
    value: "true"
  - name: backOffMaxRetries
    value: "5"
```
Subscription 1
```
apiVersion: dapr.io/v1alpha1
kind: Subscription
metadata:
  name: subscription1 
spec:
  pubsubname: testpubsub
  topic: topic1
  route: /api/route1
```
Subscription 2
```
apiVersion: dapr.io/v1alpha1
kind: Subscription
metadata:
  name: subscription2
spec:
  pubsubname: testpubsub
  topic: topic2
  route: /api/route2
```
With this configuration messages from both topics will be routed to /api/route2 which is not expected behavior. 

**Initial investigation leads to mqtt.go Subscribe function where it's registering all topics to one handler (line 216)**

rkrawiel:
@yaron2  Unfortunately, I don't have enough knowledge of the codebase and Go lang itself.

ItalyPaleAle:
Fixed by #1737

PR #1726, opened 2022-05-14
mqtt: fix multle subscribe issue

daixiang0:
Signed-off-by: Loong Dai <loong.dai@intel.com>

# Description

```
	// SubscribeMultiple starts a new subscription for multiple topics. Provide a MessageHandler to
	// be executed when a message is published on one of the topics provided, or nil for the
	// default handler.
	//
	// If options.OrderMatters is true (the default) then callback must not block or
	// call functions within this package that may block (e.g. Publish) other than in
	// a new go routine.
	// callback must be safe for concurrent use by multiple goroutines.
	SubscribeMultiple(filters map[string]byte, callback MessageHandler) Token
```
Refer to https://docs.dapr.io/reference/api/pubsub_api/, we start a new subscription for one topic one time, But SubscribeMultiple would starts a new subscription for multiple topics, so change API.

## Issue reference

Please reference the issue this PR will close: #1709

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


daixiang0:
Need some time to update conformance tests since now it does not support subscribing two topics with different callback.

yaron2:
> Need some time to update conformance tests since now it does not support subscribing two topics with different callback.
> 
> If this fix is in urgent need, we can merge it first and then add tests later.

It's needed for 1.8.

yaron2:
@daixiang0 conformance tests seem to pass..

Also what do you mean by `now it does not support subscribing two topics with different callback.` - according to the change, subscribing to two topics using two different callbacks IS supported.

pubsub/mqtt/mqtt.go
```diff
@@ -213,8 +209,9 @@ func (m *mqttPubSub) Subscribe(req pubsub.SubscribeRequest, handler pubsub.Handl
 	m.consumer = c
 
 	go func() {
-		token := m.consumer.SubscribeMultiple(
```
ItalyPaleAle, line 216: I think you need to keep SubscribeMultiple. The connection is re-created every time the `Subscribe` method is invoked, so you need to re-register all topic subscriptions. Otherwise only the last one will work.

yaron2: @daixiang0 please add a test to verify this works as intended.

daixiang0: It is what I said before (`now it does not support subscribing two topics with different callback`).

@ItalyPaleAle I check SubscribeMultiple API, it registers all topic to one subscribation, so only the last one will work:
```
	// SubscribeMultiple starts a new subscription for multiple topics. Provide a MessageHandler to
	// be executed when a message is published on one of the topics provided, or nil for the
	// default handler.
	//
	// If options.OrderMatters is true (the default) then callback must not block or
	// call functions within this package that may block (e.g. Publish) other than in
	// a new go routine.
	// callback must be safe for concurrent use by multiple goroutines.
	SubscribeMultiple(filters map[string]byte, callback MessageHandler) Token
```

ItalyPaleAle: Thanks for confirming!

Please take a look at #1737 which does support multiple handlers.

daixiang0: Oh I miss it, thank you!

daixiang0:
#1737 may fix it.

PR #1729, opened 2022-05-17
Tweaks for ASB pubsub

ItalyPaleAle:
1. Implement a rate limit when messages are retried, as per conversation with @artursouza . Because when a message is abandoned in ASB it gets re-queued immediately and sent to someone else, we need to slow down or in certain situations we can consume all max retries of messages too quickly (possibly at the rate of 100's per second)
   - This required adding a new metadata key `maxRetriableErrorsPerSec` in both pubsub and binding
1. Binding: previously, the binding was reconnecting to ASB after receiving every single message. Now the connection is kept open to avoid wasting resources. (I should have noticed that earlier).
1. Do not attempt to reconnect when context was canceled
1. Message finalization uses a background context and not the running ctx, in case the app is shutting down

### DIFF
--- a/bindings/azure/servicebusqueues/servicebusqueues.go
+++ b/bindings/azure/servicebusqueues/servicebusqueues.go
@@ -19,13 +19,13 @@ import (
 	"errors"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 	servicebus "github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus"
 	sbadmin "github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/admin"
 	"github.com/cenkalti/backoff/v4"
+	"go.uber.org/ratelimit"
 
 	azauth "github.com/dapr/components-contrib/authentication/azure"
 	"github.com/dapr/components-contrib/bindings"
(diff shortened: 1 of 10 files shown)


pubsub/azure/servicebus/servicebus.go
```diff
@@ -463,6 +463,12 @@ func (a *azureServiceBus) Subscribe(req pubsub.SubscribeRequest, handler pubsub.
 			sub.close(ctx)
 			cancel()
 
+			// If context was canceled, do not attempt to reconnect
+			if a.ctx.Err() != nil {
```
berndverst, line 467: I think it may be worthwhile to check that this specifically is `context.Canceled` and not DeadlineExceeded or other error contexts.

berndverst (changes requested):
FYI `a.ctx` is currently not receiving context from daprd. None of the pubsub brokers have the context passed through, and this component here sets this to context.background at the moment.

As a result this PR cannot actually have the complete impact (yet).

berndverst:
> Although the context isn't cascaded from Dapr, this component does use a context internally and it's canceled when the context is closed: https://github.com/dapr/components-contrib/pull/1729/files#diff-8f0a750f05222eb118aded3b8c6c3a926a2564220b5a9c2fd06e6e75c7180aafL615

@ItalyPaleAle but how is the context closed right now?

ItalyPaleAle:
It's created in the `Init` method

https://github.com/dapr/components-contrib/blob/6b99b02a2d36b69999957e3c790b46a15069e1dc/pubsub/azure/servicebus/servicebus.go#L305

It's closed in the `Close` method, which is invoked by Dapr at shutdown.

https://github.com/dapr/components-contrib/blob/6b99b02a2d36b69999957e3c790b46a15069e1dc/pubsub/azure/servicebus/servicebus.go#L621

Internally in the component, the context is used to cancel subscriptions.

When Dapr adds support for cascading the context, this should work transparently without changes.

pubsub/azure/servicebus/servicebus.go
```diff
@@ -36,38 +36,7 @@ import (
 )
 
 const (
-	// Keys.
```
ItalyPaleAle, line 39: These defaults were just moved to the `metadata.go` file so they're together with the emtadata keys

bindings/azure/servicebusqueues/servicebusqueues.go
```diff
@@ -345,19 +382,5 @@ func (a *AzureServiceBusQueues) Close() (err error) {
 			return err
 		}
 	}
-	a.setShutdown()
 	return nil
 }
-
-func (a *AzureServiceBusQueues) setShutdown() {
```
ItalyPaleAle, line 352: These 3 methods are not necessary. We have a context and looking at whether the context has been canceled is thread-safe too.

berndverst (approved):
Please open Docs PRs to document the new metadata options or any changes to metadata. E.g. `maxRetriablErrorsPerSec`

PR #1730, opened 2022-05-18
Enables Blobstorage Binding certification test & updates test due to library changes

berndverst:

# Description

All this time the Blob Storage Binding certification was not actually enabled... whoops. This enables the test.


tests/certification/bindings/azure/blobstorage/blobstorage_test.go
```diff
@@ -558,16 +558,16 @@ func TestBlobStorage(t *testing.T) {
 
 		cred, _ := azblob.NewSharedKeyCredential(os.Getenv("AzureBlobStorageAccount"), os.Getenv("AzureBlobStorageAccessKey"))
 		service, _ := azblob.NewServiceClientWithSharedKey(fmt.Sprintf("https://%s.blob.core.windows.net/", os.Getenv("AzureBlobStorageAccount")), cred, nil)
-		containerClient := service.NewContainerClient(os.Getenv("AzureBlobStorageContainer"))
+		containerClient, _ := service.NewContainerClient(os.Getenv("AzureBlobStorageContainer"))
```
shubham1172, line 561: Should we handle the error and fail the test?

tests/certification/bindings/azure/blobstorage/blobstorage_test.go
```diff
@@ -558,16 +558,16 @@ func TestBlobStorage(t *testing.T) {
...
 
-		blobClient := containerClient.NewBlockBlobClient("snapshotthis.txt")
-		uploadResp, uploadErr := blobClient.UploadBufferToBlockBlob(
+		blobClient, _ := containerClient.NewBlockBlobClient("snapshotthis.txt")
```
shubham1172, line 563: Same comment as above.

shubham1172 (approved):

tests/certification/bindings/azure/blobstorage/blobstorage_test.go
berndverst: Normally I'd say we should handle it but it really doesn't matter here which is why I didn't :)

berndverst:
The swallowing of errors added is fine FYI. This is a separate SDK I'm using to directly insert and validate blob storage contents without going through the component.
If these fail the test will fail and that's all we care to know.

artursouza (approved):

PR #1737, opened 2022-05-20
Fix: route messages to the correct publisher

### LINKED ISSUE #1735: pubsub.mqtt: all messages are sent to the same handler regardless of topic
I have pinpointed this issue in the code and will include a fix in a PR soon.

## Repro code

Component:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: orders
spec:
  type: pubsub.mqtt
  version: v1
  metadata:
    - name: url
      value: "tcp://127.0.0.1:1883"
    - name: qos
      value: 1
```

Code:

```go
package main

import (
	"context"
	"errors"
	"log"
	"math/rand"
	"net/http"

	"github.com/dapr/go-sdk/service/common"
	daprd "github.com/dapr/go-sdk/service/http"
)

var sub1 = &common.Subscription{
	PubsubName: "orders",
	Topic:      "topic1",
	Route:      "/checkout1",
}
var sub2 = &common.Subscription{
	PubsubName: "orders",
	Topic:      "topic2",
	Route:      "/checkout2",
}

func main() {
	s := daprd.NewService("127.0.0.1:6002")
	if err := s.AddTopicEventHandler(sub1, eventHandler1); err != nil {
		log.Fatalf("error adding topic 1 subscription: %v", err)
	}
	if err := s.AddTopicEventHandler(sub2, eventHandler2); err != nil {
		log.Fatalf("error adding topic 2 subscription: %v", err)
	}
	if err := s.Start(); err != nil && err != http.ErrServerClosed {
		log.Fatalf("error listenning: %v", err)
	}
}

func eventHandler1(ctx context.Context, e *common.TopicEvent) (retry bool, err error) {
	log.Printf("eventHandler1 (topic=%s): Subscriber received: %s", e.Topic, e.Data)
	log.Printf("eventHandler1 (topic=%s): Done processing: %s", e.Topic, e.Data)
	return false, nil
}

func eventHandler2(ctx context.Context, e *common.TopicEvent) (retry bool, err error) {
	log.Printf("eventHandler2 (topic=%s): Subscriber received: %s", e.Topic, e.Data)
	log.Printf("eventHandler2 (topic=%s): Done processing: %s", e.Topic, e.Data)
	return false, nil
}

```

Run:

```sh
dapr run \
    --app-id myapp \
    --app-port 6002 \
    --dapr-http-port 3602 \
    --dapr-grpc-port 60002 \
    --components-path ./components \
    --log-level debug \
    --\
        go run main.go
```

Send messages to the two topics:

```sh
curl \
    -X POST \
    "http://localhost:3602/v1.0/publish/orders/topic1" \
    -H "Content-Type: application/json" \
    -d "{\"orderId\": \"1"}"
curl \
    -X POST \
    "http://localhost:3602/v1.0/publish/orders/topic2" \
    -H "Content-Type: application/json" \
    -d "{\"orderId\": \"2"}"
```

## Expected Behavior

Should log:

```
eventHandler1 (topic=topic1): Subscriber received: map[orderId:1]
eventHandler1 (topic=topic1): Done processing: map[orderId:1]
eventHandler2 (topic=topic2): Subscriber received: map[orderId:2]
eventHandler2 (topic=topic2): Done processing: map[orderId:2]
```

## Actual Behavior

Logs:

```
eventHandler2 (topic=topic1): Subscriber received: map[orderId:1]
eventHandler2 (topic=topic1): Done processing: map[orderId:1]
eventHandler2 (topic=topic2): Subscriber received: map[orderId:2]
eventHandler2 (topic=topic2): Done processing: map[orderId:2]
```

(Sometimes all messages go to `eventHandler1` instead, depending on which one is registered first)

## Release Note

RELEASE NOTE: **FIX** Routing pubsub.mqtt messages to the correct handler when more than one is registered

ItalyPaleAle:
Fixes #1735

Also including:

- Adding a lock in `Subscribe` to prevent concurrency issues with multiple reconnections at the same time
- Increase default timeout from 3 to 30 seconds
- Avoid logging full message body in debug logs
- `connect` supports a context


yaron2:
Why open this PR when this is out? https://github.com/dapr/components-contrib/pull/1726

yaron2:
Since this PR adds several other things to MQTT, I suggest we merge #1726 first and this PR can then build on top of that. Would much rather use `Subscribe` and not `SubscribeMultiple`.

ItalyPaleAle:
I did not see that PR.

I'm working on something I stumbled upon this issue. 

As I look into #1726, I am actually not sure that is implemented correctly. The existing code is re-creating a new connection every time `Subscribe` is invoked, which means that existing subscriptions are canceled. Using `SubcribeMultiple` is required in that case. Or only the latest subscription would be working.

I tested #1726 with the repro code in #1735 and in fact it only gets messages on the second topic it's subscribed to.

yaron2:
> I did not see that PR.
> 
> I was working on something when I stumbled upon this issue.
> 
> As I look into #1726, I am actually not sure that is implemented correctly. The existing code is re-creating a new connection every time `Subscribe` is invoked, which means that existing subscriptions are canceled. Using `SubcribeMultiple` is required in that case. Or only the latest subscription would be working.
> 
> I tested #1726 with the repro code in #1735 and in fact it only gets messages on the second topic it's subscribed to.

Ok, I asked @daixiang0 to verify. We'll wait and see what comes back.

berndverst (approved):

Issue #1699, opened 2022-05-02
Allow for prepared-statements in PostgreSQL binding


sevberg:
## Problem statement

Using a PostgreSQL binding component within an app brings a lot a welcome benefits, but unfortunately it is also regressive from a safety perspective. This is because SQL statements must be supplied completely to the component endpoint via the `sql` field in the metadata ([docs](https://docs.dapr.io/reference/components-reference/supported-bindings/postgres/#exec)). For static query statements (or, at least, those which are completely controlled by the app's source code), this is no problem at all. However, for dynamic queries, this leads at best to the Dapr-user needing to redundantly-implement functionality which makes use of Query-templates and checks for safe variable inserts (i.e. avoiding SQL-injection). At worst, a user will not think about this latter concern and will blindly use basic string formatting to build-up an SQL statement, thus opening themselves up to major security vulnerabilities.

## Describe the proposal

Therefore, my proposal is that Dapr's PostgreSQL Binding should allow for the use of [Prepared Statements](https://en.wikipedia.org/wiki/Prepared_statement), as is the norm for many SQL drivers. Luckily, the driver currently used for this binging, [pgx](https://github.com/jackc/pgx) has [already implemented this feature](https://github.com/jackc/pgx/blob/7ceeea6fe604db989489de762c9d6e0a7cb2ace7/pgxpool/pool.go#L470-L503) via the `args` parameters, and all that the related [Dapr functions](https://github.com/dapr/components-contrib/blob/bfd87eb7a53bef4ddb39c38cced22aded1bccf2f/bindings/postgres/postgres.go#L151-L186) need to do is to use them.

I can see this working in the following way:

When a static query is being made, then no change to the current API is required. I.e. a post request can be made against the binding-component's end point with a data input which looks like:

```
{
  "operation": "exec",
  "metadata": {
    "sql": "INSERT INTO foo (id, c1, ts) VALUES (1, 'demo', '2020-09-24T11:45:05Z07:00')"
  }
}
```

Conversely, when a dynamic SQL statement is used, then the normal `sql` field represents a prepared statement and an additional `args` field, consisting of a list of arguments, would represent the values that are to be inserted. The result could look like such:

```
{
  "operation": "exec",
  "metadata": {
    "sql": "INSERT INTO foo (id, c1, ts) VALUES ($1, $2, $3)",
    "args": [
        1, 
        'demo', 
        '2020-09-24T11:45:05Z07:00'
    ]
  }
}
```

[...]

PR #1739, opened 2022-05-23
add mssql binding

shakirmengrani:
# Description

I have added MS-SQLbinding support

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


daixiang0:
@shakirmengrani thanks for your contribution! Please fix DCO first.

shakirmengrani (commented):
Please review

hueifeng:
ping?

bindings/mssql/mssql.go
```diff
@@ -0,0 +1,158 @@
...
+	startTime := time.Now().UTC()
+
+	resp := &bindings.InvokeResponse{
+		Metadata: map[string]string{
+			respOpKey:        string(req.Operation),
+			respSQLKey:       s,
+			respStartTimeKey: startTime.Format(time.RFC3339Nano),
+		},
+	}
+	switch req.Operation {
```
berndverst, line 87: You didn't implement the exec operation!

Also, when implementing that - please implement it using Prepared Statements. The request should additionally include the arguments to be sent to the prepared statement. See this issue for reference https://github.com/dapr/components-contrib/issues/1699

Given this is a new binding, we should not even allow anything that is not a prepared statement!

berndverst (changes requested):
Please fix linter issues - there are some bugs here.

Also please implement the `exec` operation, but only to support PreparedStatements (please check the syntax required for MSSQL) or an equivalent secure default.

bindings/mssql/mssql.go
berndverst: This example is from Postgres -- not sure if a similar format exists for SQLServer. If it does, we should use prepared statements!

```
{
  "operation": "exec",
  "metadata": {
    "sql": "INSERT INTO foo (id, c1, ts) VALUES ($1, $2, $3)",
    "args": [
        1, 
        'demo', 
        '2020-09-24T11:45:05Z07:00'
    ]
  }
}
```

berndverst:
Due to lack of response and given the code freeze now I am moving this to release 1.9

PR #1743, opened 2022-05-24
Added test for multiple pubsub handlers

ItalyPaleAle:
This PR adds conformance tests for PubSub with multiple handlers, each listening on a different topic.

It's meant to catch issues such as #1709 

Opening as a draft until I get Event Hubs to work too, which will require some changes to the infrastructure.

### DIFF
--- a/tests/conformance/pubsub/pubsub.go
+++ b/tests/conformance/pubsub/pubsub.go
@@ -16,6 +16,8 @@ package pubsub
 import (
 	"context"
 	"fmt"
+	"reflect"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -25,6 +27,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/dapr/components-contrib/pubsub"
 	"github.com/dapr/components-contrib/tests/conformance/utils"
@@ -34,6 +37,8 @@ import (
 const (
 	defaultPubsubName             = "pubusub"
 	defaultTopicName              = "testTopic"
+	defaultMultiTopic1Name        = "multiTopic1"
+	defaultMultiTopic2Name        = "multiTopic2"
 	defaultMessageCount           = 10
 	defaultMaxReadDuration        = 60 * time.Second
 	defaultWaitDurationToPublish  = 5 * time.Second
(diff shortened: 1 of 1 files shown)


pubsub/kafka/kafka.go
```diff
@@ -24,32 +26,50 @@ import (
...
 
 func (p *PubSub) Subscribe(req pubsub.SubscribeRequest, handler pubsub.Handler) error {
-	topics := p.addTopic(req.Topic)
+	// Add the topic to the list of those we subscribe to
+	p.lock.Lock()
+	p.topics[req.Topic] = handler
+	topics := p.topicList()
+	p.lock.Unlock()
 
-	return p.kafka.Subscribe(topics, req.Metadata, newSubscribeAdapter(handler).adapter)
```
skyao, line 38: I suggestion to move the update code to kafka common code, not in kafka pubsub (we have kafka binding and we need to reuse code between kafka components )

pubsub/kafka/kafka.go
```diff
@@ -75,21 +97,3 @@ func (p *PubSub) Close() (err error) {
 func (p *PubSub) Features() []pubsub.Feature {
 	return nil
 }
-
-// subscribeAdapter is used to adapter pubsub.Handler to kafka.EventHandler with the same content.
-type subscribeAdapter struct {
```
skyao, line 80: Don't remove this subscribeAdapter, you should update it. So your update can effect in all kafka components.

ItalyPaleAle: Addressed

pubsub/kafka/kafka.go
ItalyPaleAle: I moved the routing logic to the common Kafka code.

yaron2:
@ItalyPaleAle is the test failing due to anything PR related?

ItalyPaleAle:
Ok, most tests are passing now.

The Kafka certification tests (pubsub and binding) seem to be flaky on master too, with most failures happening because Kafka isn't ready yet. I increased the timeouts and this has reduced flakiness. But they're still a bit flaky [also in master](https://github.com/dapr/components-contrib/actions/runs/2418990237).

These others are expected to fail:

- `pubsub.azure.eventhubs conformance` -> expected to fail until the pipeline is merged into master, as I had to update the pipeline to set the secret name
- `bindings.azure.storagequeues certification` and `bindings.azure.eventhubs certification` -> are failing in master too due to unrelated issues

berndverst (approved):

berndverst (approved):

PR #1754, opened 2022-05-25
Fixes for SNS/SQS PubSub

ItalyPaleAle:
Including:

- Fixed adding multiple handlers for different topics (tests in #1743 are now passing)
- Fixed delivering wrong message when concurrency mode is parallel
- Avoid having multiple pollers at the same time competing for messages

& more smaller fixes

Note that because I got this from the work I've been doing for adding contexts in the Subscribe methods, there are a few more changes that are related to passing contexts around. Those will be useful for another PR coming soon :)


pubsub/aws/snssqs/metadata.go
```diff
@@ -39,7 +39,7 @@ type snsSqsMetadata struct {
 	// if sqsDeadLettersQueueName is set to a value, then the messageReceiveLimit defines the number of times a message is received
 	// before it is moved to the dead-letters queue. This value must be smaller than messageRetryLimit.
 	messageReceiveLimit int64
-	// amount of time to await receipt of a message before making another request. Default: 1.
+	// amount of time to await receipt of a message before making another request. Default: 2.
```
ItalyPaleAle, line 42: Changed to 2 seconds. Calls return sooner if there are messages, so polling every 1 second seemed very wasteful

PR #1755, opened 2022-05-25
Fixed Kafka PubSub to allow multiple handlers for different topics

ItalyPaleAle:
With this, tests from #1743 are passing

(As per #1754, the reason for the work on passing contexts is that I got this from the branch where I am doing work on contexts for Subscribe)


pubsub/kafka/kafka.go
```diff
@@ -24,32 +26,50 @@ import (
...
-	return p.kafka.Subscribe(topics, req.Metadata, newSubscribeAdapter(handler).adapter)
+	return p.kafka.Subscribe(context.Background(), topics, req.Metadata, p.handleMessage)
 }
 
-func (p *PubSub) addTopic(newTopic string) []string {
-	// Add topic to our map of topics
-	p.topics[newTopic] = true
+func (p *PubSub) handleMessage(ctx context.Context, event *kafka.NewEvent) error {
+	// Get the handler
+	p.lock.RLock()
```
yaron2, line 49: Is this read lock needed if `Subscribe` is called for all topics before `handleMessage` is ever executed?

ItalyPaleAle: Actually in this case, no. But I backported this change from the work I've been doing on #1756 and in that case it's probably necessary if the subscription is being removed.

I can delete it from here for now (I have to make a few more changes to this PR per Sky's feedback anyways)

PR #1756, opened 2022-05-25
Add contexts to pubsub.Subscribe to allow early cancelation

ItalyPaleAle:
# Description

This PR is related to dapr/dapr#4624. As noted there, we have an issue in the runtime where all components are shut down after the grace period, when the app is likely already stopped. Because of that, certain input components (the subscribe part of pubsub and the input part of bindings - the latter not in scope of this PR), can continue bringing new work when it's known to fail.

In order to fix the issue linked above properly, we need to implement a way for PubSub components to have the "publish" part closed before the "subscribe" one (and in the future that will need to be done for input bindings too).

This PR achieves precisely that by adding a context in the `Subscribe` method. When that context is canceled (which can be at any time), the subscription is removed.

PS: This API change was implemented so it can one day be used for dapr/dapr#814 too, as it allows canceling individual subscriptions by using a different context. Although that's not possible today because it requires more work on the runtime, it does implement everything that's needed in the pubsub components already.

## Checklist

* [X] Code compiles correctly
* [x] Created/updated tests

## TODO

Marking this PR as draft until the following are done (in order):

1. [x] Update conformance tests to test the new behavior
2. [x] Merge #1754 (included in this PR)
3. [x] Merge #1755 (included in this PR)
4. [x] Merge #1767 (not merged, but implemented)
5. [x] Merge #1743 (included in this PR)


go.mod
```diff
@@ -163,6 +162,7 @@ require (
 	github.com/huaweicloud/huaweicloud-sdk-go-v3 v0.0.87
 	github.com/labd/commercetools-go-sdk v0.3.2
 	github.com/nacos-group/nacos-sdk-go/v2 v2.0.1
+	github.com/rabbitmq/amqp091-go v1.3.4
```
ItalyPaleAle, line 165: `github.com/rabbitmq/amqp091-go` is the new version of `github.com/streadway/amqp` and it's maintained by RabbitMQ now

ItalyPaleAle:
Note that all certification tests will fail until the corresponding PR is merged in dapr/dapr, due to a change in the interface of the pubsub components

PR #1778, opened 2022-06-09
Add postgresql certification test + connection idle time

yaron2:
This PR adds a certification test for PostreSQL and a connection idle time metadata property consistent with the `MySQL` database implementations.

Test plan is derived from Redis and SQL Server certification tests.


artursouza (changes requested):
Overall, it looks good. Please, add it to the workflow as well along with the docker compose file.

yaron2:
> Overall, it looks good. Please, add it to the workflow as well along with the docker compose file.

Done. (docker compose file was already present).

artursouza (approved):

PR #1785, opened 2022-06-13
Add file upload to s3

yaron2:
This PR adds the ability to have Dapr read a file and upload the contents instead of having the user write boilerplate code to read the file and then send the data over the wire, which can be problematic for large file uploads.

### DIFF
--- a/bindings/aws/s3/s3.go
+++ b/bindings/aws/s3/s3.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"os"
 	"strconv"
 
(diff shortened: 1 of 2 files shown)


yaron2:
@daixiang0 please review when possible.

berndverst (approved):

chenjpu:
```
data, err = os.ReadFile(metadata.FilePath)
````
When the file is too large, all read to [] byte, will there be OOM problems?

**Comment by chenjpu** [comment]
```
data, err = os.ReadFile(metadata.FilePath)
````
When the file is too large, all read to [] byte, will there be OOM problems?

PR #1786, opened 2022-06-13
Upgrade Apache Dubbo again - pin to Dapr support branch

berndverst:

# Description

Pin Dubbo to altest commit in `dubbo.apache.org/dubbo-go/v3@dapr-support`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -155,7 +155,7 @@ require (
 )
 
 require (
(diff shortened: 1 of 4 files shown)


berndverst:
FYI @LaurenceLiZhixin -- had to make this change because we were having import problems.

artursouza (approved):

PR #1791, opened 2022-06-15
Shared implementation for Azure Service Bus subscriptions for binding and pubsub

### LINKED ISSUE #1789: binding.azure.servicebusqueues: renew lock for messages while they're being processed
From Discord:

> Hello, is the Azure Service Bus Queue binding renewing message locks or not? I was trying to understand the code and saw that it's calling 'NewReceiverForQueue' with no options. In the Azure Service Bus Go SDK, it seems like if there are no options called to that method, it defaults to ReceiveModePeekLock

I need to triage this further, but at a first look it seems to be an actual issue, even in the master branch.

The PubSub component is not impacted as it renews locks (as it's designed to be asynchronous). The binding, instead, is designed to be synchronous and today does not renew locks; however that can be a problem if the handler takes a very long time to process the message.

ItalyPaleAle:
# Description

This PR changes the Azure Service Bus components (binding and pubsub) to rely on a shared implementation for the subscription part to the queue or topic.

This makes it possible to bring into the binding a lot of the additional capabilities that were present in the pubsub component, for free, including support for multiple handlers in parallel (although it's limited to 1 by default for backwards-compatibility) and support for renewing locks (fixes #1789)

## Issue reference

Fixes #1789 by bringing in the missing features

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Extended the documentation https://github.com/dapr/docs/pull/2529


bindings/azure/servicebusqueues/metadata.go
```diff
@@ -0,0 +1,124 @@
+package servicebusqueues
```
ItalyPaleAle, line 1: Moved all metadata things to a separate file since it keeps growing

bindings/azure/servicebusqueues/metadata.go
```diff
@@ -0,0 +1,124 @@
...
+	// Default minimum and maximum recovery time while trying to reconnect
+	defaultMinConnectionRecoveryInSec = 2
+	defaultMaxConnectionRecoveryInSec = 300
+
+	// Default lock renewal interval, in seconds
+	defaultLockRenewalInSec = 20
+
+	// Default number of max active messages
+	// Max active messages should be >= max concurrent handlers
+	defaultMaxActiveMessages = 2
```
ItalyPaleAle, line 45: This and defaultMaxConcurrentHandlers are much lower than in the pubsub component to maintain compatibility with the old behavior, which was synchronous and processed one message at a time only

bindings/azure/servicebusqueues/metadata.go
```diff
@@ -0,0 +1,124 @@
...
+	// Default minimum and maximum recovery time while trying to reconnect
+	defaultMinConnectionRecoveryInSec = 2
+	defaultMaxConnectionRecoveryInSec = 300
+
+	// Default lock renewal interval, in seconds
+	defaultLockRenewalInSec = 20
+
+	// Default number of max active messages
+	// Max active messages should be >= max concurrent handlers
+	defaultMaxActiveMessages = 1
```
ItalyPaleAle, line 45: This and defaultMaxConcurrentHandlers are much lower than in the pubsub component to maintain compatibility with the old behavior, which was synchronous and processed one message at a time only

berndverst:
I hate to be the person, but can you also swap us to `azservicebus` version `1.0.1` which is out now? And that requires a swap to `azidentity` `1.0` and `azcore` 1.0?

ItalyPaleAle:
I made that change

PR #1792, opened 2022-06-16
Upgrade to Azure Table Storage Track 2 SDK in State Store and add CosmosDB Table API support

berndverst:

# Description

- Upgrades to Azure Table Storage Track 2 SDK. This includes some other library updates.
- Adds support for Azure Cosmos DB Table API
- Adds support for custom service endpoint (for other Azure environments)
- Adds support for AAD authentication

~BREAKING CHANGE: At this time this PR no longer supports using an empty row key, and instead assigns a default value `_dapr_empty_row_key_`.~


berndverst:
If this can get merged it won't be a breaking change anymore!

https://github.com/Azure/azure-sdk-for-go/pull/18425

state/azure/tablestorage/tablestorage.go
```diff
@@ -53,44 +55,56 @@ import (
...
 	valueEntityProperty = "Value"
-	operationTimeout    = 1000
 
-	accountNameKey = "accountName"
-	accountKeyKey  = "accountKey"
-	tableNameKey   = "tableName"
+	accountNameKey  = "accountName"
+	accountKeyKey   = "accountKey"
+	tableNameKey    = "tableName"
+	cosmosDBmodeKey = "cosmosDBmode"
```
ItalyPaleAle, line 62: What's the correct capitalization for this? Should this be `cosmosDbMode`, `cosmosDBMode`...?

state/azure/tablestorage/tablestorage.go
```diff
@@ -53,44 +55,56 @@ import (
...
-	client.AddToUserAgent(userAgent)
-	r.table = tables.GetTableReference(meta.tableName)
+	cred, err := aztables.NewSharedKeyCredential(meta.accountName, meta.accountKey)
+	if err != nil {
+		return err
+	}
+	r.cosmosDBmode = meta.cosmosDBmode
+	var serviceURL string
+	if r.cosmosDBmode {
+		serviceURL = fmt.Sprintf("https://%s.table.cosmos.azure.com", meta.accountName)
```
ItalyPaleAle, line 96: Wha if instead of using a boolean `cosmosDBmode` we allowed setting the endpoint directly? Right now we support only Cosmos DB and Azure Storage Tables running Azure public cloud. But there are other clouds (with different endpoints) and then there's Azure Stack etc.

state/azure/tablestorage/tablestorage.go
```diff
@@ -53,44 +55,56 @@ import (
...
+
+	client, err := aztables.NewServiceClientWithSharedKey(serviceURL, cred, nil)
+	if err != nil {
+		return err
+	}
 
-	// check table exists
-	r.logger.Debugf("using table '%s'", meta.tableName)
-	err = r.table.Create(operationTimeout, storage.FullMetadata, nil)
+	_, err = client.CreateTable(ctx.Background(), meta.tableName, nil)
```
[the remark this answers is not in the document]
ItalyPaleAle: Perhaps it should be another metadata option `timeout` to control the timeout of all requests?

state/azure/tablestorage/tablestorage.go
```diff
@@ -128,18 +143,16 @@ func (r *StateStore) Delete(req *state.DeleteRequest) error {
 
 func (r *StateStore) Get(req *state.GetRequest) (*state.GetResponse, error) {
 	r.logger.Debugf("fetching %s", req.Key)
-	pk, rk := getPartitionAndRowKey(req.Key)
-	entity := r.table.GetEntityReference(pk, rk)
-	err := entity.Get(operationTimeout, storage.FullMetadata, nil)
+	pk, rk := getPartitionAndRowKey(req.Key, r.cosmosDBmode)
+	resp, err := r.client.GetEntity(ctx.Background(), pk, rk, nil)
```
ItalyPaleAle, line 147: Same, we should always have timeouts in all contexts

state/azure/tablestorage/tablestorage.go
```diff
@@ -187,34 +200,37 @@ func getTablesMetadata(metadata map[string]string) (*tablesMetadata, error) {
 		return nil, errors.New(fmt.Sprintf("missing or empty %s field from metadata", tableNameKey))
 	}
 
+	if val, ok := metadata[cosmosDBmodeKey]; ok && val != "" {
+		meta.cosmosDBmode = val == "true"
```
ItalyPaleAle, line 204: Consider using `utils.IsTruthy`

state/azure/tablestorage/tablestorage.go
```diff
@@ -187,34 +200,37 @@ func getTablesMetadata(metadata map[string]string) (*tablesMetadata, error) {
...
-		etag = *req.ETag
+	marshalledEntity, err := r.marshal(req)
+	if err != nil {
+		return err
 	}
-	entity.OdataEtag = etag
 
 	// InsertOrReplace does not support ETag concurrency, therefore we will use Insert to check for key existence
 	// and then use Update to update the key if it exists with the specified ETag
+	_, err = r.client.AddEntity(ctx.Background(), marshalledEntity, nil)
```
ItalyPaleAle, line 220: Ditto on timeouts

state/azure/tablestorage/tablestorage.go
```diff
@@ -187,34 +200,37 @@ func getTablesMetadata(metadata map[string]string) (*tablesMetadata, error) {
...
 
-	err := entity.Insert(storage.FullMetadata, nil)
 	if err != nil {
 		// If Insert failed because item already exists, try to Update instead per Upsert semantics
 		if isEntityAlreadyExistsError(err) {
```
ItalyPaleAle, line 224: Can you short-circuit this?

```go
if err == nil {
    return nil
} else if !isEntityAlreadyExistsError(err) {
    // Any other unexpected error on Insert is propagated to the caller
	return err
}

// Rest of code

File: state/azure/tablestorage/tablestorage.go [hunk 8]
```diff
@@ -241,63 +263,94 @@ func (r *StateStore) writeRow(req *state.SetRequest) error {
...
+		Entity: aztables.Entity{
+			PartitionKey: pk,
+			RowKey:       rk,
+		},
+		Properties: map[string]interface{}{
+			valueEntityProperty: value,
+		},
+	}
+
+	var etag string
```
ItalyPaleAle, line 336: Nit. Can probably just be this?

```go
if req.ETag != nil {
	entity.ETag = *req.ETag
}
```

state/azure/tablestorage/tablestorage.go
```diff
@@ -311,7 +364,7 @@ func (r *StateStore) unmarshal(row *storage.Entity) ([]byte, *string, error) {
 	}
 
 	// use native ETag
-	etag := row.OdataEtag
+	etag := myEntity.ETag
 
 	return []byte(sv), ptr.String(etag), nil
```
ItalyPaleAle, line 369: The Azure SDK have their own `to.Ptr(etag)` if you want to avoid using a different package

state/azure/tablestorage/tablestorage.go
berndverst: I'll separately add a custom endpoint option which will take precedence if set. I think that is a better approach.

state/azure/tablestorage/tablestorage.go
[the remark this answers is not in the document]
berndverst: There previously was a default timeout -- I'll just set that one we used to have before.

state/azure/tablestorage/tablestorage.go
ItalyPaleAle: Sounds good to me!

state/azure/tablestorage/tablestorage.go
berndverst: No it really needs to be this very exact code. I know it's weird... basically, no error is code.. and a specific error is also good.. but all other errors are bad :p

state/azure/tablestorage/tablestorage.go
berndverst: Yeah - this was left over from some refactor. Updated now.

state/azure/tablestorage/tablestorage.go
berndverst: I don't need the Ptr stuff in the first place. Removed, thanks

state/azure/tablestorage/tablestorage.go
```diff
@@ -53,44 +55,62 @@ import (
...
 	valueEntityProperty = "Value"
-	operationTimeout    = 1000
 
-	accountNameKey = "accountName"
-	accountKeyKey  = "accountKey"
-	tableNameKey   = "tableName"
+	accountNameKey  = "accountName"
+	accountKeyKey   = "accountKey"
+	tableNameKey    = "tableName"
+	cosmosDbModeKey = "cosmosDbmode"
```
ItalyPaleAle, line 62: Check casing here

state/azure/tablestorage/tablestorage.go
ItalyPaleAle: Yeah that's why I was saying to format it that way. Just so there's less nesting of `if`s. But not a blocker :)

ItalyPaleAle (approved):

ItalyPaleAle (approved):

bindings/azure/servicebusqueues/servicebusqueues.go
```diff
@@ -305,13 +305,6 @@ func (a *AzureServiceBusQueues) Read(handler bindings.Handler) error {
 
 			msg := msgs[0]
 
-			body, err := msg.Body()
-			if err != nil {
```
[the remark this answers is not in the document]
ItalyPaleAle: I can explain that. The new Storage SDK pulled in a new vesion of `azcore` and forced us to upgrade the Service Bus SDK too. In the latest version of the Service Bus SDK, `msg.Body` is a propetrty and not a method.

pubsub/azure/servicebus/message.go
```diff
@@ -78,12 +78,7 @@ func NewPubsubMessageFromASBMessage(asbMsg *azservicebus.ReceivedMessage, topic
 		Topic: topic,
 	}
 
-	body, err := asbMsg.Body()
```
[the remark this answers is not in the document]
ItalyPaleAle: Same as above.

tests/conformance/common.go
```diff
@@ -419,7 +419,9 @@ func loadStateStore(tc TestComponent) state.Store {
 		store = s_postgresql.NewPostgreSQLStateStore(testLogger)
 	case "mysql":
 		store = s_mysql.NewMySQLStateStore(testLogger)
-	case "azure.tablestorage":
```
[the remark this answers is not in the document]
berndverst: Correct. Look at KeyVault

bindings/azure/servicebusqueues/servicebusqueues.go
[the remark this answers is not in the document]
berndverst: It's required because of dependency changes

pubsub/azure/servicebus/message.go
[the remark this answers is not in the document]
berndverst: Underlying library change made this tiny api change

state/azure/tablestorage/tablestorage.go
```diff
@@ -38,59 +38,108 @@ Concurrency is supported with ETags according to https://docs.microsoft.com/en-u
...
 	valueEntityProperty = "Value"
-	operationTimeout    = 1000
 
-	accountNameKey = "accountName"
-	accountKeyKey  = "accountKey"
-	tableNameKey   = "tableName"
+	accountNameKey  = "accountName"
+	accountKeyKey   = "accountKey"
+	tableNameKey    = "tableName"
+	cosmosDbModeKey = "cosmosDbMode"
```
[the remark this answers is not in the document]
berndverst: This will never support other modes. The upstream SDK only can work with two types of table storage. We want the current default to be storage accounts to avoid a breaking change.

A toggle between CosmosDB and storage accounts is all that is necessary.

I really find this solution better as a user given this will never support other table stores.

state/azure/tablestorage/tablestorage.go
```diff
@@ -241,63 +316,95 @@ func (r *StateStore) writeRow(req *state.SetRequest) error {
...
+	return err
 }
 
-func getPartitionAndRowKey(key string) (string, string) {
+func getPartitionAndRowKey(key string, cosmosDBmode bool) (string, string) {
 	pr := strings.Split(key, keyDelimiter)
 	if len(pr) != 2 {
-		return pr[0], ""
+		if cosmosDBmode {
+			return pr[0], "_dapr_empty_row_key_value_"
```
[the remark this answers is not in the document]
berndverst: Because storage account table storage allows empty row keys and CosmosDB table APi does not

state/azure/tablestorage/tablestorage.go
```diff
@@ -38,59 +38,108 @@ Concurrency is supported with ETags according to https://docs.microsoft.com/en-u
...
+		}
+		client, innerErr = aztables.NewServiceClient(serviceURL, token, nil)
+		if err != nil {
+			return innerErr
+		}
+	}
+
+	createContext, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	_, err = client.CreateTable(createContext, meta.tableName, nil)
```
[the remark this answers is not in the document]
berndverst: Let me see if with this new SDK there is a different way to query for the existence of the table.

Please note that the Shared Key Credentials always enable creation of tables, but some fine grained AAD roles could indeed now restrict table creation and therefore not give the expected error.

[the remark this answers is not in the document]
berndverst: I added a new metadata option `skipCreateTable` which can be used in the AAD case and the shared key case also. This really only makes sense in the case where someone used very fine grained AAD permissions.

artursouza (approved):

Issue #1797, opened 2022-06-18
Tracking issue: Disable built-in retries in components


ItalyPaleAle:
With the addition of resiliency in the runtime, components should not have built-in retry policies, as that multiples the attempts.

- [ ] Input bindings
- [ ] Output bindings
- [ ] Pubsub: receive
- [ ] Pubsub: publish

berndverst:
@artursouza built-in retries in components is entirely separate scope from the built-in retries that Hal is working on. Hence I'm moving this to 1.9 release.

halspang:
I wrote a proposal for how to do this consistently across components. Would be good to get some thoughts on it: https://github.com/dapr/components-contrib/issues/1863

berndverst:
Thi falls under a parent issue of Component Resiliency. https://github.com/dapr/components-contrib/issues/1863

berndverst:
This depends on https://github.com/dapr/components-contrib/issues/1863

There first needs to be a design to expand Dapr Resiliency as it exists today so it can be decided what types of behavior should be controllable from Resiliency policies. Only when the new policy type exists in Dapr Runtime can we actually remove the built in component retries and connect this to Dapr Resiliency instead.

I do not see that work being completed in 1.13, so this issue here likely cannot begin until 1.14 or later.

PR #1798, opened 2022-06-18
Disable built-in retries in pubsub subscribe

### LINKED ISSUE #1813: [pubsub][pulsar]When a message fails to be processed, subsequent messages will be blocked.
## Expected Behavior

Skip the failed messages or put the failed messages in the dead letter queue and continue consuming other messages.

## Actual Behavior

Subsequent messages will be blocked in the pulsar.
image

ItalyPaleAle:
See #1797

**Note on Pulsar**

I had to add a new metadata option `redeliveryDelay` to control the delay before messages are redelivered after a NACK. This is part of the Pulsar protocol and the default is 1 min, which was too long for our tests. Users may not dislike this either :)

PS: This also fixes #1813 

**Note on Hazelcast**

Built-in retries are preserved in Hazelcast due to https://github.com/dapr/components-contrib/issues/1808

**Note on MQTT**

The MQTT [protocol](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718043) supports ACK's but does not support NACK's. Because of that, if a message fails to be processed, we need to re-enqueue it manually before ACK'ing in the main queue.

Assuming that QOS is set to 1 (the default value now) or higher, there's no risk for messages to be lost. If the connection with the MQTT server is dropped before the ACK is sent, the server will re-send the message to another consumer (again, assuming QOS is ≠ 0).

However if the connection is dropped between the moment when the message is re-enqueued and the moment when the ACK is sent, clients may receive the same message twice.

Another downside of this approach is that messages are retried forever in case of a retriable error as we current do not implement a max retry attempt. Doing so would require modifying the payload of the message to store the retry count and that is not backwards-compatible.

(Today, messages that fail to be processed are synchronously retried in the same app, then silently dropped)

**Note on RabbitMQ**

Previously, messages that failed to be processed were being re-enqueued only once (first, the message was retried, synchronously, many times, then it was NACK'd with the order of being re-enqueued only if it was the first time it was NACK'd). Now, messages are re-enqueued unlimited times (or our tests won't pass).

Adding a limit would require modifying the payload of the message to store a retry count, and that's not backwards-compatible.


mukundansundar:
@ItalyPaleAle I have some concern regarding certification tests. Most of the certification tests for the components do test the successful redelivery of messages, wouldn't that be affected if inline retry is removed? Shouldn't the certification tests be configured to use the default resiliency mechanism also in this PR?

ItalyPaleAle:
Good question.

Right now these components include synchronous retries (meaning, the sidecar retries the operation without telling the pubsub server), which are usually not configurable and are in addition to what is controlled via resiliency policies. Not all components have those sync retries, just these that I modified.

This PR doesn't remove retries, it just makes them async. If a message fails, with this PR all components send it back to the queue (normally via a NACK, with the exception of MQTT as per comment above).
The difference is that in this case, if you have more than 1 instance of your app, the message may be redelivered to a different instance, arguably a more correct behavior.

In addition, with resiliency policies in 1.8 it is possible to ALSO add synchronous retries. These can be used by all components (not just the few I modified with this PR) and are configurable.

berndverst (approved):

Issue #1799, opened 2022-06-20
consumer rebalance failed in kakfa binding component


skyao:
## Expected Behavior

In our certification test, we have a consumer rebalance case for kafak (both in kafak pubsub component and kafka binding component). The test logic is:

-  two apps (named app-2 and app-3) with kafka component (either kafak pubsub or kafak binding) are subscribing to the same kafka topic with the same consumer group
- So the events from this topic should be sent to one of the apps (app-2 and app-3) 
- if we stopped one of  one of the apps (in cert test case, app-2) during test, the events should be sent to app-3 and none of them will be lost
- if we count all the events received by app-2 and app-3, it should match the number of the event which ware sent to the topic.

## Actual Behavior

After app-2 stopped, there's no event sent to app-3. 

So the certification test failed for this: there are some events not received.

```bash
    flow.go:228: Completed step: wait
    flow.go:220: Running step: assert messages(consumer rebalance)
    watcher.go:390: Timed out with 301 items remaining
    flow.go:228: Completed step: assert messages(consumer rebalance)
```

## Steps to Reproduce the Problem

Our certification test for kakfa binding component failed for this issue, see:

https://github.com/dapr/components-contrib/runs/6958628713?check_suite_focus=true
https://github.com/dapr/components-contrib/runs/6965196737?check_suite_focus=true

skyao:
Sometimes the certification test will be passed, this is just for good luck and depends on what app is sending to before we stop app-2:

- if the events are sent to app-3, then there is no impact to app-3 while stopping app-2, so the test case will pass
- if the events are sent to app-2, because of this issue, the app-3 won't receive the events (by kafka comsumer rebalance) , so the test case will fail.

I will add some log to show this.

skyao:
The second highlight for this bug:  this bug is not exist in kafka pubsub component!

I have tested and verified that with kafka pubsub component, if the events is be sent to app-2 and app-2 is stopped, kafka comsumer rebalance will work and app-3 will start to receive the event, and this will happens in seconds after app-2 (in fact, dapr sidecar of app-2) is stopped.

This is why the certification test of kafka pubsub component is OK.

skyao:
And it is very tricky that the number of items remaining is always be 301 (which I can't understand): 

```bash
    flow.go:228: Completed step: wait
    flow.go:220: Running step: assert messages(consumer rebalance)
    watcher.go:390: Timed out with 301 items remaining
[...]

PR #1804, opened 2022-06-21
Fix kafka binding consumer rebalance

skyao:
# Description

kafka binding component  doesn't implement io.Closer interface, so when sidecar is stopped and dapr runtime is closed, kafka binding component won't be closed. 

This will cause a bug in kafka cert test because we want to stop one the kafka sidecar and let other kafka sidecar to continue to consumer kakfa event by kafka consumer rebalance.

BTW:  in an actual daprd sidecar, after the sidecar process is quit kafka binding component will also quit and loss the connection with kafka, so this bug only appears in cert test and not in real sidecar. 

## Issue reference

https://github.com/dapr/components-contrib/issues/1799

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/bindings/kafka/kafka.go
+++ b/bindings/kafka/kafka.go
@@ -38,7 +38,7 @@ type Binding struct {
 	logger       logger.Logger
 }
 
-// NewKafka returns a new kafka pubsub instance.
+// NewKafka returns a new kafka binding instance.
 func NewKafka(logger logger.Logger) *Binding {
 	k := kafka.NewKafka(logger)
 	// in kafka binding component, disable consumer retry by default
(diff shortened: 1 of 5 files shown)


berndverst (approved):

lebatuananh:
Hello @skyao , I am using the pub/sub component with kafka. Currently, I am also experiencing the situation that after creating new pods, the group is in kafka's rebalanced state for quite a long time. I am using dapr version 1.11.2. How do you repair it?
This is my configuration file
`
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name:  ichiba-integration-layer-pubsub
spec:
  type: pubsub.kafka
  version: v1
  metadata:
    - name: brokers
      value: "xxx.xxx.xxx.xxx:9092,xxx.xxx.xxx.xxx:9092,xxx.xxx.xxx.xxx:9092"     
    - name: consumerGroup
      value: "service.integration-layer"  
    - name: clientID
      value: "service.integration-layer"      
    - name: authType
      value: "none" 

policies:
  retries:
    pubsubRetry:
      policy: constant
      duration: 5s
      maxRetries: 5
      circuitBreakers:
    pubsubCB:
      maxRequests: 1
      interval: 8s
      timeout: 60s
      trip: consecutiveFailures > 8

scopes:
  - integration-layer
`

PR #1816, opened 2022-06-25
Azure Table Storage Certification tests

tanvigour:
Signed-off-by: tanvigour <tanvi.gour@gmail.com>

# Description

## Issue reference

Please reference the issue this PR will close: #950 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


tests/certification/state/azure/tablestorage/tablestorage_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+		// delete state
+		err = client.DeleteState(ctx, "statestore-basic", certificationTestPrefix+"key1", nil)
+		assert.NoError(t, err)
+
+		return nil
+	}
+
+	NonExistingTableTest := func(ctx flow.Context) error {
+		client, err := client.NewClientWithPort(fmt.Sprint(currentGrpcPort))
```
berndverst, line 76: Don't duplicate this function. Just simply use the existing function and add a second step for the deletion of the table only. In your original function you can parameterize that function so it takes as input the name of the state store to use!

tests/certification/state/azure/tablestorage/tablestorage_test.go
```diff
@@ -0,0 +1,178 @@
...
+				secretstores_loader.New("local.env", func() secretstores.SecretStore {
+					return secretstore_env.NewEnvSecretStore(log)
+				}),
+			),
+			runtime.WithStates(
+				state_loader.New("azure.tablestorage", func() state.Store {
+					return table.NewAzureTablesStateStore(log)
+				}),
+			))).
+		Step("Run basic test with new table", NonExistingTableTest).
```
berndverst, line 156: This is where you can call the test function with the parameter, and then add a second step for the deletion

berndverst (changes requested):
One change to avoid code duplication, otherwise this looks good (assuming it actually runs :) )

Issue #1772, opened 2022-06-03
RocketMQ fix metadata properties


shubham1172:
## Problem
There are some unused and wrongly implemented metadata in RocketMQ.
| Metadata | Issue | Possible resolution |
|--|--|--|
| AccessProto | unused | Add or remove it |
| GroupName | misguiding | Should be renamed to `ProducerGroupName` |
| ConsumerBatchSize | unused | Add or remove it |
| SendTimeOut | incorrect implementation | Instead of nanoseconds, use milliseconds or seconds. |

## Release Note

RELEASE NOTE: **FIX** RocketMQ metadata properties

PR #1817, opened 2022-06-28
RocketMQ fix metadata properties

helbing:
Signed-off-by: helbing <helbingxxx@gmail.com>

# Description

fix RocketMQ pubsub metadata properties

## Issue reference

Please reference the issue this PR will close: #1772

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


pubsub/rocketmq/rocketmq.go
```diff
@@ -176,7 +182,7 @@ func (r *rocketMQ) Publish(req *pubsub.PublishRequest) error {
 				}
 			}
 
-			ctx, cancel := context.WithTimeout(r.ctx, time.Duration(r.metadata.SendTimeOut))
+			ctx, cancel := context.WithTimeout(r.ctx, time.Duration(r.metadata.SendTimeOut)*time.Second)
```
ItalyPaleAle, line 185: This is another breaking change. In 1.8 we are saying that the value of "sendTimeOut" is in nanoseconds: https://v1-8.docs.dapr.io/reference/components-reference/supported-pubsub/setup-rocketmq/

I do think that using nanoseconds is really bad however

ItalyPaleAle: @berndverst thoughts?

ItalyPaleAle: How about keeping `sendTimeOut` as nanosecond, marking it as depecated (ie. removing it from docs) and adding a new option `sendTimeOutSec`?

helbing: I think this is the better way to deal with it, so I will handle it this way?

ItalyPaleAle: Sounds good to me! You can also continue accepting `sendTimeOut` (as nanoseconds) and show a warning that the option is deprecated. See an example here (although in this case the option was completely ignored, while in your case you can still accept nanoseconds in `sendTimeOut`): https://github.com/dapr/components-contrib/blob/a08988c25ed32263809d9fdacaac90edb7b005a1/pubsub/azure/servicebus/servicebus.go#L244
 And, thank you 😊

pubsub/rocketmq/metadata.go
```diff
@@ -80,13 +85,18 @@ func (s *rocketMQMetaData) Decode(in interface{}) error {
...
 	rMetaData := getDefaultRocketMQMetaData()
 	if metadata.Properties != nil {
 		err := rMetaData.Decode(metadata.Properties)
 		if err != nil {
 			return nil, fmt.Errorf("rocketmq configuration error: %w", err)
 		}
 	}
+
+	if rMetaData.SendTimeOut != 0 {
+		logger.Warn("pubsub.rocketmq: metadata property 'sendTimeOut' has been deprecated and is now ignored - use 'sendTimeOutSec' instead. See: https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-rocketmq/")
```
ItalyPaleAle, line 98: I think the waning is fine but we can still accept this rather than ignoring it.

```
if rMetaData.SendTimeOut != 0 {
	logger.Warn("pubsub.rocketmq: metadata property 'sendTimeOut' has been deprecated - use 'sendTimeOutSec' instead. See: https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-rocketmq/")
	rMetaData.SendTimeOutSec = rMetaData.SendTimeOut / 1000000
}
```

pubsub/rocketmq/rocketmq.go
```diff
@@ -109,6 +112,9 @@ func (r *rocketMQ) setUpProducer() (mq.Producer, error) {
 	if r.metadata.GroupName != "" {
 		opts = append(opts, mqp.WithGroupName(r.metadata.GroupName))
```
ItalyPaleAle, line 113: Should probably show a warning here that groupName is deprecated

helbing: Done

pubsub/rocketmq/metadata.go
helbing: This maybe pose a problem, if both `SendTimeOut` and `SendTimeOutSec` are set, `SendTimeOutSec` will overwritte. What do you think?

addjuarez:
@helbing please look at failing checks

yaron2:
ping @helbing

yaron2:
ping @helbing

berndverst (approved):

berndverst (approved):

PR #1829, opened 2022-06-29
Ensure context propagation in MySQL binding

ItalyPaleAle:
Spin-off from PR adding contexts to input bindings


bindings/mysql/mysql.go
```diff
@@ -147,7 +145,7 @@ func (m *Mysql) Invoke(ctx context.Context, req *bindings.InvokeRequest) (*bindi
 		return nil, errors.Errorf("required metadata not set: %s", commandSQLKey)
 	}
 
-	startTime := time.Now().UTC()
+	startTime := time.Now()
```
berndverst, line 148: Arguably returning the time in a local time now is a breaking change. Why not keep it UTC?

berndverst (approved):

PR #1833, opened 2022-06-30
middleware: changes wasm basic to use waPC

codefromthecrypt:
# Description

Updates wasm basic middleware and simplifies some code inside of it,
notably by using [waPC](https://github.com/wapc/wapc-go).

This changes the entrypoint function to be named "rewrite" and
uses a pool because known memory allocators are not goroutine
safe.

FYI wazero's first beta tag will be in two weeks
https://github.com/tetratelabs/wazero/issues/506#issuecomment-1162489846

## Issue reference

N/A

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [n/a] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_


berndverst:
@daixiang0 can you review this since you know more about WASM and the Wazero library?

Also we want more information on the structure of the WASM that can be loaded. A compiled WASM sample is obviously not good enough :)

daixiang0:
@codefromthecrypt thanks for your contribution! Since `wazero` is not stable now, it will continue to iterate.

Maybe we can get UUID from context. For the logging, I would like to make it detailed to help debug.

Adding more support is my plan, such as support handling HEADER/response changes, while not having much time to do it now.

codefromthecrypt:
@daixiang0 cool I can do a UUID like https://github.com/wapc/wapc-go/blob/master/engines/wazero/wazero.go#L303

Should we start by assuming the wasm binary changes per request? or should we assume it doesn't?

daixiang0:
That is an interesting question, load once would get better performance. If `wazero` supports it, we can use it, and compare SHA256 of the wasm binary, if changes, we reload it.

codefromthecrypt:
OK I've refactored now, though feels there are some API glitches. Have a look at the impl and if there's any more stuff here I can do that before backfilling tests.

* GetHandler isn't context aware, which can make things like tracing or profiling difficult (initialization logic)
* GetHandler is a func, not an interface, which means what's returned can't be closed. This means some leaks
  * One way is to make Middleware closeable and then move the runtime to a field in that.

codefromthecrypt:
## At least a README shoud be made in how to generate the wasm

one other thing I think would be helpful is that the "run" function isn't documented or linked in a README or something (at least not that I can find). I think the source that "hello.wasm" was compiled from should be checked in, with instructions on how it was compiled. (ex. `tinygo build -o hello.wasm -scheduler=none --no-debug -target=wasi hello.go`.

## Alternative ABI

Right now, "run" is allocating memory for the URI using tinygo specific allocation. So, ex this won't work with other language compilers. Another way is to not do allocation externally.

### WASI-only

Instead of passing memory in and out, you might consider using WASI instead. Doing so would be more portable as you can use other languages.

For example, you can set moduleConfig.args = "rewrite_request_uri", uri and moduleConfig.stdout = buf

then just instantiate the module and see if the stdout is empty or not, and if not assign it to SetRequestURIBytes

### WASI+callback

reading stdout is easy, but could result in larger code depending on how people write to it. Another way is to have them call a custom host function if they need to change the URI. So, in this case a host module is instantiated per request which dispatches "set_request_uri(offset, len)" to req.SetRequestURIBytes. The wasm people generate import this function and use that instead of stdout.

### Callback only

instead of having the wasm parse os.Args (ex that "rewrite_request_uri" == arg0), the entrypoint function could be like you have now.. "run" aka "rewrite_request_uri". It wouldn't accept any args, as it is hard to know which memory offsets are unused by the compiler. Instead, there's a host function to get the current uri (ex "get_request_uri"). So on each request "rewrite_request_uri" is invoked which might call "get_request_uri" if it was enabled, and then might call "set_request_uri" if it decided to rewrite it.
 
## Your choice here

There's no perfect way, and also there are other tools that have callback flows like wapc. This is mainly to get the thinking down. Even the current approach is fine, if documented. Just it needs tinygo only for wasm generation. That may be fine!

daixiang0:
Very cool! The basic WASM support is the first step, actually, I am thinking about Envoy WASM ABI implementation, that will create a bridge between service mesh and application runtime. For `basic` design, I think it is simple enough to start with.

Could you add an example in dapr/examples for this and update related documents?

codefromthecrypt:
cool. I'll polish this up where basic means URI rewrite. I'll go ahead and update the exported function name to "rewrite_request_uri", also, unless you've strong feelings towards "run"

daixiang0:
That is ok, go ahead!

codefromthecrypt:
@daixiang0 I've updated the code here.

> Could you add an example in dapr/examples for this and update related documents?

sorry if I'm a bit daft, but can you be precise where you want to update including both the repo and the directory names?

middleware/http/wasm/basic/basic.go
```diff
@@ -137,3 +121,90 @@ func (m *Middleware) getNativeMetadata(metadata middleware.Metadata) (*middlewar
...
+	for _, fn := range []string{wasmFuncMalloc, wasmFuncFree, wasmFuncRewrite} {
+		if mod.ExportedFunction(fn) == nil {
+			return errors.Wrapf(err, "missing exported function %s", fn)
+		}
+	}
+
+	return nil
+}
+
+func (rh *wazeroRequestHandler) requestHandler(h fasthttp.RequestHandler) fasthttp.RequestHandler {
```
codefromthecrypt, line 148: trivia

one thing interesting for future work is basically I know of no allocator in wasm that's thread/goroutine safe as they don't use atomics to update the heap. This means that serious performance would rely on a ready pull who have to do one request and either reset or be thrown away. instantiation cost as it is now is ok as the wasm is simple, but even then you should be careful about the memory settings as certain toolchains/compilers reserve a lot of memory by default.

middleware/http/wasm/basic/basic.go
```diff
@@ -1,33 +1,58 @@
...
+
+	ctx := context.Background()
+
+	// TinyGo needs WASI to implement functions such as panic.
+	if _, err := wasi_snapshot_preview1.Instantiate(ctx, ret.runtime); err != nil {
+		_ = ret.Close()
+		// An error is unexpected as Instantiate won't fail on a new runtime.
+		panic(errors.Wrap(err, "failed to instantiate WASI"))
+	}
+	return
```
Taction, line 55: I prefer to return an error rather than panic, letting the dapr runtime handles the error. So I'd more like keep this func as just `return &Middleware{logger: logger}` and remove the init of the wazero to `GetHandler` func, which would also keep the same behavior with the other middleware components.

codefromthecrypt: So you are ok initializing a new runtime per middleware? That will add unnecessary overhead per call, but fine to keep things simple to start.

codefromthecrypt: I will do it a different way using namespaces. while the above will never error I can see it is distracting!

middleware/http/wasm/basic/basic.go
```diff
@@ -41,86 +57,54 @@ func (m *Middleware) GetHandler(metadata middleware.Metadata) (func(h fasthttp.R
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "error reading wasm at %s", path)
+	}
+
+	compiled, err := m.runtime.CompileModule(ctx, wasm, wazero.NewCompileConfig())
+	if err != nil {
+		return nil, errors.Wrapf(err, "error compiling wasm at %s", path)
+	}
+
+	// Create a new namespace which isolates WASI and module instantiation.
```
codefromthecrypt, line 81: @Taction so this is how I am deferring wasi to middleware instantiation, but still keeping a central runtime. wazero has a "namespace" concept which can allow you to re-instantiate the same module multiple times with the same name (as is the case when we move WASI later in the lifecycle).

daixiang0:
> sorry if I'm a bit daft, but can you be precise where you want to update including both the repo and the directory names?

I mean https://github.com/dapr/samples, you can add a WASM sample.

codefromthecrypt:
@daixiang0 cool for the sample. is there any thing else to do in this repo? I would like to have this land before updating https://github.com/dapr/samples so that the instructions can say the right function name.

middleware/http/wasm/basic/basic.go
```diff
@@ -137,3 +120,96 @@ func (m *Middleware) getNativeMetadata(metadata middleware.Metadata) (*middlewar
...
+}
+
+func (rh *wazeroRequestHandler) requestHandler(h fasthttp.RequestHandler) fasthttp.RequestHandler {
+	return func(ctx *fasthttp.RequestCtx) {
+		requestID := strconv.FormatInt(int64(ctx.ID()), 10)
+
+		mod, err := rh.ns.InstantiateModule(ctx, rh.compiled, wazero.NewModuleConfig().
+			WithName(requestID))
+		if err != nil {
+			log.Fatal(err)
```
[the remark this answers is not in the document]
codefromthecrypt: ok thanks!

codefromthecrypt:
ok ready on my side, when this is merged I'll follow up on the sample no prob!

middleware/http/wasm/basic/basic.go
```diff
@@ -1,33 +1,49 @@
...
 	"github.com/pkg/errors"
 	"github.com/tetratelabs/wazero"
-	"github.com/tetratelabs/wazero/wasi"
+	"github.com/tetratelabs/wazero/wasi_snapshot_preview1"
 	"github.com/valyala/fasthttp"
+)
 
-	"github.com/dapr/components-contrib/middleware"
-	"github.com/dapr/kit/logger"
+// These are undocumented, but exported functions. See tinygo-org/tinygo#2788
```
daixiang0, line 20: Please use the whole link here.

middleware/Readme.md
```diff
@@ -6,4 +6,4 @@ Middleware components provide a way to define middlewares that are executed in a
 
 ## Implementing a new Middleware
 
-A compliant middleware needs to implement the `Middleware` inteface included in the [`middleware.go`](middleware.go) file.
+A compliant middleware needs to implement the `Middleware` interface included in the [`middleware.go`](middleware.go) file.
```
daixiang0, line 9: Please rename it to `README.md`.

daixiang0 (changes requested):
@codefromthecrypt codes look good overall, just some nits then we can make it land.

codefromthecrypt:
@daixiang0 feedback sounds good. I generally make a sub-module readme instead of making a large one for the whole directory. I'll give it a go and if we need another revision no problem.

daixiang0:
Also please fix CI.

codefromthecrypt:
ok I've fixed everything. One thing I would like to say is that while we should merge this I think we should refactor before recutting the documentation.

Generally, I think it is too difficult to use this as generating the wasm is burdensome. People have to know too much for example memory offsets and such. I would recommend instead we either switch to a WASI approach as mentioned [here](https://github.com/dapr/components-contrib/pull/1833#issuecomment-1172811297) or switch to [wapc-go](https://github.com/wapc/wapc-go) cc @pkedy. Either one I'm happy to do as a follow-up.

I also think that it is unlikely someone will be able to use something besides wazero, so allowing that configuration while seemingly open has limited usefulness. For example, if you add wasmtime or wasmer, you'd end up in the same situation of wapc-go where they [cannot load at the same time](https://github.com/wapc/wapc-go/issues/26) due to shared library conflicts.

In summary, I think if the first goal is to show someone something easy, I think we should switch this to wapc-go+wazero and make it easy that way. This would also support languages besides tinygo. For future work, such as proxy-host, going straight to wazero makes more sense as it would be more efficient and the wasm side can leverage existing compilers. However, I'm fine to continue this course however you decide!

daixiang0:
WASM can be considered as a single field although most program things can be brought from other languanges, like Golang/Rust, it still needs some extra knowledge. 

In the first step, I introduce its support, which is really a very initial version needed to iterate. Replace request URI is one function, let's add more functions if you want.

Since now we do not want to enable CGO, the runtime config may be useless, but after CGO support, we can quickly add other WASM runtimes which need CGO.

wapc-go is really cool, you can optimize with it.

codefromthecrypt:
@daixiang0 cool so I will optimize for wapc-go and note that to use other runtimes is a *build flag* not a runtime one. wapc-go works for that (I implemented it!) I think this is a good path because it makes the first time experience less painful. I will add a commit to do this with wapc so we can validate.

codefromthecrypt:
ok I switched to wapc-go and i think you'll agree it is a lot simpler now, both in the host and also how users implement the guest side.

codefromthecrypt:
> Since now we do not want to enable CGO, the runtime config may be useless, but after CGO support, we can quickly add other WASM runtimes which need CGO.

Concretely, WASM runtimes that use CGO (wasmer, wasmtime) import the same WASI C header file. This causes a conflict which prevents them from being compiled in the same binary. This wouldn't be the case if they used different namespace, but it is also unlikely one will change. So, later, keep this in mind as basically dapr would need to be rebuilt for each runtime that conflicts like this. Some may be fine doing that as they are using custom binaries anyway.

ItalyPaleAle:
This shouldn't be a problem. We will not have CGO enabled in Dapr "ever". What we'll do is enable adding other components that communicate via gRPC. Because those will be in different processes, they can use CGO freely and don't risk conflicts.

codefromthecrypt:
thanks for the steering @ItalyPaleAle 

FYI and more for the others on this thread... I bumped https://github.com/dapr/dapr/issues/3787#issuecomment-1186061176 with some thoughts as I think exactly this plugin could help people learn a bit about WebAssembly as a potential plugin model alongside gRPC (at least for things that can be handled locally). Agreed that gRPC for plugin arch is fairly common and well understood means to control dependencies, including build or platform libraries.

daixiang0 (approved):
LGTM

/cc @dapr/maintainers-components-contrib

codefromthecrypt:
here's the doc PR https://github.com/dapr/docs/pull/2663

daixiang0:
@Taction @ItalyPaleAle please help review and merge.

middleware/http/wasm/basic/basic.go
```diff
@@ -137,3 +112,45 @@ func (m *Middleware) getNativeMetadata(metadata middleware.Metadata) (*middlewar
...
+	mod  wapc.Module
+	pool *wapc.Pool
+}
+
+func (rh *wapcRequestHandler) requestHandler(h fasthttp.RequestHandler) fasthttp.RequestHandler {
+	return func(ctx *fasthttp.RequestCtx) {
+		instance, err := rh.pool.Get(1 * time.Second)
+		if err != nil {
+			ctx.Error("wasm pool busy", fasthttp.StatusInternalServerError)
+		}
```
Taction, line 126: missing return here.
```
if err != nil {
			ctx.Error("wasm pool busy", fasthttp.StatusInternalServerError)
                         return
		}
```

Taction:
@codefromthecrypt Thanks for your great work! LGTM overall, just a nit change.

middleware/http/wasm/basic/basic.go
codefromthecrypt: good catch!

Taction (approved):
LGTM

codefromthecrypt:
hi there. I'm wondering if there are any more steps needed. I'm attending a couple devops events (wellington and singapore) over the next couple weeks, and was hoping to point to this as a neat way to start learning about webassembly and extension. Any idea when this and the docs might be mergeable cc @msfussell

middleware/http/wasm/basic/basic.go
```diff
@@ -137,3 +112,46 @@ func (m *Middleware) getNativeMetadata(metadata middleware.Metadata) (*middlewar
...
+}
+
+func (rh *wapcRequestHandler) requestHandler(h fasthttp.RequestHandler) fasthttp.RequestHandler {
+	return func(ctx *fasthttp.RequestCtx) {
+		instance, err := rh.pool.Get(1 * time.Second)
+		if err != nil {
+			ctx.Error("wasm pool busy", fasthttp.StatusInternalServerError)
+			return
+		}
+		defer instance.Close(ctx)
```
Taction, line 128: The pool used here has a minor issue, especially maybe the close of the instance, it seems that the instances resued from the pool don't work as expected.

I updated the tests file to show and catch this error:

[patch.diff.zip](https://github.com/dapr/components-contrib/files/9266031/patch.diff.zip)

Taction:
@codefromthecrypt Also please resolve the conflicts.

middleware/http/wasm/basic/basic.go
codefromthecrypt: thanks. I'll look into it and the problem wherever it is

codefromthecrypt:
there's an API change under play so I'll update this again post merge of https://github.com/wapc/wapc-go/pull/35

codefromthecrypt:
@Taction I backfilled a bunch of tests. you can decide to merge before the wapc refactor because that refactor won't effect how we present configuration here to users anyway. Thanks for the help!

berndverst:
@codefromthecrypt - only components-contrib maintainers (like myself) can merge PRs.

You mentioned you will be updating this PR once https://github.com/wapc/wapc-go/pull/35 has been merged. So for the timing being I will convert this PR to a draft. Once you have it updated, feel free to convert it to a regular PR again and I can help get this reviewed again and then merged. Feel free to mention me or the contrib maintainers tag directly.

codefromthecrypt:
@Taction @daixiang0 I added e2e test dir as I think having this in place makes any bugfixes easier later, as well if we want to transition to some more serious (non-basic) it helps people understand the infrastructure needed. As this is new, you may want to PTAL

otherwise, once https://github.com/dapr/components-contrib/pull/1975 is merged I will rebase one more time and mark for review by a maintainer

berndverst:
https://github.com/dapr/components-contrib/pull/1975 is merged now.

Please note that after rebasing you must use Go 1.19 and if you want to use the linter locally, you need golangci-lint v1.48.0.

codefromthecrypt:
ok all rebased as of.... now! 🤞

codefromthecrypt:
I think the last commit should fix the windows fail (thanks again to @pkedy for the nudge ;). We were testing error messages too strictly and they can vary per OS, such as was the case.

daixiang0 (approved):
LGTM

daixiang0:
Please fix conflicts. @codefromthecrypt

codefromthecrypt:
squashed and rebased @daixiang0!

daixiang0:
Please update tests/certification/bindings/alicloud/dubbo/go.mod as well.

codefromthecrypt:
ok I think Dubbo maybe has some deps to update as I had to use `go mod tidy -compat=1.17` to get it to work. Meanwhile 🤞

daixiang0 (dismissed):
LGTM

/cc @dapr/maintainers-components-contrib  for merge review.

Taction (dismissed):
LGTM

codefromthecrypt:
FYI, I'll update this from merge conflicts again tomorrow after wapc-go uses wazero's first tagged release (1.0.0-beta.1).

Meanwhile, we've also opened a [gophers slack wazero channel](https://gophers.slack.com/archives/C040AKTNTE0) if anyone wants to chat.

codefromthecrypt:
rebased again /cc https://github.com/orgs/dapr/teams/maintainers-components-contrib for merge review.

codefromthecrypt:
rebased :D

go.mod
```diff
@@ -506,7 +507,7 @@ require (
 	github.com/spf13/cast v1.4.1 // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
 	github.com/stretchr/objx v0.4.0 // indirect
-	github.com/tetratelabs/wazero v0.0.0-20220425003459-ad61d9a6ff43
+	github.com/tetratelabs/wazero v1.0.0-pre.1 // indirect
```
codefromthecrypt, line 510: so, on a stable tag, albeit pre-release https://github.com/tetratelabs/wazero/releases/tag/v1.0.0-pre.1

codefromthecrypt:
fossa scan looks like a timeout. could someone 🥾 it?

codefromthecrypt:
failure was Kafka maybe it can be kicked

```
=== FAIL: . TestKafka_with_retry (unknown)
[52](https://github.com/dapr/components-contrib/runs/8212909954?check_suite_focus=true#step:14:53)
```

@dapr/maintainers-components-contrib I know folks are busy, but I admit it is getting tiring to rebase or find unrelated build glitches every few days. I'd love to help in more productive ways than this, ex making a demo etc. can someone please help land this?

codefromthecrypt:
fyi @berndverst ps not sure how to tag the team as https://github.com/orgs/dapr/teams/maintainers-components-contrib is a dead link for me.

codefromthecrypt:
@daixiang0 ps since you added this comment, there's now an experimental compiler cache option https://github.com/dapr/components-contrib/pull/1833#issuecomment-1171959542

https://pkg.go.dev/github.com/tetratelabs/wazero/experimental#example-package-WithCompilationCacheDirName

note: this requires 5-7x the size of wasm to disk. The actual size is hard to say as it is more about the complexity of the code which we can't control.

If you want me to add this now I can, though I'd need to update docs etc. Otherwise, it can be done later. In any case, this only helps speed up restarts.

daixiang0:
> @daixiang0 ps since you added this comment, there's now an experimental compiler cache option [#1833 (comment)](https://github.com/dapr/components-contrib/pull/1833#issuecomment-1171959542)
> 
> https://pkg.go.dev/github.com/tetratelabs/wazero/experimental#example-package-WithCompilationCacheDirName
> 
> note: this requires 5-7x the size of wasm to disk. The actual size is hard to say as it is more about the complexity of the code which we can't control.
> 
> If you want me to add this now I can, though I'd need to update docs etc. Otherwise, it can be done later. In any case, this only helps speed up restarts.

Thanks for your input! Let's do it in another pr.

daixiang0:
> fyi @berndverst ps not sure how to tag the team as https://github.com/orgs/dapr/teams/maintainers-components-contrib is a dead link for me.

Maybe you need to become a team member first, see dapr/community.

codefromthecrypt:
ps maybe the below error can be ignored until fixed elsewhere?

```
dockercompose.go:90: Starting kafka3 ... 
[6462](https://github.com/dapr/components-contrib/runs/8219004661?check_suite_focus=true#step:14:6463)
        Host is already in use by another container
[6463](https://github.com/dapr/components-contrib/runs/8219004661?check_suite_focus=true#step:14:6464)
        Starting kafka3 ... error
[6464](https://github.com/dapr/components-contrib/runs/8219004661?check_suite_focus=true#step:14:6465)
        
[6465](https://github.com/dapr/components-contrib/runs/8219004661?check_suite_focus=true#step:14:6466)
        ERROR: for kafka3  Cannot start service kafka3: driver failed programming external connectivity on endpoint kafka-3 (c4ac5d642e0e70a8d524101dd817f3a408a977c8d1363d4085fda86102a703b0): Error starting userland proxy: listen tcp4 0.0.0.0:39092: bind: address already in use
[6466](https://github.com/dapr/components-contrib/runs/8219004661?check_suite_focus=true#step:14:6467)
        No containers to start
[6467](https://github.com/dapr/components-contrib/runs/8219004661?check_suite_focus=true#step:14:6468)
        1
```

codefromthecrypt:
I'm unsubscribing as it is just merge bot stuff and fuzz from other broke things. If someone needs my help, please mention me!

berndverst (approved):

berndverst:
@codefromthecrypt I will merge this now. Can you please make any documentation updates and create the sample you were discussing previously in this issue.

codefromthecrypt:
@berndverst thanks for owning getting this in, and of course thanks to @daixiang0 @Taction @ItalyPaleAle for review help also.

Good news is we already completed the docs based on this PR before and nothing drifted since then. https://docs.dapr.io/reference/components-reference/supported-middleware/middleware-wasm/

I'll follow-up with the sample once things merge, no prob!

codefromthecrypt:
thanks for merging @artursouza! I opened this to bump deps and once in I'll do the sample mentioned in the desc https://github.com/dapr/dapr/pull/5169

PR #1840, opened 2022-07-01
Do not hardcode specific testTopicName

ItalyPaleAle:
# Description

Makes it easier to run Event Hubs conformance tests locally (outside of CI). There is no need to use the name prefix in the topic name (in fact, the other topics do not use a prefix), so if we don't use the name prefix, we can skip hardcoding it.

Note: I have not applied the updated Bicep to the deployed infrastructure. But tested against a separate infra.


tanvigour (approved):
LGTM

berndverst (approved):

PR #1841, opened 2022-07-01
[release-1.8] Azure Blob Storage state store conformance tests

ItalyPaleAle:
# Description

Accidentally discovered that there were no conformance tests for the state.azure.blobstorage component. Once I added the tests (included in this PR), the component was failing, primarily with issues related to managing etags.

Because this component is a stable one, this PR targets release-1.8

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Extended the documentation: N/A

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -105,6 +105,8 @@ jobs:
         #
         # Only list the secrets you need for the component.
         CRON_COMPONENTS=$(yq -I0 --tojson eval - << EOF
+        - component: state.azure.blobstorage
+          required-secrets: AzureBlobStorageAccessKey,AzureBlobStorageAccount
         - component: state.azure.cosmosdb
           required-secrets: AzureCosmosDBMasterKey,AzureCosmosDBUrl,AzureCosmosDB,AzureCosmosDBCollection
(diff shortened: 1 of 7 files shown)


.github/workflows/conformance.yml
```diff
@@ -118,7 +120,7 @@ jobs:
         - component: pubsub.azure.servicebus
           required-secrets: AzureServiceBusConnectionString
         - component: bindings.azure.blobstorage
-          required-secrets: AzureBlobStorageAccessKey,AzureBlobStorageAccount,AzureBlobStorageContainer
```
ItalyPaleAle, line 121: This didn't need to be a secret and could be hardcoded (for simplicity i.e. when running locally)

berndverst:
After carefully examining this PR I have determined that this primarily addresses:

- Not throwing an error for a noop delete operation
- Not throwing the wrong error types
- And it handles an interaction between the use of ETAG and FirstWrite that was not handled before.

Otherwise this component is working as intended. I do not see sufficient impact to warrant backporting these changes to 1.7.X and 1.6.X, but we will get them into 1.8.0

berndverst (approved):

Issue #1037, opened 2021-07-28
Local keychain secret store


artursouza:
Create a local store that uses the OS' keychain solution and not simply file store

saranvasu:
is this issue still open? Can I start working on this?

msfussell:
@saranvasu - Yes, this issue is still open

helbing:
@saranvasu Do you work on this now?

ItalyPaleAle:
> @helbing and @ItalyPaleAle are there plans to give this issue a second chance? Maybe with a new implementation?

If there's a way to implement this without the need for CGo, I'd accept it. Contributions are welcome :)

PR #1848, opened 2022-07-05
Local keyring secret store

helbing:
# Description

Implement keyring secret store

## Issue reference

Please reference the issue this PR will close: #1037

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


secretstores/local/keyring/keyring.go
```diff
@@ -0,0 +1,206 @@
...
+	SecretServiceBackend: true,
+	KeychainBackend:      true,
+	KeyCtlBackend:        true,
+	KWalletBackend:       true,
+	WinCredBackend:       true,
+	FileBackend:          true,
+	PassBackend:          true,
+}
+
+type keyringSecretStoreMetaData struct {
```
daixiang0, line 52: Does it mean we can use multiple backends for one component at the same time? Could you share the cases like it?

secretstores/local/keyring/keyring.go
```diff
@@ -0,0 +1,206 @@
...
+	"github.com/dapr/kit/config"
+	"github.com/dapr/kit/logger"
+)
+
+// BackendType is alias of keyring.BackendType.
+type BackendType = keyring.BackendType
+
+// All currently supported secure storage backends.
+const (
+	SecretServiceBackend BackendType = keyring.SecretServiceBackend // secret-service
```
daixiang0, line 32: I would not alias here, the old name is meaningful enough.

helbing: Fine, I will change it.

secretstores/local/keyring/keyring.go
```diff
@@ -0,0 +1,206 @@
...
+	KeychainBackend      BackendType = keyring.KeychainBackend      // keychain
+	KeyCtlBackend        BackendType = keyring.KeyCtlBackend        // keyctl
+	KWalletBackend       BackendType = keyring.KWalletBackend       // kwallet
+	WinCredBackend       BackendType = keyring.WinCredBackend       // wincred
+	FileBackend          BackendType = keyring.FileBackend          // file
+	PassBackend          BackendType = keyring.PassBackend          // pass
+)
+
+// availableBackends is a map of backends that keyring secret store are supported.
+var availableBackends = map[BackendType]bool{
```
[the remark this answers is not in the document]
helbing: Just sync with keyring library.

secretstores/local/keyring/keyring_test.go
```diff
@@ -0,0 +1,102 @@
...
+		m.Properties = map[string]string{
+			"backendType": string(backendType),
+			"fileDir":     "./",
+		}
+		err := s.Init(m)
+		assert.Nil(t, err)
+	})
+}
+
+func TestGetSecret(t *testing.T) {
```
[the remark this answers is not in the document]
daixiang0: We have built and tested for those arch, you can detect OS and run it in the corresponding arch.

For kde we can mark TODO for not. Also you can check tests in the keyring to get some ideas.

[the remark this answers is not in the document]
helbing: Done

secretstores/local/keyring/keyring_test.go
```diff
@@ -0,0 +1,382 @@
+/*
```
ItalyPaleAle, line 1: Great that you added tests! But, would you consider moving these to be certification tests?

I am concerned with two things:

1. On local machines, if users don't have the dependencies installed (gnupg etc) tests may fail when they run `make test`. We want to make sure that contributors have the best experience even when developing locally, and we should avoid adding dependencies for running unit tests
2. On the CI, installing the dependencies will slow down the test runs for a bit.

These tests are fairly comprehensive already and I think they would be great as certification tests

secretstores/local/keyring/keyring.go
```diff
@@ -0,0 +1,192 @@
...
+limitations under the License.
+*/
+
+package keyring
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/99designs/keyring"
```
ItalyPaleAle, line 20: It looks like this depends on `github.com/99designs/go-keychain` which requires CGO.

Can you confirm that this PR works **without CGO** (ie. when building and compiling with `CGO_ENABLED=0`) on all 3 platforms? Sadly, using CGO is not possible with Dapr.

ItalyPaleAle (changes requested):

secretstores/local/keyring/keyring.go
helbing: My fault, I didn't notice that Dapr doesn't support CGO. It seems that except for backends of file and pass,  all require CGO. So this feature, local keyring secret store, should not be needed?

ItalyPaleAle:
@helbing  thanks for the contribution :)

As mentioned in the comment there, we cannot currently accept dependencies that use CGO in Dapr.

We are currently working on a new way to import components so they're out-of-process, and that would allow using CGO as well. You can track the progress here: https://github.com/dapr/dapr/issues/3787

Other proposed components are pending on that, for example SQLite: https://github.com/dapr/components-contrib/pull/1654

ItalyPaleAle (changes requested):
On hold until gRPC components are available

daixiang0:
@helbing @ItalyPaleAle I covert it as draft until gRPC components are available.

ItalyPaleAle:
👋 bot!
Keeping this alive

berndverst:
@helbing take a look at
https://github.com/dapr-sandbox/components-go-sdk
https://docs.dapr.io/developing-applications/develop-components/pluggable-components/

ItalyPaleAle:
I am going to close this PR since it cannot be merged in this repository due to the CGo requirement.

Even with external (pluggable) secret store components, we wouldn't be able to accept this in-tree, as pluggable components are released separately.

Thanks for the contribution and sorry we cannot accept it at this time!

PR #1852, opened 2022-07-05
Azure AD support in SignalR

### LINKED ISSUE #1843: Add support for Azure AD in SignalR binding
The Azure SignalR is currently not supporting Azure AD authentication.

Upon investigation, this should be possible. SignalR authenticates requests by means of a JWT, which can either be signed with HMAC using a shared key, or can be obtained through Azure AD. Relevant docs here: https://docs.microsoft.com/en-us/azure/azure-signalr/signalr-reference-data-plane-rest-api

The SignalR component doesn't use a SDK, instead it just makes REST calls to the service. So we're not dependent on any SDK update to support AAD auth.

Additionally, we already have the infrastructure in place (through the code in the `authentication/azure` folder) to authenticate with Azure AD and obtain (and refresh) access tokens.

ItalyPaleAle:
# Description

Adds support for Azure AD in SignalR and fixes a few bugs.

## Issue reference

Closes #1843

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Docs: https://github.com/dapr/docs/issues/2613


bindings/apns/authorization_builder.go
```diff
@@ -17,7 +17,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/golang-jwt/jwt"
+	jwt "github.com/golang-jwt/jwt/v4"
```
ItalyPaleAle, line 20: The package at /jwt has been deprecated and now they use versioning. Despite the version increase, it's fully backwards-compatible. Other parts of dapr already use /jwt/v4

bindings/azure/signalr/signalr.go
```diff
@@ -22,103 +22,168 @@ import (
...
-}
-
 // SignalR is an output binding for Azure SignalR.
 type SignalR struct {
-	endpoint   string
-	accessKey  string
-	version    string
-	hub        string
-	userAgent  string
-	tokens     map[string]signalrCachedToken
```
ItalyPaleAle, line 62: We don't need to cache tokens generated via SHA-256. The way the cache was implemented was not safe for concurrent use, and regardless, signing a JWT requires 1 iteration of SHA-256 which is incredibly fast... so better save the memory :)

authentication/azure/auth.go
```diff
@@ -63,6 +63,9 @@ func NewEnvironmentSettings(resourceName string, values map[string]string) (Envi
 		// The resource name to request a token is https://eventhubs.azure.net/, and it's the same for all clouds/tenants.
 		// Kafka connection does not factor in here.
 		es.Resource = "https://eventhubs.azure.net"
+	case "signalr":
+		// Azure SignalR (data plane)
+		es.Resource = "https://signalr.azure.com"
```
berndverst, line 68: Is SignalR available in other clouds and does this identifier need to change accordingly?

PR #1862, opened 2022-07-07
Changed dapr bot to merge without squashing

ItalyPaleAle:
The behavior of Dapr Bot was to squash merge PRs. Although sometimes useful, squashing rewrites history and is likely to cause merge conflicts if another PR depends on that.

Leaving this here if maintainers think this makes sense :)

### DIFF
--- a/.github/scripts/automerge.py
+++ b/.github/scripts/automerge.py
@@ -40,7 +40,7 @@ for pr in pulls:
         # Merge only one PR per run.
         print(f"Merging PR {pr.html_url}")
         try:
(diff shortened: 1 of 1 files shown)


berndverst:
Generally our team wants to squash however for cleaner history. Personally I don't want the commit history to be lots of 💄 or things like "fix linter".

I would recommend not opening tons of PRs that depend on each other @ItalyPaleAle  :)

I'm not in favor of merging without squashing.

ItalyPaleAle:
I hear you. I will close it. It was worth a shot however :)

Issue #1866, opened 2022-07-08
SendGrid Output Binding always returns 400


benc-uk:
## Expected Behavior

SendGrid output binding will send emails to SendGrid without error

## Actual Behavior

Error is always returned
```
error from SendGrid, sending email failed: 400 {Errors:[{Field:content.0.value Message:The content value must be a string at least one character in length. Help:http://sendgrid.com/docs/API_Reference/Web_API_v3/Mail/errors.html#message.content.value}]}
```

## Steps to Reproduce the Problem

- Create SendGrid output binding in your ~/.dapr/components, with your SendGrid API key, see comments in gist below
- See sample code here [sendgrid-dapr.go](https://gist.githubusercontent.com/benc-uk/1f79c53e1497e828e9df57abb4d33d5f/raw/feafe184415328fbd27d2efe29508b00a4824bad/sendgrid-dapr.go)
- Run with `dapr run`

## Release Note

RELEASE NOTE: **FIX** SendGrid output binding content error.

PR #1867, opened 2022-07-08
Fixes broken SendGrid binding

benc-uk:
# Description

The current SendGrid binding is broken, it returns a 400 error and `Field:content.0.value Message:The content value must be a string at least one character in length`

This is due to a change in data being passed into the binding and the behaviour of `strconv.Unquote` previously it was quoted, so the call to Unquote was required and it passed, now it is failing as the data is not quoted.

## Issue reference

Please reference the issue this PR will close: #1866

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


ItalyPaleAle (dismissed):

ItalyPaleAle:
Thanks for your contribution :)

artursouza (changes requested):
Is there a chance the issue is from the SDK you are using? Why is this data coming as quoted?

docs/developing-component.md
```diff
@@ -22,6 +22,8 @@ git clone https://github.com/dapr/components-contrib.git dapr/components-contrib
 
 ```
 
+Note. When adding commits to the components-contrib repo, ensure you add `--signoff` or `-s` to the commits otherwise the PR you submit may be blocked.
```
berndverst, line 25: Please add something about the DCO requirement and the actions to take as seen here instead:
https://github.com/dapr/dapr/blob/master/CONTRIBUTING.md

berndverst:
> Is there a chance the issue is from the SDK you are using? Why is this data coming as quoted?

@artursouza that's exactly the problem... whoever authored this was quoting the string, but data should generally *not* be coming in quotes, which results in this error and hence this PR is necessary.

bindings/twilio/sendgrid/sendgrid.go
```diff
@@ -185,7 +185,11 @@ func (sg *SendGrid) Invoke(ctx context.Context, req *bindings.InvokeRequest) (*b
 	}
 
 	// Email body is held in req.Data, after we tidy it up a bit
-	emailBody, _ := strconv.Unquote(string(req.Data))
+	emailBody, err := strconv.Unquote(string(req.Data))
+	if err != nil {
+		// Unquote will error if the string is not quoted (not exactly graceful!), so fallback using the string as is
```
berndverst, line 190:
```suggestion
		// Unquote will error if the string is not quoted, so fallback using the string as is
```

berndverst (changes requested):

artursouza (approved):

docs/developing-component.md
benc-uk: Done

benc-uk:
> > Is there a chance the issue is from the SDK you are using? Why is this data coming as quoted?
> 
> @artursouza that's exactly the problem... whoever authored this was quoting the string, but data should generally _not_ be coming in quotes, which results in this error and hence this PR is necessary.

I originally authored this binding, and when I did, the data was definitely quoted (hence me adding `Unquote`), but this was over 2 years ago, and the way data is passed to bindings in internally by Dapr may have changed in that time

benc-uk:
The "bindings.kafka certification" tests are failing, this seems completely unrelated to my change!

ItalyPaleAle:
Not sure what happened here, but perhaps while you were fixing DCO you accidentally brought in a lot of commits that changed history. there are now 86 files changed and including reverting changes that are in master already :( 

BUT... let's fix it :)

Looks like there's only 1 relevant commit in this PR that needs to be merged: https://github.com/dapr/components-contrib/pull/1867/commits/3a469d71f240e277d34d0aa988604ef7a836854d

So here's how you can fix it:

```sh
# Make sure you're synced with the upstream repo (ie. dapr/components-contrib)
git fetch upstream
# Reset to the upstream/master branch
git reset --hard upstream/master
# Apply your commit (note the `.patch` at the end)
curl -L "https://github.com/dapr/components-contrib/pull/1867/commits/3a469d71f240e277d34d0aa988604ef7a836854d.patch" | git am
# Now fix the DCO if needed
git commit --amend --no-edit --signoff
# Force-push to your branch
git push --force
```

This should hopefully fix it

PS: (@berndverst  need more reasons to disable squash merging in the bot? ;) )

benc-uk:
@ItalyPaleAle I'm not sure what happened either, the DCO bot started going red again, when you click the 'Details' link there's some instructions in there to do a git reset HEAD~ followed by a rebase. It seemed to be that

Anyhow, the reset hard, followed by the patch has fixed it. THANKS!

berndverst (approved):

Issue #1868, opened 2022-07-08
add a new log storage bindings component


RcXu:
## Describe the feature
This is a AliCloud Sls Log Storage Binding component for a Multi Language System，such as Nodejs as GW and Java apply the services and so on. It will take more work on store logs which are in the same format, but the Dapr outbinding make it easy.

RELEASE NOTE:
RELEASE NOTE: **ADD** New AliCloud SlsLogStorage Component

PR #1869, opened 2022-07-08
add a new log storage component

RcXu:
# Description

This is a AliCloud Sls Log Storage Binding component for a Multi Language System，such as Nodejs as GW and Java apply the services and so on. It will take more work on storing logs which are in the same format, but the Dapr outbinding make it easy.

## Issue reference
https://github.com/dapr/components-contrib/issues/1868


berndverst:
Please fix DCO by signing off on all of your commits (see the failing DCO test for details)

berndverst (changes requested):
Please fix DCO, and fix linter errors (`make lint`)

PR #1873, opened 2022-07-09
add a new binding component : alicloud sls log storage

RcXu:
Signed-off-by: RcXu <honeyxhl@163.com>

# Description
This is a AliCloud Sls Log Storage Binding component for a Multi Language System，such as Nodejs as GW and Java apply the services and so on. It will take more work on storing logs which are in the same format, but the Dapr outbinding make it easy.

## Issue reference

Please reference the issue this PR will close: [#_[issue number]_](https://github.com/dapr/components-contrib/issues/1872)


bindings/alicloud/sls/sls.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	if source := req.Metadata["source"]; source == "" {
+		return nil, fmt.Errorf("SLS binding error: source property not supplied")
+	}
+
+	ch := make(chan os.Signal, 1)
+	signal.Notify(ch, syscall.SIGTERM, os.Interrupt)
+
+	// Start producer instancce
+	s.producer.Start()
```
DeepanshuA, line 74: Should we take out Producer start in separate method to make `Invoke` look clean and crisp.

bindings/alicloud/sls/sls.go
```diff
@@ -0,0 +1,127 @@
...
+	log, err := s.parseLog(req)
+	if err != nil {
+		s.logger.Info(err)
+		return nil, err
+	}
+
```
DeepanshuA, line 80: I think it would be better to check explicitly that request only is for supported Operation (currently CreateOperation), else throw UnsupportedOperation error

bindings/alicloud/sls/sls.go
```diff
@@ -0,0 +1,127 @@
...
+	// the config properties in the component yaml file
+	producerConfig.Endpoint = m.Endpoint
+	producerConfig.AccessKeyID = m.AccessKeyID
+	producerConfig.AccessKeySecret = m.AccessKeySecret
+	s.producer = producer.InitProducer(producerConfig)
+	return nil
+}
+
+func NewAliCloudSlsLogstorage(logger logger.Logger) *AliCloudSlsLogstorage {
+	logger.Debug("new a Slslog component object")
```
ItalyPaleAle, line 48: I think the message is missing a verb? For example "initialized new slslog binding"

bindings/alicloud/sls/sls.go
ItalyPaleAle: Another thing to note is that this is working on `s.producer` which is a property of the object, and it's starting and stopping a producer in the same method.
Because there could be multiple calls to `Invoke` at the same time, this is not concurrency-safe

ItalyPaleAle (changes requested):
Please address reviewers' feedback

bindings/alicloud/sls/sls.go
RcXu: there is standard verify func int the runtime,so may be it’s not necessary at here，I think

`if binding, ok := a.outputBindings[name]; ok {
		ops := binding.Operations()
		for _, o := range ops {
			if o == req.Operation {
				var resp *bindings.InvokeResponse
				policy := a.resiliency.ComponentOutboundPolicy(a.ctx, name)
				err := policy(func(ctx context.Context) (err error) {
					resp, err = binding.Invoke(ctx, req)
					return err
				})
				return resp, err
			}
		}
		supported := make([]string, 0, len(ops))
		for _, o := range ops {
			supported = append(supported, string(o))
		}
		return nil, errors.Errorf("binding %s does not support operation %s. supported operations:%s", name, req.Operation, strings.Join(supported, " "))
	}`

RcXu: I did that before, but it didn't work. get the nonsupport error before invoke

bindings/alicloud/sls/sls.go
RcXu: I have fixed this problem

bindings/alicloud/sls/sls.go
```diff
@@ -0,0 +1,140 @@
...
+	producerConfig.AccessKeyID = m.AccessKeyID
+	producerConfig.AccessKeySecret = m.AccessKeySecret
+	s.producer = producer.InitProducer(producerConfig)
+
+	s.startSlsProducer()
+	return nil
+}
+
+func (s *AliCloudSlsLogstorage) startSlsProducer() {
+	notifyChan := make(chan os.Signal, 1)
```
ItalyPaleAle, line 53: What is this channel used for?

bindings/alicloud/sls/sls.go
```diff
@@ -0,0 +1,140 @@
...
+	var logInfo map[string]string
+	err := json.Unmarshal(req.Data, &logInfo)
+	if err != nil {
+		return nil, err
+	}
+	return producer.GenerateLog(uint32(time.Now().Unix()), logInfo), nil
+}
+
+func (s *AliCloudSlsLogstorage) parseMeta(metadata bindings.Metadata) (*SlsLogstorageMetadata, error) {
+	b, err := json.Marshal(metadata.Properties)
```
ItalyPaleAle, line 109: Instead of the two round of JSON, please use `config.Decode`. Example: https://github.com/dapr/components-contrib/blob/d38c7867715fd5d2d0d788fa440b22dd8440e49f/bindings/zeebe/client.go#L72

ItalyPaleAle (changes requested):

bindings/alicloud/sls/sls.go
RcXu: this channel is used to test if could close the sis producer. remove it  now

bindings/alicloud/sls/sls.go
RcXu: ok, fixed it already

bindings/alicloud/sls/sls.go
DeepanshuA: Ok, I get that. There are several examples like following, where-in explicitly check is done: https://github.com/dapr/components-contrib/blob/d38c7867715fd5d2d0d788fa440b22dd8440e49f/bindings/alicloud/rocketmq/rocketmq.go#L204

But, runtime should be able to take care of this.

bindings/alicloud/sls/sls.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	s.logger.Debug(log)
+	callBack := &Callback{}
+	err = s.producer.SendLogWithCallBack(req.Metadata["project"], req.Metadata["logstore"], req.Metadata["topic"], req.Metadata["source"], log, callBack)
+	if err != nil {
+		s.logger.Info(err)
+		return nil, err
+	}
+	return nil, nil
```
ItalyPaleAle, line 86: Is there really nothing we can return to the consumer? Can we return an object like `{"status":"ok"}`? No return at all seems weird

ItalyPaleAle:
Thanks for making the changes :) I just have one last comment but otherwise LGTM

bindings/alicloud/sls/sls.go
RcXu: the Sls producer api is asynchronous mode, so implemented a callback interface to accept the response of success or faild and record faliure reason to the console log. if return an ok response in the invoke func , may be not ok yet and make the caller misunderstand. 
on the other hand, at most time we may be not so mind whether the log storage is ok…so design to return nil at here.

berndverst (dismissed):

berndverst (approved):

Issue #1541, opened 2022-02-28
Azure Configuration Service component


artursouza:
## Describe the proposal

Implement Azure Configuration Service component for the configuration API.

PR #1874, opened 2022-07-11
[WIP] azure appconfig configuration store

cmendible:
Signed-off-by: cmendible <cmendible@gmail.com>

# Description

Implementing azure appconfig configuration store.

## Issue reference

Please reference the issue this PR will close: #1541

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


internal/authentication/azure/auth.go
```diff
@@ -42,30 +44,37 @@ func NewEnvironmentSettings(resourceName string, values map[string]string) (Envi
...
 		// Azure EventHubs (data plane)
 		// For documentation https://docs.microsoft.com/en-us/azure/event-hubs/authorize-access-azure-active-directory#overview
 		// The resource name to request a token is https://eventhubs.azure.net/, and it's the same for all clouds/tenants.
 		// Kafka connection does not factor in here.
 		es.Resource = "https://eventhubs.azure.net"
-	case "signalr":
+	case resourceName == "signalr":
 		// Azure SignalR (data plane)
 		es.Resource = "https://signalr.azure.com"
+	case strings.HasSuffix(resourceName, "azconfig.io"):
```
ItalyPaleAle, line 71: Please don't change this switch to look for suffixes. The idea of `resourceName` is that it should be a constant string such as `eventhubs`, `storage`, etc (perhaps we should make them real constants instead). Define a constant such as `"appconfig"` and use that.

Reason is that suffixes such as `azconfig.io` may be different in other sovereign clouds

configuration/azure/appconfig/appconfig.go
```diff
@@ -0,0 +1,190 @@
...
+
+	return m, nil
+}
+
+func (r *ConfigurationStore) Get(ctx context.Context, req *configuration.GetRequest) (*configuration.GetResponse, error) {
+	keys := req.Keys
+	items := make([]*configuration.Item, 0, len(keys))
+	for _, key := range keys {
+		resp, err := r.client.GetSetting(
+			context.TODO(),
```
ItalyPaleAle, line 159: Please pass the context here!

internal/authentication/azure/auth.go
cmendible: I have to run some test to check if asking for a token for resource `https://azconfig.io` will make the cut here.

cmendible: It does!!! 😎

cmendible (commented):
Done!

configuration/azure/appconfig/appconfig.go
```diff
@@ -0,0 +1,190 @@
...
+type ConfigurationStore struct {
+	client   *azappconfig.Client
+	json     jsoniter.API
+	metadata metadata
+
+	logger logger.Logger
+}
+
+// NewAxureAppConfigurationStore returns a new Azure App Configuration store.
+func NewAxureAppConfigurationStore(logger logger.Logger) configuration.Store {
```
JorTurFer, line 57: Is Axure a typo?

configuration/azure/appconfig/appconfig.go
```diff
@@ -0,0 +1,190 @@
...
+			return err
+		}
+
+		r.client, err = azappconfig.NewClient(r.metadata.host, cred, &options)
+		if err != nil {
+			return err
+		}
+	}
+
+	return err
```
ItalyPaleAle, line 108: Nit: if we're here, `err` is certainly `nil`, so I'd just do `return nil` to make it clear.

configuration/azure/appconfig/appconfig.go
```diff
@@ -0,0 +1,190 @@
...
+		m.retryDelay = time.Duration(parsedVal)
+	}
+
+	return m, nil
+}
+
+func (r *ConfigurationStore) Get(ctx context.Context, req *configuration.GetRequest) (*configuration.GetResponse, error) {
+	keys := req.Keys
+	items := make([]*configuration.Item, 0, len(keys))
+	for _, key := range keys {
```
ItalyPaleAle, line 157: The only problem I see here is that we're making requests to the service one at a time.

This method is primarily used to retrieve the initial config state at startup.

Assume that 2 apps are starting up and are requesting the configuration. In the meanwhile, someone is changing the configuration.

Because there's no locking, it's possible that the two apps could get a different configuration as they loop through all keys.

Is there no way to get all keys at once, atomically?

cmendible: @ItalyPaleAle  there seems to be no way to get everything at once.

We could use the [NewListRevisionsPager](https://github.com/Azure/azure-sdk-for-go/tree/sdk/appconfig/azappconfig/v0.3.2/sdk/appconfig/azappconfig/#list-configuration-setting-revisions) method together with a [filter](https://docs.microsoft.com/en-us/azure/azure-app-configuration/rest-api-key-value#filtering) by key.

Given the 5 keys per filter limitations, code would end up making 1 call for each block of 5 or less keys present in the request.

Thoughts?

ItalyPaleAle: Uhm, I don't think that would solve the issue.

I guess this is what it is. Maybe just make sure this behavior is documented?

berndverst:
Is this still work in progress? LGTM.

configuration/azure/appconfig/appconfig.go
cmendible: Referenced this review here: https://github.com/dapr/docs/issues/2660

configuration/azure/appconfig/appconfig.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	return &configuration.GetResponse{
+		Items: items,
+	}, nil
+}
+
+func (r *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.SubscribeRequest, handler configuration.UpdateHandler) (string, error) {
+	subscribeID := uuid.New().String()
+	return subscribeID, nil
```
mukundansundar, line 190: Should error be returned since u subscribe and unsubscribe are not implemented?

configuration/azure/appconfig/appconfig.go
```diff
@@ -0,0 +1,195 @@
...
+	return m, nil
+}
+
+func (r *ConfigurationStore) Get(ctx context.Context, req *configuration.GetRequest) (*configuration.GetResponse, error) {
+	keys := req.Keys
+	items := make([]*configuration.Item, 0, len(keys))
+	for _, key := range keys {
+		resp, err := r.client.GetSetting(
+			ctx,
+			key,
```
mukundansundar, line 165: There is support for querying multiple keys 5 of them at a time in the service... https://docs.microsoft.com/en-us/azure/azure-app-configuration/concept-key-value#query-key-values

Is that something that needs to be looked at for this request?

configuration/azure/appconfig/appconfig.go
```diff
@@ -0,0 +1,195 @@
...
+	maxRetryDelay        = "maxRetryDelay"
+	defaultMaxRetries    = 3
+	defaultRetryDelay    = time.Second * 4
+	defaultMaxRetryDelay = time.Second * 120
+)
+
+// ConfigurationStore is a Azure App Configuration store.
+type ConfigurationStore struct {
+	client   *azappconfig.Client
+	json     jsoniter.API
```
ItalyPaleAle, line 49: This property doesn't seem to be used anywhere, unless I'm mistaken?

configuration/azure/appconfig/appconfig.go
```diff
@@ -0,0 +1,195 @@
...
+
+	if m.connectionString != "" && m.host != "" {
+		return m, fmt.Errorf("azure store error: can't set both %s and %s fields in metadata", host, connectionString)
+	}
+
+	if m.connectionString == "" && m.host == "" {
+		return m, fmt.Errorf("azure store error: specify %s or %s field in metadata", host, connectionString)
+	}
+
+	m.maxRetries = defaultMaxRetries
```
ItalyPaleAle, line 129: These properties (maxRetries, maxRetryDelay, retryDelay) aren't used and can be removed (also the Azure SDKs handle retrying built-in)

cmendible: SDKs has the defaults values and I think we can give user the options to change them. 

I had missed setting the retry properties in code but this is how options look in the latest commit 👇

``` golang
        coreClientOpts := azcore.ClientOptions{
		Telemetry: policy.TelemetryOptions{
			ApplicationID: "dapr-" + logger.DaprVersion,
		},
		Retry: policy.RetryOptions{
			MaxRetries:    int32(m.maxRetries),
			RetryDelay:    m.maxRetryDelay,
			MaxRetryDelay: m.maxRetryDelay,
		},
	}
```

configuration/azure/appconfig/appconfig.go
```diff
@@ -0,0 +1,197 @@
...
+	if val, ok := meta.Properties[host]; ok && val != "" {
+		m.host = val
+	}
+
+	if val, ok := meta.Properties[connectionString]; ok && val != "" {
+		m.connectionString = val
+	}
+
+	if m.connectionString != "" && m.host != "" {
+		return m, fmt.Errorf("azure store error: can't set both %s and %s fields in metadata", host, connectionString)
```
mukundansundar, line 124:
```suggestion
		return m, fmt.Errorf("azure appconfig error: can't set both %s and %s fields in metadata", host, connectionString)
```

configuration/azure/appconfig/appconfig.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	if m.connectionString == "" && m.host == "" {
+		return m, fmt.Errorf("azure store error: specify %s or %s field in metadata", host, connectionString)
```
mukundansundar, line 128:
```suggestion
		return m, fmt.Errorf("azure appconfig error: specify %s or %s field in metadata", host, connectionString)
```

configuration/azure/appconfig/appconfig.go
mukundansundar: and similarly for other errors ...

configuration/azure/appconfig/appconfig_test.go
```diff
@@ -0,0 +1,178 @@
...
+		assert.True(t, ok)
+		assert.Equal(t, testProperties[connectionString], cs.metadata.connectionString)
+		assert.Equal(t, 3, cs.metadata.maxRetries)
+		assert.Equal(t, time.Second*4, cs.metadata.retryDelay)
+		assert.Equal(t, time.Second*120, cs.metadata.maxRetryDelay)
+	})
+}
+
+func Test_parseMetadata(t *testing.T) {
+	t.Run(fmt.Sprintf("parse metadatad with %s", host), func(t *testing.T) {
```
mukundansundar, line 97: nitpick:
```suggestion
	t.Run(fmt.Sprintf("parse metadata with %s", host), func(t *testing.T) {
```

configuration/azure/appconfig/appconfig_test.go
```diff
@@ -0,0 +1,178 @@
...
+		}
+
+		m, _ := parseMetadata(meta)
+		assert.NotNil(t, m)
+		if !reflect.DeepEqual(m, want) {
+			t.Errorf("parseMetadata() got = %v, want %v", m, want)
+		}
+	})
+
+	t.Run(fmt.Sprintf("parse metadatad with %s", connectionString), func(t *testing.T) {
```
mukundansundar, line 122: nitpick:
```suggestion
	t.Run(fmt.Sprintf("parse metadata with %s", connectionString), func(t *testing.T) {
```

configuration/azure/appconfig/appconfig_test.go
```diff
@@ -0,0 +1,178 @@
...
+		}
+
+		m, _ := parseMetadata(meta)
+		assert.NotNil(t, m)
+		if !reflect.DeepEqual(m, want) {
+			t.Errorf("parseMetadata() got = %v, want %v", m, want)
+		}
+	})
+
+	t.Run(fmt.Sprintf("can't set both %s and %s fields in metadata", host, connectionString), func(t *testing.T) {
```
mukundansundar, line 147:
```suggestion
	t.Run(fmt.Sprintf("both %s and %s fields set in metadata", host, connectionString), func(t *testing.T) {
```

configuration/azure/appconfig/appconfig_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+		meta := configuration.Metadata{
+			Properties: testProperties,
+		}
+
+		_, err := parseMetadata(meta)
+		assert.Error(t, err)
+	})
+
+	t.Run(fmt.Sprintf("specify %s or %s field in metadata", host, connectionString), func(t *testing.T) {
```
mukundansundar, line 163:
```suggestion
	t.Run(fmt.Sprintf("both %s and %s fields not set in metadata", host, connectionString), func(t *testing.T) {
```

internal/authentication/azure/auth.go
```diff
@@ -66,6 +66,9 @@ func NewEnvironmentSettings(resourceName string, values map[string]string) (Envi
 	case "signalr":
 		// Azure SignalR (data plane)
 		es.Resource = "https://signalr.azure.com"
+	case "appconfig":
+		// Azure App Configuration (data plane)
+		es.Resource = "https://azconfig.io"
```
mukundansundar, line 71: if possible please add relevant docs link which gives this value.

mukundansundar (commented):
please add unit tests for max retry settings parse errors.

mukundansundar (approved):
lgtm for the initial Get implementation.

mukundansundar (commented):
@berndverst can we not close #1541 till the Subscribe and Unsubscribe methods are implemented?

ItalyPaleAle:
@mukundansundar I changed the first post so it won't close the issue automatically.

berndverst (approved):

PR #501, opened 2020-10-15
Upgrade grpc version to v1.32.0

skyao:
# Description

update grpc version to v1.32.0

## Issue reference

https://github.com/dapr/dapr/issues/2101

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #1875, opened 2022-07-12
Modified envelope to accept Actors

DiegoAJR:
# Description
Modified the envelope structure of the Cloud Event to receive ActorType and ActorID. Added unit test cases to verify functionality and modified existing test cases. 

## Issue reference

Please reference the issue this PR will close: WIP #501 Pubsub for actors

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/pubsub/envelope_test.go
+++ b/pubsub/envelope_test.go
@@ -163,13 +163,13 @@ func TestCreateCloudEventsEnvelopeDefaults(t *testing.T) {
 
 	t.Run("actor type", func(t *testing.T) {
 		envelope := NewCloudEventsEnvelope("a", "source", "", "", "",
-			"mypubsub", "", []byte("data"), "1", "", "actortype", "")
-		assert.Equal(t, "actortype", envelope[ActorTypeField])
+			"mypubsub", "", []byte("data"), "1", "", "myactortype", "")
+		assert.Equal(t, "myactortype", envelope[ActorTypeField])
 	})
 
 	t.Run("actor id", func(t *testing.T) {
 		envelope := NewCloudEventsEnvelope("a", "source", "", "", "",
-			"mypubsub", "", []byte("data"), "1", "key=value", "", "1234")
+			"mypubsub", "", []byte("data"), "1", "", "", "1234")
 		assert.Equal(t, "1234", envelope[ActorIdField])
 	})
 }
@@ -261,7 +261,7 @@ func TestNewFromExisting(t *testing.T) {
 		}
 		b, _ := json.Marshal(&m)
 
-		n, err := FromCloudEvent(b, "b", "pubsub", "1", "key=value", "actortype", "1234")
+		n, err := FromCloudEvent(b, "b", "pubsub", "1", "key=value", "myactortype", "1234")
 		assert.NoError(t, err)
 		assert.Equal(t, "1.0", n[SpecVersionField])
 		assert.Equal(t, "a", n["customfield"])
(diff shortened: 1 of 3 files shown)


pubsub/envelope_test.go
```diff
@@ -103,63 +103,75 @@ func TestCreateFromJSON(t *testing.T) {
...
 	t.Run("trace state", func(t *testing.T) {
 		envelope := NewCloudEventsEnvelope("a", "source", "", "", "",
-			"mypubsub", "", []byte("data"), "1", "key=value")
+			"mypubsub", "", []byte("data"), "1", "key=value", "", "")
 		assert.Equal(t, "key=value", envelope[TraceStateField])
 	})
+
+	t.Run("actor type", func(t *testing.T) {
+		envelope := NewCloudEventsEnvelope("a", "source", "", "", "",
+			"mypubsub", "", []byte("data"), "1", "", "actortype", "")
```
berndverst, line 166: Can you make this something like "myactortype" and then check for that value below? Tiny change :)

pubsub/envelope_test.go
```diff
@@ -103,63 +103,75 @@ func TestCreateFromJSON(t *testing.T) {
...
+		assert.Equal(t, "actortype", envelope[ActorTypeField])
+	})
+
+	t.Run("actor id", func(t *testing.T) {
+		envelope := NewCloudEventsEnvelope("a", "source", "", "", "",
+			"mypubsub", "", []byte("data"), "1", "key=value", "", "1234")
```
berndverst, line 172: we don't need the strace state here.. so remove "key=value" and make it an empty string instead.

berndverst (changes requested):

DeepanshuA:
Hi @DiegoAJR, thanks for your contribution. 
I think it may be a good idea to update relevant doc alongside as well via a docs PR: https://docs.dapr.io/developing-applications/building-blocks/pubsub/pubsub-cloudevents/.

berndverst:
> Hi @DiegoAJR, thanks for your contribution. I think it may be a good idea to update relevant doc alongside as well via a docs PR: https://docs.dapr.io/developing-applications/building-blocks/pubsub/pubsub-cloudevents/.

These are optional fields and are only used for the internal implementation of Actor PubSub which isn't done yet. They don't change the behavior of cloud events themselves, but rather our internal representation. I don't think we need to update docs right now.

DeepanshuA:
> > Hi @DiegoAJR, thanks for your contribution. I think it may be a good idea to update relevant doc alongside as well via a docs PR: https://docs.dapr.io/developing-applications/building-blocks/pubsub/pubsub-cloudevents/.
> 
> These are optional fields and are only used for the internal implementation of Actor PubSub which isn't done yet. They don't change the behavior of cloud events themselves, but rather our internal representation. I don't think we need to update docs right now.

Sure, that is fine for optional fields then.

berndverst:
Please take a look at this test run: https://github.com/dapr/components-contrib/runs/7303808705?check_suite_focus=true

Essentially many of the certification tests in the `tests/certification` folder use the `NewCLoudEvent` function, so you need to update that to add the two additional parameters (empty strings is fine).

berndverst (changes requested):
Please take a look at this test run: https://github.com/dapr/components-contrib/runs/7303808705?check_suite_focus=true

Essentially many of the certification tests in the `tests/certification` folder use the NewCLoudEvent function, so you need to update that to add the two additional parameters (empty strings is fine).

berndverst (approved):

berndverst:
We ended up moving it to a feature branch since it also needs a runtime change which we cannot move into the actual runtime at this time.

PR #1329, opened 2021-11-24
Adds retry on CosmosDB Init in case of TooManyRequests error

berndverst:
# Description

Adds exponential backoff retry on Cosmos DB component initialization for up to 5 minutes total time.

Background: Cosmos DB makes a lot of Metadata requests upon client initialization. This limit is shared by all databases in a single Cosmos DB Account. It is easy to exceed this limits if multiple connections (for example via multiple sidecars) are made to the same Cosmos DB database account. To alleviate this problem we are introducing exponential backoff on init here.

As outlined in https://docs.microsoft.com/azure/cosmos-db/sql/troubleshoot-request-rate-too-large#recommended-solution-3

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [X] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #1876, opened 2022-07-12
Adding in track2 SDK support for CosmosDB Bindings

RyanLettieri:
Updating CosmosDB bindings to use track2 SDK


bindings/azure/cosmosdb/cosmosdb.go
```diff
@@ -62,56 +59,57 @@ func (c *CosmosDB) Init(metadata bindings.Metadata) error {
...
-	// disable the identification hydrator (which autogenerates IDs if missing from the request)
-	// so we aren't forced to use a struct by the upstream SDK
-	// this allows us to provide the most flexibility in the request document sent to this binding
-	config.IdentificationHydrator = nil
-	config.WithAppIdentifier("dapr-" + logger.DaprVersion)
-
-	c.client = documentdb.New(m.URL, config)
-
-	// Retries initializing the client if a TooManyRequests error is encountered
-	err = retryOperation(func() (err error) {
```
berndverst, line 91: We still need a backoff retry during the init (checking for 429 rate limit exceeded) because CosmosDB on client init downloads a ton of metadata under the hood which quickly result in API rate limits being exceeded when starting up multiple Dapr instances simultaneously. This was actually a customer issue we fixed in the past.

berndverst (changes requested):

bindings/azure/cosmosdb/cosmosdb.go
berndverst: @ItalyPaleAle can you please comment on this since you told @RyanLettieri that the new SDK handles this.

bindings/azure/cosmosdb/cosmosdb.go
```diff
@@ -62,56 +59,57 @@ func (c *CosmosDB) Init(metadata bindings.Metadata) error {
...
+		return err
+	}
+	// Create a container client
+	dbContainer, err := dbClient.NewContainer(m.Collection)
+	if err != nil {
+		return err
+	}
+	if dbContainer.ID() == "" {
+		return backoff.Permanent(
+			fmt.Errorf("container is non-existent"),
```
ItalyPaleAle, line 99: No need to use `backoff.Permanent` here because you're not inside a `retry` block

ItalyPaleAle: Just return an error.

bindings/azure/cosmosdb/cosmosdb.go
```diff
@@ -62,56 +59,57 @@ func (c *CosmosDB) Init(metadata bindings.Metadata) error {
...
+		)
+	}
+	c.client = dbContainer
+	containerRead, err := c.client.Read(context.Background(), nil)
 	if err != nil {
 		return err
 	}
 
+	if containerRead.ContainerProperties.ResourceID == "" {
+		return backoff.Permanent(
```
ItalyPaleAle, line 109: Same here: return an error not one wrapped in a backoff.

bindings/azure/cosmosdb/cosmosdb.go
```diff
@@ -62,56 +59,57 @@ func (c *CosmosDB) Init(metadata bindings.Metadata) error {
...
+	if err != nil {
+		return err
+	}
+	if dbContainer.ID() == "" {
+		return backoff.Permanent(
+			fmt.Errorf("container is non-existent"),
+		)
+	}
+	c.client = dbContainer
+	containerRead, err := c.client.Read(context.Background(), nil)
```
ItalyPaleAle, line 103: This needs a timeout because otherwise it could hang forever. For example:

```go
ctx, cancel := context.WithTimeout(context.Background(), 30 * time.Second)
containerRead, err := c.client.Read(ctx, nil)
cancel()
if err != nil {
    return err
}
```

ItalyPaleAle (changes requested):
Great work on this :) Small comments but otherwise it looks pretty solid

bindings/azure/cosmosdb/cosmosdb.go
ItalyPaleAle: Per @elsur (who maintains that SDK):

>  The Azure Go SDKs (all) have a common retry policy, which includes 429 handling: https://github.com/Azure/azure-sdk-for-go/blob/b386adc15c2e60e3812956f68d244b608330c479/sdk/azcore/runtime/policy_retry.go#L48

ItalyPaleAle: Per @ealsur (who maintains that SDK):

>  The Azure Go SDKs (all) have a common retry policy, which includes 429 handling: https://github.com/Azure/azure-sdk-for-go/blob/b386adc15c2e60e3812956f68d244b608330c479/sdk/azcore/runtime/policy_retry.go#L48

> Looks like it defaults to 3 retries but you can change it https://github.com/Azure/azure-sdk-for-go/blob/b386adc15c2e60e3812956f68d244b608330c479/sdk/azcore/runtime/policy_retry.go#L25

ItalyPaleAle: If needed, we can (should?) change the [`policy.RetryOptions` object](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime#NewRetryPolicy)

bindings/azure/cosmosdb/cosmosdb.go
```diff
@@ -62,56 +59,57 @@ func (c *CosmosDB) Init(metadata bindings.Metadata) error {
...
-		return nil
-	}, func(err error, d time.Duration) {
-		c.logger.Warnf("CosmosDB binding initialization failed: %v; retrying in %s", err, d)
-	}, 5*time.Minute)
+	dbClient, err := client.NewDatabase(m.Database)
+	if err != nil {
+		return err
+	}
+	// Create a container client
+	dbContainer, err := dbClient.NewContainer(m.Collection)
```
ealsur, line 93:
```suggestion
	dbContainer, err := client.NewContainer(m.Database, m.Collection)
```

ealsur: You don't need the previous `client.NewDatabase(m.Database)`

bindings/azure/cosmosdb/cosmosdb.go
```diff
@@ -62,56 +59,57 @@ func (c *CosmosDB) Init(metadata bindings.Metadata) error {
...
+			fmt.Errorf("container is non-existent"),
+		)
+	}
+	c.client = dbContainer
+	containerRead, err := c.client.Read(context.Background(), nil)
 	if err != nil {
 		return err
 	}
 
+	if containerRead.ContainerProperties.ResourceID == "" {
```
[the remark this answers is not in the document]
ealsur: Or the return in line 105 will simply bubble it up. This code here will never hit.

bindings/azure/cosmosdb/cosmosdb.go
```diff
@@ -85,30 +88,24 @@ func (c *CosmosDB) Init(metadata bindings.Metadata) error {
...
+	cancel()
 	if err != nil {
 		return err
 	}
 
-	if containerRead.ContainerProperties.ResourceID == "" {
-		return backoff.Permanent(
-			fmt.Errorf("attempted to read a non-existent container"),
-		)
+	if containerRead.RawResponse.StatusCode == http.StatusNotFound {
```
ealsur, line 107: On my previous comment what I meant was that if the container does not exist, you won't hit this code:

```
containerRead, err := c.client.Read(ctx, nil)
	cancel()
	if err != nil {
		return err
	}
```

A 404 is returned as `err`.

You can return the err to the caller like you currently.

bindings/azure/cosmosdb/cosmosdb.go
```diff
@@ -85,30 +88,24 @@ func (c *CosmosDB) Init(metadata bindings.Metadata) error {
...
-	if err != nil {
-		return err
-	}
 	// Create a container client
-	dbContainer, err := dbClient.NewContainer(m.Collection)
+	dbContainer, err := client.NewContainer(m.Database, m.Collection)
 	if err != nil {
 		return err
 	}
 	if dbContainer.ID() == "" {
```
ealsur, line 96: nit: You can know this by checking the `m.Collection` and/or `m.Database` parameters, the `dbContainer.ID()` is just returning the value of `m.Collection`.

bindings/azure/cosmosdb/cosmosdb.go
berndverst: We absolutely should. Let me link you to the PR where we fixed this in the Cosmos DB state store which absolutely could be an issue here. https://github.com/dapr/components-contrib/pull/1329

I don't remember whether it was Zeiss or Bosch bringing up this concern, but either way - it was an important issue which we were only able to address with exponential backoff for the client initialization.

bindings/azure/cosmosdb/cosmosdb.go
berndverst: Should such a timeout be configurable via a new metadata property?

bindings/azure/cosmosdb/cosmosdb.go
ItalyPaleAle: It does do exponential backoff already. It's configured to try 3 times with an exponential backoff of up to 2 mins. We need to tweak it only if we don't think that's good/enough.

bindings/azure/cosmosdb/cosmosdb.go
ealsur: NewContainer does not do any network calls (they don't have any `ctx` parameter), it just creates a client you can use to interact with a Container and in your case, call the `Read` method as it is done after this line.

My point is, before doing the Read call, this code is checking `dbContainer.ID() == ""`, basically validating if the `m.Collection` parameter is not `""`, it does not check that the container does not exist.

If the intent of this block is to check if the container does not exist, then it should be:

```
// Assuming m.Database and m.Collection will never be empty strings
// Otherwise, validate before calling NewContainer that m.Database or m.Collection
dbContainer, err := client.NewContainer(m.Database, m.Collection)
if err != nil {
	return err
}

c.client = dbContainer
ctx, cancel := context.WithTimeout(context.Background(), timeoutValue*time.Second)
containerRead, err := c.client.Read(ctx, nil)
cancel()
if err != nil {
    // If the container does not exist, it will return an err with Status 400
	return err
}

return nil
```

berndverst: In the past the error returned on `_, err := c.client.Read(ctx, nil)` was also the 429 error for too many metadata requests. Is this something that would now be handled internally in the SDK or could this also bubble up here?

If so, we do need add our backoff retry back to this and check for 429s specifically to reattempt the Read.

(FYI @ItalyPaleAle )

berndverst:
@ItalyPaleAle @RyanLettieri is this PR good to go? If not, maybe convert it to draft for now?

RyanLettieri:
> @ItalyPaleAle @RyanLettieri is this PR good to go? If not, maybe convert it to draft for now?

I just pushed the non-existent container check fix where I was incorrectly checking against a blank container name. I think it should be good now

ItalyPaleAle:
Let me review this on Monday one last time if you don't mind (I'll be OOF tomorrow)

@RyanLettieri  can you solve the merge conflict please?

berndverst:
@ItalyPaleAle will review Monday.

@RyanLettieri can you please fix the `go.sum` conflict?

bindings/azure/cosmosdb/cosmosdb.go
ItalyPaleAle: @RyanLettieri  can you please fix this? At the very least, sounds like the check on this line is unnecessary because it doesn't actually do anything

ItalyPaleAle (dismissed):

bindings/azure/cosmosdb/cosmosdb.go
berndverst: No need given we don't have any other such metadata. But let's define these kinds of constants at the top of the file, not inline please!

bindings/azure/cosmosdb/cosmosdb.go
berndverst: Lines 101-108 can go away and just become `return err`

So you'd have

```go
_, err := c.client.Read(ctx, nil)
return err
```

berndverst (dismissed):

berndverst (approved):

berndverst:
Make sure the Cosmos DB Binding tests also pass in this test run of conformance and certification tests:
https://github.com/dapr/components-contrib/actions/runs/2827757905
https://github.com/dapr/components-contrib/actions/runs/2827757908

PR #1886, opened 2022-07-18
pubsub.redis: better ack for messages

ItalyPaleAle:
Previously, ACKs could fail if the context was canceled.

Also, do not log errors for "context canceled" during shutdown


yaron2 (approved):
LGTM

PR #1882, opened 2022-07-15
pubsub.mqtt: support wildcard and shared subscriptions

ItalyPaleAle:
Fixes #1881

PR #1888, opened 2022-07-18
Backporting #1882 (MQTT fix) into release-1.8

### LINKED ISSUE #1881: MQTT wildcard and shared subscriptions are broken (again)
## Expected Behavior

When I subscribe to a mqtt topic with a wildcard subscription or a shared subscription, I expect to receive messages.

## Actual Behavior

No messages are received and the error `no handler defined for topic` is printed in the daprd log.

- Add a `pubsub.mqtt` component
- Subscribe with a shared subscription (`$share/group-id/topic`) or with a wildcard subscription (`topic/#`)

## Release Note

RELEASE NOTE:

ItalyPaleAle:
# Description

This PR backports #1882 into release-1.8 and fixes #1881 for the release branch.

Because of the very large number of changes in master, this is a separate PR and should not be cherry-picked back into master. For master, please merge #1882.

### DIFF
--- a/pubsub/mqtt/mqtt.go
+++ b/pubsub/mqtt/mqtt.go
@@ -18,9 +18,12 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/pem"
+	"errors"
 	"fmt"
 	"net/url"
+	"regexp"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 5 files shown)


berndverst:
MQTT Vernemq test run failed - which obviously is important for this PR. Rerunning.. hopefully it was just flaky.

ItalyPaleAle:
@berndverst  debugging now locally. This change works in #1882 against master. For some reason it doesn't work with vernemq (but works with others) against release-1.8. Looking into it

berndverst (changes requested):
Please investigate vermemq errors as discussed. (Leaving this review only so that nobody accidentally merges this)

ItalyPaleAle:
@berndverst should be fixed now. This was due to a deadlock that somehow only manifested itself with vernemq. I removed the locks, which are not necessary in 1.8 since we don't subscribe/unsubscribe dynamically, so the risk for consistency while iterating over the handler map is low.

In master, the code was reworked and there shouldn't be a risk of deadlocks even with locks enabled.

Issue #1892, opened 2022-07-19
Azure Service Bus Component: MessageID isn't guaranteed to be unique


clemensv:
The MessageID in Service Bus is a user-settable field that is not guaranteed to be unique and the service doesn't do anything to enforce uniqueness unless the user explicitly turns on duplicate detection. 

In the underlying AMQP protocol, message-id is optional and therefore cannot be relied on as a key as used here, especially when messages originate from outside of DAPR. See https://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-properties

The "SequenceNumber" system property (see https://docs.microsoft.com/en-us/azure/service-bus-messaging/message-sequencing and https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-amqp-protocol-guide#message-annotations) is a reliable identifier for uniqueness for a message within the scope of an entity and within the assumed lifetime scope of the lookup table. 

@ItalyPaleAle

artursouza:
Can you clarify the scenario you are targeting here?

clemensv:
I am reading the code and am seeing the use of a value that isn't guaranteed to be unique used as a hashtable index and that is a bug

ItalyPaleAle:
@clemensv thanks for pointing that out. Can you share a link to the line you're referring to?

jjcollinge:
Thanks @clemensv. @ItalyPaleAle the message ID is being used to track the active messages [here](https://github.com/dapr/components-contrib/blob/91ce3d3baf846a34ef6678964ee9f793c9172728/internal/component/azure/servicebus/subscription.go#L321) which is used to renew the locks on any messages we are currently handling. The use of a map here is so that we can do simpler random deletes when a message has been processed. We should move this key over to a guaranteed unique value e.g. sequence number to avoid messages overriding each other.

ItalyPaleAle:
Thanks for explaining @jjcollinge and for the PR :)

PR #1898, opened 2022-07-20
Use unique message key in Azure ServiceBus PubSub

jjcollinge:

# Description

Instead of using the message ID which is not guarenteed to be unique as a key in the map to track the active messages this PR changes to use the messages sequence 
number which is guarenteed to be unique per message.

## Issue reference

Please reference the issue this PR will close: #1892

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/internal/component/azure/servicebus/subscription.go
+++ b/internal/component/azure/servicebus/subscription.go
@@ -202,15 +202,6 @@ func (s *Subscription) handleAsync(ctx context.Context, msg *azservicebus.Receiv
 	go func() {
 		var consumeToken bool
 		var err error
-		var messageKey int64
-
-		// We check for the existence of the sequence number again just in case
-		// it has been unset by some other goroutine and we cannot process the message.
-		if msg.SequenceNumber == nil {
-			s.logger.Errorf("Message has no sequence number: %s", msg.MessageID)
-			s.AbandonMessage(ctx, msg)
-		}
-		messageKey = *msg.SequenceNumber
 
 		// If handleChan is non-nil, we have a limit on how many handler we can process
 		limitConcurrentHandlers := cap(s.handleChan) > 0
@@ -231,7 +222,7 @@ func (s *Subscription) handleAsync(ctx context.Context, msg *azservicebus.Receiv
 			}
 
 			// Remove the message from the map of active ones
-			s.removeActiveMessage(messageKey)
+			s.removeActiveMessage(*msg.SequenceNumber)
 
 			// Remove an entry from activeMessageChan to allow processing more messages
 			<-s.activeMessagesChan
@@ -330,10 +321,10 @@ func (s *Subscription) CompleteMessage(ctx context.Context, m *azservicebus.Rece
 }
 
 func (s *Subscription) addActiveMessage(m *azservicebus.ReceivedMessage) error {
-	s.logger.Debugf("Adding message %s to active messages on %s", m.MessageID, s.entity)
 	if m.SequenceNumber == nil {
 		return fmt.Errorf("message sequence number is nil")
 	}
+	s.logger.Debugf("Adding message with sequence number %d to active messages on %s", *m.SequenceNumber, s.entity)
 	s.mu.Lock()
 	s.activeMessages[*m.SequenceNumber] = m
 	s.mu.Unlock()
@@ -341,7 +332,7 @@ func (s *Subscription) addActiveMessage(m *azservicebus.ReceivedMessage) error {
 }
 
 func (s *Subscription) removeActiveMessage(messageKey int64) {
-	s.logger.Debugf("Removing message %s from active messages on %s", messageKey, s.entity)
+	s.logger.Debugf("Removing message with sequence number %d from active messages on %s", messageKey, s.entity)
 	s.mu.Lock()
 	delete(s.activeMessages, messageKey)
 	s.mu.Unlock()
--- a/internal/component/azure/servicebus/subscription.go
+++ b/internal/component/azure/servicebus/subscription.go
@@ -15,6 +15,7 @@ package servicebus
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


internal/component/azure/servicebus/subscription.go
```diff
@@ -197,6 +202,15 @@ func (s *Subscription) handleAsync(ctx context.Context, msg *azservicebus.Receiv
...
 		var err error
+		var messageKey int64
+
+		// We check for the existence of the sequence number again just in case
+		// it has been unset by some other goroutine and we cannot process the message.
+		if msg.SequenceNumber == nil {
+			s.logger.Errorf("Message has no sequence number: %s", msg.MessageID)
+			s.AbandonMessage(ctx, msg)
+		}
+		messageKey = *msg.SequenceNumber
```
ItalyPaleAle, line 213: If you're concerned about concurrency, do you think this may be an issue here? Say another goroutine changes `msg.SequenceNumber` to `nil` while this goroutine is going from line 209 to 213, causing a panic. Or is that not really a concern?

jjcollinge: This is a good point, this nil check is pretty useless without any other concurrency controls. I think in this case we are safe to not use mutexes and drop this check as long as `handleAsync` remains the only goroutine working with the message. Wdyt?

internal/component/azure/servicebus/subscription.go
```diff
@@ -175,7 +176,11 @@ func (s *Subscription) ReceiveAndBlock(handler HandlerFunc, lockRenewalInSec int
 		s.logger.Debugf("Received message: %s; current active message usage: %d/%d", msg.MessageID, len(s.activeMessagesChan), cap(s.activeMessagesChan))
```
ItalyPaleAle, line 176: Perhaps the log here could include msg.Sequencenumber too? So there's some correlation in the logs

internal/component/azure/servicebus/subscription.go
ItalyPaleAle: I think you're right. I don't think there's a risk of another goroutine changing msg, and I'd rather avoid adding another lock

jjcollinge: Thanks for the feedback. Will update later tonight 👍

berndverst:
Once the changes mentioned in the comments have been merged I will be happy to merge this. Please let me know when this is ready.

internal/component/azure/servicebus/subscription.go
jjcollinge: I've added the sequence number to the debug log when adding and removing the active message as we'll have checked for the existence of the sequence number by then and it's contextual. Happy with that?

internal/component/azure/servicebus/subscription.go
jjcollinge: Updated

ItalyPaleAle (dismissed):

ItalyPaleAle:
LGTM #ShipIt

internal/component/azure/servicebus/subscription.go
```diff
@@ -175,7 +176,11 @@ func (s *Subscription) ReceiveAndBlock(handler HandlerFunc, lockRenewalInSec int
...
 		// s.logger.Debugf("Message body: %s", string(msg.Body))
 
-		s.addActiveMessage(msg)
+		if err = s.addActiveMessage(msg); err != nil {
+			// If we cannot add the message then it is invalid and we cannot process it.
+			s.logger.Errorf("Error adding message: %s", err.Error())
```
mukundansundar, line 181:
```suggestion
			s.logger.Errorf("error adding message: %s", err.Error())
```

mukundansundar: Should msg id be part of the error message?

mukundansundar: Interesting, we might need to change all error returns and logs to not start with capital letters in this file...

internal/component/azure/servicebus/subscription.go
```diff
@@ -315,16 +320,20 @@ func (s *Subscription) CompleteMessage(ctx context.Context, m *azservicebus.Rece
 	}
 }
 
-func (s *Subscription) addActiveMessage(m *azservicebus.ReceivedMessage) {
-	s.logger.Debugf("Adding message %s to active messages on %s", m.MessageID, s.entity)
+func (s *Subscription) addActiveMessage(m *azservicebus.ReceivedMessage) error {
+	if m.SequenceNumber == nil {
```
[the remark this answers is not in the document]
mukundansundar: From https://docs.microsoft.com/en-us/azure/service-bus-messaging/message-sequencing it seems that the sequence number will always be present. Is that correct?
```

Sequencing and timestamping are two features that are always enabled on all Service Bus entities and surface through the SequenceNumber and EnqueuedTimeUtc properties of received or browsed messages.
```

internal/component/azure/servicebus/subscription.go
ItalyPaleAle: Conventions with Go is that error messages (everything that is inside `errors.New` or `fmt.Errorf` or similar) should start with lowercase letters. In Dapr, our log messages (such as `logger.Errorf`) normally start with an uppercase letter. Although, personally, I'm not picky about that :)

ItalyPaleAle: I am neutral either way, whether we have the convention or not, or whether the convention is for lowercased or uppercased.

jjcollinge: Happy to change all errors to lowercase to conform with the go convention 👌

internal/component/azure/servicebus/subscription.go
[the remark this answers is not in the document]
jjcollinge: Yes sequence number should always be present. The only reason I do a nil check is to avoid a panic if there is a bug in the underlying SDK when deserializing the message.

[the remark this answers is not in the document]
jjcollinge: Updated with the behaviour mentioned above

ItalyPaleAle (approved):

mukundansundar (approved):
Lgtm

berndverst (approved):

PR #1908, opened 2022-07-24
Add auto-completion flag for Zeebe

akkie:
# Description

The actual implementation of the Zeebe job worker binding auto-completes a job automatically if a success response will be returned or if an error occurs. In case of an error, the binding will create an incident. But there is also a case that a process can throw a business error. This can actually not be handled, because of this auto-completion functionality. 

This change introduces a new flag "autocomplete" which is enabled by default to keep backward compatibility. Auto-completion is also the most common way to handle worker results. If a worker should return a business error, then this flag can be set to false and the worker can complete the job itself with either an incident, business error or a success response.

The PR adds also new tests to test this behavior and fixes also the tests which were broken during recent Dapr changes. Maybe it makes sense to run the tests also in the pipeline.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#2673

### DIFF
--- a/bindings/zeebe/jobworker/jobworker.go
+++ b/bindings/zeebe/jobworker/jobworker.go
@@ -53,12 +53,14 @@ type jobWorkerMetadata struct {
 	PollInterval   metadata.Duration `json:"pollInterval"`
 	PollThreshold  float64           `json:"pollThreshold,string"`
 	FetchVariables string            `json:"fetchVariables"`
+	Autocomplete   *bool             `json:"autocomplete,omitempty"`
 }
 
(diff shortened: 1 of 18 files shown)


daixiang0:
@akkie please fix DCO issue :)

akkie:
@daixiang0 Fixed

akkie:
@daixiang0 Thanks. Docs are already updated and linked in the first comment

berndverst (approved):

PR #1911, opened 2022-07-25
Allow metadata to flow through service bus queue

halspang:
# Description

This commit lets metadata bind to a message and be read when
the message is received. This allows for the distributed tracing
of events.

This is required for https://github.com/dapr/dapr/issues/4884.

Other binding/pubsubs will need to be examined but this was chosen
as the first attempt. Once the dapr changes are committed, we can go
through the other input bindings.

Signed-off-by: Hal Spang <halspang@microsoft.com>

## Issue reference

This PR starts to address https://github.com/dapr/dapr/issues/4884 but requires changes in dapr/dapr after.

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/bindings/azure/servicebusqueues/servicebusqueues.go
+++ b/bindings/azure/servicebusqueues/servicebusqueues.go
@@ -154,8 +154,11 @@ func (a *AzureServiceBusQueues) Invoke(ctx context.Context, req *bindings.Invoke
 		a.senderLock.Unlock()
 	}
 
+	a.logger.Infof("Invoke Binding incoming metadata: %+v", req.Metadata)
+
 	msg := &servicebus.Message{
-		Body: req.Data,
+		Body:                  req.Data,
+		ApplicationProperties: make(map[string]interface{}),
 	}
 	if val, ok := req.Metadata[id]; ok && val != "" {
 		msg.MessageID = &val
@@ -163,6 +166,12 @@ func (a *AzureServiceBusQueues) Invoke(ctx context.Context, req *bindings.Invoke
 	if val, ok := req.Metadata[correlationID]; ok && val != "" {
 		msg.CorrelationID = &val
 	}
+
+	// Include incoming metadata in the message to be used when it is read.
+	for k, v := range req.Metadata {
+		msg.ApplicationProperties[k] = v
+	}
+
 	ttl, ok, err := contrib_metadata.TryGetTTL(req.Metadata)
 	if err != nil {
 		return nil, err
@@ -260,6 +269,13 @@ func (a *AzureServiceBusQueues) getHandlerFunc(handler bindings.Handler) impl.Ha
 		}
 		if msg.Subject != nil {
 			metadata[label] = *msg.Subject
+		}
+
+		// Passthrough any custom metadata to the handler.
+		for key, val := range msg.ApplicationProperties {
+			if stringVal, ok := val.(string); ok {
+				metadata[key] = stringVal
+			}
 		}
 
 		_, err := handler(a.ctx, &bindings.ReadResponse{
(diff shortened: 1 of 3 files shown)


bindings/azure/servicebusqueues/servicebusqueues.go
```diff
@@ -154,15 +154,24 @@ func (a *AzureServiceBusQueues) Invoke(ctx context.Context, req *bindings.Invoke
 		a.senderLock.Unlock()
 	}
 
+	a.logger.Infof("Invoke Binding incoming metadata: %+v", req.Metadata)
```
ItalyPaleAle, line 157: Can you make this a debug log please? Especially because the metadata may contain PII.

bindings/azure/servicebusqueues/servicebusqueues.go
```diff
@@ -154,15 +154,24 @@ func (a *AzureServiceBusQueues) Invoke(ctx context.Context, req *bindings.Invoke
...
 	if val, ok := req.Metadata[id]; ok && val != "" {
 		msg.MessageID = &val
 	}
 	if val, ok := req.Metadata[correlationID]; ok && val != "" {
 		msg.CorrelationID = &val
 	}
+
+	// Include incoming metadata in the message to be used when it is read.
+	for k, v := range req.Metadata {
+		msg.ApplicationProperties[k] = v
```
ItalyPaleAle, line 172: Shouldn't this contain a deny-list of keys we don't want included? For example `id` and `correlationID` (and more if needed) to avoid storing that data twice? Or perhaps it should be a sub-key, such as `req.Metadata["ApplicationProperties"]`?

ItalyPaleAle (changes requested):

bindings/azure/servicebusqueues/servicebusqueues.go
halspang: Yes, I actually meant to not include that but it missed my scan so I'm just going to remove it entirely.

bindings/azure/servicebusqueues/servicebusqueues.go
halspang: I think making it a subfield makes the incoming request too knowledgable of service bus queues. You're right we shouldn't store stuff twice though, so I'll add the deny-list there.

bindings/azure/servicebusqueues/servicebusqueues.go
```diff
@@ -155,14 +155,25 @@ func (a *AzureServiceBusQueues) Invoke(ctx context.Context, req *bindings.Invoke
...
 		msg.MessageID = &val
 	}
 	if val, ok := req.Metadata[correlationID]; ok && val != "" {
 		msg.CorrelationID = &val
 	}
+
+	// Include incoming metadata in the message to be used when it is read.
+	for k, v := range req.Metadata {
+		// Don't include the values that are saved in MessageID or CorrelationID.
+		if k == id || k == correlationID {
```
ItalyPaleAle, line 171: Should this be `k == "id" || k == "correlationID"`?

berndverst (changes requested):
Could you update the certification test to show this works with this PR?

ItalyPaleAle (approved):

ItalyPaleAle:
Approved this! But — do you think this could be implemented in the pubsub too, unless it’s not needed there?

(I’m hoping we can soon unify the two implementations too, and the closer they are already, the better)

halspang:
> Approved this! But — do you think this could be implemented in the pubsub too, unless it’s not needed there?
> 
> (I’m hoping we can soon unify the two implementations too, and the closer they are already, the better)

Yeah I can take a quick look at it and see if it's a small change. If it's bigger, I'll open it as a separate PR.

berndverst (approved):

berndverst:
@halspang (FYI @ItalyPaleAle ) please open a separate PR for service bus pubsub if this needs to be implemented there also.

PR #1920, opened 2022-07-30
Azure App Configuration: Implemented get all keys

cmendible:
Signed-off-by: cmendible <cmendible@gmail.com>

# Description

Implemented get all keys for Azure App Configuration configuration store.

## Issue reference

Please reference the issue this PR will close: #1541

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#2660


configuration/azure/appconfig/appconfig.go
```diff
@@ -160,32 +160,71 @@ func parseMetadata(meta configuration.Metadata) (metadata, error) {
...
+				})
+			if err != nil {
+				return nil, err
+			}
+
+			item := &configuration.Item{
+				Metadata: map[string]string{},
+			}
+			item.Key = key
+			item.Value = *resp.Value
```
ItalyPaleAle, line 185: Is there a chance that `resp.Value` and `resp.Label` could be nil? Should there be a nil-check before?

configuration/azure/appconfig/appconfig.go
```diff
@@ -160,32 +160,71 @@ func parseMetadata(meta configuration.Metadata) (metadata, error) {
 func (r *ConfigurationStore) Get(ctx context.Context, req *configuration.GetRequest) (*configuration.GetResponse, error) {
 	keys := req.Keys
 	items := make([]*configuration.Item, 0, len(keys))
```
ItalyPaleAle, line 162: This allocation of `items` seems to be only useful in the case where `len(keys) != 0`. Could you move it into to `else` block (to avoid allocating memory when not needed?)

configuration/azure/appconfig/appconfig.go
```diff
@@ -160,32 +160,71 @@ func parseMetadata(meta configuration.Metadata) (metadata, error) {
...
+			}
+
+			item := &configuration.Item{
+				Metadata: map[string]string{},
+			}
+			item.Key = key
+			item.Value = *resp.Value
+			item.Metadata["label"] = *resp.Label
+
+			items = append(items, item)
```
ItalyPaleAle, line 188: You have allocated `items` with a fixed size, so you shouldn't be using `append()`. A better approach would be:

```go
items := make([]*configuration.Item, 0, len(keys))
i := 0
for _, key := range keys {
	// do work here
	items[i] = item
	i++
}
// this is needed in case we are returning less items than keys
items = items[:i]
```

configuration/azure/appconfig/appconfig.go
```diff
@@ -160,32 +160,71 @@ func parseMetadata(meta configuration.Metadata) (metadata, error) {
...
-			Metadata: map[string]string{},
+	} else {
+		for _, key := range keys {
+			resp, err := r.client.GetSetting(
+				ctx,
+				key,
+				&azappconfig.GetSettingOptions{
+					Label: to.Ptr("label"),
+				})
+			if err != nil {
```
ItalyPaleAle, line 177: If a single key doesn't exist, is an error returned? In that case, what would be best:

- Returning an error (making the entire method fail)
- Or simply ignoring that key (return every other value but skipping that one)

?

ItalyPaleAle (changes requested):
Thanks for the PR :) Left some review comments!

configuration/azure/appconfig/appconfig.go
cmendible: Good one. If a value is not set API returns an empty string for Value but null for Label.

configuration/azure/appconfig/appconfig.go
cmendible: IMHO entire method should fail as **redis** implementation does. Also secret stores bulk method fail when an error is found fetching a key.

ItalyPaleAle: Ok consistency is good! So right now, if a value is not set, this returns an error?

cmendible: If a value is not set an empty string is returned. If the call fails for some reason then the method returns an error.

mukundansundar (approved):
lgtm

configuration/azure/appconfig/appconfig.go
```diff
@@ -159,33 +159,80 @@ func parseMetadata(meta configuration.Metadata) (metadata, error) {
...
-		items = append(items, item)
+	} else {
+		items = make([]*configuration.Item, 0, len(keys))
+		i := 0
+		for _, key := range keys {
+			resp, err := r.client.GetSetting(
+				ctx,
+				key,
+				&azappconfig.GetSettingOptions{
+					Label: to.Ptr("label"),
```
mukundansundar, line 177: By setting this value, we are only querying keys that have the label as `label`. For all the others only an error is returned. This either should be a filed read from the request metadata or not used at all.

mukundansundar (changes requested):

configuration/azure/appconfig/appconfig.go
mukundansundar: got the response like following when the Label `label` is not present for the key
````fail to get [test-dapr] from Configuration store test-azappconfig: GET https://{instance}.azconfig.io/kv/test-key\n--------------------------------------------------------------------------------\nRESPONSE 404: 404 Not Found\nERROR CODE UNAVAILABLE\n--------------------------------------------------------------------------------\nResponse contained no body\n--------------------------------------------------------------------------------\n"```

Comment (mukundansundar) [reply configuration/azure/appconfig/appconfig.go:177]: At the same time when a request is made for a key `test-lab` which has the Label `label` the output is the following:
```
[
    {
        "key": "test-lab",
        "value": "test-lab",
        "metadata": {
            "label": "label"
        }
    }
]
```

Comment (cmendible) [reply configuration/azure/appconfig/appconfig.go:177]: Should be `Label: to.Ptr("*")`

**Review by mukundansundar — DISMISSED** [review]
lgtm

File: configuration/azure/appconfig/appconfig.go [hunk 6]
```diff
@@ -159,33 +159,80 @@ func parseMetadata(meta configuration.Metadata) (metadata, error) {
...
+			if err != nil {
+				return &configuration.GetResponse{}, err
+			}
+
+			item := &configuration.Item{
+				Metadata: map[string]string{},
+			}
+			item.Key = key
+			item.Value = *resp.Value
+			if resp.Value != nil {
```
ItalyPaleAle, line 188: This should probably be `if resp.Label != nil` ?

configuration/azure/appconfig/appconfig.go
```diff
@@ -159,33 +159,80 @@ func parseMetadata(meta configuration.Metadata) (metadata, error) {
...
+
+	for revPgr.More() {
+		if revResp, err := revPgr.NextPage(ctx); err == nil {
+			for _, setting := range revResp.Settings {
+				item := &configuration.Item{
+					Metadata: map[string]string{},
+				}
+				item.Key = *setting.Key
+				item.Value = *setting.Value
+				if setting.Value != nil {
```
ItalyPaleAle, line 222: Same here. Should probably be `if setting.Label != nil` ?

ItalyPaleAle (changes requested):

ItalyPaleAle (approved):
LGTM

berndverst (approved):

PR #1924, opened 2022-08-02
[pubsub/jetstream]: allow tls client authentication

### LINKED ISSUE #1923: [Jetstream] Allow TLS Client Authentication
Looking at the docs, and the configuration struct for Jetstream, it doesn't seem like TLS Client Authentication can be used.

https://github.com/dapr/components-contrib/blob/e639685e83b3a2ad1a73fc89a4370b6d11ed9d17/pubsub/jetstream/metadata.go#L24-L36

I wonder if it is a trivial addition?

NickLarsenNZ:
# Description

- Configure the NATS client for TLS Client Authentication if a key pair is supplied.

## Issue reference

- Closes #1923

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#2697

### DIFF
--- a/pubsub/jetstream/jetstream.go
+++ b/pubsub/jetstream/jetstream.go
@@ -54,6 +54,9 @@ func (js *jetstreamPubSub) Init(metadata pubsub.Metadata) error {
 		}, func(nonce []byte) ([]byte, error) {
 			return sigHandler(js.meta.seedKey, nonce)
 		}))
+	} else if js.meta.tls_client_cert != "" && js.meta.tls_client_key != "" {
+		js.l.Debug("Configure nats for tls client authentication")
+		opts = append(opts, nats.ClientCert(js.meta.tls_client_cert, js.meta.tls_client_key))
 	}
 
 	js.nc, err = nats.Connect(js.meta.natsURL, opts...)
--- a/pubsub/jetstream/metadata.go
+++ b/pubsub/jetstream/metadata.go
@@ -23,8 +23,12 @@ import (
 
 type metadata struct {
 	natsURL string
+
 	jwt     string
 	seedKey string
+
+	tls_client_cert string
+	tls_client_key  string
 
 	name           string
 	durableName    string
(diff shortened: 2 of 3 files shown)


ItalyPaleAle:
Thanks @NickLarsenNZ for this contribution!

For the CA cert, do you think it would be possible to use the new feature in 1.8 to add the cert to the Dapr container rather than adding something specific to this component? https://docs.dapr.io/operations/configuration/install-certificates/

NickLarsenNZ:
@ItalyPaleAle, regarding the CA cert, yeah I guess most people will run it from a container so probably good enough to just do it as per your suggestion. I'll come back to remove this from here and the docs tomorrow.

As for v1.8, do you mean this feature doesn't have to wait until 1.9? I just went with what the docs hinted towards (see dapr/docs#2697).

yaron2:
> @ItalyPaleAle, regarding the CA cert, yeah I guess most people will run it from a container so probably good enough to just do it as per your suggestion. I'll come back to remove this from here and the docs tomorrow.
> 
> As for v1.8, do you mean this feature doesn't have to wait until 1.9? I just went with what the docs hinted towards (see [dapr/docs#2697](https://github.com/dapr/docs/issues/2697)).

To be clear, the PR will go into 1.9.

ItalyPaleAle:
@NickLarsenNZ 

> regarding the CA cert, yeah I guess most people will run it from a container so probably good enough to just do it as per your suggestion. I'll come back to remove this from here and the docs tomorrow.

Well, the solution works in non-containerized methods too. In that case, the sidecar (daprd) is a process on the operating system, and it should be able to pick up certs from the OS' trust store. So to add a new root CA, you'd add it to your OS' trust store (for example on Mac you'd add it to the Keychain, and on Windows you use mmc.exe to add it to the trust store).

Adding a cert to the root trust store in the OS has always been supported by Dapr. What we added in 1.8 was a way to make it possible to do that with a Docker container (in Kubernetes or as a standalone container or using Docker Compose) that doesn't require building custom containers.

The only reason why I brought this up is that given how we've been trying to standardize adding root CA certs through the "standard" way (of adding them to the OS trust store), I would prefer not to create a duplicate way to do that which is component-specific. Because next thing you know, every component would have to be updated to add a component-specific way to add a custom CA.... and that's something else we'll have to maintain 🙃

> As for v1.8, do you mean this feature doesn't have to wait until 1.9

No, as @yaron2  confirmed this feature will definitely have to wait for 1.9.

I mentioned 1.8 because that's when we shipped support for adding root CA certs in Docker too.

mukundansundar (dismissed):
LGTM

NickLarsenNZ:
@yaron2 @ItalyPaleAle @mukundansundar I've removed the CA section, and updated the docs. Should be good to go.

ItalyPaleAle (approved):
:shipit: thanks for the PR :)

mukundansundar (approved):
lgtm. Thanks for the contribution !!

berndverst (approved):

mukundansundar:
@NickLarsenNZ Can you fix the linter error that we are seeing?

ItalyPaleAle:
@mukundansundar the linter error seems to be due to timeouts and not @NickLarsenNZ 's fault. Hopefully #1931 can be merged to fix this

gehhilfe:
Thank you, @NickLarsenNZ wanted to add the same, but you were faster!

Issue #1752, opened 2022-05-25
Standardize behavior on component initialization failure


ItalyPaleAle:
Currently, when components fail to initialize (for example because of a misconfiguration or connection error), the behavior is undefined and depends on the component.

Some components cause the sidecar to crash; others print a warning or error message; others do nothing and try reconnecting again lazily.

To improve the user experience, we should look into standardizing this behavior.

In all cases we should print out an error message that describes the problem, at least saying that the component failed to start. If possible, then, provide more details, and it would be great to add a link to the Dapr docs to point people to the component's specs.

What happens next (crash or continue working without the component) is up for discussion.

Note that some components maintain a persistent connection, so connection failures can be determined in the "init" methods directly (e.g. databases). Other components are stateless (e.g. Azure Storage, S3), so we need to decide what to do with those: do we attempt to make a GET request in the "init" method to validate the connection parameters, or do we lazily wait for first use to report a connection issue? (Or that could be configurable?)

artursouza:
I think we should make this behavior be verified by conformance tests - that is where standard behavior is validated.

Now, when it comes to what should be the standard behavior, I think it should be a combination of all above:
1. Component should try to connect on init, optionally with built-in retries. If cannot connect, return error.
2. Component can optionally have reconnection logic in the background, in case connection is lost post init (this is validated in certification tests)
3. Dapr runtime will log error when a component cannot init.
4. Dapr runtime should emit a metric for each component that failed/succeeded to initialize, but metric should include the component type and not the component name (since name will vary per customer setup) - this should be validated in conformance tests.

ItalyPaleAle:
That makes sense to me.

So, to confirm, we should not crash when a component fails to init. Instead, we'll let the app make API calls that will fail.

yaron2:
> I think we should make this behavior be verified by conformance tests - that is where standard behavior is validated.
> 
> Now, when it comes to what should be the standard behavior, I think it should be a combination of all above:
> 
> 1. Component should try to connect on init, optionally with built-in retries. If cannot connect, return error.
> 2. Component can optionally have reconnection logic in the background, in case connection is lost post init (this is validated in certification tests)
> 3. Dapr runtime will log error when a component cannot init.
> 4. Dapr runtime should emit a metric for each component that failed/succeeded to initialize, but metric should include the component type and not the component name (since name will vary per customer setup) - this should be validated in conformance tests.

This makes sense to me.

artursouza:
[...]

PR #1935, opened 2022-08-04
Attempt to validate azure/gcp credentials for pubsub and firestore components

addjuarez:
# Description

Attempted to add credential checks in the init() methods of azure service bus and gcp pubsub/firestore components. 

For GCP Pubsub and Azure Service bus it currently involves adding an optional metadata field "initialTopic". The user would add a topic or possibly topics that we could make get/create calls on during init and could validate if the credentials were correctly configured and have sufficient permissions. We could also simplify the approach and do only a getTopic() call on a random string to see if we have permissions to read from the service.

For firestore we do a query on a dummy string. The entry should not exist and we should get a ErrNoSuchEntity if configured correctly. Any other errors would result in failure.

## Issue reference

#1752 One of multiple PRs for this issue

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/pubsub/azure/servicebus/metadata.go
+++ b/pubsub/azure/servicebus/metadata.go
@@ -34,6 +34,7 @@ type metadata struct {
 	PublishMaxRetries               int    `json:"publishMaxRetries"`
 	PublishInitialRetryIntervalInMs int    `json:"publishInitialRetryInternalInMs"`
 	NamespaceName                   string `json:"namespaceName,omitempty"`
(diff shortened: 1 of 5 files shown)


yaron2:
I support credentials validation but not like this. Making dummy calls carry hidden costs for managed services and increased load. Imagine scaling pods from 1 to 200 that cause a big charge and worse, throttling on the service level.

Issue #1922, opened 2022-08-02
Add certification tests for Azure Service Bus Pubsub


amulyavarote:
The Azure Service Bus Pubsub requires certification tests to be created.
Add conformance tests if not present.

Release Note

RELEASE NOTE: ADD Certification tests for Azure Service Bus Pubsub.

tanvigour:
Thanks for the plan Amulya, overall this looks good. Here are few suggested changes from my side
 
1. Verifying for optional parameters may not be needed as a part of certification test if already covered somewhere else, please check on that.
2. We should add an AAD testing scenario with string based authentication and service principal based authentication. please refer to azure eventhubs pubsub for example scenario for that.

PR #1937, opened 2022-08-05
Added certification tests for azure service bus pubsub

amulyavarote:
Signed-off-by: Amulya Varote <amulyavarote@microsoft.com>

# Description

Added certification tests for azure service bus for pubsub

## Issue reference

Please reference the issue this PR will close: #1922 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


tests/certification/pubsub/azure/servicebus/README.md
```diff
@@ -0,0 +1,41 @@
+# Azure Service Bus Pubsub Certification
+The purpose of this module is to provide tests that certify the Azure Service Bus Queue Input/Output Binding as a stable component.
```
ItalyPaleAle, line 2: Please change this to refer to the pubsub component and not binding

tests/certification/pubsub/azure/servicebus/servicebus_test.go
```diff
@@ -0,0 +1,266 @@
...
+
+	numMessages      = 10
+	appPort          = 8000
+	portOffset       = 2
+	messageKey       = "partitionKey"
+	pubsubName       = "messagebus"
+	topicActiveName  = "certification-pubsub-topic-active"
+	topicPassiveName = "certification-pubsub-topic-passive"
+	topicToBeCreated = "certification-topic-per-test-run"
+	iotHubNameEnvKey = "AzureIotHubName"
```
ItalyPaleAle, line 70: Extra?

tests/certification/pubsub/azure/servicebus/README.md
```diff
@@ -0,0 +1,41 @@
...
+   - Publisher publishes to 1 topic
+   - Subscriber is subscribed to 1 topic
+   - Simulate periodic errors and verify that the component retires on error
+   - Verify that all expected messages were received
+- Verify data with a topic that does not exist
+   - Run dapr application with 1 publisher and 1 subscriber
+   - Verify the creation of service bus
+   - Send messages to the service created
+   - Verify that subscriber received all the messages
+- Verify reconnection after the network interruptions
```
ItalyPaleAle, line 36: I don't see this part in the test, has it been completed?

tests/certification/pubsub/azure/servicebus/README.md
```diff
@@ -0,0 +1,41 @@
...
+   - Run dapr application with 1 publisher and 1 subscriber
+   - Publisher publishes to 1 topic
+   - Subscriber is subscribed to 1 topic
+   - Simulate network interruptions and verify that the component retires on error
```
ItalyPaleAle, line 40: One interesting thing here would be to simulate a failure for both the publisher and subscriber, separately:

1. Start both.
2. Publisher starts publishing and subscriber starts reading.
3. Simulate failure in publisher, then resume. Verify that all messages are still sent.
4. Simulate a failure in the subscriber, and verify that it resumes correctly. It should also receive the messages that were sent while the subscriber was down.

ItalyPaleAle (dismissed):

tests/certification/pubsub/azure/servicebus/README.md
amulyavarote: Added now!

berndverst (changes requested):
There should be more tests verifying the metadata options.

I want to see something that verifies the behavior of `disableEntityManagement`. When this is set to true topics which don't already exist should not be created - we should verify this behavior here.

Also I'd like for us to verify that `defaultMessageTimeToLiveInSec` actually works.

In general we should verify the behavior of all metadata options where possible.

berndverst:
@amulyavarote I marked this PR as a draft until the remaining tests have been added

berndverst:
Because this workflow was not previously enabled I need to merge this PR to make sure tests are passing. If they do not pass after merging, there will be a follow up item for you @amulyavarote. FYI

berndverst (dismissed):

tests/certification/pubsub/azure/servicebus/servicebus_test.go
```diff
@@ -0,0 +1,346 @@
...
+		// Run subscriberApplication app5
+		Step("verify if app5 has recevied messages published to newly created topic", testDefault).
+		Step(fmt.Sprintf("publish messages to topicDefault: %s", topicDefaultName), publishMessages(metadata, sidecarName5, topicDefaultName, !defaultTopicCreation, consumerGroup4)).
+
+		// Test : defaultMessageTimeToLiveInSec optional attribute, in order processing with single publisher/subscriber
+		// Run subscriberApplication app6
+		Step(app.Run(appID6, fmt.Sprintf(":%d", appPort+portOffset*5),
+			subscriberApplication(appID6, topicActiveName, consumerGroup5))).
+
+		// Run the Dapr sidecar with the component entitymanagement
```
berndverst, line 334: With entity management disabled you mean?

Also, this means that topics and subscriptions do not get automatically created. I want to see this behavior actually verified. There should be an error publishing a message to a topic which doesn't already exist if I'm understanding this correctly.

tests/certification/pubsub/azure/servicebus/servicebus_test.go
```diff
@@ -0,0 +1,346 @@
...
+	flow.New(t, "servicebus certification").
+
+		// Test : single publisher, multiple subscriber with their own consumerID
+		// Run subscriberApplication app1
+		Step(app.Run(appID1, fmt.Sprintf(":%d", appPort),
+			subscriberApplication(appID1, topicActiveName, consumerGroup1))).
+
+		// Run the Dapr sidecar with the eventhubs component 1, with permission at namespace level
+		Step(sidecar.Run(sidecarName1,
+			embedded.WithComponentsPath("./components/consumer1"),
```
berndverst, line 259: Please rename your component folders to give them more meaningful names. It should be more obvious which folder contains what.

tests/certification/pubsub/azure/servicebus/servicebus_test.go
```diff
@@ -0,0 +1,346 @@
...
+			embedded.WithDaprHTTPPort(runtime.DefaultDaprHTTPPort+portOffset*3),
+			embedded.WithProfilePort(runtime.DefaultProfilePort+portOffset*3),
+			runtime.WithSecretStores(secretStoreComponent),
+			runtime.WithPubSubs(component))).
+		Step(fmt.Sprintf("publish messages to topicToBeCreated: %s", topicToBeCreated), publishMessages(metadata, sidecarName4, topicToBeCreated, defaultTopicCreation, consumerGroup3)).
+		Step("interrupt network", network.InterruptNetwork(time.Minute, []string{}, []string{}, "5671", "5672")).
+		Step("wait", flow.Sleep(30*time.Second)).
+		Step("verify if app4 has recevied messages published to newly created topic", assertMessages(10*time.Second, consumerGroup3)).
+
+		// Test : disableEntityManagement optional attribute , Test processing with all optional attributes and single publisher/subscriber
```
berndverst, line 324: So this is only testing that by default entity management is enabled. That should be made more clear.

tests/certification/pubsub/azure/servicebus/servicebus_test.go
```diff
@@ -0,0 +1,346 @@
...
+		// Run subscriberApplication app5
+		Step("verify if app5 has recevied messages published to newly created topic", testDefault).
+		Step(fmt.Sprintf("publish messages to topicDefault: %s", topicDefaultName), publishMessages(metadata, sidecarName5, topicDefaultName, !defaultTopicCreation, consumerGroup4)).
+
+		// Test : defaultMessageTimeToLiveInSec optional attribute, in order processing with single publisher/subscriber
+		// Run subscriberApplication app6
+		Step(app.Run(appID6, fmt.Sprintf(":%d", appPort+portOffset*5),
```
berndverst, line 331: How are you verifying anything related to TTL? Somehow you need to verify that the default TTL was actually applied if this is what you intend to test here.

berndverst (changes requested):
Please restructure your tests.

For each sidecar / component I would like to see a separate flow!

```golang
// Test entity managedement disabled
flow.New(t, "entity management disabled").
.Step()....
.Run()

// test entity management enabled
flow.New(t, "entity management enabled").
.Step() ...
.Run()
```

berndverst:
@amulyavarote after discussing with you it seems you followed a lot of the structured created by the Kafka certification test. I find that test not readable and unmaintainable as is. My feedback here should make it easier to understand at a quick glance what is being tested and why / how.

tests/certification/pubsub/azure/servicebus/components/authentication/service_bus.yaml
```diff
@@ -0,0 +1,29 @@
...
+metadata:
+  name: messagebus
+  namespace: default
+spec:
+  type: pubsub.azure.servicebus
+  version: v1
+  metadata:
+    - name: consumerID
+      value: optionscertification
+    - name: connectionString
```
berndverst, line 12: If you are testing service principal authentication you must not also specify the connection string, or else it will not actually use the service principal approach. You likely need to specify other metadata options.

tests/certification/pubsub/azure/servicebus/servicebus_test.go
```diff
@@ -0,0 +1,1136 @@
...
+					err = client.PublishEvent(ctx, pubsubName, topicName, message)
+				}
+				require.Error(ctx, err, "error publishing message")
+			}
+			return nil
+		}
+	}
+
+	testDefault := func(ctx flow.Context) error {
+		ports, _ := dapr_testing.GetFreePorts(6)
```
berndverst, line 877: Why 6 ports here?

tests/certification/pubsub/azure/servicebus/servicebus_test.go
```diff
@@ -0,0 +1,1136 @@
...
+			})))
+		sidecar.Start(ctx)
+		return nil
+	}
+
+	flow.New(t, "servicebus certification - entity management enabled").
+
+		// Test : disableEntityManagement optional attribute , Test processing with all optional attributes and single publisher/subscriber
+		// Run subscriberApplication app1
+		Step("verify if app1 has recevied messages published to newly created topic", testDefault).
```
berndverst, line 902: Why are you not starting the sidecar directly in the flow like everywhere else? The sidecar should be initiated normally before you try to send an event

tests/certification/pubsub/azure/servicebus/servicebus_test.go
```diff
@@ -0,0 +1,1136 @@
...
+			// Wait for double the TTL after sending the last message.
+			time.Sleep(time.Second * 20)
+			return nil
+		}
+	}
+
+	assertMessages := func(timeout time.Duration, messageWatchers ...*watcher.Watcher) flow.Runnable {
+		return func(ctx flow.Context) error {
+			// assert for messages
+			for _, m := range messageWatchers {
```
berndverst, line 994: The goal here is to assert that there were no messages received at all because you waited twice the time of the default TTL.

I don't think you are sufficiently showing here that no message was received at all. I believe one of the other pubsub tests may have something you can use for this specific case.

tests/certification/pubsub/azure/servicebus/servicebus_test.go
```diff
@@ -0,0 +1,1136 @@
...
+			for _, message := range messages {
+				ctx.Logf("Publishing: %q", message)
+				var err error
+
+				if publishOptions != nil {
+					err = client.PublishEvent(ctx, pubsubName, topicName, message, publishOptions)
+				} else {
+					err = client.PublishEvent(ctx, pubsubName, topicName, message)
+				}
+				require.Error(ctx, err, "error publishing message")
```
berndverst, line 870: Can you add a comment here saying this error is expected because the topic does not exist?

berndverst (changes requested):

tests/certification/pubsub/azure/servicebus/servicebus_test.go
berndverst: As you explained this was done in a separate function because the sidecar cannot start up. If the sidecar cannot start (as expected) we should try to get the sidecar init failure and assert that instead. No point in attempting to publish messages which of course will fail and therefore that assert is somewhat meaningless. Any number of things could lead a sidecar to fail to startup and this test wouldn't be that meaningful.

tests/certification/pubsub/azure/servicebus/servicebus_test.go
```diff
@@ -0,0 +1,1136 @@
...
+		return func(ctx flow.Context) error {
+			// prepare the messages
+			messages := make([]string, numMessages)
+			for i := range messages {
+				messages[i] = fmt.Sprintf("partitionKey: %s, message for topic: %s, index: %03d, uniqueId: %s", metadata[messageKey], topicName, i, uuid.New().String())
+			}
+
+			// add the messages as expectations to the watchers
+			for _, messageWatcher := range messageWatchers {
+				messageWatcher.ExpectStrings(messages...)
```
berndverst, line 959: You are adding all messages to the message watcher.. even though you want to check that all (or some) of the messages are not sent because they expire.

So maybe do *not* add the message here.

tests/certification/pubsub/azure/servicebus/servicebus_test.go
berndverst: see if you can use `m.FailIfNoExpected` somehow. This should throw an error when there is an unexpected message.

Please note that your test currently seems to expect all messsages with TTL and successfully verified that they are received. What you want to test however is that they are not received due to the expiring TTL.

tests/certification/pubsub/azure/servicebus/servicebus_test.go
```diff
@@ -0,0 +1,1136 @@
...
+		Step(sidecar.Run(sidecarName1,
+			embedded.WithComponentsPath("./components/default_ttl"),
+			embedded.WithAppProtocol(runtime.HTTPProtocol, appPort+portOffset*5),
+			embedded.WithDaprGRPCPort(runtime.DefaultDaprAPIGRPCPort+portOffset*5),
+			embedded.WithDaprHTTPPort(runtime.DefaultDaprHTTPPort+portOffset*5),
+			embedded.WithProfilePort(runtime.DefaultProfilePort+portOffset*5),
+			runtime.WithSecretStores(secretStoreComponent),
+			runtime.WithPubSubs(component))).
+		Step(fmt.Sprintf("publish messages to topicToBeCreated: %s", topicActiveName), testTtlPublishMessages(metadata, sidecarName1, topicActiveName, consumerGroup1)).
+		Step("verify if app6 has recevied messages published to newly created topic", assertMessages(10*time.Second, consumerGroup1)).
```
berndverst, line 1019: Somehow you should publish the messages -- without subscribing to the messages - make note of the list of expected messages.
Then wait at least 2 * the default TTL window, then subscribe to the topic to receive the messages and use `consumerGroup1.FailIfNotExpected` or something like that.

It's important you don't subscribe to the message until the TTL has expired. If  you subscribe too early messages are delivered to your app / to the watcher.. and that defeats the purpose.

berndverst (commented):
Added a couple more thoughts on the TTL case.

tests/certification/pubsub/azure/servicebus/servicebus_test.go
```diff
@@ -0,0 +1,1127 @@
...
+					err = client.PublishEvent(ctx, pubsubName, topicName, message)
+				}
+				//Error is expected as the topic does not exist
+				require.Error(ctx, err, "error publishing message")
+			}
+			return nil
+		}
+	}
+
+	flow.New(t, "servicebus certification - entity management enabled").
```
berndverst, line 894: You mean disabled.

tests/certification/pubsub/azure/servicebus/servicebus_test.go
```diff
@@ -0,0 +1,1127 @@
...
+			subscriberApplication(appID1, topicActiveName, consumerGroup1))).
+		Step(sidecar.Run(sidecarName1,
+			embedded.WithComponentsPath("./components/entity_mgnt"),
+			embedded.WithAppProtocol(runtime.HTTPProtocol, appPort+portOffset),
+			embedded.WithDaprGRPCPort(runtime.DefaultDaprAPIGRPCPort+portOffset),
+			embedded.WithDaprHTTPPort(runtime.DefaultDaprHTTPPort+portOffset),
+			embedded.WithProfilePort(runtime.DefaultProfilePort+portOffset),
+			runtime.WithSecretStores(secretStoreComponent),
+			runtime.WithPubSubs(component))).
+		Step(fmt.Sprintf("publish messages to topicDefault: %s", topicDefaultName), publishMessages(metadata, sidecarName1, topicDefaultName, consumerGroup1)).
```
berndverst, line 907: As discussed, literally any operation on the sidecar would fail at this point because the sidecar didn't initialize correctly.

I don't particularly like this test as is given that it doesn't test the exact error in isolation. That being said in combination with your other component yaml variants and test cases we can keep this for now.

Can you add a TODO comment to instead look for the component init error in the sidecar itself? Thanks!

berndverst (changes requested):

.github/infrastructure/conformance/azure/setup-azure-conf-test.sh
```diff
@@ -613,6 +614,8 @@ echo "Configuring Service Bus test settings ..."
 SERVICE_BUS_CONNECTION_STRING="$(az servicebus namespace authorization-rule keys list --name RootManageSharedAccessKey --namespace-name "${SERVICE_BUS_NAME}" --resource-group "${RESOURCE_GROUP_NAME}" --query "primaryConnectionString" --output tsv)"
 echo export ${SERVICE_BUS_CONNECTION_STRING_VAR_NAME}=\"${SERVICE_BUS_CONNECTION_STRING}\" >> "${ENV_CONFIG_FILENAME}"
 az keyvault secret set --name "${SERVICE_BUS_CONNECTION_STRING_VAR_NAME}" --vault-name "${KEYVAULT_NAME}" --value "${SERVICE_BUS_CONNECTION_STRING}"
+SERVICE_BUS_NAMESPACE="${SERVICE_BUS_NAME}.servicebus.windows.net"
+echo export ${SERVICE_BUS_NAMESPACE_VAR_NAME}=\"${SERVICE_BUS_NAMESPACE}\" >> "${ENV_CONFIG_FILENAME}"
```
berndverst, line 618: Almost, but you forgot to export this to KeyVault, which is how you can actually reference this value in the GitHub workflow!

```suggestion
echo export ${SERVICE_BUS_NAMESPACE_VAR_NAME}=\"${SERVICE_BUS_NAMESPACE}\" >> "${ENV_CONFIG_FILENAME}"
az keyvault secret set --name "${SERVICE_BUS_NAMESPACE_VAR_NAME}" --vault-name "${KEYVAULT_NAME}" --value "${SERVICE_BUS_NAMESPACE}"
```

tests/certification/pubsub/azure/servicebus/components/authentication/service_bus.yaml
```diff
@@ -0,0 +1,29 @@
...
+  namespace: default
+spec:
+  type: pubsub.azure.servicebus
+  version: v1
+  metadata:
+    - name: consumerID
+      value: optionscertification
+    - name: namespaceName
+      secretKeyRef:
+        name: AzureServiceBusNamespace
```
berndverst, line 14: Good, but where is this coming from if you didn't actually add it in the workflow :)

.github/workflows/certification.yml
```diff
@@ -90,6 +90,8 @@ jobs:
...
         - component: pubsub.azure.eventhubs
           required-secrets: AzureEventHubsPubsubTopicActiveConnectionString,AzureEventHubsPubsubNamespace,AzureEventHubsPubsubNamespaceConnectionString,AzureBlobStorageAccount,AzureBlobStorageAccessKey,AzureEventHubsPubsubContainer,AzureIotHubName,AzureIotHubEventHubConnectionString,AzureCertificationTenantId,AzureCertificationServicePrincipalClientId,AzureCertificationServicePrincipalClientSecret,AzureResourceGroupName,AzureCertificationSubscriptionId
+        - component: pubsub.azure.servicebus
+          required-secrets: AzureServiceBusConnectionString, AzureCertificationTenantId, AzureCertificationServicePrincipalClientId, AzureCertificationServicePrincipalClientSecret
```
berndverst, line 94: You must not have spaces here. Also, you forgot to include **AzureServiceBusNamespace**

```suggestion
          required-secrets: AzureServiceBusConnectionString,AzureServiceBusNamespace, AzureCertificationTenantId,AzureCertificationServicePrincipalClientId,AzureCertificationServicePrincipalClientSecret
```

tests/certification/pubsub/azure/servicebus/components/authentication/service_bus.yaml
amulyavarote: Got it!!!

berndverst (approved):

berndverst (approved):

tests/certification/pubsub/azure/servicebus/go.mod
```diff
@@ -0,0 +1,158 @@
+module servicebusqueue_test
+
+go 1.18
```
berndverst, line 3:
```suggestion
go 1.19
```

berndverst (approved):

PR #1940, opened 2022-08-06
binding.eventhub: use same name schema for storage leases like on pubsub.eventhub

christle:
Hi,

We are currently adopting the naming change from this PR #1292 to the KEDA Eventhub Scaler. But the naming change includes only the pubsub component and not the eventhub binding. So there are two different naming schemas for storage leases, and this makes it more difficult to integrate that in the KEDA Scaler.

See  this discussion on KEDA: [https://github.com/kedacore/keda/issues/3022](https://github.com/kedacore/keda/issues/3022)

Before doing the KEDA change, it makes sense to use a unique naming schema for pubsub and binding. So I've made a proposal for it.

What do you think?

### DIFF
--- a/bindings/azure/eventhubs/eventhubs.go
+++ b/bindings/azure/eventhubs/eventhubs.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/Azure/azure-amqp-common-go/v3/aad"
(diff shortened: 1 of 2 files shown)


bindings/azure/eventhubs/eventhubs.go
```diff
@@ -154,6 +155,28 @@ func validate(connectionString string) error {
 	return err
 }
 
+func (aeh *AzureEventHubs) getStoragePrefixString() (string, error) {
```
mukundansundar, line 158:
```suggestion
func (a *AzureEventHubs) getStoragePrefixString() (string, error) {
```

bindings/azure/eventhubs/eventhubs.go
```diff
@@ -154,6 +155,28 @@ func validate(connectionString string) error {
...
+	hubName, err := aeh.validateAndGetHubName()
+	if err != nil {
+		return "", err
+	}
+
+	// empty string in the end of slice to have a suffix "-".
+	return strings.Join([]string{"dapr", hubName, aeh.metadata.consumerGroup, ""}, "-"), nil
+}
+
+func (aeh *AzureEventHubs) validateAndGetHubName() (string, error) {
```
mukundansundar, line 168:
```suggestion
func (a *AzureEventHubs) validateAndGetHubName() (string, error) {
```

mukundansundar (commented):
I agree that having consistency across binding and pubsub makes sense. 
@berndverst @tanvigour WDYT?

ItalyPaleAle (dismissed):
LGTM!

bindings/azure/eventhubs/eventhubs.go
christle: done

bindings/azure/eventhubs/eventhubs.go
christle: done

mukundansundar (approved):
lgtm.
@christle Thanks for the contribution!

tomkerkhove:
Thank you!

PR #1941, opened 2022-08-08
Initial creation of workflows building block

RyanLettieri:
Signed-off-by: Ryan Lettieri <ryanLettieri@microsoft.com>

# Description

Initial Creation of the workflows building block using temporal

## Issue reference

Related Proposal: https://github.com/dapr/dapr/issues/4576

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


berndverst:
Adding the `do-not-merge` label to make sure no maintainer merges this before the 1.9 release.

workflows/responses.go
```diff
@@ -0,0 +1,8 @@
+package workflows
+
+type StateResponse struct {
+	WfInfo    WorkflowStruct
+	StartTime string `json:"start_time"`
```
cgillum, line 5: Why `string` and not `time.Time`?

cgillum (commented):
Still making my way through this PR. Here's my first batch of comments.

tests/certification/workflows/temporal/temporal_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"time"
+)
+
+func TestTemporal(t *testing.T) {
+	log := logger.NewLogger("dapr.components")
+	ports, err := dapr_testing.GetFreePorts(2)
+	assert.NoError(t, err)
+
+	currentGrpcPort := ports[0]
+	currentHTTPPort := ports[1]
```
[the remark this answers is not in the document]
RyanLettieri: Certification test was added by mistake and was not ready. It is going to be removed in this PR and added in at a later date

workflows/responses.go
RyanLettieri: I was thinking that returning it as a string could provide more flexibility for formatting purposes or even flexibility for other backends if they return time as a string. If that doesn't make sense though, I'll gladly make the change to use time.Time

workflows/README.md
```diff
@@ -0,0 +1,11 @@
+# Workflows
+
+A workflow is custom application logic that consists of a set of tasks and or state transitions.
```
cgillum, line 3: Should we add a link to the proposal so that readers can get more context on this feature area?

workflows/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Implementing a new Workflow
+
+A compliant workflow needs to implement the `Workflow` inteface included in the [`workflow.go`](workflow.go) file.
```
cgillum, line 7:
```suggestion
A compliant workflow needs to implement the `Workflow` interface included in the [`workflow.go`](workflow.go) file.
```

workflows/requests.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+type WorkflowReference struct {
+	InstanceID string `json:"instance_id"`
+}
+
+// StartRequest is the object describing a Start Workflow request.
+type StartRequest struct {
+	Options      map[string]string `json:"workflow_options"`
+	WorkflowInfo WorkflowReference `json:"workflow_info"`
```
cgillum, line 16: Same comment I left on the other PR, but I suggest using a name such as `WorkflowReference` instead of `WorkflowInfo`. I think this will help with clarity.
```suggestion
	WorkflowReference WorkflowReference `json:"workflow_reference"`
```

workflows/responses.go
cgillum: Will there be any Go code that consumes this type? My concern is about whether 1) any Go code needs to consume this field and 2) whether it may be displayed somewhere. In the latter case, it would be a little awkward to have different components display the creation time using different formats. Not a big deal, but I wonder if it could cause problems, especially if we need to actually consume these values somewhere, like in the Dapr Dashboard. Having a strict type/format might also make it easier to display start times in a sorted manner.

workflows/temporal/temporal.go
```diff
@@ -0,0 +1,164 @@
...
+	wfStruct := workflows.WorkflowReference{InstanceID: run.GetID()}
+	return &wfStruct, nil
+}
+
+func (c *TemporalWF) Terminate(ctx context.Context, req *workflows.WorkflowReference) error {
+	c.logger.Debugf("terminating workflow")
+
+	err := c.client.TerminateWorkflow(ctx, req.InstanceID, "", "")
+	if err != nil {
+		return err
```
cgillum, line 100: Should we add a log statement for this case, like we do for the other error cases?

cgillum (commented):
Added a few more comments.

I noticed that .github/infrastructure/conformance/temporal/server/dynamicconfig/docker.yaml is empty. I'm not sure how conformance tests are supposed to work, but is this expected? Seems like there's another one for the certification test as well (also, I thought you were removing the certification tests?).

Also, there's a binary file in this PR: .github/infrastructure/conformance/temporal/worker/worker. Is this supposed to be there?

.github/infrastructure/conformance/temporal/worker/go.mod
```diff
@@ -0,0 +1,32 @@
...
+	github.com/golang/mock v1.6.0 // indirect
+	github.com/golang/protobuf v1.5.2 // indirect
+	github.com/google/uuid v1.3.0 // indirect
+	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0 // indirect
+	github.com/pborman/uuid v1.2.1 // indirect
+	github.com/pmezard/go-difflib v1.0.0 // indirect
+	github.com/robfig/cron v1.2.0 // indirect
+	github.com/stretchr/objx v0.3.0 // indirect
+	github.com/stretchr/testify v1.7.1 // indirect
+	go.temporal.io/api v1.8.0 // indirect
```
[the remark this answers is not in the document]
RyanLettieri: There seems to be a compatibility issue when running with that version:
*workflowServiceProxyServer does not implement workflowservice.WorkflowServiceServer (missing GetWorkerBuildIdOrdering method)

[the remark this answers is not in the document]
cgillum: Weird. But I guess it's not a blocker for this PR. We can revisit this later.

workflows/responses.go
RyanLettieri: Updated the temporal code to output the time using a time standard (RFC3339). It is still a string

cgillum (dismissed):
:shipit:

berndverst:
Conformance: https://github.com/dapr/components-contrib/actions/runs/3055926528
Certification: https://github.com/dapr/components-contrib/actions/runs/3055926532

berndverst (approved):

RyanLettieri:
This PR is being held open until the end of 1.9 in-case there're some hidden broken changes. Furthermore, since the API changes are not pushed into master, this isn't a strong need to have this merged into 1.9 The PR will be updated with the latest changes from master to ensure that it is not out of date. Once the code-freeze for 1.9 is complete, this can be merged into 1.10

berndverst (changes requested):
Please resolve conflicts and then we can get this merged into master.

tests/certification/go.mod
```diff
@@ -5,7 +5,7 @@ go 1.19
 require (
 	github.com/dapr/components-contrib v1.9.0-rc.1
 	github.com/dapr/dapr v1.9.0-rc.3
-	github.com/dapr/go-sdk v1.5.1-0.20221004175845-b465b1fa0721
+	github.com/dapr/go-sdk v0.8.1-0.20200925214701-2d699b6c653f
```
cgillum, line 8: I'm just now noticing this, but why are we using such an old version of the Go SDK in this PR? I would have expected us to use the latest v1.5.x version.

berndverst:
I'll merge once @cgillum can do another quick check and give an approval.

workflows/requests.go
```diff
@@ -0,0 +1,34 @@
...
+type StartRequest struct {
+	Options           map[string]string `json:"workflow_options"`
+	WorkflowReference WorkflowReference `json:"workflow_reference"`
+	WorkflowName      string            `json:"function_name"`
+	Parameters        interface{}       `json:"parameters"`
+}
+
+// CreateWorkerRequest is the object describing a Create Worker request.
+type CreateWorkerRequest struct {
+	TaskQueue     string         `json:"task_queue"`
```
cgillum, line 23: TaskQueue is a Temporal-specific concept and thus shouldn't belong outside of the temporal package. But is the `CreateWorkerRequest` being used anywhere?

workflows/requests.go
```diff
@@ -0,0 +1,34 @@
...
+	WorkerOptions worker.Options `json:"worker_options"`
+}
+
+// RegisterWorkflowAndActivitiesRequest is the object describing the wfs and acts to assign to a worker
+type RegisterWorkflowAndActivitiesRequest struct {
```
cgillum, line 28: Same question, is this being used anywhere? I couldn't find any references.

workflows/temporal/temporal.go
```diff
@@ -0,0 +1,168 @@
...
+	wfStruct := workflows.WorkflowReference{InstanceID: run.GetID()}
+	return &wfStruct, nil
+}
+
+func (c *TemporalWF) Terminate(ctx context.Context, req *workflows.WorkflowReference) error {
+	c.logger.Debugf("terminating workflow")
+
+	err := c.client.TerminateWorkflow(ctx, req.InstanceID, "", "")
+	if err != nil {
+		c.logger.Errorf("error when terminating workflow: %v", err)
```
cgillum, line 102: Now that I've had a bit more time to digest Go, I'm wondering if it would be more conventional to return a custom error instead of logging like this. The expectation is that the upstream caller takes care of logging for us:
```go
return fmt.Errorf("error terminating workflow: %w", err)
```
Same comment for all other places where we log and then return an error.

cgillum (commented):
I noticed there's a binary in this PR under /.github/infrastructure/conformance/temporal/worker named `worker`. Is that supposed to be checked into GitHub?

workflows/requests.go
berndverst: Follow up question: Is there a design doc for what the workflow interface actually should look like?

At this level there should generally only be things matching the protos and relevant helpers that apply to all components of the workflow type.

workflows/temporal/temporal.go
berndverst: I agree. Since `Terminate` is part of the Dapr runtime, the Runtime actually should log any error to the logger.

In this place we should just `return fmt.Errorf`

workflows/requests.go
berndverst: @RyanLettieri what you put here should match the proto and changes in https://github.com/dapr/dapr/pull/5158

berndverst:
> I noticed there's a binary in this PR under /.github/infrastructure/conformance/temporal/worker named `worker`. Is that supposed to be checked into GitHub?

The `worker` binary definitely needs to be removed @RyanLettieri. Seems your Docker container builds the binary from the source file anyway.

tests/certification/workflows/temporal/temporal_test.go
```diff
@@ -0,0 +1,31 @@
...
+}
```
cgillum, line 31: Does this file need to be removed? My previous comment was marked as resolved, and I was expecting this would have been removed. I'm not sure what the current expectation is. We don't appear to be actually testing anything here.

workflows/temporal/temporal.go
```diff
@@ -0,0 +1,167 @@
...
+
+	"go.temporal.io/api/enums/v1"
+	"go.temporal.io/sdk/client"
+
+	"github.com/dapr/components-contrib/workflows"
+	"github.com/dapr/kit/logger"
+)
+
+// Placeholder string for the task queue
+const TaskQueueString = "TestTaskQueue"
```
cgillum, line 31: It looks like this const isn't being used. Given the value, it looks like it doesn't belong in production code either.

cgillum (approved):
One more minor thing, but otherwise this looks good to me.

cgillum (approved):

go.mod
```diff
@@ -122,6 +122,15 @@ require (
 	k8s.io/utils v0.0.0-20221012122500-cfd413dd9e85
 )
 
+require (
```
berndverst, line 125: can you please clean this up and make sure there are only 2 `require` blocks here? The first block is the direct depedencies, the second one is the indirect ones.

go.mod
```diff
@@ -328,6 +337,8 @@ require (
 	github.com/yudai/golcs v0.0.0-20170316035057-ecda9a501e82 // indirect
 	github.com/yuin/gopher-lua v0.0.0-20210529063254-f4c35e4016d9 // indirect
 	go.opencensus.io v0.23.0 // indirect
+	go.temporal.io/api v1.12.0
```
berndverst, line 340: Don't mix direct and indirect dependencies in different blocks please.

berndverst (changes requested):
Your workflow conformance test is not running in CI/CD.

Please make sure you first run your conformance test locally.

Should be something like
`go test -v -tags=conftests -count=1 ./tests/conformance -run="TestWorkflowConformance/temporal"`

Only you will know the specific based on your PR.

image

berndverst:
I was able to verify the workflow locally with:

```
docker-compose -f ./.github/infrastructure/docker-compose-temporal.yml -p temporal up -d
go test -v -tags=conftests -count=1 ./tests/conformance -run="TestWorkflowsConformance/temporal"
=== RUN   TestWorkflowsConformance
=== RUN   TestWorkflowsConformance/temporal
=== RUN   TestWorkflowsConformance/temporal/init
time="2022-10-27T12:34:28.51554-07:00" level=debug msg="Temporal init start" instance=Bernds-MacBook-Pro.local scope=testLogger type=log ver=unknown
2022/10/27 12:34:28 INFO  No logger configured for temporal client. Created default one.
=== RUN   TestWorkflowsConformance/temporal/start
time="2022-10-27T12:34:28.526645-07:00" level=info msg="Start test running..." instance=Bernds-MacBook-Pro.local scope=workflowsTest type=log ver=unknown
time="2022-10-27T12:34:28.526677-07:00" level=debug msg="starting workflow" instance=Bernds-MacBook-Pro.local scope=testLogger type=log ver=unknown
time="2022-10-27T12:34:28.540806-07:00" level=debug msg="getting workflow data" instance=Bernds-MacBook-Pro.local scope=testLogger type=log ver=unknown
time="2022-10-27T12:34:33.546585-07:00" level=debug msg="getting workflow data" instance=Bernds-MacBook-Pro.local scope=testLogger type=log ver=unknown
time="2022-10-27T12:34:33.55706-07:00" level=debug msg="terminating workflow" instance=Bernds-MacBook-Pro.local scope=testLogger type=log ver=unknown
time="2022-10-27T12:34:33.576683-07:00" level=debug msg="getting workflow data" instance=Bernds-MacBook-Pro.local scope=testLogger type=log ver=unknown
time="2022-10-27T12:34:33.585117-07:00" level=info msg="Start test done." instance=Bernds-MacBook-Pro.local scope=workflowsTest type=log ver=unknown
--- PASS: TestWorkflowsConformance (5.07s)
    --- PASS: TestWorkflowsConformance/temporal (5.07s)
        --- PASS: TestWorkflowsConformance/temporal/init (0.01s)
        --- PASS: TestWorkflowsConformance/temporal/start (5.06s)
PASS
ok  	github.com/dapr/components-contrib/tests/conformance	5.112s
```

berndverst:
@RyanLettieri please open a Docs issue to create the new Workfows section in the Docs, and then list the temporal component as Alpha (due to the fact the building block is Alpha)

berndverst (approved):

RyanLettieri:
> @RyanLettieri please open a Docs issue to create the new Workfows section in the Docs, and then list the temporal component as Alpha (due to the fact the building block is Alpha)

@berndverst would this previously created issue suffice: https://github.com/dapr/docs/issues/2798 ?

PR #1942, opened 2022-08-08
Handle service bus exported errors

addjuarez:
# Description

Currently this component expects a wrapped amqpError typed error, but an unwrapped exported typed error can be returned. The exported error includes amqp errors but will not be found by unwrapping. We will check for the "connlost" error code instead for retries. https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/messaging/azservicebus/error.go

`error publishing event unto orders topic: rpc error: code = Internal desc = error when publish to topic orders in pubsub orderpubsub: (connlost): link detached, reason: *Error{Condition: amqp:resource-limit-exceeded, Description: The maximum entity size has been reached or exceeded for Topic: 'AAJ8779:TOPIC:ORDERS'. Size of entity in bytes:1073785042, Max entity size in bytes: 1073741824. For more information please see https://aka.ms/ServiceBusExceptions .  QuotaType: EntitySize Reference:1ebebe4c-f8ca-43ca-9f50-4ff8d4b4bf30, TrackingId:003257460000082908a6559d62f1993c_G22_B10, SystemTracker:aaj8779:Topic:orders, Timestamp:2022-08-08T23:16:13, Info: map[]}`

## Issue reference

#1899 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


ItalyPaleAle:
Thanks, this is very good. Does this need to be implemented in the binding too?

berndverst:
@addjuarez please address comment from @ItalyPaleAle

addjuarez:
From looking at the servicebus queues binding, it looks like we do not have retry logic for publishing. So, I don't think we need to add this change in the binding currently unless we want to add retries in the binding. 
 https://github.com/dapr/components-contrib/blob/master/bindings/azure/servicebusqueues/servicebusqueues.go#L139

ItalyPaleAle:
Ok perfect :)

ItalyPaleAle (approved):

berndverst (commented):
This should not have been merged, but LGTM, so it's ok.

PR #1943, opened 2022-08-09
bindings.cron: ensure it can resume correctly after context cancellation

ItalyPaleAle:
The cron binding did not resume correctly after a context cancellation and a new context. This is currently not (yet) used in Dapr so no end-user impact yet.


berndverst (approved):
👌

PR #1951, opened 2022-08-10
optimize: add e2etests buildtags for e2e

1046102779:
```shell
=== RUN   TestAuthentication
=== RUN   TestAuthentication/Valid_jwt_and_valid_seed
err: nats: no servers available for connection
    authentication_test.go:56: Did not expect error during connect nats: no servers available for connection
=== RUN   TestAuthentication/Valid_jwt_and_invalid_seed
err: dial tcp [::1]:4222: connect: network is unreachable
=== RUN   TestAuthentication/Invalid_jwt_and_valid_seed
err: nats: no servers available for connection
--- FAIL: TestAuthentication (0.00s)
    --- FAIL: TestAuthentication/Valid_jwt_and_valid_seed (0.00s)
    --- PASS: TestAuthentication/Valid_jwt_and_invalid_seed (0.00s)
    --- PASS: TestAuthentication/Invalid_jwt_and_valid_seed (0.00s)
FAIL
FAIL    github.com/dapr/components-contrib/tests/e2e/pubsub/jetstream/authentication    0.009s
FAIL
```

### DIFF
--- a/tests/e2e/pubsub/jetstream/authentication/authentication_test.go
+++ b/tests/e2e/pubsub/jetstream/authentication/authentication_test.go
@@ -1,7 +1,9 @@
+//go:build e2etests
+// +build e2etests
+
 package authentication_test
 
 import (
-	"strings"
 	"testing"
 
 	"github.com/dapr/components-contrib/pubsub"
@@ -54,11 +56,6 @@ func TestAuthentication(t *testing.T) {
 			}
 			err := js.Init(md)
 			if err != nil && !tC.expectError {
-				if strings.Contains(err.Error(), "network is unreachable") ||
-					strings.Contains(err.Error(), "no servers available for connection") {
-					log.Warn("jetstream server is unreachable")
-					return
-				}
 				t.Fatal("Did not expect error during connect", err.Error())
 			}
 			if err == nil && tC.expectError {
--- a/tests/e2e/pubsub/jetstream/authentication/authentication_test.go
+++ b/tests/e2e/pubsub/jetstream/authentication/authentication_test.go
@@ -54,7 +54,8 @@ func TestAuthentication(t *testing.T) {
 			}
 			err := js.Init(md)
 			if err != nil && !tC.expectError {
-				if strings.Contains(err.Error(), "network is unreachable") {
+				if strings.Contains(err.Error(), "network is unreachable") ||
+					strings.Contains(err.Error(), "no servers available for connection") {
 					log.Warn("jetstream server is unreachable")
 					return
 				}
--- a/tests/e2e/pubsub/jetstream/authentication/authentication_test.go
+++ b/tests/e2e/pubsub/jetstream/authentication/authentication_test.go
@@ -53,11 +53,11 @@ func TestAuthentication(t *testing.T) {
 				},
 			}
 			err := js.Init(md)
-			if strings.Contains(err.Error(), "network is unreachable") {
-				log.Warn("jetstream server is unreachable")
-				return
-			}
 			if err != nil && !tC.expectError {
+				if strings.Contains(err.Error(), "network is unreachable") {
+					log.Warn("jetstream server is unreachable")
+					return
+				}
 				t.Fatal("Did not expect error during connect", err.Error())
 			}
 			if err == nil && tC.expectError {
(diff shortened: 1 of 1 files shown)


tests/e2e/pubsub/jetstream/authentication/authentication_test.go
```diff
@@ -53,6 +54,11 @@ func TestAuthentication(t *testing.T) {
 			}
 			err := js.Init(md)
 			if err != nil && !tC.expectError {
+				if strings.Contains(err.Error(), "network is unreachable") ||
+					strings.Contains(err.Error(), "no servers available for connection") {
```
daixiang0, line 58: Why it can not start as expected?

1046102779: When using `go test -v ./...`, jetstream is not actually started locally. At present, if this unit test is to pass, the jetstream server must be started locally through ` docker-compose.yaml`.

1046102779: Another problem:

I passed the unit tests last week, but after merging the community code yesterday, I found that the unit tests could not run, and there were the following errors:

image

image

ItalyPaleAle:
Normally, tests in the `e2e` folder don't run when running `make test`.

However, I understand the issue and I think it should be fixed. But, if the server isn't ready, the test should be skipped entirely and not just avoiding a fatal log.

See how `binding.ipfs` does it: if you want to run the test, you need to pass an environmental variable; otherwise, the entire test is skipped:

https://github.com/dapr/components-contrib/blob/master/bindings/ipfs/ipfs_test.go#L36-L43

ItalyPaleAle (changes requested):
Please change to use an env var to make the tests run. See example in `bindings.ipfs`: https://github.com/dapr/components-contrib/blob/master/bindings/ipfs/ipfs_test.go#L36-L43

berndverst:
If this is an integration test I think we should use build tags! Please don't use environment variables to trigger the test unless this is needed for configuring the server etc, still I think build tags are the way to go.

FYI @ItalyPaleAle

1046102779:
[berndverst](https://github.com/berndverst)
[ItalyPaleAle](https://github.com/ItalyPaleAle)

/cc @berndverst @ItalyPaleAle. Thanks, i added a e2etest buildtag for the unit test jetstream  file.

1046102779:
> Please change to use an env var to make the tests run. See example in `bindings.ipfs`: https://github.com/dapr/components-contrib/blob/master/bindings/ipfs/ipfs_test.go#L36-L43

IPFS_TEST: "1"

```golang
      if utils.IsTruthy(os.Getenv("IPFS_TEST")) {
                  log.Println("IPFS_TEST env var is not set to a truthy value; skipping tests")
                  os.Exit(0)
      }
```

`IPFS_TEST=1` don't execute this unit test. Why not 0?

ItalyPaleAle:
> IPFS_TEST=1 don't execute this unit test. Why not 0?

It should! Did you `export IPFS_TEST=1`?

> If this is an integration test I think we should use build tags! Please don't use environment variables to trigger the test unless this is needed for configuring the server etc, still I think build tags are the way to go.

Whatever the boss (@berndverst) says :) However, the downside of build tags is that you need to configure your IDE too or the language server won't scan them...

ItalyPaleAle (approved):

mukundansundar (approved):

1046102779:
> IPFS_TEST

the env variable `IPFS_TEST` is a neutral word, if  change to `IPFS_TEST_OFF` is more appropriate?

ItalyPaleAle:
Oh I see now. It is missing a `!` before utils.IsTruthy. Want to open another PR to fix that?

1046102779:
> Oh I see now. It is missing a `!` before utils.IsTruthy. Want to open another PR to fix that?

hah, i do it.

berndverst:
Once we create certification tests for this component we should move this into a certification test instead! Does someone know how to enable NATS jetstream with a Docker image? Or ideally how to create the Docker-compose file? A lot of configuration seems to be required there.

berndverst (approved):

1046102779:
> Once we create certification tests for this component we should move this into a certification test instead! Does someone know how to enable NATS jetstream with a Docker image? Or ideally how to create the Docker-compose file? A lot of configuration seems to be required there.

We may need readme and makefile in this e2e example.

PR #1965, opened 2022-08-12
fix(kafka): ConsumeClaim should not return err when retry recovery failed

zcong1993:
ConsumeClaim should not return err when retry recovery failed.

# Description

`ConsumeClaim` return error will cause restart loop of consume, if initialOffset is oldest, this message will be consumed forever. So we should log error and let it go like no retry does.

## Issue reference

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/internal/component/kafka/consumer.go
+++ b/internal/component/kafka/consumer.go
@@ -44,7 +44,7 @@ func (consumer *consumer) ConsumeClaim(session sarama.ConsumerGroupSession, clai
 			}, func() {
 				consumer.k.logger.Infof("Successfully processed Kafka message after it previously failed: %s/%d/%d [key=%s]", message.Topic, message.Partition, message.Offset, asBase64String(message.Key))
 			}); err != nil {
-				return err
+				consumer.k.logger.Errorf("Error processing Kafka message: %s/%d/%d [key=%s]. Error: %v.", message.Topic, message.Partition, message.Offset, asBase64String(message.Key), err)
 			}
 		} else {
 			err := consumer.doCallback(session, message)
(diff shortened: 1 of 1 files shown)


internal/component/kafka/consumer.go
```diff
@@ -44,7 +44,7 @@ func (consumer *consumer) ConsumeClaim(session sarama.ConsumerGroupSession, clai
 			}, func() {
 				consumer.k.logger.Infof("Successfully processed Kafka message after it previously failed: %s/%d/%d [key=%s]", message.Topic, message.Partition, message.Offset, asBase64String(message.Key))
 			}); err != nil {
-				return err
+				consumer.k.logger.Errorf("Error processing Kafka message: %s/%d/%d [key=%s]. Error: %v.", message.Topic, message.Partition, message.Offset, asBase64String(message.Key), err)
```
ItalyPaleAle, line 47: To differentiate this error message, can you please log the number of failed attempts too? For example:

`Failed processing Kafka message after %d attempts: %s/%d/%d [key=%d]. Error: %v.`

ItalyPaleAle (changes requested):
Thanks for catching this! I agree this is more correct. However, could you please modify the error message a bit?

internal/component/kafka/consumer.go
zcong1993: > To differentiate this error message, can you please log the number of failed attempts too? For example:
> 
> `Failed processing Kafka message after %d attempts: %s/%d/%d [key=%d]. Error: %v.`

The `retry.NotifyRecover` seems can not get attempts times when failed.

zcong1993: How about this `Failed processing Kafka message after it previously failed: %s/%d/%d [key=%d]. Error: %v.` ?

ItalyPaleAle: Sounds good, but if I can wordsmith a bit: "Too many failed attempts at processing Kafka message." (+ the rest)

zcong1993: done

ItalyPaleAle (approved):

berndverst (approved):

PR #1973, opened 2022-08-17
feature(secretstore): add context to api and tencent/ssm

1046102779:
1. Add context to secretstores API;
2. Add a tencent ssm in secretstores.

/cc @berndverst @ItalyPaleAle


secretstores/tencent/ssm/ssm.go
```diff
@@ -0,0 +1,179 @@
...
+	var val string
+	ssResponse := ssResp.Response
+	if ssResponse.SecretBinary != nil {
+		bts, err := base64.StdEncoding.DecodeString(*ssResponse.SecretBinary)
+		if err != nil {
+			return response, err
+		}
+		val = string(bts)
+	}
+	if ssResponse.SecretString != nil {
```
ItalyPaleAle, line 109: Should probably be an `else if` ?

secretstores/tencent/ssm/ssm.go
```diff
@@ -0,0 +1,179 @@
...
+		val = *ssResponse.SecretString
+	}
+	response.Data[*ssResponse.SecretName] = val
```
ItalyPaleAle, line 112: Please add a nil-check before this line, in case `ssResponse.SecretName` is nil

secretstores/tencent/ssm/ssm.go
```diff
@@ -0,0 +1,179 @@
...
+	response.Data[RequestID] = *ssResponse.RequestId
```
ItalyPaleAle, line 113: Same, a nil check before this line too please

secretstores/tencent/ssm/ssm.go
```diff
@@ -0,0 +1,179 @@
...
+	ssmReq.SecretName = &req.Name
+	ssmReq.VersionId = &versionID
+	ssResp, err := s.client.GetSecretValueWithContext(ctx, ssmReq)
+	if err != nil {
+		return response, err
+	}
+	var val string
+	ssResponse := ssResp.Response
+	if ssResponse.SecretBinary != nil {
+		bts, err := base64.StdEncoding.DecodeString(*ssResponse.SecretBinary)
```
ItalyPaleAle, line 103: Wont this return a binary string that could be corrupted? Wouldn't it be best to return the secret as base64-encoded here?

For example, imagine the secret was a 256-bit symmetric key... You probably don't want that in a `string` variable. If it has to be a string, it should probably be base64-encoded

secretstores/metadata.go
```diff
@@ -20,3 +20,10 @@ const DefaultSecretRefKeyName = "_value"
 type Metadata struct {
 	Properties map[string]string `json:"properties,omitempty"`
 }
+
+type ParameterStoreMetaData struct {
```
ItalyPaleAle, line 24: If this is specific to tencent/ssm, please move it to the tencent/ssm/ssm.go file

secretstores/tencent/ssm/ssm.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func (s *ssmSecretStore) BulkGetSecret(ctx context.Context, req secretstores.BulkGetSecretRequest) (secretstores.BulkGetSecretResponse, error) {
+	response := secretstores.BulkGetSecretResponse{
+		Data: map[string]map[string]string{},
+	}
+	versionID := req.Metadata[VersionID]
+
+	var start uint64 = 0
+	names, err := s.getSecretNames(ctx, &start)
```
ItalyPaleAle, line 125: Shouldn't this read only the secret whose names is in the request object? Rather than _all_ secrets in the vault, which could be a lot?

ItalyPaleAle (changes requested):
Thanks for this PR :) Please see my review comments.

secretstores/tencent/ssm/ssm.go
1046102779: I originally thought that the upstream micrologic does not need to care about the decoding problem. After ensuring that `SecretString` and `SecretBinary` are aligned, they are returned to the micrologic server, so that when the micrologic server gets the value, they don't need to care what type it is. If we want to transfer primitive values here, we may also need to tell the type of values.

1046102779: Doing so, it may be a little burden.

secretstores/tencent/ssm/ssm.go
1046102779: Instead of getting all the key-value APIs at once, we only get the key list through the `ListKeys` API first, and then traverse to get all the values.

secretstores/metadata.go
1046102779: This is redundant. My original idea is that the current secretid and secretkey, token, and region names are not standardized, so I wanted to unify them.

Can we make some sematic conversions? Otherwise the configuration names of each component are not uniform, and the experience is very poor.

Later, I want to mention such issues and PRs separately

secretstores/tencent/ssm/ssm.go
ItalyPaleAle: It depends on what the value inside `SecretBinary` is. If it's just an ASCII (or at most, UTF-8) encoded string, then decoding the base64 and converting it to `string` should be fine.

BUT, if the value is a binary string (i.e something that could contain byte sequences that are NOT valid UTF-8 sequences), then converting to `string` could break the content.

secretstores/tencent/ssm/ssm.go
ItalyPaleAle: You're right, I misunderstood how the bulk get secrets API worked. This seems fine.

The problem is: what happens if the list of secrets changes between the time you invoke `ListKeys` and when you retrieve each secret?

1046102779: This second-level error, I think it should be acceptable.

1046102779: In fact, I think this batch access to key-values `BulkGetSecrets`API is a bit problematic. Because most secret cloud products need to specify `versionid`, in fact, this API cannot specify different versions, and users do not know which keys are there. So it feels a bit tasteless

secretstores/metadata.go
1046102779: Agree. Thanks.

secretstores/tencent/ssm/ssm.go
1046102779: > Wont this return a binary string that could be corrupted? Wouldn't it be best to return the secret as base64-encoded here?
> 
> For example, imagine the secret was a 256-bit symmetric key... You probably don't want that in a `string` variable. If it has to be a string, it should probably be base64-encoded

After base64 decoding, it is transmitted through the body, is it possible to cause damage?

ItalyPaleAle: Yes. Just converting it to `string` could cause damage. If it's binary, please leave it in a `[]byte` or leave it as a base64-encoded string.

1046102779: I added a SecretValueType Field in return map to classfiy text and binary data.

1046102779: > Yes. Just converting it to `string` could cause damage. If it's binary, please leave it in a `[]byte` or leave it as a base64-encoded string.

Ok, I see it. Thanks.

secretstores/tencent/ssm/ssm.go
```diff
@@ -0,0 +1,189 @@
...
+	ssmReq.SecretName = &req.Name
+	ssmReq.VersionId = &versionID
+	ssResp, err := s.client.GetSecretValueWithContext(ctx, ssmReq)
+	if err != nil {
+		return response, err
+	}
+	var val string
+	ssResponse := ssResp.Response
+	if ssResponse.SecretBinary != nil {
+		response.Data[ValueType] = fmt.Sprintf("%d", BINARY_SECRET_VALUE)
```
ItalyPaleAle, line 112: Nit: `strconv.Itoa(BINARY_SECRET_VALUE)` would be much faster

ItalyPaleAle (dismissed):

ItalyPaleAle:
@berndverst  this LGTM but it does change the APIs by adding contexts, so we'll need to coordinate the rollout. I think the best thing would be if one of us created a fork of dapr/dapr for this PR....

1046102779:
> @berndverst this LGTM but it does change the APIs by adding contexts, so we'll need to coordinate the rollout. I think the best thing would be if one of us created a fork of dapr/dapr for this PR....

I will create a PR to  handle it in `dapr/dapr`.

secretstores/tencent/ssm/ssm.go
1046102779: Ok. Thanks.

daixiang0 (approved):

1046102779:
Everything is ready, can you merge this PR first ?

/cc @yaron2 @artursouza @berndverst .

artursouza:
All validations are failing

yaron2:
@1046102779 does this need an updated version of dapr/dapr? See failing builds

1046102779:
> 

Ok, I try it.

1046102779:
> @1046102779 does this need an updated version of dapr/dapr? See failing builds

How is this dapr/dapr version specified? Where is it specified?

secretstores/tencentcloud/ssm/ssm.go
```diff
@@ -0,0 +1,189 @@
...
+	ListSecretsWithContext(ctx context.Context, request *ssm.ListSecretsRequest) (*ssm.ListSecretsResponse, error)
+	GetSecretValueWithContext(ctx context.Context, request *ssm.GetSecretValueRequest) (*ssm.GetSecretValueResponse, error)
+}
+
+type ssmSecretStore struct {
+	client ssmClient
+	logger logger.Logger
+}
+
+var _ secretstores.SecretStore = (*ssmSecretStore)(nil)
```
ItalyPaleAle, line 59: Nit: because `NewSSM` returns an interface now (secretstores.SecretStore), these type assertions aren't needed anymore

ItalyPaleAle (dismissed):

secretstores/tencentcloud/ssm/ssm.go
1046102779: Ok. I delete it.

1046102779:
Here I also used replace to point to temporary `1046102779/dapr`.

image

berndverst:
Certification test: https://github.com/dapr/components-contrib/actions/runs/2950848345

1046102779:
> Certification test: https://github.com/dapr/components-contrib/actions/runs/2950848345

fixed.

daixiang0 (dismissed):

secretstores/gcp/secretmanager/secretmanager_test.go
```diff
@@ -18,6 +18,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"golang.org/x/net/context"
```
berndverst, line 21: Use `context` from the standard library instead.

berndverst (changes requested):

secretstores/gcp/secretmanager/secretmanager_test.go
1046102779: done.

1046102779:
> @1046102779 see this build failure: https://github.com/dapr/components-contrib/runs/8148144426?check_suite_focus=true

Fixed.

image

daixiang0 (approved):

ItalyPaleAle (dismissed):

berndverst (dismissed):

berndverst (approved):

berndverst (approved):

berndverst (approved):

berndverst (approved):

berndverst (approved):

Issue #1970, opened 2022-08-16
Crashes and errors with statestore and CosmosDB.


alexandrejulien:
## Expected Behavior

Statestore with cosmosdb occurs crashes on dapr sidecar with Dapr 1.7.4
Using .NET App with Dapr Client for ASP.NET SDK with .Net 6 runtime.

This my dapr component :
```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: shared-statestore
  namespace: azure
spec:
  type: state.azure.cosmosdb
  version: v1
  metadata:
  - name: url
    value: https://XXXXXXXXXXXXXXX.azure.com:443/
  - name: masterKey
    value: XXXXXXXXXX
  - name: database
    value: rec-statestores
  - name: collection
    value: shared-state
  - name: keyPrefix
    value: none
```

## Actual Behavior

The dapr process crash probably when calling StateStore on CosmosDb.
No problem on Redis statestore.

2022-08-10T08:38:11.094454210Z goroutine 6630 [running]:
2022-08-10T08:38:11.094459310Z github.com/dapr/components-contrib/state/azure/cosmosdb.(StateStore).Get(0xc000b74280, 0xc000e7cab0)
2022-08-10T08:38:11.094465710Z     /home/runner/go/pkg/mod/github.com/dapr/c**@v1.7.1/state/azure/cosmosdb/cosmosdb.go:241 +0x61a
2022-08-10T08:38:11.094471010Z github.com/dapr/dapr/pkg/grpc.(api).GetState.func1({0x398e6c0, 0x4510801})
2022-08-10T08:38:11.106194011Z ℹ️
2022-08-10T08:38:11.106214211Z terminated signal received: shutting down
2022-08-10T08:38:11.106280411Z ❌  The daprd process exited with error code: exit status 2
2022-08-10T08:38:11.106290511Z ❌  Error exiting Dapr: exit status 2
2022-08-10T08:38:11.107867311Z ✅  Exited App successfully

## Steps to Reproduce the Problem

Using Statestore with CosmosDb with prefix and without prefix (2 dapr components)

berndverst:
Cosmos DB seems to indicate the data you are trying to read is binary. It tries to decode to string. The problem is that your data isn't actually binary and so decoding fails.

https://github.com/dapr/components-contrib/blob/release-1.7/state/azure/cosmosdb/cosmosdb.go#L240

[...]

PR #1974, opened 2022-08-17
Fix CosmosDB state store handling of nulls 

berndverst:
Addresses some issue related to the handling of nulls / nil values on Cosmos DB State store. #1970 

- No longer treats empty / nil values as binary data on Set operations.
- Allows returning of nil values in Get operation.
- Falls back to returning raw string if a byte string cannot be decoded (in Get operation).

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/state/azure/cosmosdb/cosmosdb.go
+++ b/state/azure/cosmosdb/cosmosdb.go
@@ -249,7 +249,18 @@ func (c *StateStore) Get(req *state.GetRequest) (*state.GetResponse, error) {
 	}
 
 	if items[0].IsBinary {
-		bytes, _ := base64.StdEncoding.DecodeString(items[0].Value.(string))
+		if items[0].Value == nil || len(items[0].Value.(string)) == 0 {
+			return &state.GetResponse{
+				Data: make([]byte, 0),
+				ETag: ptr.String(items[0].Etag),
+			}, nil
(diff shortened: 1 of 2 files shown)


state/azure/cosmosdb/cosmosdb.go
```diff
@@ -249,7 +249,18 @@ func (c *StateStore) Get(req *state.GetRequest) (*state.GetResponse, error) {
 	}
 
 	if items[0].IsBinary {
-		bytes, _ := base64.StdEncoding.DecodeString(items[0].Value.(string))
+		if items[0].Value == nil || len(items[0].Value.(string)) == 0 {
```
ItalyPaleAle, line 252: You may need to do a check if the type assertion is possible, before asserting as `(string)` right away

ItalyPaleAle (dismissed):
Approved but with one comment

state/azure/cosmosdb/cosmosdb.go
berndverst: It is always string because even binary data is saved as a base64 encoded string. This is why the decoding below always applies - except when the value in the database is literally null.

ItalyPaleAle (approved):

PR #1992, opened 2022-08-22
add nil/empty test cases for state conformance

berndverst:

# Description

Add test cases for nil / empty string state store conformance.


berndverst:
@yaron2 @artursouza

Some components allow empty string in set operations and others do not. Some handle null values, empty bytes etc well and others do not.

Additional conformance test run for Azure components etc: https://github.com/dapr/components-contrib/actions/runs/2906247028

Do we want to standardize how nil / empty values are handled, or do we want to completely disallow them?

berndverst:
Latest run: https://github.com/dapr/components-contrib/actions/runs/2906323914

PR #1997, opened 2022-08-23
Components' init functions should return interfaces

ItalyPaleAle:
# Description

Components init methods should return an interface and not a concrete type.

Companion PR for https://github.com/dapr/dapr/pull/5062
Note that merging this PR (without changes in the runtime) will make Dapr fail to compile.


berndverst (changes requested):
Can you please fix up `bindings/apns/apns_test.go` to cast the binding interface to the specific APNS binding type now.

Similar for `bindings/aws/s3/s3_test.go` with S3.

ItalyPaleAle:
Fixed @berndverst

berndverst (dismissed):

berndverst (changes requested):
```
Error: ../../embedded/components.go:32:14: undefined: nameresolution.New
Error: ../../embedded/components.go:35:14: undefined: nameresolution.New
Error: ../../embedded/components.go:38:14: undefined: nameresolution.New
```

Could you please update the component registration / loading code in the embedded sidecar?

The pinned dapr/dapr already includes your commits for this (I updated it).

berndverst (approved):

PR #2015, opened 2022-08-26
Enable conformance tests for rethinkdb

berndverst:
# Description

This makes RethinkDB a Beta component.

Note:
RethinkDB does **not support ETag**. As such the ETag capability has been removed from the set of returned capabilities.

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/.github/infrastructure/docker-compose-rethinkdb.yml
+++ b/.github/infrastructure/docker-compose-rethinkdb.yml
@@ -1,0 +1,9 @@
+version: '2'
+
+services:
+  rethinkdb:
+    image: rethinkdb:2.4
+    ports:
+      - 8081:8080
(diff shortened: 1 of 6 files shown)


berndverst:
I examined the various update policies -- `error`, `replace` and `update` but found that these only apply to primary key collisions. There is no such concept has Etag. Perhaps this can be emulated in the future, but for the time being this state store should not advertise being an actor state store or offer ETag support.

berndverst:
Also FYI, transactions on multiple documents are not actually atomic for RethinkDB according to their FAQ.
Should we remove the Transaction feature also?

ItalyPaleAle (approved):

PR #1972, opened 2022-08-16
Certification test for Azure CosmosDB Statestore

shivamkm07:
Signed-off-by: shivam <shivamkm07@gmail.com>

# Description

Certification test for Azure CosmosDB statestore

## Issue reference

Please reference the issue this PR will close: #1138 

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #2035, opened 2022-08-30
Upgrading cosmosdb statestore to track2SDK

RyanLettieri:
Signed-off-by: Ryan Lettieri <ryanLettieri@microsoft.com>

# Description

Updating cosmosdb state to use track2 sdk


berndverst:
@ItalyPaleAle can you do the first review of this?

berndverst:
Let's merge this AFTER https://github.com/dapr/components-contrib/pull/1972

This way we can make sure that the SDK upgrade didn't break the certification test.

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -15,17 +15,18 @@ package cosmosdb
 
 import (
 	// For go:embed.
```
ItalyPaleAle, line 17: Delete this comment and remove the `_ "embed"` import as those are not necessary anymore. Also, you should delete everything in the `storedprocedures` folder

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -80,31 +67,27 @@ type CosmosOperation struct {
...
-	versionSpName         = "__daprver__"
 	metadataPartitionKey  = "partitionKey"
 	unknownPartitionKey   = "__UNKNOWN__"
 	metadataTTLKey        = "ttlInSeconds"
 	statusTooManyRequests = "429" // RFC 6585, 4
 	statusNotFound        = "NotFound"
 )
 
+// Value used for timeout durations
+const timeoutValue = 20
```
ItalyPaleAle, line 87: Misses a duration. Is this in seconds? SHould be `20 * time.Second` then

I would also consider calling this something like `defaultTimeout`

ItalyPaleAle: Also, nit, but plese put all constants in the same block (see lines 79-84 above)

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -80,31 +67,27 @@ type CosmosOperation struct {
...
+
 // NewCosmosDBStateStore returns a new CosmosDB state store.
-func NewCosmosDBStateStore(logger logger.Logger) state.Store {
+func NewCosmosDBStateStore(logger logger.Logger) *StateStore {
```
ItalyPaleAle, line 90: Return type must be an interface now, so `state.Store`

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -147,63 +130,50 @@ func (c *StateStore) Init(meta state.Metadata) error {
 	}
 
 	// Create the client; first, try authenticating with a master key, if present
-	var config *documentdb.Config
+	var client *azcosmos.Client
 	if m.MasterKey != "" {
-		config = documentdb.NewConfig(&documentdb.Key{
-			Key: m.MasterKey,
-		})
+		cred, keyErr := azcosmos.NewKeyCredential(m.MasterKey)
```
ItalyPaleAle, line 135:
```go
var cred azcosmos.KeyCredential
cred, err = azcosmos.NewKeyCredential(...)
```

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -147,63 +130,50 @@ func (c *StateStore) Init(meta state.Metadata) error {
...
+		if keyErr != nil {
+			return keyErr
+		}
+		client, err = azcosmos.NewClientWithKey(m.URL, cred, nil)
+		if err != nil {
+			return err
+		}
 	} else {
 		// Fallback to using Azure AD
 		env, errB := azure.NewEnvironmentSettings("cosmosdb", meta.Properties)
```
ItalyPaleAle, line 145: Same here, and on line 149. Would be best to avoid defining new kinds of error variables.

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -80,31 +67,27 @@ type CosmosOperation struct {
...
 )
 
+// Value used for timeout durations
+const timeoutValue = 20
+
 // NewCosmosDBStateStore returns a new CosmosDB state store.
-func NewCosmosDBStateStore(logger logger.Logger) state.Store {
+func NewCosmosDBStateStore(logger logger.Logger) *StateStore {
 	s := &StateStore{
 		features: []state.Feature{state.FeatureETag, state.FeatureTransactional, state.FeatureQueryAPI},
```
ItalyPaleAle, line 92: Is there a reason why the features need to be stored as a property in the object? They could just be returned by the `Features()` method without being a property in the object.

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -287,47 +245,33 @@ func (c *StateStore) Set(req *state.SetRequest) error {
...
-			}
-			return backoff.Permanent(innerErr)
-		}
-		return nil
-	}, func(err error, d time.Duration) {
-		c.logger.Warnf("CosmosDB state store Set request failed: %v; retrying in %s", err, d)
-	}, 20*time.Second)
 
+	pk := azcosmos.NewPartitionKeyString(partitionKey)
+	_, err = c.client.UpsertItem(context.Background(), pk, marshalled, &options)
```
ItalyPaleAle, line 271: Please make sure to add a timeout (using the defaultTimeout constant) to the context

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -377,95 +311,107 @@ func (c *StateStore) Delete(req *state.DeleteRequest) error {
...
-			item, err := createUpsertItem(c.contentType, req, partitionKey)
+			marshalled, err := json.Marshal(req.Value)
 			if err != nil {
 				return err
 			}
-			upsertOperation := CosmosOperation{
-				Item: item,
-				Type: upsertOperationType,
+
+			if req.ETag != nil {
```
ItalyPaleAle, line 343: Should probably check if not empty too:

```
if req.ETag != nil && *req.ETag != ""
```

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -377,95 +311,107 @@ func (c *StateStore) Delete(req *state.DeleteRequest) error {
...
 			req := o.Request.(state.DeleteRequest)
 
-			deleteOperation := CosmosOperation{
-				Item: CosmosItem{
-					ID:           req.Key,
-					Value:        "", // Value does not need to be specified
-					PartitionKey: partitionKey,
-				},
-				Type: deleteOperationType,
+			if req.ETag != nil {
```
ItalyPaleAle, line 357: Should probably check if not empty too:

```
if req.ETag != nil && *req.ETag != ""
```

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -377,95 +311,107 @@ func (c *StateStore) Delete(req *state.DeleteRequest) error {
...
-			if isTooManyRequestsError(innerErr) {
-				return innerErr
-			}
-			return backoff.Permanent(innerErr)
-		}
-		return nil
-	}, func(err error, d time.Duration) {
-		c.logger.Warnf("CosmosDB state store Multi request failed: %v; retrying in %s", err, d)
-	}, 20*time.Second)
+	batchResponse, err := c.client.ExecuteTransactionalBatch(context.Background(), batch, nil)
```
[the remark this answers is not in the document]
ItalyPaleAle: Perhaps this should be configurable via a metadata option, since there can be many operations here?

state/azure/cosmosdb/cosmosdb_query.go
```diff
@@ -14,20 +14,26 @@ limitations under the License.
...
 
 	"github.com/dapr/components-contrib/state"
 	"github.com/dapr/components-contrib/state/query"
 )
 
+// Internal query object is created here since azcosmos has no notion of a query object
+type InternalQuery struct {
+	query      string
+	parameters []azcosmos.QueryParameter
+}
```
ItalyPaleAle, line 34: Nit: newline after this please

state/azure/cosmosdb/cosmosdb_query.go
```diff
@@ -122,45 +128,59 @@ func (q *Query) Finalize(filters string, qq *query.Query) error {
...
-	resp, err := client.QueryDocuments(collection, &q.query, &items, opts...)
-	if err != nil {
-		return nil, "", err
+
+	pk := azcosmos.NewPartitionKeyBool(true)
+	queryPager := client.NewQueryItemsPager(q.query.query, pk, opts)
+
+	token := ""
+	for queryPager.More() {
+		queryResponse, innerErr := queryPager.NextPage(context.Background())
```
ItalyPaleAle, line 162: Timeout here too please

state/azure/cosmosdb/cosmosdb_query.go
```diff
@@ -122,45 +128,59 @@ func (q *Query) Finalize(filters string, qq *query.Query) error {
...
+		}
 	}
-	token := resp.Header.Get(documentdb.HeaderContinuation)
 
 	ret := make([]state.QueryItem, len(items))
+	var err error
 	for i := range items {
 		ret[i].Key = items[i].ID
-		ret[i].ETag = ptr.String(items[i].Etag)
+		ret[i].ETag = &items[0].Etag
```
ItalyPaleAle, line 177: did you mean `&items[i].Etag`?

ItalyPaleAle (changes requested):

berndverst:
Conformance: https://github.com/dapr/components-contrib/actions/runs/3018674281
Certification: https://github.com/dapr/components-contrib/actions/runs/3018674217

berndverst:
@RyanLettieri please make sure that the conformance test for CosmosDB State store passes in this run: https://github.com/dapr/components-contrib/actions/runs/3019169073

You can manually launch another run via `/ok-to-test` in a new comment.

ealsur (dismissed):

berndverst:
@RyanLettieri converting this to draft until you get all tests to pass locally

RyanLettieri:
Cert: https://github.com/dapr/components-contrib/actions/runs/3063891151

Conformance: https://github.com/dapr/components-contrib/actions/runs/3063891192

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -377,95 +346,96 @@ func (c *StateStore) Delete(req *state.DeleteRequest) error {
 
 // Multi performs a transactional operation. succeeds only if all operations succeed, and fails if one or more operations fail.
 func (c *StateStore) Multi(request *state.TransactionalStateRequest) error {
-	operations := []CosmosOperation{}
+	if len(request.Operations) == 0 {
+		c.logger.Debugf("No Operations Provided")
+		return nil
+	}
+	partitionKey := request.Metadata["partitionKey"]
```
berndverst, line 353: You should use `populatePartitionMetadata(key, request.Metadata)`

That way it will use the item key if the `partitionKey` isn't explicitly defined.

state/azure/cosmosdb/cosmosdb.go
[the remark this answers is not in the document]
berndverst: @RyanLettieri please address Alessandro's comment.

berndverst (changes requested):
Left a comment about the partition key that needs to be addressed.

And please look at Alessandro's comment about the timeout.

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -377,95 +346,103 @@ func (c *StateStore) Delete(req *state.DeleteRequest) error {
...
+		// We can inspect the individual operation results
+		for index, operation := range batchResponse.OperationResults {
+			c.logger.Debugf("Operation %v completed with status code %d", index, operation.StatusCode)
+		}
+	} else {
+		// Transaction failed, look for the offending operation
+		for index, operation := range batchResponse.OperationResults {
+			if operation.StatusCode != http.StatusTooManyRequests {
+				c.logger.Debugf("Transaction failed due to operation %v which failed with status code %d", index, operation.StatusCode)
+				return nil
```
ItalyPaleAle, line 428: Should this return an error instead?

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -377,95 +346,103 @@ func (c *StateStore) Delete(req *state.DeleteRequest) error {
...
+	if batchResponse.Success {
+		// Transaction succeeded
+		// We can inspect the individual operation results
+		for index, operation := range batchResponse.OperationResults {
+			c.logger.Debugf("Operation %v completed with status code %d", index, operation.StatusCode)
+		}
+	} else {
+		// Transaction failed, look for the offending operation
+		for index, operation := range batchResponse.OperationResults {
+			if operation.StatusCode != http.StatusTooManyRequests {
```
ItalyPaleAle, line 426: Why this here?

ItalyPaleAle: I understand that this was here from before, but before this was in a retry block... Do we need this check now? What if the error is a 429, what should we do?

ItalyPaleAle (changes requested):

ItalyPaleAle (dismissed):

artursouza:
I propose this component comes out as v2 and we keep the old one as v1. Then, we update the tests to run against both. I don't think our certification tests for state store has validations cross versions - how data written from an older version can be read from the new version - so, given a significant change like this, there would be less risk to upgrade to a new version. I know it is a subjective call but I am concerned about the cross-version compatibility with such change refactoring (or rewrite).

/cc @yaron2 thoughts?

yaron2:
> I propose this component comes out as v2 and we keep the old one as v1. Then, we update the tests to run against both. I don't think our certification tests for state store has validations cross versions - how data written from an older version can be read from the new version - so, given a significant change like this, there would be less risk to upgrade to a new Dapr release if customers can opt in to use the new coponent. I know it is a subjective call but I am concerned about the cross-version compatibility with such change as this rewrite.
> 
> /cc @yaron2 thoughts?

As discussed, I agree. I am concernrd the new SDK might carry bugs and changes that are not cought by our tests.

ItalyPaleAle:
The concern I have with keeping the old "v1" around is that we need to maintain the dependency on the a8m SDK, which is sort-of unmaintained right now and I'd love to get rid of that for good.

Is there a way we could perhaps keep both implementations in 1.9, making the new one an "opt in" via a feature flag, and once we feel comfortable (say, 1.10), switch to the new one only?

RyanLettieri:
Cert: https://github.com/dapr/components-contrib/actions/runs/3092835163
Conf: https://github.com/dapr/components-contrib/actions/runs/3092835186

berndverst (approved):

berndverst:
In the maintainers chat we agreed we can keep the current v1 designation and don't need to introduce a new version in this case.

berndverst:
Unfortunately this PR removed the user agent so that we lost our ability to track users of Dapr + CosmosDB.

ealsur:
@ItalyPaleAle The TelemetryOptions are common to all Azure SDKs for Go, yeah

berndverst:
I already opened a PR for it. And yes we have done this.

Please do not discuss further here.

Issue #2037, opened 2022-08-31
Query API: Set filters when Unmarshalling Query struct


mcandeia:
## Describe the proposal

The default [query unmarshal](https://github.com/dapr/components-contrib/blob/master/state/query/query.go#L98) doesn't set the Filters field as it does for the others fields - sort and page.

It's not a problem for any of our current state component since filters has only been used for building the `Filter` [interface](https://github.com/dapr/components-contrib/blob/master/state/query/query.go#L45),

However if a component needs to send this data serialized as JSON this would become a problem, which is probably not the case for any of the components made within that repository.

This proposal is just a way to make it easier to serialize filters as JSON for components that need to resend them (i.e. pluggable components). Another option would be to reverse engineer the filters, but given that this modification is small enough to have no problem, I believe it would be easier to set the field.

berndverst:
Feel free to open a PR @mcandeia - just make sure it doesn't break any conformance or certification tests. If it does, you may need to update some components' Query API implementation.

PR #2038, opened 2022-08-31
Query API: Set filters when unmarshalling Query Struct

mcandeia:
Signed-off-by: Marcos Candeia <marrcooos@gmail.com>

# Description

Set filters attribute on query struct when unmarshalling it. This will allow components to serialize the filters further.

## Issue reference

Please reference the issue this PR will close: #2037 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/state/query/query.go
+++ b/state/query/query.go
@@ -106,6 +106,14 @@ func (q *Query) UnmarshalJSON(data []byte) error {
 		if err != nil {
 			return err
 		}
+
+		jdata, err := json.Marshal(elem)
+		if err != nil {
+			return err
+		}
+		if err = json.Unmarshal(jdata, &q.Filters); err != nil {
+			return err
+		}
 	}
 	// setting sorting
 	if elem, ok := m[SORT]; ok {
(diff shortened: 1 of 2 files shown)


berndverst:
Complete conformance test run: https://github.com/dapr/components-contrib/actions/runs/2966810571
Complete certification test run: https://github.com/dapr/components-contrib/actions/runs/2966810563

state/query/query.go
```diff
@@ -106,6 +106,14 @@ func (q *Query) UnmarshalJSON(data []byte) error {
 		if err != nil {
 			return err
 		}
+
+		jdata, err := json.Marshal(elem)
```
ItalyPaleAle, line 110: Please use `github.com/mitchellh/mapstructure.Decode` rather than a json Marshal/Unmarshal since this is a fairly hot path.

mcandeia: Tbh I think it could be better than that, I can simply unmarshal as json and right after build filters, I don’t know why it isn’t the actual approach, will try it out

ItalyPaleAle (approved):

berndverst (approved):

PR #2041, opened 2022-09-01
feature(middleware): add routeralias in middleware

1046102779:
Relate Issue: https://github.com/dapr/dapr/pull/5109

For gray servers, if it is an HTTP service, whether in standalone mode or k8s mode, there will be a problem of Dapr router migration, and the cost is very high. There are usually two approaches:

1. Apisix gateway or other gateways, use plugins for routing conversion; Upstream does not need to be modified.

2. Generally, the internal internet may also have HTTP services. This is usually directly exposed through the Service VIP port, and generally there is no internal internet gateway. So upstream must be modified;

Now we go directly to the Dapr runtime level and use middleware to provide a route conversion capability to convert HTTP gray server routes into runtime HTTP standard API routes.

It is necessary to support the unified locator of Restful-style resources. If no suitable route is found in the standardized configuration, the original route will be used directly.

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: routeralias 
spec:
  type: middleware.http.routeralias
  version: v1
  metadata:
  - name: "/v1.0/mall/activity/info"
    value: "/v1.0/invoke/srv.default/method/mall/activity/info"
  - name: "/v1.0/hello/activity/:id/info"
    value: "/v1.0/invoke/srv.default/method/hello/activity/info"
  - name: "/v1.0/hello/activity/:id/user"
    value: "/v1.0/invoke/srv.default/method/hello/activity/user"
```

## Test Cases
```shell
Case 1：(common router)

input：/v1.0/mall/activity/info?id=123

output：/v1.0/invoke/srv.default/method/mall/activity/info?id=123

Case 2：（Restful style router）

input：/v1.0/hello/activity/1/info
output：/v1.0/invoke/srv.default/method/mall/activity/info?id=1

Case 3：（with QueryString）

input：/v1.0/hello/activity/1/user?userid=123
output：/v1.0/invoke/srv.default/method/hello/activity/user?id=1&userid=123

Case 4：（miss router and return original router）

input：/v1.0/invoke/srv.default
output：/v1.0/invoke/srv.default
```

/cc @ItalyPaleAle @berndverst @artursouza @yaron2 @daixiang0


ItalyPaleAle:
> For gray servers, if it is an HTTP service, whether in standalone mode or k8s mode, there will be a problem of Dapr router migration, and the cost is very high.

What is a "gray server" and what is a "Dapr router migration"?

1046102779:
> > For gray servers, if it is an HTTP service, whether in standalone mode or k8s mode, there will be a problem of Dapr router migration, and the cost is very high.
> 
> What is a "gray server" and what is a "Dapr router migration"?

1. `gray server`  and `green server`: existing server and new server.
2. Dapr router migration:  existing http request server routers  need to be converted  Dapr runtime router. Such as: `/v1.o/invoke/${appid}.${namespace}/method/${method}` and etc.

1046102779:
For example:

If there is an old HTTP server, the developer wants to use dapr sidecar in standalone mode. Using `middleware.http.routeralias` component, the developer can seamlessly switch from the old server to the new server (dapr runtime in standalone mode) without loss of traffic. And the upstream's host and route do not need any changes.

an old server(http port: 8080): 
> http://127.0.0.1:8080/mall/activity/info?a=1

a new server(dapr sidecar http port: 8080)
> http://127.0.0.1:8080/mall/activity/info?a=1

CASE 1(common router):
```shell
curl -XGET "http://127.0.0.1:3500/v1.0/mall/activity/info?a=1"

Response：
{"data":"a=1","err_code":0,"err_msg":"success"}
```

CASE 2(restful router):
```shell
curl -i -XGET "http://127.0.0.1:3500/v1.0/rest/1046/info?a=1"

Response：
{"data":"a=1&id=1046","err_code":0,"err_msg":"success"}
```

1046102779:
routeralias.yaml:
```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: routeralias
spec:
  type: middleware.http.routeralias
  version: v1
  metadata:
  - name: "/v1.0/mall/activity/info"
    value: "/v1.0/invoke/server/method/mall/activity/info"
  - name: "/v1.0/rest/:id/info"
    value: "/v1.0/invoke/server/method/mall/activity/info"
```

configuration.yaml:
```yaml
apiVersion: dapr.io/v1alpha1
kind: Configuration
metadata:
  name: appconfig
spec:
  httpPipeline:
    handlers:
    - name: routeralias
      type: middleware.http.routeralias
```

middleware/http/routeralias/routeralias_middleware.go
```diff
@@ -0,0 +1,82 @@
...
+	func(h fasthttp.RequestHandler) fasthttp.RequestHandler, error,
+) {
+	if err := m.getNativeMetadata(metadata); err != nil {
+		return nil, err
+	}
+	return func(h fasthttp.RequestHandler) fasthttp.RequestHandler {
+		return func(ctx *fasthttp.RequestCtx) {
+			uri := ctx.Request.URI()
+
+			handle, params, _ := m.router.Lookup("", string(uri.Path()))
```
daixiang0, line 50: Please handle error.

middleware/http/routeralias/routeralias_middleware.go
```diff
@@ -0,0 +1,82 @@
...
+			if handle != nil {
+				req := &http.Request{}
+				req.URL = new(url.URL)
```
daixiang0, line 53: Could we make those 3 line into one code block about request assign?

middleware/http/routeralias/routeralias_middleware.go
1046102779: The last parameter returned `_` is bool type. 

```golang
// Lookup allows the manual lookup of a method + path combo.
// This is e.g. useful to build a framework around this router.
// If the path was found, it returns the handle function and the path parameter
// values. Otherwise the third return value indicates whether a redirection to
// the same path with an extra / without the trailing slash sh bool e performed.
func (r *Router) Lookup(method, path string) (Handle, Params, bool) {
        if root := r.trees[method]; root != nil {
                return root.getValue(path)
        }
        return nil, nil, false
}
```

middleware/http/routeralias/routeralias_middleware.go
1046102779: Fixed.

1046102779:
I saw that components-contrib changed fasthttp to net/http official library, so routeralias also changed.

ItalyPaleAle (changes requested):
I think this PR overall is going in a good direction, but I would ask that you use a different routing library than httprouter.

httprouter has some issues with matching partial paths, that are best explained in this issue in the go-gin repo: https://github.com/gin-gonic/gin/issues/2016 Eventually, Gin migrated off httprouter

1046102779:
> 

I read this issue [gin#2016](https://github.com/gin-gonic/gin/issues/2016), `httprouter` has conflicts and problems with some route matching, I'll try if I can migrate to gin.

I used the `httprouter` repo before because it was purely a router repo and was cited a lot, but gin is an http server repo, and it seems that its router tree has not been opened. I will take a closer look at how to apply the gin router.  

Thanks @ItalyPaleAle .

ItalyPaleAle:
I wouldn't migrate to gin because that's a full-featured framework. The reason I linked to that is because Gin used to use fasthttp and so many people (including me :) ) were having issues due to its unnatural way of supporting nested routes.

There may be other router packages out there. Or you could write one yourself, if you feel like it.

1046102779:
> 

Got it!

1046102779:
> I wouldn't migrate to gin because that's a full-featured framework. The reason I linked to that is because Gin used to use fasthttp and so many people (including me :) ) were having issues due to its unnatural way of supporting nested routes.
> 
> There may be other router packages out there. Or you could write one yourself, if you feel like it.

Finally I selected `gorilla/mux` to build a router tree.

reference: https://www.alexedwards.net/blog/which-go-router-should-i-use

Test [issue 2016](https://github.com/gin-gonic/gin/issues/2016), the result is ok.

```golang
func Route5() {
        router := mux.NewRouter()
        router.Handle("/:something/", nil)
        router.Handle("/:something/else", nil)
        router.Handle("/:something/:different/:again", nil)
        router.Handle("/:b/:a", nil)
        router.Handle("/:c/:b/:a", nil)
        router.Handle("/:d/:c/:b/:a", nil)
        router.Handle("/v1.0/id/:id", nil)
        router.Handle("/v1.0/name/:name/inspect", nil)
        router.Handle("/v1.0/id/:id/nodes", nil)
}

func main() {
        Route5()
}
```

result: **PASS**

daixiang0 (approved):

ItalyPaleAle (approved):

Issue #2045, opened 2022-09-02
Support for Distribute Transaction with Service Invoke


RcXu:
## Describe the proposal
A distributed transaction manager building block is implemented based on the service invoke API. This building block provides 4 API :
beign: open a transaction; 
commit: Submit a transaction; 
rollback: rollback the transaction; 
getState: query the execution status of the current transaction. 
And distributed transaction building block can also implement a variety of distributed transaction scheme in Dapr runtime, such as TCC, saga, etc.

The transaction component determines whether the current request is a transaction by the calling properties passed in the Service Invoke API request header, and records the header, target method, and body of the current request into persistent DB like redis, mysql, etc. Such as in TCC scheme,  when the initiator of the transaction needs to commit or rollback a transaction, Dapr runtime can use these parameters to replay the request to the Confirm or Rollback interface to implement the transaction logic within the application.

## YAML file
`
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: redis
spec:
  type: transaction.redis
  version: v1
  metadata:
  - name: scheme
    value: tcc
  - name: redisHost
    value: localhost:6379
  - name: redisPassword
    value: ""
  - name: ttlInSeconds
    value: 300
`

## Demo of Distribute transaction api

1. Begin a transaction, and store the bunch transaction state in redis. 'bunchTransactionNum' means it contain two sub transactions
`curl -X POST -d '{"bunchTransactionNum":2}' http://localhost:3500/v1.0/transaction/redis/begin`
and will get following response which will be used in the transaction schedule:
`{"transactionID":"transaction-6122d58f-27fd-4bcf-933a-5582091f6a2d-5805","bunchTransactionIDs":["bunch-1","bunch-2"]}`

2.  Service Invoke with transaction3. 
`curl -X POST  -H "Content-Type: application/json" -H "distribute-transaction-id: transaction-6122d58f-27fd-4bcf-933a-5582091f6a2d-5805" -H "distribute-bunch-transaction-id: bunch-1" -H "distribute-transaction-store: redis" -d '{"say":"hello"}' http://localhost:3500/v1.0/invoke/transactionService/method/test`
then Dapr will store the request state and query params into the reids.

3. query the state of transaction
`curl http://localhost:3500/v1.0/transaction/redis/getState/transaction-6122d58f-27fd-4bcf-933a-5582091f6a2d-5805`

4. commit or rollback the transaction
`
[...]

PR #2046, opened 2022-09-02
Support for Distribute Transaction with Service Invoke

RcXu:
# Description

A distributed transaction manager building block is implemented based on the service invoke API. This building block provides 4 API :
beign: open a transaction; 
commit: Submit a transaction; 
rollback: rollback the transaction; 
getState: query the execution status of the current transaction. 
And distributed transaction building block can also implement a variety of distributed transaction scheme in Dapr runtime, such as TCC, saga, etc.

The transaction component determines whether the current request is a transaction by the calling properties passed in the Service Invoke API request header, and records the header, target method, and body of the current request into persistent DB like redis, mysql, etc. Such as in TCC scheme,  when the initiator of the transaction needs to commit or rollback a transaction, Dapr runtime can use these parameters to replay the request to the Confirm or Rollback interface to implement the transaction logic within the application.

## Issue reference

Please reference the issue this PR will close: #[_[issue number]_](https://github.com/dapr/components-contrib/issues/2045)


ItalyPaleAle (changes requested):
Hi @RcXu thanks for this contribution!

Looks like this is not just a new component, but an entirely new building block.

Work items like these, which introduce new APIs, need to be discussed in dapr/dapr first, where you can put a proposed design for your APIs, explain the use cases, etc. We will not be able to just merge this without that discussion first and approval from the maintainers of dapr/dapr.

Also, could this be related to the work that's being done for Dapr Workflows? https://github.com/dapr/dapr/issues/4576

And this seems possibly related too: https://github.com/dapr/dapr/issues/4939

PR #2055, opened 2022-09-06
Cert tests: Use `docker compose` instead of `docker-compose`

berndverst:

# Description

Cert tests: Use `docker compose` instead of `docker-compose`

### DIFF
--- a/tests/certification/flow/dockercompose/dockercompose.go
+++ b/tests/certification/flow/dockercompose/dockercompose.go
@@ -49,7 +49,8 @@ func Up(project, filename string) flow.Runnable {
 
 func (c Compose) Up(ctx flow.Context) error {
 	out, err := exec.Command(
-		"docker compose",
+		"docker"
+		"compose",
 		"-p", c.project,
 		"-f", c.filename,
 		"up", "-d",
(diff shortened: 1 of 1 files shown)


ItalyPaleAle:
Is there a reason for this PR? `docker-compose` can usually be used as alias for `docker compose`

berndverst:
> Is there a reason for this PR? `docker-compose` can usually be used as alias for `docker compose`

This is easier for folks to run certification tests if they only have the Docker CLI. Also it is my understanding that `docker-compose` is discouraged and `docker compose` should be used.

daixiang0:
> > Is there a reason for this PR? `docker-compose` can usually be used as alias for `docker compose`
> 
> This is easier for folks to run certification tests if they only have the Docker CLI. Also it is my understanding that `docker-compose` is discouraged and `docker compose` should be used.

```

root:[~]$ docker compose
docker: 'compose' is not a docker command.
See 'docker --help'
root:[~]$ docker-compose
Define and run multi-container applications with Docker.

Usage:
  docker-compose [-f <arg>...] [--profile <name>...] [options] [--] [COMMAND] [ARGS...]
  docker-compose -h|--help
```

If apply this change, we should document the supported docker version.

PR #2058, opened 2022-09-07
Add confluent kafka to conformance test

addjuarez:
Signed-off-by: Addison Juarez <adjuarez@microsoft.com>

# Description

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


yaron2:
As part of our conformance test criteria, we should test with auth enabled also. When this is done, it will reveal the fact that Dapr today does not work with Confluent Kafka when TLS is enabled. cc @berndverst

berndverst:
@yaron2 I think confluent kafka should be supported via pluggable components and then it can use the confluent Kafka SDKs -- the Go SDK itself requires CGO but other options exist. This is also a requirement that Alibaba has for their managed Kafka service.

The auth enabled option belongs into certification tests IMO, not the conformance test.

artursouza (approved):

berndverst:
@artursouza FYI - this removed the `pubsub.redis` conformance test which is why all open PRs are currently stalled.

Please merge https://github.com/dapr/components-contrib/pull/2080 to fix this.

PR #2024, opened 2022-08-27
Introduce Basic interfaces and skeleton for Bulk PubSub

DeepanshuA:
# Description

Introduces basic interfaces and skeleton for Batch PubSub.

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

related to: https://github.com/dapr/dapr/issues/2218

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #2059, opened 2022-09-08
Kafka bulk publish

mukundansundar:
# Description

Added Bulk publish implementation for Kafka.

To be merged after #2024 

## Issue reference

related to https://github.com/dapr/dapr/issues/2218

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


mukundansundar:
Will be updating this PR to completely ignore the traceID field as it is not part of cloudEvent spec https://github.com/cloudevents/spec/blob/v1.0/extensions/distributed-tracing.md and also there is a TODO to delete traceid and keep traceparent alone ... https://github.com/dapr/components-contrib/blob/master/pubsub/envelope.go#L39

mukundansundar:
> > Will be updating this PR to completely ignore the traceID field as it is not part of cloudEvent spec https://github.com/cloudevents/spec/blob/v1.0/extensions/distributed-tracing.md and also there is a TODO to delete traceid and keep traceparent alone ... https://github.com/dapr/components-contrib/blob/master/pubsub/envelope.go#L39
> 
> How will users trace calls then?

I believe it will be through the trace parent header itself. It contains trace id and parent id in it ... 
See https://w3c.github.io/trace-context/#traceparent-header

mukundansundar:
Based on my understanding, 
`traceParent` header consists of the `traceID` and `parentID` fields apart from the version and sample details. 
The `traceID` is `ID of the whole trace forest and is used to uniquely identify a [distributed trace](https://www.w3.org/TR/trace-context/#dfn-distributed-traces) through a system` from [doc](https://www.w3.org/TR/trace-context/#trace-id). 
The `parentID` sometimes known as `spanID` is the ` ID of this request as known by the caller (in some tracing systems, this is known as the span-id, where a span is the execution of a client request)`.

So we have two options here for BulkPublish:
1. BulkPublish is supposed to be a single operation so we use the same `traceID`, `parentID` combination across all messages in that operation i.e. the same `traceParent` header.
2. The other option is to generate new child span for each event inside the BulkPublish operation since each message in itself can be consumed very differently and need not be consumed in the same fashion. This way we will also be able to identify individual events in the trace though I am not sure that is what a tracing solution should be used for. In this case, there will be a slight change in the `parentID` or `spanID` for every event but the overall `traceID` should be the same. Need to verify that scenario though. 

I personally thing we need to verify option 2 and go with that. Will try and verify 2 tomorrow morning and post update here.

Meanwhile @yaron2 @artursouza thoughts?

yaron2:
> BulkPublish is seems to be a single operation so we use the same traceID, parentID combination across all messages in that operation i.e. the same traceParent header.

It's a single operation as far as the user is concerned, but that is merely a network optimization. In terms of functionality, each message in the bulk is sent and acked individually, so it should have a separate traceID and parentID for each message.

mukundansundar:
> > BulkPublish is seems to be a single operation so we use the same traceID, parentID combination across all messages in that operation i.e. the same traceParent header.
> 
> It's a single operation as far as the user is concerned, but that is merely a network optimization. In terms of functionality, each message in the bulk is sent and acked individually, so it should have a separate traceID and parentID for each message.

@yaron2 Yes each event will have its own traceParent header ... the impl PR https://github.com/dapr/dapr/pull/5127 does that only. 

Just that acc to the trace context doc a traceParent header contains `version-traceID-parentID-flags` (parentID is actually the operation ID or the spanID)

There is no separate TraceIDField in either CloudEvent or the TraceContext docs ...

mukundansundar:
@dapr/maintainers-components-contrib Please review PR related to bulk publish implementation ...

internal/component/kafka/producer.go
```diff
@@ -74,3 +76,39 @@ func (k *Kafka) Publish(topic string, data []byte, metadata map[string]string) e
 
 	return nil
 }
+
+func (k *Kafka) BulkPublish(topic string, entries []pubsub.BulkMessageEntry, metadata map[string]string) (pubsub.BulkPublishResponse, error) {
+	if k.producer == nil {
+		err := errors.New("component is closed")
+		return pubsub.NewBulkPublishResponse(entries, pubsub.PublishFailed, err), err
+	}
+	k.logger.Debugf("Publishing on topic %v", topic)
```
ItalyPaleAle, line 85: Nit: `Bulk publishing on topic %v`

pubsub/responses.go
```diff
@@ -13,19 +13,86 @@ limitations under the License.
...
+// NewBulkPublishResponse returns a BulkPublishResponse with each entry having same status and error.
+// This method is a helper method to map a single error/success response on BulkPublish to multiple events.
+func NewBulkPublishResponse(messages []BulkMessageEntry, status BulkPublishStatus, err error) BulkPublishResponse {
+	response := BulkPublishResponse{}
+	for i, msg := range messages {
+		st := BulkPublishResponseEntry{}
+		if msg.EntryID != "" {
+			st.EntryID = msg.EntryID
+		} else {
+			msg.EntryID = fmt.Sprintf("%d", i)
```
ItalyPaleAle, line 89: `strconv.Itoa` is much faster especially in hot paths

pubsub/pubsub.go
```diff
@@ -22,16 +22,30 @@ import (
 
 // PubSub is the interface for message buses.
 type PubSub interface {
+	BulkMessager
```
ItalyPaleAle, line 25: As per my comment on https://github.com/dapr/components-contrib/pull/2024#discussion_r968381941 I wonder if this is best implemented as an optional interface. This way we don't force components to inherit the stub object and the runtime can do the check at ... runtime.

pubsub/responses.go
```diff
@@ -13,19 +13,86 @@ limitations under the License.
...
+		}
+		st.Status = status
+		if err != nil {
+			st.Error = err
+		}
+		response.Statuses = append(response.Statuses, st)
```
ItalyPaleAle, line 95: Could this array be pre-allocated, since we know the number of statuses? This way we can skip the `append` and reduce the number of allocations in a hot path.

ItalyPaleAle (changes requested):

pubsub/responses.go
mukundansundar: removed this

pubsub/pubsub.go
mukundansundar: Any decision regarding the struct will be handled as part of #2024 . Want to limit this PR specific to Kafka alone.

mukundansundar: This is done ...

pubsub/pubsub.go
```diff
@@ -29,9 +29,40 @@ type PubSub interface {
 	Close() error
 }
 
+// BulkPublisher is the interface that wraps the BulkPublish method.
+
+// BulkPublish publishes a collection of entries/messages in a BulkPublishRequest to a
+// message bus topic and returns a BulkPublishResponse with individual statuses for each message.
+type BulkPublisher interface {
+	BulkPublish(req *BulkPublishRequest) (BulkPublishResponse, error)
```
artursouza, line 37: ctx, please.

artursouza (approved):

pubsub/pubsub.go
mukundansundar: Will add it in a follow up pr

PR #2063, opened 2022-09-09
Support for Postgresql as configuration store

akhilac1:
# Description

Added Support for Postgres as configuration store

## Issue reference

related to #[4551](https://github.com/dapr/dapr/issues/4551)

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#[2779](https://github.com/dapr/docs/issues/2779)


configuration/postgres/postgres.go
```diff
@@ -0,0 +1,296 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+	"reflect"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/dapr/components-contrib/configuration"
+	"github.com/dapr/kit/logger"
```
ItalyPaleAle, line 28: Please `gofumpt` the files.

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,296 @@
...
+	"github.com/google/uuid"
+	"github.com/jackc/pgconn"
+	"github.com/jackc/pgx/v4/pgxpool"
```
ItalyPaleAle, line 31: pgx v5 is about to come out (currently it's in a RC-like state)... any chance you could update to that one right away? we'll have to upgrade eventually anyways.

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,296 @@
...
+		return fmt.Errorf(ErrorAlreadyInitialized)
+	}
+	if m, err := parseMetadata(metadata); err != nil {
+		p.logger.Error(err)
+		return err
+	} else {
+		p.metadata = m
+	}
+
+	ctx := context.Background()
```
ItalyPaleAle, line 73: Please use a timeout with the context

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+func (p *ConfigurationStore) doSubscribe(ctx context.Context, req *configuration.SubscribeRequest, handler configuration.UpdateHandler, channel string, id string, stop chan struct{}) {
+	conn, err := p.client.Acquire(ctx)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "Error acquiring connection:", err)
+	}
+	defer conn.Release()
+
+	_, err = conn.Exec(context.Background(), channel)
```
ItalyPaleAle, line 155: Please use a context with timeout (and a child of `ctx`) here too. With pgx, this context is just for the initial listen query I believe, so it should have a timeout.

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,296 @@
...
+	if err != nil {
+		p.logger.Errorf("Error listening to channel:", err)
+		return
+	}
+
+	for {
+		notification, err := conn.Conn().WaitForNotification(ctx)
+		if err != nil {
+			p.logger.Errorf("Error waiting for notification:", err)
```
ItalyPaleAle, line 164: Please don't log an error if `err` is "Context canceled" or something like that. Only log errors for actual errors. We've seen users getting confused with things like these in the logs.

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,296 @@
...
+			return
+		}
+		p.handleSubscribedChange(ctx, handler, notification, id)
+	}
+}
+
+func (p *ConfigurationStore) handleSubscribedChange(ctx context.Context, handler configuration.UpdateHandler, msg *pgconn.Notification, id string) {
+	defer func() {
+		if err := recover(); err != nil {
+			p.logger.Errorf("panic in handleSubscribedChange(）method and recovered: %s", err)
```
ItalyPaleAle, line 174: Note there's a non-ASCII character here that VS Code is highlighting.

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,296 @@
...
+		}
+	}()
+	payload := make(map[string]interface{})
+	err := json.Unmarshal([]byte(msg.Payload), &payload)
+	if err != nil {
+		p.logger.Errorf("Error in UnMarshall: ", err)
```
ItalyPaleAle, line 180: Typo: "Unmarshal"

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+	return m, nil
+}
+
+func Connect(ctx context.Context, conn string) (*pgxpool.Pool, error) {
+	pool, err := pgxpool.Connect(ctx, conn)
+	if err != nil {
+		return nil, fmt.Errorf("postgres configuration store connection error : %s", err)
+	}
+	pingErr := pool.Ping(context.Background())
```
ItalyPaleAle, line 254: Should probably use a context that is child of `ctx` and has a timeout

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,296 @@
...
+	m := metadata{}
+
+	if val, ok := cmetadata.Properties[connectionStringKey]; ok && val != "" {
+		m.connectionString = val
+	} else {
+		return m, fmt.Errorf(ErrorMissingConnectionString)
+	}
+
+	if tbl, ok := cmetadata.Properties[configtablekey]; ok && tbl != "" {
+		m.configTable = tbl
```
ItalyPaleAle, line 233: Should probably sanitize the table name to prevent SQL injections such as below, where the table name is inserted in the query as-is. See example of MySQL: https://github.com/dapr/components-contrib/blob/2b5650c2faa41f0a8d82862583497103776be49f/state/mysql/mysql.go#L624

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+func buildQuery(req *configuration.GetRequest, configTable string) (string, error) {
+	var query string
+	if len(req.Keys) == 0 {
+		query = "SELECT * FROM " + configTable
+	} else {
+		var queryBuilder strings.Builder
+		queryBuilder.WriteString("SELECT * FROM " + configTable + " WHERE KEY IN ('")
+		queryBuilder.WriteString(strings.Join(req.Keys, "','"))
```
ItalyPaleAle, line 268: This is a vector for SQL injection, since `req.Keys` comes from the user. Please use prepared statements here.

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,296 @@
...
+		queryBuilder.WriteString("')")
+		query = queryBuilder.String()
+	}
+
+	if len(req.Metadata) > 0 {
+		var s strings.Builder
+		i, j := len(req.Metadata), 0
+		s.WriteString(" AND ")
+		for k, v := range req.Metadata {
+			temp := k + "='" + v + "'"
```
ItalyPaleAle, line 278: Here too, possible SQL injection

ItalyPaleAle (changes requested):
Thanks for this! Can you please take a look at my comments? Most are nits, but the `buildQuery` method is vulnerable to SQL injections and we should use prepared statements there.

configuration/postgres/postgres.go
akhilac1: Added context with timeout

configuration/postgres/postgres.go
akhilac1: Added context with timeout

configuration/postgres/postgres.go
akhilac1: Add exclusion for error message when there is a timeout or context is cancelled

configuration/postgres/postgres.go
akhilac1: Fixed

configuration/postgres/postgres.go
akhilac1: Added check to disallow non-ASCII chars

configuration/postgres/postgres.go
akhilac1: Fixed

configuration/postgres/postgres.go
akhilac1: Agree. Since Postgres configuration will be in Alpha will move to pgx v5

configuration/postgres/postgres.go
akhilac1: Ping acquires a connection from the Pool and executes an empty sql statement against it. If the sql returns without error, the database Ping is considered successful, otherwise, the error is returned. - https://pkg.go.dev/github.com/jackc/pgx/v4@v4.17.0/pgxpool#Pool.Ping

ItalyPaleAle: @akhilac1 was this comment meant in response to something else?

akhilac1: @ItalyPaleAle  - I saw the SQL injection comment for Ping which is not using any attributes. Hence shared the reference FYI. I see that it is for querybuilder. Will push the fix. Thanks

akhilac1:
@ItalyPaleAle added non-alphanumeric character check and prepared statements to buildQuery

ItalyPaleAle:
What is the statement to create the table, by the way?

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,401 @@
...
+		return err
+	}
+	p.client = client
+	pingErr := p.client.Ping(ctx)
+	if pingErr != nil {
+		return pingErr
+	}
+	// check if table exists
+	exists := false
+	err = p.client.QueryRow(ctx, QueryTableExists, p.metadata.configTable).Scan(&exists)
```
ItalyPaleAle, line 99: Two issues with this:

1. This doesn't check the value of `exists`
2. This should probably return a more descriptive error if the table doesn't exist. In particular, if the component can't automatically create the table, it should also link to the Dapr docs with instructions to create the table

configuration/postgres/metadata.go
```diff
@@ -0,0 +1,22 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package postgres
+
+import "time"
+
+type metadata struct {
+	maxIdleTime      time.Duration
```
ItalyPaleAle, line 19: Would recommend a more descriptive name. For example "timeout" or something like that

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,401 @@
...
+			items[key] = &item
+		}
+	}
+	return &configuration.GetResponse{
+		Items: items,
+	}, nil
+}
+
+func (p *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.SubscribeRequest, handler configuration.UpdateHandler) (string, error) {
+	var triggers []string
```
ItalyPaleAle, line 148: Should this method have a lock inside to prevent multiple concurrent operations? I see that there are things that are not fully safe for concurrent use, such as `isSubscriptionActive`.

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,401 @@
...
+			trigger: trigger,
+			keys:    req.Keys,
+		}
+		go p.doSubscribe(ctx, req, handler, key, trigger, subscribeID, stop)
+	}
+	return subscribeID, nil
+}
+
+func (p *ConfigurationStore) Unsubscribe(ctx context.Context, req *configuration.UnsubscribeRequest) error {
+	if oldStopChan, ok := p.subscribeStopChanMap.Load(req.ID); ok {
```
ItalyPaleAle, line 181: Probably needs a lock here too, to prevent concurrent Subscribe/Unsubscribe calls

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,401 @@
...
+		}
+		stop := make(chan struct{})
+		subscribeID = uuid.New().String()
+		p.subscribeStopChanMap.Store(subscribeID, stop)
+		p.ActiveSubscriptions[trigger] = &subscription{
+			uuid:    subscribeID,
+			trigger: trigger,
+			keys:    req.Keys,
+		}
+		go p.doSubscribe(ctx, req, handler, key, trigger, subscribeID, stop)
```
ItalyPaleAle, line 175: This method gets a new connection from the pool for each key. Would be better to have a single connection and multiple listeners attached to that? Otherwise, with N keys, you'd need N connections.

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,401 @@
...
+	for _, trigger := range triggers {
+		key := "listen " + trigger
+		if sub, isActive := p.isSubscriptionActive(req); isActive {
+			if oldStopChan, ok := p.subscribeStopChanMap.Load(sub); ok {
+				close(oldStopChan.(chan struct{}))
+			}
+		}
+		stop := make(chan struct{})
+		subscribeID = uuid.New().String()
+		p.subscribeStopChanMap.Store(subscribeID, stop)
```
ItalyPaleAle, line 169: I see that here you're using a sync.Map. However, because there are other things in this method that are not safe for concurrent use, I'd recommend using a regular map here, and wrapping the entire Subscribe/Unsubscribe methods in a lock.

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,401 @@
...
+					s.WriteString(" AND ")
+				}
+			}
+			query += s.String()
+		}
+	}
+	return query, params, nil
+}
+
+func QueryRow(ctx context.Context, p *pgxpool.Pool, query string, tbl string) error {
```
ItalyPaleAle, line 358: This method doesn't seem to be used, unless I'm missing something

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,401 @@
...
+	}
+	return false
+}
+
+func validateInput(keys []string) error {
+	if len(keys) == 0 {
+		return nil
+	}
+	for _, key := range keys {
+		if !regexp.MustCompile(`^[a-zA-Z0-9]*$`).MatchString(key) {
```
ItalyPaleAle, line 396: This is a hot path, so consider storing the compiled regex in a variable outside of this method. No point in re-compiling it every time.

ItalyPaleAle (changes requested):

configuration/postgres/postgres.go
akhilac1: Scan raises an exception "ErrNoRows" if it does not find any rows with the error message as "no rows in result set". The documentation in Daprdocs includes table creation PR - https://github.com/dapr/docs/issues/2779

configuration/postgres/postgres.go
akhilac1: Checked on v5 and it is still in beta.So proceeding with v4

configuration/postgres/postgres.go
akhilac1: The key here is not the individual key, but the trigger. I have renamed this variable to make it obvious - https://github.com/dapr/components-contrib/blob/4f64eefae64507ca5cfd5f11044283816b126d4a/configuration/postgres/postgres.go#L167

One channel is created per trigger and when the same trigger is resubscribed to, the existing channel is closed and a new channel is created.

configuration/postgres/postgres.go
akhilac1: https://github.com/dapr/components-contrib/blob/4f64eefae64507ca5cfd5f11044283816b126d4a/configuration/postgres/postgres.go#L188

configuration/postgres/postgres.go
akhilac1: https://github.com/dapr/components-contrib/blob/4f64eefae64507ca5cfd5f11044283816b126d4a/configuration/postgres/postgres.go#L188

configuration/postgres/postgres.go
akhilac1: https://github.com/dapr/components-contrib/blob/4f64eefae64507ca5cfd5f11044283816b126d4a/configuration/postgres/postgres.go#L153

akhilac1:
@ItalyPaleAle - configuration API does not support creation on config table/triggers. So these are external pre-requisites until Admin API is supported. The steps to setup postgres as configuration store can be referenced in https://github.com/dapr/docs/pull/2800. 

Regarding Prepared statement - We donot pre-define the prepared statements as Users can query on any number of keys and/or metadata attributes. A sample generated query to GET configuration data based on Key and Metadata would be -  "SELECT * FROM cfgtbl WHERE KEY IN ($1) AND $2 = $3".  

pgxPool.Pool.Query builds a prepared statement and caches it. Every time the statement that is already cached is used, pgx uses the prepared statement from the cache.

configuration/postgres/postgres.go
ItalyPaleAle: Yes, my concern is with the usage of pgx connections. Right now, `doSubscribe` gets a new connection from the pool for each key you're subscribed to. If you subscribe to 10 connections, that requires 10 active connections. Most databases have a limit on # of active connections, so you'd be exhausting them quickly. A single connection should allow `listen`'ing to multiple keys

configuration/postgres/postgres.go
ItalyPaleAle: Correct. Please detect `ErrNoRows` and show a more user-friendly error message, with a link to the Dapr docs

configuration/postgres/postgres.go
```diff
@@ -39,6 +39,7 @@ type ConfigurationStore struct {
 	metadata             metadata
 	client               *pgxpool.Pool
 	logger               logger.Logger
+	configLock           sync.Mutex
 	subscribeStopChanMap sync.Map
```
ItalyPaleAle, line 43: If you use a lock you can now avoid using a `sync.Map` here

ItalyPaleAle (changes requested):

akhilac1:
@ItalyPaleAle  - closed comments. FYI Please

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,415 @@
...
+	if oldStopChan, ok := p.subscribeStopChanMap[req.ID]; ok {
+		delete(p.subscribeStopChanMap, req.ID)
+		close(oldStopChan.(chan struct{}))
+		for k, v := range p.ActiveSubscriptions {
+			if v.uuid == req.ID {
+				channel := "unlisten " + v.trigger
+				conn, err := p.client.Acquire(ctx)
+				if err != nil {
+					return fmt.Errorf("Error acquiring connection: %v", err)
+				}
```
mukundansundar, line 203: Small case

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,415 @@
...
+}
+
+func (p *ConfigurationStore) Unsubscribe(ctx context.Context, req *configuration.UnsubscribeRequest) error {
+	p.configLock.Lock()
+	defer p.configLock.Unlock()
+	if oldStopChan, ok := p.subscribeStopChanMap[req.ID]; ok {
+		delete(p.subscribeStopChanMap, req.ID)
+		close(oldStopChan.(chan struct{}))
+		for k, v := range p.ActiveSubscriptions {
+			if v.uuid == req.ID {
```
mukundansundar, line 198: Small case.. In general all errors within go are returned with small case starting letters

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,415 @@
...
+	}
+	p.configLock.Unlock()
+	return subscribeID, nil
+}
+
+func (p *ConfigurationStore) Unsubscribe(ctx context.Context, req *configuration.UnsubscribeRequest) error {
+	p.configLock.Lock()
+	defer p.configLock.Unlock()
+	if oldStopChan, ok := p.subscribeStopChanMap[req.ID]; ok {
+		delete(p.subscribeStopChanMap, req.ID)
```
mukundansundar, line 195: Qq should both listen and unlisten be constant templates with sprintf being used here?

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,415 @@
...
+		close(oldStopChan.(chan struct{}))
+		for k, v := range p.ActiveSubscriptions {
+			if v.uuid == req.ID {
+				channel := "unlisten " + v.trigger
+				conn, err := p.client.Acquire(ctx)
+				if err != nil {
+					return fmt.Errorf("Error acquiring connection: %v", err)
+				}
+				defer conn.Release()
+				_, err = conn.Exec(ctx, channel)
```
mukundansundar, line 205: Return after this line ? Or is it possible to have multiple subs with same id?

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,415 @@
...
+					return fmt.Errorf("Error unlistening to channel: %v", err)
+				}
+				delete(p.ActiveSubscriptions, k)
+			}
+		}
+		return nil
+	}
+	return fmt.Errorf("unable to find subscription with ID : %v", req.ID)
+}
+
```
mukundansundar, line 216: Small case https://github.com/golang/go/wiki/Errors

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,415 @@
...
+		p.logger.Errorf("Error acquiring connection:", err)
+		return
+	}
+	defer conn.Release()
+	_, err = conn.Exec(ctx, channel)
+	if err != nil {
+		p.logger.Errorf("Error listening to channel:", err)
+		return
+	}
+
```
mukundansundar, line 229: Cleaner to split ! Across both conditions and inverse the or to and?

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,415 @@
...
+					return
+				}
+			case "value":
+				value = v.Interface().(string)
+			case "version":
+				version = v.Interface().(string)
+			case "metadata":
+				a := v.Interface().(map[string]interface{})
+				for k, v := range a {
+					m[k] = v.(string)
```
mukundansundar, line 277: Should there be an else case here for error ?

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,415 @@
...
+	}()
+	payload := make(map[string]interface{})
+	err := json.Unmarshal([]byte(msg.Payload), &payload)
+	if err != nil {
+		p.logger.Errorf("Error in UnMarshal: ", err)
+		return
+	}
+
+	var key, value, version string
+	m := make(map[string]string)
```
mukundansundar, line 256: Is there a possibility that multiple values are returned for the same key and if so should we arrange by version descending and return latest?

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,415 @@
...
+		if len(tbl) > maxIdentifierLength {
+			return m, fmt.Errorf(ErrorTooLongFieldLength+" - tableName : '%v'. max allowed field length is %v ", tbl, maxIdentifierLength)
+		}
+		m.configTable = tbl
+	} else {
+		return m, fmt.Errorf(ErrorMissingTableName)
+	}
+
+	// configure maxTimeout if provided
+	if mxTimeout, ok := cmetadata.Properties[connMaxIdleTimeKey]; ok && mxTimeout != "" {
```
mukundansundar, line 320: Should a default  value be used instead of returning error?

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,415 @@
...
+				return sub.uuid, true
+			}
+		}
+	}
+	return " ", false
+}
+
+func (p *ConfigurationStore) isSubscribed(trigger string, key string, subscription string) bool {
+	for t, s := range p.ActiveSubscriptions {
+		if t == trigger && s.uuid == subscription {
```
mukundansundar, line 394: Replace with slices.Contains? https://pkg.go.dev/golang.org/x/exp/slices#Contains

mukundansundar: @berndverst @ItalyPaleAle any issue with using this method?

configuration/postgres/postgres_test.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	query, params, err := buildQuery(g, "cfgtbl")
+	_ = params
+	if err != nil {
+		t.Errorf("Error building query: %v ", err)
+	}
+	expected := "SELECT * FROM cfgtbl WHERE KEY IN ($1) AND $2 = $3"
+	if query != expected {
+		t.Errorf("Did not get expected result. Got: '%v' , Expected: '%v'", query, expected)
```
mukundansundar, line 41: Small case

configuration/postgres/postgres_test.go
```diff
@@ -0,0 +1,94 @@
...
+	}
```
mukundansundar, line 42: assert.Equals(t, expected, query, "message")

configuration/postgres/postgres_test.go
```diff
@@ -0,0 +1,94 @@
...
+	i := 0
+	for _, v := range params {
+		got := v.(string)
+		switch i {
+		case 0:
+			expected := "someKey"
+			if expected != got {
+				t.Errorf("Did not get expected result. Got: '%v' , Expected: '%v'", got, expected)
+			}
```
mukundansundar, line 51: Use assert

configuration/postgres/postgres_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"context"
+	"regexp"
+	"testing"
+
+	"github.com/pashagolub/pgxmock"
+
+	"github.com/dapr/components-contrib/configuration"
+)
+
+func TestPostgresbuildQuery(t *testing.T) {
```
mukundansundar, line 26: Add test for select * without any conditions

mukundansundar: Will there be any negative cases?

configuration/postgres/postgres_test.go
```diff
@@ -0,0 +1,94 @@
...
+func TestConnectAndQuery(t *testing.T) {
+	m := metadata{
+		connectionString: "mockConnectionString",
+		configTable:      "mockConfigTable",
+	}
+
+	mock, err := pgxmock.NewPool()
+	if err != nil {
+		t.Fatalf("an error '%s' was not expected when opening a stub database connection", err)
+	}
```
mukundansundar, line 76: assert.Nil(t, err, "message")

mukundansundar (changes requested):

configuration/postgres/postgres.go
ItalyPaleAle: I agree with @mukundansundar . This is a preview component so it's fine with using v5 right away. We'll have to upgrade eventually anyways...

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,411 @@
...
+		subscribeStopChanMap: sync.Map{},
+		configLock:           sync.Mutex{},
+	}
+}
+
+func (p *ConfigurationStore) Init(metadata configuration.Metadata) error {
+	p.logger.Debug(InfoStartInit)
+
+	if p.client != nil {
+		return fmt.Errorf(ErrorAlreadyInitialized)
```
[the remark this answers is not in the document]
ItalyPaleAle: I don't think there's any code path where `Init` would be called twice on the same object. But no harm in leaving it here imho.

configuration/postgres/postgres.go
ItalyPaleAle: None from me, we have been using generics already in Dapr.

configuration/postgres/postgres_test.go
akhilac1: this test uses the pgxmock package to test the Query. So we first set the expected query and outcome [WillReturnrows] and then check if expectation is matched after the mock executes the query. Hence assert is not being used here.

configuration/postgres/postgres.go
mukundansundar: The only caveat that I have is do we know eta... Will it be longer than the eta we plan for making this component stable?

ItalyPaleAle: > The only caveat that I have is do we know eta... Will it be longer than the eta we plan for making this component stable?

pgx AFAIK is maintained by volunteers, but it's been in beta for a while already and I doubt it will be much longer.

mukundansundar: Actually released 2days ago.. https://github.com/jackc/pgx/releases/tag/v5.0.0

configuration/postgres/postgres.go
mukundansundar: Not seeing change

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,411 @@
...
+		v := make(map[string]string)
+
+		if err := rows.Scan(&key, &item.Value, &item.Version, &metadata); err != nil {
+			return nil, err
+		}
+		if err := json.Unmarshal(metadata, &v); err != nil {
+			return nil, err
+		}
+		item.Metadata = v
+		if item.Value != "" {
```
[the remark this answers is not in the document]
akhilac1: Insert and update of config values happens outside and hence we check if the value is empty. In case it is empty we donot return this in the response

configuration/postgres/postgres.go
akhilac1: Yes. The challenge is version is a string. Issue https://github.com/dapr/dapr/issues/5183#issuecomment-1251328400 is created to mandate version to follow semantic versioning so that version can be compared. As it cannot be reliable compared today, documentation is updated to ensure single value of key to be maintained in the configuration store.

configuration/postgres/metadata.go
akhilac1: changed to maxIdleTimeout

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,411 @@
+/*
+Copyright 2021 The Dapr Authors
```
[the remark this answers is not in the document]
akhilac1: Updated in all 3 files

configuration/postgres/postgres.go
akhilac1: updated to v5 with few minor updates to connect and waitForNotification

configuration/postgres/postgres.go
akhilac1: Removed sync.Map

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,411 @@
...
+
+const (
+	configtablekey               = "table"
+	connMaxIdleTimeKey           = "connMaxIdleTime"
+	connectionStringKey          = "connectionString"
+	ErrorMissingTableName        = "missing postgreSQL configuration table name"
+	InfoStartInit                = "initializing postgreSQL configuration store"
+	ErrorMissingConnectionString = "missing postgreSQL connection string"
+	ErrorAlreadyInitialized      = "PostgreSQL configuration store already initialized"
+	ErrorMissinMaxTimeout        = "missing PostgreSQL maxTimeout setting in configuration"
```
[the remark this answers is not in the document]
akhilac1: Fixed

configuration/postgres/postgres.go
akhilac1: done

configuration/postgres/postgres.go
akhilac1: If you look at the client attribute, you will see that it is a pgxpool object. Connection can be reused only when it is closed. So we can reuse connection for get operation. But for subscribe we defer the connection close until unsubscribe. Connection gives a connection busy error when reused, since it is still in use. When we donot mention the number of connections, the default connections in pool is 5. User will not be able to subscribe beyond 5 triggers. Since the need is to listen to notifications from one table only, this should not be a blocker. Will add a new maxconnections to input arguments so that user is not limited by the default. This will also be updated in the docs.

configuration/postgres/postgres.go
akhilac1: Done

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,411 @@
...
+	p.client = client
+	pingErr := p.client.Ping(ctx)
+	if pingErr != nil {
+		return pingErr
+	}
+	// check if table exists
+	exists := false
+	err = p.client.QueryRow(ctx, QueryTableExists, p.metadata.configTable).Scan(&exists)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
akhilac1: done

configuration/postgres/postgres.go
akhilac1: done

configuration/postgres/postgres.go
akhilac1: @mukundansundar  - marked this as resolved by mistake. please comment if this needs more attention

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,401 @@
...
+	m := metadata{}
+
+	if val, ok := cmetadata.Properties[connectionStringKey]; ok && val != "" {
+		m.connectionString = val
+	} else {
+		return m, fmt.Errorf(ErrorMissingConnectionString)
+	}
+
+	if tbl, ok := cmetadata.Properties[configtablekey]; ok && tbl != "" {
+		if !regexp.MustCompile(`^[a-zA-Z0-9]*$`).MatchString(tbl) {
```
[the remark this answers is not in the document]
akhilac1: done

configuration/postgres/postgres.go
akhilac1: done

configuration/postgres/postgres.go
akhilac1: done. Used the same value as pgxpool.defaultMaxConnIdleTime

configuration/postgres/postgres.go
akhilac1: done

configuration/postgres/postgres_test.go
akhilac1: Added two scenarios for select * - where no keys are given and where no keys are provided by metadata is provided

configuration/postgres/postgres_test.go
akhilac1: done

configuration/postgres/postgres_test.go
akhilac1: @mukundansundar - assert is not used here. pgxmock functions are used to validate the query.

mukundansundar: The specified 3 lines can still be an assert without touching the pgx mock assertions

configuration/postgres/postgres.go
ItalyPaleAle: Hi @akhilac1 I understand how the pooling works. What I'm saying is that here, you are using one connection for each LISTEN. That seems wasteful, as I believe you should be able to perform multiple LISTEN on a single connection.

Each connection uses quite a bit of resources, both on the client and server (servers too have a connection limit, for example on Azure Postgres it can be as low as 50 on the server).

I think you should have a single, persistent connection that is dedicated to LISTEN'ing, and just add or remove listeners to that connection rather than creating new ones.

configuration/postgres/postgres.go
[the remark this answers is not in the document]
akhilac1: @mukundansundar - can you please respond to this.

configuration/postgres/postgres.go
akhilac1: Done

configuration/postgres/postgres_test.go
akhilac1: done

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,432 @@
...
+			return
+		}
+		p.handleSubscribedChange(ctx, handler, notification, channel, subscription)
+	}
+}
+
+func (p *ConfigurationStore) handleSubscribedChange(ctx context.Context, handler configuration.UpdateHandler, msg *pgconn.Notification, channel string, subscriptionID string) {
+	defer func() {
+		if err := recover(); err != nil {
+			p.logger.Errorf("panic in handlesubscribedchange method and recovered: %s", err)
```
mukundansundar, line 239: Nitpick camelCase for method name...

configuration/postgres/postgres.go
akhilac1: https://github.com/dapr/components-contrib/blob/90ba8a441d6ccc8987acf115ca4adf1f49fa9fba/configuration/postgres/postgres.go#L245

configuration/postgres/postgres.go
akhilac1: https://github.com/dapr/components-contrib/blob/90ba8a441d6ccc8987acf115ca4adf1f49fa9fba/configuration/postgres/postgres.go#L108-L112

configuration/postgres/postgres.go
akhilac1: @ItalyPaleAle 
https://github.com/dapr/components-contrib/blob/90ba8a441d6ccc8987acf115ca4adf1f49fa9fba/configuration/postgres/postgres.go#L130-L135

configuration/postgres/postgres.go
akhilac1: Input args are validated for sql injection scenarios
https://github.com/dapr/components-contrib/blob/90ba8a441d6ccc8987acf115ca4adf1f49fa9fba/configuration/postgres/postgres.go#L126-L131

configuration/postgres/postgres.go
akhilac1: @ItalyPaleAle - one connection allows subscription to one channel - one channel allows listening to notifications for any number of keys. Added guidance in documentation to use one subscription with multiple keys

_When calling subscribe API, metadata.pgNotifyChannel should be used to specify the name of the channel to listen for notifications from postgres configuration store.

Any number of keys can be added to a subscription request. Each subscription uses a exclusive database connectio. So, it is strongly recommended to subscribe to multiple keys within a single subscription. This helps optimize the number of connections to the database._

https://github.com/dapr/docs/blob/357ffcf6dd53d5736461c7e63050da5917aa6bbf/daprdocs/content/en/reference/components-reference/supported-configuration-stores/postgres-configuration-store.md

akhilac1:
> > Regarding Prepared statement - We donot pre-define the prepared statements as Users can query on any number of keys and/or metadata attributes. A sample generated query to GET configuration data based on Key and Metadata would be - "SELECT * FROM cfgtbl WHERE KEY IN ($1) AND $2 = $3".
> > pgxPool.Pool.Query builds a prepared statement and caches it. Every time the statement that is already cached is used, pgx uses the prepared statement from the cache.
> 
> Using `$1` etc is using prepared statements: in this case, the query doesn't contain the exact values, but they are only referenced like you're doing there. Prepared statements are needed to avoid SQL injections

@ItalyPaleAle - We cannot pre define prepared statements as there is no cap on number of keys that can be queried for and/if metadata is provided. So we build the query and depend on the functionality of pgxpool library to convert these to prepared statements. SQL Injection scenarios are addressed via input validation. 

Can you please clarify on where you see SQL Injection possibility?

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,432 @@
...
+const (
+	configtablekey               = "table"
+	connMaxIdleTimeKey           = "connMaxIdleTime"
+	connectionStringKey          = "connectionString"
+	pgNotifyChannelKey           = "pgNotifyChannel"
+	listenTemplate               = "listen %s"
+	unlistenTemplate             = "unlisten %s"
+	payloadDataKey               = "data"
+	ErrorMissingTableName        = "missing postgreSQL configuration table name"
+	ErrorMissingTable            = "postgreSQL configuration table - '%v' does not exist"
```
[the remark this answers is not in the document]
akhilac1: done

configuration/postgres/postgres.go
akhilac1: https://github.com/dapr/components-contrib/blob/90ba8a441d6ccc8987acf115ca4adf1f49fa9fba/configuration/postgres/postgres.go#L102-L104

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,432 @@
...
+
+func (p *ConfigurationStore) Get(ctx context.Context, req *configuration.GetRequest) (*configuration.GetResponse, error) {
+	if err := validateInput(req.Keys); err != nil {
+		p.logger.Error(err)
+		return nil, err
+	}
+	query, params, err := buildQuery(req, p.metadata.configTable)
+	if err != nil {
+		p.logger.Error(err)
+		return nil, fmt.Errorf("error in configuration store query: '%s' ", err)
```
akhilac1, line 133: Added context with timeout

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,432 @@
...
+		if res := strings.EqualFold(pgNotifyChannelKey, k) && !slices.Contains(pgNotifyChannels, v); res {
+			pgNotifyChannels = append(pgNotifyChannels, v) // append unique channel names only
+		}
+	}
+	if len(pgNotifyChannels) == 0 {
+		return "", fmt.Errorf("cannot subscribe to empty channel")
+	}
+	return p.subscribeToChannel(ctx, pgNotifyChannels, req, handler)
+}
+
```
akhilac1, line 177: Added context with timeout to be passed to pgx

configuration/postgres/postgres.go
akhilac1: https://github.com/dapr/components-contrib/blob/90ba8a441d6ccc8987acf115ca4adf1f49fa9fba/configuration/postgres/postgres.go#L225-L231

configuration/postgres/postgres.go
akhilac1: Fixed

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,432 @@
...
+					p.logger.Errorf("error acquiring connection:", err)
+					return fmt.Errorf("error acquiring connection: %s ", err)
+				}
+				defer conn.Release()
+				_, err = conn.Exec(ctx, pgChannel)
+				if err != nil {
+					p.logger.Errorf("error listening to channel:", err)
+					return fmt.Errorf("error listening to channel: %s", err)
+				}
+				delete(p.ActiveSubscriptions, k)
```
akhilac1, line 205: Fixed return after finding the first subscription since only one active subscription can exist for a given channel

configuration/postgres/postgres.go
akhilac1: https://github.com/dapr/components-contrib/blob/90ba8a441d6ccc8987acf115ca4adf1f49fa9fba/configuration/postgres/postgres.go#L225-L231

configuration/postgres/postgres_test.go
akhilac1: https://github.com/dapr/components-contrib/blob/90ba8a441d6ccc8987acf115ca4adf1f49fa9fba/configuration/postgres/postgres_test.go#L27-L47

configuration/postgres/postgres_test.go
akhilac1: https://github.com/dapr/components-contrib/blob/90ba8a441d6ccc8987acf115ca4adf1f49fa9fba/configuration/postgres/postgres_test.go#L98-L106

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,432 @@
...
+	} else {
+		return m, fmt.Errorf(ErrorMissingTableName)
+	}
+	// configure maxTimeout if provided
+	if mxTimeout, ok := cmetadata.Properties[connMaxIdleTimeKey]; ok && mxTimeout != "" {
+		if t, err := time.ParseDuration(mxTimeout); err == nil {
+			m.maxIdleTimeout = t
+		}
+	}
+	if m.maxIdleTimeout == 0 {
```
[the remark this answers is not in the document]
akhilac1: Only if this value is not set aka. it remains as zero we override else any value >0 set by user will be used

configuration/postgres/postgres.go
akhilac1: this method will is not part of interface and not visible outside the package. So it is named starting with a small case letter.

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,432 @@
...
+			return m, fmt.Errorf(ErrorTooLongFieldLength+" - tableName : '%v'. max allowed field length is %v ", tbl, maxIdentifierLength)
+		}
+		m.configTable = tbl
+	} else {
+		return m, fmt.Errorf(ErrorMissingTableName)
+	}
+	// configure maxTimeout if provided
+	if mxTimeout, ok := cmetadata.Properties[connMaxIdleTimeKey]; ok && mxTimeout != "" {
+		if t, err := time.ParseDuration(mxTimeout); err == nil {
+			m.maxIdleTimeout = t
```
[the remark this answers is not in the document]
akhilac1: https://github.com/dapr/components-contrib/blob/b03c77f2d66eebe404b31e148e656e73b7e28e6b/configuration/postgres/postgres.go#L312-L319

configuration/postgres/postgres.go
akhilac1: done
https://github.com/dapr/components-contrib/blob/b03c77f2d66eebe404b31e148e656e73b7e28e6b/configuration/postgres/postgres.go#L200-L206

configuration/postgres/postgres.go
akhilac1: Resolving this @ItalyPaleAle

configuration/postgres/postgres.go
akhilac1: Using the context from the input args and dropped creating a new context as pointed out in the comment -
https://github.com/dapr/components-contrib/blob/b03c77f2d66eebe404b31e148e656e73b7e28e6b/configuration/postgres/postgres.go#L199-L200

akhilac1: @ItalyPaleAle - resolving this comment

akhilac1:
> What is the statement to create the table, by the way?

https://github.com/dapr/docs/blob/357ffcf6dd53d5736461c7e63050da5917aa6bbf/daprdocs/content/en/reference/components-reference/supported-configuration-stores/postgres-configuration-store.md

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,432 @@
...
+
+func (p *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.SubscribeRequest, handler configuration.UpdateHandler) (string, error) {
+	var pgNotifyChannels []string
+	for k, v := range req.Metadata {
+		if res := strings.EqualFold(pgNotifyChannelKey, k) && !slices.Contains(pgNotifyChannels, v); res {
+			pgNotifyChannels = append(pgNotifyChannels, v) // append unique channel names only
+		}
+	}
+	if len(pgNotifyChannels) == 0 {
+		return "", fmt.Errorf("cannot subscribe to empty channel")
```
mukundansundar, line 173:
```suggestion
		return "", fmt.Errorf("cannot subscribe to empty postgres notification channel")
```

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,432 @@
...
+			break
+		} else {
+			return fmt.Errorf("unable to find subscription with ID : %v", req.ID)
+		}
+	}
+	if oldStopChan, ok := p.subscribeStopChanMap[req.ID]; ok {
+		delete(p.subscribeStopChanMap, req.ID)
+		close(oldStopChan.(chan struct{}))
+		for k, v := range p.ActiveSubscriptions {
+			if v.uuid == req.ID {
```
mukundansundar, line 192: Will it be better if the key from loop in line 181 is locally cached and we avoid this loop completely??

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,432 @@
...
+	m := make(map[string]string)
+	// trigger should encapsulate the row in "data" field in the notification
+	row := reflect.ValueOf(payload[payloadDataKey])
+	if row.Kind() == reflect.Map {
+		for _, k := range row.MapKeys() {
+			v := row.MapIndex(k)
+			strKey := k.Interface().(string)
+			switch strings.ToLower(strKey) {
+			case "key":
+				key = v.Interface().(string)
```
mukundansundar, line 258: Can this cast fail at all? Should we check if cast is OK first?
Same for other casts below...

mukundansundar (commented):
Overall lgtm... Just added few more questions....

configuration/postgres/postgres.go
akhilac1: https://github.com/dapr/components-contrib/blob/fff6c5c061e7170baaf4592a9a49af5a8882072c/configuration/postgres/postgres.go#L172-L174

configuration/postgres/postgres.go
akhilac1: Done - 
https://github.com/dapr/components-contrib/blob/fff6c5c061e7170baaf4592a9a49af5a8882072c/configuration/postgres/postgres.go#L181-L200

configuration/postgres/postgres.go
akhilac1: As discussed offline, panic will handle any edge case exceptions raised due to failing string case

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,432 @@
...
+	defaultMaxConnIdleTime = time.Minute * 30
+)
+
+func NewPostgresConfigurationStore(logger logger.Logger) configuration.Store {
+	logger.Debug("Instantiating PostgreSQL configuration store")
+	return &ConfigurationStore{
+		logger:               logger,
+		subscribeStopChanMap: make(map[string]interface{}),
+		configLock:           sync.Mutex{},
+		metadata: metadata{
```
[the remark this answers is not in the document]
akhilac1: https://github.com/dapr/components-contrib/blob/1243970b63d03c78c58f3543b7b7b5ba0f5f1d19/configuration/postgres/postgres.go#L76-L83

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,421 @@
...
+			return fmt.Errorf("invalid key : '%v'", key)
+		}
+	}
+	return nil
+}
+
+func (p *ConfigurationStore) subscribeToChannel(ctx context.Context, pgNotifyChanList []string, req *configuration.SubscribeRequest, handler configuration.UpdateHandler) (string, error) {
+	p.configLock.Lock()
+	defer p.configLock.Unlock()
+	var subscribeID string // TO_DO - duplicate trigger
```
mukundansundar, line 401: is the TODO going to be addressed in this PR or for later? If later please add a comment with issue tracking this ....

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,421 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+func validateInput(keys []string) error {
+	if len(keys) == 0 {
+		return nil
+	}
```
mukundansundar, line 389: not needed ... as the for loop on range will not be run if the length is zero

mukundansundar: Add unit test for the same ....

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,421 @@
...
+	return " ", false
+}
+
+func (p *ConfigurationStore) isSubscribed(subscriptionID string, channel string, key string) bool {
+	if val, yes := p.ActiveSubscriptions[channel]; yes {
+		if val.uuid == subscriptionID && slices.Contains(val.keys, key) {
+			return true
+		}
+	}
```
mukundansundar, line 382:
```suggestion
	if val, yes := p.ActiveSubscriptions[channel]; yes && val.uuid == subscriptionID && slices.Contains(val.keys, key) {
		return true
	}
```

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,421 @@
...
+				}
+			}
+			query += s.String()
+		}
+	}
+	return query, params, nil
+}
+
+func (p *ConfigurationStore) isSubscriptionActive(req *configuration.SubscribeRequest) (string, bool) {
+	for _, channel := range req.Metadata {
```
mukundansundar, line 367: Is `channel` actually channel or can it be other metadata value also ?

akhilac1:
@mukundansundar - can you please check the open clarifications and comment if this PR is good to merge? Please note that these are not code fixes but clarifications only.

@artursouza - FYI Please

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,421 @@
...
+func Connect(ctx context.Context, conn string, maxTimeout time.Duration) (*pgxpool.Pool, error) {
+	config, err := pgxpool.ParseConfig(conn)
+	if err != nil {
+		return nil, fmt.Errorf("postgres configuration store connection error : %s", err)
+	}
+	pool, err := pgxpool.NewWithConfig(ctx, config)
+	if err != nil {
+		return nil, fmt.Errorf("postgres configuration store connection error : %s", err)
+	}
+	pingErr := pool.Ping(ctx)
```
mukundansundar, line 321: probably not right now ... but consider implementing the Pinger interface for this component ....

configuration/postgres/postgres.go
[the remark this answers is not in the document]
mukundansundar: What does negative value imply? 
I think this condition should be 
```suggestion
	if m.maxIdleTimeout <= 0 {
```

configuration/postgres/postgres.go
akhilac1: @mukundansundar - please add this to the issue tracking configstore improvements - https://github.com/dapr/dapr/issues/5183

configuration/postgres/postgres.go
akhilac1: It is channel only. No other metadata key is considered for subscription. We use "pgNotifyChannel" as the key. Ref - [docs/2779](https://github.com/dapr/docs/issues/2779)

configuration/postgres/postgres.go
akhilac1: No pending TO_DO needed here. Removed the comment.

configuration/postgres/postgres.go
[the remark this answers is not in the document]
akhilac1: m.maxIdleTimeout will be set only when t >0. So the negative scenario, which could be due to a typo from the user - will be mitigated and only a non negative timeout will be set.

https://github.com/dapr/components-contrib/blob/2817724483e2504d36fbb4889fc76dc92bc0f878/configuration/postgres/postgres.go#L301-L309

configuration/postgres/postgres.go
akhilac1: https://github.com/dapr/components-contrib/blob/f13fcf1ef336a1bd3821264ee04eec7985710046/configuration/postgres/postgres_test.go#L105-L114

configuration/postgres/postgres.go
akhilac1: https://github.com/dapr/components-contrib/blob/f13fcf1ef336a1bd3821264ee04eec7985710046/configuration/postgres/postgres.go#L377-L379

akhilac1:
@mukundansundar  - closed comments. can you please check

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,417 @@
...
+	if m, err := parseMetadata(metadata); err != nil {
+		p.logger.Error(err)
+		return err
+	} else {
+		p.metadata = m
+	}
+	p.ActiveSubscriptions = make(map[string]*subscription)
+	ctx, cancel := context.WithTimeout(context.Background(), defaultMaxConnIdleTime)
+	defer cancel()
+	// ctx := context.Background()
```
ItalyPaleAle, line 99: Delete this

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,417 @@
...
+	ErrorAlreadyInitialized      = "postgreSQL configuration store already initialized"
+	ErrorMissingMaxTimeout       = "missing PostgreSQL maxTimeout setting in configuration"
+	QueryTableExists             = "SELECT EXISTS (SELECT FROM pg_tables where tablename = $1)"
+	ErrorTooLongFieldLength      = "field name is too long"
+	maxIdentifierLength          = 64 // https://www.postgresql.org/docs/current/limits.html
+)
+
+var (
+	allowedChars           = regexp.MustCompile(`^[a-zA-Z0-9./_]*$`)
+	defaultMaxConnIdleTime = time.Minute * 30
```
ItalyPaleAle, line 73: 30 min as a connection timeout seems a lot? 30 seconds maybe?

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,417 @@
...
+)
+
+func NewPostgresConfigurationStore(logger logger.Logger) configuration.Store {
+	logger.Debug("Instantiating PostgreSQL configuration store")
+	return &ConfigurationStore{
+		logger:               logger,
+		subscribeStopChanMap: make(map[string]interface{}),
+		configLock:           sync.Mutex{},
```
ItalyPaleAle, line 81: Nit: Can be removed

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,417 @@
...
+		return fmt.Errorf("unable to connect to configuration store: '%s'", pingErr)
+	}
+	// check if table exists
+	exists := false
+	err = p.client.QueryRow(ctx, QueryTableExists, p.metadata.configTable).Scan(&exists)
+	if err != nil {
+		if err == sql.ErrNoRows {
+			return fmt.Errorf(ErrorMissingTable, p.metadata.configTable)
+		}
+		return fmt.Errorf("error in checking if configtable exists - '%v'", p.metadata.configTable)
```
ItalyPaleAle, line 116: `return fmt.Errorf("error in checking if configtable '%s' exists: %w", p.metadata.configTable, err)`

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,417 @@
...
+func (p *ConfigurationStore) Get(ctx context.Context, req *configuration.GetRequest) (*configuration.GetResponse, error) {
+	if err := validateInput(req.Keys); err != nil {
+		p.logger.Error(err)
+		return nil, err
+	}
+	query, params, err := buildQuery(req, p.metadata.configTable)
+	if err != nil {
+		p.logger.Error(err)
+		return nil, fmt.Errorf("error in configuration store query: '%s' ", err)
+	}
```
ItalyPaleAle, line 130: With `fmt.Errorf`, when you pass errors, it's best to wrap them using `%w` which allows the error to be un-wrapped. See: https://gosamples.dev/wrap-unwrap-errors/

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,417 @@
...
+	} else {
+		p.metadata = m
+	}
+	p.ActiveSubscriptions = make(map[string]*subscription)
+	ctx, cancel := context.WithTimeout(context.Background(), defaultMaxConnIdleTime)
+	defer cancel()
+	// ctx := context.Background()
+	client, err := Connect(ctx, p.metadata.connectionString)
+	if err != nil {
+		return fmt.Errorf("error connecting to configuration store: '%s'", err)
```
ItalyPaleAle, line 102: Here too, when wrapping an error please use `%w`

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+	p.client = client
+	pingErr := p.client.Ping(ctx)
+	if pingErr != nil {
+		return fmt.Errorf("unable to connect to configuration store: '%s'", pingErr)
```
ItalyPaleAle, line 107: Here too, when wrapping an error please use `%w`

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,417 @@
...
+	p.ActiveSubscriptions = make(map[string]*subscription)
+	ctx, cancel := context.WithTimeout(context.Background(), defaultMaxConnIdleTime)
+	defer cancel()
+	// ctx := context.Background()
+	client, err := Connect(ctx, p.metadata.connectionString)
+	if err != nil {
+		return fmt.Errorf("error connecting to configuration store: '%s'", err)
+	}
+	p.client = client
+	pingErr := p.client.Ping(ctx)
```
ItalyPaleAle, line 105: Can this be just `err`?

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,417 @@
...
+		p.logger.Error(err)
+		return nil, fmt.Errorf("error in configuration store query: '%s' ", err)
+	}
+	rows, err := p.client.Query(ctx, query, params...)
+	if err != nil {
+		// If no rows exist, return an empty response, otherwise return the error.
+		if err == sql.ErrNoRows {
+			return &configuration.GetResponse{}, nil
+		}
+		return nil, fmt.Errorf("error in querying configuration store: '%s'", err)
```
ItalyPaleAle, line 137: Here too, when wrapping an error please use `%w`

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+	items := make(map[string]*configuration.Item)
+	for i := 0; rows.Next(); i++ {
```
ItalyPaleAle, line 140: Looks like the `i` isn't used? `for rows.Next() {`?

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,417 @@
...
+		var item configuration.Item
+		var key string
+		var metadata []byte
+		v := make(map[string]string)
+		if err := rows.Scan(&key, &item.Value, &item.Version, &metadata); err != nil {
+			return nil, fmt.Errorf("error in reading data from configuration store: '%s'", err)
```
ItalyPaleAle, line 146: `%w`

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,417 @@
...
+		}
+		if err := json.Unmarshal(metadata, &v); err != nil {
+			return nil, fmt.Errorf("error in unmarshalling response from configuration store: '%s'", err)
```
ItalyPaleAle, line 149: `%w`

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,417 @@
...
+	"github.com/dapr/components-contrib/configuration"
+	"github.com/dapr/kit/logger"
+)
+
+type ConfigurationStore struct {
+	metadata             metadata
+	client               *pgxpool.Pool
+	logger               logger.Logger
+	configLock           sync.Mutex
+	subscribeStopChanMap map[string]interface{}
```
ItalyPaleAle, line 43: `map[string]chan struct{}`

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,417 @@
...
+}
+
+func (p *ConfigurationStore) Unsubscribe(ctx context.Context, req *configuration.UnsubscribeRequest) error {
+	p.configLock.Lock()
+	defer p.configLock.Unlock()
+	for k, v := range p.ActiveSubscriptions {
+		if v.uuid == req.ID {
+			if oldStopChan, ok := p.subscribeStopChanMap[req.ID]; ok {
+				delete(p.subscribeStopChanMap, req.ID)
+				close(oldStopChan.(chan struct{}))
```
ItalyPaleAle, line 181: Type hint not needed if you set the correct type for `p.subscribeStopChanMap`

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,417 @@
...
+				pgChannel := fmt.Sprintf(unlistenTemplate, k)
+				conn, err := p.client.Acquire(ctx)
+				if err != nil {
+					p.logger.Errorf("error acquiring connection:", err)
+					return fmt.Errorf("error acquiring connection: %s ", err)
+				}
+				defer conn.Release()
+				_, err = conn.Exec(ctx, pgChannel)
+				if err != nil {
+					p.logger.Errorf("error listening to channel:", err)
```
ItalyPaleAle, line 191: "un-listening" perhaps?

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,417 @@
...
+	for k, v := range p.ActiveSubscriptions {
+		if v.uuid == req.ID {
+			if oldStopChan, ok := p.subscribeStopChanMap[req.ID]; ok {
+				delete(p.subscribeStopChanMap, req.ID)
+				close(oldStopChan.(chan struct{}))
+				pgChannel := fmt.Sprintf(unlistenTemplate, k)
+				conn, err := p.client.Acquire(ctx)
+				if err != nil {
+					p.logger.Errorf("error acquiring connection:", err)
+					return fmt.Errorf("error acquiring connection: %s ", err)
```
ItalyPaleAle, line 186: `%w`

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,417 @@
...
+				}
+				defer conn.Release()
+				_, err = conn.Exec(ctx, pgChannel)
+				if err != nil {
+					p.logger.Errorf("error listening to channel:", err)
+					return fmt.Errorf("error listening to channel: %s", err)
```
ItalyPaleAle, line 192: `%w`

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+	defer conn.Release()
+	if _, err = conn.Exec(ctx, command); err != nil {
+		p.logger.Errorf("error listening to channel:", err)
+		return
+	}
+	for {
+		notification, err := conn.Conn().WaitForNotification(ctx)
+		if err != nil {
+			if !pgconn.Timeout(err) && !errors.Is(ctx.Err(), context.Canceled) {
```
ItalyPaleAle, line 217: `if !pgconn.Timeout(err) && err != context.Canceled {`

Because `err` may not be `ctx.Err()`

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,417 @@
...
+			}
+			return
+		}
+		p.handleSubscribedChange(ctx, handler, notification, channel, subscription)
+	}
+}
+
+func (p *ConfigurationStore) handleSubscribedChange(ctx context.Context, handler configuration.UpdateHandler, msg *pgconn.Notification, channel string, subscriptionID string) {
+	defer func() {
+		if err := recover(); err != nil {
```
ItalyPaleAle, line 228: Why is this needed? What can panic? Normally trying to recover from panics is anti-pattern in Go

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,417 @@
...
+	if m.maxIdleTimeout <= 0 {
+		m.maxIdleTimeout = defaultMaxConnIdleTime
+	}
+	return m, nil
+}
+
+func Connect(ctx context.Context, conn string) (*pgxpool.Pool, error) {
+	config, err := pgxpool.ParseConfig(conn)
+	if err != nil {
+		return nil, fmt.Errorf("postgres configuration store connection error : %s", err)
```
ItalyPaleAle, line 316: `%w`

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+	pool, err := pgxpool.NewWithConfig(ctx, config)
+	if err != nil {
+		return nil, fmt.Errorf("postgres configuration store connection error : %s", err)
+	}
+	pingErr := pool.Ping(ctx)
+	if pingErr != nil {
+		return nil, fmt.Errorf("postgres configuration store ping error : %s", pingErr)
```
ItalyPaleAle, line 324: `%w`

ItalyPaleAle (changes requested):
Few more changes :)

configuration/postgres/postgres.go
akhilac1: api.go throws an exception in case it is not able to successfully invoke the endpoint to post the notification. If we donot handle this, side car will crash. Hence this is needed.

configuration/postgres/postgres.go
```diff
@@ -0,0 +1,417 @@
...
+		pgNotifyCmd := fmt.Sprintf(listenTemplate, channel)
+		if sub, isActive := p.isSubscriptionActive(req); isActive {
+			if oldStopChan, ok := p.subscribeStopChanMap[sub]; ok {
+				close(oldStopChan.(chan struct{}))
+				delete(p.subscribeStopChanMap, sub)
+				delete(p.ActiveSubscriptions, channel)
+			}
+		}
+		stop := make(chan struct{})
+		subscribeID = uuid.New().String()
```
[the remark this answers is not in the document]
akhilac1: Are you referring to NewUUID? Recommedation is to use New method. - https://pkg.go.dev/github.com/google/UUID#NewUUID

configuration/postgres/postgres.go
akhilac1: https://github.com/dapr/components-contrib/blob/86b77cc269f728e552f2812ef0ba1a0d3607095d/configuration/postgres/postgres.go#L71-L74

configuration/postgres/postgres.go
akhilac1: Done - 
https://github.com/dapr/components-contrib/blob/86b77cc269f728e552f2812ef0ba1a0d3607095d/configuration/postgres/postgres.go#L76-L81

configuration/postgres/postgres.go
akhilac1: Updated for all applicable fmt.Error

configuration/postgres/postgres.go
akhilac1: done

configuration/postgres/postgres.go
akhilac1: Using pgxv5 CollectRows to iterate

akhilac1:
@ItalyPaleAle  - Comments are addressed. If there are no critical issue, please approve the review.

configuration/postgres/postgres.go
akhilac1: Done

configuration/postgres/postgres.go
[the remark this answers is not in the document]
ItalyPaleAle: No I'm referring to NewRandom: https://pkg.go.dev/github.com/google/UUID#NewRandom 

Internally, uuid.New uses NewRandom but discards the error, and panics in case of error. That's not ideal. https://github.com/google/UUID/blob/v1.0.0/version4.go#L13

configuration/postgres/postgres.go
```diff
@@ -126,35 +130,34 @@ func (p *ConfigurationStore) Get(ctx context.Context, req *configuration.GetRequ
...
-	items := make(map[string]*configuration.Item)
-	for i := 0; rows.Next(); i++ {
-		var item configuration.Item
-		var key string
-		var metadata []byte
-		v := make(map[string]string)
-		if err := rows.Scan(&key, &item.Value, &item.Version, &metadata); err != nil {
-			return nil, fmt.Errorf("error in reading data from configuration store: '%s'", err)
-		}
-		if err := json.Unmarshal(metadata, &v); err != nil {
```
ItalyPaleAle: I see this was removed, we don't need to unmarshall the JSON anymore? Please confirm

configuration/postgres/postgres.go
```diff
@@ -178,18 +181,18 @@ func (p *ConfigurationStore) Unsubscribe(ctx context.Context, req *configuration
...
-					return fmt.Errorf("error acquiring connection: %s ", err)
+					return fmt.Errorf("error acquiring connection: %w ", err)
 				}
 				defer conn.Release()
 				_, err = conn.Exec(ctx, pgChannel)
 				if err != nil {
-					p.logger.Errorf("error listening to channel:", err)
-					return fmt.Errorf("error listening to channel: %s", err)
+					p.logger.Errorf("error un-listening to channel:", err)
+					return fmt.Errorf("error listening to channel: %w", err)
```
ItalyPaleAle, line 195: un-listening here too

configuration/postgres/postgres.go
ItalyPaleAle: As per chat, we should not recover in components from panics caused in the runtime.

Please:

1. Open an issue in dapr/dapr explaining where the runtime crashes and under what circumstances. We will get that fixed before the release.
2. Remove the entire `defer` block that is recovering from panics.

ItalyPaleAle (changes requested):
Almost there :)

akhilac1:
https://github.com/dapr/dapr/issues/5236 - PR to track panic in runtime

configuration/postgres/postgres.go
akhilac1: yup. a new struct is added pgNotify 

https://github.com/dapr/components-contrib/blob/96fdc8a99701b308b214fd5dc0c3d4373ae2acfd/configuration/postgres/postgres.go#L53-L56

We are using a new pgxv5 generic API - pgx.CollectRows which simplifies reading the response from pgxpool.Pool.Query 
https://pkg.go.dev/github.com/jackc/pgx/v5#CollectRows

configuration/postgres/postgres.go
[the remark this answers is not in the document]
akhilac1: Updated

ItalyPaleAle (dismissed):

ItalyPaleAle:
LGTM!

berndverst (dismissed):

ItalyPaleAle (approved):

berndverst:
@akhilac1 please create documentation if you haven't already

akhilac1:
> @akhilac1 please create documentation if you haven't already

@berndverst -  https://github.com/dapr/docs/issues/2779

PR #2065, opened 2022-09-12
[WIP] Add default implementation for bulk pubsub

shubham1172:
# Description

This PR adds a default implementation for bulk messaging in pubsub. This is used to optimize sidecar<->app communication when the pubsub component does not support bulk messaging.

TODO: add more details

## Issue reference

Main issue: https://github.com/dapr/dapr/issues/2218

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


pubsub/bulk.go
```diff
@@ -0,0 +1,154 @@
...
+func (p *DefaultBulkMessager) bulkPublishParallel(req *BulkPublishRequest) (BulkPublishResponse, error) {
+	var statuses []BulkPublishResponseEntry
+	var wg sync.WaitGroup
+
+	for _, entry := range req.Entries {
+		wg.Add(1)
+
+		go func(entry BulkMessageEntry) {
+			defer wg.Done()
+			statuses = append(statuses, p.bulkPublishSingleEntry(req, entry))
```
mukundansundar, line 54: Put status in a channel and consume after waiting....

pubsub/pubsub.go
```diff
@@ -40,3 +56,60 @@ func Ping(pubsub PubSub) error {
...
+		p: pubsub,
+	}
+}
+
+// BulkPublish publishes messages to a broker that does not support bulk publish.
+// If 'bulkPublishKeepOrder' metadata is set to true, the messages are published serially
+// in the same order, otherwise they are published in parallel.
+func (p *DefaultBulkMessager) BulkPublish(req *BulkPublishRequest) (BulkPublishResponse, error) {
+	if req.Metadata[bulkPublishKeepOrderKey] == "true" {
+		return p.bulkPublishSerial(req)
```
mukundansundar, line 77: What is this new flag? 

Currently how is in order publishing supported?

mukundansundar (changes requested):

pubsub/pubsub.go
mukundansundar: @shubham1172 my concern here is that this might confuse the user and provide a false notion that in order publishing and processing is guaranteed whereas that is only available if broker supports it like for example partition key in AZ events hub..  

@berndverst @yaron2 @artursouza i think this flag should either be renamed or removed.... Thoughts?

pubsub/bulk.go
```diff
@@ -0,0 +1,154 @@
...
+	defaultMaxBulkAwaitDurationMs int = 5 * 1000
+)
+
+// bulkPublishSerial publishes messages in serial order.
+// This is slower, but ensures that messages are published in the same order as specified in the request.
+func (p *DefaultBulkMessager) bulkPublishSerial(req *BulkPublishRequest) (BulkPublishResponse, error) {
+	var statuses []BulkPublishResponseEntry
+
+	for _, entry := range req.Entries {
+		statuses = append(statuses, p.bulkPublishSingleEntry(req, entry))
```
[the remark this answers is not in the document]
shubham1172: Do you mean pre-allocating space (the linter error)? Done that.

pubsub/bulk.go
shubham1172: Thanks for the suggestion, updated! I have also preallocated the "capacity" of the initial slice.

pubsub/pubsub.go
shubham1172: The intent was to allow a fast way to publish everything in parallel, and a slower way to publish serially (and in-order if the broker supports). We can rename this to `bulkPublishSerially` if that helps to remove the confusion.

shubham1172:
This is a sample bulk publish with 1000 cloud events. First is serial publish and second is parallel.
image

shubham1172:
This will be a part of Dapr runtime as discussed in #2024, closing this PR.

PR #2066, opened 2022-09-12
Update log level for successful retry

shivamkm07:
# Description

In case of retries, pubsub.azure.eventhub returns error even after successful event processing: https://github.com/dapr/components-contrib/blob/2b5650c2faa41f0a8d82862583497103776be49f/pubsub/azure/eventhubs/eventhubs.go#L605

It should be a Warning and not an Error

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


ItalyPaleAle (dismissed):

mukundansundar (dismissed):

shivamkm07:
> If I am seeing the original error I also want to see the recovery with the same log level. otherwise the original error should have been warn level also since we were able to recover. Maybe only a complete failure should use log level error.

So, should both be warn levels? Some of the other components are having different levels, like https://github.com/dapr/components-contrib/blob/2b5650c2faa41f0a8d82862583497103776be49f/bindings/alicloud/rocketmq/rocketmq.go#L291
https://github.com/dapr/components-contrib/blob/2b5650c2faa41f0a8d82862583497103776be49f/internal/component/kafka/consumer.go#L43

artursouza (dismissed):

artursouza:
I think we should leave the retriable error as log error because when the message is given up, we would not have any log err.

berndverst:
Can we make the log level consistent please? @artursouza - should retriable errors log as error? I think warning or debug seems better? Thoughts?

berndverst:
> I think we should leave the retriable error as log error because when the message is given up, we would not have any log err.

I see. Makes sense. I think that's not ideal when things are retried and recover. We should change that somehow in our resiliency framework.

mukundansundar:
Can we make it consistent across components? Possibly a helper function to clearly wrap/log events in a consistent fashion similar to retry notify recover wrapper?

berndverst:
> Can we make it consistent across components? Possibly a helper function to clearly wrap/log events in a consistent fashion similar to retry notify recover wrapper? 

That's a separate question which does not belong on this issue @mukundansundar

berndverst (changes requested):
Please make both a warn here.

Then below in line 608 log the actual error!

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -602,7 +602,7 @@ func (aeh *AzureEventHubs) Subscribe(subscribeCtx context.Context, req pubsub.Su
 			}, b, func(_ error, _ time.Duration) {
 				aeh.logger.Errorf("Error processing EventHubs event: %s/%s. Retrying...", req.Topic, e.ID)
```
berndverst, line 603: This should also be WarnF

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -602,7 +602,7 @@ func (aeh *AzureEventHubs) Subscribe(subscribeCtx context.Context, req pubsub.Su
...
 			}, func() {
-				aeh.logger.Errorf("Successfully processed EventHubs event after it previously failed: %s/%s", req.Topic, e.ID)
+				aeh.logger.Warnf("Successfully processed EventHubs event after it previously failed: %s/%s", req.Topic, e.ID)
 			})
 		})
 	if err != nil {
```
berndverst, line 608: This is where we want to log the only Errorf as this indicates that the retries have been exhausted.

berndverst (changes requested):
See comments.

pubsub/azure/eventhubs/eventhubs.go
shivamkm07: Done

pubsub/azure/eventhubs/eventhubs.go
shivamkm07: Done

shivamkm07:
> > I think we should leave the retriable error as log error because when the message is given up, we would not have any log err.
> 
> I checked with @halspang and this is actually incorrect. We can log something separately after the retries are exhausted. So we should update this to be WarnF in the retry block, and then below we should add a new log statement for the ErrorF. IMO that is the right way to handle this.

Done

ItalyPaleAle (approved):

berndverst (approved):

PR #2067, opened 2022-09-12
Add kafka changes for bulk subscribe 

DeepanshuA:
# Description

Adds kafka implementation for bulk subscribe

## Issue reference

Please reference the issue this PR is related to: #https://github.com/dapr/dapr/issues/2218

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


internal/component/kafka/consumer.go
```diff
@@ -35,28 +37,111 @@ type consumer struct {
...
+		}
+	}
+	if isBulkSubscribe {
 		if consumer.k.consumeRetryEnabled {
 			if err := retry.NotifyRecover(func() error {
-				return consumer.doCallback(session, message)
+				return consumer.doBulkCallback(session, messages, claim.Topic())
 			}, b, func(err error, d time.Duration) {
-				consumer.k.logger.Errorf("Error processing Kafka message: %s/%d/%d [key=%s]. Error: %v. Retrying...", message.Topic, message.Partition, message.Offset, asBase64String(message.Key), err)
+				consumer.k.logger.Errorf("Error processing Kafka bulk messages: %s. Error: %v. Retrying...", claim.Topic(), err)
```
ItalyPaleAle, line 74: As per other issue, this should be a warning?

internal/component/kafka/consumer.go
```diff
@@ -35,28 +37,111 @@ type consumer struct {
...
 	return nil
 }
 
+func (consumer *consumer) doBulkCallback(session sarama.ConsumerGroupSession, messages []*sarama.ConsumerMessage, topic string) error {
+	consumer.k.logger.Debugf("Processing Kafka bulk message: %s", topic)
+	handler, err := consumer.k.GetTopicBulkHandler(topic)
+	if err != nil {
+		return err
+	}
+	var messageValues []KafkaBulkMessageEntry
```
ItalyPaleAle, line 96: Can this be merged into a single line?

```
messageValues := make([]KafkaBulkMessageEntry, 0)
```

internal/component/kafka/consumer.go
```diff
@@ -35,28 +37,111 @@ type consumer struct {
...
+	messageValues = make([]KafkaBulkMessageEntry, 0)
```
ItalyPaleAle, line 97: Here too you should be able to pre-allocate this with the size of `messages`:

```
messageValues = make([]KafkaBulkMessageEntry, len(messages))
...
n := 0
for i, message := range messages {
    if message != nil {
        messageValues[n] = childMessage
        n++
    }
}
messageValues = messageValues[:n]
```

internal/component/kafka/consumer.go
```diff
@@ -35,28 +37,111 @@ type consumer struct {
...
+
+	entryIds := GetEntryIds(len(messages))
+	for i, message := range messages {
+		if message != nil {
+			metadata := make(map[string]string)
```
ItalyPaleAle, line 102: Please pre-alloc this too with a capacity of `len(message.Headers)`

pubsub/pubsub.go
```diff
@@ -22,16 +22,30 @@ import (
 
 // PubSub is the interface for message buses.
 type PubSub interface {
+	BulkMessager
```
ItalyPaleAle, line 25: Per the conversation on the other PR, IMHO this should be done in the runtime and not via a stub method.

pubsub/utils.go
```diff
@@ -0,0 +1,23 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package pubsub
+
+// GetEntryIds provides list of entry Ids to be used for sending messages for bulk subscribe
+func GetEntryIds(n int) []int {
```
ItalyPaleAle, line 17: `GetEntryIDs` to be consistent (surprised the linter didn't call this out)

pubsub/utils.go
```diff
@@ -0,0 +1,23 @@
...
+	entryIds := make([]int, n)
```
ItalyPaleAle, line 18: Here too

ItalyPaleAle (changes requested):

pubsub/utils.go
shubham1172, line 17: Unit test.

pubsub/kafka/kafka.go
```diff
@@ -62,6 +65,46 @@ func (p *PubSub) Subscribe(ctx context.Context, req pubsub.SubscribeRequest, han
...
 
+func (p *PubSub) BulkSubscribe(ctx context.Context, req pubsub.SubscribeRequest, handler pubsub.BulkHandler) error {
+	p.kafka.AddTopicBulkHandler(req.Topic, adaptBulkHandler(handler))
+	maxBulkCount, err := strconv.Atoi(req.Metadata["maxBulkCount"])
+	if err != nil {
+		maxBulkCount = 20
+	}
+	maxBulkLatencyMilliSeconds, err := strconv.Atoi(req.Metadata["maxBulkLatencyMilliSeconds"])
+	if err != nil {
+		maxBulkLatencyMilliSeconds = 20
```
shubham1172, line 76: Defaults should be updated, maybe even have a common default for all components.

internal/component/kafka/consumer.go
DeepanshuA: Done.

internal/component/kafka/consumer.go
DeepanshuA: Agreed. Corrected.

internal/component/kafka/consumer.go
DeepanshuA: Done.

pubsub/utils.go
DeepanshuA: Changed entryID to not remain index based anymore. So, removed this part and corresponding heloer function in `pubsub`.

pubsub/utils.go
DeepanshuA: Removed as not required.

pubsub/utils.go
DeepanshuA: Removed this one, as not required.

internal/component/kafka/consumer.go
DeepanshuA: Changed to Warnf, as it will retry.

internal/component/kafka/consumer.go
```diff
@@ -35,28 +36,110 @@ type consumer struct {
...
+		if message != nil {
+			metadata := make(map[string]string, len(message.Headers))
+			if message.Headers != nil {
+				for _, t := range message.Headers {
+					metadata[string(t.Key)] = string(t.Value)
+				}
+			}
+			entryID, entryIDErr := uuid.NewRandom()
+			if entryIDErr != nil {
+				consumer.k.logger.Errorf("Failed to generate entryID for sending message for bulk subscribe event on topic: %s. Error: %v.", topic, err)
```
artursouza, line 107: You cannot really proceed if you got an error here. I suggest we skip this message and mark it as errored in the return automatically.

pubsub/kafka/kafka.go
```diff
@@ -62,6 +64,46 @@ func (p *PubSub) Subscribe(ctx context.Context, req pubsub.SubscribeRequest, han
 	return p.kafka.Subscribe(p.subscribeCtx)
 }
 
+func (p *PubSub) BulkSubscribe(ctx context.Context, req pubsub.SubscribeRequest, handler pubsub.BulkHandler) error {
+	p.kafka.AddTopicBulkHandler(req.Topic, adaptBulkHandler(handler))
+	maxBulkCount, err := strconv.Atoi(req.Metadata["maxBulkCount"])
+	if err != nil {
+		maxBulkCount = 20
```
artursouza, line 71: Can we set these magic numbers somewhere else?

pubsub/kafka/kafka.go
```diff
@@ -62,6 +64,46 @@ func (p *PubSub) Subscribe(ctx context.Context, req pubsub.SubscribeRequest, han
...
+			return
+		}
+
+		err := p.kafka.BulkSubscribe(p.subscribeCtx)
+		if err != nil {
+			p.logger.Errorf("kafka pubsub: error re-subscribing: %v", err)
+		}
+	}()
+
+	return p.kafka.BulkSubscribe(p.subscribeCtx)
```
artursouza, line 104: What happens if the connection is reset? Will the component reconnect?

artursouza (changes requested):

pubsub/kafka/kafka.go
DeepanshuA: Yes, if the connection with broker is stopped and then started back, component is reconnected.

internal/component/kafka/kafka.go
```diff
@@ -78,6 +84,9 @@ func (k *Kafka) Init(metadata map[string]string) error {
 	config := sarama.NewConfig()
 	config.Version = meta.Version
 	config.Consumer.Offsets.Initial = k.initialOffset
+	if meta.MinFetchBytes > 0 {
+		config.Consumer.Fetch.Min = meta.MinFetchBytes
```
mukundansundar, line 88: does this line imply that a consumer once initialized for bulk cannot be used for normal subscription?

internal/component/kafka/consumer.go
```diff
@@ -189,6 +315,84 @@ func (k *Kafka) Subscribe(ctx context.Context) error {
...
+
+	ready := make(chan bool)
+	k.consumer = consumer{
+		k:       k,
+		ready:   ready,
+		running: make(chan struct{}),
+	}
+
+	go func() {
+		k.logger.Debugf("Subscribed and listening to topics: %s", topics)
```
mukundansundar, line 354:
```suggestion
		k.logger.Debugf("Bulk Subscribed and listening to topics: %s", topics)
```

internal/component/kafka/consumer.go
```diff
@@ -35,28 +36,110 @@ type consumer struct {
...
 			}); err != nil {
-				consumer.k.logger.Errorf("Too many failed attempts at processing Kafka message: %s/%d/%d [key=%s]. Error: %v.", message.Topic, message.Partition, message.Offset, asBase64String(message.Key), err)
+				consumer.k.logger.Errorf("Too many failed attempts at processing Kafka message: %s. Error: %v.", claim.Topic(), err)
 			}
 		} else {
-			err := consumer.doCallback(session, message)
+			err := consumer.doBulkCallback(session, messages, claim.Topic())
 			if err != nil {
-				consumer.k.logger.Errorf("Error processing Kafka message: %s/%d/%d [key=%s]. Error: %v.", message.Topic, message.Partition, message.Offset, asBase64String(message.Key), err)
+				consumer.k.logger.Errorf("Error processing Kafka message: %s. Error: %v.", claim.Topic(), err)
```
mukundansundar, line 82: Just a note : The whole set of messages are retried here ....

internal/component/kafka/consumer.go
```diff
@@ -101,6 +184,40 @@ func (k *Kafka) RemoveTopicHandler(topic string) {
 	k.subscribeLock.Unlock()
 }
 
+// AddBulkSubscribeConfig adds bulk subscribe config
+func (k *Kafka) AddBulkSubscribeConfig(maxBulkCount int, maxBulkAwaitDurationMilliSeconds int, maxBulkSizeBytes int) {
+	k.bulkSubscribeConfig.MaxBulkCount = maxBulkCount
+	k.bulkSubscribeConfig.MaxBulkAwaitDurationMilliSeconds = maxBulkAwaitDurationMilliSeconds
+	k.bulkSubscribeConfig.MaxBulkSizeBytes = maxBulkSizeBytes
```
mukundansundar, line 191: These two options are not used as of now.

mukundansundar (changes requested):

internal/component/kafka/consumer.go
```diff
@@ -35,28 +36,110 @@ type consumer struct {
 
 func (consumer *consumer) ConsumeClaim(session sarama.ConsumerGroupSession, claim sarama.ConsumerGroupClaim) error {
 	b := consumer.k.backOffConfig.NewBackOffWithContext(session.Context())
+	var messages []*sarama.ConsumerMessage
+	messages = make([]*sarama.ConsumerMessage, 0)
+	isBulkSubscribe := consumer.k.checkBulkSubscribe(claim.Topic())
 	for message := range claim.Messages() {
+		if isBulkSubscribe {
+			if len(messages) < consumer.k.bulkSubscribeConfig.MaxBulkCount {
```
mukundansundar, line 44: MaxBulkCount here works as the minimum count after which you break out of the loop ...

internal/component/kafka/consumer.go
DeepanshuA: I could think of 4 options here:
1. return from func doBulkCallback() with an error as system doesn't seem to have enough entropy (if I understand correctly). Not opting for it as found Option 4 to be better.
2. don't pass it on this message and mark with an error in the return automatically - Right now, I depend on runtime, to process and pass back responses to me. And, I mark the messages as per theses reponses. But, to take care of this message on the kafka implementation side, I would have to create a new response array here itself and track of the indexes which I didn't send, in some way. So, want to avoid this.
Infact, if an entryID is passed as blank, runtime is expected to pass back an error pertaining to this message at the same index in return reponse array. ()
3. Pass a blank entryID here and rely on runtime to provide back an error for this message.
4. use index as entryID here.

I am opting for 4th option. Please let me know if that works.

pubsub/kafka/kafka.go
DeepanshuA: metadata.go seems to have metadata specific to the one given in kafka component yaml. These metadata are provided per topic based in response to "/dapr/subscribe", created a struct in internal/component/kafka/kafka.go, as it already has handlers per topic based.

internal/component/kafka/consumer.go
```diff
@@ -101,6 +184,40 @@ func (k *Kafka) RemoveTopicHandler(topic string) {
...
+func (k *Kafka) checkBulkSubscribe(topic string) bool {
+	topicBulkHandler, ok := k.bulkSubscribeTopics[topic]
+	if !ok {
+		return false
+	}
+	if topicBulkHandler != nil {
+		return true
+	} else {
+		return false
+	}
```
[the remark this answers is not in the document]
DeepanshuA: Done

pubsub/kafka/kafka.go
```diff
@@ -62,6 +64,46 @@ func (p *PubSub) Subscribe(ctx context.Context, req pubsub.SubscribeRequest, han
 	return p.kafka.Subscribe(p.subscribeCtx)
 }
 
+func (p *PubSub) BulkSubscribe(ctx context.Context, req pubsub.SubscribeRequest, handler pubsub.BulkHandler) error {
+	p.kafka.AddTopicBulkHandler(req.Topic, adaptBulkHandler(handler))
+	maxBulkCount, err := strconv.Atoi(req.Metadata["maxBulkCount"])
```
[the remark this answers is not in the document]
DeepanshuA: Done

pubsub/kafka/kafka.go
```diff
@@ -62,6 +64,46 @@ func (p *PubSub) Subscribe(ctx context.Context, req pubsub.SubscribeRequest, han
...
+	}
+	maxBulkAwaitDurationMilliSeconds, err := strconv.Atoi(req.Metadata["maxBulkAwaitDurationMilliSeconds"])
+	if err != nil {
+		maxBulkAwaitDurationMilliSeconds = 20
+	}
+	maxBulkSizeBytes, err := strconv.Atoi(req.Metadata["maxBulkSizeBytes"])
+	if err != nil {
+		maxBulkSizeBytes = 20
+	}
+	p.kafka.AddBulkSubscribeConfig(maxBulkCount, maxBulkAwaitDurationMilliSeconds, maxBulkSizeBytes)
```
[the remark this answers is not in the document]
DeepanshuA: Agreed, changed it per topic based.

internal/component/kafka/metadata.go
```diff
@@ -139,6 +140,13 @@ func (k *Kafka) getKafkaMetadata(metadata map[string]string) (*kafkaMetadata, er
 
 	k.logger.Debugf("Found brokers: %v", meta.Brokers)
 
+	if val, ok := metadata["minFetchBytes"]; ok && val != "" {
```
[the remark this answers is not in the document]
DeepanshuA: As discussed offline, removed any new setting to sarama config.

internal/component/kafka/metadata.go
```diff
@@ -139,6 +140,13 @@ func (k *Kafka) getKafkaMetadata(metadata map[string]string) (*kafkaMetadata, er
...
+		val64, _ := strconv.ParseInt(val, 10, 32)
+		meta.MinFetchBytes = int32(val64)
+	} else {
+		meta.MinFetchBytes = 0
```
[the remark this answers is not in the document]
DeepanshuA: Removed minFetchBytes setting to sarama.

internal/component/kafka/kafka.go
DeepanshuA: Removed any new setting to kafka sarama config.

internal/component/kafka/consumer.go
DeepanshuA: Yes, this is on the same lines, as we had previously discussed that resiliency policies / retries will not be in effect in this release. Though resiliency and retry are different but their handling would be same in case of bulk pub/sub. 
So, in effect - this is a separate work TODO for upcoming release.

internal/component/kafka/consumer.go
DeepanshuA: MaxBulkAwaitDurationMilliSeconds is being used now - removed MaxBulkSizeBytes as it is not used as of now.

pubsub/kafka/kafka.go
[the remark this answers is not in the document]
DeepanshuA: Moved

internal/component/kafka/kafka.go
```diff
@@ -146,10 +155,28 @@ func (k *Kafka) Close() (err error) {
...
+
+// KafkaBulkMessage is a bulk event arriving from a message bus instance.
+type KafkaBulkMessage struct {
+	Entries  []KafkaBulkMessageEntry `json:"entries"`
+	Topic    string                  `json:"topic"`
+	Metadata map[string]string       `json:"metadata"`
+}
+
+// KafkaBulkMessageEntry is an item contained inside bulk event arriving from a message bus instance.
+type KafkaBulkMessageEntry struct {
```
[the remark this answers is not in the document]
DeepanshuA: The reason that I understand is that `internal/kafka` will start depending on pubsub specifc structs, which won't be desired. And, that seems to be the reason that in pubsub/kafka.go, adaption of handler is done.

internal/component/kafka/kafka.go
```diff
@@ -146,10 +155,28 @@ func (k *Kafka) Close() (err error) {
...
 // NewEvent is an event arriving from a message bus instance.
 type NewEvent struct {
 	Data        []byte            `json:"data"`
 	Topic       string            `json:"topic"`
 	Metadata    map[string]string `json:"metadata"`
 	ContentType *string           `json:"contentType,omitempty"`
 }
+
+// KafkaBulkMessage is a bulk event arriving from a message bus instance.
+type KafkaBulkMessage struct {
```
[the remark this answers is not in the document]
DeepanshuA: Same as above response.

internal/component/kafka/consumer.go
```diff
@@ -189,6 +335,84 @@ func (k *Kafka) Subscribe(ctx context.Context) error {
...
+
+			// Consume the requested topics
+			bo := backoff.WithContext(backoff.NewConstantBackOff(k.consumeRetryInterval), ctx)
+			innerErr := retry.NotifyRecover(func() error {
+				if ctxErr := ctx.Err(); ctxErr != nil {
+					return backoff.Permanent(ctxErr)
+				}
+				return k.cg.Consume(ctx, topics, &(k.consumer))
+			}, bo, func(err error, t time.Duration) {
+				k.logger.Errorf("Error consuming %v. Retrying...: %v", topics, err)
```
mukundansundar, line 393: Warnf

mukundansundar (changes requested):
Some small changes. Overall looks good

mukundansundar (approved):
Lgtm

internal/component/kafka/consumer.go
```diff
@@ -35,28 +37,138 @@ type consumer struct {
...
+		}
+	}
+	return nil
+}
+
+func (consumer *consumer) processBulkMessages(ctx context.Context, session sarama.ConsumerGroupSession, claim sarama.ConsumerGroupClaim,
+	b backoff.BackOff) error {
+	handlerConfig, err := consumer.k.GetTopicBulkHandlerConfig(claim.Topic())
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
DeepanshuA: Done

pubsub/kafka/kafka.go
DeepanshuA: Done

metadata/utils.go
```diff
@@ -37,6 +37,12 @@ const (
 
 	// QueryIndexName defines the metadata key for the name of query indexing schema (for redis).
 	QueryIndexName = "queryIndexName"
+
+	// MaxBulkCountKey is the key for the max bulk count in the metadata.
+	MaxBulkCountKey string = "maxBulkCount"
```
artursouza, line 42: Is this an attribute proposed for all components? If this is Kafka specific, just keep it in kafka.go.

internal/component/kafka/consumer.go
```diff
@@ -53,10 +59,113 @@ func (consumer *consumer) ConsumeClaim(session sarama.ConsumerGroupSession, clai
...
+			if message.Headers != nil {
+				for _, t := range message.Headers {
+					metadata[string(t.Key)] = string(t.Value)
+				}
+			}
+			var entryIDStr string
+			entryID, entryIDErr := uuid.NewRandom()
+			if entryIDErr != nil {
+				consumer.k.logger.Warnf("Failed to generate UUID for sending as entryID in message for bulk subscribe event on topic: %s. Error: %v.", topic, entryIDErr)
+				entryIDStr = strconv.Itoa(i)
```
artursouza, line 133: Why not simply default to this? What is the value to try to generate an UUID for a temporary id?

artursouza (commented):
Two minor comments.

metadata/utils.go
DeepanshuA: Yes, ideally "maxBulkCount" and "maxBulkAwaitDurationMilliSeconds" are attributes proposed for all components. As, while consuming messages from broker, there needs to be some limit, either in terms of count of messages or max await time, after which BulkSubscribe handler to send ahead messages should be called.

internal/component/kafka/consumer.go
DeepanshuA: In a particular bulk subscribe call, entryID needs to be unique - hence, I can't make it default. Ideally, NewRandom() will not error out, it is just an extra bit of precaution taken.

DeepanshuA: As discussed offline, I got your intent - Why not use index itself an entryID ad uuid seems overkill? I am perfectly fine with that, as I myself also had earlier assigned index as entryID. I have changed to that.

artursouza (approved):

Issue #1909, opened 2022-07-25
Add Dapr support for Redis 7 in cluster mode


thielepaul:
## Expected Behavior

I expect dapr to connect to the latest version of redis.

## Actual Behavior

When I connect to a redis instance deployed to kubernetes, I get the following error:
```
time="2022-07-25T14:55:59.411020038Z" level=fatal msg="process component statestore error: redis store: error connecting to redis at state-store-redis-cluster:6379: got 4 elements in cluster info address, expected 2 or 3" app_id=subscriber instance=subscriber-deployment-68b65fd66b-tp6tc scope=dapr.runtime type=log ver=1.7.2
```

## Steps to Reproduce the Problem

- deploy redis in cluster mode: https://github.com/bitnami/charts/tree/master/bitnami/redis-cluster#tldr
- configure dapr to connect to it

It seems we should use github.com/go-redis/redis/v9 to support redis 7: https://github.com/go-redis/redis/issues/2085#issuecomment-1181357149

As you can see in the log message I am currently using dapr 1.7.2, but it seems the bug is also in the current master: https://github.com/dapr/components-contrib/blob/master/go.mod#L65 

## Release Note

RELEASE NOTE:

yaron2:
Thanks, would you like to create a PR that updates `go.mod` to the `v9` version?

thielepaul:
I just hat a short look on that, and it seems it is more complicated than just updating the dependency: 
> If you are using Redis 6, install go-redis/v8
> If you are using Redis 7, install go-redis/v9
> (https://github.com/go-redis/redis#installation)

Probably, it would make sense to have the redis version configurable in the component specification and then use `go-redis/redis/v8` or `go-redis/redis/v9` depending on what is selected there.


> Thanks, would you like to create a PR that updates go.mod to the v9 version?

I'm sorry, but I currently don't have the time to do this and using redis 6 for the time being is also not an issue for us.

PR #2085, opened 2022-09-15
Adds support for Redis 7 via `redisVersion` metadata in all redis components

berndverst:
# Description

Adds support for Redis 7 via `redisVersion` metadata in all redis components

When setting `redisVersion` to `7` / `v7` or greater the new Redis Go SDK v9 will be used. For Redis 6 and below the v8 SDK will be used.

## Issue reference

#1909

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


berndverst:
More work needed to make certification tests pass.

bindings/redis/redis.go
```diff
@@ -42,14 +45,25 @@ func NewRedis(logger logger.Logger) bindings.OutputBinding {
...
+	}
 	if err != nil {
 		return err
 	}
 
 	r.ctx, r.cancel = context.WithCancel(context.Background())
 
-	_, err = r.client.Ping(r.ctx).Result()
+	if r.legacyRedis {
+		_, err = r.clientv8.Ping(r.ctx).Result()
```
artursouza, line 63: Is there a way to use Go lang to avoid these it checks? Could we declare an interface containing the methods we use here and declare a reference to that? Then, the only difference would be which instanced is referenced but they would both implement the same interface.

Issue #2060, opened 2022-09-09
Implement Subscribe and Unsubscribe for Azure App Configuration component 


pravinpushkar:
## Describe the feature
Enhance in `Azure App Configuration` component for Configuration store building block to support `Subscribe` and `Unsusbcribe` feature. Currently only `get` is supported.

Related to #1541 

RELEASE NOTE: **ADD** Support for Subscribe and Unsubscribe to Configurations in **Azure App Configuration**

pravinpushkar:
the SDK for Azure app config https://github.com/Azure/azure-sdk-for-go does not support any out of the box method for subscribing the changes to App Configuration objects.

According to the documentation here  https://docs.microsoft.com/en-us/azure/azure-app-configuration/howto-best-practices, I can see there can be two approaches - 
1. Pull based  -  Create a sentinel key and poll for the changes to this key periodically. If this key changes then get all values from configuration store. Here we can either send all values irrespective of what has changed or we can keep a local cache and send only those which has changed.

Downside - This is a poll based approach hence there will be unnecessary network calls and changed values might come with delay.

2. Push based - As mentioned here in this doc - https://docs.microsoft.com/en-us/azure/azure-app-configuration/howto-app-configuration-event , `Azure Event grid` might be used to get the changed configuration items. The event grid needs to have an `app url` configured to send the updated values.

Downside - this approach depends on another azure service plus IMO this approach can directly be leveraged by apps.

Kindly share your thoughts on this.

mukundansundar:
I belive we should implement the pull based approach with a sentinel key.

@berndverst @ItalyPaleAle thoughts?

ItalyPaleAle:
Not sure I like the idea of the "sentinel key" because that would require operators to be aware of that. Is there another approach that doesn't depend on external Azure services and yet works in a way that people would expect?

pravinpushkar:
I did not find any other approach apart from these. 
The sentinel key is just the optimisation that the doc is suggesting to limit the no of calls. The other thing I was thinking, can we mandate the sentinel key to passed as metadata if subscribe feature is to be used with azure app config store.

I have created an issue to check with the go-sdk team here for getting view as well - 
https://github.com/Azure/azure-sdk-for-go/issues/19081

ItalyPaleAle:
> can we mandate the sentinel key to be passed as metadata if subscribe feature is to be used with azure app config store

Yes, I think that would be good.

[...]

Issue #191, opened 2020-01-16
[Proposal] Blob Storage as State Store


aloneguid:
## Describe the proposal

Blob storage is currently supported as output bindings (S3, Azure etc.) however I can see scenarios when blob storage would be extremely useful for state stores as well. 

First, considering the availability of tools, it's much easier to view blob storage compared to say Dynamo DB.

Second, blob storage is extremely cheap for low traffic scenarios. I'm not quite sure whether blob storage is actually cheaper for high traffic as well.

Blob storage is a key-value storage too - key is blob name, value is blob content. One can attach key-value metadata for a blob as well. Optimistic concurrency is supported via ETags (for instance in Azure Blob Storage).

msfussell:
@aloneguid - are you interested in adding Azure Blob Storage as a state store?

ksivamuthu:
I'm interested to do if it is free to take.

aloneguid:
@msfussell I'm happy for @ksivamuthu to do it, I'm not too active for a next month or so. Looking forward to contribute to some other component soon.

youngbupark:
@ksivamuthu Can you please create docs as well ?

youngbupark:
@ksivamuthu please follow [this doc style](https://github.com/dapr/docs/tree/master/contributing#contributing-to-how-tos) when you create the howto.

PR #2088, opened 2022-09-16
Azure Appconfig configuration store - Subscribe and Unsubscribe impl

pravinpushkar:
## Issue reference

Please reference the issue this PR will close: #2060 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


berndverst:
@pravinpushkar is this still work in progress or is this ready for review? Marking as draft for now, but please note it also has a conflict.

pravinpushkar:
> @pravinpushkar is this still work in progress or is this ready for review? Marking as draft for now, but please note it also has a conflict.

Yes..this is still wip. I will resolve the conflicts and will write few unit tests.

configuration/azure/appconfig/appconfig_test.go
```diff
@@ -113,6 +113,71 @@ func Test_getConfigurationWithProvidedKeys(t *testing.T) {
...
 
+func Test_subscribeConfigurationWithProvidedKeys(t *testing.T) {
+	s := NewAzureAppConfigurationStore(logger.NewLogger("test")).(*ConfigurationStore)
+
+	s.client = &MockConfigurationStore{}
+
+	metadata := make(map[string]string)
+	metadata["sentinelKey"] = "test_sentinel_key"
+
+	t.Run("call subscribe", func(t *testing.T) {
```
shubham1172, line 124: Both the test names here are the same?

configuration/azure/appconfig/appconfig.go
```diff
@@ -248,9 +266,103 @@ func (r *ConfigurationStore) getLabelFromMetadata(metadata map[string]string) *s
 }
 
 func (r *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.SubscribeRequest, handler configuration.UpdateHandler) (string, error) {
-	return "", fmt.Errorf("Subscribe is not implemented by this configuration store")
+	subscribeID := uuid.New().String()
```
shubham1172, line 269: This may panic,  see https://pkg.go.dev/github.com/google/uuid#New

Please replace with `NewRandom` and handle the error.

configuration/azure/appconfig/appconfig.go
```diff
@@ -248,9 +266,103 @@ func (r *ConfigurationStore) getLabelFromMetadata(metadata map[string]string) *s
...
+
+func (r *ConfigurationStore) doSubscribe(ctx context.Context, req *configuration.SubscribeRequest, handler configuration.UpdateHandler, sentinelKey string, id string, stop chan struct{}) {
+	for {
+		// get sentinel key changes
+		_, err := r.Get(ctx, &configuration.GetRequest{
+			Keys:     []string{sentinelKey},
+			Metadata: req.Metadata,
+		})
+		if err != nil {
+			r.logger.Debugf("fail to get sentinel key changes or sentinel key's value is unchanged: %s", err)
```
mukundansundar, line 308: Add continue after this line and remove else nesting?

configuration/azure/appconfig/appconfig.go
```diff
@@ -248,9 +266,103 @@ func (r *ConfigurationStore) getLabelFromMetadata(metadata map[string]string) *s
...
+		} else {
+			items, err := r.Get(ctx, &configuration.GetRequest{
+				Keys:     req.Keys,
+				Metadata: req.Metadata,
+			})
+			if err != nil {
+				r.logger.Errorf("fail to get configuration key changes: %s", err)
```
mukundansundar, line 315: Add continue after this line and remove else nesting?

configuration/azure/appconfig/appconfig.go
```diff
@@ -248,9 +266,103 @@ func (r *ConfigurationStore) getLabelFromMetadata(metadata map[string]string) *s
...
+			} else {
+				r.handleSubscribedChange(ctx, req, handler, items, id)
+			}
+		}
+		select {
+		case <-stop:
+			return
+		case <-ctx.Done():
+			return
+		case <-time.After(r.metadata.subscribePollInterval):
```
mukundansundar, line 325: Just to confirm, the next iteration of for will go through only after poll interval + handler block time have passed correct? So the poll interval is in addition to how long the app takes to handle a particular key update....

configuration/azure/appconfig/appconfig.go
```diff
@@ -248,9 +266,103 @@ func (r *ConfigurationStore) getLabelFromMetadata(metadata map[string]string) *s
...
+	value, found := r.subIDToKeyListMap[subscribeID]
+	if found {
+		value = append(value, actualKeyToSubscribeSlice...)
+		r.subIDToKeyListMap[subscribeID] = value
+	} else {
+		r.subIDToKeyListMap[subscribeID] = actualKeyToSubscribeSlice
+	}
+
+	req.Keys = actualKeyToSubscribeSlice
+	go r.doSubscribe(ctx, req, handler, sentinelKey, subscribeID, stop)
```
mukundansundar, line 296: Can we pass in keys and metadata separately instead of modifying req and then passing in that?

mukundansundar (changes requested):

configuration/azure/appconfig/appconfig.go
```diff
@@ -248,9 +266,103 @@ func (r *ConfigurationStore) getLabelFromMetadata(metadata map[string]string) *s
...
+		}
+	}
+}
+
+func (r *ConfigurationStore) handleSubscribedChange(ctx context.Context, req *configuration.SubscribeRequest, handler configuration.UpdateHandler, items *configuration.GetResponse, id string) {
+	defer func() {
+		if err := recover(); err != nil {
+			r.logger.Errorf("panic in handleSubscribedChange(）method and recovered: %s", err)
+		}
+	}()
```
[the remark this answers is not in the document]
shubham1172: This should be removed per our last discussion, i.e. components should not handle panic, runtime should.

configuration/azure/appconfig/appconfig.go
pravinpushkar: @shubham1172 I think it will make sense to have retry in case of duplicate string rather than failing the subscribe. Given that probability of duplicate is very rare will this matter much. Thoughts ?

quote from new Random method - 
_//  Randomly generated UUIDs have 122 random bits.  One's annual risk of being
//  hit by a meteorite is estimated to be one chance in 17 billion, that
//  means the probability is about 0.00000000006 (6 × 10−11),
//  equivalent to the odds of creating a few tens of trillions of UUIDs in a
//  year and having one duplicate._

configuration/azure/appconfig/appconfig.go
```diff
@@ -32,14 +34,16 @@ import (
...
-	host                 = "appConfigHost"
-	connectionString     = "appConfigConnectionString"
-	maxRetries           = "maxRetries"
-	retryDelay           = "retryDelay"
-	maxRetryDelay        = "maxRetryDelay"
-	defaultMaxRetries    = 3
-	defaultRetryDelay    = time.Second * 4
-	defaultMaxRetryDelay = time.Second * 120
+	host                         = "appConfigHost"
+	connectionString             = "appConfigConnectionString"
```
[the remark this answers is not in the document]
pravinpushkar: I will change this and also will create a doc issue for this.

configuration/azure/appconfig/appconfig.go
```diff
@@ -32,14 +34,16 @@ import (
...
+	maxRetries                   = "maxRetries"
+	retryDelay                   = "retryDelay"
+	maxRetryDelay                = "maxRetryDelay"
+	subscribePollInterval        = "subscribePollInterval"
+	defaultMaxRetries            = 3
+	defaultRetryDelay            = time.Second * 4
+	defaultMaxRetryDelay         = time.Second * 120
+	defaultSubscribePollInterval = time.Second * 30
```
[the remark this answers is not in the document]
pravinpushkar: There was some discussion to keep it once in 24hrs as default So , I will change this to 24 hrs. Thought ?

configuration/azure/appconfig/appconfig.go
```diff
@@ -248,9 +266,103 @@ func (r *ConfigurationStore) getLabelFromMetadata(metadata map[string]string) *s
...
+	subscribeID := uuid.New().String()
+	stop := make(chan struct{})
+	r.subscribeStopChanMap.Store(subscribeID, stop)
+	sentinelKey := r.getSentinelKeyFromMetadata(req.Metadata)
+	if sentinelKey == "" {
+		return "", fmt.Errorf("sentinel key is not provided in metadata")
+	}
+	actualKeyToSubscribeSlice := make([]string, 0)
+	for _, key := range req.Keys {
+		_, found := r.keysToSubIDMap[key]
```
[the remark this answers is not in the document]
pravinpushkar: This was intentional actually. Since that sentinel key is already being watched, so any change to that sentinel key will fetch the new values. So there is no need to watch again for same key in a diff subscribe call.

configuration/azure/appconfig/appconfig.go
shubham1172: I think it is safe to avoid duplication handling if these UUIDs are purely in-memory and ephemeral.

configuration/azure/appconfig/appconfig.go
```diff
@@ -248,9 +266,103 @@ func (r *ConfigurationStore) getLabelFromMetadata(metadata map[string]string) *s
...
+		close(oldStopChan.(chan struct{}))
+		keys, found := r.subIDToKeyListMap[req.ID]
+		if found {
+			for _, key := range keys {
+				delete(r.keysToSubIDMap, key)
+			}
+			delete(r.subIDToKeyListMap, req.ID)
+		}
+	}
+	return nil
```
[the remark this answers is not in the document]
pravinpushkar: Thanks for catching this. It should return error and it is being handled in other redis.

configuration/azure/appconfig/appconfig.go
pravinpushkar: No need to modify the req any ore

configuration/azure/appconfig/appconfig.go
pravinpushkar: yes

configuration/azure/appconfig/appconfig.go
pravinpushkar: This will make the for loop start again but we want to start only after a timeInterval rt ?

configuration/azure/appconfig/appconfig.go
pravinpushkar: same as below

configuration/azure/appconfig/appconfig.go
```diff
@@ -248,9 +262,75 @@ func (r *ConfigurationStore) getLabelFromMetadata(metadata map[string]string) *s
...
 
 func (r *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.SubscribeRequest, handler configuration.UpdateHandler) (string, error) {
-	return "", fmt.Errorf("Subscribe is not implemented by this configuration store")
+	sentinelKey := r.getSentinelKeyFromMetadata(req.Metadata)
+	if sentinelKey == "" {
+		return "", fmt.Errorf("sentinel key is not provided in metadata")
+	}
+	uuid, err := uuid.NewRandom()
+	if err != nil {
+		return "", fmt.Errorf("failed to generate uuid, error is %s", err)
```
mukundansundar, line 271: Nitpick 
```suggestion
		return "", fmt.Errorf("failed to generate subscription id, error is %s", err)
```

mukundansundar (approved):
Lgtm

configuration/azure/appconfig/appconfig.go
ItalyPaleAle: Actually for errors better to use `%w` which wraps the original error.

configuration/azure/appconfig/appconfig.go
```diff
@@ -248,9 +262,75 @@ func (r *ConfigurationStore) getLabelFromMetadata(metadata map[string]string) *s
 }
 
 func (r *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.SubscribeRequest, handler configuration.UpdateHandler) (string, error) {
-	return "", fmt.Errorf("Subscribe is not implemented by this configuration store")
+	sentinelKey := r.getSentinelKeyFromMetadata(req.Metadata)
+	if sentinelKey == "" {
+		return "", fmt.Errorf("sentinel key is not provided in metadata")
```
mukundansundar, line 267:
```suggestion
		return "", fmt.Errorf("azure appconfig error: sentinel key is not provided in metadata")
```

configuration/azure/appconfig/appconfig.go
```diff
@@ -248,9 +262,75 @@ func (r *ConfigurationStore) getLabelFromMetadata(metadata map[string]string) *s
...
+		})
+		if err == nil {
+			r.logger.Debugf("fail to get sentinel key changes or sentinel key's value is unchanged: %s", err)
+		} else {
+			items, err := r.Get(ctx, &configuration.GetRequest{
+				Keys:     req.Keys,
+				Metadata: req.Metadata,
+			})
+			if err != nil {
+				r.logger.Errorf("fail to get configuration key changes: %s", err)
```
mukundansundar, line 295:
```suggestion
				r.logger.Errorf("azure appconfig error: fail to get configuration key changes: %s", err)
```

mukundansundar (dismissed):
lgtm. small nitpicks

configuration/azure/appconfig/appconfig.go
```diff
@@ -248,9 +262,75 @@ func (r *ConfigurationStore) getLabelFromMetadata(metadata map[string]string) *s
...
+	sentinelKey := r.getSentinelKeyFromMetadata(req.Metadata)
+	if sentinelKey == "" {
+		return "", fmt.Errorf("sentinel key is not provided in metadata")
+	}
+	uuid, err := uuid.NewRandom()
+	if err != nil {
+		return "", fmt.Errorf("failed to generate subscription id, error is %w", err)
+	}
+	subscribeID := uuid.String()
+	stop := make(chan struct{})
```
ItalyPaleAle, line 274: Rather than using a separate channel, create a context that is a child of `ctx` and store that in `subscribeStopChanMap`. To unsubscribe, then, cancel that context.

configuration/azure/appconfig/appconfig.go
```diff
@@ -248,9 +262,75 @@ func (r *ConfigurationStore) getLabelFromMetadata(metadata map[string]string) *s
...
+	for {
+		// get sentinel key changes
+		_, err := r.Get(ctx, &configuration.GetRequest{
+			Keys:     []string{sentinelKey},
+			Metadata: req.Metadata,
+		})
+		if err != nil {
+			r.logger.Debugf("fail to get sentinel key changes or sentinel key's value is unchanged: %s", err)
+		} else {
+			items, err := r.Get(ctx, &configuration.GetRequest{
```
ItalyPaleAle, line 290: Needs a timeout here. Use a sub-context with a timeout

configuration/azure/appconfig/appconfig.go
```diff
@@ -248,9 +262,75 @@ func (r *ConfigurationStore) getLabelFromMetadata(metadata map[string]string) *s
...
+	stop := make(chan struct{})
+	r.subscribeStopChanMap.Store(subscribeID, stop)
+	go r.doSubscribe(ctx, req, handler, sentinelKey, subscribeID, stop)
+	return subscribeID, nil
+}
+
+func (r *ConfigurationStore) doSubscribe(ctx context.Context, req *configuration.SubscribeRequest, handler configuration.UpdateHandler, sentinelKey string, id string, stop chan struct{}) {
+	for {
+		// get sentinel key changes
+		_, err := r.Get(ctx, &configuration.GetRequest{
```
ItalyPaleAle, line 283: Needs a timeout here. Use a sub-context with a timeout

ItalyPaleAle (changes requested):

configuration/azure/appconfig/appconfig.go
pravinpushkar: Instead of having timeout here I think we shoud have it in actual Get #191. 
But, While testing I found that when a cancelled context is passed to GetSetting#191, az appconfig sdk panics.
refer - https://github.com/Azure/azure-sdk-for-go/issues/19223
For now, I have used context.TODO in the actual GetSettings call. We need to change this when there is fix in sdk.

configuration/azure/appconfig/appconfig.go
pravinpushkar: same as above

configuration/azure/appconfig/appconfig.go
ItalyPaleAle: Sounds good. Can you please put a comment with a TODO for remind us all in the future about this? Feel free to link to the issue in the SDK repo.

Issue #2072, opened 2022-09-13
Native support for message TTL in RabbitMQ PubSub


artursouza:
## Describe the feature
RabbitMQ PubSub does not support Message TTL natively, although the broker does support it. Handle messageTTL metadata like it is done in ASB: https://github.com/dapr/components-contrib/search?l=Go&q=FeatureMessageTTL

## Release Note

RELEASE NOTE: **ADDED** native support for message TTL in RabbitMQ PubSub

Taction:
I'll add support for this.

PR #2093, opened 2022-09-17
Add pubsub rabbitmq TTL support

Taction:
# Description

Add pubsub rabbitmq TTL support.
Add a queue level TTL metadata, which can be rewritten by metadata defined in each call.

## Issue reference

Please reference the issue this PR will close: #2072

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#2808


pubsub/rabbitmq/rabbitmq.go
```diff
@@ -190,10 +193,22 @@ func (r *rabbitMQ) publishSync(req *pubsub.PublishRequest) (rabbitMQChannelBroke
...
 
+	ttl, ok, err := contribMetadata.TryGetTTL(req.Metadata)
+	if err != nil {
+		r.logger.Errorf("%s publishing to %s failed in TryGetTTL: %v", logMessagePrefix, req.Topic, err)
+		return r.channel, r.connectionCount, err
+	}
+	var expiration string
+	if ok {
+		// RabbitMQ expects the duration in ms
+		expiration = strconv.FormatInt(ttl.Milliseconds(), 10)
```
mukundansundar, line 204: Please check for negative TTL and a sane default must be set. 
Per documentation https://www.rabbitmq.com/ttl.html 
0 is expire immediately and negatives are not allowed ...

mukundansundar (changes requested):

pubsub/rabbitmq/rabbitmq.go
artursouza, line 204: This is close to what we need. Next step:
+ Write certification test to validate TTL feature.

artursouza (changes requested):

pubsub/rabbitmq/rabbitmq.go
Taction: `TryGetTTL` ensures the TTL is greater than 0(not included), and an empty `expiration` string will be ignored by the SDK when sent to the server.

mukundansundar: @Taction can you take a look into this?

Taction: I am currently looking into how to write the certification test. Since I have not written before, it will take me some time to finish that.

pubsub/rabbitmq/metadata.go
```diff
@@ -211,6 +222,12 @@ func (m *metadata) formatQueueDeclareArgs(origin amqp.Table) amqp.Table {
 		origin[argMaxLengthBytes] = m.maxLenBytes
 	}
 
+	if m.defaultQueueTTL != nil {
+		// Value in ms
+		ttl := *m.defaultQueueTTL / time.Millisecond
```
ItalyPaleAle, line 227: I don't understand why this is dividing by MS?

if `defaultQueueTTL` is in seconds, to get the number of milliseconds, shouldn't you do `ttl := *m.defaultQueueTTL * time.Millisecond`? Unless I misunderstand

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -190,10 +193,22 @@ func (r *rabbitMQ) publishSync(req *pubsub.PublishRequest) (rabbitMQChannelBroke
...
+	}
```
berndverst, line 205: There should be an `else` clause that sets `expiration` to `defaultQueueTTL` from the component metadata!

If there was no TTL specified on the request we should use the one on from the component metadata if that was specified. Only if neither were specified to we not set any expiration.

pubsub/rabbitmq/rabbitmq.go
berndverst: @Taction just add a test case here for TTL here:
https://github.com/dapr/components-contrib/tree/master/tests/certification/pubsub/rabbitmq

You need to create a new `component.yaml` file -- create a new subfolder in the components folder there and then reference that in your test flow.

You should verify several things:

- setting a default TTL at the component level expires messages correctly
- setting a TTL only at the message level but not component level expires messages correctly
- setting a TTL at the component level and message level ignores the default component level TTL and always uses the message level TTL specified

In all cases you should verify the message is gone after waiting the appropriate period of time.

berndverst (changes requested):

pubsub/rabbitmq/metadata.go
berndverst: You can literally call `*m.defaultQueueTTL.Milliseconds()`

pubsub/rabbitmq/rabbitmq.go
Taction: Thanks, this is a good point. In fact, this is related to [rabbitmq itself](https://www.rabbitmq.com/ttl.html):
A TTL can be specified on a  queue or per-message basis. When both a per-queue and a per-message TTL are specified, the lower value between the two will be chosen.

TTL from component metadata is set to rabbitmq queue, and here need not to handle it.

pubsub/rabbitmq/metadata.go
Taction: `m.defaultQueueTTL` is in time.Time format which is formatted when parsing.

> You can literally call *m.defaultQueueTTL.Milliseconds()

I will change to this, for it is easier to understand.

pubsub/rabbitmq/rabbitmq.go
Taction: After a review of this, rabbitmq will use the lower value when both are set. I will change it to what you suggested, in this way, TTL specified on the request takes priority over TTL specified on the component metadata.

pubsub/rabbitmq/metadata.go
Taction: done

pubsub/rabbitmq/rabbitmq.go
Taction: Done.

pubsub/rabbitmq/rabbitmq.go
Taction: done

Taction:
Certification tests passed, this pr is ready for review now.

yaron2:
@mukundansundar please re-review.

berndverst (approved):

Issue #2096, opened 2022-09-20
Azure config store - GetAll return outdated values


pravinpushkar:
GetAll returns older and outdated values.
Expectation  -  when keys are modified or deleted on azure config store, it should return the updated values.

Related discussions are found here also - https://github.com/dapr/components-contrib/issues/2061

See below logs - 
For update - 

➜  examples git:(feature/config_unsubscribe_api) ✗ curl -X GET http://localhost:3500/v1.0-alpha1/configuration/pravin-app-config-2\?key\=config1

**{"config1":{"value":"val111"}}**%
➜  examples git:(feature/config_unsubscribe_api) ✗ curl -X GET http://localhost:3500/v1.0-alpha1/configuration/pravin-app-config-2

**{"config1":{"value":"val1"}**,"config2":{"value":"val2"},"config3":{"value":"val3"},"config4":{"value":"1"},"sentinel-config1":{"value":"1"},"sentinel-config2":{"value":"1"}}%

For deleted keys - 
➜  examples git:(feature/config_unsubscribe_api) ✗ curl -X GET http://localhost:3500/v1.0-alpha1/configuration/pravin-app-config-2

{"config1":{"value":"val1"},"config2":{"value":"val2"},"**config3":{"value":"val3"}**,"sentinel-config1":{"value":"1"},"sentinel-config2":{"value":"1"}}%

➜  examples git:(feature/config_unsubscribe_api) ✗ curl -X GET http://localhost:3500/v1.0-alpha1/configuration/pravin-app-config-2**\?key\=config3**

{"errorCode":"ERR_CONFIGURATION_GET","message":"fail to get [config3] from Configuration store pravin-app-config-2: GET https://pravin-app-config-2.azconfig.io/kv/config3\n--------------------------------------------------------------------------------\nRESPONSE 404: 404 Not Found\

PR #2097, opened 2022-09-20
Fix getAll configuration Api for azure app config

pravinpushkar:

# Description

Azure app config getAll used to return older values that are either updated/deleted.
Changed the azure app config sdk API call to get the current values.

## Issue reference

Please reference the issue this PR will close: #2096 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


ItalyPaleAle (dismissed):

go.mod
```diff
@@ -156,7 +156,7 @@ require (
 require (
 	cloud.google.com/go/secretmanager v1.4.0
 	dubbo.apache.org/dubbo-go/v3 v3.0.3-0.20220610080020-48691a404537
-	github.com/Azure/azure-sdk-for-go/sdk/data/azappconfig v0.4.0
+	github.com/Azure/azure-sdk-for-go/sdk/data/azappconfig v0.4.1-0.20220914204640-4d445978fe75
```
berndverst, line 159: We should ask the SDK team to release 0.4.1 or 0.5.0. We should not pin a random commit, only tags.

We still have some time, please reach out to the team.

berndverst (changes requested):

go.mod
berndverst: Discussed in team chat - apparently the SDK team suggest using a direct commit in this case.

berndverst (approved):

Issue #2098, opened 2022-09-20
Support bulk subscribe in Azure Service Bus


shubham1172:
## Describe the feature
Support Bulk Subscribe API in Azure Service Bus.

New metadata to be added -

- MaxBulkSubCount: Maximum number of messages to fetch at once from the broker. If lesser messages are available, it will still fetch them without waiting to complete `MaxBulkSubCount`.

## Release Note

RELEASE NOTE: **ADD** Bulk subscribe support in Azure Service Bus

shubham1172:
Note, the timeout between messages in bulk is not configurable in the SDK yet, and the default is 20ms. This means that the SDK waits for 20ms after getting the first message before publishing the bulk of messages back.

Once it is configurable in the SDK, it can be made configurable here as well. Issue: [Azure/azure-sdk-for-go#19172](https://www.github.com/Azure/azure-sdk-for-go/issues/19172)

shubham1172:
Closed by #2098

PR #2099, opened 2022-09-20
Add bulk subscribe support to Azure Service Bus

shubham1172:
# Description

This PR allows bulk subscribe capabilities in Azure Service Bus.

## Issue reference

Please reference the issue this PR will close: #2098

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#2813

PR #2100, opened 2022-09-20
 Add bulk subscribe support to Azure Service Bus

shubham1172:
# Description

This PR allows bulk subscribe capabilities in Azure Service Bus.

Continuing this PR from https://github.com/dapr/components-contrib/pull/2099

New metadata added -
- **MaxBulkSubCount**: Maximum number of messages to fetch at once from the broker. If lesser messages are available, it will still fetch them without waiting to complete `MaxBulkSubCount`.

## Issue reference

Please reference the issue this PR will close: #2098

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: https://github.com/dapr/docs/issues/2813


shubham1172:
FYI, @ItalyPaleAle your reviews from the old PR has been addressed here, thanks for reviewing!

bindings/azure/servicebusqueues/servicebusqueues.go
```diff
@@ -268,7 +269,8 @@ func (a *AzureServiceBusQueues) Read(subscribeCtx context.Context, handler bindi
 }
 
 func (a *AzureServiceBusQueues) getHandlerFunc(handler bindings.Handler) impl.HandlerFunc {
-	return func(ctx context.Context, msg *servicebus.ReceivedMessage) error {
+	return func(ctx context.Context, asbMsgs []*servicebus.ReceivedMessage) ([]impl.HandlerResponseItem, error) {
+		msg := asbMsgs[0]
```
ItalyPaleAle, line 273: Should there be a check to make sure `asbMsgs` actually has 1 element?

pubsub/azure/servicebus/servicebus.go
```diff
@@ -360,6 +369,62 @@ func (a *azureServiceBus) Publish(req *pubsub.PublishRequest) error {
...
+		subscribeCtx,
+		a.metadata.MaxActiveMessages,
+		a.metadata.TimeoutInSec,
+		a.metadata.MaxRetriableErrorsPerSec,
+		a.metadata.MaxConcurrentHandlers,
+		"topic "+req.Topic,
+		a.logger,
+	)
+
+	receiveAndBlockFn := func(bo *backoff.ExponentialBackOff) error {
```
ItalyPaleAle, line 382: Nit: rather than passing `bo` as an argument and then creating `func() { bo.Reset()}`, pass directly a function (defined by the caller) that performs `bo.Reset()`

pubsub/azure/servicebus/servicebus.go
```diff
@@ -442,16 +490,54 @@ func (a *azureServiceBus) Subscribe(subscribeCtx context.Context, req pubsub.Sub
 }
 
 func (a *azureServiceBus) getHandlerFunc(topic string, handler pubsub.Handler) impl.HandlerFunc {
-	return func(ctx context.Context, asbMsg *servicebus.ReceivedMessage) error {
-		pubsubMsg, err := NewPubsubMessageFromASBMessage(asbMsg, topic)
+	emptyResponseItems := []impl.HandlerResponseItem{}
+	// Only the first ASB message is used in the actual handler invocation.
+	return func(ctx context.Context, asbMsgs []*servicebus.ReceivedMessage) ([]impl.HandlerResponseItem, error) {
+		pubsubMsg, err := NewPubsubMessageFromASBMessage(asbMsgs[0], topic)
```
ItalyPaleAle, line 496: Should check if `len(asbMsgs) > 0`?

internal/component/azure/servicebus/subscription.go
```diff
@@ -148,7 +199,7 @@ func (s *Subscription) ReceiveAndBlock(handler HandlerFunc, lockRenewalInSec int
 		}
```
ItalyPaleAle, line 199: (Leaving a comment here because I can't comment on L194)

There is a problem that when bulk subscribing is enabled, max active message doesn't work as expected. This is because we are taking a single entry from `s.activeMessageChan` for the entire set of bulk messages. So essentially rather than number of active messages, becomes number of active bulk messages. Not sure if that's intended.

internal/component/azure/servicebus/subscription.go
```diff
@@ -170,26 +221,84 @@ func (s *Subscription) ReceiveAndBlock(handler HandlerFunc, lockRenewalInSec int
...
+				// handle the message. The message will eventually be retried until fixed.
+				s.logger.Errorf("Error adding message: %s", err.Error())
+				<-s.activeMessagesChan
+				continue
+			}
+
+			s.logger.Debugf("Processing received message: %s", msg.MessageID)
+		}
+
+		runHandlerFn := func() {
```
ItalyPaleAle, line 245: Need to accept a `ctx` that is then passed to `handler(ctx, msgs)`

internal/component/azure/servicebus/subscription.go
```diff
@@ -170,26 +221,84 @@ func (s *Subscription) ReceiveAndBlock(handler HandlerFunc, lockRenewalInSec int
...
+				s.AbandonMessage(finalizeCtx, msg)
+				return
+			}
+
+			s.CompleteMessage(finalizeCtx, msg)
 		}
 
-		s.logger.Debugf("Processing received message: %s", msg.MessageID)
-		s.handleAsync(s.ctx, msg, handler)
+		bulkRunHandlerFunc := func() {
```
ItalyPaleAle, line 267: Need to accept a `ctx` that is then passed to `handler(ctx, msgs)`

internal/component/azure/servicebus/subscription.go
```diff
@@ -204,69 +313,60 @@ func (s *Subscription) Close(closeCtx context.Context) {
...
-		if err != nil {
-			// Log the error only, as we're running asynchronously
-			s.logger.Errorf("App handler returned an error for message %s on %s: %s", msg.MessageID, s.entity, err)
-			s.AbandonMessage(finalizeCtx, msg)
-			return
-		}
-
-		s.CompleteMessage(finalizeCtx, msg)
+		// Invoke the handler to process the message.
+		runHandlerFn()
```
ItalyPaleAle, line 369: Pass `ctx` to this

internal/component/azure/servicebus/subscription.go
```diff
@@ -204,69 +313,60 @@ func (s *Subscription) Close(closeCtx context.Context) {
...
+					before := time.Now()
+					_ = s.retriableErrLimit.Take()
+					s.logger.Debugf("Resumed after pausing for %v", time.Since(before))
+				}
+
+				// Remove the message from the map of active ones
+				s.removeActiveMessage(msg.MessageID, *msg.SequenceNumber)
+
+				// Remove an entry from activeMessageChan to allow processing more messages
+				<-s.activeMessagesChan
```
ItalyPaleAle, line 347: Here's where the issue I was mentioning before happens. You are here inside a loop for each message, and this may cause issues.

ItalyPaleAle (changes requested):
There are some issues now with handling concurrency with `maxActiveMessages` due to the fact that now you pass more than 1 message to the handler, when it was expecting only one. I _believe_ you now have a situation where `s.activeMessageChan` in `subscription.go` is drained more quickly than it is populated. In fact, with bulk messages, it is populated once per each set (bulk set), and drained once per every message

berndverst:
Fix linter issues

```
pubsub/azure/servicebus/servicebus.go:427     gofumpt     File is not `gofumpt`-ed
pubsub/azure/servicebus/message_test.go:42:2  gosec       G101: Potential hardcoded credentials
pubsub/azure/servicebus/message_test.go:46:2  stylecheck  ST1003: var testSampleTimeHttpFormat should be testSampleTimeHTTPFormat
```

internal/component/azure/servicebus/subscription.go
shubham1172: Hello @ItalyPaleAle, thanks for catching this, this has been updated now. 

Instead of `activeMessagesChan`, it's now `activeOperationsChan`. The channel is now used to control concurrent active operations, rather than messages. The size of `activeOperationsChan` is calculated as `maxActiveMessages/maxBulkCount`, where `maxBulkCount` can be `[1, maxActiveMessages]`

In case of normal subscribe, `maxBulkCount` is 1, so `activeOperationsChan` has the size `maxActiveMessages` (same as before). Here, each operation is just one receive request. 

In case of bulk subscribe, if `maxBulkCount` is 100 and `maxActiveMessages` is 10000, `activeOperationsChan`'s size is 100.

Let me know what you think about this!

/cc @mukundansundar @DeepanshuA

internal/component/azure/servicebus/subscription.go
```diff
@@ -168,28 +226,90 @@ func (s *Subscription) ReceiveAndBlock(handler HandlerFunc, lockRenewalInSec int
...
+			finalizeCtx, finalizeCancel := context.WithTimeout(context.Background(), s.timeout)
+			defer finalizeCancel()
+
+			if err != nil {
+				// Handle the error and mark messages accordingly.
+				// Note, the order of the responses match the order of the messages.
+				for i, resp := range resps {
+					if resp.Error != nil {
+						// Log the error only, as we're running asynchronously.
+						s.logger.Errorf("App handler returned an error for message %s on %s: %s", msgs[i].MessageID, s.entity, err)
```
ItalyPaleAle, line 293: May be wrong, but shouldn't `err` be `resp.Error` here?

internal/utils/utils.go
```diff
@@ -14,3 +15,14 @@ func IsTruthy(val string) bool {
 		return false
 	}
 }
+
+// GetIntOrDefFromMap returns the value of a key in a map as an int,
+// or a default value if the key is not present or a valid int.
+func GetIntOrDefFromMap(m map[string]string, key string, def int) int {
```
ItalyPaleAle, line 21: Nit: `GetIntOrDefaultFromMap`

internal/component/azure/servicebus/subscription.go
ItalyPaleAle: Thanks for making this change @shubham1172 ! Looks good to me now :)

ItalyPaleAle (changes requested):
Just a few more smaller changes please!

internal/utils/utils.go
ItalyPaleAle, line 21: Also, in the other PR (https://github.dev/shubham1172/components-contrib/blob/68704c23e04333e6790ba0924d1c24f83cd1113c/internal/utils/utils.go#L21-L22) there's an `GetUint64OrDefFromMap`... Can the two be implemented just once, with a generic?

internal/component/azure/servicebus/subscription.go
shubham1172: Yup, thanks!

internal/utils/utils.go
shubham1172: Replaced with generics implementation.

ItalyPaleAle (approved):

ItalyPaleAle (approved):

ItalyPaleAle (approved):

PR #2103, opened 2022-09-21
Adding conformance for bulk publish and properly map kafka errors on bulk publish

mukundansundar:
# Description

Adding conformance for bulk publish. 
This is currently using the existing subscriber to subscribe to the bulk published events as well.

For the mapping of errors from kafka, I was able to manually modify the bulk publish conf test locally and simulate a partial failure for a single message. I did modify the max message size for producer also locally and made the single failure message larger than the accepted size. 

The BulkPublishResponse got from BulkPublish is as following 
```
{[{11 FAILED kafka server: Message was too large, server rejected it to avoid allocation error.} {12 SUCCESS <nil>} {13 SUCCESS <nil>} {14 SUCCESS <nil>} {15 SUCCESS <nil>} {16 SUCCESS <nil>} {17 SUCCESS <nil>} {18 SUCCESS <nil>} {19 SUCCESS <nil>} {20 SUCCESS <nil>}]}
```
Here 11 -> 20 are the entry IDs and the 11 message alone failed with message was too large error.
The overall error was `kafka: Failed to deliver 1 messages.`

This I believe should only be part of certification test later on for kafka as it is too specific to kafka (as in modifying producer message size and simulating partial failure on publish).

Later on will figure out a way to add it to conformance tests.

## Issue reference

related to https://github.com/dapr/dapr/issues/2218

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -22,7 +22,8 @@ on:
   pull_request:
     branches:
       - master
-      - release-*
+      - 'release-*'
+      - 'feature/*'
 
 jobs:
   # Based on whether this is a PR or a scheduled run, we will run a different
--- a/tests/config/pubsub/tests.yml
+++ b/tests/config/pubsub/tests.yml
@@ -1,4 +1,5 @@
-# Supported operation: publish, subscribe
+# Supported operation: publish, subscribe, multiplehandlers, bulkpublish
+# bulkpublish is only run for components that implement pubsub.BulkPublisher interface
 # Config map:
 ## pubsubName : name of the pubsub
 ## testTopicName: name of the test topic to use
--- a/tests/config/pubsub/tests.yml
+++ b/tests/config/pubsub/tests.yml
@@ -1,5 +1,5 @@
 # Supported operation: publish, subscribe, multiplehandlers, bulkpublish
-# bulkpublish is only run for components that implement pubsub.BulkPublisher interface
+# bulkpublish should only be run for components that implement pubsub.BulkPublisher interface
 # Config map:
 ## pubsubName : name of the pubsub
 ## testTopicName: name of the test topic to use
@@ -11,7 +11,7 @@
 componentType: pubsub
 components:
   - component: azure.eventhubs
-    allOperations: true
+    operations: ["publish", "subscribe", "multiplehandlers"]
     config:
       pubsubName: azure-eventhubs
       testTopicName: eventhubs-pubsub-topic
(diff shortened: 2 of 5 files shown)


.github/workflows/conformance.yml
```diff
@@ -22,7 +22,8 @@ on:
   pull_request:
     branches:
       - master
-      - release-*
+      - 'release-*'
+      - 'feature/*'
```
mukundansundar, line 26: @dapr/maintainers-components-contrib Please let me know if this enable is OK ... This will only run PR conformance tests for PRs targeting feature branch also ...

berndverst: I suppose this is fine as it will only run the Docker bases conformance tests. Everything else requires manual `/ok-to-test`. Not all Azure tests perform well with multiple runs in parallel.

tests/config/pubsub/tests.yml
```diff
@@ -1,4 +1,5 @@
-# Supported operation: publish, subscribe
+# Supported operation: publish, subscribe, multiplehandlers, bulkpublish
+# bulkpublish is only run for components that implement pubsub.BulkPublisher interface
```
berndverst, line 2: don't you need to update this file and set `allOperations: false` and list the explicit operations everywhere where you don't have bulk publish?

berndverst: It's not clear how `bulkpublish` is added to the list of operations run by `allOperations: true`. Additionally, all operations do need to be listed explicitly where this test should not be run.

mukundansundar: I agree it's not clear.... 
What I am doing now is that I am skipping the tests for those components that are not implementing this interface.... I think this can lead to cases where silently tests are skipped and not run.... 
I will instead change to scenario where when this operation is specified and the component does not implement the interface,the conf test actually fails.... 

Thanks for noticing this @berndverst

mukundansundar: Done ....

tests/config/pubsub/tests.yml
```diff
@@ -10,7 +11,7 @@
 componentType: pubsub
 components:
   - component: azure.eventhubs
-    allOperations: true
+    operations: ["publish", "subscribe", "multiplehandlers"]
```
berndverst, line 14: Thanks for doing this.

berndverst (approved):

PR #2112, opened 2022-09-22
sanitize createBlob request metadata in AzStorageBlob

berndverst:

# Description

Sanitize blob creation request metadata in Azure Storage Blob Binding


bindings/azure/blobstorage/blobstorage.go
```diff
@@ -462,3 +463,30 @@ func (a *AzureBlobStorage) isValidDeleteSnapshotsOptionType(accessType azblob.De
...
+			delete(metadata, oldkey)
+			metadata[key] = val
+		}
+
+		// Remove all non-ascii characters and replace all colons with underscores
+		metadata[key] = strings.Map(func(r rune) rune {
+			if r > unicode.MaxASCII {
+				return -1
+			}
+			return r
```
shubham1172, line 487: "Replace colon with underscore" needs to be added.

berndverst: No the comment needs to be updated. I used to have this before, but it isn't needed :)

shubham1172: Sure, let's update the `special_value` test as well then.

berndverst: Done.

shubham1172 (dismissed):

mukundansundar (approved):
Lgtm

PR #2113, opened 2022-09-22
Conf test for Bulk Subscribe and Kafka fixes

DeepanshuA:
# Description

Intends to add Conf Tests for Bulk Subscribe Implementation.
While writing bulkSubscribe Conf tests, found out 2 issues with Kafka bukSubscribe implementation, so fixing those with this PR itself:
1. There was an issue while concurrently subscribing on one topic and bulkSubscribing on another topic.
2. Issue was there in ConsumeClaim where-in for bulkSubscribe topic messages, old method processBulkMessages was returning and thus ConsumeClaim was called again, it led to failure of Conf Test for PubSub where-in few messages (1, 2 and 5 out of 10 messages are failed for few times).

These issues are fixed.

## Issue reference

Please reference the issue this PR is linked to: https://github.com/dapr/dapr/issues/2218

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


berndverst:
@ItalyPaleAle could you review this for us please.

bindings/kafka/kafka.go
```diff
@@ -89,9 +89,12 @@ func (b *Binding) Read(ctx context.Context, handler bindings.Handler) error {
 		return nil
 	}
 
-	ah := adaptHandler(handler)
+	handlerConfig := kafka.SubscriptionHandlerConfig{
+		IsBulkSubscribe: false,
```
ItalyPaleAle, line 93: Nit: is this necessary, since `false` is the zero value? (although adding it for clarity may help)

internal/component/kafka/consumer.go
```diff
@@ -32,60 +32,63 @@ type consumer struct {
...
+		ticker := time.NewTicker(time.Duration(handlerConfig.SubscribeConfig.MaxBulkAwaitDurationMilliSeconds) * time.Millisecond)
+		defer ticker.Stop()
+		messages := make([]*sarama.ConsumerMessage, 0, handlerConfig.SubscribeConfig.MaxBulkCount)
+		for {
+			select {
+			case <-session.Context().Done():
+				return consumer.flushBulkMessages(claim, messages, session, handlerConfig.BulkHandler, b)
+			case message := <-claim.Messages():
+				if message != nil {
+					consumer.mutex.Lock()
```
ItalyPaleAle, line 56: Shouldn't the lock be placed outside of the `if message != nil` block?

internal/component/kafka/consumer.go
```diff
@@ -32,60 +32,63 @@ type consumer struct {
...
+					if len(messages) >= handlerConfig.SubscribeConfig.MaxBulkCount {
+						consumer.flushBulkMessages(claim, messages, session, handlerConfig.BulkHandler, b)
+						messages = make([]*sarama.ConsumerMessage, 0, handlerConfig.SubscribeConfig.MaxBulkCount)
+					}
+					consumer.mutex.Unlock()
+				}
+			case <-ticker.C:
+				consumer.mutex.Lock()
+				consumer.flushBulkMessages(claim, messages, session, handlerConfig.BulkHandler, b)
+				messages = make([]*sarama.ConsumerMessage, 0, handlerConfig.SubscribeConfig.MaxBulkCount)
```
ItalyPaleAle, line 67: Here too, just reset the slice without allocating new memory if possible.

internal/component/kafka/consumer.go
```diff
@@ -32,60 +32,63 @@ type consumer struct {
...
+			select {
+			case <-session.Context().Done():
+				return consumer.flushBulkMessages(claim, messages, session, handlerConfig.BulkHandler, b)
+			case message := <-claim.Messages():
+				if message != nil {
+					consumer.mutex.Lock()
+					messages = append(messages, message)
+					if len(messages) >= handlerConfig.SubscribeConfig.MaxBulkCount {
+						consumer.flushBulkMessages(claim, messages, session, handlerConfig.BulkHandler, b)
+						messages = make([]*sarama.ConsumerMessage, 0, handlerConfig.SubscribeConfig.MaxBulkCount)
```
ItalyPaleAle, line 60: No need to re-allocate this. Instead, just do: `messages = messages[:0]` which keeps the allocated memory. https://yourbasic.org/golang/clear-slice/

internal/component/kafka/consumer.go
```diff
@@ -190,62 +197,40 @@ func (consumer *consumer) Setup(sarama.ConsumerGroupSession) error {
...
-	k.subscribeLock.Lock()
-	delete(k.subscribeTopics, topic)
-	k.subscribeLock.Unlock()
-}
-
-// AddTopicBulkHandler adds a bulk handler and configuration for a topic
-func (k *Kafka) AddTopicBulkHandler(topic string, handlerConfig BulkSubscriptionHandlerConfig) {
+// AddTopicHandler adds a bulk handler and configuration for a topic
+func (k *Kafka) AddTopicHandler(topic string, handlerConfig SubscriptionHandlerConfig) {
 	k.bulkSubscribeLock.Lock()
```
ItalyPaleAle, line 202: Do we need a `bulkSubscribeLock`, or could `subscribeLock` be used instead? Since you're modifying the same `subscribeTopics` map

internal/component/kafka/utils.go
```diff
@@ -62,24 +62,14 @@ func isValidPEM(val string) bool {
 	return block != nil
 }
 
-// Map of topics and their handlers
-type TopicHandlers map[string]EventHandler
+// // Map of topics and their handlers
+// type TopicHandlers map[string]EventHandler
```
ItalyPaleAle, line 66: Delete?

ItalyPaleAle (changes requested):
Small changes please!

bindings/kafka/kafka.go
DeepanshuA: Yeah, this is by default, but just want to extra clear here, so wanted to set.

internal/component/kafka/consumer.go
DeepanshuA: Actually, I thought that lock is basically required when accessing `messages` and not `message`. But, it is perfectly fine to be cautious here as its just an if check. Moved it out.

internal/component/kafka/consumer.go
DeepanshuA: Good catch, thanks, done

internal/component/kafka/consumer.go
DeepanshuA: Done.

internal/component/kafka/consumer.go
DeepanshuA: My bad, I should have removed it. Done.

internal/component/kafka/utils.go
DeepanshuA: Yeah, done.

ItalyPaleAle (approved):
LGTM! 🚀

Issue #1719, opened 2022-05-12
Eliminate the existing inconsistency in the declaration of connection metadata between components


sleipnir:
## Describe the feature
I think there is a very big inconsistency between the metadata of each pubsub component. For example the pubsub component for redis allows the host, user and password properties to be set separately, which is good when we deploy in environments like kubernetes where we use secrets to manage sensitive properties.

RabbitMQ for example has an Operator that allows the credentials of clusters created via Operator to be stored in a kubernetes secret, this secret in turn has separate user, password, and host keys, and does not have a full url property. . That is, I can't use Dapr's pubsub component with RabbitMQ and reuse this secret automatically generated by RabbitMQ's Operator simply because the Dapr component doesn't follow the same pattern as the Redis component.

I believe it would be the component's responsibility to assemble the correct urls based on the separate values ​​of host, user and password. That way it would be much better to manage these properties securely with Kubernetes.

Currently I need to manually generate a new secret with a url property that connects these parts. But this is bad because since this process is manual I cannot guarantee that one day this process will fail. Being that if I could reuse the secret generated by the Operator of RabbitMQ I would simply trust my automation to generate and manage these properties, without me having to save this secret in git (we use gitops in our operation) or simply create it manually in the cluster .

I believe all components should declare host, user and password properties separately instead a full qualified connection string url

## Release Note

RELEASE NOTE:

sleipnir:
Ping

sleipnir:
Anyone with an answer here?

PR #2129, opened 2022-09-23
Added username and password arguments to pubsub.rabbitmq metadata

marioreggiori:
# Description

Added `username` and `password` arguments to pubsub.rabbitmq metadata and provide it to `amqp.Dial`.

**Caution:** This is probably a breaking change because the `host` argument was missused to also include credentials.

## Issue reference

Please reference the issue this PR will close: #1719

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/pubsub/rabbitmq/metadata.go
+++ b/pubsub/rabbitmq/metadata.go
@@ -15,17 +15,23 @@ package rabbitmq
 
 import (
 	"fmt"
+	"net/url"
 	"strconv"
 	"time"
 
 	amqp "github.com/rabbitmq/amqp091-go"
 
 	"github.com/dapr/components-contrib/pubsub"
+	"github.com/dapr/kit/logger"
 )
 
 type metadata struct {
 	consumerID       string
-	host             string
+	connectionString string
+	protocol         string
+	hostname         string
+	username         string
+	password         string
 	durable          bool
 	enableDeadLetter bool
 	deleteWhenUnused bool
@@ -42,8 +48,16 @@ type metadata struct {
 }
 
 const (
-	metadataConsumerIDKey           = "consumerID"
-	metadataHostKey                 = "host"
+	metadataConsumerIDKey = "consumerID"
+
+	metadataConnectionStringKey = "connectionString"
+	metadataHostKey             = "host"
+
+	metadataProtocolKey = "protocol"
+	metadataHostnameKey = "hostname"
+	metadataUsernameKey = "username"
+	metadataPasswordKey = "password"
+
 	metadataDurableKey              = "durable"
 	metadataEnableDeadLetterKey     = "enableDeadLetter"
 	metadataDeleteWhenUnusedKey     = "deletedWhenUnused"
@@ -61,8 +75,10 @@ const (
 )
 
 // createMetadata creates a new instance from the pubsub metadata.
-func createMetadata(pubSubMetadata pubsub.Metadata) (*metadata, error) {
+func createMetadata(pubSubMetadata pubsub.Metadata, log logger.Logger) (*metadata, error) {
 	result := metadata{
+		protocol:         "amqp",
+		hostname:         "localhost",
 		durable:          true,
 		deleteWhenUnused: true,
 		autoAck:          false,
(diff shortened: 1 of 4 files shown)


berndverst:
How about when username and password are empty we fallback to the existing misused behavior to avoid the breaking change?

Could you add that please?

marioreggiori:
We could omit username and password if the arguments haven't been provided via metadata. Unfortunately this is still a breaking change because we have to prepend the protocol `amqp://`.

marioreggiori:
Hi @ItalyPaleAle, the idea was to split the connection URI into multiple arguments to support secrets generated by the rabbitmq-operator (containing username and password separately). It appears that referencing other variables with `$(PREVIOUSLY_DEFINED_VAR)` as possible with Kubernetes envs from secrets does not work within dapr `Components`.

marioreggiori:
> I don't think there is any other protocol (like amqps) :)
Never mind.. I've should have read further :D

berndverst:
Linter failure

```
  Error: amqp_authority contains underscore. You should use mixedCap or MixedCap. (nosnakecase)
  Error: amqp_authority contains underscore. You should use mixedCap or MixedCap. (nosnakecase)
  Error: amqp_authority contains underscore. You should use mixedCap or MixedCap. (nosnakecase)
```

ItalyPaleAle:
@marioreggiori here's my proposal:

1. Create a new property "connectionString" which behaves like "host" does today.
1. Keep "host" as it is today and mark it as deprecated. It's an alias of "connectionString", taht will keep working but will not be in docs. If people use that, show a warning in the logs that they should change how they configure the component and use "connectionString" instead. Example of how we do that: https://github.com/dapr/dapr/blob/80f7e4ff9f478275d022cf2d0643028c36d43875/pkg/runtime/runtime.go#L2625-L2626
1. Create a new set of options: "hostname", "username", "password", "protocol". If there's no "connectionString", then construct it from reading from these.

berndverst:
FYI our code freeze for Dapr 1.9 (when PRs must have been reviewed, reviews sufficiently addressed, and the PR merged) is a week from today.

This PR might not make this next release. The sooner it is completed, the better the chance.

berndverst:
Please make sure to signoff on all of your commits too -- the DCO check is failing.

marioreggiori:
@berndverst I'm having difficulties rewording the previous commit which I forgot to sign-off without having to resolve conflicts. Can you please give me a hint on how to solve this?

pubsub/rabbitmq/metadata.go
```diff
@@ -76,10 +87,19 @@ func createMetadata(pubSubMetadata pubsub.Metadata) (*metadata, error) {
...
 
-	if val, found := pubSubMetadata.Properties[metadataHostKey]; found && val != "" {
-		result.host = val
-	} else {
-		return &result, fmt.Errorf("%s missing RabbitMQ host", errorMessagePrefix)
+	if val, found := pubSubMetadata.Properties[metadataConnectionStringKey]; found && val != "" {
+		result.connectionString = val
+	} else if val, found := pubSubMetadata.Properties[metadataHostKey]; found && val != "" {
+		result.connectionString = val
+		log.Warn("[DEPRECATION NOTICE] The 'host' argument is deprecated. Use 'connectionString' or individual connection arguments instead.")
```
ItalyPaleAle, line 94: Can you add a link to the Dapr docs too in the log messages, which we'll update? https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-rabbitmq/

pubsub/rabbitmq/metadata.go
```diff
@@ -184,3 +217,25 @@ func (m *metadata) formatQueueDeclareArgs(origin amqp.Table) amqp.Table {
...
+		if m.password != "" {
+			amqpAuthority.WriteString(":" + m.password)
+		}
+
+		amqpAuthority.WriteString("@")
+	}
+
+	amqpAuthority.WriteString(m.hostname)
+
+	return fmt.Sprintf("%s://%s", m.protocol, amqpAuthority.String())
```
ItalyPaleAle, line 240: I also wonder if we should use `url` to build this, which would also take care of correctly escaping sequences when needed? WDYT?

Something like:

```go
		u := url.URL{}
		u.Scheme = m.protocol
		u.Host = m.hostname
		if m.username != "" {
			u.User = ...
		}
		return u.String()
```

ItalyPaleAle (changes requested):
LGTM, just a couple of small comments :)

ItalyPaleAle:
>  I'm having difficulties rewording the previous commit which I forgot to sign-off without having to resolve conflicts. Can you please give me a hint on how to solve this?

Yeah, that's a bit of a pain in the bottom to solve. When I run into those situations, personally what I find easiest is to..... restart from scratch and re-apply the commits.

If you don't care about preserving the previous commits and are ok squashing them:

```sh
git reset --hard upstream/master
curl -L "https://patch-diff.githubusercontent.com/raw/dapr/components-contrib/pull/2129.diff" | git apply
```

Then commit with `git commit -s` as usual.

ItalyPaleAle (approved):

berndverst (approved):

shubham1172:
@marioreggiori @berndverst this seems to be missing in the docs, is there a docs PR/issue to track this?

berndverst:
I haven't seen anything. @marioreggiori did you open a PR against the docs repo? We reserve the right to revert any changes that are now documented.

marioreggiori:
@ItalyPaleAle said that you will take care of the docs. See https://github.com/dapr/components-contrib/pull/2129#discussion_r979124785

> We reserve the right to revert any changes that are not documented.

I currently don't have time to work on this.. So feel free to revert.

shubham1172:
Created https://github.com/dapr/docs/issues/3127 to track the docs changes.

PR #2151, opened 2022-10-05
Change hunter007 fork to latest go-sdk

yaron2:
Updates certification tests to use the latest Go-SDK version instead of a forked go-sdk version.

### DIFF
--- a/tests/certification/bindings/alicloud/dubbo/go.mod
+++ b/tests/certification/bindings/alicloud/dubbo/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/dapr/components-contrib v1.9.0-rc.1
 	github.com/dapr/components-contrib/tests/certification v0.0.0-20211026011813-36b75e9ae272
 	github.com/dapr/dapr v1.9.0-rc.3
-	github.com/dapr/go-sdk v1.4.0
+	github.com/dapr/go-sdk v1.5.1-0.20221004175845-b465b1fa0721
 	github.com/dapr/kit v0.0.3-0.20220930182601-272e358ba6a7
 	github.com/stretchr/testify v1.8.0
 )
@@ -164,5 +164,3 @@ require (
 replace github.com/dapr/components-contrib => ../../../../..
 
 replace github.com/dapr/components-contrib/tests/certification => ../../..
-
-replace github.com/dapr/go-sdk => github.com/hunter007/dapr-go-sdk v1.3.1-0.20220709114046-2f2dc4f9a684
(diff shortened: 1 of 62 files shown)


mukundansundar (approved):

berndverst (approved):

yaron2:
There's an issue now that we import proto files in the Go-SDK directly - we have proto namespace conflicts with packages coming from `dapr/dapr`. Info here: https://developers.google.com/protocol-buffers/docs/reference/go/faq#namespace-conflict.

I'm thinking about trying to add `GOLANG_PROTOBUF_REGISTRATION_CONFLICT=warn` to the test step. Thoughts @ItalyPaleAle @berndverst?

ItalyPaleAle:
@yaron2 what's the issue with failing tests? I don't understand where the conflict is

yaron2:
> @yaron2 what's the issue with failing tests? I don't understand where the conflict is

https://github.com/dapr/components-contrib/actions/runs/3191305895/jobs/5207465716

mukundansundar:
@yaron2 isn't this a breaking change from go-sdk? Can there be this scenario happening in non test env with a release of new version of go-sdk which imports its own protocol files ?

yaron2:
> @yaron2 isn't this a breaking change from go-sdk? Can there be this scenario happening in non test env with a release of new version of go-sdk which imports its own protocol files ?

No, I don't see how that's a breaking change.

mukundansundar:
> > @yaron2 isn't this a breaking change from go-sdk? Can there be this scenario happening in non test env with a release of new version of go-sdk which imports its own protocol files ?
> 
> 
> 
> No, I don't see how that's a breaking change.

I am just wondering whether except for this specific test scenario it is failing can there be other production scenarios where go sdk is used where this conflict may occur ?

yaron2:
> > > @yaron2 isn't this a breaking change from go-sdk? Can there be this scenario happening in non test env with a release of new version of go-sdk which imports its own protocol files ?
> > 
> > 
> > No, I don't see how that's a breaking change.
> 
> I am just wondering whether except for this specific test scenario it is failing can there be other production scenarios where go sdk is used where this conflict may occur ?

It's not a breaking change per-se, but if you have a Go binary that's importing the Go-SDK and `dapr/dapr` packages and you're using protobuf > 1.26, you'll get this error when you build your Go binary.

ItalyPaleAle:
Ok I see the error now. I agree with the proposed solution and also agree it's not a breaking change, unless someone is doing something really weird (and in that case they're on their own)

Issue #1999, opened 2022-08-23
[redis Configuration]After the dapr is started, the newly added key cannot be subscribed


saber-wang:
## Expected Behavior

If do not specify a key, you can subscribe to the newly added configuration item.

## Actual Behavior

Cannot subscribe to the newly added configuration item

## Steps to Reproduce the Problem

1. Do not specify key subscription configuration
```
using var client = new DaprClientBuilder().Build();
var _configurationResponse = await client.SubscribeConfiguration("configuration",new List<string>());
 await foreach (var items in _configurationResponse.Source.WithCancellation(cancellationToken))
     {
        _logger.LogInformation("{@items}", items);
     }
```
2. Add a key that does not exist in redis
3. Observation log (no output)

berndverst:
Sounds like whenever a key is not found there should be an attempt to refresh the list of known keys.

Which configuration store is this? Redis? This problem will be specific to the configuration store. (We have multiple configuration stores now)

saber-wang:
@berndverst  Yes, the current test is redis.

pravinpushkar:
Related issue is here in runtime - https://github.com/dapr/dapr/issues/5183
IMO the problem is in current impln for subscribe all in runtime. The solution may span both runtime and specific components.

PR #2157, opened 2022-10-06
fix redis subscribe all keys

pravinpushkar:

# Description

Using `PSubscribe` when a pattern is provided, else use `Subscribe`

## Issue reference

Please reference the issue this PR will close: #1999


ItalyPaleAle:
LGTM but while reviewing this, I noticed that on L279 there is an issue with concurrency. First, you check if the key exists in the map, and then you add it. Two concurrent goroutines could be writing at the same time. Can you change that to use an atomic operation LoadOrStore instead? https://github.dev/pravinpushkar/components-contrib/blob/443a9d6b1b5ae09c87a57b9a6916cb3f30ba173b/configuration/redis/redis.go#L279

ItalyPaleAle (approved):

berndverst (approved):

pravinpushkar:
> LGTM but while reviewing this, I noticed that on L279 there is an issue with concurrency. First, you check if the key exists in the map, and then you add it. Two concurrent goroutines could be writing at the same time. Can you change that to use an atomic operation LoadOrStore instead? https://github.dev/pravinpushkar/components-contrib/blob/443a9d6b1b5ae09c87a57b9a6916cb3f30ba173b/configuration/redis/redis.go#L279
> 
> (That can be done in a separate PR)

Right. Thanks for noticing.

PR #2165, opened 2022-10-08
optimize(contenttype): add contenttype ut

1046102779:
Signed-off-by: 1046102779 <seachen@tencent.com>

# Description

Add contenttype unit test and add grpc contenttype. And remove related codes in `dapr/dapr` repo.

[dapr/pkg/messaging/v1/utils.go:line 74](https://github.com/dapr/dapr/blob/master/pkg/messaging/v1/util.go#L74)
```golang
// IsJSONContentType returns true if contentType is the mime media type for JSON.
func IsJSONContentType(contentType string) bool {
        return strings.HasPrefix(strings.ToLower(contentType), JSONContentType)
}
```

[dapr/pkg/messaging/v1/utils.go:line 189](https://github.com/dapr/dapr/blob/master/pkg/messaging/v1/util.go#L189)
```golang
// IsGRPCProtocol checks if metadata is originated from gRPC API.
func IsGRPCProtocol(internalMD DaprInternalMetadata) bool {
        originContentType := ""
        if val, ok := internalMD[ContentTypeHeader]; ok {
                originContentType = val.Values[0]
        }
        return strings.HasPrefix(originContentType, GRPCContentType)
}
```
## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


daixiang0 (approved):

contenttype/contenttype.go
```diff
@@ -32,18 +38,23 @@ func IsJSONContentType(contentType string) bool {
 	return isContentType(contentType, JSONContentType)
 }
 
+// IsGRPCContentType checks for content type.
```
berndverst, line 41: Please remove this -- gRPC is not a content type but a transport type. There is no reason for us to have this method.

Everything else is fine, but the gRPC content type additions should be remove. Thanks.

berndverst (changes requested):

PR #2169, opened 2022-10-10
Azure Service Bus: consider receiving 0 messages as an error that requires reconnecting

ItalyPaleAle:
(In draft until we complete validation with @halspang in the long-haul cluster)

In the long-haul clusters, we have seen that the pubsub app sometimes enters into a weird state.

The only relevant log message is that it begins receiving payloads from Azure Service Bus with 0 messages. When that happens, something else gets messed up.

This is an attempt to fix the issue by causing the ASB components to force-reconnect to ASB when they receive 0 messages. This means changing the behavior to assume the connection is down.

### DIFF
--- a/internal/component/azure/servicebus/subscription.go
+++ b/internal/component/azure/servicebus/subscription.go
@@ -15,6 +15,7 @@ package servicebus
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


ItalyPaleAle:
As per discussion in the endgame call, we will merge this into 1.9 CC @yaron2 @berndverst 

Summary:

- The long-haul clusters were running a very old version of the sidecar (1.8.0-rc.4). Since then, this code hasn't changed much, but we have made other fixes in ASB and the SDK has had some minor version bumps. The error appeared after many weeks and we aren't sure what caused it.
- Since yesterday, our LH clusters are now running 2 versions of the pubsub test side-by-side: 1.9.0-rc.5 and 1.9.0-rc.5 + this patch. Both have been behaving fine for the last 24 hours, with no error.
- Because of the above, we don't really know if this fixes the issue. However, we do know that this does not cause negative impacts. It feels like it's a low-risk change which makes sense, and we believe it would be beneficial to ship this in 1.9 regardless, to be on the safe side.

yaron2 (approved):

PR #2181, opened 2022-10-12
Fix for handling "context deadline exceeded" on ASB publishing

ItalyPaleAle:
We are seeing more errors with Azure Service Bus pubsub in the long-haul tests, in which messages fail to be delivered and the component enters into a "broken" state:

```text
{"app_id":"pubsub-workflow","instance":"pubsub-workflow-app-664b8f4765-7fqcm","level":"debug","msg":"rpc error: code = Internal desc = error when publish to topic mediumtopic in pubsub longhaul-sb: context deadline exceeded","scope":"dapr.runtime.grpc.api","time":"2022-10-11T21:52:00.492233096Z","type":"log","ver":"1.9.0-rc.5"}
{"app_id":"pubsub-workflow","instance":"pubsub-workflow-app-664b8f4765-7fqcm","level":"debug","msg":"rpc error: code = Internal desc = error when publish to topic rapidtopic in pubsub longhaul-sb: context deadline exceeded","scope":"dapr.runtime.grpc.api","time":"2022-10-11T21:52:00.492232996Z","type":"log","ver":"1.9.0-rc.5"}
{"app_id":"pubsub-workflow","instance":"pubsub-workflow-app-664b8f4765-7fqcm","level":"debug","msg":"rpc error: code = Internal desc = error when publish to topic rapidtopic in pubsub longhaul-sb: context deadline exceeded","scope":"dapr.runtime.grpc.api","time":"2022-10-11T21:52:10.480573665Z","type":"log","ver":"1.9.0-rc.5"}
```

After the above, no more messages can be delivered and all invocations fail with the same error.

This seems to be caused by the *publish* connection to ASB entering into a "broken" state, where it's not dropped, but just "hanging". That's why the error is a "context deadline exceeded" (i.e. timeout - as we observed it uses the default timeout) and not something reported as "connection lost".

This PR changes the behavior of the component so in case of network errors (both "connlost" and "context deadline exceeded"), the sender is deleted so it can be re-created in the subsequent call, with a new connection. Hopefully this fixes the issues we've been observing.

PS: This is different from yesterday's fix (#2169) as that was on the receiving side. But both are similar in the sense that we're now more aggressively forcing a reconnection if there are certain errors.

### DIFF
--- a/bindings/azure/servicebusqueues/servicebusqueues.go
+++ b/bindings/azure/servicebusqueues/servicebusqueues.go
@@ -16,13 +16,13 @@ package servicebusqueues
 import (
 	"context"
 	"errors"
+	"fmt"
 	"sync"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 	servicebus "github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus"
 	sbadmin "github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/admin"
-	"github.com/Azure/go-amqp"
 	backoff "github.com/cenkalti/backoff/v4"
 
 	"github.com/dapr/components-contrib/bindings"
@@ -141,22 +141,10 @@ func (a *AzureServiceBusQueues) Operations() []bindings.OperationKind {
 	return []bindings.OperationKind{bindings.CreateOperation}
 }
 
-func (a *AzureServiceBusQueues) Invoke(ctx context.Context, req *bindings.InvokeRequest) (*bindings.InvokeResponse, error) {
-	var err error
-
-	a.senderLock.RLock()
-	sender := a.sender
-	a.senderLock.RUnlock()
-
-	if sender == nil {
-		a.senderLock.Lock()
-		sender, err = a.client.NewSender(a.metadata.QueueName, nil)
-		if err != nil {
-			a.senderLock.Unlock()
-			return nil, err
-		}
-		a.sender = sender
-		a.senderLock.Unlock()
+func (a *AzureServiceBusQueues) Invoke(invokeCtx context.Context, req *bindings.InvokeRequest) (*bindings.InvokeResponse, error) {
+	sender, err := a.getSender()
+	if err != nil {
+		return nil, fmt.Errorf("failed to create a sender for the Service Bus queue: %w", err)
 	}
 
 	msg := &servicebus.Message{
@@ -187,10 +175,19 @@ func (a *AzureServiceBusQueues) Invoke(ctx context.Context, req *bindings.Invoke
 		msg.TimeToLive = &ttl
 	}
 
-	ctx, cancel := context.WithTimeout(ctx, a.timeout)
-	defer cancel()
+	// Send the message
+	ctx, cancel := context.WithTimeout(invokeCtx, a.timeout)
+	err = sender.SendMessage(ctx, msg, nil)
+	cancel()
+	if err != nil {
+		if impl.IsNetworkError(err) {
+			// Force reconnection on next call
+			a.deleteSender()
+		}
+		return nil, err
+	}
 
-	return nil, sender.SendMessage(ctx, msg, nil)
+	return nil, nil
 }
 
 func (a *AzureServiceBusQueues) Read(subscribeCtx context.Context, handler bindings.Handler) error {
@@ -219,7 +216,7 @@ func (a *AzureServiceBusQueues) Read(subscribeCtx context.Context, handler bindi
 			})
 			if err != nil {
 				// Realistically, the only time we should get to this point is if the context was canceled, but let's log any other error we may get.
-				if err != context.Canceled {
+				if errors.Is(err, context.Canceled) {
 					a.logger.Warnf("Error reading from Azure Service Bus Queue binding: %s", err.Error())
 				}
 				return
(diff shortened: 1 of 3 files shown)


bindings/azure/servicebusqueues/servicebusqueues.go
```diff
@@ -187,10 +175,19 @@ func (a *AzureServiceBusQueues) Invoke(ctx context.Context, req *bindings.Invoke
 		msg.TimeToLive = &ttl
 	}
 
-	ctx, cancel := context.WithTimeout(ctx, a.timeout)
-	defer cancel()
+	// Send the message
+	ctx, cancel := context.WithTimeout(invokeCtx, a.timeout)
+	err = sender.SendMessage(ctx, msg, nil)
+	cancel()
```
artursouza, line 181: Why no more defer on cancel()?

ItalyPaleAle: Just my old habits of avoiding defers :) (defer used to be slow -  although it shouldn't be the case anymore).

bindings/azure/servicebusqueues/servicebusqueues.go
```diff
@@ -267,6 +257,48 @@ func (a *AzureServiceBusQueues) Read(subscribeCtx context.Context, handler bindi
...
+	// Check if the sender already exists
+	a.senderLock.RLock()
+	if a.sender != nil {
+		a.senderLock.RUnlock()
+		return a.sender, nil
+	}
+	a.senderLock.RUnlock()
+
+	// Acquire a write lock then try checking a.sender again in case another goroutine modified that in the meanwhile
+	a.senderLock.Lock()
```
mcandeia, line 271: nit: you can defer unlock here

bindings/azure/servicebusqueues/servicebusqueues.go
```diff
@@ -267,6 +257,48 @@ func (a *AzureServiceBusQueues) Read(subscribeCtx context.Context, handler bindi
...
+	return sender, nil
+}
+
+// deleteSender deletes the sender, closing the connection
+func (a *AzureServiceBusQueues) deleteSender() {
+	a.senderLock.Lock()
+	if a.sender != nil {
+		closeCtx, closeCancel := context.WithTimeout(context.Background(), time.Second)
+		_ = a.sender.Close(closeCtx)
+		closeCancel()
```
mcandeia, line 296: did you consider close it in background?

Given the fact that you're ignoring the error and the timeout is actually 1s (so we can have 1s lock here?)

ItalyPaleAle: A timeout is still needed or we could leak a goroutine if the connection just hanged forever

bindings/azure/servicebusqueues/servicebusqueues.go
```diff
@@ -267,6 +257,48 @@ func (a *AzureServiceBusQueues) Read(subscribeCtx context.Context, handler bindi
...
+	a.senderLock.Unlock()
+
+	return sender, nil
+}
+
+// deleteSender deletes the sender, closing the connection
+func (a *AzureServiceBusQueues) deleteSender() {
+	a.senderLock.Lock()
+	if a.sender != nil {
+		closeCtx, closeCancel := context.WithTimeout(context.Background(), time.Second)
```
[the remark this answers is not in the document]
ItalyPaleAle: No, the usage of `background` here is on purpose, with a timeout of 1s. `TODO` would indicate that this is a placeholder for future use.

pubsub/azure/servicebus/servicebus.go
```diff
@@ -486,6 +486,20 @@ func (a *azureServiceBus) senderForTopic(ctx context.Context, topic string) (*se
...
 }
 
+// deleteSenderForTopic deletes a sender for a topic, closing the connection
+func (a *azureServiceBus) deleteSenderForTopic(topic string) {
+	a.topicsLock.Lock()
+	defer a.topicsLock.Unlock()
+
+	sender, ok := a.topics[topic]
+	if ok && sender != nil {
+		closeCtx, closeCancel := context.WithTimeout(context.Background(), time.Second)
```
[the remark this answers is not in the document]
ItalyPaleAle: Same as above :)

bindings/azure/servicebusqueues/servicebusqueues.go
mcandeia: yes, you can keep the timeout being 1s though

ItalyPaleAle: Oh I misunderstood. I thought you meant just using `context.Background()` with no timeout.

I think this needs to happen in foreground. Reason is that after this method returns, the caller usually creates a new connection. It may be best to make sure the old one is actually closed before creating a new one.

bindings/azure/servicebusqueues/servicebusqueues.go
ItalyPaleAle: Yeah i should have

bindings/azure/servicebusqueues/servicebusqueues.go
ItalyPaleAle: Changed

bindings/azure/servicebusqueues/servicebusqueues.go
mcandeia: if you want to keep it in that way, so ideally you should receive a parent context and compose with the 1s timeout,

What I'm trying to say is that the `deleteSender` function can block for a second arbitrarily and its caller has no option to prevent this from happening

ItalyPaleAle: > What I'm trying to say is that the deleteSender function can block for a second arbitrarily and its caller has no option to prevent this from happening

Yup, that's by design. So there's some slowdown

artursouza (approved):

berndverst (approved):

Issue #2068, opened 2022-09-12
Add support for operations like "DELETE" and "GET" for Redis output binding


yash-nisar:
## Describe the feature

Currently, Redis only supports the [`CREATE` operation](https://github.com/dapr/components-contrib/blob/master/bindings/redis/redis.go#L75). We can perhaps add the following additional operations for Output Bindings as well:

`DELETE` operation: https://redis.io/commands/del/
`GET` operation: https://redis.io/commands/get/

Following this, the certification tests and docs should be updated as well.

## Release Note

RELEASE NOTE: **ADD** New feature in Dapr.

wxbty:
#2094 did this, maybe it should be separated from the rocketmq issue

PR #2186, opened 2022-10-13
Add general command capabilities to redis bindings

wxbty:
# Description

As pr #2068 said, redis binder lacks other operation commands besides set

## Issue reference

Please reference the issue this PR will close: #2068


ItalyPaleAle (changes requested):
Hi @wxbty thanks for the contribution!

I’m afraid I am not a fan of this approach however. I agree that the Redis binding needs lots of work and needs to support at least the operations described in #2068.

However, I don’t think that we should offer the ability to execute general commands. Three reasons:

1. First and foremost, security. Being  able to execute any command comes with a lot of power to do a lot of things, including things that developers shouldn’t be able to do.
2. Then, when we offer operations in bindings, we try to have them accept parameters in a consistent structure and return an optimized data structure. To do that, we need to know the exact operation that’s being executed
3. Lastly, the ability to perform general command goes against Dapr’s philosophy of abstracting components. Although bindings aren’t as abstract as, say, state stores, this really requires understanding the Redis commands and it’s not what Dapr aims to do.

If you make a PR to accept a subset of commands like in #2068, however, that would be more than welcome!

wxbty:
> Hi @wxbty thanks for the contribution!
> 
> I’m afraid I am not a fan of this approach however. I agree that the Redis binding needs lots of work and needs to support at least the operations described in #2068.
> 
> However, I don’t think that we should offer the ability to execute general commands. Three reasons:
> 
> 1. First and foremost, security. Being  able to execute any command comes with a lot of power to do a lot of things, including things that developers shouldn’t be able to do.
> 2. Then, when we offer operations in bindings, we try to have them accept parameters in a consistent structure and return an optimized data structure. To do that, we need to know the exact operation that’s being executed
> 3. Lastly, the ability to perform general command goes against Dapr’s philosophy of abstracting components. Although bindings aren’t as abstract as, say, state stores, this really requires understanding the Redis commands and it’s not what Dapr aims to do.
> 
> If you make a PR to accept a subset of commands like in #2068, however, that would be more than welcome!

Hi @ItalyPaleAle thanks for your reply!
I hope to continue contributing。As you said, there are still a few doubts。
1、Pr @2068 only mentioned “delete” and “get” ，do we not need other frequently used operations?
2、Now there is only one invoke method, do you want to add deleteInvoke and getInvoke methods? Or an invoke is distinguished by parameters

Issue #2056, opened 2022-09-07
Cron binding as Stable


sarsharma:
* The [Cron Binding](https://docs.dapr.io/reference/components-reference/supported-bindings/cron/) requires certification tests to be created.
* Add conformance tests if not present.

mukundansundar:
👋

nyemade-uversky:
@berndverst would you be able to double check on this and mark this as closed if the changes look good?

DeepanshuA:
Docs issue for change from Input/Output Binding to Input only Binding: https://github.com/dapr/docs/pull/2971

DeepanshuA:
> Docs PR for change from Input/Output Binding to Input only Binding: [dapr/docs#2971](https://github.com/dapr/docs/pull/2971)

berndverst:
This issue would have auto-closed if PR 2191 had not said "see 2056" but instead "fixes 2056" or "closes 2056".

PR #2191, opened 2022-10-14
Add certification tests for Cron Binding

### LINKED ISSUE #2308: Change Cron from input/output binding to only input binding component
Currently cron is an `input/output` binding component which supports `delete` operation.
This behaviour is unique to cron where the output binding is trying to stop the read operation of the input binding.
Cron is registered as both output and input binding in the dapr runtime
https://github.com/dapr/dapr/blob/410caa9a36725e43d0469b85e8f93958bdece3ce/cmd/daprd/components/binding_cron.go#L24-L29
The binding object is therefore different when calling Read() and Invoke() methods 
https://github.com/dapr/components-contrib/blob/ead9ffa665dd15474ab04392309789c7754aaf8d/bindings/cron/cron.go#L71
https://github.com/dapr/components-contrib/blob/ead9ffa665dd15474ab04392309789c7754aaf8d/bindings/cron/cron.go#L104

Cancelling ctx from Invoke() method would not cancel the context for the input binding.
Therefore, we would either need to use a global variable to communicate between the two binding objects using cron binding's name or change the way that `input/output` binding components are registered in the runtime.

Using global variable also introduces issue of having unique component names. https://github.com/dapr/components-contrib/pull/2191#discussion_r1018512436

For certifying cron as a stable component, we could change cron as an input binding only component for now.

sarsharma:
# Description

Adding certification test for cron binding.

## Issue reference

See #2056
Fixes #2308

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


tests/certification/bindings/cron/components/cron.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: Component
+metadata:
+  name: cron
+spec:
+  type: bindings.cron
+  version: v1
+  metadata:
+  - name: schedule
+  # test non-standard macro format
+    value: "@every 1s"
```
shubham1172, line 12:
```suggestion
    # test non-standard macro format
    value: "@every 1s"
```

tests/certification/bindings/cron/components/cron3s.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: Component
+metadata:
+  name: cron3s
+spec:
+  type: bindings.cron
+  version: v1
+  metadata:
+  - name: schedule
+  # test standard crontab format
+    value: "*/3 * * * * *"
```
shubham1172, line 12:
```suggestion
    # test standard crontab format
    value: "*/3 * * * * *"
```

shubham1172, line 12:
```suggestion
    # test standard crontab format
    value: "*/3 * * * * *"
```

tests/certification/bindings/cron/components/cron.yaml
shubham1172, line 12:
```suggestion
    # test non-standard macro format
    value: "@every 1s"
```

tests/certification/bindings/cron/config.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: dapr.io/v1alpha1
+kind: Configuration
+metadata:
+  name: cronbindingconfig
+spec:
+  features:
```
shubham1172, line 6: Did you intend to add something here?

tests/config/bindings/tests.yml
```diff
@@ -33,12 +33,14 @@ components:
     operations: ["create", "operations", "read"]
   - component: azure.servicebusqueues
     operations: ["create", "operations", "read"]
+  - component: cron
+    operations: ["operations", "read", "delete"]
```
shubham1172, line 37:
```suggestion
    operations: ["delete", operations", "read"]
```

tests/certification/bindings/cron/cron_test.go
```diff
@@ -0,0 +1,223 @@
...
+	flow.New(t, "test cron trigger schedule @every 3s with sidecar restart").
+		Step(app.Run("cronapp", fmt.Sprintf(":%d", appPort), application)).
+		Step(sidecar.Run("cron-sidecar",
+			embedded.WithComponentsPath("./components"),
+			embedded.WithDaprGRPCPort(grpcPort),
+			embedded.WithAppProtocol(runtime.HTTPProtocol, appPort),
+			embedded.WithDaprHTTPPort(httpPort),
+			componentRuntimeOptions(),
+		)).
+		Step("allow cron to trigger once", flow.Sleep(time.Second*4)).
```
shubham1172, line 192: Don't use magic numbers (applicable at multiple places).

tests/certification/bindings/cron/cron_test.go
```diff
@@ -0,0 +1,223 @@
...
+		Step(sidecar.Run("cron-sidecar",
+			embedded.WithComponentsPath("./components"),
+			embedded.WithDaprGRPCPort(grpcPort),
+			embedded.WithAppProtocol(runtime.HTTPProtocol, appPort),
+			embedded.WithDaprHTTPPort(httpPort),
+			componentRuntimeOptions(),
+		)).
+		Step("assert cron triggered within deadline", testAssert).
+		Step("test trigger delete operation", testDelete).
+		Step("stop sidecar", sidecar.Stop("cron-sidecar")).
```
shubham1172, line 97: Once deleted, should we also check if cron was not triggered again?

tests/certification/bindings/cron/cron_test.go
```diff
@@ -0,0 +1,223 @@
...
+	flow.New(t, "test cron trigger schedule @every 1s").
+		Step(app.Run("cronapp", fmt.Sprintf(":%d", appPort), application)).
+		Step(sidecar.Run("cron-sidecar",
+			embedded.WithComponentsPath("./components"),
+			embedded.WithDaprGRPCPort(grpcPort),
+			embedded.WithAppProtocol(runtime.HTTPProtocol, appPort),
+			embedded.WithDaprHTTPPort(httpPort),
+			componentRuntimeOptions(),
+		)).
+		Step("assert cron triggered within deadline", testAssert).
```
shubham1172, line 95: I am wondering if this will be flaky, since we need to test it in exactly 10 seconds. If it gets delayed by a bit and another cron is triggered, then assertion will fail. Increasing the trigger time can help with it.

tests/certification/bindings/cron/cron_test.go
mukundansundar: Yes we should.

tests/certification/bindings/cron/cron_test.go
mukundansundar: +1

tests/certification/bindings/cron/README.md
```diff
@@ -0,0 +1,16 @@
+# Cron certification testing
+
+This project aims to test the cron component under various conditions.
+
+## Test plan
+
+### Input Binding Test
+
```
mukundansundar, line 8: Please add more details in the Test Plan for each test.

tests/certification/bindings/cron/README.md
```diff
@@ -0,0 +1,16 @@
...
+* Test cron trigger with deadline limits
+* Test cron trigger with different schedules and cron formats
+* Test cron trigger on app restart
+* Test cron trigger on sidecar restart
+
```
mukundansundar, line 13: Add a test which tests same function being called from two different cron bindings on two different schedules.

tests/certification/bindings/cron/cron_test.go
```diff
@@ -0,0 +1,223 @@
...
+		triggerWatcher.ExpectInts(i)
+	}
+
+	// Test if cron triggers 5 times within 15 seconds
+	testAssert := func(ctx flow.Context) error {
+		triggerWatcher.Assert(ctx, time.Second*15)
+		return nil
+	}
+
+	application := func(ctx flow.Context, s common.Service) error {
```
DeepanshuA, line 120: `application` and other code pieces like `testAssert` are getting repeated in each func, can we rather take them out and just parametrize accordingly to return appropriate `application` or `testAssert`?

tests/certification/bindings/cron/cron_test.go
ItalyPaleAle: 100% agree. Actually, please consider replacing the timer in the Cron binding with one that allows faking time. Just measuring time is going to lead to very flaky tests (we see that all the time in the runtime). See: https://github.com/benbjohnson/clock
Then, in tests you can control time passing by with the mock clock

tests/certification/bindings/cron/config.yaml
berndverst: It's ok to keep it like this. Some other tests have this too.

tests/certification/bindings/cron/go.mod
```diff
@@ -0,0 +1,149 @@
...
+	sigs.k8s.io/json v0.0.0-20220713155537-f223a00ba0e2 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
+	sigs.k8s.io/yaml v1.3.0 // indirect
+)
+
+replace github.com/dapr/components-contrib/tests/certification => ../../
+
+replace github.com/dapr/components-contrib => ../../../../
+
+replace github.com/dapr/go-sdk => github.com/hunter007/dapr-go-sdk v1.3.1-0.20220709114046-2f2dc4f9a684
```
berndverst, line 144: I don't think we need this replace statement anymore

berndverst (changes requested):
Please address changes requested by @shubham1172 @ItalyPaleAle and @mukundansundar 

Not sure why these reviewers didn't request the changes officially themselves.

I would like to see explicit approval from them on this PR. Thanks!

tests/certification/bindings/cron/config.yaml
shubham1172: It's ok, but why keep it if it's not doing anything? 🤔

ItalyPaleAle: It you remove this, Dapr will crash. Don't ask Me why.

tests/certification/bindings/cron/cron_test.go
sarsharma: updated

tests/certification/bindings/cron/cron_test.go
sarsharma: This would require forking `github.com/robfig/cron/v3` used for cron binding
This PR was never merged https://github.com/robfig/cron/pull/327

tests/certification/bindings/cron/cron_test.go
sarsharma: The delete operation is broken for latest dapr.
This line never gets called when the runningCtx is cancelled
https://github.com/dapr/components-contrib/blob/ede060959a0a1deb1e7980aac6a4f5199b31e6a2/bindings/cron/cron.go#L91
While calling `b.Read()` and `b.Invoke()`, `b` is different since cron is both input/output binding.
Hence, reverting to using `stopCh map[string]chan bool` to signal ctx cancellation and stopping cron.

bindings/cron/cron.go
```diff
@@ -27,14 +27,14 @@ import (
...
-	parser        cron.Parser
-	runningCtx    context.Context
-	runningCancel context.CancelFunc
+	logger   logger.Logger
+	name     string
+	schedule string
+	parser   cron.Parser
 }
 
+var stopCh = make(map[string]chan bool)
```
ItalyPaleAle, line 36: Please keep this as a property of the object and not a global variable

bindings/cron/cron.go
```diff
@@ -27,14 +27,14 @@ import (
 
 // Binding represents Cron input binding.
 type Binding struct {
-	logger        logger.Logger
-	name          string
-	schedule      string
-	parser        cron.Parser
-	runningCtx    context.Context
```
ItalyPaleAle, line 34: Why getting rid of the context?

tests/certification/bindings/cron/cron_test.go
```diff
@@ -0,0 +1,266 @@
...
+	httpPort := ports[1]
+	appPort := ports[2]
+
+	cronName := "cron3s"
+	appName := "cronapp3s"
+	sidecarName := "cron-sidecar"
+
+	// Test if cron triggers 5 times within 15 seconds
+	expectedTriggerCount := 5
+	timeoutToObeserveTrigger := time.Second * 15
```
ItalyPaleAle, line 158: This is going to be very flaky as the wall clock isn't going to give you the accuracy you need. I am speaking by experience because I wrote some tests in Dapr/dapr that relied on the wall clock to measure if a background goroutine was working as intended..... and they were very flaky and had to rewrite them.

Please use a mock clock such as the clock library to mock the clock and advance it during your test execution: https://github.com/benbjohnson/clock

ItalyPaleAle (changes requested):

tests/certification/bindings/cron/cron_test.go
ItalyPaleAle: Sounds good to me. Let's fork it, since it's not actively maintained.

Looks like the code of the package is pretty short, so you may just import the code into our tree by creating a sub-package internal/cron or something like that (just make sure you report the original license correctly!)

bindings/cron/cron.go
sarsharma: https://github.com/dapr/components-contrib/pull/2191#discussion_r1016806187
It is because we cannot rely on context as property of the cron object itself

bindings/cron/cron.go
sarsharma: Same reason as above https://github.com/dapr/components-contrib/pull/2191#discussion_r1016806187

ItalyPaleAle: I see the comment now. I don’t believe the component is broken, however.

in the runtime, input and output bindings have a separate lifecycle.
Input bindings are stopped via cancelation of the context passes to the Read method.
Output bindings are stopped via invocation of the (optional) Close() method.

You shouldn’t use a single context (or cancellation token) for both. They are meant to have two separate cancellation tokens.

as for using a global variable, that’s going to be a problem when you initialize multiple instances of the component. We’re also going to start (very soon) doing work to allow components to be de-initialized dynamically, so we can’t use global variables.

mukundansundar: I believe that the intent for the output operation here is that the scheduled InputBinding for that component is stopped. 
Essentially OutputBinding operation controlling the execution of an InputBinding. In my opinion this should not be a valid scenario in the way dapr Bindings are defined, i.e. outputBinding should not affect inputBinding. 
Though it is a completely valid scenario for Cron.

ItalyPaleAle: I agree with @mukundansundar, this doesn't sound like the correct behavior. I don't believe there's another component which allows stopping (but interestingly, not restarting - so then you have no way to restart a cron except crashing the app) the Read operation using an invocation.

Also, the operation isn't even [clearly documented](https://docs.dapr.io/reference/components-reference/supported-bindings/cron/#binding-support).

My vote (if @berndverst agrees) would be to remove the capability and make the cron binding an input binding only. When we add support for dynamic components (it'll come - eventually) people can use that to start/stop cron tasks. And in the meanwhile, if developers want to ignore cron invocations, they can do that in their own app.

tests/certification/bindings/cron/cron_test.go
sarsharma: Forked cron library to use mock clock and placed the library under `internal/cron`
I am now using mock clock for the tests

bindings/cron/cron.go
sarsharma: @ItalyPaleAle I have changed cron to input binding only component for now.

ItalyPaleAle: Thanks! Please don't forget to update the docs

internal/cron/README.md
```diff
@@ -0,0 +1,125 @@
+[![GoDoc](http://godoc.org/github.com/robfig/cron?status.png)](http://godoc.org/github.com/robfig/cron)
```
ItalyPaleAle, line 1: Can you update this file and explain why it's here? Don't think we need the full README. For example:

```
# cron

This package contains a fork of [`github.com/robfig/cron/v3@v3.0.0`](https://github.com/robfig/cron/) which adds support for mocking the time.

See LICENSE for the license of the original package
```

bindings/cron/cron_test.go
```diff
@@ -64,61 +64,18 @@ func TestCronInitFailure(t *testing.T) {
...
-		i++
-		if i == testsNum {
-			cancel()
-		}
-
 		return nil, nil
 	})
-	time.Sleep(time.Duration(testsNum+3) * time.Second)
-	assert.Equal(t, testsNum, i)
+	time.Sleep(time.Second * 5)
```
ItalyPaleAle, line 77: This test is a place to use the mock clock too  :) Then you can test to make sure it ran exactly 5 times!

ItalyPaleAle (changes requested):

internal/cron/README.md
```diff
@@ -0,0 +1,19 @@
+# Cron
+
+This package is a fork of [`github.com/robfig/cron/v3@v3.0.0`](https://github.com/robfig/cron/) which adds support for mocking the time using [`github.com/benbjohnson/clock`](https://github.com/benbjohnson/clock)
```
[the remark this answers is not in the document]
sarsharma: @DeepanshuA  This fork is to allow the cron library being used in the cron binding to use the mock clock instead of the usual clock. It adds a cron.WithClock() method for the cron initialiser, to which we can pass a mock clock. 
If other tests only want to use a mock clock, they could directly use https://github.com/benbjohnson/clock by importing it.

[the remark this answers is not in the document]
ItalyPaleAle: I agree with @sarsharma

Also wonder if this should go into Dapr/kit however.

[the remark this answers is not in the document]
DeepanshuA: Got it. Saw previous comment. The required PR is still open there and nobody seems to be merging it for long.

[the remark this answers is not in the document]
DeepanshuA: Can we explicitly put this comment with a TODO marker in README, so that anyone looking at code may get to know the reason of forking in first glance. And, if PR https://github.com/robfig/cron/pull/327 gets merged in future (magically :) ), then any other dev may also try to cover that TODO.
I still feel it can be part of dapr/kit. Rest as Approvers/maintainers of contrib decide.

bindings/cron/cron_test.go
```diff
@@ -64,61 +69,19 @@ func TestCronInitFailure(t *testing.T) {
 
 // TestLongRead
 // go test -v -count=1 -timeout 15s -run TestLongRead ./bindings/cron/.
-func TestCronReadWithDeleteInvoke(t *testing.T) {
```
shubham1172, line 67: Aren't `TestCronInitSuccess` and `TestCronInitWithSeconds` the same? It's unrelated to this PR, but maybe we can follow table driven tests for check for different formats within a single `TestCronInit` method,

bindings/cron/cron_test.go
```diff
@@ -64,61 +69,19 @@ func TestCronInitFailure(t *testing.T) {
...
-			assert.Equal(t, schedule, scheduleVal)
-		}
-
 		return nil, nil
 	})
-	time.Sleep(time.Duration(testsNum+3) * time.Second)
-	assert.Equal(t, testsNum, i)
+	// Check if cron triggers 5 times in 5 seconds
+	clk.Add(5 * time.Second)
+	assert.Equal(t, i, 5, "Cron did not trigger expected number of times, expected 5, got %d", i)
```
shubham1172, line 85: Please also add a test to ensure that it stops once the context is cancelled.

tests/certification/bindings/cron/README.md
```diff
@@ -0,0 +1,21 @@
...
+* Test cron trigger with deadline limits  
+> Example: a cron trigger with a schedule `@every 1s` should trigger 10 times in the next 10 seconds
+* Test cron trigger with different schedules and cron formats  
+> Test cron trigger with schedule written in the standard crontab format `*/3 * * * * *` and another cron trigger with schedule written in non-standard macro format `@every 1s`
+* Test cron trigger on app restart
+> Check if cron will still trigger the app in case the app listening to cron trigger crashes and restarts
+* Test cron trigger on sidecar restart
+> Check if the app is still triggered by cron in case of dapr sidecar restart.
+
+### Output Binding Test
```
shubham1172, line 18: Clean up.

tests/certification/bindings/cron/cron_test.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/dapr/components-contrib/tests/certification/flow"
+	"github.com/dapr/components-contrib/tests/certification/flow/app"
+	"github.com/dapr/components-contrib/tests/certification/flow/sidecar"
+
+	bindings_loader "github.com/dapr/dapr/pkg/components/bindings"
+	dapr_testing "github.com/dapr/dapr/pkg/testing"
+)
+
+// Test cron trigger with most frequent schedule possible : @every 1s with a deadline
+func TestCronBindingFrequentTrigger(t *testing.T) {
```
shubham1172, line 42: Why do we need `FrequentTrigger` test? Should it be `TestCronBindingTrigger`? Also based on the test plan, do you want to add other schedules too?

shubham1172 (changes requested):
Added some comments.

internal/cron/README.md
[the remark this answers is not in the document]
sarsharma: The PR https://github.com/robfig/cron/pull/327 uses https://github.com/microsoft/clock which is an archived package 
and the PR seems unlikely to be merged :(  
With this internal fork, we are using https://github.com/benbjohnson/clock which seems to be better supported.
I am still adding a TODO comment in the README

tests/certification/bindings/cron/cron_test.go
sarsharma: adding a test matrix to test various cron formats and schedules

bindings/cron/cron_test.go
sarsharma: added!

bindings/cron/cron_test.go
sarsharma: removed these tests and replaced with a common test containing various cron format and schedules

sarsharma:
@shubham1172 @mukundansundar @ItalyPaleAle

tests/certification/bindings/cron/cron_test.go
```diff
@@ -0,0 +1,306 @@
...
+	appName := "cronapp"
+	sidecarName := "cron-sidecar"
+
+	testMatrix := []cronTest{
+		{
+			cronName:                 "cron1s",
+			schedule:                 "@every 1s", // Test macro cron format
+			expectedTriggerCount:     10,
+			timeoutToObserveTriggers: time.Second * 10,
+			startTime:                time.Now(),
```
shubham1172, line 62: nit: since it's a mock clock, maybe you can use a constant here like `time.UnixDate`.

tests/certification/bindings/cron/cron_test.go
```diff
@@ -0,0 +1,306 @@
...
+	return func(ctx flow.Context) error {
+		// allow up to 1 extra trigger to account for additional timeout(@schedule interval of cron trigger) provided in the tests
+		if *observedTriggerCount != expectedTriggerCount && *observedTriggerCount != expectedTriggerCount+1 {
+			t.Errorf("expected %d triggers, got %d", expectedTriggerCount, *observedTriggerCount)
+		}
+		return nil
+	}
+}
+
+func assertTestCase(t *testing.T, ct cronTest, observedTriggerCount *int) func(ctx flow.Context) error {
```
shubham1172, line 283: nit: Had to scroll down to look at the function implementation, please update the function name. It looks like the same implementation as `assertTriggerCount`, why not reuse it? You can update `ct.clk` in the main test code itself.

tests/certification/bindings/cron/cron_test.go
```diff
@@ -0,0 +1,306 @@
...
+	observedTriggerCount := 0
+	// total time for all triggers to be observed
+	timeoutToObserveTriggers := time.Second * 15
+
+	clk := clock.NewMock()
+
+	flow.New(t, "test cron bindings with different schedules and same route").
+		Step(app.Run(appName, fmt.Sprintf(":%d", appPort), appWithTriggerCounter(t, cronName, &observedTriggerCount))).
+		Step(sidecar.Run(sidecarName,
+			embedded.WithComponentsPath("./components2"),
```
shubham1172, line 146:
```suggestion
			embedded.WithComponentsPath("./components_sameroute"),
```

shubham1172 (approved):
Added a few nits @sarsharma, LGTM, thanks for incorporating all the changes!

tests/certification/bindings/cron/cron_test.go
shubham1172: Or something similar, anything more intuitive than `components2`.

tests/certification/bindings/cron/cron_test.go
sarsharma: using a constant time for all test cases

tests/certification/bindings/cron/cron_test.go
sarsharma: updated to reuse same function

shubham1172 (approved):
LGTM! Thanks @sarsharma 🥳

tests/certification/bindings/cron/README.md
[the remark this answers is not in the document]
sarsharma: this was already removed @mukundansundar, can you please try refreshing once..

ItalyPaleAle (dismissed):

shubham1172 (dismissed):

internal/cron/README.md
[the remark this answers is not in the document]
berndverst: I think it's very weird to have this fork inline in `internal/cron`. Can you make a PR against `dapr/kit` and move that there? You can make it `dapr/kit/cron`?

berndverst (changes requested):
Can you move the internal cron fork stuff into a PR against `dapr/kit` - create a new package `cron` there? Then import latest `dapr/kit` and use that package `cron`.

berndverst (approved):

PR #2193, opened 2022-10-14
Update more libraries

berndverst:
# Description

Update more libraries and clean up the `go.mod` files.

Came up with this handy command:
```bash
go list -mod=mod -f '{{if not .Indirect}}{{.}}{{end}}' -u -m all | grep "\[.*\]" | awk '{gsub("\\[|\\]","",$3); print $1 "@" $3}' | xargs go get
```

Now all `go.mod` files have exactly one `require` block for direct dependencies and one `require` block for indirect dependencies.


ItalyPaleAle (approved):
No idea what your command does (that's a lot of bash-fu there), but the go.mod LGTM :)

ItalyPaleAle (approved):

berndverst:
> No idea what your command does (that's a lot of bash-fu there), but the go.mod LGTM :)

It finds all upgradeable direct dependencies and their latest minor version. Then one by one runs `go get` to upgrade that dependency.

ItalyPaleAle (approved):

berndverst:
Not sure why postgresql is failing here but it seems consistent. At first I thought it was due to `github.com/jackc/pgx/v5/stdlib` but it doesn't seem to be occurring elsewhere.

berndverst:
The Postgresql State Certification Etag test was wrong. I fixed it. Not sure how this even worked up to this point. Maybe the underlying Etag behavior where subsequent etags were incremental changed. The test was erroneously making the assumption subsequent etags are sequential. (FYI @yaron2 :) )

Instead I added a step to fetch the the written key to get the latest Etag value.

ItalyPaleAle (approved):

ItalyPaleAle (approved):

ItalyPaleAle (approved):

PR #2211, opened 2022-10-21
New PubSub component: Azure Service Bus Queues

### LINKED ISSUE #2130: Azure Service Bus - Is it possible to use a "Queue" resource? ("error: failed to subscribe to topic")
I know that the "Pub/Sub" connector implies the use of topics, but does the connector architecture allow the use of a service bus Queue, or only Topics are supported? I checked the documentation but couldn't find this explicitly declared.

In any case, I leave it here for anyone that might get into this same issue when accidentally using a queue instead of a topic.

```
ERRO[0002] failed to subscribe to topic queue1: azure service bus error: could not get subscription subscriber, GET https://bus-microservices.servicebus.windows.net/queue1/Subscriptions/subscriber
--------------------------------------------------------------------------------
RESPONSE 400: 400 Bad Request
ERROR CODE: 400
--------------------------------------------------------------------------------
<Error><Code>400</Code><Detail>The specified HTTP verb (GET) is not valid. To know more visit https://aka.ms/sbResourceMgrExceptions. . TrackingId:e9ccd839-1aa3-4f75-978f-8c811a387e3a_G27, SystemTracker:bus-microservices:Queue:queue1, Timestamp:2022-09-24T18:44:17</Detail></Error>
--------------------------------------------------------------------------------  app_id=subscriber instance=DESKTOP-DP18TG3 scope=dapr.runtime type=log ver=1.8.4
```

`pubsub.yaml`

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: messages-pub-sub
spec:
  type: pubsub.azure.servicebus
  version: v1
  metadata:
  - name: connectionString
    value: "Endpoint=sb://{ServiceBusNamespace}.servicebus.windows.net/;SharedAccessKeyName={PolicyName};SharedAccessKey={Key};EntityPath={ServiceBus}"
  scopes:
    - publisher
    - subscriber
```

ItalyPaleAle:
# Description

Creates a new component for using Azure Service Bus Queues as pubsub (before it was only available as a binding). Will be called `pubsub.azure.servicebus.queues`.

This PR's intent is to rename the existing `pubsub.azure.servicebus` component to `pubsub.azure.servicebus.topics` (maintaining the old name as alias for backwards-compatibility).

I also added conformance tests for the new component, which are passing, but won't be able to run with `/ok-to-test` in this PR because they require a pipeline change and the pipeline always runs from master. Here's the log when I run them locally: https://pastebin.com/ywQnHc66 (too big for GitHub to post here)

## Issue reference

Fixes #2130

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Docs issue: https://github.com/dapr/docs/issues/2912

### DIFF
--- a/.build-tools/component-folders.json
+++ b/.build-tools/component-folders.json
@@ -22,6 +22,7 @@
         "configuration/redis/internal",
         "pubsub/aws",
         "pubsub/azure",
+        "pubsub/azure/servicebus",
         "pubsub/gcp",
         "secretstores/alicloud",
         "secretstores/aws",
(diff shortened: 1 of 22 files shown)


ItalyPaleAle:
All Service Bus conf tests are expected to fail because they need an updated pipeline in master :)

berndverst (approved):

PR #2215, opened 2022-10-22
Add new PubSub component - KubeMQ Kubernetes message broker

kubemq:
# Description

This PR for adding new PubSub component, KubeMQ , Kubernetes Message Broker

Conformance tests can be done by using KubeMQ community docker image:

 docker run -d -p 8080:8080 -p 50000:50000 -p 9090:9090 kubemq/kubemq-community:latest  

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


go.mod
```diff
@@ -3,364 +3,377 @@ module github.com/dapr/components-contrib
 go 1.19
 
 require (
-	cloud.google.com/go/datastore v1.8.0
```
ItalyPaleAle, line 6: This change to dependencies seems to be reverting a lot of packages to an older version. Could you please reset all go.mod files to the master branch, then just do a “go mod tidy”?

pubsub/kubemq/kubemq.go
```diff
@@ -0,0 +1,73 @@
...
+	logger           logger.Logger
+	ctx              context.Context
+	ctxCancel        context.CancelFunc
+	eventsClient     *kubeMQEvents
+	eventStoreClient *kubeMQEventStore
+}
+
+func NewKubeMQ(logger logger.Logger) pubsub.PubSub {
+	return &kubeMQ{
+		metadata:         nil,
```
ItalyPaleAle, line 20: Nit: no need to initialize variables as `nil`.

pubsub/kubemq/kubemq_events.go
```diff
@@ -0,0 +1,169 @@
+package kubemq
+
+import (
+	"context"
+	"github.com/dapr/components-contrib/pubsub"
+	"github.com/dapr/kit/logger"
+	"github.com/kubemq-io/kubemq-go"
+	"github.com/kubemq-io/kubemq-go/pkg/uuid"
```
ItalyPaleAle, line 8: Please use `google/uuid` which is what we use throughout Dapr. This package also has methods that return errors (`NewRandom()`) rather than panic in case of errors.

pubsub/kubemq/metadata.go
```diff
@@ -0,0 +1,85 @@
...
+			return nil, err
+		}
+	} else {
+		return nil, fmt.Errorf("invalid kubeMQ address, address is empty")
+	}
+	if val, found := pubSubMetadata.Properties["clientID"]; found && val != "" {
+		result.clientID = val
+	}
+
+	if val, found := pubSubMetadata.Properties["authToken"]; found && val != "" {
```
ItalyPaleAle, line 56: Other Dapr components read the Kubernetes auth token from the volume that is mounted by Kubernetes. I would recommend doing that rather than requiring an auth token passed as metadata - unless KubeMQ works differently

pubsub/kubemq/metadata.go
```diff
@@ -0,0 +1,85 @@
...
+		switch val {
+		case "false":
+			result.isStore = false
+		case "true":
+			result.isStore = true
+		default:
+			return nil, fmt.Errorf("invalid kubeMQ store value, store can be true or false")
+		}
+	}
+	if val, found := pubSubMetadata.Properties["useMock"]; found && val != "" {
```
ItalyPaleAle, line 74: What does this do?

pubsub/kubemq/kubemq_events.go
```diff
@@ -0,0 +1,169 @@
...
+	if err := k.setPublishStream(); err != nil {
+		k.logger.Errorf("error init kubemq client error: %w", err.Error())
+		return err
+	}
+	k.isInitialized = true
+	return nil
+}
+func (k *kubeMQEvents) Init(meta *metadata) error {
+	k.metadata = meta
+	_ = k.init()
```
ItalyPaleAle, line 75: Why is the error discarded?

pubsub/kubemq/kubemq_events.go
```diff
@@ -0,0 +1,169 @@
...
+	return nil
+}
+func (k *kubeMQEvents) setPublishStream() error {
+	var err error
+	k.publishFunc, err = k.client.Stream(k.ctx, func(err error) {
+		select {
+		case k.resultChan <- err:
+		default:
```
ItalyPaleAle, line 83: Why this? Wouldn't this cause  errors to be discarded?

pubsub/kubemq/kubemq_events.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+	})
+	return err
+}
+func (k *kubeMQEvents) Publish(req *pubsub.PublishRequest) error {
+	if !k.isInitialized {
```
ItalyPaleAle, line 89: This can lead to race conditions if multiple goroutines are calling `Publish` at the same time. Consider using a lock or something else to ensure this is safe for concurrent use.

pubsub/kubemq/kubemq_events.go
```diff
@@ -0,0 +1,169 @@
...
+		return err
+	}
+	return nil
+}
+func (k *kubeMQEvents) Features() []pubsub.Feature {
+	return nil
+}
+
+func (k *kubeMQEvents) Subscribe(ctx context.Context, req pubsub.SubscribeRequest, handler pubsub.Handler) error {
+	if !k.isInitialized {
```
ItalyPaleAle, line 114: Same - risk of race conditions

pubsub/kubemq/kubemq_events.go
```diff
@@ -0,0 +1,169 @@
...
+	err := k.client.Subscribe(ctx, &kubemq.EventsSubscription{
+		Channel:  req.Topic,
+		Group:    k.metadata.group,
+		ClientId: clientID,
+	}, func(event *kubemq.Event, err error) {
+		if err != nil {
+			k.logger.Errorf("kubemq pub/sub error: subscribing to %s failed with error: %s", req.Topic, err.Error())
+			return
+		}
+		if ctx.Err() != nil {
```
ItalyPaleAle, line 133: Wouldn't this be captured by the `if err != nil` above?

pubsub/kubemq/kubemq_eventstore.go
```diff
@@ -0,0 +1,199 @@
+package kubemq
```
ItalyPaleAle, line 1: On this file I have the same feedback I had on `kubemq_events`

ItalyPaleAle (changes requested):
Thanks for this component, looks exciting to have another interesting one!

There are a few things before we can get this merged. First, please take a look at the initial review comments.

Aside from that, our PRs need to have committees accept the DCO (https://docs.dapr.io/contributing/contributing-overview/#developer-certificate-of-origin-signing-your-work). Also, please make sure to run `make lint` before you commit too.

pubsub/kubemq/metadata.go
kubemq: This is the client side. if you want to provide a auth token, you should provide this.
it works little bit different.

pubsub/kubemq/metadata.go
kubemq: This is for mock client in UnitTests

pubsub/kubemq/kubemq_events.go
kubemq: This is because in events mode the SDK is never return error. this was done to be align with SDK signature

pubsub/kubemq/kubemq_events.go
kubemq: his is because in events mode the SDK is never return error. this was done to be align with SDK signature

pubsub/kubemq/kubemq_events.go
kubemq: Not always, a message can be retrieved and at the same time the ctx will be cancelled.
(It happened in conformance test in some runs)

go.mod
kubemq: Changed as requested

pubsub/kubemq/kubemq.go
kubemq: Changed as requested

pubsub/kubemq/kubemq_events.go
kubemq: Changed as requested

pubsub/kubemq/kubemq_events.go
kubemq: Changed as requested

pubsub/kubemq/kubemq_events.go
kubemq: Changed as requested

pubsub/kubemq/kubemq_eventstore.go
kubemq: Changed as requested for the file kubemq_events_store

pubsub/kubemq/metadata.go
ItalyPaleAle: Ok so this is not the same token that is used to access the K8s APIs?

pubsub/kubemq/metadata.go
ItalyPaleAle: For unit tests, we normally don't enable mocking through metadata options, as those are meant to be public. Right now, someone could create a component and set `useMock` to `true` and could break things.
You should be able to set some property directly in the object, rather than going through the metadata.

pubsub/kubemq/kubemq_events.go
ItalyPaleAle: Ok, as long as the error is logged

pubsub/kubemq/kubemq_events.go
ItalyPaleAle: I'm more interested in the fact that you either write the error to the channel or, if the channel is full, discard it. I'm not sure I understand the flow.

pubsub/kubemq/kubemq_events.go
ItalyPaleAle: Although this fixes the condition for reading/writing `k.isInitialized`, it doesn't fix the race condition when 2 goroutines are trying to initialize at the same time.

Imagine two goroutines, A and B, calling `Publish` (or `Subscribe`) at the exact same time.

- A loads `k.isInitialized` (atomically) and sees it's false, so calls `init()`
- B loads `k.isInitialized` (atomically) and sees it's false, so calls `init()`
- In both cases, the `init` function is executed till the end before it sets the boolean `k.isInitialized`, so you end up with race conditions there

I think you need to use locks here, so when A starts, it checks if there's a client, then gets a lock, and if that fails, it creates it. At the same time, B is trying to acquire the lock and would block until A is done.

Example is here: https://github.com/dapr/components-contrib/blob/ed745763a0e029f7d7294d867bada99bbcd09827/internal/component/azure/servicebus/client.go#L100-L126

Note that this uses RWMutex locks for performance, which also means that after switching from a read lock to a write loick, it needs to check if the client was created again.

pubsub/kubemq/kubemq_events.go
ItalyPaleAle: Sounds like a bug in the upstream SDK that should handle context cancellation :) But ok, I suppose.

pubsub/kubemq/metadata.go
kubemq: No, it's not. this is auth token to access KubeMQ server

pubsub/kubemq/kubemq_events.go
kubemq: Changed to RWMutex as requested

pubsub/kubemq/metadata.go
kubemq: Got it,
Removed

pubsub/kubemq/kubemq_events.go
kubemq: yes, bad error text

pubsub/kubemq/kubemq_events.go
kubemq: changed as requested to RWMutext

pubsub/kubemq/kubemq_events.go
kubemq: This is the in-memory Pub/Sub which not return any error when you stream to a channel ( legacy decision)

berndverst:
@kubemq any chance you can give Dapr maintainers access to update your PR? Otherwise it's tricky to coordinate keeping this PR up to date with master for merging.

Can you also please ensure the DCO check passes? https://docs.dapr.io/contributing/contributing-overview/#developer-certificate-of-origin-signing-your-work

berndverst (changes requested):
Please run `make modtidy-all` from the root of the contrib repo, and then commit all `go.mod` and `go.sum` changes.

Currently tests fail because the go.sum files aren't updated.

berndverst:
For the component conformance, please create a Docker compose file here:
`.github/infrastructure/`

Then update the workflow at `.github/workflows/conformance.yml`, just list `pubsub.kubemq` in this block:
https://github.com/dapr/components-contrib/blob/master/.github/workflows/conformance.yml#L72

And further below, please add the loading of your docker compose file:
https://github.com/dapr/components-contrib/blob/master/.github/workflows/conformance.yml#L300

Something like

```
    - name: Start kubemq
      run: docker-compose -f ./.github/infrastructure/docker-compose-kubemq.yml -p kubemq up -d
      if: contains(matrix.component, 'kubemq')
```

tests/config/pubsub/tests.yml
```diff
@@ -12,7 +10,7 @@
 componentType: pubsub
 components:
   - component: azure.eventhubs
-    operations: ["publish", "subscribe", "multiplehandlers", "bulkpublish"]
+    allOperations: true
```
berndverst, line 13: Please revert these changes. `allOperations` now includes additional operations that these components do not support, such as BulkPublish and BulkSubscribe.

All edits in this file must be reverted, except for your addition of KubeMQ

It seems you actually forgot to add an entry for `kubemq`, so the conformance tests would never actually run against `kubemq`!

berndverst (changes requested):

berndverst:
Not sure what's up with the linter failures, but I suspect those will actually go away after you run `make modtidy-all` and commit those changes go.mod/go.sum files (all of them).

Please do run `make lint` locally as well.

berndverst:
I opened a PR (linked above) to fix up parts of your PR.

Make sure the following passes:
- `make lint`
- `make test`
-  Run your conformance test

Some issues @kubemq:

You updated a commercetools dependency which broke the commerce tools binding. Please investigate and fix.

Please make sure the conformance test passes. Once you merge my PR, you can run the test via:
```
go test -v -tags=conftests -count=1 ./tests/conformance -run="TestPubsubConformance/kubemq"
```

you may find that "all operations" do not actually work, if so you need to explicitly list the supported operations in `tests/config/pubsub/tests.yml`

kubemq:
Hi @berndverst,
I have merged your PR (thank you !), added the conformance tests files.
Couple Issues:
1. make lint - fails on:
ERRO Running error: 1 error occurred:
        * can't run linter goanalysis_metalinter: inspect: failed to load package kubernetes: could not load export data: no export data for "k8s.io/client-go/kubernetes"
2. make test fails too on   github.com/dapr/components-contrib/bindings/kubernetes [build failed]
3.  conformance tests fails too :
/root/go/pkg/mod/k8s.io/client-go@v0.23.0/applyconfigurations/meta/v1/unstructured.go:64:38: cannot use doc (variable of type *"github.com/googleapis/gnostic/openapiv2".Document) as type *"github.com/google/gnostic/openapiv2".Document in argument to proto.NewOpenAPIData

Can you PR again the updated stream ?

kubemq:
Thank you @berndverst ,
all is working include conformance tests

berndverst (approved):

berndverst:
@kubemq what remains now is to create a documentation PR against the v1.10 branch of the docs to add this component. Without the docs PR nobody will find this component.

berndverst:
Docs tracking issue: https://github.com/dapr/docs/issues/2947

PR #2217, opened 2022-10-24
Pin postgres version in test infra

ItalyPaleAle:
Postgres releases a new major version every year or so, and they sometimes have backwards-incompatible changes. When Postgres 15 came out a few days ago, our cert tests started to fail right away (the bug was in our tests making assumptions they should not have made).

This pins postgres to a specific version in our test infra so we can upgrade at our own pace.

Also adds a (commented-out) service for deploying pgadmin, useful for local debugging.

### DIFF
--- a/.github/infrastructure/docker-compose-postgresql.yml
+++ b/.github/infrastructure/docker-compose-postgresql.yml
@@ -1,7 +1,7 @@
 version: '2'
 services:
   db:
-    image: postgres
+    image: postgres:15
     restart: always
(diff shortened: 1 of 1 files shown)


berndverst:
Don't you think it helps to know as an early warning sign that we are incompatible with a newer Postgres version?

If we pin a version we might fail to ensure compatibility with newer Postgres versions.

berndverst (approved):

PR #2218, opened 2022-10-24
Fix Azure Service Bus cert/conf tests

ItalyPaleAle:
After merging #2211 , some tests are now failing in master (that we couldn't check before because they required a pipeline merged into master). This fixes the issues introduced by that PR.


berndverst (dismissed):

berndverst (approved):

1046102779 (commented):
`test.log`  is  a ignore file?

berndverst:
> `test.log`  is  a ignore file?

It was accidentally committed, but will not be ignored.

ItalyPaleAle:
It's been deleted now

PR #2219, opened 2022-10-24
More fixes to ASB cert tests

ItalyPaleAle:
Last one (hopefully)

### DIFF
--- a/tests/certification/pubsub/azure/servicebus/topics/components/authentication/service_bus.yaml
+++ b/tests/certification/pubsub/azure/servicebus/topics/components/authentication/service_bus.yaml
@@ -4,7 +4,7 @@ metadata:
   name: messagebus
   namespace: default
 spec:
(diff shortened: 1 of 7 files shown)


berndverst (approved):

tests/certification/pubsub/azure/servicebus/topics/components/authentication/service_bus.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: messagebus
   namespace: default
 spec:
-  type: pubsub.azure.servicebus
+  type: pubsub.azure.servicebus.topics
```
berndverst, line 7: should not be necessary due to the alias right?

ItalyPaleAle: Alias is defined in the runtime, not here in the test

PR #2220, opened 2022-10-24
Fixes to mDNS resolver

ItalyPaleAle:
1. If the host doesn't support IPv6 (or supports only IPv6), allow the client to resolve IPs
3. Better concurrency handling with picking an address from the list via round-robin

### DIFF
--- a/nameresolution/mdns/mdns.go
+++ b/nameresolution/mdns/mdns.go
@@ -120,11 +120,11 @@ func (a *addressList) next() *string {
 	}
 
 	if a.counter.Load() == math.MaxUint32 {
+		// This will only reset unless another goroutine has done that already
 		a.counter.CompareAndSwap(math.MaxUint32, 0)
 	}
 	counter := a.counter.Add(1) - 1
-	index := counter % l
-	addr := a.addresses[index]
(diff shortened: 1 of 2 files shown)


nameresolution/mdns/mdns.go
```diff
@@ -300,9 +292,35 @@ func (m *Resolver) Init(metadata nameresolution.Metadata) error {
...
+		zeroconf.SelectIPTraffic(zeroconf.IPv6),
+	}
+	for i := 0; i < len(opts); i++ {
+		resolver, err = zeroconf.NewResolver(opts[i])
+		if err == nil {
+			break
+		}
+	}
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize resolver: %w", err)
```
berndverst, line 319: This will only return the ipv6 specific error which could mask the fact that ipv4 is also broken. Can you update the error message to make this clearer?

berndverst (approved):

berndverst:
Lint

ItalyPaleAle:
@berndverst fixed and ready for review

berndverst (approved):

PR #2228, opened 2022-10-26
Add support for Redis 7 in all Dapr Components

### LINKED ISSUE #1909: Add Dapr support for Redis 7 in cluster mode

berndverst:

# Description

Adds support for Redis 7.

- Dynamically queries the Redis Server to obtain the Redis server and instantiate the appropriate Redis Client (v8 or v9).
- Because there are no Redis 7 images with JSON support at this time, also adds ability to dynamically determine JSON support and handle JSON content type data accordingly.

Fixes #1909 

This reworks PR #2085, which I will abandon.


internal/component/redis/settings.go
```diff
@@ -32,6 +32,10 @@ type Settings struct {
 	DB int `mapstructure:"redisDB"`
 	// The redis type node or cluster
 	RedisType string `mapstructure:"redisType"`
+	// The redis version - controls which redis client to use
+	// >= 7 uses go-redis v9
+	// < 7 uses go-redis v8, this is also the default
+	RedisVersion int `mapstructure:"redisVersion"`
```
ItalyPaleAle, line 38: Is there a way perhaps this could be determined at runtime? Maybe Dapr can ping the Redis server and determine its version? Not sure if that's even possible, just asking.

berndverst: Not from what I can tell.

berndverst:
@ItalyPaleAle I'll convert this back to draft. I added conformance tests later last night and I find some failures with Redis 7 that require investigation.

internal/component/redis/settings.go
ItalyPaleAle: Are you able to give it a shot?

For example use the new SDK and if the version is old, reconnect with the old one. If it works, it would make the component much simpler as it wouldn't require operators to know the redis version beforehand

.github/infrastructure/docker-compose-redis7.yml
```diff
@@ -0,0 +1,8 @@
+version: '2'
+services:
+  redis:
+    image: redis:7.0.5
```
berndverst, line 4: Note there is no available redisjson image for Redis 7, so consequently Query API cannot be supported with Redis 7 at this time (it may be possible, but requires manual set up of the server)

internal/component/redis/settings.go
berndverst: Done and working.

berndverst:
Because I renamed some tests I will have to get a STC member to remove some required tests and add new tests.

Remove
```
bindings.redis conformance
pubsub.redis conformance
state.redis conformance
```

and add
```
bindings.redis.v6 conformance
pubsub.redis.v6 conformance
state.redis.v6 conformance
bindings.redis.v7 conformance
pubsub.redis.v7 conformance
state.redis.v7 conformance
```

internal/component/redis/settings.go
ItalyPaleAle: Noice!!!!

.github/infrastructure/docker-compose-redis7.yml
ItalyPaleAle: SHould this be just redis:7 however? So it gets upgraded to say 7.0.6, 7.1.0...

internal/component/redis/redis.go
```diff
@@ -14,20 +14,62 @@ limitations under the License.
...
+	DoErr(ctx context.Context, args ...interface{}) error
+	Close() error
+	PingResult(ctx context.Context) (string, error)
+	SetNX(ctx context.Context, key string, value interface{}, expiration time.Duration) (*bool, error)
+	EvalInt(ctx context.Context, script string, keys []string, args ...interface{}) (*int, error, error)
+	XAdd(ctx context.Context, stream string, maxLenApprox int64, values map[string]interface{}) (string, error)
+	XGroupCreateMkStream(ctx context.Context, stream string, group string, start string) error
+	XAck(ctx context.Context, stream string, group string, messageID string) error
+	XReadGroupResult(ctx context.Context, group string, consumer string, streams []string, count int64, block time.Duration) ([]RedisXStream, error)
+	XPendingExtResult(ctx context.Context, stream string, group string, start string, end string, count int64) ([]RedisXPendingExt, error)
```
berndverst, line 66: This method and its implementations in the two clients must be slow -- which is why PubSub cannot finish reading the remaining messages in time.

Would appreciate it if you can take a look at my conversion to the generic types.

@ItalyPaleAle

internal/component/redis/redis.go
```diff
@@ -37,110 +79,71 @@ func ParseClientFromProperties(properties map[string]string, defaultSettings *Se
...
-			IdleCheckFrequency: time.Duration(s.IdleCheckFrequency),
-			IdleTimeout:        time.Duration(s.IdleTimeout),
-		}
-		/* #nosec */
-		if s.EnableTLS {
-			options.TLSConfig = &tls.Config{
-				InsecureSkipVerify: s.EnableTLS,
-			}
-		}
+	err := c.DoErr(context.Background(), "JSON.GET")
```
ItalyPaleAle, line 117: SHould this have a timeout?

internal/component/redis/redis.go
```diff
@@ -37,110 +79,71 @@ func ParseClientFromProperties(properties map[string]string, defaultSettings *Se
...
 
-		return redis.NewClusterClient(options)
+	if err == nil {
+		trueVal := true
+		clientHasJSONSupport = &trueVal
```
ItalyPaleAle, line 121: Nit: use `ptr.Of(true)` from dapr/kit/ptr

internal/component/redis/redis.go
```diff
@@ -37,110 +79,71 @@ func ParseClientFromProperties(properties map[string]string, defaultSettings *Se
...
-		ReadTimeout:        time.Duration(s.ReadTimeout),
-		WriteTimeout:       time.Duration(s.WriteTimeout),
-		PoolSize:           s.PoolSize,
-		MaxConnAge:         time.Duration(s.MaxConnAge),
-		MinIdleConns:       s.MinIdleConns,
-		PoolTimeout:        time.Duration(s.PoolTimeout),
-		IdleCheckFrequency: time.Duration(s.IdleCheckFrequency),
-		IdleTimeout:        time.Duration(s.IdleTimeout),
+	if strings.HasPrefix(err.Error(), "ERR unknown command") {
+		falseVal := false
```
ItalyPaleAle, line 126: Here too

internal/component/redis/redis.go
```diff
@@ -37,110 +79,71 @@ func ParseClientFromProperties(properties map[string]string, defaultSettings *Se
...
+		clientHasJSONSupport = &falseVal
+		return false
 	}
+	trueVal := true
+	clientHasJSONSupport = &trueVal
```
ItalyPaleAle, line 131: Here too

internal/component/redis/redis.go
```diff
@@ -37,110 +79,71 @@ func ParseClientFromProperties(properties map[string]string, defaultSettings *Se
...
+	return true
+}
 
-	/* #nosec */
-	if s.EnableTLS {
-		options.TLSConfig = &tls.Config{
-			InsecureSkipVerify: s.EnableTLS,
+func GetServerVersion(c RedisClient) (string, error) {
+	ctx := c.Context()
```
ItalyPaleAle, line 136: Should this have a timeout too?

internal/component/redis/redis.go
berndverst: I refactored the existing code. Right now my goal wasn't to add new functionality such as adding timeouts where there were none. We can do that in another pass if desired though.

internal/component/redis/redis.go
```diff
@@ -37,110 +79,71 @@ func ParseClientFromProperties(properties map[string]string, defaultSettings *Se
...
+	res, err := c.DoResult(ctx, "INFO", "server")
+	if err != nil {
+		return "", err
+	}
+	// get row in string res beginning with "redis_version"
+	rows := strings.Split(res.(string), "\n")
+	for _, row := range rows {
+		if strings.HasPrefix(row, "redis_version:") {
+			return strings.TrimSpace(strings.Split(row, ":")[1]), nil
```
ItalyPaleAle, line 145: Nit: strings.Cut ? In case there are more `:` after the first=

internal/component/redis/v8client.go
```diff
@@ -0,0 +1,279 @@
...
+func (c v8Client) Context() context.Context {
+	return c.client.Context()
+}
+
+func (c v8Client) Close() error {
+	return c.client.Close()
+}
+
+func (c v8Client) PingResult(ctx context.Context) (string, error) {
+	return c.client.Ping(ctx).Result()
```
ItalyPaleAle, line 61: Should there be a timeout?

internal/component/redis/v8client.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+func (c v8Client) EvalInt(ctx context.Context, script string, keys []string, args ...interface{}) (*int, error, error) {
+	eval := c.client.Eval(ctx, script, keys, args...)
```
ItalyPaleAle, line 65: Here too. And eery other operation in thsi file

internal/component/redis/v9client.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+// v9Client is an interface implementation of RedisClient
+
+type v9Client struct {
+	client v9.UniversalClient
+}
+
+func (c v9Client) DoErr(ctx context.Context, args ...interface{}) error {
+	return c.client.Do(ctx, args...).Err()
```
ItalyPaleAle, line 45: In this file too... SHould we have timeouts?

lock/redis/standalone.go
```diff
@@ -135,37 +133,35 @@ func (r *StandaloneRedisLock) parseConnectedSlaves(res string) int {
...
 	// 1.Setting redis expiration time
-	nx := r.client.SetNX(r.ctx, req.ResourceID, req.LockOwner, time.Second*time.Duration(req.ExpiryInSeconds))
-	if nx == nil {
+	nxval, err := r.client.SetNX(r.ctx, req.ResourceID, req.LockOwner, time.Second*time.Duration(req.ExpiryInSeconds))
+	if nxval == nil {
 		return &lock.TryLockResponse{}, fmt.Errorf("[standaloneRedisLock]: SetNX returned nil.ResourceID: %s", req.ResourceID)
 	}
 	// 2. check error
-	err := nx.Err()
 	if err != nil {
```
ItalyPaleAle, line 141: Should error check be before thecking for nil nxval?

lock/redis/standalone.go
```diff
@@ -135,37 +133,35 @@ func (r *StandaloneRedisLock) parseConnectedSlaves(res string) int {
...
 	// 1. delegate to client.eval lua script
-	eval := r.client.Eval(r.ctx, unlockScript, []string{req.ResourceID}, req.LockOwner)
+	evalInt, parseErr, err := r.client.EvalInt(r.ctx, unlockScript, []string{req.ResourceID}, req.LockOwner)
 	// 2. check error
-	if eval == nil {
+	if evalInt == nil {
 		return newInternalErrorUnlockResponse(), fmt.Errorf("[standaloneRedisLock]: Eval unlock script returned nil.ResourceID: %s", req.ResourceID)
 	}
-	err := eval.Err()
 	if err != nil {
```
ItalyPaleAle, line 158: Same here - should error check be first? Or errors could be eaten

lock/redis/standalone.go
```diff
@@ -135,37 +133,35 @@ func (r *StandaloneRedisLock) parseConnectedSlaves(res string) int {
...
 	}
 	// 3. parse result
-	i, err := eval.Int()
+	i := *evalInt
 	status := lock.InternalError
-	if err != nil {
+	if parseErr != nil {
 		return &lock.UnlockResponse{
 			Status: status,
 		}, err
```
ItalyPaleAle, line 167: Can't leave a comment, but on L 193 (`c.client.Close()`) consider setting c.client to nil. Same on L189 after cancling the context

pubsub/redis/redis.go
```diff
@@ -144,7 +144,7 @@ func (r *redisStreams) Init(metadata pubsub.Metadata) error {
 
 	r.ctx, r.cancel = context.WithCancel(context.Background())
 
-	if _, err = r.client.Ping(r.ctx).Result(); err != nil {
+	if _, err = r.client.PingResult(r.ctx); err != nil {
```
ItalyPaleAle, line 147: Does this need a timeout?

state/redis/redis.go
```diff
@@ -20,7 +20,6 @@ import (
 	"strings"
 
 	"github.com/agrea/ptr"
```
ItalyPaleAle, line 22: Nit: swap out agea/ptr for dapr/kit/ptr.Of?

ItalyPaleAle (approved):

ItalyPaleAle (approved):
LGTM with some nits

internal/component/redis/redis.go
ItalyPaleAle: Can you walk me through the problem tomorrow? I’m not sure I understand the issue.

berndverst: Conformance test now times out for PubSub. Only explanation can be my code that converts the v8 and v9 client specific types for messages and streams to my generic interface types.

internal/component/redis/redis.go
berndverst: Not possible given the output returned by this Redis command.

lock/redis/standalone.go
berndverst: 193 makes sense, but I think for 189 we should still let it attempt to close the client first.

lock/redis/standalone.go
berndverst: `err` and `nx` are returned by the same internal function, so even though it looks like an error would be swallowed, that isn't the case here.

Changing the order would change the specific error messages returned, so will keep it as is.

lock/redis/standalone.go
berndverst: `err` and `evalInt` are returned by the same internal function, so even though it looks like an error would be swallowed, that isn't the case here.

Changing the order would change the specific error messages returned, so will keep it as is.

berndverst:
@ItalyPaleAle seems the timeouts even still exist in v7. We were just lucky when we investigated this earlier.

ItalyPaleAle (approved):

PR #2234, opened 2022-10-27
PubSub: Fall back to 'consumerID' when group not present

### LINKED ISSUE #2233: All Dapr PubSub Components should make consumer group optional and use runtime provided group name if not present
Dapr Runtime PubSub checks for the presence of `consumerID` metadata, and if not present overwrites this to be the app ID. This allows every Dapr app to receive a PubSub event at least once.

If the consumer ID is manually overridden, of course this will not work.

Unfortunately, many Dapr PubSub component have not made consumer ID / consumer Group name an optional metadata attribute, and do not consistently check for the presence of `consumerID` when the component specific attribute name is missing.

Every Dapr PubSub component must be reviewed such that:
If the consumer ID / group name metadata attribute isn't called `consumerID`, the attribute should be set to the value found in `consumerID` (which is provided by the runtime).

berndverst:

Fixes #2233

RocketMQ, AWS SNS/SQS, Jetstream were the only components not using the metadata key `consumerID` which is used by the runtime to inject a default consumer group if none is present.


berndverst:
Ideally `consumerID` should be a required field of the PubSub base metadata in `pubsub/metadata.go`, all code should be refactored to use that value - but existing aliases can be maintained.

For now this PR is acceptable as a quick fix, but I worry future components may make the same mistake.

ItalyPaleAle (approved):

PR #2236, opened 2022-10-28
adds wasm benchmarks and updates wapc-go to v0.5.5

codefromthecrypt:
# Description

This updates wapc-go to v0.5.5 which updates [wazero](https://wazero.io/) to [1.0.0-pre.3](https://github.com/tetratelabs/wazero/releases/tag/v1.0.0-pre.3). This is the last release that will build with Go 1.17.

Notably, this improves performance and changes host function syntax.

Here is the before performance:
```
Benchmark_native/rewrite/rewrite
Benchmark_native/rewrite/rewrite-16 	 1092189	      1037 ns/op
Benchmark_basic/rewrite
Benchmark_basic/rewrite-16          	  230673	      5055 ns/op
```

Here is the new performance:
```
Benchmark_native/rewrite/rewrite
Benchmark_native/rewrite/rewrite-16 	 1133650	      1008 ns/op
Benchmark_basic
Benchmark_basic/rewrite
Benchmark_basic/rewrite-16          	  390234	      2802 ns/op
```
## Issue reference

n/a

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


middleware/http/wasm/internal/test/logger.go
```diff
@@ -0,0 +1,81 @@
...
+	"bytes"
+	"fmt"
+
+	"github.com/dapr/kit/logger"
+)
+
+// compile-time check to ensure testLogger implements logger.testLogger.
+var _ logger.Logger = &testLogger{}
+
+func NewLogger() logger.Logger {
```
ItalyPaleAle, line 13: Why this logger?

codefromthecrypt: for benchmarks, so that they don't end up using a real one

codefromthecrypt: @ItalyPaleAle sorry I added a commit, I left the old copy. The benchmarks are in the root directory of wasm as they will be re-used when we add the flexible wasm thing next (ex compare existing basic vs the new http-wasm one)

ItalyPaleAle: Dapr/kit/logger now supports setting a buffer as log destination. Would that work? (Just fetch the latest version from main)
I am just worried that this logger will need to be updated too every time the log interface changes.

codefromthecrypt: yeah I can refactor it out. I'll try it!

codefromthecrypt: I think the other thing was testing was more difficult due to the amount of formatting logic internal to it. we'll see

codefromthecrypt: It looks like this isn't exported @ItalyPaleAle 

I can see where some code sets does this, but it is within package. Am I missing something?

```
func getTestLogger(buf io.Writer) *daprLogger {
	l := newDaprLogger(fakeLoggerName)
	l.logger.Logger.SetOutput(buf)

	return l
}
```

codefromthecrypt: in any case, the logger here was moved not written from scratch. can we decouple this PR from making the upstream logger capable of being used in test mode maybe?

codefromthecrypt: ah I see you added here.. if it is ok to use a non-tagged version I can do that in this PR. otherwise we need a tag on that repo https://github.com/dapr/kit/pull/25

ItalyPaleAle: Using a non-tagged version is fine

codefromthecrypt: hmm this isn't working out well, as something in the path of the request is setting the internal level to panic. I've not hunted down what, yet.

codefromthecrypt: I got it. I'll use a master commit for dapr/kit for now.

middleware/http/wasm/basic/internal/e2e_test.go
```diff
@@ -61,22 +67,21 @@ func Test_EndToEnd(t *testing.T) {
...
-				require.Equal(t, `Info(main ConsoleLog)
-Info(request[0] ConsoleLog)
-Debug(wasm stdout: main Stdout
-request[0] Stdout
-)
-Debug(wasm stderr: main Stderr
-request[0] Stderr
-)
-`, log.String())
+				for _, s := range []string{
```
codefromthecrypt, line 70: this is a little less precise than before, but I'm perfectly fine with it!

middleware/http/wasm/benchmark_test.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func Benchmark_basic(b *testing.B) {
+	md := metadata.Base{Properties: map[string]string{
+		"path":     "./basic/example/example.wasm",
+		"poolSize": strconv.Itoa(parallel),
+	}}
+
+	l := logger.NewLogger(b.Name())
+	l.SetOutput(io.Discard)
```
codefromthecrypt, line 39: fyi benchmarks don't do any logging, and I checked there's no perf difference after using dapr/kit logging

codefromthecrypt:
rebased

codefromthecrypt:
once this is in I'll rebase https://github.com/dapr/components-contrib/pull/2239 and mark it not draft

codefromthecrypt:
#2239 is replacing basic so this isn't needed anymore

PR #2237, opened 2022-10-29
Crypto building block: Initial implementation + first 4 components

ItalyPaleAle:
# Description

This PR includes the initial implementation for the crypto building block as described in dapr/proposals#3

It includes:

1. The new building block's interface, as described in the issue in the proposal
1. 4 component implementations:
    - `crypto.kubernetes.secrets`: stores keys as Kubernetes secrets
    - `crypto.jwks`: retrieves keys from a JWKS, either from a HTTP(S) URL or from the local disk
    - `crypto.localstorage`: uses keys stored in files on a locally-mounted folder
    - `crypto.azure.keyvault`: uses keys stored in Azure Key Vault (as keys)
1. Support for local crypto (where crypto operations are performed locally in the runtime) via a new "Dapr crypto" in dapr/kit: dapr/kit#38
   - The 3 non-Azure components are based on a shared, abstract "local crypto base component" which is based on this package.
1. Full conformance tests

## TODO

* [x] Add more unit tests, especially for the local crypto packages
* [X] Create conformance tests for the new components
* [ ] Docs for the new components


berndverst:
To my knowledge the crypto building block and design have not been approved yet. Until that is the case this PR will not be merged.

ItalyPaleAle:
> To my knowledge the crypto building block and design have not been approved yet. Until that is the case this PR will not be merged.

Yup. It’s in draft until all is signed off

yaron2:
> To my knowledge the crypto building block and design have not been approved yet. Until that is the case this PR will not be merged.

It's a bad practice to create PRs before there's a proposal sign off. This API is unlikely to find it's way to 1.10 with DocumentStore and bulk pub/sub, unless we somehow find ourselves doing very well on these fronts simultaneously and holistically (PRs and vetted docs).

ItalyPaleAle:
Happy to make the required changes when needed. In the meanwhile, this hopefully showcases the proposed implementation, the end-user benefits (a simple addition was a JWKS component which is very useful for things like JWTs)... and get things moving a bit :)

.github/infrastructure/conformance/azure/setup-azure-conf-test.sh
```diff
@@ -213,11 +213,11 @@ IOT_HUB_EVENT_HUB_CONNECTION_STRING_VAR_NAME="AzureIotHubEventHubConnectionStrin
 IOT_HUB_BINDINGS_CONSUMER_GROUP_VAR_NAME="AzureIotHubBindingsConsumerGroup"
 IOT_HUB_PUBSUB_CONSUMER_GROUP_VAR_NAME="AzureIotHubPubsubConsumerGroup"
 
-KEYVAULT_CERT_NAME="AzureKeyVaultSecretStoreCert"
-KEYVAULT_CLIENT_ID_VAR_NAME="AzureKeyVaultSecretStoreClientId"
-KEYVAULT_SERVICE_PRINCIPAL_CLIENT_SECRET_VAR_NAME="AzureKeyVaultSecretStoreServicePrincipalClientSecret"
-KEYVAULT_SERVICE_PRINCIPAL_CLIENT_ID_VAR_NAME="AzureKeyVaultSecretStoreServicePrincipalClientId"
-KEYVAULT_TENANT_ID_VAR_NAME="AzureKeyVaultSecretStoreTenantId"
+KEYVAULT_CERT_NAME="AzureKeyVaultCert"
```
ItalyPaleAle, line 216: Reviewers: I've manually copied over the secrets/certs, so they exist with both names in the Key Vault (for both old and new tests)

crypto/jwks/component.go
```diff
@@ -0,0 +1,248 @@
...
+
+	// Check if the value is a valid path to a local file
+	stat, err := os.Stat(md)
+	if err == nil && stat != nil && !stat.IsDir() {
+		return k.initJWKSFromFile(ctx, md)
+	}
+
+	// Treat the value as the actual JWKS
+	// First, check if it's base64-encoded (remove trailing padding chars if present first)
+	mdJSON, err := base64.RawStdEncoding.DecodeString(strings.TrimRight(md, "="))
```
mukundansundar, line 127: is a trim right needed here?

crypto/jwks/component.go
```diff
@@ -0,0 +1,248 @@
...
+			case <-eventCh:
+				// When there's a change, reload the JWKS file
+				err := k.parseJWKSFile(file)
+				if loaded != nil {
+					// The first time, signal that the initialization was complete and pass the error
+					loaded <- err
+					close(loaded)
+					loaded = nil
+				} else {
+					// Log errors only
```
mukundansundar, line 191:
```suggestion
					// Only log errors
```

crypto/jwks/component.go
```diff
@@ -0,0 +1,248 @@
...
+	return nil
+}
+
+func (k *jwksCrypto) initJWKSFromFile(ctx context.Context, file string) error {
+	// Get the path to the folder containing the file
+	path := filepath.Dir(file)
+
+	// Start watching for changes in the filesystem
+	eventCh := make(chan struct{})
+	loaded := make(chan error)
```
mukundansundar, line 171: potential place for memory leak? as in if the init starts but Init context is completed and the `k.ctx` is not yet canceled. 
Should this be a buffered channel?

crypto/jwks/component.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	k.jwksLock.Lock()
+	k.jwks = jwks
+	k.jwksLock.Unlock()
+
+	return nil
+}
+
+// Retrieves a key (public or private or symmetric) from a Kubernetes secret.
```
mukundansundar, line 233: qq why is this a Kubernetes secret?

crypto/kubernetes/secrets/component.go
```diff
@@ -0,0 +1,135 @@
...
+	"k8s.io/client-go/kubernetes"
+
+	contribCrypto "github.com/dapr/components-contrib/crypto"
+	kubeclient "github.com/dapr/components-contrib/internal/authentication/kubernetes"
+	internals "github.com/dapr/kit/crypto"
+	"github.com/dapr/kit/logger"
+)
+
+const (
+	requestTimeout              = 30 * time.Second
```
mukundansundar, line 35: should this be configurable?

.github/scripts/test-info.mjs
```diff
@@ -183,7 +198,8 @@ const components = {
         requireAWSCredentials: true,
         requireTerraform: true,
         conformanceSetup: 'conformance-pubsub.aws.snssqs.terraform-setup.sh',
-        conformanceDestroy: 'conformance-pubsub.aws.snssqs.terraform-destroy.sh',
+        conformanceDestroy:
+            'conformance-pubsub.aws.snssqs.terraform-destroy.sh',
```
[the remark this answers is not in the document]
ItalyPaleAle: That looks like prettier's doing (the linter). I ran it and it must have changed that line.

crypto/azure/keyvault/algorithms.go
```diff
@@ -0,0 +1,84 @@
...
+
+// GetJWKEncryptionAlgorithm returns a JSONWebKeyEncryptionAlgorithm constant is the algorithm is a supported one.
+func GetJWKEncryptionAlgorithm(algorithm string) *azkeys.JSONWebKeyEncryptionAlgorithm {
+	// Special case for AES-CBC, since we treat A[NNN]CBC as having PKCS#7 padding, and A[NNN]CBC-NOPAD as not using padding
+	switch algorithm {
+	case internals.Algorithm_A128CBC, internals.Algorithm_A192CBC, internals.Algorithm_A256CBC:
+		// Append "PAD", e.g. "A128CBCPAD"
+		algorithm += "PAD"
+	case internals.Algorithm_A128CBC_NOPAD, internals.Algorithm_A192CBC_NOPAD, internals.Algorithm_A256CBC_NOPAD:
+		// Remove the "-NOPAD" prefix, e.g. "A128CBC"
```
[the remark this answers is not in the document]
ItalyPaleAle: I'll change to suffix.

As for `-NOPAD`, yes it's a constant I made up to smoothen the differences between Azure Key Vault and the RFC 7518 spec (JSON Web Algorithms). 

- In RFC7518, algorithms like `A128CBC` use AES-CBC with PKCS#7 padding (note: PKCS#5 and PKCS#7 padding are essentially equivalent).
  - The RFC doesn't actually define `A128CBC` because the RFC only includes authenticated ciphers. So it actually defines `A128CBC-HS256`, which we will support once https://github.com/dapr/kit/pull/40 is merged. Nevertheless, the specs for `A128CBC-HS256` do require PKCS#7 padding
- Azure Key Vault supports two algorithms (source: https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/keyvault/azkeys#JSONWebKeyEncryptionAlgorithm ):
   - `A128CBC` does **not** use padding
   - `A128CBCPAD` **does** use padding
   - Most people will want to use padding (not using padding means that the plaintext message's length must be a multiple of 16). So for Azure Key Vault we convert `A128CBC` into `A128CBCPAD`, and allow `-NOPAD` algorithms if one really, really doesn't want to use padding

crypto/azure/keyvault/algorithms.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+func init() {
+	listEncryption := azkeys.PossibleJSONWebKeyEncryptionAlgorithmValues()
+	validEncryptionAlgs = make(map[string]struct{}, len(listEncryption))
+	for _, v := range listEncryption {
+		validEncryptionAlgs[string(v)] = struct{}{}
+	}
+
+	listSignature := azkeys.PossibleJSONWebKeySignatureAlgorithmValues()
```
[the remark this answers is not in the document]
ItalyPaleAle: They do not call the network.

crypto/azure/keyvault/component.go
```diff
@@ -0,0 +1,418 @@
...
+
+type keyID struct {
+	Version string
+	Name    string
+}
+
+func newKeyID(val string) keyID {
+	obj := keyID{}
+	idx := strings.IndexRune(val, '/')
+	// Can't be on position 0, because the key name must be at least 1 character
```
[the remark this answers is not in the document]
ItalyPaleAle: The network operation will fail after this call. Would rather keep `newKeyID` returning a single value so it's simpler to use.

crypto/jwks/component.go
ItalyPaleAle: No, with base64 the `=` can only be present at the end. See: https://en.wikipedia.org/wiki/Base64

crypto/jwks/component.go
ItalyPaleAle: Actually I think "Log errors only" is more correct, althought it's subtle

- "Only log errors" means -> We will log only in case of error
- "Log errors only" means -> With errors, we'll only show them as logs and not take further action

crypto/kubernetes/secrets/component.go
ItalyPaleAle: I don't think for K8s this needs to be configurable, since the communication should be within the cluster IMHO. In case of failures here, users can retry (e.g. with resiliency).

crypto/subtlecrypto.go
```diff
@@ -0,0 +1,162 @@
...
+		pubKey jwk.Key,
+		err error,
+	)
+
+	// Encrypt a small message and returns the ciphertext.
+	Encrypt(ctx context.Context,
+		// Input plaintext
+		plaintext []byte,
+		// Encryption algorithm to use
+		algorithm string,
```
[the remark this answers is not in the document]
ItalyPaleAle: They are all defined here: https://github.com/dapr/kit/blob/b5bafe889a818601bcd22acb3807d21069be7d69/crypto/consts.go#L38 But each component may offer only a subset of this list. (No component implements them all, at the moment)

crypto/azure/keyvault/component.go
```diff
@@ -0,0 +1,418 @@
...
+	"github.com/lestrrat-go/jwx/v2/jwk"
+
+	contribCrypto "github.com/dapr/components-contrib/crypto"
+	azauth "github.com/dapr/components-contrib/internal/authentication/azure"
+	internals "github.com/dapr/kit/crypto"
+	"github.com/dapr/kit/logger"
+)
+
+const (
+	requestTimeout = 30 * time.Second
```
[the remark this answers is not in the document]
ItalyPaleAle: Yes i think it should

ItalyPaleAle:
@mukundansundar  RE:

> Are the list of algorithms constrained or is it going to depend on underlying implementation?

Makes me wonder if perhaps we should include 2 new methods in the `SubtleCrypto` interface called:

- `SupportedEncryptionAlgorithms`
- `SupportedSignatureAlgorithms`

They each return a slice with the list of algorithms supported. Those can then be featured in the Metadata API.

Thoughts? (If you think it's useful, I'll add it in a separate PR to not make this even bigger)

crypto/jwks/component.go
ItalyPaleAle: Yes, good catch!

.github/infrastructure/conformance/azure/setup-azure-conf-test.sh
tmacam: Will they auto-expire or should we have something to remind us to remove them in the future?

crypto/azure/keyvault/component.go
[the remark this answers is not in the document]
ItalyPaleAle: Done

crypto/subtlecrypto.go
```diff
@@ -0,0 +1,162 @@
...
+		// Name (or name/version) of the key to use in the key vault
+		key string,
```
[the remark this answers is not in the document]
ItalyPaleAle: Done

.github/infrastructure/conformance/azure/setup-azure-conf-test.sh
ItalyPaleAle: It will expire but this is not a new thing for this PR. Just re-using the key that was there before.

We will need to rotate the keys in the future. The reminder will be that the CI will fail 🤣

crypto/azure/keyvault/component.go
```diff
@@ -0,0 +1,418 @@
...
+	if algorithm == nil {
+		return false, fmt.Errorf("invalid algorithm: %s", algorithmStr)
+	}
+
+	// Verifying with non-cacheable keys must happen in the vault
+	if !kid.Cacheable() {
+		return k.verifyInVault(parentCtx, digest, signature, algorithm, kid)
+	}
+
+	// Using a cacheable, asymmetric key, we can verify the data directly here
```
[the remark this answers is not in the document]
ItalyPaleAle: Azure Key Vault only supports asymmetric algorithms for digital signatures (and `GetJWKSignatureAlgorithm` enforces that). A mismatch between key type and algorithm would cause an error in the vault.

crypto/azure/keyvault/algorithms.go
```diff
@@ -0,0 +1,84 @@
...
+
+	internals "github.com/dapr/kit/crypto"
+)
+
+var (
+	validEncryptionAlgs map[string]struct{}
+	validSignatureAlgs  map[string]struct{}
+)
+
+func init() {
```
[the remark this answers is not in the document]
ItalyPaleAle: This is currently added to the _package_ `init()` function, which is executed when the code is loaded in memory (i.e. when daprd starts). It's a common pattern for putting code that initializes singleton.

The other approach is to load that lazily. In the _component's_ `Init()` method, I can use `sync.Once` (with the sync.Once token stored as a global variable in the package) to initialize the map the first time a component of this kind is loaded.

The current method is simpler as it doesn't require meddling with sync.Once, but the maps (which contain about a dozen keys each, with `struct{}{}` as value) will use a few bytes of memory even when users don't use this component. Using sync.Once requires less memory but adds a bit more complexity. Neither should matter for dynamic loading of component.

To be honest I don't have strong opinions, so whatever you folks think is best.

.github/infrastructure/conformance/azure/setup-azure-conf-test.sh
berndverst: LGTM

pubsub/feature.go
```diff
@@ -25,11 +29,5 @@ type Feature string
...
 // IsPresent checks if a given feature is present in the list.
 func (f Feature) IsPresent(features []Feature) bool {
-	for _, feature := range features {
-		if feature == f {
-			return true
-		}
-	}
-
-	return false
+	return slices.Contains(features, f)
```
berndverst, line 32: Why was this modified? This kind of change can also be the cause of subtle bugs.

Also, you are using the experimental slices packages. Maybe best to avoid `golang.org/x/exp`

ItalyPaleAle:
After talking to @berndverst , I'll split this into 3 PRs tomorrow:

1. Crypto interface + conf tests
2. Azure Key Vault component
3. The 3 local components

pubsub/feature.go
ItalyPaleAle: Was requested by @mukundansundar here: https://github.com/dapr/components-contrib/pull/2237#discussion_r1142498284

Not concerned about it being "experimental" however. It's in `exp` because it uses generics, but otherwise it's exactly the same code we had before. We use it in other parts of Dapr.

Anyways, i'm happy to revert if you'd like.

berndverst:
> After talking to @berndverst , I'll split this into 3 PRs tomorrow:
> 
> 1. Crypto interface + conf tests
> 2. Azure Key Vault component
> 3. The 3 local components

Thanks, that will help a lot with the review.

crypto/azure/keyvault/algorithms.go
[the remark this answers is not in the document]
berndverst: We wouldn't allow any other package to initialize code in `init()` so I think we should also not use that approach here. Using the component's `Init()` method seems like the way to go.

pubsub/feature.go
berndverst: @mukundansundar in the future let's be consistent. Refactoring this to using `slices` could have been elsewhere.

ItalyPaleAle:
Opened:

- #2690
- #2691 
- #2692

Will close this PR after I've addressed the remaining feedback

crypto/azure/keyvault/algorithms.go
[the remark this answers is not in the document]
ItalyPaleAle: Addressed in #2692

PR #2239, opened 2022-10-29
Adds http-wasm middleware to replace basic

codefromthecrypt:
# Description

This adds a new httpwasm middleware based on the [http-wasm][1] Application Binary Interface (ABI), specifically using the [http-handler][2] ABI via [http-wasm-host-go][3]. The net result is far more flexible programming, with even faster execution than before. Currently, the only guest language is [TinyGo][4], which the examples are written in, but others can be requested [here][1].

The example code now can do a lot more, including full access to write responses in wasm, as needed. To that end the example is no longer just showing rewrite, but showing a route function which executes in about 2 microseconds on my laptop benchmarks:

```go
// handleRequest implements a simple HTTP router.
func handleRequest(req api.Request, resp api.Response) (next bool, reqCtx uint32) {
	// If the URI starts with /host, trim it and dispatch to the next handler.
	if uri := req.GetURI(); strings.HasPrefix(uri, "/host") {
		req.SetURI(uri[5:])
		next = true // proceed to the next handler on the host.
		return
	}

	// Serve a static response
	resp.Headers().Set("Content-Type", "text/plain")
	resp.Body().WriteString("hello")
	return // skip the next handler, as we wrote a response.
}
```

## Issue reference

closes https://github.com/dapr/dapr/issues/5003

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: https://github.com/dapr/docs/pull/2922

[1]: https://http-wasm.io
[2]: https://http-wasm.io/http-handler-abi/
[3]: https://github.com/http-wasm/http-wasm-host-go
[4]: https://github.com/http-wasm/http-wasm-guest-tinygo


codefromthecrypt:
I will update this to https://http-wasm.io once the pending PR is merged

codefromthecrypt:
Thanks for the feedback @daixiang0 

> middleware.http.wasm.httpwasm is kind of duplicate, how about general?

So the spec name is http-wasm .. https://http-wasm.io

how about `middleware.http.wasm.http-wasm` or just have this as simply `middleware.http.wasm` and stop nesting? wdyt?

daixiang0:
`middleware.http.wasm` looks good since we do not want to implement it in another way.

I will review it after it is ready.

codefromthecrypt:
ok I will change this PR to replace the other one. will click ready after

codefromthecrypt:
@daixiang0 ok it is flat now, and I also updated docs PR with same

middleware/http/wasm/README.md
```diff
@@ -1,5 +1,34 @@
-# WebAssembly Middleware
+## http-wasm WebAssembly Middleware
 
-Middleware components in this directory load WebAssembly binaries (wasm) to
-modify requests and responses. Each middleware describes its features and
-constraints such as required function exports.
+This component lets you manipulate an incoming request or serve a response with
+custom logic compiled using the [http-wasm][1] Application Binary Interface
```
ItalyPaleAle, line 4: Nit: links should be inline, e.g. `[htp-wasm](https://http-wasm.io/)`

middleware/http/wasm/httpwasm.go
```diff
@@ -0,0 +1,162 @@
...
+		}
+		if stderr := rh.stderr.String(); len(stderr) > 0 {
+			rh.logger.Debugf("wasm stderr: %s", stderr)
+		}
+	})
+}
+
+// Close implements io.Closer
+func (rh *requestHandler) Close() error {
+	// TODO: we have to use a finalizer as there's no way in dapr to close middleware, yet.
```
ItalyPaleAle, line 157: I think you can remove the TODO. Although Dapr doesn't close middleware components yet, we will implement it with `Close() error` (= `io.Closer`).

middleware/http/wasm/httpwasm.go
```diff
@@ -0,0 +1,162 @@
...
+
+type requestHandler struct {
+	mw             wasmnethttp.Middleware
+	logger         logger.Logger
+	stdout, stderr *bytes.Buffer
+}
+
+func (rh *requestHandler) requestHandler(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		h := rh.mw.NewHandler(ctx, next)
```
ItalyPaleAle, line 138: Can't you get the context from `r.Context()`? This would be a context tied to the request (well, in theory.... in practice it isn't while we're [stuck with fasthttp](https://github.com/dapr/dapr/issues/4979), but that's a different story :) )

middleware/http/wasm/httpwasm.go
```diff
@@ -0,0 +1,190 @@
...
+	var stdout, stderr bytes.Buffer
+	mw, err := wasmnethttp.NewMiddleware(ctx, meta.guest,
+		handler.Logger(m),
+		// TODO: currently, the pool is unbounded
+		// See https://github.com/http-wasm/http-wasm-host-go/issues/42
+		// handler.PoolMax(meta.PoolSize)
+		handler.ModuleConfig(wazero.NewModuleConfig().
+			WithStdout(&stdout). // reset per request
+			WithStderr(&stderr). // reset per request
+			// The below violate sand-boxing, but allow code to behave as expected.
```
codefromthecrypt, line 99: @ItalyPaleAle are you alright with these? wasm has two purposes we are using.

* decoupling (maybe even from programming language, but also) from the binary
* sandboxing (preventing plugins from crashing the process or causing memory problems)

strict sandboxing is default in wazero meaning fake time, fake random no-op sleep.

middleware/http/wasm/README.md
```diff
@@ -1,5 +1,23 @@
-# WebAssembly Middleware
+## http-wasm WebAssembly Middleware
```
daixiang0, line 1: `WebAssembly Middleware` is enough.

codefromthecrypt: agreed!

middleware/http/wasm/httpwasm.go
ItalyPaleAle: I do not understand enough about WASM to have an opinion on this sorry. I lean towards trusting you.

In general, I do think that we do want real times.

If I had to guess, fake timers etc are in place to protect against side-channel attacks. In the context of Dapr, at least while we have 1 daprd process per app in the sidecar model, I don't know how much of a concern side-channel attacks are.

CC: @yaron2

codefromthecrypt: FWIW, this is commented in a way that if later someone wants to be more careful, they can. Mostly it is about how pessimistic people want to be with wasm plugins. However, since they are loaded from the FS, it is implicitly more controlled than purely random source!

codefromthecrypt:
ps @anuraaga reminded me that we don't need to maintain poolSize here anymore. We used to need that here because in waPC the only way to do pooling required also managing its capacity. However, the better choice is to stick with dapr's controls for concurrency vs make a different knob here.

https://docs.dapr.io/operations/configuration/control-concurrency/
https://github.com/http-wasm/http-wasm-host-go/blob/main/RATIONALE.md#guest-pool

I'll update the docs PR as well.

ItalyPaleAle (dismissed):
LGTM 

But I'd like @berndverst  and/or @yaron2  to confirm they're good with skipping the security measures for side channel attacks, per: https://github.com/dapr/components-contrib/pull/2239#discussion_r1009967030

middleware/http/wasm/README.md
```diff
@@ -1,5 +1,23 @@
...
-Middleware components in this directory load WebAssembly binaries (wasm) to
-modify requests and responses. Each middleware describes its features and
-constraints such as required function exports.
+This component lets you manipulate an incoming request or serve a response with custom logic compiled using the [htp-wasm](https://http-wasm.io/) Application Binary Interface (ABI). The `handle_request` function receives an incoming request and can manipulate it or serve a response as necessary.
+
+Please see the [documentation](https://github.com/dapr/docs/blob/v1.9/daprdocs/content/en/reference/components-reference/supported-middleware/middleware-wasm.md) for general configuration.
+
+### Generating Wasm
+
+To compile your wasm, you must compile source using a http-wasm guest SDK such as [TinyGo](https://github.com/http-wasm/http-wasm-guest-tinygo). You can also make a copy of [hello.go](./example/example.go) and replace the `handler.HandleFn` function with your custom logic.
```
daixiang0, line 9: Better to use [http-wasm-guest-tinygo](https://github.com/http-wasm/http-wasm-guest-tinygo) or something else since TinyGO usually mean the tiny Go Compiler.

middleware/http/wasm/httpwasm.go
```diff
@@ -0,0 +1,161 @@
...
+type middlewareMetadata struct {
+	// Path is where to load a `%.wasm` file that implements the guest side of
+	// the handler protocol. No default.
+	Path string `json:"path"`
+
+	// guest is WebAssembly binary implementing the waPC guest, loaded from Path.
+	guest []byte
+}
+
+// middleware is a wasm basic middleware.
```
daixiang0, line 42: Not basic now, maybe we can just delete this comment.

middleware/http/wasm/httpwasm.go
```diff
@@ -0,0 +1,161 @@
...
+type middleware struct {
+	logger logger.Logger
+}
+
+// NewMiddleware returns a new wasm basic middleware.
```
daixiang0, line 47: same as above...

middleware/http/wasm/httpwasm.go
```diff
@@ -0,0 +1,161 @@
...
+func NewMiddleware(logger logger.Logger) dapr.Middleware {
+	return &middleware{logger: logger}
+}
+
+// GetHandler returns the HTTP handler provided by wasm basic middleware.
```
daixiang0, line 52: same as above...

middleware/http/wasm/httpwasm.go
```diff
@@ -0,0 +1,161 @@
...
+		h := rh.mw.NewHandler(r.Context(), next)
+		defer func() {
+			rh.stdout.Reset()
+			rh.stderr.Reset()
+		}()
+
+		h.ServeHTTP(w, r)
+
+		if stdout := rh.stdout.String(); len(stdout) > 0 {
+			rh.logger.Debugf("wasm stdout: %s", stdout)
```
daixiang0, line 150: Is the log level always DEBUG or sync with Dapr itself?

middleware/http/wasm/internal/testdata/rewrite.wat
```diff
@@ -0,0 +1,92 @@
+;; This example module is written in WebAssembly Text Format to show the
```
daixiang0, line 1: Is it generated as well?

middleware/http/wasm/httpwasm.go
anuraaga: Hi @daixiang0 - the SDK includes a `log` method that is synced with Dapr

https://github.com/http-wasm/http-wasm-guest-tinygo/blob/main/handler/api/api.go#L48
https://github.com/dapr/components-contrib/pull/2239/files/d54eb0d1c97f1a6c2b0e1bfd320eab01ee10e479#diff-90c834bbe0d45903a30fffa3a23ebed48a983e5e9b32868d5b513d8c09dcad8dR93

However the middleware may write directly to `stdout`/`stderr` without using that function, which is what is logged here. We need to pick a level for logging these as the middleware didn't.

Does `Debug` look OK for that?

middleware/http/wasm/internal/testdata/rewrite.wat
codefromthecrypt: no I wrote it. this is the baseline for benchmarks as compiled with tinygo etc won't be faster than this.

codefromthecrypt: I will change the comment!

middleware/http/wasm/README.md
codefromthecrypt: ok!

middleware/http/wasm/httpwasm.go
daixiang0: I know, it is ok.

daixiang0:
Please fix conflicts.

codefromthecrypt:
@daixiang0 ok rebased and squashed.

codefromthecrypt:
sorry about broke. forgot the dapr/kit log apis are not released yet. should be fixed now.

daixiang0 (dismissed):

middleware/http/wasm/httpwasm.go
daixiang0: I think the attack surface is small enough.

daixiang0 (approved):
LGTM

/cc @dapr/maintainers-components-contrib

codefromthecrypt:
@yaron2 I think we're at 🚢 moment. once in, I'll update any doc errata. I'm looking forward to mentioning this at [oredev](https://oredev.org/sessions) next week!

yaron2:
Reviewed comments, LGTM. Thanks much for this contribution @codefromthecrypt

anuraaga:
Thanks everyone, great stuff!

codefromthecrypt:
https://github.com/dapr/dapr/pull/5458

PR #2241, opened 2022-10-31
Added cert test for state.mysql + Fixes including transactions

### LINKED ISSUE #2232: Stable certification tests for MySQL state store
Add stable certification tests for MySQL State Store component.

RELEASE NOTE: **CHANGED** MySQL State Store not certified as stable.

ItalyPaleAle:
Adds certification test for state.mysql.

It also includes two major bug fixes:

- Actually use transactions. Before, transactions were started but then queries were executed outside of transactions.
- Added timeout to queries, configurable with new metadata property `timeoutInSeconds`

Tests are executed against both MySQL 8 and MariaDB 10.

Fixes #2232

Tests not included:

- Testing for timeouts. There was an attempt at this, but it can't be done because the `comcast` library we use to simulate network failures doesn't seem to work with apps running in Docker containers. Multiple attempts have failed very miserably. Seems to be due to namespacing.
- Connecting with PEM. That requires setting up a PKI and will be done in a separate PR.


.github/infrastructure/docker-compose-mysql.yml
```diff
@@ -1,7 +1,7 @@
 version: '2'
 services:
   db:
-    image: mysql
+    image: mysql:8
```
ItalyPaleAle, line 4: Better to pin a specific version. We have learnt that lesson with Redis earlier today...

tests/certification/state/mysql/README.md
```diff
@@ -0,0 +1,32 @@
+# MySQL State Store certification testing
+
+This project aims to test the MySQL State Store component under various conditions. Tests are run against both MySQL 8 and MariaDB 10.
+
+## Test plan
+
+## Test for CRUD operations
+
+1. Able to create and test connection.
```
berndverst, line 9: We do not need to test these things here because it is covered as part of conformance tests, and it really irks me to have duplicate test cases in certification tests :)

tests/certification/state/mysql/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+* Not prone to SQL injection on write
+* Not prone to SQL injection on read
+* Not prone to SQL injection on delete
+
+## Connection Recovery
+
+1- When MySQL/MariaDB goes down and then comes back up - client is able to connect
+
+## Concurrency
```
berndverst, line 23: These cases should already be covered by conformance tests.

berndverst (changes requested):
Please add a test case for these metadata options:

```
  - name: pemPath # Required if pemContents not provided. Path to pem file.
    value: "<PEM PATH>"
  - name: pemContents # Required if pemPath not provided. Pem value.
    value: "<PEM CONTENTS>"    
```


https://docs.dapr.io/reference/components-reference/supported-state-stores/setup-mysql/

tests/certification/state/mysql/README.md
berndverst: Ok, to leave here - but please be mindful of cert test bloat. Conformance tests already cover a lot. This makes maintenance harder.

ItalyPaleAle: Gotcha. I based the test on the Postgres one, I can remove them if needed

ItalyPaleAle:
Authenticating with PEM certificates will require managing a PKI for mTLS. https://smallstep.com/hello-mtls/doc/combined/mysql/mysql-client

Unless we are comfortable keeping certs and keys in Git (I am not really, even if they're just for tests), this will require a bit of work. We'll need to generate the certs (CA, server, client) before starting the Docker Compose. It's doable, just a bit of work.

Could that be done in a separate PR?

berndverst (approved):

PR #2245, opened 2022-10-31
Updated depguard linter config

ItalyPaleAle:
1. Add older (pre-v4) versions of `github.com/cenkalti/backoff` to the denylist
2. Add `github.com/agrea/ptr` to the denylist (use `github.com/dapr/kit/ptr` instead)
3. Fixed error messages for denied deps

Port of dapr/dapr#5438


mukundansundar (commented):
Can you add https://github.com/pkg/errors to the deny list as well?

berndverst (dismissed):

berndverst (approved):

PR #2255, opened 2022-11-04
Fix to Hashicorp secret store vault getSecret error

shivanisinghnitp:
Signed-off-by: Shivani Singh <shivasingh@microsoft.com>

# Description

For the issue 4779 -> default value of vaultKVUsePrefix is true, due to which, BulkGetSecret method can't be used until it's set to false. Instead of having the user change that in the YAML, if we by default keep the value to be false, it would not cause issues such as 4779 to the users

## Issue reference

#4779
https://github.com/dapr/dapr/issues/4779

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


secretstores/hashicorp/vault/vault.go
```diff
@@ -166,7 +166,7 @@ func (v *vaultSecretStore) Init(metadata secretstores.Metadata) error {
 		}
 	}
 
-	if !convertedVaultKVUsePrefix {
+	if convertedVaultKVUsePrefix {
```
pravinpushkar, line 169: @shivanisinghnitp this change looks similar to the existing one. Is default value of `convertedVaultKVUsePrefix` being read somewhere else. Please make sure this is tested out.

akhilac1 (changes requested):
@shivanisinghnitp  - Hashicorp vault documentation states the follows - 
'The path prefix (or alternatively the -mount flag for vault kv commands) tells Vault which secrets engine to which it should route traffic. When a request comes to Vault, it matches the initial path part using a longest prefix match and then passes the request to the corresponding secrets engine enabled at that path. Vault presents these secrets engines similar to a filesystem.'

The existing default of 'true' is appropriate considering the way Hashicopr Vault is designed. Also, the referenced issue does not discuss any limitations related to this attribute. 

Default is also clearly documented in dapr docs - https://docs.dapr.io/reference/components-reference/supported-secret-stores/hashicorp-vault/

I do not see this as a right approach to reset the default to false.Please revisit the proposed PR.

berndverst:
I agree with @akhilac1. Also this would be a breaking change for several users of this component today.

I do not agree with changing the default value.

Surely, there is a different way to make `BulkGetSecret` work -- perhaps the implementation of that function must change.

secretstores/hashicorp/vault/vault.go
berndverst: Should this line remain unchanged?

berndverst (changes requested):
On behalf of Contrib maintainers I do not approve of this change.

`BulkGetSecret` can be used to fetch all secrets with the `dapr` prefix by default. That is the correct behavior. If you wish to not use a prefix you need to configure the component to disable the prefix.

By default the `dapr` prefix (or provided prefix) should be continued to be used unless use of prefix is explicitly disabled.

cc @yaron2

Issue #2014, opened 2022-08-25
Proposal: Structured component metadata info requirement


berndverst:
## Describe the proposal

This is a proposal to capture and require component metadata information to be provided alongside each component in a structured manner according to a defined schema.

## Why
This has several advantages:
- Automatically generate component documentation to be in line with components-contrib code. No more manual creation of docs PRs for component metadata.
- We can generate a schema of all known components which can be used to provide better developer tooling, e.g. this could enable the validation of component.yaml files.

## How
Alongside the code of each component a `metadata.yaml` should be included which describes all metadata options, whether they are required, what type of values they accept, etc
A schema definition additional enables authoring of these `metadata.yaml` files with tools such as VS Code.

An example example schema for how component metadata can and should be described can be found here: https://github.com/ItalyPaleAle/dapr-components-contrib/blob/component-info-sample/.build-tools/pkg/metadataschema/schema.go

Here are some specific examples of what this would look like.

Example state store with multiple connection options:

```yaml
schemaVersion: v1
type: state
name: azure.cosmosdb
version: v1
status: stable
title: "Azure Cosmos DB (SQL API)"
urls:
  - title: Reference
    url: https://docs.dapr.io/reference/components-reference/supported-state-stores/setup-azure-cosmosdb/
capabilities:
  # If actorStateStore is present, the metadata key actorStateStore can be used
  - actorStateStore
  - crud
  - transactional
  - etag
  - ttl
  - query
authenticationProfiles:
  - title: "Master key"
    description: "Authenticate using a pre-shared \"master key\"."
    metadata:
      - name: masterKey
        required: true
        sensitive: true
        description: "The key to authenticate to the Cosmos DB account."
        example: '"my-secret-key"'
  - title: "Azure AD: Managed identity"
    description: "Authenticate using Azure AD and a managed identity."
    metadata:
      - name: azureClientId
        required: false
        description: "Client ID (application ID). Required if the service has multiple identities assigned."
        example: '"c7dd251f-811f-4ba2-a905-acd4d3f8f08b"'
  - title: "Azure AD: Client credentials"
    description: "Authenticate using Azure AD with client credentials, also known as \"service principals\"."
    metadata:
      - name: azureTenantId
        required: true
        description: "ID of the Azure AD tenant"
        example: '"cd4b2887-304c-47e1-b4d5-65447fdd542b"'
      - name: azureClientId
        required: true
        description: "Client ID (application ID)"
[...]

PR #2257, opened 2022-11-05
Refactors all state store to standardize metadata parsing

berndverst:
# Description

This relates to the larger theme of #2014 which is about defining component metadata.

Standardizes all state store component metadata definitions. Now all component metadata is obtained via the helper method `metadata.DecodeMetadata(input interface{}, result interface{}) error`

Additionally, this decoding function was updated to properly handle string arrays, and additionally to use our `isTruthy` function by default when converting from string to `bool` or `*bool`.

The state store interface now has a function `GetComponentMetadata() map[string]string` which returns a map of metadata fields with their respective types. This honors the `mapstructure` tag overrides.

Under the hood the following internal helper function is used to traverse a struct to list struct fields and tags:
`metadata.GetMetadataInfoFromStructType(t reflect.Type, metadataMap *map[string]string) error`

Please note:
For metadata parsing to work correctly all struct fields must be exported!


berndverst:
Let's see whether I broke any conformance or certification tests... This is my first run :)

berndverst:
Ah.. because I modified the interface it seems I need to update certification tests!

berndverst:
FYI Re ran the eventhubs bindings test above

artursouza (approved):

berndverst:
In a few cases I added new metadata, such as the ability to set the tableName instead of the default `dapr`. This needs to be documented. (Postgres is an example)

Issue #2261, opened 2022-11-08
kubernetes nameresolution customize configuration not working


havardzzl:
## Expected Behavior

change cluster domain configuration works properly.

## Actual Behavior

change kubernetes nameresolution cluster domain configuration not working.

## Steps to Reproduce the Problem

set a Configuration spec as follows:
spec:
  nameResolution:
    component: kubernetes
    version: v1
    configuration:
      clusterDomain: fake-cluster.local
use it in a dapr pod with sidecar. It should not works

## Release Note

RELEASE NOTE:**FIX** Bug in kubernetes nameresolution initialize.

havardzzl:
I read the code and fix it in my own project.
https://github.com/dapr/components-contrib/compare/master...havardzzl:components-contrib:master

PR #2262, opened 2022-11-08
fix kubernetes nameresolution customize configuration

havardzzl:
# Description

fix bug in kubernetes nameresolution initialization

## Issue reference

https://github.com/dapr/components-contrib/issues/2261

Please reference the issue this PR will close: #2261

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


nameresolution/kubernetes/kubernetes.go
```diff
@@ -45,10 +45,13 @@ func (k *resolver) Init(metadata nameresolution.Metadata) error {
...
 	}
-	if config, ok := configInterface.(map[string]string); ok {
-		clusterDomain := config[ClusterDomainKey]
-		if clusterDomain != "" {
-			k.clusterDomain = clusterDomain
+	if config, ok := configInterface.(map[string]interface{}); ok {
+		clusterDomainPtr := config[ClusterDomainKey]
+		if clusterDomainPtr != nil {
+			clusterDomain, _ := clusterDomainPtr.(string)
+			if clusterDomain != "" {
```
daixiang0, line 52: Could we add this case to test?

havardzzl: I change the test case, the bug is because when the dapr initialize the nameResolution resolver it pass the metadata.Configuration as map[string]interface{} instead of map[string]string

ItalyPaleAle (approved):

ItalyPaleAle (dismissed):
lgtm but I wonder if this should be fixed in the runtime (Dapr/dapr) to pass a map[string]string like we do for all other components?

nameresolution/kubernetes/kubernetes.go
```diff
@@ -45,10 +45,13 @@ func (k *resolver) Init(metadata nameresolution.Metadata) error {
 	if err != nil {
 		return err
 	}
-	if config, ok := configInterface.(map[string]string); ok {
-		clusterDomain := config[ClusterDomainKey]
-		if clusterDomain != "" {
-			k.clusterDomain = clusterDomain
+	if config, ok := configInterface.(map[string]interface{}); ok {
+		clusterDomainPtr := config[ClusterDomainKey]
```
berndverst, line 49: `Ptr` is a bad name for this. It's not a pointer but an interface value that has not been cast yet.

Can you please change the variable name?

berndverst (changes requested):

berndverst:
> lgtm but I wonder if this should be fixed in the runtime (Dapr/dapr) to pass a map[string]string like we do for all other components?

@ItalyPaleAle 
The problem is the `config.Normalize` function. I don't think that's even necessary.. this function is what converts the `map[string]string` to `map[string]interface`. I have never seen this function used in any other component!

berndverst (approved):

berndverst:
@ItalyPaleAle I patched up this PR to use our standardized metadata parsing.

Interestingly, the `clusterDomain` metadata option is not documented anywhere.

berndverst (approved):

berndverst:
Oh my bad -- this is not component metadata but from Dapr Config. This PR is ok as is.

Issue #2266, opened 2022-11-10
[PubSub AWS/SqsSns] daprd shutdown leaves remaining competing consumers orphaned from Topic


olitomlinson:
## Expected Behaviour

Given App X is subscribed to Topic A
Given App X has 2 or more replicas active
When any one of the replicas is shutdown (scale in for example)
**Then the remaining active replicas are still able to consume events from the Topic in a competing consumer fashion**

## Actual Behaviour

Given App X is subscribed to Topic A
Given App X has 2 or more replicas active
When any one of the replicas is shutdown (scale in for example)
**Then the remaining replicas are orphaned from the Topic and no longer consume events in ANY fashion from that Topic**

## Steps to Reproduce the Problem

- Scale an App which is consuming from a Dapr Topic to 2 or more replicas
- Scale down the App by one or more replicas
- Send a message to the topic, which should be consumed by one of the remaining replicas
- Observe that the message is never consumed (It can't be consumed because the queue is deleted so the message is dropped by the SNS Topic)

- If you perform the above operations while refreshing the AWS console, you can see the subscription disappear after scale-in, and re-appear on scale-out.

---
**Code area**

This looks to be the code relating to deleting the Subscription on daprd shutdown(?) 

https://github.com/dapr/components-contrib/blame/ce87730a7362a71a1a0b81619a0703c8ae7b65e9/pubsub/aws/snssqs/snssqs.go#L848

---

**Additional detail** 

The same outcome is observed when redeploying just a single pod (new pod comes up first along side the old, old pod is disposed, deleting the subscription in the process, new pod has no subscription)

We are using the `consumerId : {namespace}` metadata configuration for this component

This is blocking our adoption of AWS Sqs/Sns for PubSub.

olitomlinson:
cc @yaron2

ItalyPaleAle:
I think your analysis is correct and that may be the cause of the issue. I will need to investigate this once I’m back, we can triage it for Dapr 1.10 as this is an alpha component

yaron2:
Beta component since 1.6.

ItalyPaleAle:
My bad. Still, not a stable component.
How Is this beta by the way since we don’t have a conf test for this?

PR #2268, opened 2022-11-10
Fix deletion of sqs queue regression

yaron2:
Multiple users of the AWS SQS component reported a regression caused by this commit: https://github.com/dapr/components-contrib/commit/704f4dd7307d96252d7cc92cf792fcfdebea3c6c.

Fixes https://github.com/dapr/components-contrib/issues/2266.


ItalyPaleAle (approved):

berndverst:
I did troubleshoot this with the affect user in depth.

Unfortunately the same ARN is always returned for the same Topic / Queue combo. So it is not possible to have multiple clients using the same queue and topic.

The unsubscribe request would disconnect all clients using the same queue and topic.

berndverst (approved):

PR #2269, opened 2022-11-10
Updating Temporal Init to allow for custom namespaces

RyanLettieri:
# Description

Updated Temporal Init to allow for custom namespaces

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/workflows/temporal/temporal.go
+++ b/workflows/temporal/temporal.go
@@ -87,7 +87,7 @@ func (c *TemporalWF) Start(ctx context.Context, req *workflows.StartRequest) (*w
 	taskQ := req.Options["task_queue"]
 
 	opt := client.StartWorkflowOptions{ID: req.WorkflowReference.InstanceID, TaskQueue: taskQ}
-	run, err := c.client.ExecuteWorkflow(ctx, opt, req.WorkflowName, req.Parameters)
+	run, err := c.client.ExecuteWorkflow(ctx, opt, req.WorkflowName, req.Input)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


workflows/temporal/temporal.go
```diff
@@ -58,6 +59,9 @@ func (c *TemporalWF) Init(metadata workflows.Metadata) error {
 	if m.Identity != "" {
 		cOpt.Identity = m.Identity
 	}
+	if m.Namespace != "" {
+		cOpt.Namespace = m.Namespace
```
johnewart, line 63: Do we need to explicitly specify a default, or will it default to 'default'? ... That's a lotta defaults

johnewart (dismissed):
Assuming we don't need to specify a default, this looks fine, otherwise we should add one

workflows/temporal/temporal.go
RyanLettieri: By default, it will default to 'default': https://pkg.go.dev/go.temporal.io/sdk@v1.17.0/client#pkg-constants

berndverst: So `parseMetadata` should actually be using `contrib/metadata`'s `DecodeMetadata` function which takes the `metadata.Properties` from `Init` and unmarshals them into the struct.

The default value in the struct will be the empty string.

Please make sure that when passing an empty string it defaults to whatever you need it to be. Otherwise you do indeed need to assign defaults!

cgillum (dismissed):
LGTM, assuming the defaulting works as intended.

artursouza (approved):

berndverst (approved):

PR #2273, opened 2022-11-11
[azappconfig] incorporate OnlyIfChanged flag from sdk

pravinpushkar:

# Description

This PR incorporates the fix created for this issue - https://github.com/Azure/azure-sdk-for-go/issues/19297
Basically, `GetSetting` will only fetch the new values if the value has been modified in az appconfig store.


configuration/azure/appconfig/appconfig.go
```diff
@@ -295,15 +295,21 @@ func (r *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.S
...
 func (r *ConfigurationStore) doSubscribe(ctx context.Context, req *configuration.SubscribeRequest, handler configuration.UpdateHandler, sentinelKey string, id string) {
+	var etagVal *azcore.ETag
 	for {
 		// get sentinel key changes
-		_, err := r.Get(ctx, &configuration.GetRequest{
-			Keys:     []string{sentinelKey},
-			Metadata: req.Metadata,
-		})
+		resp, err := r.client.GetSetting(
+			context.TODO(),
```
DeepanshuA, line 302: We have a context already available here, do we require TODO here?

configuration/azure/appconfig/appconfig.go
```diff
@@ -295,15 +295,21 @@ func (r *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.S
...
+			sentinelKey,
+			&azappconfig.GetSettingOptions{
+				Label:         r.getLabelFromMetadata(req.Metadata),
+				OnlyIfChanged: etagVal,
+			},
+		)
 		if err != nil {
 			r.logger.Debugf("azure appconfig error: fail to get sentinel key changes or sentinel key's value is unchanged: %s", err)
```
DeepanshuA, line 310: Should we print for which `sentinelKey`, value is unchanged.

configuration/azure/appconfig/appconfig.go
berndverst: Probably should use the context provided here.

configuration/azure/appconfig/appconfig.go
```diff
@@ -295,15 +295,21 @@ func (r *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.S
 }
 
 func (r *ConfigurationStore) doSubscribe(ctx context.Context, req *configuration.SubscribeRequest, handler configuration.UpdateHandler, sentinelKey string, id string) {
+	var etagVal *azcore.ETag
```
berndverst, line 298: This is always an empty Etag value... doesn't that mean you keep getting updates all the time?

It seems you should be storing the etag somewhere and be using this in subsequent calls using this `OnlyIfChanged` option.

berndverst (changes requested):

berndverst:
Also run `make modtidy-all`

configuration/azure/appconfig/appconfig.go
ItalyPaleAle: Actually should be a children context from `ctx` that has a timeout.

`getCtx, getCancel := context.WithTimeout(ctx, timeoutInSeconds)`

configuration/azure/appconfig/appconfig.go
```diff
@@ -295,15 +295,21 @@ func (r *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.S
...
-		})
+		resp, err := r.client.GetSetting(
+			context.TODO(),
+			sentinelKey,
+			&azappconfig.GetSettingOptions{
+				Label:         r.getLabelFromMetadata(req.Metadata),
+				OnlyIfChanged: etagVal,
+			},
+		)
 		if err != nil {
```
ItalyPaleAle, line 309: Please add a line here to not log errors if `errors.Is(err, context.Canceled)` or when the component is shutting down, an error would be logged

configuration/azure/appconfig/appconfig.go
```diff
@@ -295,15 +295,21 @@ func (r *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.S
...
 			r.logger.Debugf("azure appconfig error: fail to get sentinel key changes or sentinel key's value is unchanged: %s", err)
 		} else {
+			etagVal = resp.ETag
 			items, err := r.Get(ctx, &configuration.GetRequest{
```
ItalyPaleAle, line 313: Here too the context should use a timeout perhaps, or it could hang forever

configuration/azure/appconfig/appconfig.go
```diff
@@ -295,15 +295,21 @@ func (r *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.S
...
 				Keys:     req.Keys,
 				Metadata: req.Metadata,
```
ItalyPaleAle, line 315: On the error check below, please there too ignore errors for context canceled (when the component is shutting down)

configuration/azure/appconfig/appconfig.go
pravinpushkar: Thanks for catching. There was one issue that I had created regarding context in az SDK sometime back - https://github.com/Azure/azure-sdk-for-go/issues/19223
[inline comments] https://github.com/dapr/components-contrib/blob/7295b3acc16c6dea35371da39a9df6ef90af16b9/configuration/azure/appconfig/appconfig.go#L206

It was fixed few weeks back, will make the change for this.

pravinpushkar:
@berndverst @ItalyPaleAle @DeepanshuA Updated the PR with a new commit. Please check.

configuration/azure/appconfig/appconfig.go
```diff
@@ -248,10 +247,8 @@ func (r *ConfigurationStore) getAll(ctx context.Context, req *configuration.GetR
 		},
 		nil)
 
-	// TODO: here contxt.TODO() is used because the SDK panics when a cancelled context is passed in NextPage
-	// Issue - https://github.com/Azure/azure-sdk-for-go/issues/19223 . It needs to be modified to use ctx once the SDK is fixed
 	for allSettingsPgr.More() {
-		if revResp, err := allSettingsPgr.NextPage(context.TODO()); err == nil {
+		if revResp, err := allSettingsPgr.NextPage(timeoutContext); err == nil {
```
ItalyPaleAle, line 251: I think you need a new `timeoutCtx` for each call, not one for the method. This is because each `NextPage` invocation makes a network call, so if you have a single timeout, getting a page after the first may fail.

configuration/azure/appconfig/appconfig.go
```diff
@@ -295,21 +292,32 @@ func (r *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.S
...
+			ctx,
+			sentinelKey,
+			&azappconfig.GetSettingOptions{
+				Label:         r.getLabelFromMetadata(req.Metadata),
+				OnlyIfChanged: etagVal,
+			},
+		)
 		if err != nil {
-			r.logger.Debugf("azure appconfig error: fail to get sentinel key changes or sentinel key's value is unchanged: %s", err)
+			if !errors.Is(err, context.Canceled) {
```
ItalyPaleAle, line 307: If the error is context.Canceled(), shouldn't the method return?

configuration/azure/appconfig/appconfig.go
```diff
@@ -295,21 +292,32 @@ func (r *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.S
...
 		} else {
+			// if sentinel key has changed then update the Etag value.
+			etagVal = resp.ETag
 			items, err := r.Get(ctx, &configuration.GetRequest{
 				Keys:     req.Keys,
 				Metadata: req.Metadata,
 			})
 			if err != nil {
-				r.logger.Errorf("azure appconfig error: fail to get configuration key changes: %s", err)
+				if !errors.Is(err, context.Canceled) {
```
ItalyPaleAle, line 318: Here too, should this return if the context is canceled?

ItalyPaleAle (changes requested):

configuration/azure/appconfig/appconfig.go
berndverst: cool, somehow missed that even though I was looking for it.

configuration/azure/appconfig/appconfig.go
pravinpushkar: Yes, agree. My mistake

pravinpushkar: @ItalyPaleAle

ItalyPaleAle: You're not wrong, but I do believe that we should short-circuit when we can. This way if someone adds more code in the future, they are less likely to accidentally introduce a bug.

pravinpushkar:
@berndverst @ItalyPaleAle all review comments are addressed .

ItalyPaleAle (approved):

berndverst (approved):

Issue #2008, opened 2022-08-15
Does dapr support the AMPQs (Secure AMPQ) protocol when using rabbitmq for pubsub?


bab5470:
Does dapr support the AMPQs (Secure AMPQ) protocol when using rabbitmq for pubsub?

When we specify the following in our dapr pubsub config our setup works:

```
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: pubsub
  namespace: backend
spec:
  type: pubsub.rabbitmq
  version: v1
  metadata:
  - name: host
    value: "amqp://username:password@rabbitmq.data.svc.cluster.local:5672"
  - name: exchangeKind
    value: topic
  - name: durable
    value: true
  - name: deleteWhenUnused
    value: false
```

When we switch the config as follows it breaks:

```
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: pubsub
  namespace: backend
spec:
  type: pubsub.rabbitmq
  version: v1
  metadata:
  - name: host
    value: "amqps://username:password@rabbitmq.data.svc.cluster.local:5671"
  - name: exchangeKind
    value: topic
  - name: durable
    value: true
  - name: deleteWhenUnused
    value: false
```

Note all we've changed is the protocol from ampq to ampqs and the port from 5672 to 5671.

We are using our own internal CA when securing rabbitmq so I don't know if there is some way to get the daprd sidecar to trust our internal CA's cert when connecting to rabbitmq?

Thanks,
Brad

yaron2:
Moved to correct repository. Thanks for reporting this @bab5470, it's currently not supported it seems.

anoymouscoder:
According to my findings, currently secure amqp is supported. Based on [official rabbitmq documentation](https://www.rabbitmq.com/uri-query-parameters.html), you can set the parameters for tls in the uri, eg. `amqps://localhost:5671?cacertfile=/path/to/ca_cert.pem&keyfile=/path/to/key.pem&certfile=/path/to/cert.pem`.
We are currently using the `dial` method of amqp library(ref: https://github.com/dapr/components-contrib/blob/bcea284c7b0c1a5a07294021a2bb6cc74909da35/bindings/rabbitmq/rabbitmq.go#L79), which supports the establishment of TLS connections. 
I've done tests in my local development env and found nothing wrong.

anoymouscoder:
@yaron2 @bab5470 Could you confirm it, not sure I fully understand the intent of this issue.

[...]

PR #2276, opened 2022-11-13
Add rabbitmq tls metadata

SpikeWong:
Signed-off-by: spike <hello@spike.wiki>

# Description

Add TLS related metadata configuration to rabbitmq

## Issue reference

 #2008
 #2246 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#2953


pubsub/rabbitmq/metadata.go
```diff
@@ -97,6 +101,14 @@ func createMetadata(pubSubMetadata pubsub.Metadata, log logger.Logger) (*metadat
...
+		uri, err := amqp.ParseURI(result.connectionString)
+		if err != nil {
+			return &result, fmt.Errorf("%s invalid connection string: %s, err: %w", errorMessagePrefix, result.connectionString, err)
+		}
+		result.protocol = uri.Scheme
+	}
+
 	if val, found := pubSubMetadata.Properties[metadataProtocolKey]; found && val != "" {
 		result.protocol = val
 	}
```
anoymouscoder, line 114: Not sure if I need to check that the protocol parsed from the uri is the same as the configured protocol. If not, raised an error?

berndverst: I don't think that's necessary. The protocol metadata key can be an explicit override for users who know what protocol they want to use.

berndverst (changes requested):
Thanks for the PR!

Please address the linter issues. (See build Linux_amd64 action run)

SpikeWong:
> Can you please add a unit test for the following cases:
> 
> * Connection URI set (with protocol) & protocol is not set -> protocol is correctly inferred
> * Connection URI with missing protocol part & protocol is not set -> some error (or default?)
> * Connection URI (with protocol) & protocol is set to something else -> the explicit protocol value is used

I'm not sure if we should override the protocol in the connectionString when both the connectionString and the protocol are set.
According to this func, the corresponding value in connectionString is currently not overwritten if username, host and connectionString are set at the same time.
So if we were to do this with the protocol, would we also change the behaviour of the other fields https://github.com/dapr/components-contrib/blob/9870c5e33beb8706a6e76e5c95d55a39b40d7fa3/pubsub/rabbitmq/metadata.go#L233-L250

berndverst:
@SpikeWong ok, in that case you need to compare the protocol between `connectionString` and `protocol` metadata and throw an error if these mismatch.

SpikeWong:
> @SpikeWong ok, in that case you need to compare the protocol between `connectionString` and `protocol` metadata and throw an error if these mismatch.

finished

pubsub/rabbitmq/metadata.go
```diff
@@ -203,6 +218,11 @@ func createMetadata(pubSubMetadata pubsub.Metadata, log logger.Logger) (*metadat
 		result.defaultQueueTTL = &ttl
 	}
 
+	result.TLSProperties, err = pubsub.TLS(pubSubMetadata.Properties)
+	if err != nil {
+		return &result, fmt.Errorf("%s invalid TLS configuration: %w", errorMessagePrefix, err)
+	}
+
```
Taction, line 225: If the protocol is AMQPS, check if TLS Properties are provided. If not, return the error info to inform users that the tls config is missed for amqps protocol.

SpikeWong: Based on https://www.rabbitmq.com/ssl.html#enabling-tls, if rabbitmq server configure `ssl_options.verify` as `verify_none` or `ssl_options.fail_if_no_peer_cert` as `false`, the client is allowed to use amqps protocol without certificates.

yaron2 (approved):

yaron2:
LGTM. @berndverst see if you're happy.

berndverst (approved):

Issue #2271, opened 2022-11-11
Add bulk publish support to in-memory pubsub component


shubham1172:
## Describe the feature

Add Bulk Publish support (dapr/dapr#2218) to the in-memory pubsub component.

## Release Note

RELEASE NOTE: **ADD** Bulk Publish support to in-memory pubsub component

shubham1172:
Closing this since the current in-mem package EventBus does not have a good way to do this, and the workaround is similar to the BulkPublish implementation provided by Dapr runtime for its components.

shubham1172:
Closing this since the current in-mem package EventBus does not have a good way to do this, and the workaround is similar to the BulkPublish implementation provided by Dapr runtime for its components.

PR #2277, opened 2022-11-14
[WIP] Add bulk publish support to in-memory pubsub component

shubham1172:
# Description

**Note**, still evaluating if this PR is needed since the default implementation in runtime already does the same. Since EventBus does not have support for bulk-publish directly, this might be unnecessary.

This PR adds a bulk publish support to the in-memory pubsub component. It publishes multiple messages by invoking the EventBus package's [publish](https://pkg.go.dev/github.com/asaskevich/EventBus#EventBus.Publish) operation multiple times.

It also onboards the "bulkpublish" operation to the conformance test for in-memory pubsub.

## Issue reference

Please reference the issue this PR will close: #2271 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: https://github.com/dapr/docs/issues/2954

### DIFF
--- a/pubsub/in-memory/in-memory.go
+++ b/pubsub/in-memory/in-memory.go
@@ -53,6 +53,13 @@ func (a *bus) Publish(req *pubsub.PublishRequest) error {
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


shubham1172:
Closing this since it's not providing any value over the existing default implementation.

PR #2286, opened 2022-11-16
ASQ Binding: Cert Test: Add queue visibilityTimeout test

berndverst:

# Description

Adds certification test case for Azure Storage Queues Binding's queue visibilityTimeout option.

Also fixes major bug in certification test framework. Previously `watcher.FailIfNotExpected` which is used to ensure no unexpected data / events were delivered in the case of an Unordered Queue never actually failed when unexpected data was sent.

This will likely result in other certification tests failing, but if so I will be opening separate issues for those failures and ask various team members to address accordingly.


ItalyPaleAle (approved):

berndverst:
FYI @mukundansundar

Issue #2264, opened 2022-11-09
Mqtt pubsub components fails to read messages on startup when multiple subscriptions configured


bwojdyla:
## Expected Behavior

Messages that are persisted on queue while dapr sidecar was down should be processed after dapr sidecar is started.

## Actual Behavior

After darp sidecar is started messages which are one queue are not processed.

## Steps to Reproduce the Problem

Configure 2 subscribtions for mqtt component:
```
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: testapp-pubsub
spec:
  type: pubsub.mqtt
  version: v1
  metadata:
  - name: url
    value: "abc"
  - name: qos
    value: 1
  - name: retain
    value: "false"
  - name: cleanSession
    value: "false"
  - name: backOffMaxRetries
    value: "5"
```

Subscription 1

```
apiVersion: dapr.io/v1alpha1
kind: Subscription
metadata:
  name: subscription1 
spec:
  pubsubname: testapp-pubsub
  topic: topic1
  route: /api/route1
```

Subscription 2

```
apiVersion: dapr.io/v1alpha1
kind: Subscription
metadata:
  name: subscription2
spec:
  pubsubname: testapp-pubsub
  topic: topic2
  route: /api/route2
```
1. Start dapr + application for first time so that it creates queue
2. Stop dapr
3. Send at least 2 messages for topic1
4. Start dapr for second time
5. You should see similar behavior:
```
time="2022-11-08T16:27:09.771556034Z" level=info msg="app is subscribed to the following topics: [topic1 topic2] through pubsub=testapp-pubsub" app_id=test.testapp.api instance=536c465db880 scope=dapr.runtime type=log ver=1.9.3
time="2022-11-08T16:27:09.771606899Z" level=debug msg="subscribing to topic='topic1' on pubsub='testapp-pubsub'" app_id=test.testapp.api instance=536c465db880 scope=dapr.runtime type=log ver=1.9.3
time="2022-11-08T16:27:09.771680668Z" level=info msg="initializing the subscriber" app_id=test.testapp.api instance=536c465db880 scope=dapr.contrib type=log ver=1.9.3
time="2022-11-08T16:27:09.777840715Z" level=debug msg="subscribing to topic='topic2' on pubsub='testapp-pubsub'" app_id=test.testapp.api instance=536c465db880 scope=dapr.runtime type=log ver=1.9.3
[...]

PR #2291, opened 2022-11-17
[hotfix-1.9] Upgrade mqtt component

pravinpushkar:
# Description

There has been changes in  upstream mqtt sdk to handle a race condition in https://github.com/dapr/components-contrib/blob/84cd16677fc555570f7f261f0a6bb0bb1c774b42/pubsub/mqtt/mqtt.go#L193

This issue is very specific to steps described in the issue #2264 i.e. When there are two subscriptions and messages are being published only to the 1st topic(If we reverse it or introduced even an extra route for any 3rd topic for which there is no subscription it does not get reproduced) and subscriber gets killed after publishing few message. If the subscriber comes online and there has been few (>2) messages published already then it only delivers the 1st pending message and gets stuck.

The reason is when it will try to subscribe for the 2nd topic, it will try to reset the subscription and call `m.consumer.Disconnect` where it will get stuck.

Below are the logs observed after subscriber is restarted - (logs statements are introduced just for understanding purpose)

msg="app is subscribed to the following topics: [A B] 

msg="subscribing to topic='A' 

msg="initializing the subscriber" 

handlerForTopic: Acquiring lock for ###########  A

msg="subscribing to topic='B' 

handlerForTopic: Acquired lock for ############ A

msg="Processing MQTT message A#4 (retained=false)" 

msg="re-initializing the subscriber" 

mqtt closing subscription through m.consumer.Disconnect ############ - (this never gets completed)
```go
func (m *mqttPubSub) closeSubscription() {
	fmt.Println("mqtt closing subscription through m.consumer.Disconnect ############")
	m.consumer.Disconnect(5)
	fmt.Println("mqtt closed subscription through m.consumer.Disconnect #############")
	m.consumer = nil
}
```

msg="Done processing MQTT message A#4; sending ACK" (sends the 1st pending message here)

handlerForTopic: Acquiring lock for ###########  A. - (it gets stuck here)

## Issue reference

Please reference the issue this PR will close: #2264

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/pubsub/mqtt/mqtt.go
+++ b/pubsub/mqtt/mqtt.go
@@ -247,7 +247,6 @@ func (m *mqttPubSub) startSubscription(ctx context.Context) error {
 // onMessage returns the callback to be invoked when there's a new message from a topic
 func (m *mqttPubSub) onMessage(ctx context.Context) func(client mqtt.Client, mqttMsg mqtt.Message) {
 	return func(client mqtt.Client, mqttMsg mqtt.Message) {
-
 		ack := false
 		defer func() {
 			// Do not send N/ACKs on retained messages
--- a/pubsub/mqtt/mqtt.go
+++ b/pubsub/mqtt/mqtt.go
@@ -247,8 +247,6 @@ func (m *mqttPubSub) startSubscription(ctx context.Context) error {
 // onMessage returns the callback to be invoked when there's a new message from a topic
 func (m *mqttPubSub) onMessage(ctx context.Context) func(client mqtt.Client, mqttMsg mqtt.Message) {
 	return func(client mqtt.Client, mqttMsg mqtt.Message) {
-		// Turn off auto-ACK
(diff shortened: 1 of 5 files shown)


berndverst (approved):

pubsub/mqtt/mqtt.go
```diff
@@ -336,6 +333,8 @@ func (m *mqttPubSub) connect(ctx context.Context, clientID string) (mqtt.Client,
 		return nil, err
 	}
 	opts := m.createClientOptions(uri, clientID)
+	// Turn off auto-ack
+	opts.SetAutoAckDisabled(true)
```
ItalyPaleAle, line 337: Does this need to be changed in master too?

pravinpushkar: It is already there in master.

Issue #2101, opened 2022-09-20
EventHubs components to use Azure SDK for Go ("Track 2 SDK")


berndverst:
The EventHubs components should be refactored to use the Azure SDK for Go (Track 2 SDK)

https://github.com/Azure/azure-sdk-for-go/tree/sdk/messaging/azeventhubs/v0.2.0

(Check for a newer version when working on this issue)

As part of this issue support for AAD auth should be added - see #1103.

berndverst:
Consider adding https://github.com/dapr/components-contrib/issues/2176 (Websocket connection support) with the new SDK.

berndverst:
Ping @ItalyPaleAle  :)

ItalyPaleAle:
It's on my TODO for next week!

PR #2298, opened 2022-11-21
Receiving of custom properties from IoT Hub through endpoint that is compatible with Event Hubs

mecoding1:
Currently, we can receive only the system properties from IOT Hub. Following changes should allow the retrieval of all custom properties from Azure IoT Hub through [endpoint that is compatible with Event Hubs](https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-messages-read-builtin#read-from-the-built-in-endpoint) . It is based on metadata key : _**requireAllProperties**_.

Signed-off-by: mecoding1 <118708378+mecoding1@users.noreply.github.com>

# Description

Currently,[ IoT Hub supported headers for the events hub component only forwards the Application properties](https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-azure-eventhubs/#subscribing-to-azure-iot-hub-events).
For the Custom Properties , added a new key under metadata which if true retrieves & forwards all the custom properties which was shared. 

If _requireAllProperties: true_  (  accepts either true or false ) , then it looks for `Properties   map[string]interface{}` in the [Event  structure](https://github.com/Azure/azure-event-hubs-go/blob/571caa480fa9150983303f8cbc19147b027d0fec/event.go#L93) , where the custom properties of IoT Hub are assigned.
In such scenarios both [System Properties of D2C IoT Hub messages](https://learn.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-messages-construct#system-properties-of-d2c-iot-hub-messages) and [Application Properties of D2C IoT Hub messages](https://learn.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-messages-construct#application-properties-of-d2c-iot-hub-messages) will be received and forwarded by the component to the route.

By default or if not provided, it works as it is used in the previous implementation i.e. only the Application Properties are forwarded.

Sample _subscription.yaml_ file will look like the following if we want to extract the custom properties of IoT Hub:

```
apiVersion: dapr.io/v2alpha1
 kind: Subscription
 metadata:
   name: <myevent-subscription-name>
 spec:
   pubsubname: <eventhubs-pubsub-name>
   topic: <topic-name>
   metadata:
     rawPayload: false
     requireAllProperties: true
   route: /endpoint
 scopes:
   - <dapr-scope>
```

## Issue reference

This PR will address : [#3642](https://github.com/dapr/dapr/issues/3642) for Azure Event Hub compatible endpoint for Azure IoT Hub

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -124,6 +127,16 @@ func subscribeHandler(ctx context.Context, topic string, e *eventhub.Event, hand
 	if e.ID != "" {
 		res.Metadata[sysPropMessageID] = e.ID
 	}
+	// added properties if any ( includes application properties from iot-hub)
+	if getAllProperties{
+		if e.Properties != nil && len(e.Properties) > 0{
+			for key, value := range e.Properties {
+				if str, ok := value.(string); ok {
+						res.Metadata[key] = str
```
berndverst, line 135: is it possible for these property names to conflict with other metadata names? If so this step would be replacing them...

berndverst:
We are going to work on #2101 during this cycle which will replace the EventHubs SDK. This change might more sense after that is complete.

@ItalyPaleAle since you will be working on those tasks, what is your preference here?

ItalyPaleAle:
I’m ok with merging this now (once your feedback is addressed). I haven’t started work on #2101 yet

berndverst (changes requested):
Please add an integration test (certification test case) here: https://github.com/dapr/components-contrib/tree/master/tests/certification/pubsub/azure/eventhubs

You will need to define a new `component.yaml` file (look at the other examples) and then explicitly set the new metadata field.

In your test you will need to send some data to IoT hub (there should also be a test script for this) and then assert that with your new metadata option turned on you are receiving the expected IoT Hub properties.

berndverst:
@ItalyPaleAle sounds good. Will not merge this without new certification test case however. That test case will also help with retaining the functionality when the SDK migration occurs.

mecoding1:
@berndverst For the certification tests, could you pls point towards where we are checking for the specific system properties shared in the [script  ](https://github.com/dapr/components-contrib/blob/4dde34c2d14b377aa9c52630951beea9f6f6b8c2/tests/certification/pubsub/azure/eventhubs/send-iot-device-events.sh#L44) system properties as mentioned [here ](https://github.com/dapr/components-contrib/blob/d8c2ce47fdb49db8e9ba3ad825d0f9492c0e6f27/tests/certification/pubsub/azure/eventhubs/README.md?plain=1#L64) ? 

As for the new component.yaml , the new metadata key  is part of the subscription.yaml ( similarly as rawPayload ).

pubsub/azure/eventhubs/eventhubs.go
mecoding1: The custom application properties will be a dictionary of string properties that the device ( or client ) will define itself other than the ones already pre-defined under system properties. I believe that the possibility of chances of intentional conflict among them is maybe impractical. 

List of predefined system properties : [System Properties of D2C IoT Hub messages](https://learn.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-messages-construct#system-properties-of-d2c-iot-hub-messages)

berndverst:
It seems the individual properties on the IoT Hub messages are never verified anywhere.

This is where the messages are sent and the list of expected messages is created.
https://github.com/dapr/components-contrib/blob/4dde34c2d14b377aa9c52630951beea9f6f6b8c2/tests/certification/pubsub/azure/eventhubs/eventhubs_test.go#L175

I think what you should do is rename appID5 to be something like `iotApp` so the test is clearer, then towards the top in the `subscriber` application near the section where the message is "observed", also add an assertion that the properties for event hubs exist and have the expected value. But do this is only if the appID is `iotApp`.

https://github.com/dapr/components-contrib/blob/4dde34c2d14b377aa9c52630951beea9f6f6b8c2/tests/certification/pubsub/azure/eventhubs/eventhubs_test.go#L109

hope that makes sense

mecoding1:
Thanks @berndverst for the clarification. 

Currently I think we are only checking the properties over at [eventhubs_integration_test.go](https://github.com/dapr/components-contrib/blob/4dde34c2d14b377aa9c52630951beea9f6f6b8c2/pubsub/azure/eventhubs/eventhubs_integration_test.go ).

https://github.com/dapr/components-contrib/blob/71d7acb506faa2574781ad5548679e7b45e083ca/pubsub/azure/eventhubs/eventhubs_integration_test.go#L116-L122

While have updated the scripts and also the [eventhubs_integration_test.go](https://github.com/dapr/components-contrib/blob/4dde34c2d14b377aa9c52630951beea9f6f6b8c2/pubsub/azure/eventhubs/eventhubs_integration_test.go ) , might a require a few follow up when it comes to certification tests.

If we have a look here in [certification/pubsub/azure/eventhubs/eventhubs_test.go](https://github.com/dapr/components-contrib/blob/4dde34c2d14b377aa9c52630951beea9f6f6b8c2/tests/certification/pubsub/azure/eventhubs/eventhubs_test.go#L93-L115) , we are observing the Data of _**[TopicEvent](https://github.com/dapr/go-sdk/blob/014b4d6836b3e921670721132c145a252d591395/service/common/type.go#L21-L48)**_.

https://github.com/dapr/components-contrib/blob/4dde34c2d14b377aa9c52630951beea9f6f6b8c2/tests/certification/pubsub/azure/eventhubs/eventhubs_test.go#L99-L103

https://github.com/dapr/components-contrib/blob/4dde34c2d14b377aa9c52630951beea9f6f6b8c2/tests/certification/pubsub/azure/eventhubs/eventhubs_test.go#L109
And type of [AddTopicEventHandler ](https://github.com/dapr/go-sdk/blob/014b4d6836b3e921670721132c145a252d591395/service/common/service.go#L37 ) requires definition of [TopicEventHandler ](https://github.com/dapr/go-sdk/blob/014b4d6836b3e921670721132c145a252d591395/service/common/service.go#L52)  which deals with [TopicEvent](https://github.com/dapr/go-sdk/blob/014b4d6836b3e921670721132c145a252d591395/service/common/type.go#L21-L48), 
which brings me to ask do we have any definition or a way of reading headers through the current implementation of [handler](https://github.com/dapr/go-sdk/blob/014b4d6836b3e921670721132c145a252d591395/service/common/service.go#L52) ?
Since, both application & system properties are passed on through headers.

berndverst:
Yes headers = request metadata. So if you examine the metadata map you should be able to find the desired properties on the message.

FYI I will not be working for the remainder of the week because I'm both out sick, but also it's the thanksgiving holidays here in the US.

mecoding1:
@berndverst I believe we are _not receiving headers or metadata_ as part of neither the context nor the TopicEvent in handler.  

https://github.com/dapr/components-contrib/blob/ead9ffa665dd15474ab04392309789c7754aaf8d/tests/certification/pubsub/azure/eventhubs/eventhubs_test.go#L103-L109

In the main/service/http/topic.go , [we are sharing only the context. ](https://github.com/dapr/go-sdk/blob/014b4d6836b3e921670721132c145a252d591395/service/http/topic.go#L272-L273) , whereas we need to get the headers as well ( i.e. context with the value ).
`retry, err := fn(context.WithValue(
				r.Context(),
				"EventMetadata", r.Header), &te)` instead of `retry, err := fn(r.Context(), &te)`

For example I tried a bit as follow, ( with `context.WithValue(
				r.Context(),
				"EventMetadata", r.Header), &te)`  & I receive the headers as key 'EventMetadata' :
image

image

Whereas if we have only `r.Context()` as in [here](https://github.com/dapr/go-sdk/blob/014b4d6836b3e921670721132c145a252d591395/service/http/topic.go#L272-L273), it forwards / inputs a blank context ( without headers ) as follows :

image
image

Without this forwarding of headers from [dapr/go-sdk](https://github.com/dapr/go-sdk) dont think we can parse the system/application properties ( i.e. the headers ) ?

berndverst:
@mecoding1 can you please use App Protocol `HTTP` for the test, and then have a manual http server and endpoint for the PubSub event? That way you should be able to examine the request headers for these additional properties.

For now you cannot use the convenience method of the Go SDK `AddTopicEventHandler` so you have to manually subscribe to a topic and set up a HTTP server and endpoint to receive the event.

The problem is that this method here never examines the headers of the incoming request and doesn't provide a way to retain those headers or custom properties.
https://github.com/dapr/go-sdk/blob/014b4d6836b3e921670721132c145a252d591395/service/http/topic.go#L196

The Go SDK definitely needs to be updated to make this easier, but that should be a separate effort. https://github.com/dapr/dapr/pull/5516 this PR add the ability to pass custom attributes via gRPC to the SDK, but the SDK needs to actually be updated to expose this.

berndverst (changes requested):
Fix linter issue please.

berndverst (approved):

MattCosturos:
Is it possible to get Application Properties (enriched messages) from an Event hub via a Dapr binding now?

yahanda:
+1, as far as the docs say [here](https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-azure-eventhubs/#subscribing-to-azure-iot-hub-events), it seems that only System Properties can be retrieved. Are there any plans to support retrieving Application Properties (message enrichments) in the near future? Thanks!

berndverst:
Docs might not have been updated. Has anyone tried with Dapr 1.10?

also please don't comment on a PR. Comment on the issue or open a new issue.

berndverst:
@MattCosturos @yahanda the way this is implemented you need to add `"requireAllProperties": "true"` to the subscription metadata - either when using the SDK (or manually exposing the `/dapr/config` JSON response) or when using the declarative (`yaml`) subscription.

As of Dapr 1.10 this PR *should* return all data if you provided this setting.

Issue #2288, opened 2022-11-17
Add support for TTL in state stores for all relational databases


berndverst:
Add support for TTL for all relational databases.

Use the native TTL features where possible, and otherwise simulate in Dapr.

To update:

- [X] CockroachDB: #2618
- [X] MongoDB: #2535
- [x] MySQL: #2641
- [X] PostgreSQL: #2302
- [x] Azure SQL Server: #2682
- [X] SQLite

ItalyPaleAle:
In relational databases, we should be able to emulate TTL for state stores fairly easily. Here's how it will work:

- A new "expire_at" column (name can be changed depending on conventions for that database) is added to the table storing data which contains the time the record expires at.
  - The data type depends on what the database supports. If the database has native datetime types, it will use that. Otherwise, a bigint storing a UNIX timestamp can work.
  - A value <= 0 means the record never expires
  - The default value is 0
  - The column will have an index to speed up purging
- When a record with a TTL is added, the database adds the TTL (in seconds) to the current time and stores that for the "expire_at" column
  - Note that the value for "expire_at" MUST be computed in the database (e.g. the query to insert the value would be something like "CURRENT_TIMESTAMP + ttl", so the database computes the current timestamp). This is to avoid issues with potential clock skews.
- When a record is queried, a new WHERE clause is added to the query to make sure that records are returned only if "expire_at" is <= 0 OR "expire_at" < CURRENT_TIME (the time must use the database's clock, to avoid issues with clock skews). Records that have expired are thus silently ignored.
- A background goroutine will periodically purge expired records by running a query like `DELETE FROM table WHERE expire_at > 0 AND expire_at < CURRENT_TIME`.
   - The interval will be configurable via a metadata option
   - The default value can be fairly high, for example 10 minutes (since expired records are silently ignored anyways)
   - If the interval set in the metadata is <= 0, then there's no background garbage collection. This can be used by developers who are sure that they will not use TTLs.

ItalyPaleAle:
Interestingly, looks like the Oracle state store already implemented TTLs in pretty much the same way described above.

berndverst:
@ItalyPaleAle do you plan to run DB migrations (with ALTER table statements) to add the `ttl` column?

ItalyPaleAle:
Yes. Although because we don't keep a `migrations` table today, it may need to do some stuff to see if the column already exists.

olitomlinson:
[...]

PR #2302, opened 2022-11-22
Add TTL to postgres state store

ItalyPaleAle:
Adds TTL support to Postgres state store. Part of #2288, and it's been implemented just like designed in that issue. It includes updates to certification tests.

It also fixes a number of bugs in the Postgres state store (such as transactions not actually using a single transaction that could be rolled back atomically in case of failure).

Docs issue: https://github.com/dapr/docs/issues/2966


state/azure/cosmosdb/cosmosdb.go
```diff
@@ -481,7 +480,7 @@ func createUpsertItem(contentType string, req state.SetRequest, partitionKey str
 		isBinary = false
 	}
 
-	ttl, err := parseTTL(req.Metadata)
+	ttl, err := stateutils.ParseTTL(req.Metadata)
```
ItalyPaleAle, line 483: Note that this method ("parseTTL") was common among many components, so I've moved it to a shared package

berndverst: Prefer using https://github.com/dapr/components-contrib/blob/master/metadata/utils.go#L58 `TryGetTTL`

berndverst (dismissed):

state/utils/ttl.go
```diff
@@ -0,0 +1,40 @@
...
+package utils
+
+import (
+	"fmt"
+	"math"
+	"strconv"
+)
+
+// Key used for "ttlInSeconds" in metadata.
+const MetadataTTLKey = "ttlInSeconds"
```
berndverst, line 23: package contrib `metadata` `TryGetTTL` already does all of this. If necessary just fix up that method instead please.

berndverst: Prefer using https://github.com/dapr/components-contrib/blob/master/metadata/utils.go#L58 `TryGetTTL`

ItalyPaleAle: This method can't be used because it rejects values < 0. For TTLs in state store set requests, we support -1 to indicate to not set a TTL. This is an explicit contract in our docs: https://docs.dapr.io/developing-applications/building-blocks/state-management/state-store-ttl/#persisting-state-ignoring-an-existing-ttl

For that reason, fixing-up the method in contrib metadata is not an option IMHO.

berndverst: Aren't there other components supporting TTL? If so, let's at least move your code into the metadata package so it can be found again.

state/azure/cosmosdb/cosmosdb.go
ItalyPaleAle: Responded below. Don't think that's possible

state/utils/ttl.go
ItalyPaleAle: > Aren't there other components supporting TTL?

What do you mean?

I put it in `state/utils` because it's specific to state stores.

ItalyPaleAle:
Back in draft state so I can make 2 changes.

1. Create a table for "metadata" (make name configurable) that is used to store the migration version AND the last time the GC ran.
2. Change GC so it checks the metadata table to see if it runs, so multiple Dapr sidecars don't all have the GC running periodically.

ItalyPaleAle:
Updated the PR and it's now ready

daixiang0:
pubsub.kafka certification seems flaky.

state/postgresql/migrations.go
```diff
@@ -0,0 +1,227 @@
...
+// Perform the required migrations
+func (m *migrations) Perform(ctx context.Context) error {
+	// Use an advisory lock (with an arbitrary number) to ensure that no one else is performing migrations at the same time
+	// This is the only way to also ensure we are not running multiple "CREATE TABLE IF NOT EXISTS" at the exact same time
+	// See: https://www.postgresql.org/message-id/CA+TgmoZAdYVtwBfp1FL2sMZbiHCWT4UPrzRLNnX1Nb30Ku3-gg@mail.gmail.com
+	const lockID = 42
+
+	// Long timeout here as this query may block
+	queryCtx, cancel := context.WithTimeout(ctx, time.Minute)
+	_, err := m.Conn.ExecContext(queryCtx, "SELECT pg_advisory_lock($1)", lockID)
```
DeepanshuA, line 45: Just thinking aloud: Should we use `pg_try_advisory_lock` instead of `pg_advisory_lock` and return right away if lock already acquired by other session, and let migration be re-triggered later (need to give this part a thought, how re-trigger).

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -534,20 +533,6 @@ func populatePartitionMetadata(key string, requestMetadata map[string]string) st
 	return key
 }
 
-func parseTTL(requestMetadata map[string]string) (*int, error) {
```
DeepanshuA, line 537: I like this part: having a common method in ttl/utils, instead of repetitive method everywhere.

state/postgresql/migrations.go
ItalyPaleAle: I like that this blocks so we don't have to be concerned with retrying after a delay. Solves all the concurrency issues.

state/azure/cosmosdb/cosmosdb.go
berndverst: we already had a common method in `metadata/utils.go`, but that one just can't be used in this instance

berndverst:
We might want to move the TTL utility method into `metadata/utils.go` instead because bindings might benefit from it too, but for now this is fine.

berndverst (approved):

PR #2321, opened 2022-11-28
Allow non-200 responses to be treated as non-errors

johnewart:
# Description

Adding support for an optional metadata field `treatNon200AsError` that will prevent the binding from returning an error object when the upstream service returns a non-2xx error; this component keeps the existing behavior by default (i.e a 500 or other status code will result in an error being returned), which will unfortunately swallow the HTTP server's response. 

## Issue reference

This is related to an issue opened in the Java SDK (https://github.com/dapr/java-sdk/issues/783) -- while this won't fix the open issue immediately it will allow for clients to decide if they want to receive the response object even if the response is not 2xx. 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation (will open a PR from https://github.com/dapr/docs/compare/v1.9...johnewart:http_binding_update?expand=1)

### DIFF
--- a/bindings/http/http.go
+++ b/bindings/http/http.go
@@ -28,6 +28,7 @@ import (
 	"github.com/mitchellh/mapstructure"
 
 	"github.com/dapr/components-contrib/bindings"
+	"github.com/dapr/components-contrib/internal/utils"
 	"github.com/dapr/kit/logger"
 )
 
@@ -35,10 +36,10 @@ import (
 //
 //revive:disable-next-line
 type HTTPSource struct {
-	metadata httpMetadata
-	client   *http.Client
-
-	logger logger.Logger
+	metadata      httpMetadata
+	client        *http.Client
+	errorIfNot2XX bool
+	logger        logger.Logger
 }
 
 type httpMetadata struct {
@@ -70,6 +71,13 @@ func (h *HTTPSource) Init(metadata bindings.Metadata) error {
 		Transport: netTransport,
 	}
 
+	if val, ok := metadata.Properties["errorIfNot2XX"]; ok {
+		h.errorIfNot2XX = utils.IsTruthy(val)
+	} else {
+		// Default behavior
+		h.errorIfNot2XX = true
+	}
+
 	return nil
 }
 
@@ -91,6 +99,9 @@ func (h *HTTPSource) Operations() []bindings.OperationKind {
 // Invoke performs an HTTP request to the configured HTTP endpoint.
 func (h *HTTPSource) Invoke(ctx context.Context, req *bindings.InvokeRequest) (*bindings.InvokeResponse, error) {
 	u := h.metadata.URL
+
+	errorIfNot2XX := h.errorIfNot2XX // Default to the component config (default is true)
+
 	if req.Metadata != nil {
 		if path, ok := req.Metadata["path"]; ok {
 			// Simplicity and no "../../.." type exploits.
@@ -99,6 +110,13 @@ func (h *HTTPSource) Invoke(ctx context.Context, req *bindings.InvokeRequest) (*
 				return nil, fmt.Errorf("invalid path: %s", path)
 			}
 		}
+
+		if _, ok := req.Metadata["errorIfNot2XX"]; ok {
+			errorIfNot2XX = utils.IsTruthy(req.Metadata["errorIfNot2XX"])
+		}
+	} else {
+		// Prevent things below from failing if req.Metadata is nil.
+		req.Metadata = make(map[string]string)
 	}
 
 	var body io.Reader
@@ -164,8 +182,8 @@ func (h *HTTPSource) Invoke(ctx context.Context, req *bindings.InvokeRequest) (*
 		metadata[key] = strings.Join(values, ", ")
 	}
 
-	// Create an error for non-200 status codes.
-	if resp.StatusCode/100 != 2 {
+	// Create an error for non-200 status codes unless suppressed.
+	if errorIfNot2XX && resp.StatusCode/100 != 2 {
 		err = fmt.Errorf("received status code %d", resp.StatusCode)
 	}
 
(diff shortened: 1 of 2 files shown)


bindings/http/http.go
```diff
@@ -164,9 +174,11 @@ func (h *HTTPSource) Invoke(ctx context.Context, req *bindings.InvokeRequest) (*
 		metadata[key] = strings.Join(values, ", ")
 	}
 
-	// Create an error for non-200 status codes.
-	if resp.StatusCode/100 != 2 {
-		err = fmt.Errorf("received status code %d", resp.StatusCode)
+	// Create an error for non-200 status codes unless suppressed.
+	if treatNon200AsError {
```
ItalyPaleAle, line 178: Can you merge these ifs? `if treatNon200AsError && resp.StatusCode…..`

bindings/http/http.go
```diff
@@ -99,6 +102,13 @@ func (h *HTTPSource) Invoke(ctx context.Context, req *bindings.InvokeRequest) (*
 				return nil, fmt.Errorf("invalid path: %s", path)
 			}
 		}
+
+		if treatNon200AsErrorStr, ok := req.Metadata["treatNon200AsError"]; ok {
```
ItalyPaleAle, line 106: I am not a big fan of this property name, as it’s long and it’s technically not just 200, but 2xx status codes.
Can you suggest another name?

artursouza: What about this being a metadata of the component and not of the request? Or, even better, allow both. I predict most people want this behavior to be consistent and not per request.

artursouza (dismissed):
Also, isn't this about the error payload being returned? Why do we need to change the status code behavior?

bindings/http/http.go
berndverst: ErrorIfNot2XX

bindings/http/http.go
```diff
@@ -99,6 +102,13 @@ func (h *HTTPSource) Invoke(ctx context.Context, req *bindings.InvokeRequest) (*
...
+			treatNon200AsError = treatNon200AsErrorStr == "true"
```
berndverst, line 107: please use `IsTruthy` from `package metadata` for this

berndverst (changes requested):
Let's be clear, this should be an optional feature. And I agree with @artursouza it should really be a setting at the component metadata level. It doesn't make sense at the request level in my opinion as it so fundamentally changes the behavior of the component.

I'm not a fan of this change but understand some might want this. It makes troubleshooting Dapr more difficult in my opinion as it is then unclear whether the binding invocation had an error, or whether an external resource correctly returned this status. Even if the HTTP requests returns 500, the binding invocation itself was successful. The request metadata indicates the actual status code.

Also, shouldn't this PR set `treatNon200AsError` (better name this `FailIfStatusNot2xx`) to `false` by default for backwards compatibility?

Right now this PR always enable this new behavior it seems. That is bad.

johnewart:
Currently, the behavior is such that when the response code from the downstream request is anything outside the range 200-299 the binding will return an error (and therefore no response data will be passed along), which means that you cannot act on anything outside of that range as a result. 

This has the side-effect of silently making a decision how to handle the result - the binding itself did not encounter an error (i.e network connectivity issue, etc.) Arguably, the right thing to do is pass along the response (regardless of the downstream status code) unless an actual error happened between the binding and the HTTP endpoint. However, I don't think we can default to that because it's possible that code paths depend on the existing behavior and changing the default would break that.

bindings/http/http.go
johnewart: I would advocate both, changing just the component setting could relate to wider reaching changes. I can change the name to `errorIfNot2XX` both for the request and component metadata.

bindings/http/http.go
johnewart: Changed to use `IsTruthy` in places where string truthiness is evaluated

johnewart:
> Also, isn't this about the error payload being returned? Why do we need to change the status code behavior?

We can't return the payload if the status code is outside 200-299 as the code exists today, gRPC drops the response object if the error is non-nil (I double checked using the gRPC path example and forcibly returned errors, the response object was `nil` in those cases), and also in our HTTP API we do not pass along the response if there is an error returned from the Invoke. 

The current code always passes along the response body, and status code as long as there is no error returned: 

```go
b, err := io.ReadAll(resp.Body)

// skip some bits ...

return &bindings.InvokeResponse{
    Data:     b,
    Metadata: metadata,
}, err
```

bindings/http/http.go
johnewart: Done.

berndverst (approved):

berndverst:
@johnewart IMO this binding should have sent the error as data.

In JSON I would expected something like:

```json
{
error:"some internal server error",
status:"500",
body: ""
}
```

But for backwards compatibility we cannot mange this change now I think.

bindings/http/http_test.go
```diff
@@ -172,6 +172,69 @@ func TestInit(t *testing.T) {
 			path:      "/",
 			err:       "received status code 500",
 		},
+		"internal server error suppressed": {
```
ItalyPaleAle, line 175: Not going to block the PR on this, but it would be great if you could add a test case for 2xx status codes too, such as 200 and 201 (for example).

ItalyPaleAle (approved):

ItalyPaleAle:
> But for backwards compatibility we cannot mange this change now I think.

Maybe this is the time we make a "v2" component? :)

bindings/http/http_test.go
johnewart: Sure, I will add some and push with the DCO issue solved -- can't hurt :D

johnewart:
Squashed the previous changes into one commit (with signoff) and also added a second commit re: @ItalyPaleAle's request for non-200 responses (and to also improve isolation and clean it up a bit)

ItalyPaleAle (dismissed):

ItalyPaleAle (dismissed):

ItalyPaleAle (approved):

berndverst (approved):

berndverst:
@johnewart, as the PR author can you update the docs with this change? It's not in 1.10 and 1.11 docs.

PR #2322, opened 2022-11-28
Hotfix: Cosmosdb StateTransactionRequest with etag crash

RyanLettieri:
Signed-off-by: Ryan Lettieri <ryanLettieri@microsoft.com>

# Description

Changed an asterisk to an ampersand to fix a nil pointer being dereferenced.

## Issue reference

When attempting to create a state transaction request in cosmosDB with an etag, a nil pointer was being dereferenced.

This PR will close: #[_[2319]_](https://github.com/dapr/components-contrib/issues/2319)

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #2323, opened 2022-11-28
Cherry-pick from release into master: Cosmos DB Etag hotfix

berndverst:
# Description

This is a cherry-pick from release into master of the Cosmos DB Etag hotfix. #2322


ItalyPaleAle (approved):

artursouza (changes requested):
Can this be a merge instead?

berndverst:
@artursouza no because there are commits in release 1.9 that do not belong in master.

artursouza:
What are those? I don't recall us getting into that situation before.

Issue #2309, opened 2022-11-24
Mqtt PubSub multiplies messages when application unhealthy detected


bwojdyla:
## Expected Behavior

If dapr sidecar detects application is unhealthy it should unsubscribe from message broker without multiplying messages on queue.

## Actual Behavior

If dapr sidecar detects application is unhealthy it multiplies messages on consumer's queue.

## Steps to Reproduce the Problem
Enable app healthchecks https://docs.dapr.io/developing-applications/building-blocks/observability/app-health/
Configure 3 subscribtions for mqtt component (generally configure more than one subscribtion, The more subscripption the more copies of messages there would be)

```
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: testapp-pubsub
spec:
  type: pubsub.mqtt
  version: v1
  metadata:
  - name: url
    value: "abc"
  - name: qos
    value: 2
  - name: retain
    value: "false"
  - name: cleanSession
    value: "false"
  - name: backOffMaxRetries
    value: "5"
```

Subscription 1
```
apiVersion: dapr.io/v1alpha1
kind: Subscription
metadata:
  name: subscription1 
spec:
  pubsubname: testapp-pubsub
  topic: A
  route: /api/route1
```

Subscription 2
```
apiVersion: dapr.io/v1alpha1
kind: Subscription
metadata:
  name: subscription2
spec:
  pubsubname: testapp-pubsub
  topic: B
  route: /api/route2
```

Subscription 3
```
apiVersion: dapr.io/v1alpha1
kind: Subscription
metadata:
  name: subscription3
spec:
  pubsubname: testapp-pubsub
  topic: C
  route: /api/route2
```

In application prepare message handler so that it last some time ~5s. Later, while handler is running application must be killed.

1. Start dapr and application

```
time="2022-11-24T08:22:23.562665598Z" level=info msg="app is subscribed to the following topics: [C A B] through pubsub=testapp-pubsub" app_id=testapp instance=dc02d2b2cc3c scope=dapr.runtime type=log ver=1.9.4
time="2022-11-24T08:22:23.562701843Z" level=debug msg="subscribing to topic='A' on pubsub='testapp-pubsub'" app_id=testapp instance=dc02d2b2cc3c scope=dapr.runtime type=log ver=1.9.4
[...]

PR #2333, opened 2022-11-30
re-connecting to broker in case of mqtt handler failure

shivamkm07:
Signed-off-by: shivam <shivamkm07@gmail.com>

# Description

MQTT pub-sub currently re-publishes the messages once the handler returns error. This causes same message being re-delivered indefinitely even to other applications which are healthy and subscribed to same topic. This PR removes re-publishing logic and replaces with re-connecting logic where dapr sidecar disconnects and re-connects to broker to trigger broker re-sending unacknowledged messages.

Note that mqtt as per it's spec doesn't re-send unacknowledged messages unless there is re-connection. 

## Issue reference

Please reference the issue this PR will close: #2309 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/pubsub/mqtt/mqtt.go
+++ b/pubsub/mqtt/mqtt.go
@@ -245,6 +245,15 @@ func (m *mqttPubSub) startSubscription(ctx context.Context) error {
 	return nil
 }
 
+func (m *mqttPubSub) reconnect() error {
+	m.subscribingLock.Lock()
+	defer m.subscribingLock.Unlock()
+
+	m.resetSubscription()
+
+	return m.startSubscription(m.ctx)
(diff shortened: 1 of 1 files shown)


shivamkm07:
Closing this as re-connection in case of handler failures would lead to other issues and doesn't look like a good idea. Some of the possible issues would be:
1. Retained messages would be re-sent again and again in case of handler failure
2. Due to re-connection, assuming 100 messages are in-flight and one of the message handlers failed, all of the messages would be re-sent due to re-connection
3. Might not be an efficient solution for brokers which support re-sending unacked messages after a time.

PR #2337, opened 2022-12-02
Upgrade a bunch of SDKs

berndverst:

# Description

Upgrade a bunch of SDKs


internal/component/azure/servicebus/errors.go
```diff
@@ -36,7 +36,8 @@ func IsNetworkError(err error) bool {
 	}
 
 	// Context deadline exceeded errors often happen when the connection is just "hanging"
-	if errors.Is(err, amqp.ErrConnClosed) || errors.Is(err, context.DeadlineExceeded) {
+	var connErr *amqp.ConnError
```
ItalyPaleAle, line 39: This changes the behavior. Before we were only looking for `amqp.ErrConnClosed`, now this will return true for all connection errors

ItalyPaleAle: Looks like this doesn't change the behavior after reading the code. All good

ItalyPaleAle (dismissed):

internal/component/azure/servicebus/errors.go
berndverst: ErrConnClosed doesn't exist anymore and now `amqp.ConnError` is thrown for those situations where connections fail or are closed. There is a more specific inner error but no constants for those are exposed. We don't really need those inner errors however.

berndverst: Reference https://github.com/Azure/go-amqp/commit/1d0e36752c2b095fd8ca31826e3d665b14eef392

ItalyPaleAle (dismissed):

ItalyPaleAle (approved):

berndverst:
cert tests failing were already failing in master

PR #2342, opened 2022-12-06
Merge release-1.9 into master

ItalyPaleAle:
Merges the latest fixes from release-1.9 into master


mukundansundar:
@ItalyPaleAle do we need the with 0 files changed and only commits are there? cc @dapr/maintainers-components-contrib

ItalyPaleAle:
> do we need the with 0 files changed and only commits are there? cc 

Ugh, just realized that. I was wondering why another PR was not closed automatically and had assumed it was due to this. Never mind then :)

PR #2349, opened 2022-12-07
Fix ratelimit middleware not working

### LINKED ISSUE #2343: "ratelimit" middleware strips URL path
## In what area(s)?

/area runtime

## What version of Dapr?

CLI version: 1.9.1
Runtime version: 1.9.4

## Expected Behavior

Setting rate-limiting middleware should not remove the path of the request URI when forwarding the request to the destination app. IOW the path and query parameters provided by the caller should be received by the callee app.

## Actual Behavior

It seems that adding rate-limiting middleware strips the path from the request URL, resulting in the destination app only receiving `/` as path.

Create the following configuration:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: default-rate-limit
spec:
  type: middleware.http.ratelimit
  version: v1
  metadata:
    - name: maxRequestsPerSecond
      value: 1
```

Then I have a very simple app that has an endpoint `/info` and if requests are sent to any other path, it returns the path it received.

This app is configured as follows:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Configuration
metadata:
  name: echo-app
spec:
  appHttpPipeline:
    handlers:
    - name: default-rate-limit
      type: middleware.http.ratelimit
```

When I send a request to the app (`http://localhost:3500/v1.0/invoke/echo-app/method/info`) the URL the app receives is `/` instead of `/info`.

Then I removed the `handlers` section above in `Configuration` and sent the same request again: Now the app receives path `/info` correctly.

### Additional information

I also tried to configure a different middleware ([`uppercase`](https://docs.dapr.io/reference/components-reference/supported-middleware/middleware-uppercase/) instead of rate-limiting): In this case I did **not** observe the described issue.

For reference, here the `Component` used for the `uppercase` middleware test case:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: default-uppercase
spec:
  type: middleware.http.uppercase
  version: v1
  metadata:
    - name: foo # needed as otherwise K8S complains about 'spec.metadata' being a required section.
      value: bar
```

And here the app configuration (just replaced `handlers` section compared to above:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Configuration
metadata:
  name: echo-app
spec:
  appHttpPipeline:
    handlers:
    - name: default-uppercase
      type: middleware.http.uppercase
```

And here the test:

```text
curl -X POST http://localhost:3500/v1.0/invoke/echo-app/method/info -H "Content-Type: text/plain" -d "this is the body"

{
  "path": "/info",
  "body": "THIS IS THE BODY"
}
```

The app echoes back a small JSON with the received path as well as the request body: As you can see the returned request body is all uppercase (to proof that the `uppercase` middleware was indeed invoked).

ItalyPaleAle:
We need to set a value for the X-Forwarded-For header or the upstream library won't rate-limit requests. Additionally, upgraded the tollbooth library to v7.

Fixes #2343

### DIFF
--- a/middleware/http/ratelimit/ratelimit_middleware.go
+++ b/middleware/http/ratelimit/ratelimit_middleware.go
@@ -18,7 +18,8 @@ import (
 	"net/http"
 	"strconv"
 
-	"github.com/didip/tollbooth"
+	tollbooth "github.com/didip/tollbooth/v7"
+	libstring "github.com/didip/tollbooth/v7/libstring"
 
 	"github.com/dapr/components-contrib/middleware"
 	"github.com/dapr/kit/logger"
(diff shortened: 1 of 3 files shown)


DeepanshuA:
LGTM overall, just thought to ask: Should we incorporate a UT for this?

middleware/http/ratelimit/ratelimit_middleware.go
```diff
@@ -54,7 +54,26 @@ func (m *Middleware) GetHandler(metadata middleware.Metadata) (func(next http.Ha
 	limiter := tollbooth.NewLimiter(meta.MaxRequestsPerSecond, nil)
 
 	return func(next http.Handler) http.Handler {
-		return tollbooth.LimitHandler(limiter, next)
+		// Adapted from toolbooth.LimitHandler
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// We need to set a value for X-Forwarded-For or the rate limiter won't work
+			r.Header.Set("X-Forwarded-For", "0.0.0.0")
```
yaron2, line 60: Can this override a user defined X-Forwarded-For?

ItalyPaleAle: In theory, yes. In practice, the requests that hit the Dapr sidecar only come from client apps and there shouldn't be a X-Forwarded-For - if I'm not mistaken

ItalyPaleAle: Just to be safe, I changed the code to try and look up the IP (like the tollbooth library does), and only set "X-Forwarded-For" if strictly necessary.

yaron2 (approved):

middleware/http/ratelimit/ratelimit_middleware.go
yaron2: Thanks, this change makes sense.

yaron2:
@ItalyPaleAle please fix DCO and we can merge.

yaron2 (approved):

PR #2350, opened 2022-12-07
MQTT: Resume subscriptions after a reconnection

### LINKED ISSUE #2229: MQTT Missing Resubscribe on Reconnect
## Expected Behavior

When I restart my mqtt-broker, the dapr sidecar should reconnect, resubscribe and the receive messsages again

## Actual Behavior

The sidecar does reconnect but not resubscribe

* subscribe with dapr to mqtt broker
* restart mqtt broker

## Release Note

RELEASE NOTE:

## Possible Solution

As proposed in the upstream issue (https://github.com/eclipse/paho.mqtt.golang/issues/22), we could put the subscribe call in the `onConnect` handler this should fix the issue.

ItalyPaleAle:
# Description

Fixes #2229 by making MQTT re-subscribe to topics after the connection to the MQTT broker is interrupte

## Issue reference

Fixes #2229

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


ItalyPaleAle:
Moving back to draft because there are still some issues causing tests to fail in the CI (which I am not able to reproduce locally :( )

ItalyPaleAle:
@thielepaul thanks for the reminder. I updated the binding too and fixed the race condition in the tests. This is ready for review now.

bindings/mqtt/mqtt.go
```diff
@@ -243,58 +243,96 @@ func (m *MQTT) handleMessage(ctx context.Context, handler bindings.Handler, mqtt
...
 
 func (m *MQTT) Read(ctx context.Context, handler bindings.Handler) error {
-	// reset synchronization
-	if m.consumer != nil {
-		m.logger.Warnf("re-initializing the subscriber")
-		m.consumer.Disconnect(5)
-		m.consumer = nil
+	// mqtt broker allows only one connection at a given time from a clientID
+	topicHash := sha256.Sum256([]byte(m.metadata.topic))
+	consumerClientID := fmt.Sprintf("%s-consumer-%x", m.metadata.clientID, topicHash[0:4])
```
artursouza, line 248: This is a big change in how consumerID are calculated. Why is this needed?

artursouza (changes requested):

bindings/mqtt/mqtt.go
ItalyPaleAle: There can only be 1 active consumer with the same name. If you subscribe to multiple topics with the same binding, then it would fail.

Note that this change impact the binding only, which is not a stable component. If that's a concern for a hotfix, I can move that into a separate PR and have the pubsub only in a hotfix

pubsub/mqtt/mqtt.go
```diff
@@ -324,20 +313,61 @@ func (m *mqttPubSub) handlerForTopic(topic string) pubsub.Handler {
...
+			tokenErr := token.Error()
+			// In case of timeout, ok will be false and there won't be an error in the token
+			if tokenErr == nil && !ok {
+				tokenErr = context.DeadlineExceeded
+			}
+			if tokenErr != nil {
+				// Cause a disconnection and wait for reconnection
+				m.logger.Errorf("MQTT error while trying to subscribe to topics; will reconnect. Error: %v", tokenErr)
+				client.Disconnect(5)
+				m.connect(ctx, clientID, isConsumer, readyCh)
```
shivamkm07, line 356: This would essentially lead to recursive calls to connect, and indefinite retries in case of subscribe failures. Is this expected behavior?

ItalyPaleAle: I would say so, until it recovers. There’s a timeout in starting subscriptions anyways.

pubsub/mqtt/mqtt.go
```diff
@@ -88,6 +88,7 @@ func (m *mqttPubSub) Init(metadata pubsub.Metadata) error {
 	}
 
 	m.ctx, m.cancel = context.WithCancel(context.Background())
+	m.topics = make(map[string]mqttPubSubSubscription)
```
artursouza, line 91: If this is the same change in reelase branch, it should be removed from here to avoid conflict. We merge release branch into master.

ItalyPaleAle: It's the same. I will change that once it's merged in master

pubsub/mqtt/mqtt.go
```diff
@@ -213,34 +213,23 @@ func (m *mqttPubSub) startSubscription(ctx context.Context) error {
...
 	select {
-	case <-token.Done():
-		// Subscription went through
-	case <-subscribeCtx.Done():
-		return subscribeCtx.Err()
-	}
-	if err := token.Error(); err != nil {
-		return fmt.Errorf("mqtt error from subscribe: %v", err)
+	// Double the wait time here as we've seen that some brokers just need a bit more time
+	case <-time.After(defaultWait * 2):
```
DeepanshuA, line 226: Is it correct to double the defaultTime here OR increase/double the default wait time in metadata.go itself?

pubsub/mqtt/mqtt.go
DeepanshuA: Should we have some control in terms of number of retries? Let's say, if it doesn't timeout but fails due to some non-retriable reason, it will just keep on re-trying. If we can't differentiate between retriable and non-retriable errors here, may be having limited number of attempts may help.

bindings/mqtt/mqtt.go
```diff
@@ -243,58 +243,96 @@ func (m *MQTT) handleMessage(ctx context.Context, handler bindings.Handler, mqtt
...
-	c, err := m.connect(consumerClientID)
+	m.logger.Infof("mqtt subscribing to topic '%s' with consumer client ID '%s'", m.metadata.topic, consumerClientID)
+	client, err := m.connect(consumerClientID, onConn)
 	if err != nil {
 		return err
 	}
-	m.consumer = c
+	// m.connect returns after the connection has been established, now we need to wait for the subscription to be up
+	select {
+	case <-time.After(defaultWait):
```
DeepanshuA, line 270: if I understand correctly, I would recommend that it would be better to use time.NewTimer instead of time.After, to avoid memory leak scenarios. (https://pkg.go.dev/time#After)

ItalyPaleAle: I will change it, but the risk of leaks is not a problem here. Per the docs:

> The underlying Timer is not recovered by the garbage collector until the timer fires.

Which does indeed mean there's a "leak" but only "until the timer fires". In this case, the timer is short-lived (`defaultWait` is 30s), so it's not a big concern here.

pubsub/mqtt/mqtt.go
ItalyPaleAle: Let me look at how it can be improved

berndverst:
@ItalyPaleAle is this ready to be merged?

pubsub/mqtt/mqtt.go
berndverst: The change was merged in the release branch first in this case.

pubsub/mqtt/mqtt.go
berndverst: @ItalyPaleAle anything remaining here?

berndverst:
Ping.

Does this need to go into 1.10? If so can we get this finalized?

ItalyPaleAle:
I think I actually want to rewrite this. I'm not very happy with how this looks. Will find the time for this.

ItalyPaleAle:
Moving back to draft because I want to rewrite this

ItalyPaleAle:
2nd attempt in #2467

PR #2352, opened 2022-12-08
Check the correct error object in a few places inside blobstorage client

johnewart:
# Description

Fix a few places where we get a new error object (i.e `newSharedKeyErr`) but then check if `err` is non-nil, when we should check the error we just got. 

## Checklist

* [x] Code compiles correctly
* [NA] Created/updated tests
* [NA] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_


ItalyPaleAle (approved):

DeepanshuA (approved):

berndverst:
Not sure how you found this, but thanks for catching! An artifact of addressing linter stuff due to "shadowed" variables etc

berndverst (approved):

PR #2355, opened 2022-12-08
[Hotfix release-1.9] MQTT PubSub: Resume subscriptions after a reconnection

ItalyPaleAle:
Extracted from #2350 with the fix for the MQTT PubSub component only

The fix for the MQTT binding component will not be included in the hotfix

### DIFF
--- a/pubsub/mqtt/mqtt.go
+++ b/pubsub/mqtt/mqtt.go
@@ -89,14 +89,9 @@ func (m *mqttPubSub) Init(metadata pubsub.Metadata) error {
 	m.ctx, m.cancel = context.WithCancel(context.Background())
 	m.topics = make(map[string]mqttPubSubSubscription)
 
-	// mqtt broker allows only one connection at a given time from a clientID.
-	producerClientID := m.metadata.producerID
-	if producerClientID == "" {
-		// for backwards-compatibility; see: https://github.com/dapr/components-contrib/pull/2104
-		producerClientID = m.metadata.consumerID + "-producer"
-	}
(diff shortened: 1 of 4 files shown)


artursouza (approved):

pubsub/mqtt/mqtt.go
```diff
@@ -212,34 +212,23 @@ func (m *mqttPubSub) startSubscription(ctx context.Context) error {
 		// for backwards-compatibility; see: https://github.com/dapr/components-contrib/pull/2104
 		consumerClientID += "-consumer"
 	}
-	connCtx, connCancel := context.WithTimeout(ctx, defaultWait)
-	c, err := m.connect(connCtx, consumerClientID)
-	connCancel()
+
+	readyCh := make(chan struct{})
```
mukundansundar, line 216: should this be buffered ?
```suggestion
	readyCh := make(chan struct{}, 1)
```

mukundansundar: @ItalyPaleAle ^

ItalyPaleAle: No, it's ok for this to be unbuffered so it blocks until it's consumed

ItalyPaleAle: Changed to use `close()` rather than sending a message, which won't block

ItalyPaleAle:
I made a small change following feedback by @mukundansundar

ItalyPaleAle:
For some cert tests that are failing, it's due to external changes in the test infra. I've opened #2356 to backport those into release-1.9

PR #2363, opened 2022-12-13
Cloudflare KV state store and Queues binding - new components

ItalyPaleAle:
# Description

This PR adds two new components:

- `state.cloudflare.workerskv` is a state store backed by [Cloudflare Workers KV](https://developers.cloudflare.com/workers/learning/how-kv-works/)
- `bindings.cloudflare.queues` is a binding that allows *publishing* messages to [Cloudflare Queues](https://developers.cloudflare.com/queues/)
   - This is implemented as a binding and not pubsub because (at least at the moment), subscribers to Queues can only be Workers, so Dapr cannot receive messages from a Queue.

They are implemented in the same PR because they share the entire underlying "infrastructure". Both components, in fact, require a Cloudflare Worker to be deployed to work.

The Worker can be deployed automatically by the components. A new internal package takes care of deploying the Worker (and upgrading it if needed) by invoking the Cloudflare APIs. Additionally, all requests from Dapr to the Worker must be authenticated using a JWT signed by Dapr using EdDSA (using Ed25519).

## Metadata properties

Both components share a number of required metadata properties to work:

- `workerName`: This is the name of the Worker that is deployed. This should be unique per each component: that is, users should not use the same worker to bind to two different KV namespaces or two different Queues, or to bind to a Queue and a KV namespace
- `key`: Ed25519 private key (PEM-encoded) that is used by Dapr to issue JWT tokens to authenticate with the Worker.
   - This is generated with `openssl genpkey -algorithm ed25519 -out private.pem` 
   - Note: on MacOS, this is not available with the built-in LibreSSL that is invoked with `openssl`; instead, install OpenSSL via Homebrew and then use `/opt/homebrew/opt/openssl@3/bin/openssl

There are then two options:

- The user can pre-create the Worker and have it deployed manually (I'll write docs for that).
   -   In this case, the user must set `workerUrl` to the URL of the worker, for example `https://daprdemo.mydapr.workers.dev`
- Dapr can create a Worker (and upgrade it if needed) automatically. Required metadata properties:
   - `cfAPIToken`: API token for Cloudflare (created with the "Workers" template)
   - `cfAccountID`: Cloudflare account ID

For the Queues binding, this is needed too:

- `queueName`: Name of the Cloudflare Queue
   - Must be pre-created for example with `npx wrangler queues create <queuename>`

For the KV state store, this is needed too:

- `kvNamespaceID`: ID of the KV namespace
  - Must be pre-created, for example with `npx wrangler kv:namespace create <namespacename>` 
  - The ID is a hex string returned by the command above, not the name

## Checklist

* [X] Code compiles correctly
* [x] Enable Conformance tests
   * For the KV state store only. For Queues, we need an upgraded (paid) Cloudflare account
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: @ItalyPaleAle TODO


yaron2:
Great components to have, thanks for the PR. I do have questions around the expected usability aspect when working with these components from compute (Dapr in this case) that's not deployed in a Cloudflare edge data centers - and if it's recommended or expected for users running compute outside of Cloudflare to use Cloudflare's KV and queue systems in regards to latency.

ItalyPaleAle:
Those are good questions.

The latency should not be a major concern. I haven't done accurate measurements, but Dapr would make requests to the closest CF datacenter, so latency would not be worse than any other cloud state store we support (like hosted SQL databases) - actually, given how many datacenters Cloudflare has, I would expect latency to be among the best of the cloud-hosted services.

With regards to KV, one thing to keep in mind is that its consistency model is closer to a "session-level" consistency (although it's "session-less"). Requests that hit the same Cloudflare datacenter see the data immediately updated. But if a request were to hit a different datacenter, there could be a delay of up to 30-60s before KV data is consistent. This shouldn't be a big issue in the case of Dapr, since the compute is normally running in a single DC (and if people use geo-redundancy... then that's on them to figure it out :) )

As for the use case, I see the KV component more like a step into integrating Dapr with the Workers ecosystem and enabling more interoperability with code that's running on Workers. Outside of users that are also using Workers for other things, I wouldn't otherwise recommend using KV as a general purpose state store however, unless users have some very specific needs.

Different story is for Queues. The Dapr integration in this case is, in my mind, a way to enable scaling Dapr apps so they can leverage Workers. The idea is that you have some work that you want to perform in a Worker, so you use Dapr to add that to a queue, and then a Worker would take it up. I think there's a lot of use for this.

For context, the Queues binding was the first thing I added (and you can see that in the name of the branch). I added the KV state store later because I wanted to create a "common infra" for Workers-based components, and adding KV was the forcing function for me to do that: after I created the internal package with the common infra, adding the KV state store was a very low-hanging fruit. I am envisioning that the "common infra" could then be used for a bunch of other things, including other components (maybe Cloudflare D1?) and even for the "compute" component I've proposed in dapr/proposals#16

ItalyPaleAle:
@yaron2 

Those are good questions.

The latency should not be a major concern. I haven't done accurate measurements, but Dapr would make requests to the closest CF datacenter, so latency would not be worse than any other cloud state store we support (like hosted SQL databases) - actually, given how many datacenters Cloudflare has, I would expect latency to be among the best of the cloud-hosted services.

With regards to KV, one thing to keep in mind is that its consistency model is closer to a "session-level" consistency (although it's "session-less"). Requests that hit the same Cloudflare datacenter see the data immediately updated. But if a request were to hit a different datacenter, there could be a delay of up to 30-60s before KV data is consistent. This shouldn't be a big issue in the case of Dapr, since the compute is normally running in a single DC (and if people use geo-redundancy... then that's on them to figure it out :) )

As for the use case, I see the KV component more like a step into integrating Dapr with the Workers ecosystem and enabling more interoperability with code that's running on Workers. Outside of users that are also using Workers for other things, I wouldn't otherwise recommend using KV as a general purpose state store however, unless users have some very specific needs.

Different story is for Queues. The Dapr integration in this case is, in my mind, a way to enable scaling Dapr apps so they can leverage Workers. The idea is that you have some work that you want to perform in a Worker, so you use Dapr to add that to a queue, and then a Worker would take it up. I think there's a lot of use for this.

For context, the Queues binding was the first thing I added (and you can see that in the name of the branch). I added the KV state store later because I wanted to create a "common infra" for Workers-based components, and adding KV was the forcing function for me to do that: after I created the internal package with the common infra, adding the KV state store was a very low-hanging fruit. I am envisioning that the "common infra" could then be used for a bunch of other things, including other components (maybe Cloudflare D1?) and even for the "compute" component I've proposed in dapr/proposals#16

ItalyPaleAle:
We now have a (free) Cloudflare account, which is enough for the Conf test for the KV state store (but not for Queues).

I have updated the PR to add the test. They're passing when I run them locally, but because the test requires secrets, we won't know if it works here until the PR is merged in "master".

yaron2:
> Different story is for Queues.

Agree, queues are less latency sensitive from the publisher perspective so it makes sense to have Dapr invoke queues that worker functions then consume.

> but Dapr would make requests to the closest CF datacenter, so latency would not be worse than any other cloud state store we support (like hosted SQL databases) - actually, given how many datacenters Cloudflare has, I would expect latency to be among the best of the cloud-hosted services.

Latency within an AWS region is ~3ms, where latency from code (Dapr) deployed in an AWS region to a CF DC is ~50ms, so that's considerably slower writing to CF KV than to DynamoDB within the same region, for example.

In any case I don't see this as a showstopper for the component.

.github/workflows/certification.yml
```diff
@@ -70,15 +70,15 @@ jobs:
         - bindings.postgres
         EOF
         )
-        echo "::set-output name=pr-components::$PR_COMPONENTS"
```
ItalyPaleAle, line 73: Reviewers: The old syntax is being deprecated and GitHub was complaining

ItalyPaleAle:
Update:

> I have updated the PR to add the test. They're passing when I run them locally, but because the test requires secrets, we won't know if it works here until the PR is merged in "master".

In coordination with @berndverst we created a new branch in dapr/components-contrib (so not a fork) and run the conf tests there, where secrets are available. Tests are passing and the cleanup steps work too. All good :)

internal/component/cloudflare/workers/workers.go
```diff
@@ -0,0 +1,395 @@
...
+	"github.com/dapr/kit/ptr"
+)
+
+const (
+	// Minimum version required for the running Worker.
+	minWorkerVersion = 20221209
+	// Issuer for JWTs
+	tokenIssuer = "dapr.io/cloudflare" //nolint:gosec
+	// JWT token expiration
+	tokenExpiration = 5 * time.Minute
```
DeepanshuA, line 43: Do we want these to be configurable? And for default should we place defaults in metadata?

internal/component/cloudflare/workers/workers.go
```diff
@@ -0,0 +1,395 @@
...
+	logger               logger.Logger
+	ctx                  context.Context
+	cancel               context.CancelFunc
+}
+
+// Init the base class.
+func (w *Base) Init(workerBindings []Binding, componentDocsURL string, infoResponseValidate func(*InfoEndpointResponse) error) (err error) {
+	w.ctx, w.cancel = context.WithCancel(context.Background())
+	w.client = &http.Client{
+		Timeout: time.Second * 30,
```
DeepanshuA, line 61: Same

tests/config/state/tests.yml
```diff
@@ -39,6 +39,9 @@ components:
   - component: cassandra
     allOperations: false
     operations: [ "set", "get", "delete", "bulkset", "bulkdelete", "ttl" ]
+  - component: cloudflare.kv
+    allOperations: false
+    operations: [ "set", "get", "delete", "bulkset", "bulkdelete"]
```
DeepanshuA, line 44: As the basis is a key value store primarily, should we support timeToLive since beginning?

state/cloudflare/kv/cfkv.go
```diff
@@ -0,0 +1,217 @@
...
+	return []state.Feature{}
+}
+
+func (q *CFKV) Delete(parentCtx context.Context, stateReq *state.DeleteRequest) error {
+	token, err := q.metadata.CreateToken()
+	if err != nil {
+		return fmt.Errorf("failed to create authorization token: %w", err)
+	}
+
+	ctx, cancel := context.WithTimeout(parentCtx, 30*time.Second)
```
DeepanshuA, line 101: Should we make timeout duration configurable, instead of hard-coding here?

state/cloudflare/kv/cfkv.go
```diff
@@ -0,0 +1,217 @@
...
+	return nil
+}
+
+func (q *CFKV) Get(parentCtx context.Context, stateReq *state.GetRequest) (*state.GetResponse, error) {
+	token, err := q.metadata.CreateToken()
+	if err != nil {
+		return nil, fmt.Errorf("failed to create authorization token: %w", err)
+	}
+
+	ctx, cancel := context.WithTimeout(parentCtx, 30*time.Second)
```
DeepanshuA, line 132: Same as above.

internal/component/cloudflare/workers/workers.go
ItalyPaleAle: I don't think this specific value should be configurable. The component is generating a new JWT on every request, which are considered disposable.

Actually, I may reduce this to 1 min for better security.

internal/component/cloudflare/workers/workers.go
```diff
@@ -0,0 +1,395 @@
...
+	infoResponseValidate func(*InfoEndpointResponse) error
+	componentDocsURL     string
+	client               *http.Client
+	logger               logger.Logger
+	ctx                  context.Context
+	cancel               context.CancelFunc
+}
+
+// Init the base class.
+func (w *Base) Init(workerBindings []Binding, componentDocsURL string, infoResponseValidate func(*InfoEndpointResponse) error) (err error) {
```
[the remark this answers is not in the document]
ItalyPaleAle: It's used to show a log message if users need to manually set up the Worker.

internal/component/cloudflare/workers/workers.go
```diff
@@ -0,0 +1,395 @@
...
+
+type deployWorkerMetadata struct {
+	Main              string    `json:"main_module"`
+	CompatibilityDate string    `json:"compatibility_date"`
+	UsageModel        string    `json:"usage_model"`
+	Bindings          []Binding `json:"bindings,omitempty"`
+}
+
+// Binding contains a binding that is attached to the worker
+type Binding struct {
```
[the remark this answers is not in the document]
ItalyPaleAle: Will do

tests/config/state/tests.yml
ItalyPaleAle: Right, looks like TTLs should be supported: https://developers.cloudflare.com/workers/runtime-apis/kv/#expiring-keys I need to add support for them

internal/component/cloudflare/workers/workers.go
ItalyPaleAle: Done

state/cloudflare/kv/cfkv.go
ItalyPaleAle: This is during the step that's creating the worker during init, and I think hard-coding a timeout here makes sense.

tests/config/state/tests.yml
ItalyPaleAle: I added support for TTLs. However, it seems that KV requires TTLs to be > 60 seconds. For that reason, I am not enabling them on conf tests. It should be tested in cert tests in the future.

berndverst (approved):

berndverst:
@ItalyPaleAle don't forget the docs issue please!

ItalyPaleAle:
@berndverst done: https://github.com/dapr/docs/issues/3026

Issue #2282, opened 2022-11-15
AWS Component Testing Infrastructure


msfussell:
## AWS Infrastructure
AWS Infrastructure needs to be set up for stable certification for components

## Release Note
Using the AWS supplied infrastructure account, this ensuring that AWS components are tested on AWS infra.

berndverst:
This is a prerequisite for the AWS component certification tests (and ideally conformance tests).

A service account equivalent will need to be created and the credentials must be added to the Dapr GitHub organization by the STC.

The same service account will need to be used to run all tests, so for any future tests new permissions must be granted to this service account.

It is important not to permanently provision resources or choosing the smallest tier possible to reduce cost.

Additionally, it may be advisable to keep configurations in a AWS secret store instead - similarly to the way we keep things in Azure KeyVault. It would be better if the AWS tests do not depend on any Azure infrastructure in any way shape or form.

ItalyPaleAle:
> Additionally, it may be advisable to keep configurations in a AWS secret store instead - similarly to the way we keep things in Azure KeyVault. It would be better if the AWS tests do not depend on any Azure infrastructure in any way shape or form.

IMHO I think it may be useful to keep all secrets in the same place. Since they're opaque, it doesn't matter if they're stored in AWS or Azure Key Vault or somewhere else... Having less places where secrets are stored makes management and access control easier.

berndverst:
From a maintainer perspective I would like to envision a future where the Azure, AWS and other infra pieces can be managed by distinct people with distinct sets of permissions.

I still am strongly in favor of *not* mixing AWS and Azure components in the same test Workflows and not sharing any infra between these.

addjuarez:
General AWS conformance/certification workflow

1. Set aws github secrets AWS_SECRET_KEY, AWS_ACCESS_KEY as env variables.
2. Utilize **Terraform** to deploy and manage resources needed for tests for cloud service components. Terraform Templates should be added to the repo when conformance or certifications tests are added for cloud components.
3. After the service is deployed and ready, we run the tests. Component yamls will reference the env variables for the AWS credentials
4. Clean up resources using terraform

There are github actions plugins for terraform so the only code changes needed should be inside conformance/certification workflow yaml and terraform files will be needed for cloud service components. These workflow changes could also be easily updated to support azure and gcp.

https://www.terraform.io/
https://github.com/hashicorp/setup-terraform

ItalyPaleAle:
> Terraform

[...]

PR #2365, opened 2022-12-13
AWS Certification Infrastructure 

addjuarez:
# Description

Added terraform infrastructure for certification tests and conformance tests. Currently this support only AWS but It can be easily configured to support GCP or Azure with some small updates. 

AWS components can use the following env variables for the credentials AWS_ACCESS_KEY_ID and AWS_ACCESS_KEY_SECRET

Example of aws conformance test run https://github.com/addjuarez/components-contrib/actions/runs/3707312106/jobs/6283587598

**Conformance Tests:**

- To make use of the infrastructure with conformance tests you will have to add your component to the CRON_COMPONENT group in the conformance.yaml file. 
- You will also need to add an argument to your component called _terraform-dir_. This Variable should have the directory path from the .github/infrastructure/terraform/conformance folder. 
- You may also need set env variables in the workflow for your resource names so the conformance tests can reference them.
- In previous folder you will need to have a terraform file that supports the resources the conformance tests will use.
- There is an example with aws snssqs pubsub component that shows how to support the following.

**Certification Tests:**

- To make use of the infrastructure with conformance tests you will have to add your component to the CLOUD_COMPONENT group in the certification .yaml file. 
- You will also need to add an argument to your component called _terraform-dir_. This Variable should have the directory path from the .github/infrastructure/terraform/certification folder. 
- In previous folder you will need to have a terraform file that supports the resources the certification tests will use.
- Certification tests have an env variable called $UNIQUE_ID created from the github run id. You can reference this value in your terraform file and tests to create non conflicting resources for your tests. 

```
variable "UNIQUE_ID " {
    type        = string
    description = "This is an example input variable using env variables."
}

resource "aws_sns_topic" "testTopic" {
  name = "certification-pubsub-topic-${var.UNIQUE_ID }"
  tags = {
    dapr-topic-name ="certification-pubsub-topic-${var.UNIQUE_ID }"

  }
}

```
`var topicActiveName = "certification-pubsub-topic-" + os.Getenv("UNIQUE_ID ")
`
## Issue reference

#2282 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


.github/workflows/conformance.yml
```diff
@@ -176,6 +177,8 @@ jobs:
       run:
         shell: bash
     needs: generate-matrix
+    env:
+      UUID: ${{github.run_id}}
```
ItalyPaleAle, line 181: github.run_id is not a UUID however, is it? Can you use a different name so it's not misleading?

tmacam:
Not to be a naming purist but why `CRON_COMPONENT` membership? Seems a bit odd that to run things in a different cloud you need permissions to run things in the Cron component.

tmacam (changes requested):

.github/infrastructure/terraform/conformance/pubsub/aws/snssqs/snssqs.tf
```diff
@@ -0,0 +1,94 @@
...
+      "Action": "sqs:SendMessage",
+      "Resource": "${aws_sqs_queue.testQueue.arn}",
+      "Condition": {
+        "ArnEquals": {
+          "aws:SourceArn": "${aws_sns_topic.multiTopic1.arn}"
+        }
+      }
+    },
+    {
+      "Sid": "First",
```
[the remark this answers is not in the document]
tmacam: From https://docs.aws.amazon.com/sns/latest/dg/sns-access-policy-use-cases.html#sns-publish-messages-to-sqs-queue and https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_multi-value-conditions.html, I think we can tighten and simplify this SQS policy to a single statement such as:

```json
{
  "Statement": [{
    "Sid": "Allow-SNS-SendMessage",
    "Effect": "Allow",
    "Principal": {
      "Service": "sns.amazonaws.com"
    },
    "Action": "sqs:SendMessage",
    "Resource": "${aws_sqs_queue.testQueue.arn}",
    "Condition": {
      "ArnEquals": {
        "aws:SourceArn": [
          "${aws_sns_topic.testTopic.arn}",
          "${aws_sns_topic.multiTopic1.arn}",
          "${aws_sns_topic.multiTopic2.arn}"
        ]
      }
    }
  }]
}
```


This will only grant SNS, _the service_, to publish to that SQS queue IFF the source SNS topic is one of three defined in that list. (Might IAM might be a bit rusty so... better double check that).

.github/infrastructure/terraform/conformance/pubsub/aws/snssqs/snssqs.tf
```diff
@@ -0,0 +1,94 @@
+resource "aws_sns_topic" "testTopic" {
+  name = "testTopic"
```
tmacam, line 2: do we need enforce uniqueness on these identifiers? What would happen if two PRs try to run certification/conformance tests at the same time? Would those collide? If so, should we suffix those with `${github.run_id}`

ItalyPaleAle:
> Not to be a naming purist but why CRON_COMPONENT membership? Seems a bit odd that to run things in a different cloud you need permissions to run things in the Cron component.

I found that today too and my understanding is that it's called "cron" because it's the list of components that are tested by the cron task (every X hours) - and also by "/ok-to-test" :)

.github/infrastructure/terraform/conformance/pubsub/aws/snssqs/snssqs.tf
```diff
@@ -0,0 +1,94 @@
...
+}
+
+resource "aws_sns_topic" "multiTopic2" {
+  name = "multiTopic2"
+  tags = {
+    dapr-topic-name = "multiTopic2"
+  }
+}
+
+resource "aws_sqs_queue" "testQueue" {
```
DeepanshuA, line 22: Just cross-checking: if we also need to configure a dlq for any of the test

.github/workflows/conformance.yml
```diff
@@ -254,6 +257,42 @@ jobs:
...
+    - name: Setup Terraform
+      uses: hashicorp/setup-terraform@v2
+      if: matrix.terraform-dir != ''
+
+    - name: Configure AWS Credentials
+      uses: aws-actions/configure-aws-credentials@v1
+      with:
+        aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY }}
+        aws-secret-access-key: ${{ secrets.AWS_SECRET_KEY }}
+        aws-region: us-west-2
```
robertojrojas, line 269: Suggestion - If the AWS Region is required, it should be provided dynamically instead of hard-coded.

.github/infrastructure/terraform/conformance/pubsub/aws/snssqs/snssqs.tf
```diff
@@ -0,0 +1,94 @@
+resource "aws_sns_topic" "testTopic" {
```
robertojrojas, line 1: Suggestion - It's customary in Terraform to add the provider block similar to the following.
This could be added to the top of this file or as a separate file usually called **versions.tf**
```terraform
terraform {
  required_version = ">=0.13"

  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "~> 4.0"
    }
  }
}

provider "aws" {
}
```

.github/infrastructure/terraform/conformance/pubsub/aws/snssqs/snssqs.tf
```diff
@@ -0,0 +1,94 @@
...
+  name = "testTopic"
+  tags = {
+    dapr-topic-name = "testTopic"
```
robertojrojas, line 4: Suggestion - Specifying the values for **name** and **dapr-topic-name** as a Terraform variable with default values would allow modification at runtime.
Something like:
```terraform
#variables.tf
variable "test_topic" {
  type            = string
  description = "The Pub/Sub topic name."
  default        = "testTopic"
}

#snssqs.tf
 name = var.test_topic
 tags = {
   dapr-topic-name = var.test_topic
}


```

.github/workflows/certification.yml
```diff
@@ -223,6 +237,42 @@ jobs:
...
+    - name: Setup Terraform
+      uses: hashicorp/setup-terraform@v2
+      if: matrix.terraform-dir != ''
+
+    - name: Configure AWS Credentials
+      uses: aws-actions/configure-aws-credentials@v1
+      with:
+        aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY }}
+        aws-secret-access-key: ${{ secrets.AWS_SECRET_KEY }}
+        aws-region: us-west-2
```
robertojrojas, line 249: Suggestion -  If required, the AWS Region should be provided by as a variable.

.github/infrastructure/terraform/conformance/pubsub/aws/snssqs/snssqs.tf
tmacam: YES! Good catch!

.github/workflows/conformance.yml
addjuarez: Sure, I can rename it to simply RUN_ID.

.github/infrastructure/terraform/conformance/pubsub/aws/snssqs/snssqs.tf
```diff
@@ -0,0 +1,94 @@
...
+      "Condition": {
+        "ArnEquals": {
+          "aws:SourceArn": "${aws_sns_topic.multiTopic1.arn}"
+        }
+      }
+    },
+    {
+      "Sid": "First",
+      "Effect": "Allow",
+      "Principal": "*",
```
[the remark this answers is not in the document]
addjuarez: > Not to be a naming purist but why `CRON_COMPONENT` membership? Seems a bit odd that to run things in a different cloud you need permissions to run things in the Cron component.

CRON_COMPONENT here means the tests run on a schedule instead of on every pr. For conformance tests that will use terraform to provision resources this is currently necessary. The conformance framework doesn't support dynamic component names so we cant avoid resource conflicts if multiple of these jobs run. Certification tests this isn't an issue.

.github/workflows/certification.yml
```diff
@@ -336,6 +389,12 @@ jobs:
         name: ${{ matrix.component }}_certification_test
         path: ${{ env.TEST_OUTPUT_FILE_PREFIX }}_certification.*
 
+    - name: Terraform Destroy
+      continue-on-error: true    
+      run: terraform destroy -auto-approve -var="UUID=${{github.run_id}}"
+      working-directory: "./.github/infrastructure/terraform/certification/${{ matrix.terraform-dir }}"
+      if: matrix.terraform-dir != ''
```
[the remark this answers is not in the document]
addjuarez: Yes if terraform fails to destroy the resources they will remain there. We would probably need to make a script to delete the orphan resources.

.github/infrastructure/terraform/conformance/pubsub/aws/snssqs/snssqs.tf
addjuarez: The conformance tests do not require a dlq. The terraform file for the certification tests should have one however.

.github/infrastructure/terraform/conformance/pubsub/aws/snssqs/snssqs.tf
addjuarez: Sure I can add this for clarity.

.github/workflows/conformance.yml
addjuarez: A default region is required. It can be overwritten in the terraform file.

.github/infrastructure/terraform/conformance/pubsub/aws/snssqs/snssqs.tf
```diff
@@ -0,0 +1,94 @@
...
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~> 4.0"
+    }
+  }
+}
+
+provider "aws" {
+  region = "us-east-1"
+}
```
tmacam, line 14: Just to comment something we mentioned on the call: it would be cool to tag resources created by this with a creation timestamp and perhaps something else to identify them as being created for automated-testing. This will help us identify them and "garbage collect" them if the "Terraform Destroy" step fails for some reason.

According to https://engineering.deptagency.com/best-practices-for-terraform-aws-tags you can add default tags in this section and all resources created will have those tags.

yaron2:
@addjuarez please resolve the conflict.

yaron2:
@ItalyPaleAle are we good to merge this?

.github/workflows/conformance.yml
```diff
@@ -319,10 +370,6 @@ jobs:
       run: docker-compose -f ./.github/infrastructure/docker-compose-rabbitmq.yml -p rabbitmq up -d
       if: contains(matrix.component, 'rabbitmq')
 
-    - name: Start aws snssqs
```
artursouza, line 322: Should we keep it too? It can be helpful to run the tests for PRs and not wait for the code to be merged.

.github/workflows/conformance.yml
```diff
@@ -256,6 +259,42 @@ jobs:
           echo "$CERT_NAME=$CERT_FILE" >> $GITHUB_ENV
         done
 
+    - name: Setup Terraform
+      uses: hashicorp/setup-terraform@v2
+      if: matrix.terraform-dir != ''
```
ItalyPaleAle, line 264: Here too - I don't see this variable defined anywhere?

tests/config/pubsub/aws/snssqs/pubsub.yml
```diff
@@ -26,4 +24,6 @@ spec:
     - name: messageMaxNumber
       value: 10
     - name: concurrencyMode
-      value: "single"
\ No newline at end of file
+      value: "single"
+    - name: disableEntityManagement # Optional
```
ItalyPaleAle, line 28: Nit: remove the "Optional"

tests/conformance/common.go
```diff
@@ -160,6 +160,10 @@ func ParseConfigurationMap(t *testing.T, configMap map[string]interface{}) {
 				val = uuid.New().String()
 				t.Logf("Generated UUID %s", val)
 				configMap[k] = val
+			} else if strings.Contains(val, "${{") {
```
ItalyPaleAle, line 163: I wonder if maybe we should change the parsing to use Go templates, which would allow us to put variables also in-between. And also define a function to generate UUIDs.
https://pkg.go.dev/text/template

ItalyPaleAle:
I have a couple of comments but overall LGTM.

I wonder if before we merge this, we can merge it into a branch in this repo (not in a fork) so we can actually test the pipeline. @berndverst can we do that?

.github/workflows/certification.yml
```diff
@@ -223,6 +225,42 @@ jobs:
           echo "$CERT_NAME=$CERT_FILE" >> $GITHUB_ENV
         done
 
+    - name: Setup Terraform
+      uses: hashicorp/setup-terraform@v2
+      if: matrix.terraform-dir != ''
```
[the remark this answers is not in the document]
addjuarez: Its not defined yet since we have no certification components that currently use this. It should only run on components that need the resources

mukundansundar:
@addjuarez please add details in the conf test Readme file on how to setup and run the aws test infra and also if there are any specific conventions to follow for the test

tests/conformance/README.md
```diff
@@ -119,4 +119,63 @@ If you want to combine VS Code & dlv for debugging so you can set breakpoints in
...
+    operations: ["publish", "subscribe", "multiplehandlers"]
+    config:
+    pubsubName: aws-snssqs
+    testTopicName: ${{PUBSUB_AWS_SNSSQS_TOPIC}}
+    testMultiTopic1Name: ${{PUBSUB_AWS_SNSSQS_TOPIC_MULTI_1}}
+    testMultiTopic2Name: ${{PUBSUB_AWS_SNSSQS_TOPIC_MULTI_2}}
+```
+
+3. When writing your `component.yml` you should reference your credentials using env variables and any resources specified in the yaml should use env variables as well just as you did in the `test.yml`. Also
+`disableEntityManagement` should be set to `true` since we want to use only terraform to provision resources.
```
berndverst, line 149: Not every component even has entity management, so it is best if you explain what this is and that this option might be called something else, but should be deactivated.

tests/conformance/README.md
```diff
@@ -119,4 +119,63 @@ If you want to combine VS Code & dlv for debugging so you can set breakpoints in
...
+    description = "Unique Id of the github worklow run."
+}
+```
+
+5. The component should be added to the `cron-components` step in conformance test workflow `.github/conformance.yml`. The component should have a variable named `terraform-dir` and the value should be the relative path from `.github/infrastructure/terraform/conformance` to the folder which the tests personal terraform files are located such as `pubsub/aws/snsqsq`.
+
+```
+- component: pubsub.aws.snssqs.terraform      
+  terraform-dir: pubsub/aws/snssqs    
 ```
```
berndverst, line 181: Can you please create a separate section in the README for AWS specifically that explains what needs to be done when a new AWS conformance or certification test is being added? Isn't there a service principal / service account that needs to be updated? Who can do this and how?

berndverst (changes requested):

.github/infrastructure/terraform/conformance/pubsub/aws/snssqs/snssqs.tf
```diff
@@ -0,0 +1,99 @@
...
+    }
+  }
+}
+
+provider "aws" {
+  region = "us-east-1"
+  default_tags {
+    tags = {
+      Purpose  = "AutomatedTesting"
+    }
```
tmacam, line 17: Could we add the CreationDate here too? So we could GC anything that is 2d old with no risk of impacting running tests.

berndverst: I thought we tear down everything immediately after the test run anyway?

tmacam: Ideally, yes. But if the `Terraform Destroy` step fails or isnt reached for whatever reason, there will be left over resources lingering around.

By tagging them we can identify those resources more easily from, say, any resource in that account that needs to be long lived. And it also allows us to limit the scope of the permissions required for the garbage collector.

berndverst: I think we should do that but it can be a follow-up PR

berndverst (approved):

tests/config/pubsub/aws/snssqs/pubsub.yml
```diff
@@ -7,14 +7,12 @@ spec:
...
-      value: "http://localhost:4566"
     - name: accessKey
-      value: "my-access"
+      value: ${{AWS_ACCESS_KEY_ID}}
     - name: secretKey
-      value: "my-secret"
+      value: ${{AWS_SECRET_ACCESS_KEY}}
     - name: region
-      value: "us-east-1"
+      value: "us-west-2"
```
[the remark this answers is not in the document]
berndverst: Most of us are you on the West coast I guess :)

tests/conformance/common.go
```diff
@@ -167,6 +167,10 @@ func ParseConfigurationMap(t *testing.T, configMap map[string]interface{}) {
...
+				s := strings.TrimSpace(strings.TrimSuffix(strings.TrimPrefix(val, "${{"), "}}"))
```
berndverst, line 171: What if you have `${{bla}} ${stuff}}`? This would fail then.

berndverst: I will accept this for now -- but I will break the moment someone tries to use two variables. So I'm with @ItalyPaleAle that we should use Go templates instead.

tests/conformance/common.go
```diff
@@ -195,6 +199,10 @@ func parseConfigurationInterfaceMap(t *testing.T, configMap map[interface{}]inte
...
```
berndverst, line 203: same issue here -- it only supports a single variable substitution and will break if there are multiple such vars

berndverst (approved):

tests/conformance/README.md
```diff
@@ -119,4 +119,76 @@ If you want to combine VS Code & dlv for debugging so you can set breakpoints in
...
+5. The component should be added to the `cron-components` step in conformance test workflow `.github/conformance.yml`. The component should have a variable named `terraform-dir` and the value should be the relative path from `.github/infrastructure/terraform/conformance` to the folder which the tests personal terraform files are located such as `pubsub/aws/snsqsq`.
+
+    ```
+    - component: pubsub.aws.snssqs.terraform      
+    terraform-dir: pubsub/aws/snssqs    
+    ```
+
+## Adding new AWS component in github actions
+
+1. For tests involving aws components we use a service account to provision the resources needed. If you are contributing a brand new component you will need to make sure our account has sufficient permissions to provision resources and use handle component. A Dapr STC/Maintainer will have to update the service account so contact them for assistance.
```
berndverst, line 184:
```suggestion
1. For tests involving aws components we use a service account to provision the resources needed. If you are contributing a brand new component you will need to make sure our account has sufficient permissions to provision resources and use handle component. A Dapr STC member will have to update the service account so contact them for assistance.
```

berndverst, line 184:
```suggestion
1. For tests involving aws components we use a service account to provision the resources needed. If you are contributing a brand new component you will need to make sure our account has sufficient permissions to provision resources and use handle component. A Dapr STC member will have to update the service account so contact them for assistance.
```

berndverst (approved):

berndverst (approved):

tests/config/pubsub/aws/snssqs/terraform/pubsub.yml
```diff
@@ -0,0 +1,29 @@
...
+spec:
+  type: pubsub.aws.snssqs
+  version: v1
+  metadata:
+    - name: accessKey
+      value: ${{AWS_ACCESS_KEY_ID}}
+    - name: secretKey
+      value: ${{AWS_SECRET_ACCESS_KEY}}
+    - name: region
+      value: "us-east-1"
```
berndverst, line 15: shouldn't this also be `us-west-1`?

.github/infrastructure/terraform/conformance/pubsub/aws/snssqs/snssqs.tf
```diff
@@ -0,0 +1,105 @@
...
+    description = "Timestamp of the github worklow run."
+}
+
+variable "UNIQUE_ID" {
+    type        = string
+    description = "Unique Id of the github worklow run."
+}
+
+provider "aws" {
+  region = "us-east-1"
```
berndverst, line 23: should this have been `us-west-1`?

PR #2379, opened 2022-12-20
Changed postgres state store to use pgx instead of db/sql

ItalyPaleAle:
# Description

We currently have 3 Postgres components: state store, binding, and config store. The state store was the only one that used the database/sql interface to connect to Postgres rather than the "native" pgx interface (pgx was still used as the driver, but with the database/sql interface on top). 

Per [pgx docs](https://github.com/jackc/pgx#choosing-between-the-pgx-and-databasesql-interfaces), the pgx interface is recommended in our case as it's faster and has more features.

This PR changes the postgres state store to use the pgx interface. This comes with no user-facing changes (ie. feature set remains the same).

**EDIT:** Per review feedback, this adds a new metadata option `timeoutInSeconds` to control timeouts for operations.

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Docs issue: https://github.com/dapr/docs/issues/2966


state/postgresql/postgresdbaccess.go
```diff
@@ -246,32 +265,37 @@ func (p *PostgresDBAccess) doSet(parentCtx context.Context, db dbquerier, req *s
...
+	if result.RowsAffected() != 1 {
 		return errors.New("no item was updated")
 	}
 
 	return nil
 }
 
 func (p *PostgresDBAccess) BulkSet(parentCtx context.Context, req []state.SetRequest) error {
-	tx, err := p.db.BeginTx(parentCtx, nil)
+	ctx, cancel := context.WithTimeout(parentCtx, 30*time.Second)
```
mukundansundar, line 285: can we have all timeouts be defined as constants at least? no magic numbers please.

mukundansundar: the same question here, should this be a configurable param?

state/postgresql/postgresdbaccess.go
```diff
@@ -246,32 +265,37 @@ func (p *PostgresDBAccess) doSet(parentCtx context.Context, db dbquerier, req *s
...
+	tx, err := p.db.Begin(ctx)
+	cancel()
 	if err != nil {
 		return fmt.Errorf("failed to begin transaction: %w", err)
 	}
-	defer tx.Rollback()
+	defer func() {
+		rollbackCtx, rollbackCancel := context.WithTimeout(parentCtx, 30*time.Second)
```
mukundansundar, line 292: same as above.

state/postgresql/postgresdbaccess.go
```diff
@@ -299,20 +325,19 @@ func (p *PostgresDBAccess) Get(parentCtx context.Context, req *state.GetRequest)
 	var (
 		value    []byte
 		isBinary bool
-		etag     uint64 // Postgres uses uint32, but FormatUint requires uint64, so using uint64 directly to avoid re-allocations
```
mukundansundar, line 302: as seen in lines below, are reallocations ok?

state/postgresql/postgresdbaccess.go
```diff
@@ -389,11 +408,20 @@ func (p *PostgresDBAccess) doDelete(parentCtx context.Context, db dbquerier, req
...
 	if err != nil {
 		return fmt.Errorf("failed to begin transaction: %w", err)
 	}
-	defer tx.Rollback()
+	defer func() {
+		rollbackCtx, rollbackCancel := context.WithTimeout(parentCtx, 30*time.Second)
+		rollbackErr := tx.Rollback(rollbackCtx)
+		rollbackCancel()
+		if rollbackErr != nil && !errors.Is(rollbackErr, pgx.ErrTxClosed) {
+			p.logger.Errorf("Failed to rollback transaction in BulkDelete: %v", rollbackErr)
```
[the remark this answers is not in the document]
mukundansundar: should we let the user control the context timeouts if any?

state/postgresql/postgresdbaccess.go
ItalyPaleAle: It wasn't a choice as pgx doesn't support decoding a XID into uint64. Not sure if this allocation stays on the stack or escapes to the heap, however.

state/postgresql/postgresdbaccess.go
```diff
@@ -389,11 +408,20 @@ func (p *PostgresDBAccess) doDelete(parentCtx context.Context, db dbquerier, req
...
+		}
```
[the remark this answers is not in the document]
ItalyPaleAle: Changed

ItalyPaleAle:
@mukundansundar changed per your feedback, making timeouts configurable

state/postgresql/postgresql_query.go
```diff
@@ -140,8 +139,8 @@ func (q *Query) Finalize(filters string, qq *query.Query) error {
 	return nil
 }
 
-func (q *Query) execute(ctx context.Context, logger logger.Logger, db *sql.DB) ([]state.QueryItem, string, error) {
-	rows, err := db.QueryContext(ctx, q.query, q.params...)
+func (q *Query) execute(ctx context.Context, logger logger.Logger, db dbquerier) ([]state.QueryItem, string, error) {
+	rows, err := db.Query(ctx, q.query, q.params...)
```
mukundansundar, line 143: should `defer rows.Close()` be called before error check? 
```
The returned Rows must be closed before the connection can be used again. It is safe to attempt to read from the returned Rows even if an error is returned. The error will be the available in rows.Err() after rows are closed. It is allowed to ignore the error returned from Query and handle it in Rows.
```
See https://pkg.go.dev/github.com/jackc/pgx/v5#Conn.Query

ItalyPaleAle: Looks like the Close() call should be done after the error check. See example: https://pkg.go.dev/github.com/jackc/pgx/v5#example-Conn.Query

mukundansundar (approved):
Lgtm

state/postgresql/postgresdbaccess.go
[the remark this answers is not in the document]
DeepanshuA: Sorry, what was the consensus here? Shouldn't we have a metric representing rollback failures, as it can be an important aspect for user to look into.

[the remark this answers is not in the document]
ItalyPaleAle: I am not sure what conditions would cause the rollback to fail. aside from the transaction already being closed (which is an error we expect to get here and it's fine), I can only think of a network issue. In that case, however, if the connection is dropped while a transaction is active (and un-committed), then it's automatically rolled back.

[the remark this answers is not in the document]
ItalyPaleAle: Worth adding also that this behavior isn't new and the current code behaves the same way - except there errors were completely ignored and now are at least logged.

berndverst (approved):

Issue #2364, opened 2022-12-12
multi environment on aws sqs/sns


Sbolder:
## In what area(s)?

> /area runtime

## What version of Dapr?

> Runtime version: 1.9.4

## Expected Behavior
In the same aws account i have created two different eks cluster for env (dev - preprod).
With correct naming convetion dapr on start-up create:
1 sqs with two subscriber
example
sqs: event-service
sns : rn:aws:sns:eu-central-1:<account-id>:preprodEvent
sns : rn:aws:sns:eu-central-1:<account-id>:devEvent
and create correct policy to permit at sqs to send message at the correct sns

polixy expected:
```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "Service": "sns.amazonaws.com"
      },
      "Action": "sqs:SendMessage",
      "Resource": "arn:aws:sqs:eu-central-1:397194520445:bnpl-contracts-engine-service",
      "Condition": {
        "ArnEquals": {
          "aws:SourceArn": "arn:aws:sns:eu-central-1:397194520445:devContractEntityChangedEvent"
        }
      }
    },
    {
      "Effect": "Allow",
      "Principal": {
        "Service": "sns.amazonaws.com"
      },
      "Action": "sqs:SendMessage",
      "Resource": "arn:aws:sqs:eu-central-1:397194520445:bnpl-contracts-engine-service",
      "Condition": {
        "ArnEquals": {
          "aws:SourceArn": "arn:aws:sns:eu-central-1:397194520445:uatContractEntityChangedEvent"
        }
      }
    }
  ]
}
```

## Actual Behavior

Dapr create all resource correclty but the second service (subscriber) change sqs policy and overwrite the existing policy.
Unfortunatly if i reboot all services, the second services win and work done, but the first loose thei subscription and dosn't work.

the second service that startup on daprd log:
```
time="2022-12-07T10:59:14.760932156Z" level=error msg="error setting sns-sqs subscription policy: error setting queue subscription policy: InvalidAttributeValue: Invalid value for the parameter Policy.\n\tstatus code: 400, request id: 2e698073-5ee2-51b7-9a8c-dc792aa0c78c" app_id=event-service instance=event-webapi-8598984688-lqdcs scope=dapr.contrib type=log ver=1.9.4
[...]

PR #2386, opened 2022-12-22
fix bug multi environment on aws sqs/sns #2364

Sbolder:
Signed-off-by: Davide Boldrin <davi1994@hotmail.it>

# Description

I have changed the metod to create a policy on pubsub aws subcription, and now support multiEnvironment.

## Issue reference

Please reference the issue this PR will close: #2364
https://github.com/dapr/components-contrib/issues/2364
## Checklist

* [X ] Code compiles correctly
* [ X] Created/updated tests


pubsub/aws/snssqs/policy.go
```diff
@@ -74,18 +44,17 @@ type policy struct {
...
-					return true
-				}
+			if s.Condition.ValueArnEquals.AwsSourceArn == snsArn {
+				return true
 			}
+
 			// insert it if it does not exist
-			p.Statement[i].Condition.ForAllValuesArnEquals.AwsSourceArn = append(p.Statement[i].Condition.ForAllValuesArnEquals.AwsSourceArn, snsArn)
-			return false
+			//p.Statement[i].Condition.ForAllValuesArnEquals.AwsSourceArn = append(p.Statement[i].Condition.ForAllValuesArnEquals.AwsSourceArn, snsArn)
```
ItalyPaleAle, line 56: Please remove code that's now unused, rather than just commenting it out.

ItalyPaleAle (changes requested):
Ciao Davide, thanks for this PR.

Can you please explain how this fixes the issue you've reported?

Also, please fix the DCO issue: https://docs.dapr.io/contributing/contributing-overview/#developer-certificate-of-origin-signing-your-work

Sbolder:
@ItalyPaleAle 
 in the previous version the method "tryInsertCondition" try to create one statement for each resource, and append a new condition for each new sns, if you have one aws account with two environment this generate a bad request on startup of second service.

With my fix, now the method create one statement for each resource so, the relation until sqs ans sns is 1..1 on policy, and is more clear to understand.
Otherwhise if you have for example:
on env "dev" one publisher and one subscriber
on env "test" one publisher and one subscriber
with my fix each services write their policy and the second service dosn't change other service policy, is more linear and clear.

please feel free to contact me for any questions.
D.

berndverst:
Will consider merging this after we merge the SNS/SQS certification test. I don't know enough about this change so will need input from community members with more knowledge of SNS/SQS

yaron2:
cc @amimo

amimimor:
> #2364

@mukundansundar, @yaron2, others, please correct me if I'm wrong: issue #2364 describes a feature that can be summarized as a "Fan-In" pattern, meaning: two producer topics (2 SNSs) possibly originating from different environments (2 EKSs), and one subscriber queue (SQS, possibly in a different cluster). 

Just to clarify, currently, the AWS SNS/SQS PubSub component supports either an at-least 1:1 ratio of sns-topic:sqs-queue, which *possibly can support Fan-In* (if many publisher apps push messages to a given SNS topic), or at-least 1:many (many on the consuming side), using the component default behavior (it would connect 1 SNS topic to many SQSs, each dedicated per Dapr consuming app). The problem that #2364 wants to solve is therefore quite unique, where many publisher apps,  want to push messages to a queue by using different topics - which I fail to understand why does one need 2 or more topics to send messages that would eventually reach a single queue, if they could send, from many producer apps, messages to a given topic.

From my experience, the way we use EKS and AWS environments in general, we would never allow cross-EKS resource sharing and *personally* I am strongly against encouraging behaviors that might, due to lack of awareness/knowledge, result in the creation of a security vulnerability - e.g. data leakage from one EKS cluster to another. For that reason, if the user wishes to do whatever they want, we have introduced the ability to disable the automatic provisioning and configuration of resources and their security policies.

yaron2:
cc @Sbolder

Sbolder:
Hello  @yaron2 , @amimimor 

I think I didn't explain myself well.
I'll try to summarize my pull request.

The requirement is to have two EKS clusters on the same account (eks-develop and eks-test).
Now the problem arises when I deploy the first environment, the following are created:
1 sqs (writer)
1 sns (reader-dev)
with all the policies created correctly.

When I deploy the second test environment, it is recreated (or maybe not because it already exists):
1 sqs (writer) with the same name as the entity managed by Dapr
1 sns (reader-test)

The problem at this point is that the policies are generated incorrectly, not allowing both sns to subscribe to the sqs, but only one of them is able to do it and it works.

I understand the concern that there may be a exchange of messages between different EKS clusters (something to be avoided at all costs), but as far as I understand how Dapr works, it always reads only from the sns queue that it creates at startup, so I don't think a exchange of messages between different EKS clusters is possible.

I hope I was clearer.

Issue #2325, opened 2022-11-29
Conformance Test for Config API


shivamkm07:
## Describe the feature
Currently we don't have a conformance test for Config API at all. It needs to be added as per this [guide](https://docs.dapr.io/operations/components/certification-lifecycle/#conformance-tests)

## Release Note

RELEASE NOTE: **ADD** Conformance test for Config API

shivamkm07:
@berndverst Can you please add this to 1.10?

shivamkm07:
Conformance flow:

1. Init component
2. Initialize configuration store with some keys
3. Get
   - Single key: get data for key
   - Multiple keys: get data for all the requested keys
   - Empty list: get data for all keys in config store
     - For multiple versions of key in config store, get the latest semantic version: Will be added after this is merged: https://github.com/dapr/components-contrib/issues/2187

4. Subscribe

   - Subscribe with a set of keys:
     -  Update one subscribed key: get update for the key
     - Update multiple subscribed keys: get update for all the subscribed keys updated
     - Update non-subscribed keys: no updates
   - Subscribe with empty keys:
     - Update one subscribed key: get update for the key
     - Update multiple subscribed keys: get update for all the subscribed keys updated
     - Add a new key: get updates for the key
  
5. Unsubscribe

   - Unsubscribe to all set of subscribed keys
   - Update the keys from the set of subscribed keys: no update
   - Add a new key: no updates

PR #2387, opened 2022-12-23
Conformance test for Config API

shivamkm07:
Signed-off-by: shivam <shivamkm07@gmail.com>

# Description
Adds Conformance test for Config API testing Redis Configuration Store (both v6 and v7). It also fixes a bug in redis unsubscribe operation

Also currently `configuration.UpdateEvent` doesn't have a way to notify user of the operation performed (e.g. set, delete etc.). Added this as well as logic for handling delete key scenario in redis. 

## Issue reference

Please reference the issue this PR will close: #2325 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


configuration/redis/redis.go
```diff
@@ -265,32 +265,34 @@ func (r *ConfigurationStore) Get(ctx context.Context, req *configuration.GetRequ
 
 func (r *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.SubscribeRequest, handler configuration.UpdateHandler) (string, error) {
 	subscribeID := uuid.New().String()
+	keyStopChanMap := make(map[string]chan struct{})
 	if len(req.Keys) == 0 {
 		// subscribe all keys
 		stop := make(chan struct{})
-		r.subscribeStopChanMap.Store(subscribeID, stop)
+		keyStopChanMap[keySpaceAny] = stop
```
pravinpushkar, line 272: Should we use synchronised map here also ?

configuration/redis/redis.go
```diff
@@ -324,19 +327,30 @@ func (r *ConfigurationStore) handleSubscribedChange(ctx context.Context, req *co
...
-		Metadata: req.Metadata,
-		Keys:     []string{targetKey},
-	})
-	if err != nil {
-		r.logger.Errorf("get response from redis failed: %s", err)
-		return
+	operation := msg.Payload
+	var items map[string]*configuration.Item
+
+	if operation == "del" {
```
pravinpushkar, line 333: why we are checking for del operation? AFAIK config api does not support `del` ops.

shivamkm07: Config API doesn't support delete as well as add/update keys. But we expect notification after subscribing if a key is updated/deleted. Currently if a redis subscribed key is deleted, it errors out in redis as it tries to `get` the keyValue but it is deleted. So it needs separate handling

configuration/redis/redis.go
shivamkm07: `keyStopChanMap` is unique for a given subscription ID, unlike `subscribeStopChanMap` which is a common map for all subscription ids. imo it should be okay for this to be normal map.

pravinpushkar: got it. make sense.

shivamkm07:
@berndverst @ItalyPaleAle tagging for early reviews.

configuration/redis/redis.go
```diff
@@ -243,6 +243,10 @@ func (r *ConfigurationStore) Get(ctx context.Context, req *configuration.GetRequ
 
 		redisValue, err := r.client.Get(ctx, redisKey).Result()
 		if err != nil {
+			if err == redis.Nil {
+				r.logger.Warnf("redis key %s does not exist, ignore it\n", redisKey)
```
pravinpushkar, line 247:
```suggestion
				r.logger.Warnf("redis key %s does not exist, ignoring it\n", redisKey)
```

tests/utils/configupdater/redis/redis.go
```diff
@@ -0,0 +1,107 @@
...
+	separator = "||"
+)
+
+type metadata struct {
+	Host     string
+	Password string
+}
+
+type ConfigUpdater struct {
+	client redis.UniversalClient
```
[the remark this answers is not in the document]
berndverst: You should add ConfigUpdater as an interface, and then implement in both the v8 and v9 client files within the internal package.

berndverst (changes requested):

tests/utils/configupdater/redis/redis.go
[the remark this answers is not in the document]
shivamkm07: Resolved using internal RedisClient interface.

berndverst:
Please run your conformance test locally with redis 7. You will see it failing.

Similar to the RedisClient interface, you need to create a ConfigUpdater interface here:
https://github.com/dapr/components-contrib/blob/master/internal/component/redis/redis.go#L57

And then you need to provide implementations in both the v8 and v9 client (which correspond to redis 6 and 7).

https://github.com/dapr/components-contrib/blob/master/internal/component/redis/v8client.go
https://github.com/dapr/components-contrib/blob/master/internal/component/redis/v9client.go

berndverst (approved):

berndverst:
I ran the conformance tests locally with Redis 6 and Redis 7

PR #2391, opened 2022-12-27
Update & Fix Actions

ItalyPaleAle:
- Correct the path for the cloudflare.workerskv config file (fixes conformance tests failures in master branch currently)
- Update all Actions to newer versions to avoid warnings about Node.js 12 being deprecated soon (examples: https://github.com/dapr/components-contrib/actions/runs/3788549043 )
- Removed the deprecated Azure/get-keyvault-secrets action


berndverst (approved):

.github/scripts/dapr_bot.js
```diff
@@ -145,35 +142,14 @@ async function cmdAssign(github, issue, username, isFromPulls) {
...
+    await github.rest.issues.addAssignees({
         owner: issue.owner,
         repo: issue.repo,
         issue_number: issue.number,
         assignees: [username],
     });
 }
 
-/**
- * Comment a funny joke.
```
berndverst, line 157: You cannot remove this without discussing with maintainers first. That's not ok.

Now I happen to know that this isn't being used at all and I agree that we can remove it, but please run this kind of change by maintainers first in the future.

berndverst (approved):

PR #2395, opened 2022-12-29
optimize: get logger from context and remove unused logger params.

1046102779:
Signed-off-by: seachen <seachen@tencent.com>

### DIFF
--- a/bindings/azure/servicebusqueues/servicebusqueues.go
+++ b/bindings/azure/servicebusqueues/servicebusqueues.go
@@ -50,7 +50,8 @@ func NewAzureServiceBusQueues(logger logger.Logger) bindings.InputOutputBinding 
 
 // Init parses connection properties and creates a new Service Bus Queue client.
 func (a *AzureServiceBusQueues) Init(metadata bindings.Metadata) (err error) {
-	a.metadata, err = impl.ParseMetadata(metadata.Properties, a.logger, (impl.MetadataModeBinding | impl.MetadataModeQueues))
+	ctx := logger.NewContext(context.Background(), a.logger)
+	a.metadata, err = impl.ParseMetadata(ctx, metadata.Properties, (impl.MetadataModeBinding | impl.MetadataModeQueues))
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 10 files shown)


ItalyPaleAle:
Hi @1046102779 I am not sure I support this PR, sorry :(

It seems that it changes the signature of methods that parse metadata to accept a context rather than a logger. Because the context is not used for anything else, I'm not sure if this PR actually brings any benefit - it's just replacing one object with another, and it may actually add more overhead.

ItalyPaleAle (changes requested):
See comment

1046102779:
> Hi @1046102779 I am not sure I support this PR, sorry :(
> 
> It seems that it changes the signature of methods that parse metadata to accept a context rather than a logger. Because the context is not used for anything else, I'm not sure if this PR actually brings any benefit - it's just replacing one object with another, and it may actually add more overhead.

The life cycle of a service, including initialization, running and ending, if each process uses the Context mechanism to add init timeout control, it means that the service is better. I see that some methods of third-party components do not have context parameters in the initialization process. It is a good thing to add it. At the same time, logger is written into a context as an instance parameter and transmitted in the process. It is also very simple to use. At the same time, this logger also carries information, such as: service name, scope, or trace information, etc.

berndverst:
I also do not see the value here. The particular components you changed do not need a context during Init. The parsing of component metadata itself doesn't require context. Many components do not make any calls against the service during Init - therefore they also do not need a context there.

This new context addition in service bus Init case for example only adds more confusion for me.

1046102779:
> I also do not see the value here. The particular components you changed do not need a context during Init. The parsing of component metadata itself doesn't require context. Many components do not make any calls against the service during Init - therefore they also do not need a context there.
> 
> This new context addition in service bus Init case for example only adds more confusion for me.

How about this?
state/cockroachdb/cockroachdb_access.go:385
```golang
-         data, token, err := stateQuery.execute(ctx, p.logger, p.db)
+	  data, token, err := stateQuery.execute(ctx, p.db)
```

---

I also feel strange to pass `logger.Logger` as a parameter into other methods. My understanding is that the logger.Logger instance is not really a valid business parameter, it is passed in purely for log output.

berndverst:
Removing unused arguments is fine.

In some cases we defined helper methods where we might want to log something. The only way to do that is to explicitly pass in the logger.

Issue #2284, opened 2022-11-14
Two Way SSL configuration for HTTP Binding 


vikas2k14:
We are using dapr HTTP binding for all the external calls in our application. Our external application needs two-way SSL for communication in one of our use cases.

I wanted to check if it's supported right now and how to configure it.

cc : @rahulopengts @akhilac1 @mukundansundar @pravinpushkar

rabollin:
@vikas2k14  - Would you please add few more details on why mTLS will not solve the problem of 2 way SSL.

vikas2k14:
> @vikas2k14 - Would you please add more details on why mTLS will not solve the problem of 2 way SSL.

@rabollin as discussed offline, it is an HTTP binding call that will go outside the cluster (not service-to-service invocation ); hence, I don't think dapr mTLS will handle it.


Attaching a sample diagram for the same
![SSL drawio-2](https://user-images.githubusercontent.com/37506429/207555028-eb4ae944-0064-462f-b250-7352f2731e93.png)

pravinpushkar:
I think we will have to provide 3 options to pass `ca certificate`, `client certificate` and `client key` to the tls config option.
We can pass these options as encoded strings in the component yaml files and read the it `http.go`. This way it will work both in self hosted as well as in k8s mode.

ItalyPaleAle:
@pravinpushkar agree with your plan.

Please consider allowing the values of those metadata properties to be either:

- a local file: first, test if there's a file at the given location
- if there's no existing file, assume it's a PEM-encoded value directly

pravinpushkar:
> @pravinpushkar agree with your plan.
> 
> Please consider allowing the values of those metadata properties to be either:
> 
> * a local file: first, test if there's a file at the given location
> * if there's no existing file, assume it's a PEM-encoded value directly

Sure.. sounds good!

PR #2399, opened 2023-01-02
HTTP binding: add TLS config options for client cert verification

pravinpushkar:

# Description

Adding tls config options for http client

## Issue reference

Please reference the issue this PR will close: #2284 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


bindings/http/http.go
```diff
@@ -53,9 +68,17 @@ func NewHTTP(logger logger.Logger) bindings.OutputBinding {
 
 // Init performs metadata parsing.
 func (h *HTTPSource) Init(metadata bindings.Metadata) error {
-	if err := mapstructure.Decode(metadata.Properties, &h.metadata); err != nil {
+	var err error
+	if err = mapstructure.Decode(metadata.Properties, &h.metadata); err != nil {
 		return err
 	}
+	var tlsConfig *tls.Config
+	if h.metadata.MTLSEnable == "true" {
```
ItalyPaleAle, line 76: Do we need this option? Or can we just infer it if mtlsClientCert, mtlsClientKey, mtlsRootCA are set?

*If* we do need this option, please use `utils.IsTruthy` to check for truthy values

bindings/http/http_test.go
```diff
@@ -364,3 +369,66 @@ func TestNon2XXErrorsSuppressed(t *testing.T) {
...
+			err:        "x509: certificate signed by unknown authority",
+			statusCode: 200,
+		},
+	}
+
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			req := tc.ToInvokeRequest()
+			response, err := hs.Invoke(context.TODO(), &req)
+			// It will error out in the 1st phase of TLS handshake only as the server certificate is not trusted by the client.
```
ItalyPaleAle, line 428: Can you also add a test that succeeds please?

bindings/http/http.go
```diff
@@ -81,6 +107,59 @@ func (h *HTTPSource) Init(metadata bindings.Metadata) error {
...
+
+// getPemBytes returns the PEM encoded bytes from the provided certName and certData.
+// If the certData is a file path, it reads the file and returns the bytes.
+// Else if the certData is a PEM encoded string, it returns the bytes.
+// Else it returns an error.
+func (h *HTTPSource) getPemBytes(certName, certData string) ([]byte, error) {
+	// Read the file
+	pemBytes, err := os.ReadFile(certData)
+	// If there is an error assume it is already PEM encoded string not a file path.
+	if err != nil {
```
ItalyPaleAle, line 148: Check if error is specifically `errors.Is(err, os.ErrNotExist)` to exclude other errors such as permissions ones

ItalyPaleAle (changes requested):

bindings/http/http.go
```diff
@@ -81,6 +107,59 @@ func (h *HTTPSource) Init(metadata bindings.Metadata) error {
...
+			return nil, fmt.Errorf("provided %q value is neither a valid file path or nor a valid pem encoded string", certName)
+		}
+		return []byte(certData), nil
+	}
+	return pemBytes, nil
+}
+
+// isValidPEM validates the provided input has PEM formatted block.
+func isValidPEM(val string) bool {
+	block, _ := pem.Decode([]byte(val))
```
DeepanshuA, line 159: Do we also need to validate further here to check `x509.ParseCertificate(block.Bytes)` ?
ref: https://pkg.go.dev/crypto/x509#example-Certificate.Verify

ItalyPaleAle: IMHO this is fine, if the PEM is not a valid X.509 certificate, it will fail parsing and will throw an error later on (plus, not all PEM files are certificates - some are keys). I don't even consider checking if it's a valid PEM necessary, since they will fail later...

bindings/http/http_test.go
ItalyPaleAle: I think `SSL_CERT_DIR` needs to be set before the tests start actually :(

bindings/http/http.go
pravinpushkar: removed, can be inferred from provided certs

bindings/http/http_test.go
pravinpushkar: not needed. client can add new root certs via - 
```
caCertPool := x509.NewCertPool()
ok := caCertPool.AppendCertsFromPEM(caCertBytes)
```

bindings/http/http.go
```diff
@@ -81,6 +107,59 @@ func (h *HTTPSource) Init(metadata bindings.Metadata) error {
...
+		RootCAs:      caCertPool,
+		Certificates: []tls.Certificate{cert},
+	}, nil
+}
+
+// getPemBytes returns the PEM encoded bytes from the provided certName and certData.
+// If the certData is a file path, it reads the file and returns the bytes.
+// Else if the certData is a PEM encoded string, it returns the bytes.
+// Else it returns an error.
+func (h *HTTPSource) getPemBytes(certName, certData string) ([]byte, error) {
```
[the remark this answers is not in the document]
pravinpushkar: This is to make consistent in both cases where cert files or cert pem data is given.

pravinpushkar:
@dapr/approvers-components-contrib Review comments are addressed.

bindings/http/http_test.go
```diff
@@ -147,7 +147,163 @@ func TestDefaultBehavior(t *testing.T) {
...
+}
+
+func TestHTTPSBinding(t *testing.T) {
+	handler := http.NewServeMux()
+	handler.HandleFunc("/testhttps", httpsHandler)
+	server := setupHTTPSServer(t, true, handler)
+	defer server.Close()
+	t.Run("get with https with valid client cert and clientAuthEnabled true", func(t *testing.T) {
+		certMap := map[string]string{
+			"MTLSEnable":     "true",
```
ItalyPaleAle, line 225: Nit: remove this

bindings/http/http_test.go
```diff
@@ -147,7 +147,163 @@ func TestDefaultBehavior(t *testing.T) {
...
+
+		req := TestCase{
+			input:      "GET",
+			operation:  "get",
+			metadata:   map[string]string{"path": "/testhttps"},
+			path:       "/testhttps",
+			err:        "",
+			statusCode: 200,
+		}.ToInvokeRequest()
+		response, err := hs.Invoke(context.TODO(), &req)
```
ItalyPaleAle, line 241: In tests, please use `context.Background()` because it's not a TODO that needs to be fixed in the future

bindings/http/http_test.go
```diff
@@ -147,7 +147,163 @@ func TestDefaultBehavior(t *testing.T) {
...
+		assert.Nil(t, err)
```
ItalyPaleAle, line 242: Nit:

```suggestion
		assert.NoError(t, err)
```

Better semantics also in case oferrors

ItalyPaleAle: (Same below many times)

bindings/http/http.go
```diff
@@ -73,7 +73,7 @@ func (h *HTTPSource) Init(metadata bindings.Metadata) error {
 		return err
 	}
 	var tlsConfig *tls.Config
-	if h.metadata.MTLSEnable == "true" {
+	if h.metadata.MTLSClientCert != "" && h.metadata.MTLSClientKey != "" && h.metadata.MTLSRootCA != "" {
```
ItalyPaleAle, line 76: Is the `rootCA` actually required? If you don't pass a CA cert, you can skip creating a cert pool, and the app will use the system's default cert pool (the CA trust store) to validate certificates. It should be a valid (albeit possibly uncommon) scenario

ItalyPaleAle (changes requested):

bindings/http/http.go
pravinpushkar: agree, this might be a possible scenario.

ItalyPaleAle (approved):

artursouza (approved):

Issue #1971, opened 2022-08-16
AWS SQS/SNS Pubsub as Stable


mcandeia:
- The [AWS SQS/SNS Pubsub](https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-aws-snssqs/) requires certification tests to be created.
- Add conformance tests if not present.
- TODO - Complete the list of items here to get this component to stable

mukundansundar:
todo

olitomlinson:
+1 for this

Issue #1339, opened 2021-11-26
PubSub AWS SNS/SQS Certification Tests


amimimor:
## Describe the feature
implement certification tests and stabilize AWS Pubsub SNS/SQS component

## Release Note

RELEASE NOTE:

berndverst:
This is sort of a sub item of #1971

PR #2416, opened 2023-01-06
AWS SNS/SQS PubSub Certification Tests - Exiting Queue/Topics tests

robertojrojas:
# Description
Provides the AWS SNS/SQS PubSub Certification Test Plan and Test Code. 
Tests for Existing Queue and Topics.
It depends on issue #2282 

## Issue reference
This PR makes progress towards #1339 and #1971 but does not complete these. More component metadata tests will need to be added.

### DIFF
--- a/.github/workflows/certification.yml
+++ b/.github/workflows/certification.yml
@@ -286,11 +286,17 @@ jobs:
     
     - name: Create aws.snssqs specific variables
       if: contains(matrix.component, 'snssqs')
+      working-directory: "./.github/infrastructure/terraform/certification/${{ matrix.terraform-dir }}"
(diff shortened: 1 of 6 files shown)


.github/workflows/certification.yml
```diff
@@ -286,11 +286,17 @@ jobs:
     
     - name: Create aws.snssqs specific variables
       if: contains(matrix.component, 'snssqs')
+      working-directory: "./.github/infrastructure/terraform/certification/${{ matrix.terraform-dir }}"
+      if: matrix.terraform-dir != ''
```
addjuarez, line 290: Two if statements here will cause an error in the workflow.

berndverst (approved):

Issue #2370, opened 2022-12-15
how to rewrite Component properties in RocketMq pubsub


naah69:
We want create a public rocketmq pubsub component in k8s crd without `producerGroup` and `consumerGroup`.

And we will set both of them on annotations or envirement vairables of depolyment.yaml.

is the way available？

naah69:
@jinjianfeng-chn 
now we have to add many rocketmq crd to implement that every service set different consumerGroup and producerGroup.
I am not good at Golang
Can u help me to add the feature that envirement vairable can rewrite consumerGroup and producerGroup?

berndverst:
@naah69 I will not accept the PR you opened. Loading random environment variables from within the component is now the correct pattern.

You can change your PR such that when no consumer group is specified it uses the default value that is autogenerated (which is the same as the app ID). This metadata key will be `consumerID`.

Another option is something like this:
Instead, what you can do is something with this.
https://docs.dapr.io/developing-applications/building-blocks/pubsub/howto-namespace/#with-namespace-consumer-groups

berndverst:
I have a better idea for you @naah69:

Configure the environment variable secret store.
https://docs.dapr.io/reference/components-reference/supported-secret-stores/envvar-secret-store/

Then configure your Dapr component,
and for the consumerGroup you would do something like in your component:

```
 metadata:

// other component metadata here

    - name: consumerGroup //
      secretKeyRef:
        name: MYCONSUMERGROUPENVVAR // can be anything you want
        key: MYCONSUMERGROUPENVVAR
    - name: producerGroup //
      secretKeyRef:
        name: MYPRODUCERGROUPENVVAR // can be anything you want
        key: MYPRODUCERGROUPENVVAR
auth:
  secretStore: envvar-secret-store
```

This assumed you have the following secret store also loaded:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: envvar-secret-store
  namespace: default
spec:
  type: secretstores.local.env
  version: v1
  metadata:
```

berndverst:
Keep in mind you can only reference a single secret store in a component at a time.

PR #2420, opened 2023-01-08
Add get rocketmq group from env to support multi apps use the same component and difference group 

naah69:
# Description

Add get rocketmq group from env to support multi apps use the same component and difference group

## Issue reference

https://github.com/dapr/components-contrib/issues/2370

### DIFF
--- a/pubsub/rocketmq/rocketmq.go
+++ b/pubsub/rocketmq/rocketmq.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"os"
 	"strconv"
 	"strings"
 	"sync"
@@ -119,7 +120,9 @@ func (r *rocketMQ) setUpConsumer() (mq.PushConsumer, error) {
 	if r.metadata.InstanceName != "" {
 		opts = append(opts, mqc.WithInstance(r.metadata.InstanceName))
 	}
-	if r.metadata.ConsumerGroup != "" {
+	if os.Getenv("ConsumerGroup") != "" {
+		opts = append(opts, mqc.WithGroupName(os.Getenv("ConsumerGroup")))
+	} else if r.metadata.ConsumerGroup != "" {
 		opts = append(opts, mqc.WithGroupName(r.metadata.ConsumerGroup))
 	} else if r.metadata.GroupName != "" {
 		r.metadata.ConsumerGroup = r.metadata.GroupName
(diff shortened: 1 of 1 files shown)


berndverst (changes requested):
As a maintainer I will not accept any changes that read environment variables. This is not the correct way to accomplish what you want to do.

You can change the code such that when no consumer group is specified it uses the default value that is autogenerated (which is the same as the app ID). This metadata key will be `consumerID`.

Another option is something like this:
Instead, what you can do is something with this.
https://docs.dapr.io/developing-applications/building-blocks/pubsub/howto-namespace/#with-namespace-consumer-groups

naah69:
> As a maintainer I will not accept any changes that read environment variables. This is not the correct way to accomplish what you want to do.
> 
> You can change the code such that when no consumer group is specified it uses the default value that is autogenerated (which is the same as the app ID). This metadata key will be `consumerID`.
> 
> Another option is something like this: Instead, what you can do is something with this. https://docs.dapr.io/developing-applications/building-blocks/pubsub/howto-namespace/#with-namespace-consumer-groups

The second way is a way to reslove my issue too,  i  will try it.
If it is not valid , i will try to develop the first idea.

berndverst:
@naah69 I commented on the issue you filed with another solution -- using the local secret store (environment variables). This way you can use environment variables directly as the value of the properties in the RocketMQ component yaml. That may be better for you! Take a look!

I'll close this PR for now. Please open a new PR if you want to implement the default consumer group name (by taking the value from `metadata.properties['consumerID']`)

naah69:
> @naah69 I commented on the issue you filed with another solution -- using the local secret store (environment variables). This way you can use environment variables directly as the value of the properties in the RocketMQ component yaml. That may be better for you! Take a look!
> 
> I'll close this PR for now. Please open a new PR if you want to implement the default consumer group name (by taking the value from `metadata.properties['consumerID']`)

thanks,my  issue had be resolved by the second way.

naah69:
> @naah69 I commented on the issue you filed with another solution -- using the local secret store (environment variables). This way you can use environment variables directly as the value of the properties in the RocketMQ component yaml. That may be better for you! Take a look!
> 
> I'll close this PR for now. Please open a new PR if you want to implement the default consumer group name (by taking the value from `metadata.properties['consumerID']`)

I had a new issue ,where is the document that describe the all vairable that can be use in yaml

Issue #2424, opened 2023-01-09
[AWS DynamoDB] Incorrect Implementation of the Partition Key Metadata Field


robertojrojas:
## Expected Behavior

In DynamoDB, the _**Primary Key**_ or _**Partition Key**_ attribute name does not always have to be named **`key`**.
By providing the Metadata Field named **`partitionKey`** with a different name, the DynamoDB component should
send the _partitionKey_ name instead of `key` for CRUD operations against the data.
See [AWS DynamoDB Developer Guide](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.CoreComponents.html#HowItWorks.CoreComponents.PrimaryKey)

For example:
```json
[
  {
    "value": "Bruce Wayne"
    "metadata": {
       "partitionKey": "personName"
    }
  }
]
```

## Actual Behavior
Currently, the DynamoDB State Store component incorrectly replaces the value of the **`key`** instead of the name of the _**key**_ with the value specified in the **`partitionKey`** metadata field, effectively sending the **`partitionKey`** as the value of to store instead of **`value`**.
This is occurring in the master branch [commit ID aae1f6c064191183dccf5fa5ac77c41d90eead3e](https://github.com/dapr/components-contrib/blob/aae1f6c064191183dccf5fa5ac77c41d90eead3e/state/aws/dynamodb/dynamodb.go#L93) 
The following PR  https://github.com/dapr/components-contrib/pull/2389 added the PartitionKey feature.

## Steps to Reproduce the Problem

1. Build dapr using the above components-contrib commit version.
2. Run dapr 
3. Using the following payload to save a state:
```json
[
  {
    "value": "Bruce Wayne"
    "metadata": {
       "partitionKey": "personName"
    }
  }
]

curl -X POST -H "Content-Type: application/json" -d '....' http://localhost:3500/v1.0/state/statestore
```

4. Get the state: 
```
curl http://localhost:3500/v1.0/state/statestore/name 
 ````
6. The state shows `personName` instead of `Bruce Wayne`

## Release Note

**FIX** Bug in runtime.

PR #2426, opened 2023-01-10
[AWS DynamoDB] Fixes Incorrect Implementation of the Partition Key Metadata Field

robertojrojas:
Signed-off-by: Roberto J Rojas <robertojrojas@gmail.com>

# Description
This PR fixes the implementation of Partition Key replacement. 
When the DynamoDB component is configured with metadata field that contains the property **`partitionKey`**, the code will replace the default Partition Key **`"key"`** with the value provided in the **`partitionKey`** property.

## Issue reference
PR will close: #2424

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#3052


state/aws/dynamodb/dynamodb.go
```diff
@@ -227,8 +233,8 @@ func (d *StateStore) BulkSet(ctx context.Context, req []state.SetRequest) error
 func (d *StateStore) Delete(ctx context.Context, req *state.DeleteRequest) error {
 	input := &dynamodb.DeleteItemInput{
 		Key: map[string]*dynamodb.AttributeValue{
-			"key": {
-				S: aws.String(populatePartitionMetadata(req.Key, req.Metadata)),
+			d.partitionKey: {
```
ItalyPaleAle, line 236: In the current version, the partition key is read from the request's metadata. Now, it's read from the component's metadata.

Shouldn't we continue allowing passing a partition key for each request?

ItalyPaleAle (changes requested):

state/aws/dynamodb/dynamodb.go
robertojrojas: Since the DynamoDB table needs to be created before hand and once the table is created the partition key can not be changed through the request, I felt that specifying the partition key only through the component's metadata made more sense. Unless, I'm missing something.

ItalyPaleAle: Oh I see. That's ok then.

ItalyPaleAle (approved):
Please open a PR in https://github.com/dapr/docs to document the changes to the metadata properties too

berndverst (approved):

Issue #2247, opened 2022-11-01
AWS S3 binding as Stable


artursouza:
Create a certification test plan.
Implement certification tests (must cover all component metadata options)
Mark component as "Stable".

robertojrojas:
Hi @johnewart Are you still working on this? if not, please let me know so I can assign to myself.

PR #2428, opened 2023-01-10
bindings.aws.s3: Improvements & fixes

ItalyPaleAle:
# Description

Some various improvements & fixes for the AWS S3 binding, based on static code review:

1. Use `metadata.DecodeMetadata` for parsing metadata, which also supports case-insensitive keys and uses `IsTruthy` for bools
2. Ensure error messages are consistent
3. Fixes possible (albeit not very likely) panic with UUID generation if kernel doesn't have enough entropy
4. Use a more JSON-aware Unquote method, taken from #2427

## Issue reference

Part of #2247


yaron2:
cc @robertojrojas

robertojrojas (commented):
@ItalyPaleAle  Does it make sense to split this PR into 2 PRs?  PR-1 for S3 Fixes and PR-2 Test certification go module updates?

ItalyPaleAle:
The go module updates are because I ran "make modtidy-all" due to moving a dependency from indirect to direct and upgrading dapr/kit. Can't have this PR without those changes I'm afraid, or the linter will fail.

robertojrojas:
> 



> The go module updates are because I ran "make modtidy-all" due to moving a dependency from indirect to direct and upgrading dapr/kit. Can't have this PR without those changes I'm afraid, or the linter will fail.

Sounds good. 👍

bindings/aws/s3/s3.go
```diff
@@ -152,40 +153,39 @@ func (s *AWSS3) create(ctx context.Context, req *bindings.InvokeRequest) (*bindi
...
-	} else {
-		key = uuid.New().String()
-		s.logger.Debugf("s3 binding: key not found. generating key %s", key)
-	}
 
-	d, err := strconv.Unquote(string(req.Data))
-	if err == nil {
-		req.Data = []byte(d)
+	key := req.Metadata[metadataKey]
+	if key == "" {
```
robertojrojas, line 158: I wonder if the following _if_ block should be extracted to its own function so it can be properly unit tested. On test scenario that came to me is the one where the `key` is just spaces. Would that be valid and/or appropriate in the context of S3?

ItalyPaleAle: IMHO this is one of those cases in which we should just accept whatever the user passes and if that's invalid, the service will complain. Otherwise, next we'd find ourselves dealing with users trying to send UTF-8 sequences and having to address that case too, where maybe one S3 implementation supports them and the other one doesn't.

(Also note that the behavior hasn't changed, I just simplified the check since retrieving a value that doesn't exist from a map returns the zero value automatically, in this case the empty string)

robertojrojas: Yeah I agree.

berndverst (approved):

berndverst:
@ItalyPaleAle update the PR and run `make modtidy-all` again.

berndverst:
This PR cannot be merged right now because bindings certification tests for most components are broken in master. Will wait for @DeepanshuA to address.

ItalyPaleAle:
@berndverst  is this ready to be merged now?

berndverst (approved):

berndverst:
@ItalyPaleAle we need another `make modtidy-all`

berndverst (approved):

Issue #2404, opened 2023-01-04
Add Bulk Subscribe Request Params in SubscribeRequest


DeepanshuA:
## Describe the feature
As per consensus reached in https://github.com/dapr/dapr/issues/5499, user would be able to provide bulkSubscribe options in its own dedicated specification, instead of being a part of metadata. Like this:
```
apiVersion: dapr.io/v1alpha1
kind: Subscription
metadata:
  name: orders-subscription-mqtt
spec:
  topic: orders
  route: /orders
  pubsubname: mqtt-pubsub
  bulkSubscribe: (optional)
    enabled: "true" (required)
    maxBulkSubCount: "4" (optional)
    maxBulkSubAwaitDurationMs: "10000" (optional)
  metadata:
    rawPayload: "true"
```
So, change is required to add BulkSubscribeRequest with requisite params as a part of `SubscribeRequest`.

## Release Note

RELEASE NOTE:

PR #2433, opened 2023-01-12
Set Dapr Running Flag and Drop dapr runtime fork replacement

DeepanshuA:
Signed-off-by: Deepanshu Agarwal <deepanshu.agarwal1984@gmail.com>

# Description

This PR is to:
1. Remove reference to DeepanshuA dapr runtime fork, as we have now required PRs merged in for dapr runtime. 
2. Set Running flag to true for dapr runtime, before calling shutdown.

Successful Run:
https://github.com/dapr/components-contrib/actions/runs/3902146129/jobs/6664858638

## Issue reference

Please reference the issue this PR is related to: #2404 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/tests/certification/flow/sidecar/sidecar.go
+++ b/tests/certification/flow/sidecar/sidecar.go
@@ -145,6 +145,7 @@ func Stop(appID string) flow.Runnable {
 func (s Sidecar) Stop(ctx flow.Context) error {
 	var client *Client
 	if ctx.Get(s.appID, &client) {
+		client.rt.SetRunning(true)
 		client.rt.Shutdown(2 * time.Second)
 
 		return client.rt.WaitUntilShutdown()
--- a/.build-tools/go.mod
+++ b/.build-tools/go.mod
@@ -18,5 +18,3 @@ require (
 	github.com/xeipuuv/gojsonreference v0.0.0-20180127040603-bd5ef7bd5415 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
-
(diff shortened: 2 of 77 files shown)


ItalyPaleAle (approved):

tests/certification/flow/sidecar/sidecar.go
```diff
@@ -145,6 +145,7 @@ func Stop(appID string) flow.Runnable {
 func (s Sidecar) Stop(ctx flow.Context) error {
 	var client *Client
 	if ctx.Get(s.appID, &client) {
+		client.rt.SetRunning(true)
```
ItalyPaleAle, line 148: This works here but I wonder if it should go in the `Start` method, just in case it gets used for other things in the runtime in the future

DeepanshuA: ohh, I assumed that we are already setting running flag to `true` in runtime.go's Run method, so when runtime would be invoked via cert test, then it would be set already - And, that we are just fixing some race conditions while stopping. Let me check, from your comment it seems that we are not at all setting `running` flag to true in case of cert tests.

ItalyPaleAle: Correct. The running flag is set in the runtime's 'Run' method. But that method is not invoked in cert tests because the "fake runtime" does its own initialization. And that's why our tests were failing, because the running flag was unset.

berndverst (approved):

PR #2438, opened 2023-01-13
Fix bug in ASB message renewal

jjcollinge:
- Measure length after taking lock

# Description

The length of the active messages was read before taking a lock on it. It was used to init the length a map that was then populated by reading the active messages after the lock was taken.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


berndverst (approved):

internal/component/azure/servicebus/subscription.go
```diff
@@ -361,8 +361,8 @@ func (s *Subscription) RenewLocksBlocking(ctx context.Context, receiver Receiver
 				s.logger.Debugf("Renewed session %s locks for %s", sessionReceiver.SessionID(), s.entity)
 			} else {
 				// Snapshot the messages to try to renew locks for.
-				msgs := make([]*azservicebus.ReceivedMessage, len(s.activeMessages))
 				s.mu.RLock()
+				msgs := make([]*azservicebus.ReceivedMessage, len(s.activeMessages))
 				for i, m := range s.activeMessages {
 					msgs[i] = m
```
mukundansundar, line 367: @jjcollinge Should this be `msgs = append(msgs, m)`?
Since activeMessages is a map and key is a sequenceNumber which can be more than the index value of the list `msgs`.

internal/component/azure/servicebus/subscription.go
```diff
@@ -361,8 +361,8 @@ func (s *Subscription) RenewLocksBlocking(ctx context.Context, receiver Receiver
 				s.logger.Debugf("Renewed session %s locks for %s", sessionReceiver.SessionID(), s.entity)
 			} else {
 				// Snapshot the messages to try to renew locks for.
-				msgs := make([]*azservicebus.ReceivedMessage, len(s.activeMessages))
 				s.mu.RLock()
+				msgs := make([]*azservicebus.ReceivedMessage, len(s.activeMessages))
```
mukundansundar, line 365: this can then be `msgs := make([]*azservicebus.ReceivedMessage, 0, len(s.activeMessages))`

internal/component/azure/servicebus/subscription.go
jjcollinge: Ah yes, good spot - I will revert to append 👌

ItalyPaleAle: No need for append then, which is not efficient with allocs 

use a separate I counter variable

PR #2440, opened 2023-01-14
Fix bug in ASB pubsub

jjcollinge:

# Description
[Reverts changes to active message snapshotting](https://github.com/dapr/components-contrib/pull/2346#discussion_r1052712437) as `activeMessages` is a map of message sequence number -> message and is being flattened to a slice. Reference: https://github.com/dapr/components-contrib/pull/2438#discussion_r1070259105. cc: @mukundansundar 

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/internal/component/azure/servicebus/subscription.go
+++ b/internal/component/azure/servicebus/subscription.go
@@ -363,8 +363,10 @@ func (s *Subscription) RenewLocksBlocking(ctx context.Context, receiver Receiver
 				// Snapshot the messages to try to renew locks for.
 				s.mu.RLock()
 				msgs := make([]*azservicebus.ReceivedMessage, 0, len(s.activeMessages))
+				var i int
 				for _, m := range s.activeMessages {
-					msgs = append(msgs, m)
+					msgs[i] = m
+					i++
 				}
 				s.mu.RUnlock()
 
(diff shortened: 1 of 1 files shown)


jjcollinge:
I've not been able to spin up the Azure test infra to run the Azure E2E tests myself (I am looking into doing this on Monday) so I'm not sure if this will fix the current failed tests. I am running some smoke tests against ASB - please hold off merging this until I've confirmed they are working ok.

mukundansundar (dismissed):
lgtm

ItalyPaleAle (dismissed):
LGTM but you could also use an external i counter (and then do i++ in the loop)

jjcollinge:
@ItalyPaleAle I've updated to directly assign to the map rather than using append. Are these cert tests failures related to these change or are they flakey tests?

jjcollinge:
So I've ran the Azure E2E tests against these changes and do get the same errors as seen in the workflow... However, even if I run the tests against a cluster with Dapr 1.9.5 installed I still get the same error (I checked the sidecar image was `docker.io/daprio/daprd:1.9.5` to ensure it's not running an edge build). Could someone check if this is actually isolated to these changes please?

ItalyPaleAle:
These test failures seem unrelated, so possibly flakiness.

jjcollinge:
So I think we should merge this change but this will not fix the E2E Azure tests (at least based on my testing). Fixing the E2E Azure tests is going to have to be followed up. As mentioned above, the tests appeared to fail against ASB for Dapr 1.7.x,1.8.x,1.9.x for me so that will require more investigation. Is it possible to run the "stable" versions of Dapr against the Azure test infra to confirm?

mukundansundar (approved):

PR #2446, opened 2023-01-18
Fix mqtt pubsub: remove republishing when application is unhealthy

shivamkm07:
# Description

MQTT pub-sub currently re-publishes the messages once the handler returns error. This causes same message being re-delivered indefinitely even to other applications which are healthy and subscribed to same topic. This PR removes re-publishing logic as well as associated metadata. 

**It also removes mosquitto from the list of supported brokers** because mosquitto doesn't retry messages after a timeout and so in case of handler failures, messages are not retried. 

## Issue reference

Please reference the issue this PR will close: #2309 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


pubsub/mqtt3/metadata.go
```diff
@@ -100,15 +97,6 @@ func parseMQTTMetaData(md pubsub.Metadata, log logger.Logger) (*metadata, error)
...
 	}
 
-	m.maxRetriableErrorsPerSec = defaultMaxRetriableErrorsPerSec
-	if val, ok := md.Properties[mqttMaxRetriableErrorsPerSec]; ok && val != "" {
-		var err error
-		m.maxRetriableErrorsPerSec, err = strconv.Atoi(val)
-		if err != nil {
-			return &m, fmt.Errorf("%s invalid maxRetriableErrorsPerSec %s, %s", errorMsgPrefix, val, err)
-		}
-	}
```
mukundansundar, line 110: aren't we removing the property now? I think this is a breaking change?

mukundansundar (commented):
I think this is a breaking change. And why are we removing inline retires? Do we have default pubsub resiliency policies in place instead of the inline retries? 
Also even if we have resiliency policies in place, we should not straight away remove the metadata property function. We should deprecate it and remove in 2 release as per dapr support policy.

shivamkm07:
> I think this is a breaking change. And why are we removing inline retires? Do we have default pubsub resiliency policies in place instead of the inline retries? Also even if we have resiliency policies in place, we should not straight away remove the metadata property function. We should deprecate it and remove in 2 release as per dapr support policy.

Inline retry was actually removed much earlier in a separate PR: https://github.com/dapr/components-contrib/pull/1798
The built-in retries (from sidecar to application) was removed and re-publishing logic was added, but re-publishing messages in case of failures led to clogging the whole broker in case even one application fails (See https://github.com/dapr/components-contrib/issues/2309#issuecomment-1328215282). The current PR just removes the re-publishing logic to fix this bug. 

Note that this would disable the retry completely unless user specifically applies the resiliency policies, and this is what is expected behaviour as well: https://github.com/dapr/components-contrib/issues/2309#issuecomment-1328323617. The same logic has also been implemented in rabbitmq: [enabling-message-delivery-retries](https://github.com/dapr/docs/blob/v1.9/daprdocs/content/en/reference/components-reference/supported-pubsub/setup-rabbitmq.md#enabling-message-delivery-retries.) 

As for if the change is breaking or not, again the current PR just removes the `maxRetriableErrorsPerSec` metadata added in https://github.com/dapr/components-contrib/pull/1798 for re-publishing logic. The original flag(present in v1.8.x releases) was `backOffMaxRetries` removed in above PR, and so if retries need to be supported for 2 releases, `backoffMaxRetries` need to be supported to account for breaking changes. But I am not sure if the change comes under breaking change. cc @ItalyPaleAle @berndverst

shivamkm07:
Removing backoffMaxRetries` or `maxRetriableErrorsPerSec` shouldn't break any user program though. It's just that these options would be ignored and so no retries would happen. Does that come under breaking change?

pubsub/mqtt3/metadata.go
ItalyPaleAle: Not necessarily a breaking change.

pubsub/mqtt3/metadata.go
```diff
@@ -121,5 +109,11 @@ func parseMQTTMetaData(md pubsub.Metadata, log logger.Logger) (*metadata, error)
 		log.Warnf("Metadata property 'backOffMaxRetries' for component pubsub.mqtt has been deprecated and will be ignored. See: https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-mqtt/")
 	}
 
+	// Deprecated config option
+	// TODO: Remove in the future
+	if _, ok := md.Properties["maxRetriableErrorsPerSec"]; ok {
+		log.Warnf("Metadata property 'maxRetriableErrorsPerSec' for component pubsub.mqtt has been deprecated and will be ignored. See: https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-mqtt/")
```
ItalyPaleAle, line 115:
```suggestion
		log.Warnf("Metadata property 'maxRetriableErrorsPerSec' for component pubsub.mqtt3 has been deprecated and will be ignored. See: https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-mqtt/")
```

.github/workflows/conformance.yml
```diff
@@ -67,7 +67,6 @@ jobs:
         - pubsub.hazelcast
         - pubsub.in-memory
         - pubsub.mqtt3-emqx
-        - pubsub.mqtt3-mosquitto
```
ItalyPaleAle, line 70: Should Mosquitto be removed from `bindings.mqtt3-mosquitto` too?

ItalyPaleAle (approved):

.github/workflows/conformance.yml
berndverst: Bindings is different because we don't make the same guarantees there. Should be fine to leave in bindings IMO.

berndverst (approved):

ItalyPaleAle (approved):

ItalyPaleAle:
Ok.

@shivamkm07 please, do update the docs to highlight:

- The removal of the metadata option
- That Mosquitto is not supported

PR #2447, opened 2023-01-18
Use sync.Map for ASB active messages

jjcollinge:
# Description
Moves the ASB implementation to use sync.Map instead of native map and mutex to handle active message tracking. This is because we've been seeing crashes in the E2E related to the active message access (even when all writes are protected a write mutex and all read protected by a read mutex 🤔). I can confirm that this change DOES stop the sidecar crashing during the Azure E2E. However, it DOES NOT fix the Azure E2E tests. They still fail. I've also ran the E2E tests against 1.9.5 and they also fail and do not include the most recent changes.

UPDATE: Interestingly reverting to using `append` also stops the crash.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/internal/component/azure/servicebus/subscription.go
+++ b/internal/component/azure/servicebus/subscription.go
@@ -363,8 +363,10 @@ func (s *Subscription) RenewLocksBlocking(ctx context.Context, receiver Receiver
 				// Snapshot the messages to try to renew locks for.
 				s.mu.RLock()
 				msgs := make([]*azservicebus.ReceivedMessage, 0, len(s.activeMessages))
+				var i int
 				for _, m := range s.activeMessages {
-					msgs = append(msgs, m)
+					msgs[i] = m
+					i++
 				}
 				s.mu.RUnlock()
 
--- a/internal/component/azure/servicebus/subscription.go
+++ b/internal/component/azure/servicebus/subscription.go
@@ -362,9 +362,9 @@ func (s *Subscription) RenewLocksBlocking(ctx context.Context, receiver Receiver
 			} else {
 				// Snapshot the messages to try to renew locks for.
 				s.mu.RLock()
-				msgs := make([]*azservicebus.ReceivedMessage, len(s.activeMessages))
-				for i, m := range s.activeMessages {
-					msgs[i] = m
+				msgs := make([]*azservicebus.ReceivedMessage, 0, len(s.activeMessages))
+				for _, m := range s.activeMessages {
+					msgs = append(msgs, m)
 				}
 				s.mu.RUnlock()
 
--- a/internal/component/azure/servicebus/subscription.go
+++ b/internal/component/azure/servicebus/subscription.go
@@ -49,8 +49,7 @@ type HandlerFn func(ctx context.Context, msgs []*azservicebus.ReceivedMessage) (
 // Subscription is an object that manages a subscription to an Azure Service Bus receiver, for a topic or queue.
 type Subscription struct {
 	entity               string
-	mu                   sync.RWMutex
-	activeMessages       map[int64]*azservicebus.ReceivedMessage
+	activeMessages       sync.Map
 	activeOperationsChan chan struct{}
 	requireSessions      bool // read-only once set
 	timeout              time.Duration
@@ -99,7 +98,6 @@ func NewSubscription(
 
 	s := &Subscription{
 		entity:          opts.Entity,
-		activeMessages:  make(map[int64]*azservicebus.ReceivedMessage),
 		timeout:         time.Duration(opts.TimeoutInSec) * time.Second,
 		maxBulkSubCount: *opts.MaxBulkSubCount,
 		requireSessions: opts.RequireSessions,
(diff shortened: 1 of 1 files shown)


mukundansundar:
> UPDATE: Interestingly reverting to using `append` also stops the crash.

Interesting. Just using `append` also works, but since acc to the PR all map access was behind locks, why did it panic when using index? 
Should we continue to use map and mutex locks instead of sync.Map? I think sync.Map is a better option. 
@jjcollinge thoughts?

jjcollinge:
I think it'd be safer to move to `sync.Map` as it's not clear to me why we were seeing the panic on the native map and how using `append` is avoiding that. Just to make it super clear, the E2E tests are still failing for me against these changes and on older stable versions of Dapr. It'd be useful if others could confirm whether they see the Azure E2E tests fail on versions of Dapr prior to this change set too.

mukundansundar:
> I think it'd be safer to move to `sync.Map` as it's not clear to me why we were seeing the panic on the native map and how using `append` is avoiding that. Just to make it super clear, the E2E tests are still failing for me against these changes and on older stable versions of Dapr. It'd be useful if others could confirm whether they see the Azure E2E tests fail on versions of Dapr prior to this change set too.

I do have a theory for `append`. When using `append`, if there is not enough memory allocated, a new underlying array is created. 
`The append built-in function appends elements to the end of a slice. If it has sufficient capacity, the destination is resliced to accommodate the new elements. If it does not, a new underlying array will be allocated. Append returns the updated slice.` - From [docs](https://pkg.go.dev/builtin#append)
That still does not explain why len(s.activeMessages) is zero in the line before and then in for-range loop it contains messages.

mukundansundar (dismissed):
lgtm

internal/component/azure/servicebus/subscription.go
```diff
@@ -361,14 +359,11 @@ func (s *Subscription) RenewLocksBlocking(ctx context.Context, receiver Receiver
...
 				// Snapshot the messages to try to renew locks for.
-				s.mu.RLock()
-				msgs := make([]*azservicebus.ReceivedMessage, 0, len(s.activeMessages))
-				var i int
-				for _, m := range s.activeMessages {
-					msgs[i] = m
-					i++
-				}
-				s.mu.RUnlock()
+				msgs := make([]*azservicebus.ReceivedMessage, 0)
```
DeepanshuA, line 362: Should we initialize it with some initial default capacity? Just thinking so, as with initial capacity of 0, re-allocations of slice will definitely happen.

DeepanshuA:
> I think it'd be safer to move to `sync.Map` as it's not clear to me why we were seeing the panic on the native map and how using `append` is avoiding that. Just to make it super clear, the E2E tests are still failing for me against these changes and on older stable versions of Dapr. It'd be useful if others could confirm whether they see the Azure E2E tests fail on versions of Dapr prior to this change set too.

I agree that we should stick to using sync.Map here. 
Just left a small question/comment.

Also, @jjcollinge when you say that e2e tests are still failing, are these still the ones mentioned in https://github.com/dapr/dapr/pull/5769#issuecomment-1385143072 and what are the errors reported for this failure in App/daprd logs?

internal/component/azure/servicebus/subscription.go
jjcollinge: Yh makes sense to have an initial buffer, does 64 sound reasonable?

jjcollinge:
I'm running the E2E tests against my own Azure infra. I'm not seeing any errors (other than lock renewal for missing messages) but many test cases fail due the assertions themselves. I've tried uninstall my custom Dapr version and installing stable versions via the CLI and rerunning the tests whilst validating that the `pubsub` apps that get deployed are running the stable `daprd` versions and they result in the same test case failures.

internal/component/azure/servicebus/subscription.go
DeepanshuA: This slice is being sent to RenewLockMessage for ASB, I am not very sure that how many number of messages can usually be the minimal kind of value for it. So, yeah 64 sounds fine or may be if you think to lower the value to like 16 or so. @ItalyPaleAle @berndverst suggestions?

ItalyPaleAle: How about rather than guessing, we initialize it with size 0. Then, if we need a bigger slice in the future (e.g. 1), we replace the slice with newly-allocated memory. Every time we encounter a situation where we need more memory, we replace the slice with a bigger one and keep the bigger one.

ItalyPaleAle:
> UPDATE: Interestingly reverting to using append also stops the crash.

I think this may be because the slice was initialized with a length of 0 and a capacity of non-0? Wonder if changing from make(type, 0, capacity) to make(type, length) would work.

I personally do prefer not using sync.Map because doing things with a lock ensure better consistency.

This is from the docs of sync.Map’s Range function:

> Range does not necessarily correspond to any consistent snapshot of the Map's contents: no key will be visited more than once, but if the value for any key is stored or deleted concurrently (including by f), Range may reflect any mapping for that key from any point during the Range call. Range does not block other methods on the receiver; even f itself may call any method on m.

jjcollinge:
Yh thinking about this, this isn't even a concurrency issue, this is just [incorrectly using the slice](https://play.golang.com/p/5LyxH45qLVV). I do typically prefer using native maps and a mutex (hence it was this before) and given the above I'm happy to revert to using those without initializing the map to a length of 0.

ItalyPaleAle (approved):

berndverst:
> I'm running the E2E tests against my own Azure infra. I'm not seeing any errors (other than lock renewal for missing messages) but many test cases fail due the assertions themselves. I've tried uninstall my custom Dapr version and installing stable versions via the CLI and rerunning the tests whilst validating that the `pubsub` apps that get deployed are running the stable `daprd` versions and they result in the same test case failures.

The E2E tests if modified are not necessarily backwards compatible with a previous version of Dapr.

You would need to check out the dapr/dapr 1.9 release branch and run the E2E tests from there if you are using Dapr 1.9.X.

E2E tests in master are only guaranteed to work with the master build of dapr/dapr, unless of course we merged something without running the E2E test.

berndverst (approved):

berndverst:
Will update https://github.com/dapr/dapr/pull/5780 to include this commit now.

PR #2449, opened 2023-01-18
bindings.localstorage fixes

### LINKED ISSUE #2374: Manage Local Storage binding file not found
I started to use the [Local Storage binding](https://docs.dapr.io/reference/components-reference/supported-bindings/localstorage/) for my service application.
During my tests I notice that if I try to retrive a file that doesn't exist yet, I expect to receive a `NotFound` exception or a `BindingResponse` with the `Data` field null, instead I received an `Internal` exception just like in the image.
Would it be possible to fix to receive one of the two proposals?
Also is it possible to add in the yaml component a metadata field, time to live, for the files?

MicrosoftTeams-image

Another bug I found during my tests is that if I get the data inside a file and then delete this, I receive an exception in which is described how my file cannot be delete because it is used by another process:

image

Maybe is it cause because the file isn't close after getting the data information?

### LINKED ISSUE #2444: [Security] `bindings.localstorage`: disallow accessing `/var/run/secrets` by default
## Description

This is a potential security issue.

Using the `bindings.localstorage` binding, users can access any file mounted inside the daprd container. The folder `/var/run/secrets` is where Dapr mount secrets such as the bound token for Sentry, as well as other secrets, and it should not be accessible directly to applications.

This is potentially a breaking change, although it's very unlikely someone will have a legit reason for doing this.

## Release Note

RELEASE NOTE: **FIX** Disallow accessing `/var/run/secrets` using `bindings.localstorage` (security)

ItalyPaleAle:
1. Adds checks and enforcements on rootPath. Also includes ability to disallow certain root paths such as `/var/run/secrets` (also if there are symlinks to those paths)
2. Closes file descriptor that was left open after reading a file
3. If the file doesn't exist when reading it, returns a more specific error message

Fixes #2444
Fixes #2374

### DIFF
--- a/bindings/localstorage/localstorage.go
+++ b/bindings/localstorage/localstorage.go
@@ -210,8 +210,12 @@ func (ls *LocalStorage) get(filename string, req *bindings.InvokeRequest) (*bind
 
 	f, err := os.Open(absPath)
 	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, fmt.Errorf("file not found: %s", absPath)
+		}
 		return nil, fmt.Errorf("error opening path %s: %w", absPath, err)
 	}
+	defer f.Close()
 
 	b, err := io.ReadAll(f)
 	if err != nil {
--- a/bindings/localstorage/localstorage.go
+++ b/bindings/localstorage/localstorage.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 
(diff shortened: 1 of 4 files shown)


berndverst (dismissed):
Please update docs also.

ItalyPaleAle:
@berndverst  I don't think this requires any doc upgrade. Users should not have attempted to access denied folders to begin with, and would have experienced undefined behaviors if rootPath was not absolute

berndverst (approved):

Issue #2445, opened 2023-01-18
Propagation of tracing headers in HTTP Binding component


drewby:
## Describe the feature

Propagation of tracing headers in HTTP Binding component in Dapr to associate traces with the correct parent span. This feature will improve traceability and aid in debugging when calling external HTTP services that provide trace data to a central collector.

There is a workaround, where the client application can add the Traceparent to the metadata in the request payload. However, this means the upstream traces will be parented to the original request to Dapr, not the Dapr request to the endpoint. 

Since a motivation for Dapr is effortless distributed tracing, this capability should be built into the HTTP binding. There may be cases where a customer does not want to propagate trace headers, so perhaps a flag in the binding configuration could be set whether to propagate headers. 

## Release Note

RELEASE NOTE: **ADD** Propagation of tracing headers in HTTP Binding component.

msfussell:
@drewby - Can you update the docs issue for this?

PR #2452, opened 2023-01-20
Add tracing headers to HTTP request

drewby:
# Description

Adds tracing headers (traceparent and tracecontext) to the outbound request in the HTTP Binding. Adding tracing headers is controlled by a traceHeaders option in the binding config. For backward compatibility, the default behavior is 'false' (do not include trace headers). 

## Issue reference

This PR will close: #2445

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#3076


bindings/http/http.go
```diff
@@ -104,6 +107,13 @@ func (h *HTTPSource) Init(metadata bindings.Metadata) error {
 		h.errorIfNot2XX = true
 	}
 
+	if val, ok := metadata.Properties["traceHeaders"]; ok {
+		h.traceHeaders = utils.IsTruthy(val)
+	} else {
+		// Default behavior
+		h.traceHeaders = false
+	}
```
pravinpushkar, line 115: We do not need the else block here. default will be false only IMO.

bindings/http/http.go
```diff
@@ -38,10 +38,12 @@ import (
...
 const (
-	MTLSEnable     = "MTLSEnable"
-	MTLSRootCA     = "MTLSRootCA"
-	MTLSClientCert = "MTLSClientCert"
-	MTLSClientKey  = "MTLSClientKey"
+	MTLSEnable        = "MTLSEnable"
+	MTLSRootCA        = "MTLSRootCA"
+	MTLSClientCert    = "MTLSClientCert"
+	MTLSClientKey     = "MTLSClientKey"
+	traceparentHeader = "traceparent"
```
ItalyPaleAle, line 45: Nit: separate this with a newline

bindings/http/http.go
```diff
@@ -104,6 +107,13 @@ func (h *HTTPSource) Init(metadata bindings.Metadata) error {
 		h.errorIfNot2XX = true
 	}
 
+	if val, ok := metadata.Properties["traceHeaders"]; ok {
```
ItalyPaleAle, line 110: Could be simplified to:

utils.IsTruthy(metadata.Properties["traceHeader"])

No need for the if block at all: if the value doesn't exist, the function returns false by default.

bindings/http/http.go
```diff
@@ -186,6 +196,7 @@ func (h *HTTPSource) Operations() []bindings.OperationKind {
 func (h *HTTPSource) Invoke(ctx context.Context, req *bindings.InvokeRequest) (*bindings.InvokeResponse, error) {
 	u := h.metadata.URL
 
+	traceHeaders := h.traceHeaders
```
ItalyPaleAle, line 199: No need for this, just access h.traceHeader

bindings/http/http.go
```diff
@@ -234,6 +245,16 @@ func (h *HTTPSource) Invoke(ctx context.Context, req *bindings.InvokeRequest) (*
 		request.Header.Set("Accept", "application/json; charset=utf-8")
 	}
 
+	// HTTP binding needs to inject traceparent header for proper tracing stack.
+	if traceHeaders {
+		if tp, ok := req.Metadata[traceparentHeader]; ok {
```
ItalyPaleAle, line 250: Rather than checking if it existsC, here and in the next block please check if the value is not "". Otherwise this could set the header to an empty value, but present.

bindings/http/http.go
ItalyPaleAle: Also please make the key a constant

ItalyPaleAle (changes requested):
Thanks for this :) looks good but some feedback on the code

ItalyPaleAle:
I actually wonder... if we should allow setting arbitrary headers rather than having a specific set? What if people want to set content-type or other headers?

bindings/http/http.go
ItalyPaleAle: On second thought I don't think we need this metadata key. Seems that the headers need to be passed in the metadata of each request so the default behavior is to not include them anyways.

bindings/http/http.go
berndverst, line 45:
```suggestion
	TraceparentHeaderKey = "traceparent"
```

Constants should be named as such and should always be capitalized.

berndverst (changes requested):
Please address feedback from @ItalyPaleAle and @pravinpushkar.

The trace header should only be checked in request metadata. There is no meaning to use this as a component metadata option.

bindings/http/http.go
drewby: Daprd is currently passing the trace headers to Bindings by adding them to metadata, but the HTTP binding is not adding them to the request. These trace headers exist when Dapr tracing is configured and regardless of whether the client calling Dapr includes them in metadata. And they have the advantage of getting the current Span id from Dapr (instead of the span id of the client). 

So the existing behavior will change for some. It seems like a minor change of behavior, but I wasn’t sure. Thoughts?

bindings/http/http.go
drewby: I’ll remove. I was trying to match current code, thought maybe it was convention. :)

drewby:
> I actually wonder... if we should allow setting arbitrary headers rather than having a specific set? What if people want to set content-type or other headers?

HTTP Binding allows arbitrary headings by adding them to Metadata with a Capitalized key. However, the daprd main code adds traceparent and tracestate to metadata without capitalization. So this is very specific to just these two headers getting forwarded.

drewby:
> Please address feedback from @ItalyPaleAle and @pravinpushkar.
> 
> The trace header should only be checked in request metadata. There is no meaning to use this as a component metadata option.

The purpose of the component metadata is to keep existing behavior, which is to not include traceparent/tracestate from the daprd context.  

Based on forums and some experience, there are users who will enable Daprd tracing and not instrument their own code. So nothing will change about request metadata, but they could start sending these headers.  I don’t think that breaks anything, but wasn’t sure so decided to default to an explicit enable of the new behavior.   

If we think sending the headers is the expected behavior, than this could be a bug fix instead of a new feature :)

bindings/http/http.go
drewby: Done

drewby:
@ItalyPaleAle, @berndverst I thought of another case over the weekend, I want to ask your input. 

Today, if a developer wants to send traceheaders to the http endpoint, they will manually add Traceparent to the request metadata. This is a workaround for the HTTP Binding not including the daprd trace context. We discussed this workaround in the discord.

An example, if the client includes a "Traceparent" in the metadata, darpd will add its one "traceparent" and "tracestate" and the incoming metadata to HTTP Binding will look something like this:

```
"path": "/"            // from client
"Traceparent": "1234"   // from client
"traceparent": "5678"   // from daprd
"tracestate": "9012"      // from daprd
```

With the code from this PR, this will result in the outbound HTTP request having headers "Traceparent": "1234" (from the client) and "Tracestate":"9012" (from daprd). Client header wins only because of order of execution. 

What do you think the correct behavior should be? Should we ignore the client trace headers in the metadata?

berndverst:
> @ItalyPaleAle, @berndverst I thought of another case over the weekend, I want to ask your input.
> 
> Today, if a developer wants to send traceheaders to the http endpoint, they will manually add Traceparent to the request metadata. This is a workaround for the HTTP Binding not including the daprd trace context. We discussed this workaround in the discord.
> 
> An example, if the client includes a "Traceparent" in the metadata, darpd will add its own "traceparent" and "tracestate" and the incoming metadata to HTTP Binding will look something like this:
> 
> ```
> "path": "/"            // from client
> "Traceparent": "1234"   // from client
> "traceparent": "5678"   // from daprd
> "tracestate": "9012"      // from daprd
> ```
> 
> With the code from this PR, this will result in the outbound HTTP request having headers "Traceparent": "1234" (from the client) and "Tracestate":"9012" (from daprd). Client header wins only because of order of execution.
> 
> What do you think the correct behavior should be? Should we ignore the client trace headers in the metadata? For proper distributed tracing, the parent should be the daprd traceparent.

Anything a client sets must be allowed as an explicit override - or maybe you should show a warning that these particular properties cannot be overridden. But you should not silently ignore the client's specified trace metadata.

berndverst:
> > Please address feedback from @ItalyPaleAle and @pravinpushkar.
> > The trace header should only be checked in request metadata. There is no meaning to use this as a component metadata option.
> 
> The purpose of the component metadata is to keep existing behavior, which is to not include traceparent/tracestate from the daprd context.
> 
> Based on forums and some experience, there are users who will enable Daprd tracing and not instrument their own code. So nothing will change about request metadata, but the binding might start sending these headers (unexpected by user). I don’t think sending the headers would break anything, but wasn’t sure so decided to default to an explicit enable of the new behavior.
> 
> If we think sending the headers is the expected behavior, than this could be a bug fix instead of a new feature :)

We don't make any guarantees about the headers, so adding additional trace headers is perfectly acceptable and doesn't require opt in.

However, if the user overrides trace headers themselves, you probably should alert the user that they are overriding trace headers if they have tracing enabled in Dapr. I'd use `warn` level logs for this.

bindings/http/http.go
drewby: Removed component metadata based on discussion below.

bindings/http/http.go
```diff
@@ -243,6 +247,22 @@ func (h *HTTPSource) Invoke(ctx context.Context, req *bindings.InvokeRequest) (*
 		}
 	}
 
+	// HTTP binding needs to inject traceparent header for proper tracing stack.
+	if tp, ok := req.Metadata[TraceparentHeaderKey]; ok && tp != "" {
+		if _, ok := request.Header[http.CanonicalHeaderKey(TraceparentHeaderKey)]; ok {
+			h.logger.Warn("tracing enabled, overwriting Traceparent in request headers")
```
berndverst, line 253:
```suggestion
			h.logger.Warn("When tracing is enabled a custom Traceparent request header cannot be specified and is ignored.")
```

bindings/http/http.go
```diff
@@ -243,6 +247,22 @@ func (h *HTTPSource) Invoke(ctx context.Context, req *bindings.InvokeRequest) (*
...
+		}
+
+		request.Header.Set(TraceparentHeaderKey, tp)
+	}
+	if ts, ok := req.Metadata[TracestateHeaderKey]; ok && ts != "" {
+		if _, ok := request.Header[http.CanonicalHeaderKey(TracestateHeaderKey)]; ok {
+			h.logger.Warn("tracing enabled, overwriting Tracestate in request headers")
```
berndverst, line 260:
```suggestion
			h.logger.Warn("Tracing is enabled. A custom Tracestate request header cannot be specified and is ignored.")
```

berndverst (approved):

PR #2453, opened 2023-01-20
Event Hubs: upgrade SDK to "track 2"

### LINKED ISSUE #2101: EventHubs components to use Azure SDK for Go ("Track 2 SDK")

### LINKED ISSUE #2378: The setup-azure-conf-test script contains some deprecated attributes, due to which it fails after running
## Expected Behavior

https://github.com/dapr/components-contrib/tree/master/tests/certification/pubsub/azure/eventhubs 
Referring to the above documentation, to run conformance test using azure portal, we need to run the script "setup-azure-conf-test.sh" in ".github/infrastructure/conformance/azure" and it should generate a bash rc file.

## Actual Behavior

The above script fails stating - "Could not find user with upn shivasingh@microsoft.com"

Login to az cli
Go to .github/infrastructure/conformance/azure and run setup-azure-conf-test.sh script

## Fix 
The script uses "objectId" attribute, which is deprecated in "az ad user list", it needs to be changed to "id"
https://learn.microsoft.com/nb-no/cli/azure/microsoft-graph-migration

ItalyPaleAle:
# Description

Updated the Azure Event Hubs PubSub and Binding components to the new "track 2" SDK.

This should come with no functional changes to end users (and no changes in metadata), but there are some significant performance and stability improvements thanks to using the new SDK (as well as some fixes in our component's code, including - _lesigh_ - race conditions).

EDIT:
- For the binding, there's a small change in that the `partitionID` metadata property has been deprecated in favor of listening on all partitions

## ⚠️ Important note

Azure Event Hub maintains "checkpoints" in Azure Blob Storage, and that's something that's managed by the SDK. The new SDK uses a different way of storing checkpoints that is **not compatible** with the old SDK.

This can be a problem during the rollout of a new version of Dapr: if the app is scaled horizontally, on an environment like K8s the rollout updates one instance of Dapr at a time. This means that for a brief time there can be apps using Dapr 1.9 (old SDK) and Dapr 1.10 (new SDK), and that will cause issues.

The way we're addressing that is a bit "hacky" but should do the job:

- Before starting a new subscription, we check if there's a file in Azure Storage with the checkpoint created by the old SDK (Dapr 1.9). If the file exists AND has an active lock, it means that there's another app running concurrently that's using the old SDK.
- We wait for up to 2 minutes to see if the lock is released (locks are requested for 60s so if they exist after 2 minutes, it means that the other app is still running).
- If after 2 minutes the lock is still active, we return an error starting the subscription to have Dapr complete the initialization anyways and report a healthy status to K8s. However, the subscription would not be active.
- At the same time, we also schedule a background timer that forces Dapr to crash after 2 more minutes. 2 minutes should be enough time for K8s to have completed the rollout of more instances of the app, and seeing Dapr panicking would cause K8s to restart the pod, and thus restarting the subscription. Hopefully by now all old apps have been updated so we can subscribe successfully.

This is not ideal but seems better than the only alternative @berndverst and I discussed of creating a "v2" of the component, which would require more manual work for users AND would increase the burden on us to support 2 versions of the same component (including one using deprecated SDKs).

This behavior will be temporary and will be removed in Dapr 1.13.

Note: given the temporary nature of this fix, and the complexity of creating a test running 2 versions of Dapr (we do not have the infra to do that at the moment), I will not be able to create an automated test for this. I have validated it manually.

## Issue reference

Fixes #2101 
Fixes #2378 incidentally


.github/infrastructure/conformance/azure/setup-azure-conf-test.sh
```diff
@@ -228,7 +228,7 @@ STORAGE_CONTAINER_VAR_NAME="AzureBlobStorageContainer"
 STORAGE_QUEUE_VAR_NAME="AzureBlobStorageQueue"
 
 # Derived variables
-ADMIN_ID="$(az ad user list --filter "userPrincipalName eq '${ADMIN_UPN}'" --query "[].objectId" --output tsv)"
```
ItalyPaleAle, line 231: Reviewers: the script was not working with the updated Azure SDK

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -15,709 +15,618 @@ package eventhubs
...
-	backOffConfig.Policy = retry.PolicyExponential
-	backOffConfig.MaxInterval = resourceCheckMaxRetryInterval
-	backOffConfig.MaxRetries = resourceCheckMaxRetry
+// BulkPublish sends data to Azure Event Hubs in bulk.
+func (aeh *AzureEventHubs) BulkPublish(ctx context.Context, req *pubsub.BulkPublishRequest) (pubsub.BulkPublishResponse, error) {
+	res := pubsub.BulkPublishResponse{}
 
-	b := backOffConfig.NewBackOffWithContext(parentCtx)
+	if req.Topic == "" {
+		return res, errors.New("parameter 'topic' is required")
```
mukundansundar, line 151: do we need to send a response of all failures so that it is retried again by dapr?

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -15,709 +15,618 @@ package eventhubs
...
 
-func (aeh *AzureEventHubs) ensurePublisherClient(ctx context.Context, hubName string) error {
-	if aeh.metadata.EnableEntityManagement {
-		if err := aeh.ensureEventHub(ctx, hubName); err != nil {
-			return err
-		}
+	// Build the batch of messages
+	batch, err := client.NewEventDataBatch(ctx, batchOpts)
+	if err != nil {
+		return fmt.Errorf("error creating batch: %w", err)
```
mukundansundar, line 202:
```suggestion
		return fmt.Errorf("error creating event batch: %w", err)
```

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -15,709 +15,618 @@ package eventhubs
...
+		// TODO: Support setting a batch size
+		const batchSize = 1
+		ctx, cancel = context.WithCancel(subscribeCtx)
+		events, err = partitionClient.ReceiveEvents(ctx, batchSize, nil)
+		cancel()
 
-	aeh.metadata = m
-	aeh.eventProcessors = map[string]*eph.EventProcessorHost{}
-	aeh.hubClients = map[string]*eventhub.Hub{}
+		// A DeadlineExceeded error means that the context timed out before we received the full batch of messages, and that's fine
```
mukundansundar, line 364: where is the timeout setting for this particular context `cix`? we are only initialising it with a cancel right?

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -15,709 +15,618 @@ package eventhubs
...
+		if err != nil && !errors.Is(err, context.DeadlineExceeded) {
+			// If we get an error like ErrorCodeOwnershipLost, it means that the partition was rebalanced and we lost it
+			// We'll just stop this subscription and return
+			eventHubError := (*azeventhubs.Error)(nil)
+			if errors.As(err, &eventHubError) && eventHubError.Code == azeventhubs.ErrorCodeOwnershipLost {
+				aeh.logger.Debug("Client lost ownership of partition %s for topic %s", partitionClient.PartitionID(), topic)
```
mukundansundar, line 370: Debugf?

.github/infrastructure/conformance/azure/setup-azure-conf-test.sh
berndverst: I can confirm this is necessary.

pubsub/azure/eventhubs/eventhubs.go
ItalyPaleAle: Yes you're right. Let me fix that

pubsub/azure/eventhubs/eventhubs.go
ItalyPaleAle: Good catch

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -620,6 +630,89 @@ func (aeh *AzureEventHubs) createCheckpointStore() (checkpointStore azeventhubs.
...
+		// Authenticate with a shared key
+		credential, newSharedKeyErr := azblob.NewSharedKeyCredential(aeh.metadata.StorageAccountName, aeh.metadata.StorageAccountKey)
+		if newSharedKeyErr != nil {
+			return nil, fmt.Errorf("invalid Azure Storage shared key credentials with error: %w", newSharedKeyErr)
+		}
+		client, err = azblob.NewClientWithSharedKeyCredential(accountURL, credential, &options)
+		if err != nil {
+			return nil, fmt.Errorf("error creating Azure Storage client from shared key credentials: %w", err)
+		}
+	} else {
```
mukundansundar, line 697: if else if else can be simplified, return after getting the client immediately.

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -620,6 +630,89 @@ func (aeh *AzureEventHubs) createCheckpointStore() (checkpointStore azeventhubs.
...
+		},
+	}
+
+	var (
+		err    error
+		client *azblob.Client
+	)
+	// Use the global URL for Azure Storage
+	accountURL := fmt.Sprintf("https://%s.blob.%s", aeh.metadata.StorageAccountName, "core.windows.net")
+
```
mukundansundar, line 680: can move this after connection string check

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -620,6 +630,89 @@ func (aeh *AzureEventHubs) createCheckpointStore() (checkpointStore azeventhubs.
...
+	defer cancel()
+	_, err = client.CreateContainer(ctx, aeh.metadata.StorageContainerName, &container.CreateOptions{
+		// Default is private
+		Access: nil,
+	})
+	if err != nil {
+		// Check if it's an Azure Storage error
+		resErr := &azcore.ResponseError{}
+		// If the container already exists, return no error
+		if errors.As(err, &resErr) && (resErr.ErrorCode == "ContainerAlreadyExists" || resErr.ErrorCode == "ResourceAlreadyExists") {
```
mukundansundar, line 654: create constants. 
Can we use any lib constants for this? As if the string changes, then this path is affected.

pubsub/azure/eventhubs/eventhubs.go
```diff
@@ -620,6 +630,89 @@ func (aeh *AzureEventHubs) createCheckpointStore() (checkpointStore azeventhubs.
...
+		}
+		client, err = azblob.NewClientWithSharedKeyCredential(accountURL, credential, &options)
+		if err != nil {
+			return nil, fmt.Errorf("error creating Azure Storage client from shared key credentials: %w", err)
+		}
+	} else {
+		// Use Azure AD
+		settings, err := azauth.NewEnvironmentSettings("storage", aeh.metadata.properties)
+		if err != nil {
+			return nil, err
```
mukundansundar, line 701: does this err need to be wrapped?

tests/certification/pubsub/azure/eventhubs/eventhubs_test.go
```diff
@@ -131,16 +131,16 @@ func TestEventhubs(t *testing.T) {
 			client := sidecar.GetClient(ctx, sidecarName)
 
 			// publish messages
-			ctx.Logf("Publishing messages. sidecarName: %s, topicName: %s", sidecarName, topicName)
+			logs := fmt.Sprintf("Published messages. sidecarName: %s, topicName: %s", sidecarName, topicName)
```
mukundansundar, line 134: This is still either present tense or future, cannot be past as the op is not complete yet.

pubsub/azure/eventhubs/eventhubs.go
ItalyPaleAle: I looked but sadly the constants are in the library but are not exported, they're in an internal package only (no idea why).

However these constants are defined in the Azure REST API specs so they will not change.

tests/certification/pubsub/azure/eventhubs/eventhubs_test.go
ItalyPaleAle: It's past as the log is printed after the messages have ben published only!

pubsub/azure/eventhubs/eventhubs.go
ItalyPaleAle: Fixed

ItalyPaleAle:
I've updated the PR and now changes the binding too, although it's not fully working yet - I'll debug more on Monday.

In the meanwhile, I've also identified the issue with IoT Hub that wasn't working, and it was because we had to enable the "fallback route". I've updated the Bicep to include that, and manually applied the change to the infra using the Azure Portal.

berndverst (approved):

tomkerkhove:
Out of curiosity, how is this not a breaking change?

This change most probably broke KEDA's integration.

berndverst:
Its not a breaking change and we made sure of it. We ensured to manually map existing properties to the new SDK for backwards compatibility. All existing auth methods continue to work in the exact same manner and we have E2E / integration tests and lots of customer evidence to prove it.

ItalyPaleAle:
@tomkerkhove To be honest I wasn't aware of the KEDA integration until yesterday :( Our team was never involved with that.

I would also argue that accessing the data in the checkpoint store without using Dapr is not an officially-supported scenario, so it is not covered by our backwards-compatibility promise.

That said, how can we help ensuring that KEDA integration works?

berndverst:
What even is the KEDA integration? Clearly as a maintainer I am completely unaware. @tomkerkhove 

Can you contribute a test or something?

tomkerkhove:
Sorry for the 0 context - I was referring to KEDA using the Dapr checkpointing pattern for https://keda.sh/docs/2.9/scalers/azure-event-hub/

PR #2462, opened 2023-01-24
Support PEM certificates from Azure AD

### LINKED ISSUE #2396: Azure Auth: failed to decode pkcs12 certificate while creating spt: pkcs12
## Expected Behavior
Creds file should work

## Actual Behavior
Creds file fails with:

```
process component localsecrets error: [INIT_COMPONENT_FAILURE]: initialization error occurred for send-email (bindings.azure.storagequeues/v1): invalid credentials with error: failed to decode pkcs12 certificate while creating spt: pkcs12: error reading P12 data: asn1: structure error: tags don't match (16 vs {class:0 tag:13 length:45 isCompound:true}) {optional:false explicit:false application:false private:false defaultValue:<nil> tag:<nil> stringType:0 timeType:0 set:false omitEmpty:false} pfxPdu @2" app_id=c2c-exchange-actors instance=COMP1 scope=dapr.runtime type=log ver=1.9.5
```

Seems to be an issue referenced here: https://github.com/Azure/azure-sdk-for-go/issues/5412

Create a credential with certificate, following this guide: https://docs.dapr.io/developing-applications/integrations/azure/authenticating-azure/#creating-an-azure-ad-application
```
az ad sp create-for-rbac --name ... --role contributor ...  --create-cert 
```

Copy the file from `fileWithCertAndPrivateKey` to home directory, reference it in component.yml

```yml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: send-email
spec:
  type: bindings.azure.storagequeues
  version: v1
  metadata:
  - name: accountName
    secretKeyRef:
      name: storageAccount
  - name: queueName
    value: send-email
  - name: azureTenantId
    secretKeyRef:
      name: azure__tenantId
  - name: azureClientId
    secretKeyRef:
      name: azure__clientId
  - name: azureCertificateFile
    value: "C:\\Users\\myuser\\.dapr\\certs\\myfile.pem"
```

Run application with dapr

## Release Note

RELEASE NOTE:

ItalyPaleAle:
The Azure CLI is now issuing client certificates in an "unusual" format: a single file containing 2 PEM-encoded DER blocks with both the certificate and key. This updates our Azure AD auth layer to support that format too, in addition to the "usual" PKCS#12/PFX.

Fixes #2396

### DIFF
--- a/internal/authentication/azure/auth.go
+++ b/internal/authentication/azure/auth.go
@@ -16,6 +16,7 @@ package azure
 import (
 	"crypto/rsa"
 	"crypto/x509"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"os"
@@ -313,7 +314,7 @@ func (c CertConfig) ServicePrincipalTokenByCertBytes() (*adal.ServicePrincipalTo
 		return nil, err
 	}
 
-	certificate, rsaPrivateKey, err := c.decodePkcs12(c.CertificateData, c.CertificatePassword)
+	certificate, rsaPrivateKey, err := c.decodeCertificate(c.CertificateData, c.CertificatePassword)
 	if err != nil {
 		return nil, fmt.Errorf("failed to decode pkcs12 certificate while creating spt: %v", err)
 	}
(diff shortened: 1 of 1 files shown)


berndverst (approved):
LGTM - that's how I would have done that too

ItalyPaleAle:
> that's how I would have done that too

Great minds think alike? 😉

Issue #2313, opened 2022-11-25
MQTT PubSub Component To Support ONLY QoS 1


DeepanshuA:
### Dapr to Remove Support of QoS 0 and 2 from MQTT PubSub

**Dapr works on the concept of "at least once"**

Now, let's try to map it with MQTT qos concept and see if Dapr can support all QoS 0, 1 or 2.

**QoS 0: At MOST once**

Once App receives a message, it may assume that it doesn't need to respond back with any ACK, as it is not required for qos 0.
But, dapr tries to simulate broker for retries (Why? What I understand, it does so to keep retries per secod in check, so that throttling limit is not reached). Hence, when dapr will not receive an ACK, it will get it replayed somehow. Is it fine to get it replayed somehow? Yes, it is, as Dapr promises at once delivery. Thus, Dapr should NOT support QoS 0.
It also falls in sync with https://github.com/dapr/components-contrib/issues/1808.

**QoS 1: At least once**
In sync with dapr contract.

**QoS 2: Exactly once.**
Again, dapr works with simulating broker at few places, like retry logic and all.
For this to happen, it is a possibility that in some race conditions , message is delivered MORE than once. Thus, the contrcat pf QoS can not be fulfilled. Thus, dapr should not support QoS 0 or 2.
And, this should be explicit in docs as well.

Ref: https://mntolia.com/mqtt-qos-levels-explained/

_But, we already have a parameter `qos` in our MQTT spec, which user can mention._
We can add a warning for 1.10, where-in when dapr is coming up with with this component and with qos specified, we will give a warning that it will be deprecated in next release and right now, it will be defaulted to 1.
Then, in next release, we can remove qos flag.
**Why to remove?**
We can keep this option, but it will confuse users giving impression that different qos is supported.

ItalyPaleAle:
I think removing support for QoS = 0 makes a lot of sense (it actually used to be the default until Dapr 1.7 I believe!), and making QoS = 1 the new default.

Not sure if removing 2 is needed however. We guarantee “at least once” delivery, so if the message were to be delivered twice in some situations, it may not necessarily be bad?

shivamkm07:
> Not sure if removing 2 is needed however. We guarantee “at least once” delivery, so if the message were to be delivered twice in some situations, it may not necessarily be bad?

But if message is redelivered, that would break the semantics of QOS 2. Then essentially qos 2 would be same in behavior as qos 1, which wouldn't be what user would be expecting when using qos 2.

[...]

PR #2467, opened 2023-01-25
Refactored MQTT to better handle subscribptions and fix bugs

### LINKED ISSUE #2229: MQTT Missing Resubscribe on Reconnect

ItalyPaleAle:
This is my 2nd attempt at refactoring MQTT (pubsub and binding) to fix bugs such as #2229 and other issues.

Pretty much refactors how the components manage the connection to the broker and rely more on the upstream SDK for managing connections and re-connecting.

- Both: Fixes re-subscribing after a connection is dropped
- Pubsub: Uses a single connection for consumer and publisher
- Pubsub: Avoids reconnecting when the list of subscriptions changes
- Binding: Create producer connection only on output binding, otherwise the component would create 2 connections with the same consumer group and one would enter in a failed state
- Binding: 2 metadata properties have changed:
  - "qos" has been deprecated and is now fixed to "1" (see #2313)
  - "cleanSession" now defaults to "false"
  - (Note that docs have an issue for the binding because "consumerID" is required in the binding, but not in pubsub)

Fixes #2229
Fixes #2350

Docs issue: https://github.com/dapr/docs/issues/3072


ItalyPaleAle:
CC: @DeepanshuA

berndverst (approved):
Thanks for this PR - and good collaboration on it (off-GitHub ☺️)

Issue #2468, opened 2023-01-25
Add options for token authentication and automatic JWT tokens for HTTP binding


yaron2:
Support for the following:

```yaml
kind: Component
metadata:
  name: mybinding
spec:
  type: bindings.http
  version: v1
  metadata:
    - name: url
      value: http://something.com/
# Fixed security token
    - name: securityToken
      secretKeyRef:
        name: securitytoken
    - name: securityHeader
      value: x-token
# JWT header
    - name: jwtIssuer
      value: http://ibm.jwtsissuer.com/v1/issue
    - name: jwtIssuerToken
      secretKeyRef:
        name: issuertoken
    - name: jwtIssuerHeader
      value: x-token
```

PR #2470, opened 2023-01-25
Add security token header support

yaron2:
Partially closes https://github.com/dapr/components-contrib/issues/2468.

This functionality is independent and can work standalone from the JWT token part described on the issue.


berndverst (approved):

ItalyPaleAle:
@yaron2 @berndverst don't you think we should look into a way to pass arbitrary headers rather than adding each header one-by-one?

Made sense (perhaps) for trace headers, but seems that this could have been better as a generic one?

(I also see that this requires passing both the header name and value as metadata properties, so there's a very high chance it will be abused by users to pass any kind of header)

berndverst:
That's exactly what @yaron2 and I discussed actually in Discord. But the importance right here is that you should be able to get the security token from a SecretStore reference. This is why I am fine with this feature as is.

PR #2472, opened 2023-01-26
Dapr 1.11: Support Azure CLI Credential for Azure Track 2 components

### LINKED ISSUE #2469: Support azure-cli credentials for local development using azure-based components
## Describe the feature
It would be a huge time saver for developers to be able to use azure-cli based credentials to authenticate azure-based dapr components; this would allow developers to having to create service principles and managing secrets or certificates just to run their components locally.

The analog here for .NET developers would be use of the `DefaultAzureCredential`, which automatically checks several sources on your machine for azure credentials (including VS Code, CLI, token cache, and managed identity endpoints).

## Release Note
**ADD** Support for Azure-CLI credentials with azure-based dapr components

- Depends on #1103

berndverst:

# Description

This adds an additional authentication method after MSI attempted: Azure CLI credential

This is particularly helpful for using Dapr locally in development.

Fixes #2469

Because the ChainedTokenCredential attempts auth methods sequentially we do not obtain the ManagedIdentityCredential and CLICredential this way -- instead we run these in parallel in different go routines. Whichever method finishes first is used.

NOTE
This PR needs to be revert if Azure SDK for GO / azidentity 1.3.0 is not released and used by Dapr 1.11 release time. This PR uses 1.3.0-Beta.3 and needs to be updated.

### DIFF
--- a/internal/authentication/azure/auth.go
+++ b/internal/authentication/azure/auth.go
@@ -14,6 +14,7 @@ limitations under the License.
 package azure
 
 import (
+	"context"
 	"crypto/rsa"
 	"crypto/x509"
 	"encoding/pem"
@@ -21,9 +22,11 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"golang.org/x/crypto/pkcs12"
 
@@ -34,6 +37,36 @@ import (
 type EnvironmentSettings struct {
 	Metadata map[string]string
 	Cloud    *cloud.Configuration
+}
+
+// timeoutWrapper prevents a potentially very long timeout when managed identity or CLI credential aren't available
+type timeoutWrapper struct {
+	cred azcore.TokenCredential
+	// timeout applies to all auth attempts until one doesn't time out
+	timeout    time.Duration
+	authmethod string
+}
+
+// GetToken wraps Token acquisition attempts with a short timeout because managed identity or CLI credential
+// may not be available and connecting to IMDS can take several minutes to time out.
+func (w *timeoutWrapper) GetToken(ctx context.Context, opts policy.TokenRequestOptions) (azcore.AccessToken, error) {
+	var tk azcore.AccessToken
+	var err error
+	// no need to synchronize around this value because it's written only within ChainedTokenCredential's critical section
+	if w.timeout > 0 {
+		c, cancel := context.WithTimeout(ctx, w.timeout)
+		defer cancel()
+		tk, err = w.cred.GetToken(c, opts)
+		if ce := c.Err(); errors.Is(ce, context.DeadlineExceeded) {
+			err = azidentity.NewCredentialUnavailableError(w.authmethod)
+		} else {
+			// some managed identity implementation is available, so don't apply the timeout to future calls
+			w.timeout = 0
+		}
+	} else {
+		tk, err = w.cred.GetToken(ctx, opts)
+	}
+	return tk, err
 }
 
 // NewEnvironmentSettings returns a new EnvironmentSettings configured for a given Azure resource.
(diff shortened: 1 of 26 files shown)


ItalyPaleAle:
@berndverst I think we can merge this in master since we're working on 1.11 now. I saw there's a new beta 3: https://github.com/Azure/azure-sdk-for-go/releases/tag/sdk%2Fazidentity%2Fv1.3.0-beta.3  WDYT

internal/authentication/azure/auth.go
```diff
@@ -103,7 +136,9 @@ func (s EnvironmentSettings) GetAzureEnvironment() (*azure.Environment, error) {
 // GetTokenCredential returns an azcore.TokenCredential retrieved from, in order:
 // 1. Client credentials
 // 2. Client certificate
-// 3. MSI
+// 3. MSI with 1 second timeout
+// 4. CLI credential with 5 second timeout
```
ItalyPaleAle, line 140: What if we moved CLI credentials before MSI? THis way you wouldn't need to try MSI twice with different timeouts. I think it's acceptable to have CLI have priority over MSI.

onionhammer: It would not be ideal to delay startup in production because you're waiting for an azure CLI timeout, does these fail quickly or do they do HTTP requests that eventually timeout when an MSI endpoint isnt available?

ItalyPaleAle: I'm actually not even sure why the CLI's credential check needs a context. All methods in the Azure SDKs that use a context are making network calls, but the CLI credentials are stored on a local file (unless maybe the method is making a call to exchange a refresh token?)

Another option could be to trigger both these checks in parallel (each on its own goroutine) and then the first one that returns (successfully) wins. Kind of like Promise.race in JavaScript!

berndverst: I just followed what the DefaultAzureCredential essentially does. All this code is pretty much copied from the `azidentity` package.

berndverst: @onionhammer @ItalyPaleAle 

The `DefaultAzureCredential` tries Managed Identity with a time out of 1 (one) second before trying the CLI Credential. This will usually be no problem. However, in the case this does time out I added an indefinite attempt to connect after the Azure CLI Credential. I think this is a good compromise that also matches what the DefaultAzureCredential does - but with the additional Managed Identity Retry without the 1 second timeout.

ItalyPaleAle: Oh, if this is what the Azure team is doing too, then I have no more concerns.

ItalyPaleAle (dismissed):

internal/authentication/azure/auth.go
```diff
@@ -130,40 +97,50 @@ func (s EnvironmentSettings) GetTokenCredential() (azcore.TokenCredential, error
...
 		cred, credErr := azidentity.NewAzureCLICredential(nil)
 		if credErr == nil {
-			creds = append(creds, &timeoutWrapper{cred: cred, authmethod: "Azure CLI", timeout: 5 * time.Second})
+			ch <- cred
 		} else {
 			errs = append(errs, credErr)
 		}
+	}(&wg, ch)
+
+	go func() {
```
ItalyPaleAle, line 133: You can put `wg.Wait()` outside of a background goroutine, so then you don't need the `close(ch)`

internal/authentication/azure/auth.go
```diff
@@ -130,40 +97,50 @@ func (s EnvironmentSettings) GetTokenCredential() (azcore.TokenCredential, error
...
+	go func(wg *sync.WaitGroup, ch chan<- azcore.TokenCredential) {
+		defer wg.Done()
+		// sleep for 1 second to allow MSI to win before trying Azure CLI
+		time.Sleep(1 * time.Second)
 		cred, credErr := azidentity.NewAzureCLICredential(nil)
 		if credErr == nil {
-			creds = append(creds, &timeoutWrapper{cred: cred, authmethod: "Azure CLI", timeout: 5 * time.Second})
+			ch <- cred
 		} else {
 			errs = append(errs, credErr)
```
ItalyPaleAle, line 129: This is not concurrency-safe when running in a goroutine. If both goroutines try to append an error at the same time, it will panic.

Instead, consider having a separate channel that is used to send errors

internal/authentication/azure/auth.go
```diff
@@ -130,40 +97,50 @@ func (s EnvironmentSettings) GetTokenCredential() (azcore.TokenCredential, error
...
-		msiCred, err := c.GetTokenCredential()
+		cred, err := c.GetTokenCredential()
 		if err == nil {
-			creds = append(creds, &timeoutWrapper{cred: msiCred, authmethod: "managed identity", timeout: 1 * time.Second})
+			ch <- cred
 		} else {
 			errs = append(errs, err)
 		}
-	}
+	}(&wg, ch)
```
ItalyPaleAle, line 119: No need to pass `wg` as an argument to the function, can just refer to the one in the parent scope

ItalyPaleAle (changes requested):

berndverst:
@ItalyPaleAle after much investigation doing things concurrently isn't possible.

Every Track 2 SDK takes the `acore.TokenCredential` and calls `GetToken()` which takes some options, including a required scope. The scope isn't known in any of our auth code.

It is only the `GetToken()` call which actually verifies whether a given auth method like CLI and MSI will work.

For this reason I am reverting back to the original version you had approved using the sequential ChainedTokenCredential matching the DefaultAzureCredential implementation with the additional fallback MSI option.

ItalyPaleAle (approved):

PR #2487, opened 2023-01-31
[release-1.10] Fix "Azure Service Bus does not recover from failures when publishing"

### LINKED ISSUE #2483: [Azure Service Bus] Components do not recover from failures when publishing
## Description

Issue reported on Discord: https://discord.com/channels/778680217417809931/1067897796725841930

Seems that after publishing fails on Azure Service Bus with a connection error, the connection enters in a "bad" state and Dapr isn't able to recover from it.

We have some code in place that forces a reconnection if the subscribing connection gets into a "bad" state. However we have nothing in place that causes the connection to be recreated if the publishing side continues to fail.

## Release Note

RELEASE NOTE: **FIX** Force reconnecting to Azure Service Bus after a persistent publish failure

ItalyPaleAle:
# Description

This PR is an attempt at fixing #2483 , although I am not able to repro the issue so I can't validate this. The approach is two-fold:

1. We are already treating "context deadline exceeded" errors as errors that would require disconnecting and reconnecting. However, it seems that the check `errors.Is(err, context.DeadlineExceeded)` may not be detecting those errors, perhaps because an upstream library is wrapping the errors some way. Based on looking into how the Service Bus SDK works, I've seen a sign that something like that may be happening. So, I've changed the code to be more lax and also accept errors whose stringified message is "context deadline exceeded" (but which are not  context.DeadlineExceeded) as errors that would require reconnecting.
2. In parallel, I've added more logs that, should the error persist, would better help us understanding what is going on, and if the connection is indeed being re-created.

## Issue reference

(Hopefully) fixes #2483

### DIFF
--- a/bindings/azure/servicebusqueues/servicebusqueues.go
+++ b/bindings/azure/servicebusqueues/servicebusqueues.go
@@ -92,7 +92,7 @@ func (a *AzureServiceBusQueues) Invoke(invokeCtx context.Context, req *bindings.
 	if err != nil {
 		if impl.IsNetworkError(err) {
 			// Force reconnection on next call
-			a.client.CloseSender(a.metadata.QueueName)
+			a.client.CloseSender(a.metadata.QueueName, a.logger)
 		}
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


berndverst (approved):

MattCosturos:
I see this is (hopefully) fixed in the 1.10 series.
We are running into this issue with 1.9.5
Until we update is there any information we could provide that would be of assistance in debugging? I noticed the comment that you can't repo the issue, same here, we can't repo the issue on demand, but after some time the calls to publish fail. Seems to take days till failure

PR #2499, opened 2023-02-02
Add `Close() error` function io input binding interface.

JoshVanL:
Branched from https://github.com/dapr/components-contrib/pull/2474 which should be merged first.

Part of https://github.com/dapr/dapr/issues/2478

PR adds a `Close() error` function to the `InputBinding` interface. Close will release all resources associated with the input binding, and blocks until all go routines have returned.

`Close` is intended to always be called by the caller.

### DIFF
--- a/bindings/input_binding.go
+++ b/bindings/input_binding.go
@@ -23,7 +23,7 @@ import (
 // InputBinding is the interface to define a binding that triggers on incoming events.
 type InputBinding interface {
 	// Init passes connection and properties metadata to the binding implementation.
-	Init(metadata Metadata) error
+	Init(ctx context.Context, metadata Metadata) error
 	// Read is a method that runs in background and triggers the callback function whenever an event arrives.
 	Read(ctx context.Context, handler Handler) error
 }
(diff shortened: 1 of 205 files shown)


bindings/input_binding.go
```diff
@@ -26,6 +26,9 @@ type InputBinding interface {
 	Init(ctx context.Context, metadata Metadata) error
 	// Read is a method that runs in background and triggers the callback function whenever an event arrives.
 	Read(ctx context.Context, handler Handler) error
+	// Close is a method that closes the connection to the binding. Must be
+	// called when the binding is no longer needed to free up resources.
+	Close() error
```
ItalyPaleAle, line 31: Rather than making this a custom method, import `io.Closer` in the interface...

```go
type InputBinding interface {
    io.Closer
//...
```

ItalyPaleAle (changes requested):
Since this PR is tied to #2474 can you do a single PR?

It's just that updating an interface is a schlep... and best to do it all at once :) (I'll ping you on Discord to explain the next steps once #2474 is reviewed and approved)

bindings/input_binding.go
JoshVanL: Sounds good to me :+1:

JoshVanL:
/close

Closing in favor of moving to a single PR [here](https://github.com/dapr/components-contrib/pull/2474).

PR #2507, opened 2023-02-04
SQLite state store: certification tests + various updates and fixes

### LINKED ISSUE #2485: SQLite state store and concurrent access by multiple Dapr instances
## Description

SQLite does not support concurrent writes to the database, and trying to execute multiple queries that modify the database can corrupt the DB.

Within a single Dapr instance, the component is using Go locks to ensure that no more than one operation on the database happens at the same time. This works really well to prevent issues when there's only one Dapr sidecar accessing the database.

However, if multiple Dapr sidecars are initialized with the same SQLite database, there's no protection against concurrent writes.

This could be implemented by adding a transaction around each SQL query. We could also add a metadata option to make this optional, if users developing locally don't need concurrent access to the same DB.

## Release Note

RELEASE NOTE: **ADD** Protections against concurrent access to the same SQLite database in `state.sqlite`

ItalyPaleAle:
This PR contains certification tests for the SQLite state store as well as various updates and fixes.

Among the fixes, it ensures the component works well even in a scenario where multiple Dapr processes are using it, and this is certified in the cert tests. Fixes #2485

Things that will need to be changed in the docs (I can't do that yet because the docs for 1.10 hasn't been merged yet :) ):

- New or changed metadata properties:
   - `metadataTableName` name of the metadata table (default: "metadata")
   - `cleanupInterval` replaces the old `cleanupIntervalInSeconds` and is parsed as a Go time.Duration. This seems better for handling long durations such as "1d" (`cleanupIntervalInSeconds` remains available for backwards-compatibility)
   - `cleanupInterval` now defaults to `0` which means that it's disabled by default. Unlike in Postgres, in SQLite the entire database gets locked during cleanups, so seems appropriate to not run this unless the user wants it
   - `busyTimeout` duration (Go time.Duration) that makes SQLite wait in case another thread/process is accessing the same database, before returning a "database busy" error (default: `2s`)
   - `disableWAL` if set to true, disables Write-Ahead Logging for journaling. This should always be enabled (the default - value is false) unless the database is stored on a network file system. This has no impact on read-only or in-memory databases.
- Misc: now the component automatically adds `?cache=shared` for in-memory databases, so users don't need to do it themselves anymore 

PS: a couple of small fixes were made to the Postgres component too, since I was looking at that code for reference, but nothing major

Docs issue: https://github.com/dapr/docs/issues/3138


state/sqlite/sqlite_metadata.go
```diff
@@ -79,23 +83,59 @@ func (m *sqliteMetadataStruct) InitWithMetadata(meta state.Metadata) error {
 
 	// Cleanup interval
 	// Nil duration means never clean up expired data
-	if m.CleanupIntervalInSeconds != "" {
+	if m.CleanupIntervalStr != "" {
```
berndverst, line 86: Unnecessary - our decode function (metadata.DecodeMetadata) already handles durations just fine.

state/sqlite/sqlite_metadata.go
```diff
@@ -21,34 +21,38 @@ import (
...
 
 type sqliteMetadataStruct struct {
-	ConnectionString         string `json:"connectionString" mapstructure:"connectionString"`
-	TableName                string `json:"tableName" mapstructure:"tableName"`
-	TimeoutInSeconds         string `json:"timeoutInSeconds" mapstructure:"timeoutInSeconds"`
+	ConnectionString   string `json:"connectionString" mapstructure:"connectionString"`
+	TableName          string `json:"tableName" mapstructure:"tableName"`
+	MetadataTableName  string `json:"metadataTableName" mapstructure:"metadataTableName"`
+	TimeoutInSeconds   string `json:"timeoutInSeconds" mapstructure:"timeoutInSeconds"`
+	CleanupIntervalStr string `json:"cleanupInterval" mapstructure:"cleanupInterval"` // Cleanup interval as a time.Duration string
```
berndverst, line 42: Just make this `time.Duration` or if preferred `*time.Duration` directly -- our `metadata.DecodeMetadata` function supports this out of the box.

state/sqlite/sqlite_metadata.go
```diff
@@ -21,34 +21,38 @@ import (
...
+	BusyTimeoutStr     string `json:"busyTimeout" mapstructure:"busyTimeout"`         // Busy timeout as a time.Duration string
```
berndverst, line 43: Ditto.

state/sqlite/sqlite_metadata.go
```diff
@@ -79,23 +83,59 @@ func (m *sqliteMetadataStruct) InitWithMetadata(meta state.Metadata) error {
...
-		} else {
-			m.cleanupInterval = nil
+			m.cleanupInterval = time.Duration(cleanupIntervalInSec) * time.Second
+		}
+	}
+
+	// Busy timeout
+	// Values must be in milliseconds. Values <= 0 do not set any timeout
+	if m.BusyTimeoutStr != "" {
+		parsed, err := time.ParseDuration(m.BusyTimeoutStr)
```
berndverst, line 111: ditto

berndverst (changes requested):

state/sqlite/sqlite_metadata.go
ItalyPaleAle: Done

state/sqlite/sqlite_metadata.go
ItalyPaleAle: Done

state/sqlite/sqlite_metadata.go
ItalyPaleAle: Done

state/sqlite/sqlite_metadata.go
ItalyPaleAle: Done

berndverst (approved):

Issue #2213, opened 2022-10-21
Pass kafka record key and other metadata to pubsub subscribers


ctoomey:
## Describe the feature
Currently kafka pubsub subscribers only get passed the kafka record value (payload), record headers, and topic.

Many non-dapr kafka consumer applications make use of other information that's not provided to dapr subscribers: record key, partition #, offset #, timestamp, timestamp type. These are all available when consuming from kafka using kafka's SDKs, e.g, Java's [`ConsumerRecord`](https://javadoc.io/static/org.apache.kafka/kafka-clients/3.3.1/org/apache/kafka/clients/consumer/ConsumerRecord.html).

This additional metadata could be added to the event metadata sent to subscribers that's currently just populated with record headers. Optionally, passing this additional info could be made an opt-in feature requiring subscribers to set a flag in their subscription yaml file.

This change should also be made to the kafka input binding unless for some reason that's not supposed to have feature parity.

## Release Note

RELEASE NOTE: **ADD** Pass kafka record key and other metadata to pubsub subscribers

berndverst:
Sounds like what is required here is:

During the Kafka PubSub subscription handler additional information should be added to the cloudevent JSON.

Related / Required but a larger scope than this issue here:
Only as of this coming release can additional metadata fields be passed through Dapr PubSub via gRPC (HTTP was supported already) [this applies to all components], however note that no SDK has support for reading this data natively at this time (the mapping between incoming gRPC request to SDK native object is a manual mapping at this time which would need to be updated).

Please feel free to open a PR for the desired Kafka data:
You can see messages headers are set as metadata here. Any additional properties would also need to be added to the map of metadata.

PR #2510, opened 2023-02-05
Kafka: pass partition key inside Metadata

yosiat:
# Description

I am using components-contrib directly in my application, and having access to message's key will help me a lot.
Pass "partition key" in message's metadata for Kafka bindings/subscribers.

## Issue reference

This issue - https://github.com/dapr/components-contrib/issues/2213 shows an acceptance for such a PR, the issue discuss on additional fields. I can add more fields, but would love guidance on which fields you would like to me add.

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


ItalyPaleAle (commented):
LGTM but would love input from @skyao who is more familiar with Kafka :)

berndverst (changes requested):
Firstly, using `components-contrib `directly is not a supported use case. We may break your use case at any point and may not accept PRs that fix it. So please be aware.

Now for the specific PR here:

Are `message.Key` and `message.Partition` always set by Kafka? Are there situations where these could be empty or nil? If so your code would fail. Not knowing Kafka well enough I would like to see error checking in your type conversion.

yosiat:
@berndverst I worked on the past month moving from using dapr/components-contrib to using sarama directly, so I actually don't need this PR.

But anyway, I can do the necessary adjustments and fix this PR, if you want to land this PR in the future.. please let me know.

PR #2513, opened 2023-02-08
Workflow improvements - enable workflows for merge queues.

berndverst:

# Description

Added a PR comment to the complete build matrix action run when using `/ok-to-test`

Also enabled the Linux cache again for the PR workflow.

### DIFF
--- a/.github/workflows/components-contrib-all.yml
+++ b/.github/workflows/components-contrib-all.yml
@@ -29,12 +29,13 @@ on:
       - release-*
 jobs:
   post-comment:
-    name: Post PR comment - repository dispatch only
+    name: Post comment on Repository Dispatch
     runs-on: ubuntu-latest
     steps:
     - name: Parse repository_dispatch payload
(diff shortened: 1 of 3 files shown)


.github/workflows/components-contrib-all.yml
```diff
@@ -28,9 +28,38 @@ on:
     branches:
       - release-*
 jobs:
+  post-comment:
```
ItalyPaleAle, line 31: Consider adding an `if` directly in the job to avoid the entire job from running if the event is not repository_dispatch? https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif

berndverst: I just made another update -- I do need this job to run now (you will see why).

berndverst: Oh I think I also might need to use outputs / inputs. One moment

ItalyPaleAle (approved):

Issue #2518, opened 2023-02-09
[Proposal] OSS / cloud agnostic components should not have vendor specific naming


berndverst:
All open source components that are not cloud specific should not be named with a specific cloud provider.

E.g. Apache Dubbo is currently named `alicloud.dubbo` but the component is cloud agnostic. There are several other such components.

We should rename these components to standard OSS components. This means using an OSS name to instantiate the component in Dapr, and also moving the source code folder for clarity. Additionally documentation should only refer to the OSS / vendor neutral name.

For backwards compatibility we will maintain the existing vendor specific registry name as an alias.

berndverst:
Does this sound good to you all @yaron2 @artursouza @ItalyPaleAle @msfussell ?

berndverst:
1. Move code in contrib and rename variables as necessary.
2. Rename component name in dapr/dapr and create alias.
3. Update docs to refer to new name only.

berndverst:
> LGTM. What are the components that would be impacted, @berndverst ?
> 
> I think one more could be Azure SQL (= SQL Server)?

Azure SQL is only a docs issue! In code and the registry name is already vendor neutral. This component is also a bit weird because the component doesn't work correctly with SQL server installation locally on Windows for example from what I recall -- but in theory that component would fall under this too - that is correct.

yaron2:
If a component is not using a vendor SDK to interact with it then it should not be prefixed with a vendor name.

berndverst:
> If a component is not using a vendor SDK to interact with it then it should not be prefixed with a vendor name.
> 
> 

@yaron2 exactly, but that isn't the case right now - hence this issue.

I will do the work to rename things for 1.11.

ItalyPaleAle:
This was completed, right? @berndverst

PR #2519, opened 2023-02-10
Move vendor neutral OSS components

berndverst:

# Description

Renames `alicloud.dubbo`, `alicloud.nacos` and `alicloud.dubbo` to drop the `alicloud` prefix. These components are vendor/cloud agnostic

Part of #2518


.github/workflows/certification.yml
```diff
@@ -62,7 +62,7 @@ jobs:
         - state.memcached
         - state.mysql
         - state.sqlite
-        - bindings.alicloud.dubbo
+        - bindings.dubbo
```
ItalyPaleAle, line 65: Please fix the dubbo cert test: https://github.com/dapr/components-contrib/actions/runs/4139639228/jobs/7157452071

ItalyPaleAle (changes requested):

.github/workflows/certification.yml
berndverst: Please leave a suggestion comment I can just apply next time...

berndverst: Not possible until dapr/dapr is updated and then imported in cert tests. So I intend to temporarily break this test.

This is because of the local replace of contrib.

```
replace github.com/dapr/components-contrib => ../../../..
```

Note that inside the include dapr/dapr runtime the path is different and so this replacement is off by one folder.

berndverst:
> LGTM. This would need a corresponding dapr/dapr and dapr/docs PRs as well.

No need to state that here since I already wrote that in the issue ;)

ItalyPaleAle (approved):

PR #2523, opened 2023-02-11
[release-1.10] Azure Service Bus: fix goroutine "explosion" after recovering from failure and many other bugs

ItalyPaleAle:
This PR hopefully* fixes the issue with the goroutine "explosion" caught in our long-haul test cluster after Service Bus components attempt to recover from failure.

(* "hopefully" because goroutines seem to be under control in my environment when I "force" failures to happen, but we need to test this in an actual long-haul environment).

Includes also fixes for a lot more bugs discovered during an intense code review of the components:

- Re-add lock renewal to ASB Queue components (both pubsub and binding) which was removed - only the Topics pubsub component was renewing locks now
- Fixed: context leak when sessions are enabled
- Fixed: if number of messages received in bulk is greater than maxConcurrentHandlers, a deadlock happens
- Fixed: maxRetriableErrorsPerSec was not working as expected when receiving bulk messages
- Avoid checking if topic/queue exists every time we send a message, but only when creating a new sender
- Updated upstream SDK to 1.2.0 which contains a [relevant fix](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/messaging/azservicebus/CHANGELOG.md#120-2023-02-07)

Additionally, I've made an effort to clean up the code significantly (removing duplications etc) and to make sure the components share as much code as possible in the shared implementation package.


tests/certification/pubsub/azure/servicebus/topics/servicebus_test.go
```diff
@@ -1355,6 +1355,7 @@ func TestServicebusWithSessionsRoundRobin(t *testing.T) {
 
 func componentRuntimeOptions() []runtime.Option {
 	log := logger.NewLogger("dapr.components")
+	log.SetOutputLevel(logger.DebugLevel)
```
artursouza, line 1358: Do you need this to be merged?

ItalyPaleAle: I think it's helpful to debug failures, some logs are only debug level. But I can revert if you prefer

internal/component/azure/servicebus/client.go
```diff
@@ -174,7 +192,27 @@ func (c *Client) CloseAllSenders(log logger.Logger) {
 	close(workersCh)
 
 	// Clear the map
-	c.senders = make(map[string]*servicebus.Sender)
+	maps.Clear(c.senders)
+}
+
+// Close the client and every sender or consumer created by the connnection.
+func (c *Client) Close(log logger.Logger) {
```
mukundansundar, line 199: qq, back in the `GetSender` function, no context is passed to the newly created sender.
When the client is closed, does the SDK automatically close all senders or do we need to closed the senders also?

ItalyPaleAle: Yes https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus#Client.Close

tests/certification/pubsub/azure/servicebus/topics/servicebus_test.go
artursouza: Please, revert this to avoid regression to log levels.

artursouza: Oh. It is just cert tests. LGTM

artursouza (approved):

PR #2535, opened 2023-02-15
Add TTL support to mongodb state store

JoshVanL:
# Description

Adds TTL support to the mongodb state store using the first class TTL document index functionality.

Adds appropriate tests to cover the TTL. FYI, mongodb runs a TTL garbage collector every 60s, so a value may live past its TTL for a maximum of 60 seconds.

Part of https://github.com/dapr/dapr/issues/5899 and https://github.com/dapr/components-contrib/issues/2288

Docs PR [here](https://github.com/dapr/docs/pull/3192)

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation https://github.com/dapr/docs/pull/3192


state/mongodb/mongodb.go
```diff
@@ -146,9 +148,18 @@ func (m *MongoDB) Init(metadata state.Metadata) error {
...
-	collection := m.client.Database(meta.DatabaseName).Collection(meta.CollectionName, opts)
-
-	m.collection = collection
+	m.collection = m.client.Database(meta.DatabaseName).Collection(meta.CollectionName, opts)
+
+	// Set expireAfterSeconds index on ttl field with a value of 0 to delete
+	// values immediately when the TTL value is reached.
+	// MongoDB TTL Indexes: https://docs.mongodb.com/manual/core/index-ttl/
+	// TTL fields are deleted at most 60 seconds after the TTL value is reached.
+	if _, err := m.collection.Indexes().CreateOne(context.Background(), mongo.IndexModel{
```
ItalyPaleAle, line 157: What happens if the index already exists? Is this a no-op? And what happens if the collection already exists and has a lot of data - will this block? And what if 2 instances of Dapr create the index at the same time?

Also, nit, but I find putting complex, multi-line function invocations inside an `if` to be really hard to read

state/mongodb/mongodb.go
```diff
@@ -195,10 +206,28 @@ func (m *MongoDB) setInternal(ctx context.Context, req *state.SetRequest) error
...
-	_, err := m.collection.UpdateOne(ctx, filter, update, options.Update().SetUpsert(true))
+	reqTTL, err := stateutils.ParseTTL(req.Metadata)
+	if err != nil {
+		return fmt.Errorf("failed to parse TTL: %w", err)
+	}
 
-	return err
+	var ttl any
+	if reqTTL != nil {
+		ttl = primitive.NewDateTimeFromTime(time.Now().Add(time.Second * time.Duration(*reqTTL)))
```
ItalyPaleAle, line 216: Most other components when they set the TTL, use the time from the DB server. The problem of using the time from the app like here is that if there are multiple apps accessing the data, and there's a clock skew (not that uncommon), this may cause undefined behavior. Can the TTL be set in a relative way, such as "10 seconds from now" so MongoDB uses its own clock?

state/mongodb/mongodb.go
```diff
@@ -195,10 +206,28 @@ func (m *MongoDB) setInternal(ctx context.Context, req *state.SetRequest) error
...
+	}
+
+	update := bson.D{{"$set", bson.D{
+		{id, req.Key},
+		{value, v},
+		{etag, uuid.NewString()},
```
ItalyPaleAle, line 222: `uuid.NewString()` can panic. Please use `obj, err := uuid.NewRandom(); str := obj.String()` so you can check the error

state/mongodb/mongodb.go
```diff
@@ -195,10 +206,28 @@ func (m *MongoDB) setInternal(ctx context.Context, req *state.SetRequest) error
 		filter[etag] = uuid.NewString()
```
ItalyPaleAle, line 206: Here too actually, same issue

ItalyPaleAle (changes requested):

ItalyPaleAle:
Thanks for this PR!

Left some comments on the code. As for this:

> FYI, mongodb runs a TTL garbage collector every 60s, so a value may live past its TTL for a maximum of 60 seconds.

I think we should make sure that when the TTL is passed, the value is not returned to the app. How other components do it, after they Get a value, they check the TTL and if it's past expiration, they just return a "not found" to the caller.

(And in this case too we should use the clock of the DB server and not the app server)=

state/mongodb/mongodb.go
JoshVanL: > What happens if the index already exists? Is this a no-op? 
no-op, unless the index already exists _and_ has different options, in which case it errors out. I feel like erroring out is the safest option? Maybe it doesn't matter since the index name is already namespaced with dapr (`_dapr_ttl`) ...

>  And what happens if the collection already exists and has a lot of data - will this block?
I have added `SetBackground(true)`.

`And what if 2 instances of Dapr create the index at the same time?`
Same as above, noop.

state/mongodb/mongodb.go
JoshVanL: Done

state/mongodb/mongodb.go
ItalyPaleAle: Thanks for confirming :)

> unless the index already exists and has different options, in which case it errors out.

I am not worried about that. Users should not be modifying the schema of the collection outside of Dapr.

JoshVanL:
@ItalyPaleAle Although I have resolved or said done to the above, I haven't pushed yet.

I agree having the TTL check on gets is a good idea to get around the 60s problem.. I am just currently pulling my hair out trying to get `$$NOW` to work in the new filter :sweat_smile:

JoshVanL:
@ItalyPaleAle what a headache. Got there in the end :slightly_smiling_face: 

I have updated the test to ensure that 1. state it not returned soon after TTL, and 2. the data is _actually_ deleted from the underlying state store to prevent regression.
PTAL!

tests/conformance/state/state.go
```diff
@@ -801,12 +801,13 @@ func ConformanceTests(t *testing.T, props map[string]string, statestore state.St
 			assertEquals(t, "⏱️", res)
 
 			// Wait for the object to expire and request again
-			time.Sleep(3 * time.Second)
-			res, err = statestore.Get(context.Background(), &state.GetRequest{
-				Key: key + "-ttl",
-			})
-			require.NoError(t, err)
-			assert.Nil(t, res.Data)
+			assert.Eventually(t, func() bool {
```
ItalyPaleAle, line 804: Do we still need this change?

tests/config/state/tests.yml
```diff
@@ -44,6 +44,10 @@ components:
   - component: cloudflare.workerskv
     allOperations: false
     # Although this component supports TTLs, the minimum TTL is 60s, which makes it not suitable for our conformance tests
+    # TODO: perhaps create a special case `ttl60` operation for this component
```
ItalyPaleAle, line 47: 👍

state/mongodb/mongodb.go
```diff
@@ -192,20 +205,70 @@ func (m *MongoDB) setInternal(ctx context.Context, req *state.SetRequest) error
...
+
+	return nil
 }
 
 // Get retrieves state from MongoDB with a key.
 func (m *MongoDB) Get(ctx context.Context, req *state.GetRequest) (*state.GetResponse, error) {
+	// Since MongoDB doesn't delete the document immediately when the TTL value
+	// is reached, we need to filter out the documents with TTL value less than
+	// the current time.
+	filter := bson.D{
```
ItalyPaleAle, line 260: Random thought: "NoSQL" may be nice but looking at this really makes me miss a nice `SELECT * FROM table WHERE ttl < CURRENT_TIME` 🙃

state/mongodb/mongodb.go
```diff
@@ -192,20 +205,70 @@ func (m *MongoDB) setInternal(ctx context.Context, req *state.SetRequest) error
...
 
-	update := bson.M{"$set": bson.M{id: req.Key, value: v, etag: uuid.NewString()}}
-	_, err := m.collection.UpdateOne(ctx, filter, update, options.Update().SetUpsert(true))
+	etagV, err := uuid.NewRandom()
+	if err != nil {
+		return err
+	}
 
-	return err
+	update := mongo.Pipeline{
```
ItalyPaleAle, line 225: A bit of a nit, but doing this initialization + an `append` later will cause Go to allocate the slice with capacity 1 and then will need to re-allocate when you do an `append`. Example: https://go.dev/play/p/uryppFGd3li

Consider doing something like (I haven't tested it)

```go
	update := make(mongo.Pipeline, 2)
	update[0] = primitive.D{{"$set", bson.D{
			{id, req.Key},
			{value, v},
			{etag, etagV.String()},
		}}}

	if reqTTL != nil {
		update[1] = primitive.D{{"$addFields", bson.D{
			{daprttl, bson.D{
				{"$dateAdd",
					bson.D{{"startDate", "$$NOW"}, {"unit", "second"}, {"amount", *reqTTL}},
				},
			}},
		}}}
	} else {
		update[1] = primitive.D{{"$addFields", bson.D{
			{daprttl, nil},
		}}}
	}
```

ItalyPaleAle (approved):
Thanks for figuring this out, I saw the query and it looks like it was quite a bit of work :)

LGTM, just a couple of suggestions (non-blocking)

tests/conformance/state/state.go
JoshVanL: Thanks, I'll revert the 70 seconds to 3, but will keep the Eventually block since it speeds up the test

state/mongodb/mongodb.go
JoshVanL: :upside_down_face: :upside_down_face: :upside_down_face:

state/mongodb/mongodb.go
JoshVanL: Thanks, done

ItalyPaleAle (dismissed):

ItalyPaleAle:
@JoshVanL please fix the linter issues: https://github.com/dapr/components-contrib/actions/runs/4206220094/jobs/7299567920

JoshVanL:
@ItalyPaleAle done

state/mongodb/mongodb.go
```diff
@@ -51,6 +52,8 @@ const (
 	id               = "_id"
 	value            = "value"
 	etag             = "_etag"
+	daprttl          = "_dapr_ttl"
```
artursouza, line 55: Any reason why this attribute is prefixed with dapr while the others don't?

JoshVanL: I would like to have seen the others prefixed as well, but that cannot be changed now without a proper migration strategy.

ItalyPaleAle (approved):

state/mongodb/mongodb.go
ItalyPaleAle: IMHO because the collection is not meant for accessing outside of Dapr, so the prefix isn't needed. But I'm fine with it

JoshVanL: Let's remove the prefix while we can since its a bit of a eye sore.

ItalyPaleAle (approved):

ItalyPaleAle (approved):

ItalyPaleAle:
Good catch!

image

berndverst (approved):

berndverst:
FYI - this PR seems to have create an issue when using TTL with Mongo against Azure CosmosDB.

PR #2536, opened 2023-02-15
AWS S3 Bindings Certification Tests

robertojrojas:
# Description

Provides the AWS S3 Bindings Conformance and Certification Test Plan and Test Code.

## Issue reference

Please reference the issue this PR will close: #2247

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


.github/scripts/test-info.mjs
```diff
@@ -82,6 +82,20 @@ const components = {
             'AzureBlobStorageQueue',
         ],
     },
+    'bindings.aws.s3': {
```
berndverst, line 85: Something seems off here.

For the Docker (localstack) case it should say `certification: false` and also `requireAWSCredentials: false` and `requireTerraform: false`

You you should only run local stack for conformance test.

Of course you will also run the terraform conformance test with regular AWS. Certification is AWS + Terraform only.

In general I highly recommend you refactor `bindings.aws.s3` to `bindings.aws.s3.localstack` to make it clear what is what. I have seen strange issues with nested tests executing all the tests. It is best if these are in sibling folders.

berndverst (changes requested):

.github/scripts/test-info.mjs
berndverst, line 85: Let's remove this entry as discussed? We don't want to run the tests with docker in conformance and certification test.

robertojrojas: @berndverst absolutely! I'll take care of that today.

.github/scripts/test-info.mjs
```diff
@@ -530,6 +555,11 @@ function GenerateMatrix(testKind, enableCloudTests) {
             ) {
                 continue
             }
+        } else {
+            // For conformance tests, avoid running Docker and Cloud Tests together.
+            if (comp.conformance && comp.requireDocker) {
```
ItalyPaleAle, line 560: I think it's ok to run Docker and Cloud  tests together. The two things aren't mutually exclusive.

ItalyPaleAle: After speaking with @robertojrojas i see the benefit of having 2 separate tests

tests/certification/bindings/aws/s3/go.mod
```diff
@@ -0,0 +1,147 @@
+module s3_test
+
+go 1.19
```
berndverst, line 3:
```suggestion
go 1.20
```

berndverst (changes requested):
Can you please use `encodeBase64` in one of your certification tests and verify the behavior?

berndverst:
@robertojrojas seems our AWS service account doesn't have the necessary permissions. Perhaps @artursouza can help you get that taken care of.

robertojrojas:
@berndverst @ItalyPaleAle thanks for reviewing this PR. I've added a cert test for `encodeBase64` and I'll coordinate with @artursouza regarding the AWS S3 permissions.

tests/certification/bindings/aws/s3/s3_test.go
```diff
@@ -0,0 +1,426 @@
...
+		Operation: "create",
+		Data:      dataBytes,
+		Metadata:  invokeCreateMetadata,
+	}
+
+	return client.InvokeBinding(ctx, invokeCreateRequest)
+}
+
+// listObejctRequest is used to make a common binding request for the list operation.
+func listObejctRequest(ctx flow.Context, client daprsdk.Client) (out *daprsdk.BindingEvent, err error) {
```
artursouza, line 84: nit: typo

tests/certification/bindings/aws/s3/s3_test.go
```diff
@@ -0,0 +1,426 @@
...
+	}
+
+	out, invokeDeleteErr := client.InvokeBinding(ctx, invokeGetRequest)
+	if invokeDeleteErr != nil {
+		return nil, fmt.Errorf("%w", invokeDeleteErr)
+	}
+	return out, nil
+}
+
+// Verify S3 Basic Binding Support (Create, Get, List, Delete)
```
artursouza, line 146: What is the difference between these tests and the existing conformance tests?

tests/conformance/bindings/bindings.go
```diff
@@ -237,7 +237,7 @@ func ConformanceTests(t *testing.T, props map[string]string, inputBinding bindin
 			req.Operation = bindings.GetOperation
 			resp, err := outputBinding.Invoke(context.Background(), &req)
 			assert.NoError(t, err, "expected no error invoking output binding")
-			if createPerformed {
+			if createPerformed && resp != nil {
```
artursouza, line 240: Is this a legit behavior change? Can resp be nil even when create was performed?

Usually, changes in conformance tests need to be carefully reviewed to understand if the variation of the component is acceptable.

artursouza (changes requested):

tests/certification/bindings/aws/s3/s3_test.go
robertojrojas: One difference I see is the fact the conformance tests require either `docker` or `cloud` resources to be ready prior testing.

tests/conformance/bindings/bindings.go
robertojrojas: I rememeber see the `resp` with nil, but I can't reproduce the issue any longer, so I'm removing that change.

tests/certification/bindings/aws/s3/s3_test.go
artursouza: Is there any different in the test scenarios?

robertojrojas: For S3 Cert test in particular, it will test certain metadata tags that are not necessarily  common with other components like **forcePathStyle**, **encodeBase64**,  and **decodeBase64**.  I hope that answers your question.

artursouza: It does. Thanks.

artursouza (dismissed):
Please, wait for Bernd to approve or dismiss his review.

berndverst (approved):

PR #2538, opened 2023-02-16
Updated and trimmed dependencies

ItalyPaleAle:
# Description

> It all started with me updating Azure Event Hubs..... and then I ended up down a rabbit hole 🐰

This PR updates a bunch of dependencies and removes some redundant ones. It actually contains a bunch of code changes that were required in the process

PS: Also had to update linter to 1.51.1 or it was failing on Go 1.20

## Notable dependencies updated

- Azure Event Hubs SDK.
  - The new SDK [0.5.0](https://github.com/Azure/azure-sdk-for-go/releases/tag/sdk%2Fmessaging%2Fazeventhubs%2Fv0.5.0) fixed two bugs we reported ([1](https://github.com/Azure/azure-sdk-for-go/issues/19842) [2](https://github.com/Azure/azure-sdk-for-go/issues/19840)) 
  - It also included some breaking changes, most notably the fact that the checkpointer (which we rely on a lot) uses the "regular" Azure Blob Storage SDK rather than a fork they were using before, which leads to...
- Azure Blob Storage SDK. 
  - Updating Event Hubs brought in a new SDK for Azure Blob Storage too (1.0.0) which did contain a couple of API changes.
  - Because now Event Hubs depends on this SDK too, to avoid having the authentication code repeated in multiple places, I've refactored `internal/component/azure/blobstorage` so Event Hubs can use the some auth code
  - Also the code that Event Hubs was using to ensure storage accounts existed was more robust than the one in the blobstorage package, so now we actually check if the creation fails because the container already exists (not an error) or if it fails for other reasons (this is also a good way to test that the storage account key is valid on initialization)
- A new version of `github.com/bradfitz/gomemcache` (for Memcached) was released that also added a `Close()` method to the connection  - very useful to us to be able to actually close the connection when the component shuts down

## Dependencies removed

A bunch of dependencies have been removed, which should also help reducing the size of `daprd` (and amount of code kept in memory):

- To start, we were importing three (3!) libraries to validate or generate JWT tokens across different components: `github.com/golang-jwt/jwt/v4`, `github.com/coreos/go-oidc`, and `github.com/lestrrat-go/jwx/v2`. I've changed the code to only use the last one, which is the most complete of the three as it supports the entire JWX suite and not just JWTs (it's also the library I'm using in #2237)
- `github.com/pkg/errors` is deprecated and replaced by the standard library; this was already removed in dapr/dapr, but was not complete here.
- Same story for  `go.uber.org/atomic`, which again was removed in the runtime months ago but was still lingering around here.


ItalyPaleAle:
The linter is failing due to a bug in golangci-lint, which should be fixed in an upcoming release 1.51.2

berndverst (approved):

PR #2542, opened 2023-02-17
Refactored Actions for certification and conformance tests to better run in PRs and branches

ItalyPaleAle:
> Note: this PR is created from a branch in the dapr/components-contrib repo and not in my fork, otherwise there'd be no way to run these new pipelines with secrets (until the PR is merged). This was done with permission from @berndverst 

Problem:

- Many of our cert and conformance tests require credentials to access cloud services, which are stored as GitHub Actions secrets (or fetched from Azure Key Vault, but doing so requires the secret in Actions)
- These secrets are not available in pipelines that run in PRs created from other people's forks (_that's why this PR is created from a branch in the official repo_), so in order to run certification and conformance tests we use `/ok-to-test`
- However, when using `/ok-to-test` the YAML files of the pipelines are always fetched from the "master" branch
- This means that running tests in a release branch is almost guaranteed to fail, as tests are added/removed/changed in master. For example, if you opened a PR against release-1.10 today (which was just released yesterday), 2 certification tests would fail already because one didn't exist in release-1.10, and one was renamed.

A secondary (and smaller) problem is that our YAML files are getting busy, with lots of setup steps inlined for each component that needs it, such as starting a Docker Compose project, fetching secrets, etc.

This PR changes the workflows to make it possible to work with release branches too, and keeps things tidier:

- Tests are not defined in the file `.github/scripts/test-info.mjs`, where there's an array of objects that includes each component, whether we have a conformance or certification test, whether it needs secrets (so it can't run in PRs), etc.
- Each test can also define the name of a script to invoke before and after certification or conformance tests. Components can use that to set up the environment, such as bringing up a Docker Compose project (there's a shared `docker-compose.sh` script for that to avoid repetition, since it's very common), running Terraform, or any other setup that may be needed
- Given the kind of tests (certification or conformance) and whether secrets are available or not, the JS file generates the correct matrix of tests to run, with info such as what scripts need to be executed, what secrets need fetching, etc.

Because the `test-info.mjs` file lives in the code, it can be different in release branches to specify different sets of tests.

PS: Also added prettier to lint JS code like in dapr/dapr#5948


ItalyPaleAle:
Tests seem to be passing! SQLite is failing until #2541 is merged, but it's unrelated from the new pipelines

berndverst (approved):

Issue #2544, opened 2023-02-17
rabbitmq pubsub - possibility of using quorum queues (instead of classic queues)


VuiDJi:
## Describe the feature
We want to use rabbitmq production ready cluster with 3 replicas for dapr pusub component.

For fault tolerance, we are forced to configure the replication policy for classic queues, but [rabbitmq developers recommend to avoid using classic mirrored queues](https://github.com/rabbitmq/cluster-operator/issues/666#issuecomment-826574713). They will be deprecated and removed in the future.
They think that if you care about data safety, use quorum queues instead (quorum queues don't require policies like that at all).

Please add the ability to use quorum queues, which is extremely relevant for production scenarios.

yaron2:
Hint for implementation here: https://stackoverflow.com/questions/73670885/create-queue-quorum-rabbitmq-in-golan-by-streadway-module.

jackycsl:
Can I work on this issue?

yaron2:
> Can I work on this issue?

Of course.

thevaruntiwari:
May I know when we will receive the new release integrated with this feature support.

juanrevello:
Hi! Is there any news regarding this change? I need to create quorum queues, since rabbitmq indicates that the classic queues are to be removed. Thanks!

ItalyPaleAle:
Done in #2816

PR #2556, opened 2023-02-18
[pubsub/rabbitmq]add queue type options - classic, qourum, stream

jackycsl:
# Description

Add the options for pub-sub Rabbitmq - "classic", "qourum", or "stream". Default to "classic" mode. 

## Issue reference

Please reference the issue this PR will close: #2544

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: https://github.com/dapr/docs/pull/3193

### DIFF
--- a/pubsub/rabbitmq/metadata.go
+++ b/pubsub/rabbitmq/metadata.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"net/url"
 	"strconv"
+	"strings"
 	"time"
 
 	amqp "github.com/rabbitmq/amqp091-go"
@@ -208,6 +209,7 @@ func createMetadata(pubSubMetadata pubsub.Metadata, log logger.Logger) (*metadat
 	}
 
 	if val, found := pubSubMetadata.Properties[metadataQueueType]; found && val != "" {
+		val = strings.ToLower(val)
 		if queueTypeValid(val) {
 			result.queueType = val
 		} else {
--- a/pubsub/rabbitmq/metadata.go
+++ b/pubsub/rabbitmq/metadata.go
@@ -46,6 +46,8 @@ type metadata struct {
 	maxLen           int64
 	maxLenBytes      int64
 	exchangeKind     string
+	queueType        string
+
 	publisherConfirm bool
 	concurrency      pubsub.ConcurrencyMode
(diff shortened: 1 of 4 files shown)


pubsub/rabbitmq/metadata.go
```diff
@@ -203,6 +207,14 @@ func createMetadata(pubSubMetadata pubsub.Metadata, log logger.Logger) (*metadat
 		}
 	}
 
+	if val, found := pubSubMetadata.Properties[metadataQueueType]; found && val != "" {
+		if queueTypeValid(val) {
```
ItalyPaleAle, line 211: Can you lowercase "val" to be safer? So it would also accept "Classic"

jackycsl: Sure. I added the lowercase.

ItalyPaleAle (approved):

pubsub/rabbitmq/metadata.go
```diff
@@ -46,6 +47,8 @@ type metadata struct {
 	maxLen           int64
 	maxLenBytes      int64
 	exchangeKind     string
+	queueType        string
```
DeepanshuA, line 50: While allowing Stream queue types as well, it seems from documentation that it may be best practice to set max-age and max-length-bytes for it (https://rabbitmq.com/streams.html#retention):
```
Streams are implemented as an immutable append-only disk log. This means that the log will grow indefinitely until the disk runs out. To avoid this undesirable scenario it is possible to set a retention configuration per stream which will discard the oldest data in the log based on total log data size and/or age.
```
max-length-bytes already seems to be supported here, Should we add at least `max-age` also here (if not others like x-stream-max-segment-size-bytes)?

berndverst (changes requested):
The certification tests also need to be updated to cover any new features (and metadata options) added. Please see here https://github.com/dapr/components-contrib/blob/master/tests/certification/pubsub/rabbitmq/rabbitmq_test.go

pubsub/rabbitmq/metadata.go
jackycsl: Sure. I will add in the max-age arguments.

PR #2560, opened 2023-02-21
Fix broken links to bindings tutorial

matzew:
# Description

Fixing broken links on the development doc:

## Issue reference

#2559 was created for the defect

Issue #2559, opened 2023-02-21
Broken links on development guide


matzew:
## Expected Behavior

No broken (404) links on the https://github.com/dapr/components-contrib/blob/master/docs/developing-component.md#validating-with-dapr-core 

## Actual Behavior

the links to the quickstarts / tutorials are broken, after the "re org" of the `quickstarts` repo

berndverst:
Quickstarts are changing too much to maintain these links. You can pick any app - quickstart apps are just examples here. I suggest we link to the repo only, but not to any specific app.

matzew:
Sounds good, I will tweak the doc for this, @berndverst

matzew:
/close

this was now addressed

PR #2561, opened 2023-02-22
Just pointing to the bindings tutorial

matzew:
# Description

As suggested by @berndverst just linking to the repo. I linked to the bindings quickstart, instead of a specific app (E.g. the node app like before)

## Issue reference

Follow up PR from https://github.com/dapr/components-contrib/pull/2560, based on feedback from @berndverst in #2559


berndverst:
Your previous PR was also ok. I hadn't yet seen that you opened it. But this works, thanks!

berndverst (approved):

PR #2563, opened 2023-02-22
Feat: Add CloudWeGo kitex output binding

liu-song:
# Description
Add output binding for cloudwego/kitex ,https://github.com/cloudwego/kitex ,kitex is a Go RPC framework with high-performance and strong-extensibility for building micro-services. Integration kitex output binding has referenced apache dubbo binding code . for now , kitex generic call protocol is supported Thrift .
https://www.cloudwego.io/docs/kitex/tutorials/advanced-feature/generic-call/
 

## Issue reference
Please reference the issue this PR will close: 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


berndverst (changes requested):
Please drop `cloudwego` from all file paths -- move things to `bindings/kitex` instead.

Please see https://github.com/dapr/components-contrib/issues/2518

We will also register kitex as `bindings.kitex` instead of `bindings.cloudwego.kitex`.

bindings/cloudwego/kitex/context.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func (d *kitexContext) Init(metadata map[string]string) error {
+	if d.inited {
+		return nil
+	}
+	var destService, hostPorts string
+	destService, ok := metadata[metadataRPCDestService]
+	if !ok {
+		return fmt.Errorf("dapr bingdings cloudwego/kitex metadataRPCDestService isn't exist")
```
berndverst, line 58:
```suggestion
		return fmt.Errorf("dapr binding kitex metadataRPCDestService isn't exist")
```

bindings/cloudwego/kitex/context.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	hostPorts, ok = metadata[metadataRPCHostports]
+	if !ok {
+		return fmt.Errorf("dapr bingdings cloudwego/kitex metadataRPCHostPorts isn't exist")
```
berndverst, line 62: Same for these error messages

bindings/cloudwego/kitex/context.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	_, ok = metadata[metadataRPCMethodName]
+	if !ok {
+		return fmt.Errorf("dapr bingdings cloudwego/kitex metadataRPCMethodName isn't exist")
+	}
+
+	genericCli, err := genericclient.NewClient(destService, generic.BinaryThriftGeneric(), client.WithHostPorts(hostPorts))
+	if err != nil {
+		return fmt.Errorf("get gerneric service of kitex failed %w", err)
```
berndverst, line 71:
```suggestion
		return fmt.Errorf("get generic service of kitex failed %w", err)
```

What does this actually mean? Does a user of this binding know what a generic client is? And can there be any explanation why it failed?

go.mod
```diff
@@ -40,12 +40,15 @@ require (
 	github.com/apache/dubbo-go-hessian2 v1.11.5
 	github.com/apache/pulsar-client-go v0.9.0
 	github.com/apache/rocketmq-client-go/v2 v2.1.1-rc2
-	github.com/aws/aws-sdk-go v1.44.202
+	github.com/apache/thrift v0.13.0
+	github.com/aws/aws-sdk-go v1.44.204
```
DeepanshuA, line 44: Just curious: why do we need to upgrade aws-sdk-go version with kitex?

berndverst (changes requested):
Can you write an end to end test in the certification test folder which uses the kitex binding to call another service and then validate the the server received the expected response / kitex also received the expected response etc?

And normally we also have a binding conformance test you should try to pass:
https://github.com/dapr/components-contrib/blob/master/tests/conformance/bindings/bindings.go

To do that add your component yaml file here:
https://github.com/dapr/components-contrib/tree/master/tests/config/bindings

And do edit the `tests.yml` there. You can then run your conformance test with:
`go test -v -tags=conftests -count=1 ./tests/conformance -run="TestBindingsConformance/kitex"`

liu-song:
End-to-end testing has been completed, please review again, thanks

bindings/kitex/kitex_output.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/dapr/kit/logger"
+)
+
+type kitexOutputBinding struct {
+	ctxCache  map[string]*kitexContext
+	cacheLock sync.RWMutex
+	logger    logger.Logger
+}
+
+var kitexBinding *kitexOutputBinding
```
ItalyPaleAle, line 30: We should not be using singletons because:

1. We can have multiple components initialized of the same type
2. This is not safe for concurrent access
3. We should be able to Close components

ItalyPaleAle (changes requested):

bindings/kitex/kitex_output.go
liu-song: Got it，I've learned it.

tests/conformance/bindings/bindings.go
```diff
@@ -103,6 +104,13 @@ func startHTTPServer(url string) {
 }
 
 func (tc *TestConfig) createInvokeRequest() bindings.InvokeRequest {
+	if tc.ComponentType == "bindings" && tc.ComponentName == "kitex" {
```
berndverst, line 107: @ItalyPaleAle thoughts on this approach for conformance test?

berndverst (approved):

ItalyPaleAle (approved):

tests/conformance/bindings/bindings.go
ItalyPaleAle: I think it's ok given that it's a unique case. Although maybe unnecessary given that there's a certification test?

ItalyPaleAle: As discussed with @berndverst , we've dropped the conformance test since the same things were covered by the certification and unit tests. Thanks for submitting the PR, we're good to merge now :)

berndverst (approved):

Issue #2509, opened 2023-02-05
Support reconnection between sidecar and broker for RabbitMQ bindings


Syuparn:
## Describe the feature

Currently, when RabbitMQ broker is restarted, the binding component connections between Dapr sidecar and RabbitMQ are closed and not reconnected anymore. Once the connection is closed, applications will not able to use binding components unless the sidecar is restarted.

```
{"errorCode":"ERR_INVOKE_OUTPUT_BINDING","message":"error when invoke output binding myevent: Exception (504) Reason: \"channel/connection is not open\""}
```

RabbitMQ pub/sub component can reconnect to RabbitMQ (https://github.com/dapr/dapr/issues/1688) using `reconnectWait` metadata. This should also be implemented for binding.

## Release Note

RELEASE NOTE: ADDED RabbitMQ's binding implementation to automatically reconnect on 'channel/connection is not open'

PR #2565, opened 2023-02-22
Support reconnection between sidecar and broker for RabbitMQ bindings

Taction:
# Description

Support reconnect to RabbitMQ when RabbitMQ restarts or closes the channel with daprd. 
Also, continue consuming messages after reconnected. 

## Issue reference

Please reference the issue this PR will close: #2509

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/bindings/rabbitmq/rabbitmq.go
+++ b/bindings/rabbitmq/rabbitmq.go
@@ -25,10 +25,11 @@ import (
 
 	amqp "github.com/rabbitmq/amqp091-go"
 
+	"github.com/dapr/kit/logger"
+
 	"github.com/dapr/components-contrib/bindings"
 	"github.com/dapr/components-contrib/internal/utils"
 	contribMetadata "github.com/dapr/components-contrib/metadata"
-	"github.com/dapr/kit/logger"
 )
 
 const (
@@ -43,7 +44,13 @@ const (
 	rabbitMQMaxPriorityKey     = "x-max-priority"
 	defaultBase                = 10
 	defaultBitSize             = 0
+
+	errorChannelConnection = "channel/connection is not open"
+	logMessagePrefix       = "rabbitmq binding "
+	defaultReconnectWait   = 5 * time.Second
 )
+
+var errClosed = errors.New("component is stopped")
 
 // RabbitMQ allows sending/receiving data to/from RabbitMQ.
 type RabbitMQ struct {
@@ -55,6 +62,9 @@ type RabbitMQ struct {
 	closed     atomic.Bool
 	closeCh    chan struct{}
 	wg         sync.WaitGroup
+
+	// used for reconnect
+	notifyRabbitChannelClose chan *amqp.Error
 }
 
 // Metadata is the rabbitmq config.
@@ -84,27 +94,67 @@ func (r *RabbitMQ) Init(_ context.Context, metadata bindings.Metadata) error {
 		return err
 	}
 
-	conn, err := amqp.Dial(r.metadata.Host)
+	err = r.connect()
 	if err != nil {
 		return err
 	}
-
-	ch, err := conn.Channel()
-	if err != nil {
-		return err
-	}
-	ch.Qos(r.metadata.PrefetchCount, 0, true)
-	r.connection = conn
-	r.channel = ch
-
-	q, err := r.declareQueue()
-	if err != nil {
-		return err
-	}
-
-	r.queue = q
-
+	r.notifyRabbitChannelClose = make(chan *amqp.Error, 1)
+	r.channel.NotifyClose(r.notifyRabbitChannelClose)
+	go r.reconnectWhenNecessary()
 	return nil
+}
+
+func (r *RabbitMQ) reconnectWhenNecessary() {
+	for {
+		select {
+		case <-r.closeCh:
+			return
+		case _ = <-r.notifyRabbitChannelClose:
+			// this is called when the server restart or the channel is closed by server.
+			if r.connection != nil && !r.connection.IsClosed() {
+				ch, err := r.connection.Channel()
+				if err == nil {
+					r.channel = ch
+					r.notifyRabbitChannelClose = make(chan *amqp.Error, 1)
+					r.channel.NotifyClose(r.notifyRabbitChannelClose)
+					continue
+				}
+				// if encounter err fallback to reconnect connection
+			}
+			// keep trying to reconnect
+			for {
+				err := r.connect()
+				if err == nil {
+					break
+				}
+				if err == errClosed {
+					return
+				}
+				r.logger.Warnf("reconnect failed: %s", err.Error())
+
+				time.Sleep(defaultReconnectWait)
+			}
+
+			r.notifyRabbitChannelClose = make(chan *amqp.Error, 1)
+			r.channel.NotifyClose(r.notifyRabbitChannelClose)
+		}
+	}
+}
+
+func dial(uri string) (conn *amqp.Connection, ch *amqp.Channel, err error) {
+	conn, err = amqp.Dial(uri)
+
+	if err != nil {
+		return nil, nil, err
+	}
+
+	ch, err = conn.Channel()
+	if err != nil {
+		conn.Close()
+		return nil, nil, err
+	}
+
+	return conn, ch, nil
 }
 
 func (r *RabbitMQ) Operations() []bindings.OperationKind {
@@ -112,6 +162,10 @@ func (r *RabbitMQ) Operations() []bindings.OperationKind {
 }
 
 func (r *RabbitMQ) Invoke(ctx context.Context, req *bindings.InvokeRequest) (*bindings.InvokeResponse, error) {
+	// check if connection channel to rabbitmq is open
+	if r.channel == nil {
+		return nil, errors.New(errorChannelConnection)
+	}
 	pub := amqp.Publishing{
 		DeliveryMode: amqp.Persistent,
 		ContentType:  "text/plain",
@@ -238,21 +292,7 @@ func (r *RabbitMQ) Read(ctx context.Context, handler bindings.Handler) error {
 		return errors.New("binding already closed")
 	}
 
-	msgs, err := r.channel.Consume(
-		r.queue.Name,
-		"",
-		false,
-		false,
-		false,
-		false,
-		nil,
-	)
-	if err != nil {
-		return err
-	}
-
 	readCtx, cancel := context.WithCancel(ctx)
-
 	r.wg.Add(2)
 	go func() {
 		defer r.wg.Done()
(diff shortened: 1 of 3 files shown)


bindings/rabbitmq/rabbitmq.go
```diff
@@ -55,17 +63,21 @@ type RabbitMQ struct {
...
-	PrefetchCount    int    `json:"prefetchCount"`
-	MaxPriority      *uint8 `json:"maxPriority"` // Priority Queue deactivated if nil
+	Host             string        `json:"host"`
+	QueueName        string        `json:"queueName"`
+	Exclusive        bool          `json:"exclusive,string"`
+	Durable          bool          `json:"durable,string"`
+	DeleteWhenUnused bool          `json:"deleteWhenUnused,string"`
+	PrefetchCount    int           `json:"prefetchCount"`
+	MaxPriority      *uint8        `json:"maxPriority"` // Priority Queue deactivated if nil
+	ReconnectWait    time.Duration `json:"reconnectWait"`
```
ItalyPaleAle, line 79: Because this is parsed as `reconnectWaitSeconds`, please make this property unexported, i.e.:

```suggestion
	reconnectWait    time.Duration
```

bindings/rabbitmq/rabbitmq.go
```diff
@@ -84,34 +96,78 @@ func (r *RabbitMQ) Init(_ context.Context, metadata bindings.Metadata) error {
...
-	if err != nil {
-		return err
+func (r *RabbitMQ) reconnectWhenNecessary() {
+	for {
+		select {
+		case <-r.closeCh:
+			return
+		case _ = <-r.notifyRabbitChannelClose:
+			// this is called when the server restart or the channel is closed by server.
+			if r.connection != nil && !r.connection.IsClosed() {
```
ItalyPaleAle, line 115: This probably needs to be wrapped in a lock or it can cause race conditions

bindings/rabbitmq/rabbitmq.go
```diff
@@ -84,34 +96,78 @@ func (r *RabbitMQ) Init(_ context.Context, metadata bindings.Metadata) error {
...
+	return conn, ch, nil
 }
 
 func (r *RabbitMQ) Operations() []bindings.OperationKind {
 	return []bindings.OperationKind{bindings.CreateOperation}
 }
 
 func (r *RabbitMQ) Invoke(ctx context.Context, req *bindings.InvokeRequest) (*bindings.InvokeResponse, error) {
+	// check if connection channel to rabbitmq is open
+	if r.channel == nil {
```
ItalyPaleAle, line 167: Ditto. This probably needs to be wrapped in a lock or it can cause race conditions

ItalyPaleAle (changes requested):

bindings/rabbitmq/rabbitmq.go
berndverst: Agreed with @ItalyPaleAle

Taction: Done.

bindings/rabbitmq/rabbitmq.go
Taction: Done.

bindings/rabbitmq/rabbitmq.go
Taction: Done.

bindings/rabbitmq/rabbitmq.go
```diff
@@ -84,34 +97,81 @@ func (r *RabbitMQ) Init(_ context.Context, metadata bindings.Metadata) error {
 		return err
 	}
 
-	conn, err := amqp.Dial(r.metadata.Host)
+	err = r.connect()
 	if err != nil {
 		return err
 	}
+	r.notifyRabbitChannelClose = make(chan *amqp.Error, 1)
+	r.channel.NotifyClose(r.notifyRabbitChannelClose)
```
DeepanshuA, line 105: NotifyClose returns an Error struct(https://pkg.go.dev/github.com/rabbitmq/amqp091-go#Error), which also has a Recover attribute, which returns `true when this error can be recovered by retrying later or with different parameters`. Should we check Error.Retry's value before retrying to connect? And, if we still want to re-connect, should we at least wait for sometime, so that we don't keep retrying a connection which may be un-retriable at the moment.

Taction: Yes, you are right! If the recover attribute is false, we should wait for sometime. 
Since this is only notified when the channel is closed, and the error struct is a common struct, so I'd more like to keep the reconnect part.

ItalyPaleAle (dismissed):

DeepanshuA (approved):

ItalyPaleAle:
@berndverst good to merge?

ItalyPaleAle:
@Taction can you please open a PR in dapr/docs to document the new metadata?

@berndverst is this good to merge? Would appreciate your review too

Taction:
Docs issue and pr created.

bindings/rabbitmq/rabbitmq.go
```diff
@@ -230,66 +291,153 @@ func (r *RabbitMQ) declareQueue() (amqp.Queue, error) {
...
+				false,
+				false,
+				false,
+				false,
+				nil,
+			)
+			if err != nil {
+				r.logger.Errorf("Error consuming messages from queue [%s]: %v", r.queue.Name, err)
+				goto Retry
+			}
```
JoshVanL, line 343: To prevent us from needing to use a `goto`. This also removes the sleep below which currently always prevents `Close` from returning up to `r.metadata.reconnectWait` duration.

```suggestion
			if err != nil {
				r.logger.Errorf("Error consuming messages from queue [%s]: %v", r.queue.Name, err)
				select {
				  case <-time.After(r.metadata.reconnectWait):
				  case <-readCtx.Done():
				    r.logger.Info("Input binding closed, stop fetching message")
				    return
				}
			}
```

bindings/rabbitmq/rabbitmq.go
```diff
@@ -230,66 +291,153 @@ func (r *RabbitMQ) declareQueue() (amqp.Queue, error) {
...
+				err               error
+				declaredQueueName string
+				ch                *amqp.Channel
+			)
+			r.channelMutex.RLock()
+			declaredQueueName = r.queue.Name
+			ch = r.channel
+			r.channelMutex.RUnlock()
+			if ch == nil {
+				goto Retry
```
JoshVanL, line 329:
```suggestion
				select {
				  case <-time.After(r.metadata.reconnectWait):
				    continue
				  case <-readCtx.Done():
				    r.logger.Info("Input binding closed, stop fetching message")
				    return
				}
```

bindings/rabbitmq/rabbitmq.go
```diff
@@ -230,66 +291,153 @@ func (r *RabbitMQ) declareQueue() (amqp.Queue, error) {
...
+			)
+			if err != nil {
+				r.logger.Errorf("Error consuming messages from queue [%s]: %v", r.queue.Name, err)
+				goto Retry
+			}
+			r.handleMessage(readCtx, handler, msgs, ch)
+
+		Retry:
+			if r.closed.Load() || readCtx.Err() != nil {
+				r.logger.Info("Input binding closed, stop fetching message")
```
JoshVanL, line 348:
```suggestion

```

bindings/rabbitmq/rabbitmq.go
```diff
@@ -230,66 +291,153 @@ func (r *RabbitMQ) declareQueue() (amqp.Queue, error) {
...
+	if err != nil {
+		return err
+	}
+
+	r.notifyRabbitChannelClose = make(chan *amqp.Error, 1)
+	ch.NotifyClose(r.notifyRabbitChannelClose)
+
+	r.channelMutex.Lock()
+	// try to close the old channel and connection, ignore the error
+	_ = r.reset() //nolint:errcheck
```
JoshVanL, line 413: Do we not want to at least log this error message?

JoshVanL (changes requested):
Some nits from me :slightly_smiling_face: 

We should add to the wait group for the reconnect routine so it is waited for in `Close`, and use `select`s rather than `time.Sleep`s so that `closeCh` can be responded to immediately. If we don't use `select`, then `Close` will needlessly hang up to `r.metadata.reconnectWait` duration.

bindings/rabbitmq/rabbitmq.go
Taction: The error message is logged inside the `reset` function. So here do not need logs.

bindings/rabbitmq/rabbitmq.go
```diff
@@ -84,53 +95,100 @@ func (r *RabbitMQ) Init(_ context.Context, metadata bindings.Metadata) error {
...
-		return err
+func (r *RabbitMQ) reconnectWhenNecessary() {
+	for {
+		select {
+		case <-r.closeCh:
+			return
+		case e := <-r.notifyRabbitChannelClose:
+			// If this error can not be recovered, first wait and then retry.
+			if e != nil && !e.Recover {
+				time.Sleep(r.metadata.reconnectWait)
```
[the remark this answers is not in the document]
Taction: Done.

bindings/rabbitmq/rabbitmq.go
```diff
@@ -84,53 +95,100 @@ func (r *RabbitMQ) Init(_ context.Context, metadata bindings.Metadata) error {
...
+				err := r.connect()
+				if err == nil {
+					break
+				}
+				if err == errClosed {
+					return
+				}
+				r.logger.Warnf("Reconnect failed: %v", err)
+
+				time.Sleep(r.metadata.reconnectWait)
```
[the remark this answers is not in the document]
Taction: Done.

bindings/rabbitmq/rabbitmq.go
Taction: To avoid the goto and the duplicate select, I slightly changed the code. When there's any error or `r.handleMessage` finished, it will eventually execute the select part.

bindings/rabbitmq/rabbitmq.go
Taction: Same as above.

bindings/rabbitmq/rabbitmq.go
Taction: Same as above.

bindings/rabbitmq/rabbitmq.go
```diff
@@ -230,66 +303,155 @@ func (r *RabbitMQ) declareQueue() (amqp.Queue, error) {
...
+					false,
+					false,
+					false,
+					false,
+					nil,
+				)
+				if err == nil {
+					// all good, handle messages
+					r.handleMessage(readCtx, handler, msgs, ch)
+				} else {
```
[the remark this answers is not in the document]
Taction: From my point of view, when`r.handleMessage` is finished there is more likely the ctx is canceled or the client lost its connection with the server, so it is worth checking the ctx or waiting for the reconnect.

ItalyPaleAle (approved):

PR #2594, opened 2023-02-23
Improvements to Consul nameresolver

### LINKED ISSUE #2489: Consul nameresolution resolves conflicts
Consul does not automatically delete or overwrite service registrations. When services `service-a` and `service-b` exist and the following behavior occurs.
A parsing error occurs

`service-a` = 192.168.1.2:3500

`service-b` = 192.168.1.3:3500

At this point consul has the following parsing

```
service-a -> 192.168.1.2:3500 (health check passing)

service-b -> 192.168.1.3:3500(Health check passing)
```

Restart both services and reassign service ip

`service-a` -> 192.168.1.4:3500

`service-b` -> 192.168.1.2:3500

At this point consul has the following resolution

```
service-a -> 192.168.1.4:3500 (health check passing)
service-a -> 192.168.1.2:3500(Health check passing)

service-b -> 192.168.1.2:3500(Health check passing)
service-b -> 192.168.1.3:3500(Health check failed)
```

Now when we access `service-a` there is a chance that it will resolve to `192.168.1.2:3500` resulting in the server actually accessing `service-b`

## Release Note

RELEASE NOTE: Fix consul name-resolution resolves conflicts

ItalyPaleAle:
# Description

This PR contains improvements to the Consul nameresolver, especially the `ResolveID` method which was more complex than it needed to be. In particular, the algorithm for picking a random entry from the list of results could be replaced with a much simpler one (complexity: O(n) to O(1), and removes N allocations on the heap), plus other small tweaks.

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


ItalyPaleAle:
Moving this back to draft as I realized there's a better way to do this: https://developer.hashicorp.com/consul/api-docs/agent/check#deregistercriticalserviceafter

The risk is that if we delete a service, we may delete multiple instances when we just need to delete one. I'll also test in a Load balanced scenario

ItalyPaleAle:
After further investigation, looks like the initial approach of unregistering the service won't work and can't be made to work. I'm leaving this PR with only the "ancillary" improvements to the component.

nameresolution/consul/consul.go
```diff
@@ -81,96 +79,84 @@ type resolverConfig struct {
...
 		return fmt.Errorf("failed to init consul client: %w", err)
 	}
 
-	// register service to consul
+	// Register service to consul
 	if r.config.Registration != nil {
-		if err := r.client.Agent().ServiceRegister(r.config.Registration); err != nil {
+		agent := r.client.Agent()
+
+		err = agent.ServiceRegister(r.config.Registration)
```
DeepanshuA, line 108: I saw your comment about using `deregistercriticalserviceafter` - but that fix not working.

Should we rather deregister before registering?
Documentation for `ServiceDeregister(serviceID string) error` (https://developer.hashicorp.com/consul/commands/services/deregister) also recommends that it ` is meant to be paired with services register`.

It may not completely solve the issue but would certainly be a good practice most probably and may also help in solving issues like https://github.com/dapr/components-contrib/issues/2489

ItalyPaleAle: That was actually what I first tried doing, and it didn't work because a service can have multiple instances if the app is scaled horizontally. Calling "ServiceDeregister" removed the entire service and so un-registered other instances of the app too.

I spent 2 hours on this yesterday and doesn't really look like Consul has any way to remove only one instance :(

DeepanshuA: Ohh ok, got it.

berndverst (approved):

PR #2596, opened 2023-02-24
Resolve long-standing TODO in mDNS component

ItalyPaleAle:
Changes the mDNS component to resolve a long-standing TODO in the runtime:

https://github.com/dapr/dapr/blob/77a3f1669c83253c203b9e30b9c8d7f59284c66d/pkg/runtime/runtime.go#L2101-L2111


nameresolution/metadata.go
```diff
@@ -16,6 +16,8 @@ package nameresolution
 import "github.com/dapr/components-contrib/metadata"
 
 const (
+	// TODO: REMOVE THESE AFTER RUNTIME IS CHANGED
```
ItalyPaleAle, line 19: Reviewers: The MDNS constants below are unused but I'm not removing them in this PR because they have to be changed in the runtime first by pinning the latest components-contrib. Otherwise the runtime will break and so will our certification tests

berndverst (approved):

PR #2597, opened 2023-02-24
Remove deprecated components: pubsub.hazelcast, bindings.twitter

### LINKED ISSUE #1808: Deprecate pubsub.hazelcast due to no "at least once delivery" guarantee
Currently, Hazelcast does not seem to provide a way for a consumer to submit an ACK or NACK after processing a message. As such, messages that fail to be processed are not sent back to the broker and are discard (also if the runtime crashes).

This seems to be a limitation in Hazelcast itself, as topics do not seem to have a way for consumer to n/ack messages: https://docs.hazelcast.com/imdg/4.2/data-structures/topic

This should at least prevent us from making pubsub.hazelcast a stable component. @yaron2 suggested we may consider dropping this component entirely as we can't offer the "at least once delivery" guarantee.

### LINKED ISSUE #2503: [Proposal] Deprecate Twitter Binding component [approved by all maintainers]
## Describe the proposal

Deprecate the Twitter binding component.

Why:

- The industry and developers are increasingly less interested in Twitter
- Twitter is planning to charge for API access beginning next week, further decreasing the usage of Twitter APIs: https://arstechnica.com/tech-policy/2023/02/cash-strapped-twitter-to-start-charging-developers-for-api-access-next-week/
- New/Current Twitter API accounts are not supported by the existing binding - for these users the component is not usable.
- Maintainer burden: Current Twitter API accounts can only use the v2 API and Dapr's Twitter component is incompatible with the v2 API. However v2 clients are incompatible with the v1 API. We would need to create a new implementation for the v2 client but also maintain the v1 client. This creates an unnecessary maintenance burden for an integration that primarily is used in demos, but not in business situations.

Benefits:
+ No need to maintain to different Twitter integrations
+ Free up memory for the Dapr sidecar

ItalyPaleAle:
# Description

Removes the source code of pubsub.hazelcast and bindings.twitter which are slated to be removed in Dapr 1.11

## Issue reference

Fixes #2503
Fixes #1808


berndverst:
We might need to revert this PR if we get too much pushback about Twitter. Ok for now.

berndverst (approved):

Issue #2615, opened 2023-03-02
Oauth2 Not Adding Token to Request Header


drewby:
## Expected Behavior

Oauth2 Middleware component for Authorization Code Grant flow is supposed to add the Bearer token to a request header after authentication.  This header is then sent along with the request to the app for validation and claims lookup. This works correctly in v1.9.

## Actual Behavior

In v1.10, the Oauth2 Middleware component is adding the Bearer token to the Response headers instead of the Request headers. The token is going back to the client application which is not expected.

## Steps to Reproduce the Problem

1. Simple controller to return headers (ASP.NET core / C#):

```csharp
var app = WebApplication.Create(args);

app.MapGet("/Headers", (HttpRequest request) =>
{
    var headers = request.Headers;
    return new Microsoft.AspNetCore.Mvc.ObjectResult(headers);    
});

app.Run();
```

2. Create component config

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: oauth2
spec:
  type: middleware.http.oauth2
  version: v1
  metadata:
  - name: clientId
    value: <clientId>
  - name: clientSecret
    value: <clientSecret>
  - name: scopes
    value: "api://appId/TodoList.Read, api://appId/TodoList.ReadWrite"
  - name: authURL
    value: "https://login.microsoftonline.com/tenantId/oauth2/v2.0/authorize"
  - name: tokenURL
    value: "https://login.microsoftonline.com/tentId/oauth2/v2.0/token"
  - name: redirectURL
    value: "http://localhost:8080/v1.0/invoke/test/method/Headers"
  - name: authHeaderName
    value: "Authorization" 
```

3. Configure middleware

```yaml
apiVersion: dapr.io/v1alpha1
kind: Configuration
metadata:
  namespace: nginx
  name: dapr-config
spec: 
  httpPipeline:
    handlers:
      - name: oauth2 
        type: middleware.http.oauth2
```

4.  Run dapr

```bash
dapr run --app-id test --config config.yaml --components-path ./components/ --app-port 5000 -H 8080 -- dotnet run --project ./src --urls http://localhost:5000
```

With daprd 1.9, we see the Bearer token header. In daprd 1.10, there is no Bearer token header.

## Release Note

RELEASE NOTE: **FIX** Oauth2 middleware adds Bearer token to request headers.

ItalyPaleAle:
[...]

PR #2616, opened 2023-03-02
Oauth2 Middleware fix for Bearer token header

drewby:
# Description

Fixes issue where the Bearer token is not being added to Authorization request header. 

## Issue reference

close: #2615

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [N/A] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_


middleware/http/oauth2/oauth2_middleware.go
```diff
@@ -83,7 +83,7 @@ func (m *Middleware) GetHandler(ctx context.Context, metadata middleware.Metadat
 			session := sessions.Start(w, r)
 
 			if session.GetString(meta.AuthHeaderName) != "" {
-				w.Header().Add(meta.AuthHeaderName, session.GetString(meta.AuthHeaderName))
+				r.Header.Add(meta.AuthHeaderName, session.GetString(meta.AuthHeaderName))
```
DeepanshuA, line 86: We have another place as well, where this change would be required: https://github.com/dapr/components-contrib/blob/907b6b9ebeeba9c4f74f4df1ae3316eae3cb784e/middleware/http/oauth2/oauth2_middleware.go#L139

Can we have a unit test for it, so that we don't get this issue at least again?

ItalyPaleAle:
Thanks, I left a comment on the associated issue.

middleware/http/oauth2/oauth2_middleware.go
drewby: > We have another place as well, where this change would be required:
> 
> https://github.com/dapr/components-contrib/blob/907b6b9ebeeba9c4f74f4df1ae3316eae3cb784e/middleware/http/oauth2/oauth2_middleware.go#L139
> 
> Can we have a unit test for it, so that we don't get this issue at least again?

I was thinking the same but it was going to require injecting a mock for getting tokens (similar to oauth2clientcredentials component). Its a more significant code change. I'll give it a go today and update the PR.

drewby: FYI. 
I updated the second location where header is added. Thanks for pointing that out. 
I created a test for the first case which doesn't require a Mock. 
I'll work on the second case which requires a Mock.

drewby: @DeepanshuA There are tests added for both places where a Request header is added.

Do you see a purpose for the Header being added after the token Exchange? The next line sends a redirect response to the client, so the request header has no purpose. 

The header (before the redirect) has been added that way all the back to v0.1, but I'm not sure the reason.

ItalyPaleAle: I agree the header on (current) L148 is irrelevant

ItalyPaleAle (approved):

middleware/http/oauth2/oauth2_middleware.go
drewby: I removed the unneeded header and the test.

ItalyPaleAle (approved):

middleware/http/oauth2/oauth2_middleware.go
DeepanshuA: Yup, agreed - it is a redirect - I missed that. request header is not required at that place then.

PR #2618, opened 2023-03-02
Adds TTL support to the CockroachDB state.

JoshVanL:
Part of the larger [issue](https://github.com/dapr/dapr/issues/5899) and https://github.com/dapr/components-contrib/issues/2288.

Adds TTL support to the cockroachDB state store by running a garbage collector which runs at a configurable interval, which deletes items saved.

Cleanups are coordinated in a new `dapr_metadata` table.

Follows much of the same implementation as [postgres](https://github.com/dapr/components-contrib/pull/2302)

Docs PR:  https://github.com/dapr/docs/pull/3226

/cc @ItalyPaleAle

### DIFF
--- a/state/cockroachdb/cockroachdb_access.go
+++ b/state/cockroachdb/cockroachdb_access.go
@@ -68,6 +68,13 @@ type cockroachDBMetadata struct {
 	MaxConnectionAttempts *int
 }
 
+// Interface that contains methods for querying.
+type dbquerier interface {
+	ExecContext(context.Context, string, ...any) (sql.Result, error)
+	QueryContext(context.Context, string, ...any) (*sql.Rows, error)
+	QueryRowContext(context.Context, string, ...any) *sql.Row
+}
+
 // newCockroachDBAccess creates a new instance of CockroachDBAccess.
 func newCockroachDBAccess(logger logger.Logger) *CockroachDBAccess {
 	logger.Debug("Instantiating new CockroachDB state store")
@@ -175,6 +182,10 @@ func (c *CockroachDBAccess) Init(ctx context.Context, metadata state.Metadata) e
 
 // Set makes an insert or update to the database.
 func (c *CockroachDBAccess) Set(ctx context.Context, req *state.SetRequest) error {
+	return c.set(ctx, c.db, req)
+}
+
+func (c *CockroachDBAccess) set(ctx context.Context, d dbquerier, req *state.SetRequest) error {
 	c.logger.Debug("Setting state value in CockroachDB")
 
 	value, isBinary, err := validateAndReturnValue(req)
@@ -203,7 +214,7 @@ func (c *CockroachDBAccess) Set(ctx context.Context, req *state.SetRequest) erro
 	if req.ETag == nil {
 		query = `
 INSERT INTO %[1]s
-  (key, value, isbinary, etag, expire_at)
+  (key, value, isbinary, etag, expiredate)
 VALUES
   ($1, $2, $3, 1, %[2]s)
 ON CONFLICT (key) DO UPDATE SET
@@ -211,7 +222,7 @@ ON CONFLICT (key) DO UPDATE SET
   isbinary = $3,
   updatedate = NOW(),
   etag = EXCLUDED.etag + 1,
-  expire_at = %[2]s
+  expiredate = %[2]s
 ;`
 		params = []any{req.Key, value, isBinary}
 	} else {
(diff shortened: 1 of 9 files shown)


state/cockroachdb/cockroachdb_access.go
```diff
@@ -310,66 +405,68 @@ func (p *cockroachDBAccess) BulkDelete(ctx context.Context, req []state.DeleteRe
...
-			if err != nil {
-				tx.Rollback()
-				return err
-			}
-
-			err = p.Delete(ctx, &delReq)
-			if err != nil {
-				tx.Rollback()
-				return err
+	performTx := func() error {
```
ItalyPaleAle, line 416: I would really like to avoid using lambdas when it's just to be able to `defer` a call inside, because they are not light on the heap. Was there an issue with the previous code?

ItalyPaleAle (changes requested):
Initial pass

JoshVanL:
@ItalyPaleAle honestly I don't know enough about CockroachDB to say if there are any real differences in implementations that would effect us without a bit of research, however given a lot of code _looks_ mostly the same I would assume we could generalize most of it. A surface level look at the docs would suggest we can reuse almost everything outside of superficial naming etc.

I'll let others chime in in case they have some more authoritative knowledge on this.

berndverst:
Can we just run the postgres conformance and certification tests against cockroachdb to see if they work? That would be telling :)

I'm all for removing the CockroachDB component and using the Postgres component instead if that works!

ItalyPaleAle:
If I were to _guess_, a conflict could be with the use of the `xmin` column for determining the ETag, like the Postgres component does.

So I don't think the components are exactly interchangeable today (it's more likely you can use the CockroachDB component against postgres however). But If the changes are as small as the `xmin` thing, there could be a lot of code reuse possible.

But maybe that could be a separate PR :)

berndverst:
@JoshVanL linter error - something about SQL string formatting

state/cockroachdb/cockroachdb_access.go
```diff
@@ -310,66 +405,68 @@ func (p *cockroachDBAccess) BulkDelete(ctx context.Context, req []state.DeleteRe
...
+			switch o.Operation {
+			case state.Upsert:
+				var setReq state.SetRequest
+
+				setReq, err = getSet(o)
+				if err != nil {
+					return err
+				}
+
+				err = c.Set(ctx, &setReq)
```
ItalyPaleAle, line 427: One more thing on this method, as well as all the Bulk* like BulkSet and BulkDelete... Not something added in this PR, but this is actually NOT using transactions, so it's not bulk and can't be rolled back in case of errors.

Although we do start a transaction (see `c.db.Begin()` above), the transaction is not passed to the methods that do Get, Set, Delete, etc. So it's not being used....

state/cockroachdb/cockroachdb_access.go
```diff
@@ -411,54 +508,93 @@ func (p *cockroachDBAccess) Ping(ctx context.Context) error {
...
-		_, err = p.db.Exec(createTable)
+		c.logger.Info("Creating CockroachDB state table")
+		_, err = c.db.ExecContext(ctx, fmt.Sprintf(`CREATE TABLE %s (
+  key text NOT NULL PRIMARY KEY,
+  value jsonb NOT NULL,
+  isbinary boolean NOT NULL,
+  etag INT,
+  insertdate TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+  updatedate TIMESTAMP WITH TIME ZONE NULL,
+  expire_at TIMESTAMP WITH TIME ZONE NULL,
```
ItalyPaleAle, line 552: Can you call this `expiredate` so it's the same as in the Postgres component? (all other columns are the same except etag which doesn't exist in postgres since we use the xid)

ItalyPaleAle (changes requested):

JoshVanL:
Yep, the `xmin` part was the first thing I ran into when running the CockroachDB tests against the postgresql implementation.

I'm in favor of having a shared internal implementation with things like `xmin`/`etag` being component specific. Let's do this in a followup PR, and just have lots of duplicated code for the moment.

state/cockroachdb/cockroachdb_access.go
JoshVanL: Done

state/cockroachdb/cockroachdb_access.go
JoshVanL: Done.

state/cockroachdb/cockroachdb_access.go
JoshVanL: Done.

ItalyPaleAle (approved):

PR #2631, opened 2023-03-04
[release-1.10] Backport OAuth middlewares fixes into 1.10 [DO NOT SQUASH]

ItalyPaleAle:
Backports #2616 and #2629 into release-1.10

### DIFF
--- a/middleware/http/oauth2/oauth2_middleware.go
+++ b/middleware/http/oauth2/oauth2_middleware.go
@@ -43,7 +43,9 @@ type oAuth2MiddlewareMetadata struct {
 
 // NewOAuth2Middleware returns a new oAuth2 middleware.
 func NewOAuth2Middleware(log logger.Logger) middleware.Middleware {
-	return &Middleware{logger: log}
+	m := &Middleware{logger: log}
+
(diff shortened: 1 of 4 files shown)


berndverst:
ok to include in a hotfix from my perspective due to security concerns (and we did include a workflow [alpha api] change in runtime as well...)

Thoughts @yaron2 @artursouza ?

berndverst (approved):

ItalyPaleAle:
> ok to include in a hotfix from my perspective due to security concerns

Note this won't address any security concern. We will address those for 1.11 (and address other issues that were raised about the components, such as that they are keeping the JWTs in an in-memory cache that is not shared across multiple processes). I made an attempt to see if we could chain failures from the Bearer MW, but it's not possible to catch 401 status codes issued by that MW and change them, as by the time they hit the OAuth2 middleware they have been flushed to the client already.

However, these commits are recommended to be included because the 2 middlewares are currently completely broken in 1.10 for the scenario they are designed for. Meaning, the MWs are currently unusable.

artursouza (approved):

PR #2632, opened 2023-03-04
Consolidate CockroachDB and PostgreSQL into single internal implementation

JoshVanL:
Based on the comment [here](https://github.com/dapr/components-contrib/pull/2618#issuecomment-1452702942).

This PR consolidates the implementations for CockroachDB and PostrgreSQL into a single internal package.

The main difference between the two are 1. how table schemas are created/columns, and 2. postgresql makes use of the internal `xmin` column for etag, where cockroach uses a custom column. To account for this, each package will pass a `EnsureTable` and `SetQuery` function with the database specific implementation.

I have used PostgreSQL as the base implementation to both, and have attempted to minimize the changes made to the logic. You'll notice that certification tests are unchanged, and only type + func signatures are updated in integration tests.

With this change, CockroachDB gets feature parity with PostgreSQL for free.

Docs PR: TBD


state/postgresql/migrations.go
```diff
@@ -23,27 +23,33 @@ import (
...
+
 	// Use an advisory lock (with an arbitrary number) to ensure that no one else is performing migrations at the same time
 	// This is the only way to also ensure we are not running multiple "CREATE TABLE IF NOT EXISTS" at the exact same time
 	// See: https://www.postgresql.org/message-id/CA+TgmoZAdYVtwBfp1FL2sMZbiHCWT4UPrzRLNnX1Nb30Ku3-gg@mail.gmail.com
 	const lockID = 42
 
 	// Long timeout here as this query may block
 	queryCtx, cancel := context.WithTimeout(ctx, time.Minute)
-	_, err := m.Conn.Exec(queryCtx, "SELECT pg_advisory_lock($1)", lockID)
+	_, err := tx.Exec(queryCtx, "SELECT pg_advisory_lock($1)", lockID)
```
ItalyPaleAle, line 52: Please do not use a transaction here, but rather the actual connection. As explained in the link above, "CREATE TABLE" statements in Postgres do not respect transactions, so we must use a different kind of locking

internal/component/postgresql/postgresdbaccess.go
```diff
@@ -270,15 +260,15 @@ func (p *PostgresDBAccess) Get(parentCtx context.Context, req *state.GetRequest)
 	var (
 		value    []byte
 		isBinary bool
-		etag     uint32
+		etag     sql.NullInt64
```
ItalyPaleAle, line 263: I think pgx has a specific nullable type: https://pkg.go.dev/gopkg.in/jackc/pgx.v2#hdr-Null_Mapping

internal/component/postgresql/postgresdbaccess.go
```diff
@@ -55,18 +56,25 @@ type PostgresDBAccess struct {
 	metadata postgresMetadataStruct
 	db       pgxPoolConn
 
+	ensureTableFn func(context.Context, pgx.Tx, EnsureTableOptions) error
```
ItalyPaleAle, line 59: Can you rename to "migrateFn" or something that indicates it is more than just ensuring a table?

internal/component/postgresql/postgresdbaccess.go
```diff
@@ -304,13 +299,13 @@ func (p *PostgresDBAccess) Get(parentCtx context.Context, req *state.GetRequest)
 
 		return &state.GetResponse{
 			Data: data,
-			ETag: ptr.Of(strconv.FormatUint(uint64(etag), 10)),
+			ETag: ptr.Of(etagS),
```
ItalyPaleAle, line 302: If `etagS` is an empty string, this should probably be `nil`, right?

internal/component/postgresql/postgresdbaccess.go
```diff
@@ -304,13 +299,13 @@ func (p *PostgresDBAccess) Get(parentCtx context.Context, req *state.GetRequest)
...
 		}, nil
 	}
 
 	return &state.GetResponse{
 		Data: value,
-		ETag: ptr.Of(strconv.FormatUint(uint64(etag), 10)),
+		ETag: ptr.Of(etagS),
```
ItalyPaleAle, line 308: Same as above

internal/component/postgresql/postgresdbaccess.go
```diff
@@ -335,7 +330,7 @@ func (p *PostgresDBAccess) doDelete(parentCtx context.Context, db dbquerier, req
 			return state.NewETagError(state.ETagInvalid, err)
 		}
 
-		result, err = db.Exec(parentCtx, "DELETE FROM state WHERE key = $1 AND xmin = $2", req.Key, uint32(etag64))
+		result, err = db.Exec(parentCtx, fmt.Sprintf("DELETE FROM state WHERE key = $1 AND %s = $2", p.etagColumn), req.Key, uint32(etag64))
```
ItalyPaleAle, line 333: In hot paths we prefer to use string concatenation rather than fmt.Sprintf for simple concatenation operations

state/cockroachdb/cockroachdb.go
```diff
@@ -15,111 +15,111 @@ package cockroachdb
...
-		dbaccess: dba,
+	return postgresql.NewPostgreSQLStateStore(logger, postgresql.Options{
+		ETagColumn:    "etag",
+		EnsureTableFn: ensureTables,
+		SetQueryFn: func(req *state.SetRequest, opts postgresql.SetQueryOptions) string {
+			// Sprintf is required for table name because sql.DB does not substitute
+			// parameters for table names.
+			// Other parameters use sql.DB parameter substitution.
+			if req.ETag == nil || *req.ETag == "" {
+				return fmt.Sprintf(`
```
ItalyPaleAle, line 37: As mentioned, consider using string concatenation rather than fmt.Sprintf for concatenating strings **in hot paths**. The perf difference is really significant.

state/cockroachdb/cockroachdb.go
```diff
@@ -15,111 +15,111 @@ package cockroachdb
...
+			}
+
+			// When an etag is provided do an update - no insert.
+			return fmt.Sprintf(`
+UPDATE %[1]s
+SET
+  value = $2,
+  isbinary = $3,
+  updatedate = NOW(),
+  etag = etag + 1,
```
ItalyPaleAle, line 58: I like this solution, simple yet elegant :)

ItalyPaleAle (changes requested):
Thanks for starting on this, it was another large effort!

internal/component/postgresql/postgresdbaccess.go
JoshVanL: Thanks, makes sense

JoshVanL:
Thanks @ItalyPaleAle, PTAL

I think it might still be worth it to keep the cert tests to cover the `postgresql.Options` (`ETagColumn`, `MigrationFn`, and `SetQueryFn`).

ItalyPaleAle:
Thanks for the changes! Please fix the linter issues then LGTM

PR #2633, opened 2023-03-04
Consolidate SQL expired date garbage collector

JoshVanL:
Based on #2632 which should be reviewed and merged first.

PR adds a single internal package for handling the garbage collection of SQL data which has a TTL.

Package is used by `sqlite`, `CockroachDB`, and `PostgreSQL`.


ItalyPaleAle:
Thanks, always welcoming PRs that unify code :)

Marking as draft until #2632 is merged

internal/component/sql/cleanup.go
```diff
@@ -0,0 +1,241 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to start transaction: %w", err)
+		}
+		defer txwc.Rollback()
+	}
+
+	// Check if the last iteration was too recent
+	// This performs an atomic operation, so allows coordination with other daprd processes too
+	// We do this before beginning the transaction
+	canContinue, err := g.updateLastCleanup(ctx, tx, txwc)
```
ItalyPaleAle, line 158: This needs to be called with the DB connection (e.g. `g.db`) and not a transaction, before the transaction lock is acquired.

The reason is that the cleanup operation (so the transaction) can take a long time, so we want to make sure that other daprd processes aren't blocked waiting on the transaction and can hit this "semaphore" before. `updateLastCleanup` is atomic, so there's no risk there.

ItalyPaleAle: (I changed it already)

internal/component/sql/cleanup.go
```diff
@@ -34,29 +35,43 @@ type GarbageCollector interface {
...
-	SQLDB                    SQLDB
-	SQLDBWithContext         SQLDBWithContext
+
+	// Interval to perfm the cleanup.
+	CleanupInterval time.Duration
+
+	// Database connection when using pgx.
+	DBPgx pgxConn
+	// Database connection when using database/sql.
+	DBSql sqlConn
```
JoshVanL, line 53: These types should be exported if part of the caller's `Options`.

internal/component/sql/cleanup.go
```diff
@@ -34,29 +35,43 @@ type GarbageCollector interface {
...
 }
 
-type SQLDB interface {
+// Interface for connections that use pgx.
+type pgxConn interface {
 	Begin(context.Context) (pgx.Tx, error)
+	Exec(context.Context, string, ...interface{}) (pgconn.CommandTag, error)
```
JoshVanL, line 59:
```suggestion
// Interface for connections that use pgx.
type pgxConn interface {
	Begin(context.Context) (pgx.Tx, error)
	Exec(context.Context, string, ...any) (pgconn.CommandTag, error)
```

JoshVanL:
Thanks @ItalyPaleAle, LGTM! :+1:

PR #2637, opened 2023-03-06
Remove now-unused metadata properties

ItalyPaleAle:
Completes the work from #2596 now that the runtime has been updated

Updates the pinned runtime in tests too

### DIFF
--- a/nameresolution/metadata.go
+++ b/nameresolution/metadata.go
@@ -16,17 +16,6 @@ package nameresolution
 import "github.com/dapr/components-contrib/metadata"
 
 const (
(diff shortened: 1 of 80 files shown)


nameresolution/metadata.go
```diff
@@ -16,17 +16,6 @@ package nameresolution
 import "github.com/dapr/components-contrib/metadata"
 
 const (
-	// TODO: REMOVE THESE AFTER RUNTIME IS CHANGED
```
ItalyPaleAle, line 19: Reviewers: this is the only "real" change

berndverst (approved):

PR #2641, opened 2023-03-07
Add MySQL TTL and Cleanup

DeepanshuA:
# Description

Adding MySQL TTL and cleanup.
It basically mimics what has been done so far for PostgreSQL and CockroachDB.
Uses cleanup code from https://github.com/dapr/components-contrib/pull/2633.

## Issue reference

Please reference the issue this PR is linked to: #2288 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


ItalyPaleAle:
Thanks @DeepanshuA , let's wait for #2633. Hoping to get that merged in the next couple of days :)

JoshVanL:
@DeepanshuA @ItalyPaleAle Just pushed up a pass at adding ttl support for mysql. Feel free to use some/none of that :slightly_smiling_face: 

https://github.com/dapr/components-contrib/commit/dde93ce09cb4813b45564fabcd08bd557342f3b3

ItalyPaleAle:
@DeepanshuA #2633 has been merged, can you use that in this PR?

DeepanshuA:
> @DeepanshuA #2633 has been merged, can you use that in this PR?

Yeah, I will do that today.

DeepanshuA:
> @DeepanshuA @ItalyPaleAle Just pushed up a pass at adding ttl support for mysql. Feel free to use some/none of that 🙂
> 
> This is based on the [cleanup package PR](https://github.com/dapr/components-contrib/pull/2633).
> 
> [dde93ce](https://github.com/dapr/components-contrib/commit/dde93ce09cb4813b45564fabcd08bd557342f3b3)

Oops, looks like we worked almost concurrently on this one. Sorry.
I am using your change to verify if I have missed something. Thanks for posting that.

JoshVanL:
@DeepanshuA, no problem! Just a heads up that I'm going to be working on `sqlserver` next, which should be the last SQL server to implement TTL using the garbage collector.

DeepanshuA:
> @DeepanshuA, no problem! Just a heads up that I'm going to be working on `sqlserver` next, which should be the last SQL server to implement TTL using the garbage collector.

Sure @JoshVanL, please go ahead!

DeepanshuA:
Sorry, I was out sick for 2-3 days, and couldn't fix the issue in cert test. I will try to find sometime for this one in next 2-3 days, fix that up and open it for Review. Hope that is fine.

state/mysql/mysql.go
```diff
@@ -231,7 +267,35 @@ func (m *MySQL) finishInit(ctx context.Context, db *sql.DB) error {
...
+	if err = m.ensureStateTable(ctx, m.schemaName, m.tableName); err != nil {
+		return err
+	}
+
+	if err = m.ensureMetadataTable(ctx, m.schemaName, m.metadataTableName); err != nil {
+		return err
+	}
+
+	if m.cleanupInterval != nil {
+		gc, err := sqlCleanup.ScheduleGarbageCollector(sqlCleanup.GCOptions{
```
ItalyPaleAle, line 279: To my understanding, this may not work with how the GC works.

Per a comment here: https://stackoverflow.com/a/16588699

> This is leaving the fields the same, but it's still doing an update either way.

The problem is that the GC library expects UpdateLastCleanupQuery to return `rowsAffected == 0` if the update isn't needed. According to docs, this one always updates the database, so I presume that `rowsAffected == 1`, which would cause the cleanup to run every time.

Perhaps this should use a stored procedure instead? Or something else

state/mysql/mysql.go
```diff
@@ -171,6 +190,23 @@ func (m *MySQL) parseMetadata(md map[string]string) error {
 	}
 	m.connectionString = meta.ConnectionString
 
+	// Cleanup interval
+	m.cleanupInterval = meta.CleanupInterval
+	s, ok := md[cleanupIntervalKey]
+	if ok && s != "" {
```
ItalyPaleAle, line 196: Nit:

```suggestion
	if s := md[cleanupIntervalKey]; s != "" {
```

> The reason why we have an `ok` as second argument is because reading a key that doesn't exist always returns the zero value for the type, which for a string is the empty string. This make it impossible to know if the value is defined but empty, or not present. In this case, we just care if there's a value and we treat undefined and defined but empty the same.

state/mysql/mysql.go
```diff
@@ -305,6 +371,44 @@ func (m *MySQL) ensureStateTable(ctx context.Context, stateTableName string) err
 		}
 	}
 
+	// Check if expiredate column exists - to cater cases when table was created before v1.11.
+	columnExists, err := columnExists(ctx, m.db, schemaName, stateTableName, "expiredate", m.timeout)
```
ItalyPaleAle, line 375: there's a potential for a race condition here if 2 sidecars are starting up at the same time.

Because you check for the column's existence and then after that you add the column, it's possible that 2 sicecars starting up at the same time will both detect the column doesn't exist, and will both try to create that.

In the Postgres component, we solve that by acquiring a lock (not using transactions because, at least in Postgres, table creation doesn't happen in a transaction). I don't know if MySQL (and MariaDB) support something like that.

Worst case, I guess, one of the two sidecars will fail to initialize and will need to be restarted? This should only happen during upgrades from <1.11 anyways

state/mysql/mysql.go
```diff
@@ -305,6 +371,44 @@ func (m *MySQL) ensureStateTable(ctx context.Context, stateTableName string) err
...
+func (m *MySQL) ensureMetadataTable(ctx context.Context, schemaName, metaTableName string) error {
+	exists, err := tableExists(ctx, m.db, schemaName, metaTableName, m.timeout)
+	if err != nil {
+		return err
+	}
+
+	if !exists {
+		m.logger.Info("Creating MySQL metadata table")
+		_, err = m.db.ExecContext(ctx, fmt.Sprintf(`CREATE TABLE IF NOT EXISTS %s (
+			id VARCHAR(255) NOT NULL PRIMARY KEY, value VARCHAR(255) NOT NULL);`, metaTableName))
```
ItalyPaleAle, line 406: For `value`, consider using `TEXT` since we may need more than 255 chars in the future. (It's probably ok for the key)

state/mysql/mysql.go
```diff
@@ -305,6 +371,44 @@ func (m *MySQL) ensureStateTable(ctx context.Context, stateTableName string) err
...
+
+func (m *MySQL) ensureMetadataTable(ctx context.Context, schemaName, metaTableName string) error {
+	exists, err := tableExists(ctx, m.db, schemaName, metaTableName, m.timeout)
+	if err != nil {
+		return err
+	}
+
+	if !exists {
+		m.logger.Info("Creating MySQL metadata table")
+		_, err = m.db.ExecContext(ctx, fmt.Sprintf(`CREATE TABLE IF NOT EXISTS %s (
```
ItalyPaleAle, line 405: Nit: I would prefer name `key` rather than `id` for the primary key, given this is a "key-value".

state/mysql/mysql.go
```diff
@@ -528,28 +664,29 @@ func (m *MySQL) setValue(parentCtx context.Context, querier querier, req *state.
...
+		params = []any{enc, eTag, isBinary, req.Key, *req.ETag}
 	} else {
 		// If this is a duplicate MySQL returns that two rows affected
 		maxRows = 2
-		query := fmt.Sprintf(
-			`INSERT INTO %s (value, id, eTag, isbinary) VALUES (?, ?, ?, ?) on duplicate key update value=?, eTag=?, isbinary=?;`,
-			m.tableName, // m.tableName is sanitized
-		)
-		result, err = querier.ExecContext(ctx, query, enc, req.Key, eTag, isBinary, enc, eTag, isBinary)
+		query = `INSERT INTO %s (value, id, eTag, isbinary, expiredate) VALUES (?, ?, ?, ?, %[2]s)
```
ItalyPaleAle, line 677:
```suggestion
		query = `INSERT INTO %[1]s (value, id, eTag, isbinary, expiredate) VALUES (?, ?, ?, ?, %[2]s) 
```

state/mysql/mysql.go
```diff
@@ -528,28 +664,29 @@ func (m *MySQL) setValue(parentCtx context.Context, querier querier, req *state.
...
+		query = `INSERT INTO %[1]s (value, id, eTag, isbinary, expiredate) VALUES (?, ?, ?, ?, %[2]s)`
+		params = []any{enc, req.Key, eTag, isBinary}
 	} else if req.ETag != nil && *req.ETag != "" {
 		// When an eTag is provided do an update - not insert
-		query := fmt.Sprintf(
-			`UPDATE %s SET value = ?, eTag = ?, isbinary = ? WHERE id = ? AND eTag = ?;`,
-			m.tableName, // m.tableName is sanitized
-		)
-		result, err = querier.ExecContext(ctx, query, enc, eTag, isBinary, req.Key, *req.ETag)
+		query = `UPDATE %[1]s SET value = ?, eTag = ?, isbinary = ?, expiredate = %[2]s
```
ItalyPaleAle, line 671: What happens if we try to update a row that is expired but not yet garbage-collected? In that case the ETag check should fail? Or...?

ItalyPaleAle (changes requested):

state/mysql/mysql.go
ItalyPaleAle: See: https://github.com/dapr/components-contrib/pull/2688

state/mysql/mysql.go
DeepanshuA: Name `key` is not allowed. I started off with `key` only but it gives an error.

state/mysql/mysql.go
DeepanshuA: Yup, the two different logs above are for mysql and maridb.
Just, for re-confirming, I added connection string as well to the log:
For mysql:
```
{"Time":"2023-03-23T06:59:23.453716692Z","Action":"output","Package":"github.com/dapr/components-contrib/tests/certification/state/mysql","Test":"TestMySQL/cleanup/cleanup_concurrency","Output":"time=\"2023-03-23T06:59:23.453657992Z\" level=info msg=\"Testing CleanupExpired - no rows should be affected, %sroot:root@tcp(localhost:3306)/?allowNativePasswords=true\" instance=fv-az365-158 scope=dapr.components type=log ver=unknown\n"}
{"Time":"2023-03-23T06:59:23.454251799Z","Action":"output","Package":"github.com/dapr/components-contrib/tests/certification/state/mysql","Test":"TestMySQL/cleanup/cleanup_concurrency","Output":"time=\"2023-03-23T06:59:23.454176798Z\" level=info msg=\"=========================Rows affected: 0\" instance=fv-az365-158 scope=dapr.components type=log ver=unknown\n"}
```
For mariadb:
```
{"Time":"2023-03-23T06:59:31.507769379Z","Action":"output","Package":"github.com/dapr/components-contrib/tests/certification/state/mysql","Test":"TestMySQL/cleanup#01/cleanup_concurrency","Output":"time=\"2023-03-23T06:59:31.507718579Z\" level=info msg=\"Testing CleanupExpired - no rows should be affected, %sroot:root@tcp(localhost:3307)/\" instance=fv-az365-158 scope=dapr.components type=log ver=unknown\n"}
{"Time":"2023-03-23T06:59:31.508253385Z","Action":"output","Package":"github.com/dapr/components-contrib/tests/certification/state/mysql","Test":"TestMySQL/cleanup#01/cleanup_concurrency","Output":"time=\"2023-03-23T06:59:31.508174984Z\" level=info msg=\"=========================Rows affected: 0\" instance=fv-az365-158 scope=dapr.components type=log ver=unknown\n"}
```

state/mysql/mysql.go
DeepanshuA: Done

state/mysql/mysql.go
DeepanshuA: I have added `IF NOT EXISTS` to column and index addition - that way, in case of such a race condition, only first one will be able to add column and index and other one will NOT fail, it will simply skip past. Hope it suffices.

state/mysql/mysql.go
DeepanshuA: I have changed that to `TEXT` for Value. But, just wanted to confirm one thing. Currently, we are storing last-cleanup timestamp in the value, do you foresee any particular use case for metadata table here?

state/mysql/mysql.go
DeepanshuA: Done

state/mysql/mysql.go
ItalyPaleAle: That should do it, thanks

state/mysql/mysql.go
```diff
@@ -528,28 +663,29 @@ func (m *MySQL) setValue(parentCtx context.Context, querier querier, req *state.
 
 	if req.Options.Concurrency == state.FirstWrite && (req.ETag == nil || *req.ETag == "") {
 		// With first-write-wins and no etag, we can insert the row only if it doesn't exist
-		query := fmt.Sprintf(
-			`INSERT INTO %s (value, id, eTag, isbinary) VALUES (?, ?, ?, ?);`,
-			m.tableName, // m.tableName is sanitized
-		)
-		result, err = querier.ExecContext(ctx, query, enc, req.Key, eTag, isBinary)
+		query = `INSERT INTO %[1]s (value, id, eTag, isbinary, expiredate) VALUES (?, ?, ?, ?, %[2]s)`
```
ItalyPaleAle, line 666: Just realized this could have the same issues with rows that are expired (but not yet GC'd).

If the row is expired, a first-write-wins should work. In this case, it would fail.

ItalyPaleAle (dismissed):

state/mysql/mysql.go
ItalyPaleAle: I found a fix - need this merged soon :)

ItalyPaleAle:
@berndverst since I made the last commits, I'd like to get you to approve this please!

berndverst (approved):

PR #2659, opened 2023-03-10
Add `connectionString` metadata option for MongoDB

berndverst:
Instead of crafting from various metadata fields the MongoDB connection string, this adds a `connectionString` metadata option that allows using a complete connection string. If `connectionString` is specified, no other connection metadata like `host` or `server` are used.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


state/mongodb/mongodb.go
```diff
@@ -416,6 +417,10 @@ func (m *MongoDB) Query(ctx context.Context, req *state.QueryRequest) (*state.Qu
 }
 
 func getMongoURI(metadata *mongoDBMetadata) string {
```
ItalyPaleAle, line 419: Can you rename to `getMongoDBConnString`?

ItalyPaleAle (dismissed):

state/mongodb/mongodb.go
berndverst: Use the suggestion feature so I can apply it in the future?

ItalyPaleAle (approved):

Issue #2661, opened 2023-03-10
GetDel support for Redis Bindings


Vishal07051999:
We have started using Dapr in our company and, we were trying to use it for one of our use case where we need GetDel of Redis.

Have created a implementation for the same : https://github.com/dapr/components-contrib/pull/2662

berndverst:
Please do not start a PR immediately after creating an issue. It is best to first have some discussion as your PR (or the enhancement) might get rejected and get maintainers to voice their support on the issue.

Personally I do not see the value in `GetDel`. You can already use the `Get` operation and the `Del` operation. If necessary we can add ETAG support which we already have in other parts of Dapr.

Could you please share more about your use case in your company and why using `Get` and `Delete` operations today do not suffice?

Vishal07051999:
GetDel help us in making keys one time retrival only even in case of consurrent program.

GetDel is a very frequesnt usecase when you need to create a Payment OTP or, Login OTP as it prevent concurrent calls from succeding as only one call gets data.

I agree using metadata with delete set as true will work and is better as it will simplify API surface area.

Will make a change in same PR.

Thanks for the suggestion.

Vishal07051999:
Updated PR with all requested changes

roozbehid-ic:
I am also upvoting this feature. I actually came here to check if this is supported or not. Atomicity of it, is what matters that cannot be done with two seperate API calls of GET and DEL

berndverst:
@Vishal07051999 please assign the issue https://github.com/dapr/docs/issues/3258 to yourself to complete the documentation for your implementation.

Vishal07051999:
Unable to assign or edit issue create.

PR #2662, opened 2023-03-10
Adding GetDel support for Redis Bindings

Vishal07051999:
# Description

Get Del Support for Redis Bindings

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/components-contrib/issues/2661

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


berndverst:
As discussed on the issue, it is unclear whether we want to accept this feature. You opened a PR immediately after opening an issue - that is a risky move.

We will have to decide first whether we want to add this feature.

Now from a PR perspective, this PR would also need a test covering this new operation in `tests/certification/binding/redis`

berndverst (changes requested):
Please provide business / architecture background on the issue to explain the need for this.

If we decide to proceed, then I suggest that we do not make this a new operation, instead we add a new metadata property for the `Get` operation `delete: true`. Only if the Get operation is invoked and the delete metadata property is set and true, then under the hood you would invoke the `GetDel` function. Does that make sense?

Having too many operations is not good for users or maintainers, and we do not want to be just a SDK passthrough tool. We can choose an opinionated approach here :)

Oh and by the way as I mentioned in another comment, you also must update the certification tests at `tests/certification/binding/redis`

Vishal07051999:
> Please provide business / architecture background on the issue to explain the need for this.
> 
> If we decide to proceed, then I suggest that we do not make this a new operation, instead we add a new metadata property for the `Get` operation `delete: true`. Only if the Get operation is invoked and the delete metadata property is set and true, then under the hood you would invoke the `GetDel` function. Does that make sense?
> 
> Having too many operations is not good for users or maintainers, and we do not want to be just a SDK passthrough tool. We can choose an opinionated approach here :)
> 
> Oh and by the way as I mentioned in another comment, you also must update the certification tests at `tests/certification/binding/redis`

Updated Certification Tests.

Also, agree with the point of delete being a MetaData Parameter instead of a separate operation. Under the hood, it will use `Redis GetDel Operation when delete flag is 'true' in metadata`.

Vishal07051999:
Lint Issues fixed.

yaron2:
> Lint Issues fixed.

One more linter issue: https://github.com/dapr/components-contrib/actions/runs/4410063633/jobs/7727120068.

berndverst (approved):

berndverst:
I fixed the linter error for you to get this merged.

Vishal07051999:
Thanks @yaron2 .

PR #2543, opened 2023-02-17
Merge Certification Test and Conformance test code coverage

shivanisinghnitp:
# Description

_Please explain the changes you've made_

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #_[issue number]_

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #2681, opened 2023-03-17
Merge conformance and certification coverage report

shivamkm07:
# Description

This PR carries forward the changes done in https://github.com/dapr/components-contrib/pull/2543. It helps publishing coverage reports for conformance test and then merges the certification and conformance test coverage report to publish it to our discord monitoring channel. 

The sample run in my fork: https://github.com/shivamkm07/components-contrib/actions/runs/4444112664

The discord notification looks like this: 
image

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


.github/scripts/test-info.mjs
```diff
@@ -67,11 +70,13 @@ const components = {
             'AzureIotHubName',
             'AzureIotHubBindingsConsumerGroup',
         ],
+        sourcePkg: 'azure/eventhubs',
     },
     'bindings.azure.servicebusqueues': {
         conformance: true,
         certification: true,
         requiredSecrets: ['AzureServiceBusConnectionString'],
+        sourcePkg: 'azure/servicebusqueues',
```
ItalyPaleAle, line 79: A lot of components, like servicebus, redis, etc, have a lot of code in a shared package. It would not be a fair calculation of code coverage if you didn't include those.

Please, allow passing multiple `sourcePkg` (type would be`string | string[]` to allow for an array too) and consider those too.

ItalyPaleAle (changes requested):

.github/scripts/test-info.mjs
shivamkm07: The shared packages in `internal` folder are already taken into account. The `sourcePkg` is just the relative path of the component from the parent folder. 
See `-coverpkg=github.com/dapr/components-contrib/internal/component/${{ matrix.source-pkg }}/...,github.com/dapr/components-contrib/${SRC_PKG}/...`

.github/workflows/conformance.yml
```diff
@@ -242,10 +255,17 @@ jobs:
...
+        else
+          SRC_PKG=${KIND}/${NAME}  
+        fi
+
         set +e
         gotestsum --jsonfile ${{ env.TEST_OUTPUT_FILE_PREFIX }}_conformance.json \
           --junitfile ${{ env.TEST_OUTPUT_FILE_PREFIX }}_conformance.xml --format standard-verbose -- \
-          -p 2 -count=1 -timeout=15m -tags=conftests ./tests/conformance --run="Test${KIND_UPPER}Conformance/${NAME}"
+          -p 2 -count=1 -timeout=15m -tags=conftests ./tests/conformance --run="Test${KIND_UPPER}Conformance/${NAME}" -coverprofile=cover.out \
+          -covermode=set -coverpkg=github.com/dapr/components-contrib/internal/component/${{ matrix.source-pkg }}/...,github.com/dapr/components-contrib/${SRC_PKG}/...
```
shivamkm07, line 268: This is where `sourcePkg` is being used

.github/workflows/coverage-reports.yml
```diff
@@ -0,0 +1,118 @@
...
+# ------------------------------------------------------------
+
+name: Publishing test coverage reports of components
+
+on:
+  repository_dispatch:
+    types: [coverage-report]
+  workflow_dispatch:
+  schedule:
+    - cron: '0 */24 * * *'
```
DeepanshuA, line 21: Instead of every 24 hours i.e. daily, we may want this report every 3 or 7 days, something like: "0 3 * * 2,5" (at 3 A.M. on Tuesday and Friday).

.github/workflows/conformance.yml
ItalyPaleAle: But in this case, `internal/component` may reference a folder that doesn’t exist. For example, if `matrix.source-pkg` is `azure/servicebusqueues` as in the example above, that folder doesn’t exist.

.github/workflows/coverage-reports.yml
rabollin: I too agree that once a week is good enough.

.github/workflows/conformance.yml
DeepanshuA: Yeah, if a folder doesn't exist, that folder would be ignored in that case. And, coverage will factor in only present folders. Do you think there can be any problem in that case?

ItalyPaleAle: No concerns with including additional folders, but the current code is not including all folders.

As per https://github.com/dapr/components-contrib/pull/2681#discussion_r1139750172 I would recommend making `sourcePkg` of type `string | string[]` to allow passing an array too, and then include the correct folders. `sourcePkg` could be an absolute package such as `github.com/dapr/components-contrib/foo`, since components may have shared code outside of `github.com/dapr/components-contrib/internal`

shivamkm07: That makes sense, Will update the `sourcePkg` to array of strings

.github/scripts/test-info.mjs
shivamkm07: Done

.github/workflows/coverage-reports.yml
shivamkm07: Done

.github/scripts/test-info.mjs
```diff
@@ -86,95 +99,136 @@ const components = {
...
         ],
-        sourcePkg: 'azure/storagequeues',
+        sourcePkg: [
+            'github.com/dapr/components-contrib/bindings/azure/storagequeues',
+        ],
     },
     'bindings.cron': {
         conformance: true,
         certification: true,
+        sourcePkg: ['github.com/dapr/components-contrib/bindings/cron'],
```
ItalyPaleAle, line 109: Can we make this optional, or auto-generated, when the component uses only "standard" paths? This will make it easier to maintain.

The GenerateMatrix function below should be able to generate `github.com/dapr/components-contrib/bindings/cron` from `bindings.cron` without the user needing to type this.

ItalyPaleAle (changes requested):

.github/scripts/test-info.mjs
shivamkm07: Please see the comment below

.github/workflows/conformance.yml
```diff
@@ -242,10 +255,18 @@ jobs:
 
         echo "Running tests for Test${KIND_UPPER}Conformance/${KIND}/${NAME} ... "
 
+        if [ ${{ matrix.source-pkg }} != '' ]; then
+          SRC_PKG=${{ matrix.source-pkg }}
+        else
+          SRC_PKG=github.com/dapr/components-contrib/${KIND}/${NAME}
```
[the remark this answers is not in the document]
ItalyPaleAle: Given how "standard" doesn't include `bindings.azure.cosmosdb`, I would recommend instead:

1. In the YAML file, only set `SRC_PKG=${{ matrix.source-pkg }}` (without `SRC_PKG=github.com/dapr/components-contrib/${KIND}/${NAME}`)
2. Change the `test-info.mjs` file to always generate `source-pkg`, taking into account all possible "standard" paths (so also `bindings.azure.cosmosdb`: https://github.com/dapr/components-contrib/pull/2681/files/ef1340d6b31a41ea6bfa654d223dfaec7bd6a487..2a91f9bf3075050fd268deaa6fd450fb845738b2#diff-3a6fa7800d7ed529cdd84fb582d261df6e24170634ce6cc14fd884fc2d57ad7aR757-R759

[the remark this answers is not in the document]
shivamkm07: Done

.github/workflows/coverage-reports.yml
```diff
@@ -0,0 +1,118 @@
...
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * 0'
+
+jobs:
+  coverage:
+      name: Download and merge gocov reports
+      runs-on: ubuntu-latest
+      env:
+        GOVER: "1.20"
```
ItalyPaleAle, line 28: Please don't hardcode the version of Go. Instead, read it from the `go.mod` file. See example: https://github.com/dapr/components-contrib/blob/33f9c6ecef9f8f5065b6fd417e80f4a17e48a357/.github/workflows/conformance.yml#L202-L205

.github/workflows/coverage-reports.yml
```diff
@@ -0,0 +1,118 @@
...
+        - name: Check out code
+          uses: actions/checkout@v3
+          with:
+            repository: ${{ env.CHECKOUT_REPO }}
+            ref: ${{ env.CHECKOUT_REF }}
+
+        - name: Setup go dependencies
+          run: |
+            go install github.com/axw/gocov/gocov@v1.1.0
+            go install gotest.tools/gotestsum@latest
```
ItalyPaleAle, line 45: Should we pin a version for gotestsum and gocovmerge? Also for security reasons

.github/workflows/conformance.yml
```diff
@@ -255,6 +270,7 @@ jobs:
         fi
         set -e
 
+        cat cover.out
```
ItalyPaleAle, line 273: Is this needed? May create a lot of noise

ItalyPaleAle (changes requested):

.github/workflows/coverage-reports.yml
shivamkm07: Done

.github/workflows/coverage-reports.yml
shivamkm07: Done

.github/workflows/conformance.yml
```diff
@@ -217,7 +228,9 @@ jobs:
     - name: Download Go dependencies
       run: |
         go mod download
+        go install github.com/axw/gocov/gocov@v1.1.0
```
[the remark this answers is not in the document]
shivamkm07: Done

.github/workflows/conformance.yml
shivamkm07: Removed

ItalyPaleAle:
Looks like the SQL Server cert tests are failing, but seems unrelated from this PR. So, LGTM.

ItalyPaleAle (approved):

PR #2682, opened 2023-03-18
Adds TTL support to SQL Server using SQL Cleanup Strategy

JoshVanL:
Adds TTL support to the SQL Server state store using the dapr SQL cleanup strategy.

Update Upset procedure name to `sp_Upsert_v3`.

Part of https://github.com/dapr/dapr/issues/5899 and https://github.com/dapr/components-contrib/issues/2288

Docs PR: https://github.com/dapr/docs/pull/3277


state/sqlserver/migration.go
```diff
@@ -108,23 +109,23 @@ func (m *migration) executeMigrations() (migrationResult, error) {
 		defer db.Close()
 	}
 
-	err = m.ensureSchemaExists(db)
+	err = m.ensureSchemaExists(ctx, db)
```
ItalyPaleAle, line 112: Is there a way all these `ensure` statements can be put in a lock? Otherwise there could be race conditions when multiple sidecars are starting up at the same time.

state/sqlserver/migration.go
```diff
@@ -159,38 +160,39 @@ func (m *migration) ensureIndexedPropertyExists(ix IndexedProperty, db *sql.DB)
...
-func (m *migration) ensureTableExists(db *sql.DB, r migrationResult) error {
+func (m *migration) ensureTableExists(ctx context.Context, db *sql.DB, r migrationResult) error {
 	tsql := fmt.Sprintf(`
 	IF NOT EXISTS (SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = '%s' AND TABLE_NAME = '%s')
     	CREATE TABLE [%s].[%s] (
 			[Key] 			%s CONSTRAINT PK_%s PRIMARY KEY,
 			[Data]			NVARCHAR(MAX) NOT NULL,
-			[InsertDate] 	DateTime2 NOT NULL DEFAULT(GETDATE()),
-			[UpdateDate] 	DateTime2 NULL,`,
+			[InsertDate] 	DateTime2 NOT NULL DEFAULT(CURRENT_TIMESTAMP),
```
ItalyPaleAle, line 193: Is this a breaking change? `GETDATE()` to `CURRENT_TIMESTAMP`?

state/sqlserver/migration.go
ItalyPaleAle: Uhm looks like most (all) statements perform atomic queries with `IF NOT EXIST`, so it's probably ok.

state/sqlserver/migration.go
```diff
@@ -273,92 +299,88 @@ func (m *migration) createStoredProcedureIfNotExists(db *sql.DB, name string, es
...
 }
 
 /* #nosec. */
 //nolint:dupword
-func (m *migration) ensureUpsertStoredProcedureExists(db *sql.DB, mr migrationResult) error {
+func (m *migration) ensureUpsertStoredProcedureExists(ctx context.Context, db *sql.DB, mr migrationResult) error {
 	tsql := fmt.Sprintf(`
-			CREATE PROCEDURE %s (
-				@Key 			%s,
+			CREATE PROCEDURE %[1]s (
```
ItalyPaleAle, line 309: What happens if the sproc already exists for 1.10 so this needs to be updated?

state/sqlserver/migration.go
```diff
@@ -273,92 +299,88 @@ func (m *migration) createStoredProcedureIfNotExists(db *sql.DB, name string, es
...
-								IF NOT EXISTS (SELECT * FROM [%s] WHERE [KEY]=@KEY AND RowVersion = @RowVersion)
+								IF NOT EXISTS (SELECT * FROM [%[3]s] WHERE [KEY]=@KEY AND RowVersion = @RowVersion)
 									BEGIN
 										THROW 2601, ''FIRST-WRITE: COMPETING RECORD ALREADY WRITTEN.'', 1
 									END
 								BEGIN
-									UPDATE [%s]
-									SET [Data]=@Data, UpdateDate=GETDATE()
+									UPDATE [%[3]s]
+									SET [Data]=@Data, UpdateDate=CURRENT_TIMESTAMP, ExpireDate=CASE WHEN @TTL IS NULL THEN NULL ELSE DATEADD(SECOND, @TTL, CURRENT_TIMESTAMP) END
```
ItalyPaleAle, line 328: I think ExpireDate needs to be updated regardless, not just if null?

state/sqlserver/migration.go
```diff
@@ -273,92 +299,88 @@ func (m *migration) createStoredProcedureIfNotExists(db *sql.DB, name string, es
...
 								END TRY
-					
+
 								BEGIN CATCH
 									IF ERROR_NUMBER() IN (2601, 2627)
-										UPDATE [%s]
-										SET [Data]=@Data, UpdateDate=GETDATE()
+										UPDATE [%[3]s]
+										SET [Data]=@Data, UpdateDate=CURRENT_TIMESTAMP, ExpireDate=CASE WHEN @TTL IS NULL THEN NULL ELSE DATEADD(SECOND, @TTL, CURRENT_TIMESTAMP) END
 										WHERE [Key]=@Key AND RowVersion = ISNULL(@RowVersion, RowVersion)
```
ItalyPaleAle, line 374: Probably needs a check here where if the row is expired (TTL is not NULL and is < current time), then it's not updated. Otherwise, this would return true for rows that are expired but not yet GC'd.

state/sqlserver/sqlserver.go
```diff
@@ -65,22 +68,26 @@ const (
...
+	defaultSchema          = "dbo"
+	defaultDatabase        = "dapr"
+	defaultTable           = "state"
+	defaultMetaTable       = "dapr_metadata"
+	defaultCleanupInterval = time.Hour
 )
 
-// NewSQLServerStateStore creates a new instance of a Sql Server transaction store.
-func NewSQLServerStateStore(logger logger.Logger) state.Store {
+// New creates a new instance of a Sql Server transaction store.
```
ItalyPaleAle, line 89:
```suggestion
// New creates a new instance of a SQL Server transaction store.
```

I'm a former Product Marketing Manager, please forgive me :)

state/sqlserver/sqlserver.go
```diff
@@ -228,6 +269,24 @@ func (s *SQLServer) parseMetadata(meta map[string]string) error {
 		return err
 	}
 
+	// Cleanup interval
+	v, ok := meta[cleanupIntervalKey]
+	if ok && v != "" {
```
ItalyPaleAle, line 274:
```suggestion
	if v := meta[cleanupIntervalKey]; v != "" {
```

state/sqlserver/sqlserver.go
```diff
@@ -341,14 +400,24 @@ func (s *SQLServer) setTable(tableName string) error {
...
+
 // Features returns the features available in this state store.
 func (s *SQLServer) Features() []state.Feature {
 	return s.features
 }
 
 // Multi performs multiple updates on a Sql server store.
 func (s *SQLServer) Multi(ctx context.Context, request *state.TransactionalStateRequest) error {
-	tx, err := s.db.Begin()
+	tx, err := s.db.BeginTx(ctx, nil)
```
ItalyPaleAle, line 420: Don't we need a `defer tx.Rollback()` here?

state/sqlserver/sqlserver.go
```diff
@@ -466,7 +535,7 @@ type TvpDeleteTableStringKey struct {
 
 // BulkDelete removes multiple entries from the store.
 func (s *SQLServer) BulkDelete(ctx context.Context, req []state.DeleteRequest) error {
-	tx, err := s.db.Begin()
+	tx, err := s.db.BeginTx(ctx, nil)
```
ItalyPaleAle, line 538: Ditto... Don't we need a `defer tx.Rollback()` here?

state/sqlserver/sqlserver.go
```diff
@@ -606,15 +682,15 @@ func (s *SQLServer) executeSet(ctx context.Context, db dbExecutor, req *state.Se
...
+		return errors.New("no item was updated")
 	}
 
 	return nil
 }
 
 // BulkSet adds/updates multiple entities on store.
 func (s *SQLServer) BulkSet(ctx context.Context, req []state.SetRequest) error {
-	tx, err := s.db.Begin()
+	tx, err := s.db.BeginTx(ctx, nil)
```
ItalyPaleAle, line 693: Ditto

ItalyPaleAle (changes requested):

state/sqlserver/migration.go
JoshVanL: Reverted.

state/sqlserver/migration.go
JoshVanL: Yes, `ExpireDate` is being `SET` to `NULL` when `@TTL` is `NULL`.

state/sqlserver/migration.go
JoshVanL: Is this not the behavior we want? I understood we always SET the value, regardless of what the row currently is?

JoshVanL:
Thanks @ItalyPaleAle, I have applied or responded to your comments. PTAL :slightly_smiling_face:

state/sqlserver/migration.go
ItalyPaleAle: Ok thanks for confirming, I did not understand the T-SQL :)

state/sqlserver/migration.go
ItalyPaleAle: I don't think so. Here's my thinking:

- A row that is expired is to be considered "deleted", even though it may not (yet) have been GC'd.
- This query uses UPDATE which means that it should fail if the row doesn't exist
- If we try to perform an update of a row that is expired, it should then fail (i.e. not update anything).

ItalyPaleAle: You have not misunderstood, but this is a unique case.

If you are passing an Etag, you are telling Dapr to [observe the "first-write-wins" pattern](https://docs.dapr.io/developing-applications/building-blocks/state-management/state-management-overview/#why-dapr-chooses-optimistic-concurrency-control-occ) and you are instructing it to update a record only if the Etag matches.

In case of an expired row, although it's still in the database, it's "soft deleted", so the Etag should never match, and the user should get an error indicating Etag mismatch. With this query, a "soft deleted" record would be updated, however.

(I think the Postgres and SQLite implementation have the same bug, I can look into fixing those)

JoshVanL: Thanks for the explanation @ItalyPaleAle, I understand now!

ItalyPaleAle (approved):

state/sqlserver/migration.go
ItalyPaleAle: Thanks for the change.

I've updated Postgres, Cockroach, SQLite, and Oracle to fix this issue too: https://github.com/dapr/components-contrib/pull/2682#discussion_r1142299949

ItalyPaleAle (changes requested):
The SQL Server tests are failing, can you please take a look?

JoshVanL:
@ItalyPaleAle Looks to be happy

ItalyPaleAle (approved):

PR #2685, opened 2023-03-20
Add output binding wasm

### LINKED ISSUE #2274: New output binding for Wasm
## Describe the feature

We currently have a middleware that allows modifying requests using Wasm.

Since Dapr implements now a Wasm runtime to support that, would it make sense to have the ability to create a binding to allow running code as Wasm?

This would be an output binding with an `invoke` method that allows triggering code from a Wasm file. It would allow turning Dapr into a Wasm runtime easily.

CC: @codefromthecrypt  @daixiang0 @artursouza 

## Release Note

RELEASE NOTE: **ADD** Wasm output binding

codefromthecrypt:
# Description

This adds an output binding for wasm, initially implementing for WASI preview 1. We can later add old WASI (snapshot) or GOOS=js, and in the future new WASI. Also, [soon](https://github.com/golang/go/issues/58141) we can use normal Go to compile as well as TinyGo.

## Issue reference

Closes #2274

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


bindings/wasm/benchmark_test.go
```diff
@@ -0,0 +1,42 @@
...
+	defer output.(io.Closer).Close()
+
+	ctx := context.Background()
+	err := output.Init(ctx, bindings.Metadata{Base: md})
+	if err != nil {
+		b.Fatal(err)
+	}
+
+	request := &bindings.InvokeRequest{
+		Metadata:  map[string]string{"args": "salaboy"},
```
codefromthecrypt, line 32: @salaboy will become immortal

codefromthecrypt:
fyi for bench results. Note they are a *lot* worse if you use stdin instead of args, but I didn't add a dimension about that.

```
goos: darwin
goarch: arm64
pkg: github.com/dapr/components-contrib/bindings/wasm
BenchmarkExample
BenchmarkExample-12    	  101841	     14366 ns/op
PASS
```

bindings/wasm/output.go
```diff
@@ -0,0 +1,192 @@
...
+		return fmt.Errorf("wasm: error compiling binary: %w", err)
+	}
+
+	switch detectImports(out.module.ImportedFunctions()) {
+	case modeWasiP1:
+		_, err = wasi_snapshot_preview1.Instantiate(ctx, out.runtime)
+	}
+
+	if err != nil {
+		return fmt.Errorf("wasm: error instantiating host functions: %w", err)
```
ItalyPaleAle, line 98: Is `_ = out.runtime.Close(ctx)` not needed here?

bindings/wasm/output.go
```diff
@@ -0,0 +1,192 @@
...
+func (out *outputBinding) Close() error {
+	// wazero's runtime closes everything.
+	if rt := out.runtime; rt != nil {
+		return rt.Close(context.Background())
+	}
+	return nil
+}
+
+func (out *outputBinding) getInitMetadata(metadata bindings.Metadata) (*initMetadata, error) {
+	b, err := json.Marshal(metadata.Properties)
```
ItalyPaleAle, line 152: Rather than encoding to JSON and back, please use our `DecodeMetadata` function (example: https://github.com/dapr/components-contrib/blob/76e33137d95f60f182a21505d361ba633afd44ed/state/sqlite/sqlite_metadata.go#L57) which has a bunch of advantages

bindings/wasm/output.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	var stdout bytes.Buffer
+	moduleConfig = moduleConfig.WithStdout(&stdout)
+
+	if args, ok := req.Metadata["args"]; ok {
+		argsSlice := strings.SplitN(args, ",", -1)
+		// prepend arg0 (program name) unless we want users to supply it.
+		argsSlice = append([]string{out.module.Name()}, argsSlice...)
+		moduleConfig = moduleConfig.WithArgs(argsSlice...)
```
ItalyPaleAle, line 120: If `req.Metadata["args"]` is not defined, you don't want to pass `out.module.Name()` as args?

ItalyPaleAle (changes requested):
This is awesome, just some small tweaks please :)

bindings/wasm/output.go
```diff
@@ -0,0 +1,192 @@
...
+
+	// Handle any error, noting exit zero is not one.
+	if exitErr, ok := err.(*sys.ExitError); ok && exitErr.ExitCode() == 0 {
+		return &bindings.InvokeResponse{Data: stdout.Bytes()}, nil
+	}
+	return nil, err
+}
+
+func (out *outputBinding) Operations() []bindings.OperationKind {
+	return []bindings.OperationKind{bindings.GetOperation}
```
yaron2, line 139: Is this really a `Get` operation? we can support any operation name that matches what is being done as closely as possible - perhaps `Execute`?

codefromthecrypt:
Thanks for all the awesome feedback. I will work it in.

I have a question about an important missing detail, which are ENV variables. Most of the time, the only ENV you need are at init scope. e.g. setting TMPDIR to something besides /tmp (though we haven't plugged any persistence in, yet). Some things use them at invoke time, but yeah using them at request time can also be problematic as people can override your init-level settings (unless we forbid that)

So, two questions

1. how should we allow env variables to be passed through this? (I think in one case we'd have to serialize as a string)?
2. If so, should we allow invoke to override the same env key set during init?

I think for first commit we can stick with args/env and later I can help with filesystem. For example, if we want to allow output bindings to mount / to a directory on the host, and if so do we want to magically handle /tmp for invocations so each has their own.. basically, this discussion is big enough to be another PR, but I wanted to start the brain nag.

bindings/wasm/output.go
```diff
@@ -0,0 +1,192 @@
...
+func (out *outputBinding) Init(ctx context.Context, metadata bindings.Metadata) (err error) {
+	meta, err := out.getInitMetadata(metadata)
+	if err != nil {
+		return fmt.Errorf("wasm: failed to parse metadata: %w", err)
+	}
+
+	out.runtime = wazero.NewRuntimeWithConfig(ctx, out.runtimeConfig)
+	out.module, err = out.runtime.CompileModule(ctx, meta.guest)
+	if err != nil {
+		_ = out.runtime.Close(ctx)
```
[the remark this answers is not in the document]
codefromthecrypt: so this context was made to allow trace context propagation. e.g. know which request closed your runtime, so I think we'll document that close will operate even if the context is canceled. OTOH it is fine do do what you are saying as we haven't actually exposed a trace interface, yet.

bindings/wasm/output.go
```diff
@@ -0,0 +1,192 @@
...
+	moduleConfig := out.moduleConfig.WithName(moduleName)
+
+	if len(req.Data) > 0 {
+		moduleConfig = moduleConfig.WithStdin(bytes.NewReader(req.Data))
+	}
+
+	var stdout bytes.Buffer
+	moduleConfig = moduleConfig.WithStdout(&stdout)
+
+	if args, ok := req.Metadata["args"]; ok {
```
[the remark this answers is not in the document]
codefromthecrypt: I'll add tests. no, an empty isn't desired indeed!

bindings/wasm/output.go
codefromthecrypt: so arg0 is the program name.. it is a bit arbitrary what we choose here, but some things expect that even if there are no args... wdyt?

bindings/wasm/output.go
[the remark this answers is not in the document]
codefromthecrypt: good eye.

codefromthecrypt:
another thing I need feedback on. context cancelation/timeout integration has a high overhead, some factor of ten percent. Most runtimes don't enable things like this by default due to the performance, but it is certainly safer to do that.

Should we enable context cancelation/timeout integration by default in dapr? or should we make this a config?

Either way, we can with a config, and see results even though our demo wasm is a toy. It would be nice to see some representative WASI wasm someone wants to run when this feature is out.

bindings/wasm/output.go
ItalyPaleAle: I don't have a problem with arg0.

What I was trying to say is that you set `moduleConfig = moduleConfig.WithArgs(argsSlice...)` only within the block with `if req.Metadata["args"]`. If `req.Metadata["args"]` is empty, you don't want to set `moduleConfig.WithArgs`?

ItalyPaleAle:
@codefromthecrypt 

With regards to filesystem access and env vars, I think we can postpone that for later. We ought to be careful with exposing env vars and allowing unfettered access to the filesystem, because that could allow accessing secrets that may not be desirable. Do you think that env var and fs access are a must?

As for context cancellation, are you talking about cancelling Invoke() via cancelling the context? Ideally, we'd like to have that supported. Maybe it could be an opt-out thing, which is enabled by default but you can pass a metadata option (in the request) to not enable that?

bindings/wasm/output.go
codefromthecrypt: made it more explicit, and tested it. please re-open this if the direction isn't correct.

codefromthecrypt:
PTAL: we're ready except how to handle ENV unless we want to keep it out of this PR.

bindings/wasm/benchmark_test.go
```diff
@@ -26,10 +25,7 @@ func BenchmarkExample(b *testing.B) {
 		b.Fatal(err)
 	}
 
-	request := &bindings.InvokeRequest{
-		Metadata:  map[string]string{"args": "salaboy"},
```
ItalyPaleAle: No more @salaboy? :)

codefromthecrypt:
whoops I forgot to integrate context cancelation. Lemme do that now.

codefromthecrypt:
@ItalyPaleAle I left this enabled and we can later decide to add/name a property to disable context cancelation. on my laptop the dummy test wasm takes 14us with cancelation enabled and 12us without.

Lemme know what we need to do next on this PR and thanks for the lightning feedback!

bindings/wasm/benchmark_test.go
codefromthecrypt: :) I decided to use the benchmark to test baseline overhead, like the minimal for any wasi function. salaboy still lives forever in the test cases :D

ItalyPaleAle:
I fixed the linter issues, and now LGTM! Thanks for the changes

However, I'd like to get @berndverst's approval on this PR too before we merge it :)

bindings/wasm/benchmark_test.go
salaboy: 😂

bindings/wasm/output.go
```diff
@@ -0,0 +1,208 @@
...
+		// The below ensures context cancels in-flight wasm functions.
+		runtimeConfig: wazero.NewRuntimeConfig().
+			WithCloseOnContextDone(true),
+
+		// The below violate sand-boxing, but allow code to behave as expecteout.
+		moduleConfig: wazero.NewModuleConfig().
+			WithRandSource(rand.Reader).
+			WithSysNanosleep().
+			WithSysWalltime().
+			WithSysNanosleep(),
```
pkedy, line 81: Nit: duplicate `WithSysNanosleep()`. No negative affect though.

bindings/wasm/output.go
```diff
@@ -0,0 +1,208 @@
...
+	argsSlice := []string{out.binaryName}
+
+	// Get any remaining args from configuration
+	if args := req.Metadata["args"]; args != "" {
+		argsSlice = append(argsSlice, strings.Split(args, ",")...)
+	}
+	moduleConfig = moduleConfig.WithArgs(argsSlice...)
+
+	// Instantiating executes the guest's main function (exported as _start).
+	_, err := out.runtime.InstantiateModule(ctx, out.module, moduleConfig)
```
pkedy, line 142: @codefromthecrypt You know best. Even though `_start` runs to completion, do you need to call `module.Close()` to clean up any resources in wazero?

Would it make sense to add `defer module.Close()`?

pkedy (dismissed):
A few comments (mainly curious about the need for `module.Close()`) but LGTM.

bindings/wasm/output.go
```diff
@@ -0,0 +1,208 @@
...
+		logger: logger,
+
+		// The below ensures context cancels in-flight wasm functions.
+		runtimeConfig: wazero.NewRuntimeConfig().
+			WithCloseOnContextDone(true),
+
+		// The below violate sand-boxing, but allow code to behave as expecteout.
+		moduleConfig: wazero.NewModuleConfig().
+			WithRandSource(rand.Reader).
+			WithSysNanosleep().
```
berndverst, line 79: Remove this duplicate option method

```suggestion
```

berndverst (approved):

bindings/wasm/output.go
codefromthecrypt: good eye!

codefromthecrypt: https://github.com/dapr/components-contrib/pull/2697

PR #2688, opened 2023-03-20
Fix inconsistency with TTL and Etags in SQL databases + fix SQL Server cert tests

ItalyPaleAle:
Changes these state stores:

- Postgres
- CockroachDB
- SQLite
- ~~Oracle~~ (moved to #2702)

When a row is expired, it is considered "soft deleted", so performing a set with an Etag should not update anything if the row has expired. Otherwise, there'd be inconsistencies between updating rows that have expired ("soft-deleted") and actually "hard-deleted"

See also: https://github.com/dapr/components-contrib/pull/2682#discussion_r1142299949

Additionally, this fixes another issue with "soft deleted" rows and first-write concurrency: rows that are expired but not yet garbage-collected should allow first-write wins.

> Also fixes SQL Server certification tests that were broken in master


JoshVanL:
lgtm :+1:

berndverst (approved):

state/sqlite/sqlite_dbaccess.go
```diff
@@ -377,7 +377,8 @@ func (a *sqliteDBAccess) doSet(parentCtx context.Context, db querier, req *state
 				expiration_time = %s
 			WHERE
 				key = ?
-				AND eTag = ?`,
+				AND etag = ?
+				AND (expiration_time IS NULL OR expiration_time > CURRENT_TIMESTAMP)`,
```
DeepanshuA, line 381:
```suggestion
				AND (expiration_time IS NULL OR expiration_time >= CURRENT_TIMESTAMP)`,
```

ItalyPaleAle:
Moving back to draft as I found a few bugs in the Oracle state store component, and need to investigate the failures with SQL server

berndverst (approved):

berndverst (approved):

berndverst (approved):

Issue #2461, opened 2023-01-24
Certify GCP PubSub as Stable


artursouza:
Certify GCP PubSub as Stable

PR #2689, opened 2023-03-20
GCP PubSub - Certification Tests

robertojrojas:
# Description

Provides the GCP PubSub Conformance and Certification Test Plan and Test Code.

This PR depends on PR: #2714

## Issue reference

This PR will close: #2461

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


tests/certification/pubsub/gcp/pubsub/gcppubsub_test.go
```diff
@@ -0,0 +1,799 @@
...
+			}
+
+			return nil
+		}
+	}
+
+	flow.New(t, "GCPPubSub certification - Existing Topic").
+
+		// Run subscriberApplication app1
+		Step(app.Run(appID1, fmt.Sprintf(":%d", appPort+portOffset*3),
```
artursouza, line 751: Why "*3" ? Isn't it supposed to be "+3"?

artursouza:
Why is this PR titled "Part 1"?

tests/certification/pubsub/gcp/pubsub/gcppubsub_test.go
robertojrojas: well it turns out that `*3` what's used in number of certification tests.

robertojrojas:
@artursouza In previous Cert tests I had to divide them into multiple PRs due to having too many tests and other configuration to be done. This one can be just one so I'm changing the name to remove the `Part 1`.

.github/scripts/test-info.mjs
```diff
@@ -232,6 +232,25 @@ const components = {
...
         sourcePkg: 'pubsub/aws/snssqs',
     },
+    'pubsub.gcp.pubsub': {
+        certification: true,
+        requireTerraform: true,
+        requireGCPCredentials: true,
+        certificationSetup: 'certification-pubsub.gcp.pubsub-setup.sh',
+        certificationDestroy: 'certification-pubsub.gcp.pubsub-destroy.sh',
+    },
+    'pubsub.gcp.pubsub.docker': {
```
berndverst, line 242: Could we please comment out the docker case? Just like the Azure components we do not run a dockerized emulator, and it would be best to to minimize the number of task runners every PR executes.

berndverst: I made this update.

berndverst (approved):

PR #2690, opened 2023-03-21
Crypto building block: core parts

ItalyPaleAle:
This is a spin-off from #2237 that includes the "core parts" only.

Contains:

- Interfaces for the building block
- Some shared code
- Conformance tests
- Changes to the Azure infrastructure to enable the Azure Key Vault conformance tests. Some properties have been renamed since they're not specific to the secret store anymore.
   - Note that I've already applied those changes manually in the Azure Portal. The values in the Azure Portal at the moment are available under both the old and new names.


ItalyPaleAle:
One comment from #2237:

Should we also include 2 new methods in the SubtleCrypto interface called:

- SupportedEncryptionAlgorithms
- SupportedSignatureAlgorithms

They each return a slice with the list of algorithms supported. Those can then be featured in the Metadata API. So a user could call the metadata endpoint and know for the crypto components that are enabled, what algorithms are supported.  This is different from components' features because they're 2 separate lists.

If you think it's useful, I'll add it in a separate PR to not make this even bigger

crypto/key.go
```diff
@@ -0,0 +1,114 @@
...
+
+	if k.nbf != nil && k.nbf.After(t) {
+		return false
+	}
+
+	return true
+}
+
+// KeyCanPerformOperation returns true if the key can be used to perform a specific operation.
+func KeyCanPerformOperation(key jwk.Key, op jwk.KeyOperation) bool {
```
mukundansundar, line 70: does this need to be an exported function? 
Can we use the above function `CanPerformOperation` on a `jwk.Key` instead?

ItalyPaleAle: Yes there is a reason for this (and `KeyCanPerformAlgorithm` below) to be exported functions.

These functions need to be able to support any object that implements `jwk.Key`. The `crypto.Key` struct defined in this package implements `jwk.Key`, but extends it too. Some components do not have a `crypto.Key`, but only a `jwk.Key`.

crypto/subtlecrypto.go
```diff
@@ -0,0 +1,162 @@
...
+package crypto
+
+import (
+	"context"
+
+	"github.com/lestrrat-go/jwx/v2/jwk"
+)
+
+// SubtleCrypto offers an interface to perform low-level ("subtle") cryptographic operations with keys stored in a vault.
+type SubtleCrypto interface {
```
berndverst, line 23: Can you please add GetComponentMetadata the same way we have in State Store and Secret Store?

berndverst (dismissed):
LGTM, but please add `GetComponentMetadata()` which returns `map[string]string` to the interface.

crypto/subtlecrypto.go
ItalyPaleAle: Done

berndverst (approved):

PR #2696, opened 2023-03-23
Add metadata analyzer tools

berndverst:
# Description

These tools in `.metadata-tools` check that `metadata.yaml` files are up to date with the component implementation, by leveraging the `GetComponentMetadata()` method.

The intended purpose is to use this a CI/CD check in the future which prevents any contrib contributor from adding a new metadata option without also updating the corresponding `metadata.yaml` file.

Usage:
```bash
make check-component-metadata
```

Currently this will return the following:

```

{
  "state/azure/cosmosdb": {
    "contenttype": "string"
  },
  "state/mongodb": {
    "connectionstring": "string"
  }
}
```

How this is done:
1. We traverse all Go files in contrib with the exclusion of some folders.
2. We parse the file into an abstract syntax tree and search for any constructors creating StateStores or SecretStores (at this time only these have the GetComponentMetadata method in their interface)
3. We load a component's metadata.yaml and track all metadata field names as well as their data type (the type is currently not used by this tool, but will be used in the future)
4. We write to file a new Go program which imports and Constructs all SecretStore and StateStore components found before calling the `GetComponentMetadata()` function. Then we check that all the metadata fields are present in the corresponding yaml file if that file exists.

Future steps:
1. Also throw an error for metadata fields present in the yaml but not present in the component -- this will need special handling for the standard Azure auth metadata options.
2. Also list incorrect data types, e.g. matadata.yaml states a field is of type string but should be number because the underlying type is int.
3. This will need to be added to CI/CD


berndverst:
FYI @tmacam @ItalyPaleAle

.metadata-tools/metadataanalyzer/metadatainfo.go
berndverst, line 1: Note that this is just a partial program. The `metadataanalyzer.go` main application here is created programmatically and written to file via the `metadataanalyzercreator.go` application.

From the contrib root, run `go run .metadata-tools/metadataanalyzercreator.go ./`

berndverst:
Fun fact, the binary created by the above steps is about 110MB at this time. That's how much it takes just to import all state stores and secret stores!

.metadata-tools/metadataanalyzer/go.mod
```diff
@@ -0,0 +1,199 @@
+module metadataanalyzer
+
+go 1.20
+
+replace github.com/dapr/components-contring => /Users/bverst/src/dapr/components-contrib
```
berndverst, line 5:
```suggestion
replace github.com/dapr/components-contring => ../../
```

berndverst:
I might dynamically create the go.mod / go.sum files too (in the instructions and CI/CD). That will avoid the modtidy-diffcheck failure

.metadata-tools/metadataanalyzercreator.go
```diff
@@ -0,0 +1,178 @@
...
+			if !skip {
+				packageName = filepath.Base(dir) + "/" + packageName
+			} else {
+				packageName = filepath.Base(dir)
+			}
+			skip = false
+			dir = filepath.Dir(dir)
+
+			switch filepath.Base(dir) {
+			case "bindings", "pubsub", "state", "secretstores", "middleware", "workflows", "lock", "configuration", "crypto", "nameresolution":
```
ItalyPaleAle, line 53: Can you use `.build-tools/component-folders.json` instead?

.metadata-tools/metadataanalyzercreator.go
```diff
@@ -0,0 +1,178 @@
...
+		return nil
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// now we have a list of packages that implement the GetComponentMetadata method
+	// now write our go program to check the metadata of each component and compare it againts the yaml metadata
+
+	importBlock := ""
```
ItalyPaleAle, line 95: Although this works, I wonder if putting the code in a template is better. Like, create a separate file, load it from disk, and use the Go template syntax. May be easier to maintain long-term. https://pkg.go.dev/text/template

.metadata-tools/metadataanalyzercreator.go
```diff
@@ -0,0 +1,178 @@
...
+						continue
+					}
+					if selExpr.Sel.Name == typeSplit[1] {
+						return f.Name.Name, nil
+					}
+				}
+			}
+		}
+	}
+	return "", fmt.Errorf("could not find constructor method")
```
ItalyPaleAle, line 177: Total nit :)

```suggestion
	return "", errors.New("could not find constructor method")
```

ItalyPaleAle (commented):
This is good stuff

tmacam:
> We parse the file into an abstract syntax tree and search for any constructors creating StateStores or SecretStores (at this time only these have the GetComponentMetadata method in their interface)

So we should add `GetComponentMetadata` method to the remaining interfaces? Or is the plan to, somehow, do some black-magic to bypass this limitation?

.metadata-tools/metadataanalyzer/go.mod
```diff
@@ -0,0 +1,199 @@
+module metadataanalyzer
+
+go 1.20
+
+replace github.com/dapr/components-contring => ../../
```
tmacam, line 5: components-contring -> components-contrib ?

.metadata-tools/metadataanalyzer/go.mod
```diff
@@ -0,0 +1,199 @@
...
+	cloud.google.com/go/compute v1.18.0 // indirect
+	cloud.google.com/go/compute/metadata v0.2.3 // indirect
+	cloud.google.com/go/datastore v1.10.0 // indirect
+	cloud.google.com/go/iam v0.11.0 // indirect
+	cloud.google.com/go/secretmanager v1.10.0 // indirect
+	github.com/Azure/azure-sdk-for-go v68.0.0+incompatible // indirect
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.5.0-beta.1 // indirect
+	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.3.0-beta.4 // indirect
+	github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos v0.3.3 // indirect
+	github.com/Azure/azure-sdk-for-go/sdk/data/aztables v1.0.1 // indirect
```
tmacam, line 24: Yeah, auto-generating this will be great because this is covered by .gitignore  and it will be easy to forget to update these

.metadata-tools/metadataanalyzercreator.go
berndverst: I will take a look. Wasn't aware of that file

.metadata-tools/metadataanalyzercreator.go
```diff
@@ -0,0 +1,178 @@
+package main
```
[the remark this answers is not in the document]
berndverst: Let me try - should be possible, yes.

.metadata-tools/metadataanalyzercreator.go
berndverst: I'll take a look.

.metadata-tools/metadataanalyzer/go.mod
berndverst: Please use the suggestion feature in the future so it can be applied @tmacam

berndverst:
```suggestion
replace github.com/dapr/components-contrib => ../../
```

.metadata-tools/metadataanalyzer/go.mod
berndverst: Yup - I will cover that via a Makefile command instead.

berndverst:
> > We parse the file into an abstract syntax tree and search for any constructors creating StateStores or SecretStores (at this time only these have the GetComponentMetadata method in their interface)
> 
> So we should add `GetComponentMetadata` method to the remaining interfaces? Or is the plan to, somehow, do some black-magic to bypass this limitation?

@tmacam yes that is correct, all other interfaces need to have GetComponentMetadata() added, and then that needs to be implemented in every component. I did so for SecretStore and StateStore before. This still remains to be done for other components.

Please do not further comment on that part here - as that is not up for discussion in this PR :)

metadata/duration.go
```diff
@@ -83,21 +83,26 @@ func toTimeDurationHookFunc() mapstructure.DecodeHookFunc {
...
 		case reflect.Float64:
-			val := time.Duration(data.(float64))
+			val := time.Duration(data.(float64) * float64(time.Second))
 			if t != reflect.TypeOf(Duration{}) {
 				return val, nil
 			}
 			return Duration{Duration: val}, nil
 		case reflect.Int64:
-			val := time.Duration(data.(int64))
+			val := time.Duration(data.(int64) * int64(time.Second))
```
ItalyPaleAle, line 105: Isn't this a breaking change?

Before it would parse the time as nanoseconds, now as seconds.

internal/component/postgresql/metadata.go
```diff
@@ -63,32 +62,20 @@ func (m *postgresMetadataStruct) InitWithMetadata(meta state.Metadata) error {
...
-
+	if m.CleanupInterval != nil {
 		// Non-positive value from meta means disable auto cleanup.
-		if cleanupIntervalInSec > 0 {
-			m.cleanupInterval = ptr.Of(time.Duration(cleanupIntervalInSec) * time.Second)
-		} else {
-			m.cleanupInterval = nil
+		if *m.CleanupInterval <= 0 {
+			if meta.Properties[cleanupIntervalKey] == "" {
+				// unfortunately the mapstructure decoder decodes an empty string to 0, a missing key would be nil however
```
ItalyPaleAle, line 74: Is there a way we can change this behavior so we don't need to go back to check if the string is empty? Such a behavior seems dangerous.

ItalyPaleAle (commented):
Have one question otherwise LGTM

ItalyPaleAle (changes requested):
Have one question otherwise LGTM

metadata/duration.go
berndverst: Correct but this was never actually used anywhere - that's why this is fine!

internal/component/postgresql/metadata.go
berndverst: No we cannot - I tried for a few hours, but this is not a feature of the library today unfortunately.

berndverst: I don't think it's a real issue - the test case of empty string is unrealistic. That's different from omitting the field.

metadata/duration.go
berndverst: Reverted this particular change as it isn't strictly necessary. We are never encountering this situation because we are always marshaling a string (via map[string]string anyway).

ItalyPaleAle (approved):

PR #2704, opened 2023-03-24
Add BulkGet support for SQLite and Postgres state stores

ItalyPaleAle:
# Description

This PR implements proper BulkGet for the SQLite and Postgres (so, also CockroachDB) state stores. Also adds a conformance test for BulkGet.

This is particularly helpful for actor reminders. When BulkGet isn't implemented, the runtime performs multiple Get operations in sequence, causing more I/O than needed. A "proper" BulkGet performs a single SELECT query that retrieves multiple rows.

## Issue reference

Related to dapr/dapr#6121


state/sqlite/sqlite_dbaccess.go
```diff
@@ -246,50 +247,92 @@ func (a *sqliteDBAccess) Get(parentCtx context.Context, req *state.GetRequest) (
...
+}
+
+func (a *sqliteDBAccess) BulkGet(parentCtx context.Context, req []state.GetRequest) (bool, []state.BulkGetResponse, error) {
+	if len(req) == 0 {
+		return true, []state.BulkGetResponse{}, nil
+	}
+
+	// SQLite doesn't support passing an array for an IN clause, so we need to build a custom query
+	inClause := strings.Repeat("?,", len(req))
+	inClause = inClause[:(len(inClause) - 1)]
```
JoshVanL, line 281: Maybe do this instead
```suggestion
	inClause := strings.Join(strings.Repeat("?", len(req)), ",")
```

state/sqlite/sqlite_dbaccess.go
```diff
@@ -246,50 +247,92 @@ func (a *sqliteDBAccess) Get(parentCtx context.Context, req *state.GetRequest) (
...
+	for rows.Next() {
+		r := state.BulkGetResponse{}
+		r.Key, r.Data, etag, err = readRow(rows)
+		if err != nil {
+			r.Error = err.Error()
+		}
+		r.ETag = &etag
+		res[n] = r
+		n++
+	}
```
JoshVanL, line 313:
```suggestion
	for ; rows.Next(); n++ {
		r := state.BulkGetResponse{}
		r.Key, r.Data, etag, err = readRow(rows)
		if err != nil {
			r.Error = err.Error()
		}
		r.ETag = &etag
		res[n] = r
	}
```

JoshVanL (changes requested):
Some nits from me, but otherwise lgtm

state/sqlite/sqlite_dbaccess.go
ItalyPaleAle: I'm afraid this won't work as `strings.Repeat` returns a string, while `strings.Join` wants a slice. We could of course allocate a slice, but that would require more allocations on the heap and doesn't seem as efficient as the code above (which I reckon isn't the prettiest, but should be fairly efficient and this is a hot path).

ItalyPaleAle:
Thanks for the review @JoshVanL

berndverst (approved):
Nice!

PR #2708, opened 2023-03-24
Fix Postgres state store hardcoding table name in Delete method

### LINKED ISSUE #2707: Postgres state store: DELETE queries use hardcoded table name
In Dapr 1.10 and master branch, the Postgres state component (stable) has a bug in the `doDelete` method, where DELETE queries use hardcoded table name:

https://github.com/dapr/components-contrib/blob/release-1.10/state/postgresql/postgresdbaccess.go#L326

This means that the Delete method is not usable when the state table has a different name.

ItalyPaleAle:
Fixes #2707

### DIFF
--- a/internal/component/postgresql/postgresdbaccess.go
+++ b/internal/component/postgresql/postgresdbaccess.go
@@ -329,7 +329,7 @@ func (p *PostgresDBAccess) doDelete(parentCtx context.Context, db dbquerier, req
 
 	var result pgconn.CommandTag
 	if req.ETag == nil || *req.ETag == "" {
-		result, err = db.Exec(parentCtx, "DELETE FROM state WHERE key = $1", req.Key)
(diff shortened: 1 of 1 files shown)


internal/component/postgresql/postgresdbaccess.go
```diff
@@ -329,7 +329,7 @@ func (p *PostgresDBAccess) doDelete(parentCtx context.Context, db dbquerier, req
 
 	var result pgconn.CommandTag
 	if req.ETag == nil || *req.ETag == "" {
-		result, err = db.Exec(parentCtx, "DELETE FROM state WHERE key = $1", req.Key)
+		result, err = db.Exec(parentCtx, "DELETE FROM "+p.metadata.TableName+" WHERE key = $1", req.Key)
```
berndverst, line 332: Per conversation here, the table name is sanitized so cannot cause a SQL injection. Looks good as a result.

berndverst (approved):

berndverst:
Should cherry pick this into release-1.10 to include with the next hotfix. I'll do that.

PR #2710, opened 2023-03-25
Standardize metadata parsing in configstores and lockstore

berndverst:
# Description

Standardized the metadata parsing in configstores and lockstores.

Adds the `GetComponentMetadata` interface method required for checking the validity of `metadata.yaml` files.

### DIFF
--- a/configuration/azure/appconfig/appconfig.go
+++ b/configuration/azure/appconfig/appconfig.go
@@ -17,7 +17,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strconv"
+	"reflect"
 	"sync"
 	"time"
 
@@ -30,6 +30,7 @@ import (
 
 	"github.com/dapr/components-contrib/configuration"
 	azauth "github.com/dapr/components-contrib/internal/authentication/azure"
+	contribMetadata "github.com/dapr/components-contrib/metadata"
 
 	"github.com/dapr/kit/logger"
 )
@@ -85,9 +86,9 @@ func (r *ConfigurationStore) Init(_ context.Context, metadata configuration.Meta
 			ApplicationID: "dapr-" + logger.DaprVersion,
 		},
 		Retry: policy.RetryOptions{
-			MaxRetries:    int32(m.maxRetries),
-			RetryDelay:    m.maxRetryDelay,
-			MaxRetryDelay: m.maxRetryDelay,
+			MaxRetries:    int32(m.MaxRetries),
+			RetryDelay:    m.internalMaxRetryDelay,
+			MaxRetryDelay: m.internalMaxRetryDelay,
 		},
 	}
 
@@ -95,8 +96,8 @@ func (r *ConfigurationStore) Init(_ context.Context, metadata configuration.Meta
 		ClientOptions: coreClientOpts,
 	}
 
-	if r.metadata.connectionString != "" {
-		r.client, err = azappconfig.NewClientFromConnectionString(r.metadata.connectionString, &options)
+	if r.metadata.ConnectionString != "" {
+		r.client, err = azappconfig.NewClientFromConnectionString(r.metadata.ConnectionString, &options)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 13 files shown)


metadata/utils.go
```diff
@@ -140,6 +140,14 @@ func GetMetadataProperty(props map[string]string, keys ...string) (val string, o
 // DecodeMetadata decodes metadata into a struct
 // This is an extension of mitchellh/mapstructure which also supports decoding durations
 func DecodeMetadata(input interface{}, result interface{}) error {
+	// avoid a common of passing the metada object, instead of the properties map
```
berndverst, line 143:
```suggestion
	// avoids a common mistake of passing the metadata object, instead of the properties map
```

configuration/azure/appconfig/appconfig.go
```diff
@@ -123,67 +124,34 @@ func (r *ConfigurationStore) Init(_ context.Context, metadata configuration.Meta
...
-	if val, ok := meta.Properties[connectionString]; ok && val != "" {
-		m.connectionString = val
+	m := metadata{
+		MaxRetries:                    defaultMaxRetries,
+		internalMaxRetryDelay:         defaultMaxRetryDelay,
+		internalRetryDelay:            defaultRetryDelay,
+		internalSubscribePollInterval: defaultSubscribePollInterval,
+		internalRequestTimeout:        defaultRequestTimeout,
 	}
+	contribMetadata.DecodeMetadata(meta.Properties, &m)
```
ItalyPaleAle, line 134: Err check?

configuration/postgres/postgres.go
```diff
@@ -278,31 +279,27 @@ func (p *ConfigurationStore) handleSubscribedChange(ctx context.Context, handler
...
 
 func parseMetadata(cmetadata configuration.Metadata) (metadata, error) {
-	m := metadata{}
-	if val, ok := cmetadata.Properties[connectionStringKey]; ok && val != "" {
-		m.connectionString = val
-	} else {
+	m := metadata{
+		MaxIdleTimeout: defaultMaxConnIdleTime,
+	}
+	contribMetadata.DecodeMetadata(cmetadata.Properties, &m)
```
ItalyPaleAle, line 285: Err check?

configuration/redis/redis.go
```diff
@@ -74,69 +76,26 @@ func NewRedisConfigurationStore(logger logger.Logger) configuration.Store {
...
-		}
-		m.MaxRetries = int(parsedVal)
+	m := metadata{
+		EnableTLS:               defaultEnableTLS,
+		MaxRetries:              defaultMaxRetries,
+		internalMaxRetryBackoff: defaultMaxRetryBackoff,
+		Failover:                false,
+		DB:                      defaultDB,
 	}
+	contribMetadata.DecodeMetadata(meta.Properties, &m)
```
ItalyPaleAle, line 86: Err check?

metadata/utils.go
```diff
@@ -140,6 +140,14 @@ func GetMetadataProperty(props map[string]string, keys ...string) (val string, o
 // DecodeMetadata decodes metadata into a struct
 // This is an extension of mitchellh/mapstructure which also supports decoding durations
 func DecodeMetadata(input interface{}, result interface{}) error {
+	// avoids a common mistake of passing the metadata object, instead of the properties map
+	// if input is not of type map[string]string, then cast to metadata.Base and access the Properties
+	if _, ok := input.(map[string]string); !ok {
```
ItalyPaleAle, line 145: Why not just changing the signature to `input map[string]string`?

berndverst: Now it's for convenience - being able to support the metadata.base class is actually much nicer.

ItalyPaleAle: One very convenient option (that would involve some refactoring) would be to change `GetComponentMetadata()` to return `any`. The component would simply return whatever metadata struct they use (for example `return redisMetadata{}`), and the caller would be doing all the reflection magic.

(Bonus: move the code that does `DecodeMetadata` into a file compiled only with `//go:build +metadata` so it's not even included in the built `daprd`)...

berndverst: Sorry, not going to do that in this PR.

ItalyPaleAle: To be clear I was not suggesting any change in this PR. Just thinking out loud, maybe for the future!

ItalyPaleAle (approved):

PR #2720, opened 2023-03-29
Refactor of Metadata parsing in Bindings

berndverst:
# Description

Refactor of metadata parsing in Bindings.

I also added the ability to specify that metadata only applies to certain component types.

Example:

```go
type ComponentMetadata struct {
  MyField `mapstructure:"somefield" only:"pubsub,binding"`
}
```

This would only include this field "somefield" in Pubsub and Binding components metadata, but not any other type.


berndverst:
FYI this PR didn't touch the components that failed, so it should be good to merge if it passes code review!

bindings/kubernetes/kubernetes.go
```diff
@@ -66,22 +74,23 @@ func (k *kubernetesInput) Init(ctx context.Context, metadata bindings.Metadata)
...
-	if val, ok := metadata.Properties["resyncPeriodInSec"]; ok && val != "" {
-		intval, err := strconv.ParseInt(val, 10, 64)
-		if err != nil {
-			k.logger.Warnf("invalid resyncPeriodInSec %s; %v; defaulting to 10s", val, err)
-			k.resyncPeriod = time.Second * 10
+func (k *kubernetesInput) parseMetadata(meta bindings.Metadata) error {
+	m := kubernetesMetadata{}
+	err := metadata.DecodeMetadata(meta.Properties, &m)
+	if err != nil {
+		if strings.Contains(err.Error(), "decoding 'resyncPeriodInSec': time: invalid duration") {
```
ItalyPaleAle, line 81: I would change this to:

```suggestion
		if strings.Contains(err.Error(), "resyncPeriodInSec") {
```

This is so if the upstream library (mapdecode) changes how it formats errors, this is less likely to break

metadata/utils.go
```diff
@@ -204,9 +204,35 @@ func toStringArrayHookFunc() mapstructure.DecodeHookFunc {
 	}
 }
 
+type ComponentTypeValue string
+
+var ComponentType = struct {
```
ItalyPaleAle, line 209: Does this need to be a struct? Can it be a set of constants instead?

As per our conversation on a recent PR, using a struct increases the memory usage, while consts should be easier on the memory.

ItalyPaleAle (changes requested):

metadata/utils.go
berndverst: Changed it to constants with an addition `switch` statement check.

ItalyPaleAle (approved):

ItalyPaleAle (approved):

ekjuanrejon:
hello, does the zeebee component support basic authentication?

PR #2724, opened 2023-03-30
[Release 1.10] fix support for custom state store table names

berndverst:
# Description

Fixes the ability to use a custom state store table name. Without this fix only some operations work correctly with a custom state store table name.

Components fixed:
- CockroachDB
- PostgreSQL
- Oracle Database

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
~* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_~


state/postgresql/metadata.go
```diff
@@ -57,6 +58,20 @@ func (m *postgresMetadataStruct) InitWithMetadata(meta state.Metadata) error {
 		return err
 	}
 
+	if m.TableName != "" {
```
ItalyPaleAle, line 61: This is making the postgres cert test fail.

The reason is that some DBs have different rules for valid table names. In the case of Postgres (and probably others), `.` is a valid character if you want to specify the schema name.

ItalyPaleAle (changes requested):

state/postgresql/metadata.go
berndverst: Then I will just remove this validation for table names again and leave it the way it was. In 1.11 it can then be done properly.  What do you think? Not likely someone can edit the yaml but doesn't have the credentials to the DB anyway.

I certainly won't be writing custom validation code for each DB here.

berndverst:
Should be good now @ItalyPaleAle

ItalyPaleAle (approved):

PR #2733, opened 2023-03-31
RethinkDB: Disable transactions that were not working

ItalyPaleAle:
The implementation of the "Multi" method in RethinkDB was not correct, as it caused operations to be executed in the wrong order and it was not using a transaction (per comment on code: `best effort, no transacts supported`)

In essence, RethinkDB was not supporting transactions, and advertising the existence of the Multi method can cause users to believe it does, with dangerous potentials for data loss.

### DIFF
--- a/state/rethinkdb/rethinkdb.go
+++ b/state/rethinkdb/rethinkdb.go
@@ -52,10 +52,10 @@ type stateConfig struct {
 }
 
 type stateRecord struct {
-	ID   string      `json:"id" rethinkdb:"id"`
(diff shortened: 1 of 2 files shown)


ItalyPaleAle:
>  also docs needs to be changed.

Docs did not mention RethinkDB as supporting transactions: https://docs.dapr.io/reference/components-reference/supported-state-stores/

Because of that, and the fact that it's beta, I think this is not a breaking change (or at least not one we should be concerned with).

PR #2734, opened 2023-04-01
Redis state store: fixed JSON handling in Multi

ItalyPaleAle:
Also fixed a race condition in ClientHasJSONSupport, which was storing the result in a global variable (so it was shared among all components) and in a non-thread safe way.


state/redis/redis.go
```diff
@@ -344,43 +347,40 @@ func (r *StateStore) Set(ctx context.Context, req *state.SetRequest) error {
 		firstWrite = 0
 	}
 
-	var bt []byte
-	var setQuery string
-	if contentType, ok := req.Metadata[daprmetadata.ContentType]; ok && contentType == contenttype.JSONContentType && rediscomponent.ClientHasJSONSupport(r.client) {
-		setQuery = setJSONQuery
-		bt, _ = utils.Marshal(&jsonEntry{Data: req.Value}, r.json.Marshal)
+	if req.Metadata[daprmetadata.ContentType] == contenttype.JSONContentType && r.clientHasJSON {
+		bt, _ := utils.Marshal(jsonEntry{Data: req.Value}, r.json.Marshal)
```
mukundansundar, line 351:
```suggestion
		bt, _ := utils.Marshal(&jsonEntry{Data: req.Value}, r.json.Marshal)
```
why was pointer ref removed from older code?

mukundansundar:
also should this be ported back into previous versions of Dapr?

ItalyPaleAle:
> also should this be ported back into previous versions of Dapr?

I don't think that's necessary. I discovered this issue while working on something else, which is coming very soon. This should not have a major impact on code used today.

mukundansundar (approved):

PR #2735, opened 2023-04-01
Crypto components: add methods that return the list of supported algorithms

### LINKED ISSUE #2693: Crypto building block: implement `Supported(Encryption|Signature)Algorithm` for metadata API
https://github.com/dapr/components-contrib/pull/2237#issuecomment-1476791268 

> Makes me wonder if perhaps we should include 2 new methods in the SubtleCrypto interface called:
`SupportedEncryptionAlgorithms`
`SupportedSignatureAlgorithms`
They each return a slice with the list of algorithms supported. Those can then be featured in the Metadata API.

> Thoughts? (If you think it's useful, I'll add it in a separate PR to not make this even bigger)

I think providing this detail will be really useful in the Metadata API, allowing users to know which algorithms can be used with which component.

_Originally posted by @mukundansundar in https://github.com/dapr/components-contrib/issues/2690#issuecomment-1478159832_

ItalyPaleAle:
Fixes #2693

Also adds conformance tests

PS: The change is not nearly as large as this PR makes it look like :) It's just because of the updated dapr/kit dependency causing lots of changes in go.mod/go.sum files


crypto/azure/keyvault/component.go
```diff
@@ -401,6 +399,14 @@ func (k *keyvaultCrypto) getVaultURI() string {
 	return fmt.Sprintf("https://%s.%s", k.md.VaultName, k.md.vaultDNSSuffix)
 }
 
+func (keyvaultCrypto) SupportedEncryptionAlgorithms() []string {
```
mukundansundar, line 402: should we let the user know that they have to call init before calling this method? 
```suggestion
func (keyvaultCrypto) SupportedEncryptionAlgorithms() ([]string, error) {
```
This way we can return error if this method is called without calling Init.

mukundansundar: Or do something similar to local base component, call algsParsed.Do() before returning the value. same below.

crypto/subtlecrypto.go
```diff
@@ -163,3 +165,9 @@ type SubtleCrypto interface {
 	// GetComponentMetadata returns information on the component's metadata.
 	GetComponentMetadata() map[string]string
 }
+
+// SubtleCryptoMetadata is an extension to SubtleCrypto that includes methods to return information on the supported algorithms.
+type SubtleCryptoMetadata interface {
```
mukundansundar, line 170:
```suggestion
type SubtleCryptoAlgorithms interface {
```

crypto/azure/keyvault/component.go
ItalyPaleAle: I think having the assumption that `Init()` be called before this method is fair, and I don't think we need to make the method return an error.

The way this method is meant to be used is by the component metadata API, the same one that calls `Features()`: 
https://github.com/dapr/dapr/blob/263a21cf1a3d1bad5a4f261d1785fbe3eab09785/pkg/runtime/runtime.go#L3117

crypto/subtlecrypto.go
ItalyPaleAle: Thanks

mukundansundar (approved):

Issue #2187, opened 2022-10-13
[Proposal]: Semantic versioning for configuration items in Configuration Building Block API


akhilac1:
## Describe the proposal
Configuration building block GET API allows a user to get 'all' the configuration data when no specific key is specified.
Owing to the current structure of response, only one value can be returned for a key.

https://github.com/dapr/components-contrib/blob/3427cc1c590f1e05cbf202a94f409d042b68bc00/configuration/responses.go#L17-L19

So, when there are multiple versions of a key and no filtering criteria is provided, the configuration building block should to return the latest version. 

To enable reliable comparison of the versions  for configuration items, following are required -
- Specification should be updated to clearly mandate adherence to semantic versioning standards 
- Configuration components should be enhanced to compare and return the configuration item with the latest version, when no filter is specified 
- To support backward compatibility, if version items does not follow standards of semantic versioning, system should compare major and minor versions and return the item with highest version.

rabollin:
> ## Describe the proposal
> Configuration building block GET API allows a user to get 'all' the configuration data when no specific key is specified. Owing to the current structure of response, only one value can be returned for a key.
> 
> https://github.com/dapr/components-contrib/blob/3427cc1c590f1e05cbf202a94f409d042b68bc00/configuration/responses.go#L17-L19
> 
> So, when there are multiple versions of a key and no filtering criteria is provided, the configuration building block should to return the latest version.
> 
> To enable reliable comparison of the versions for configuration items, following are required -
> 
> * Specification should be updated to clearly mandate adherence to semantic versioning standards
> * Configuration components should be enhanced to compare and return the configuration item with the latest version, when no filter is specified
> * To support backward compatibility, if version items does not follow standards of semantic versioning, system should compare major and minor versions and return the item with highest version.

Versioning would help with differentiating the configuration changes and App may have more control now, however
1. Does the App need to keep track of the versioning?
2. How do you handle the existing implementations?

berndverst:
Does this require an API change or is this just in the implementation in each component?

If this is required of every component, consider enforcing this in the API somehow (and hence the proposal must be in dapr/dapr).

Changes to the API or building block behaviors in general should be discussed via proposal in the `dapr/dapr` repo. You can ask a STC member to transfer this issue there.

akhilac1:
> Does this require an API change or is this just in the implementation in each component?
> 
> If this is required of every component, consider enforcing this in the API somehow (and hence the proposal must be in dapr/dapr).
> 
[...]

PR #2740, opened 2023-04-03
Adding postgres config store in conformance tests

shivamkm07:
# Description

This PR adds postgres config store in conformance tests. It also fixes some bugs required in the store implementation and includes the updater implementation for postgres. 

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


configuration/postgres/postgres.go
```diff
@@ -69,7 +69,7 @@ const (
 	ErrorMissingConnectionString = "missing postgreSQL connection string"
 	ErrorAlreadyInitialized      = "postgreSQL configuration store already initialized"
 	ErrorMissingMaxTimeout       = "missing PostgreSQL maxTimeout setting in configuration"
-	QueryTableExists             = "SELECT EXISTS (SELECT FROM pg_tables where tablename = $1)"
+	QueryTableExists             = "SELECT EXISTS (SELECT FROM pg_tables where tablename = lower($1))"
```
ItalyPaleAle, line 72: Why are you lowercasing the table?

configuration/postgres/postgres.go
```diff
@@ -162,42 +162,40 @@ func (p *ConfigurationStore) Get(ctx context.Context, req *configuration.GetRequ
 }
 
 func (p *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.SubscribeRequest, handler configuration.UpdateHandler) (string, error) {
-	var pgNotifyChannels []string
+	pgNotifyChannel := ""
 	for k, v := range req.Metadata {
-		if res := strings.EqualFold(pgNotifyChannelKey, k) && !slices.Contains(pgNotifyChannels, v); res {
-			pgNotifyChannels = append(pgNotifyChannels, v) // append unique channel names only
+		if res := strings.EqualFold(pgNotifyChannelKey, k); res {
```
ItalyPaleAle, line 167:
```suggestion
		if strings.EqualFold(pgNotifyChannelKey, k) {
```

tests/conformance/configuration/configuration.go
```diff
@@ -142,13 +147,21 @@ func ConformanceTests(t *testing.T, props map[string]string, store configuration
...
 	t.Run("init", func(t *testing.T) {
-		err := store.Init(context.Background(), configuration.Metadata{
+		// Initializing config updater. It has to be initialized before the store to create the table
+		err := updater.Init(props)
+		assert.Nil(t, err)
+
+		// Creating trigger for postgres config updater
+		if component == postgres {
+			err = updater.(*postgres_updater.ConfigUpdater).CreateTrigger(pgNotifyChannel)
+			assert.Nil(t, err)
```
ItalyPaleAle, line 157:
```suggestion
			assert.NoError(t, err)
```

ItalyPaleAle (changes requested):

configuration/postgres/postgres.go
shivamkm07: This is because table names in pg_tables are stored in lowercase only, and SELECT clause does the case sensitive comparison, so otherwise the query output is always false:

image

tests/conformance/configuration/configuration.go
shivamkm07: Done

tests/conformance/configuration/configuration.go
```diff
@@ -142,13 +147,21 @@ func ConformanceTests(t *testing.T, props map[string]string, store configuration
 	processedC3 := make(chan *configuration.UpdateEvent, keyCount*4)
 
 	t.Run("init", func(t *testing.T) {
-		err := store.Init(context.Background(), configuration.Metadata{
+		// Initializing config updater. It has to be initialized before the store to create the table
+		err := updater.Init(props)
+		assert.Nil(t, err)
```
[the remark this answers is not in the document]
shivamkm07: Done

configuration/postgres/postgres.go
shivamkm07: @ItalyPaleAle Please note that this would be a breaking change in postgres metadata. cc @mukundansundar

configuration/postgres/postgres.go
```diff
@@ -162,42 +163,40 @@ func (p *ConfigurationStore) Get(ctx context.Context, req *configuration.GetRequ
 }
 
 func (p *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.SubscribeRequest, handler configuration.UpdateHandler) (string, error) {
-	var pgNotifyChannels []string
```
shivamkm07, line 165: Removing the list as there could be only one channel passed in metadata in one subscribe call.

configuration/postgres/postgres.go
```diff
@@ -388,32 +376,23 @@ func validateInput(keys []string) error {
...
-		go p.doSubscribe(ctx, req, handler, pgNotifyCmd, channel, subscribeID, stop)
+	pgNotifyCmd := fmt.Sprintf(listenTemplate, pgNotifyChannel)
+	stop := make(chan struct{})
+	subscribeUID, err := uuid.NewRandom()
+	if err != nil {
+		return "", fmt.Errorf("unable to generate subscription id - %w", err)
+	}
+	subscribeID = subscribeUID.String()
+	p.subscribeStopChanMap[subscribeID] = stop
+	p.ActiveSubscriptions[subscribeID] = &subscription{
```
shivamkm07, line 391: Making `ActiveSubscriptions` with subscribeIDs as key instead of channel, because there could be multiple subscription IDs for a common channel (multiple subscribe calls with same channel), but there would be a unique subscriptionID for a given subscribe call with a specific channel

tests/conformance/configuration/configuration.go
```diff
@@ -203,12 +216,16 @@ func ConformanceTests(t *testing.T, props map[string]string, store configuration
 	}
 
 	if config.HasOperation("subscribe") {
+		subscribeMetadata := make(map[string]string)
+		if component == postgres {
```
shivamkm07, line 220: Ideally Conformance tests should not have component specific code as they test the interface, but subscribe call for postgres required channel in the metadata and so it's added here

configuration/postgres/postgres.go
ItalyPaleAle: It's an alpha component so ok to add breaking changes.

.github/scripts/test-info.mjs
```diff
@@ -185,6 +185,10 @@ const components = {
         certification: true,
         sourcePkg: ['configuration/redis', 'configuration/redis/internal'],
     },
+    'configuration.postgres': {
```
ItalyPaleAle, line 188: Please run `make prettier-format`. A bit of a nit but this should go before `configuration.redis.v6` to keep alphabetic order :)

configuration/postgres/postgres.go
```diff
@@ -162,42 +163,40 @@ func (p *ConfigurationStore) Get(ctx context.Context, req *configuration.GetRequ
 }
 
 func (p *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.SubscribeRequest, handler configuration.UpdateHandler) (string, error) {
-	var pgNotifyChannels []string
+	pgNotifyChannel := ""
 	for k, v := range req.Metadata {
-		if res := strings.EqualFold(pgNotifyChannelKey, k) && !slices.Contains(pgNotifyChannels, v); res {
-			pgNotifyChannels = append(pgNotifyChannels, v) // append unique channel names only
+		if strings.EqualFold(pgNotifyChannelKey, k) {
+			pgNotifyChannel = v
```
ItalyPaleAle, line 169: Add a `break` after the match to short-circuit

configuration/postgres/postgres.go
```diff
@@ -162,42 +163,40 @@ func (p *ConfigurationStore) Get(ctx context.Context, req *configuration.GetRequ
 }
 
 func (p *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.SubscribeRequest, handler configuration.UpdateHandler) (string, error) {
-	var pgNotifyChannels []string
+	pgNotifyChannel := ""
 	for k, v := range req.Metadata {
-		if res := strings.EqualFold(pgNotifyChannelKey, k) && !slices.Contains(pgNotifyChannels, v); res {
-			pgNotifyChannels = append(pgNotifyChannels, v) // append unique channel names only
+		if strings.EqualFold(pgNotifyChannelKey, k) {
```
ItalyPaleAle, line 168: Since these keys are all ASCII-only (or anyways we don't care if they're not ASCII), please do:

```suggestion
		if strings.ToLower(k) == pgNotifyChannelKey {
```

The linter will complain so you may need to add `//nolint:gocritic`

> strings.EqualFold is useful when the strings contain non-ASCII chars, but in this case is just slower.

configuration/postgres/postgres.go
```diff
@@ -162,42 +163,40 @@ func (p *ConfigurationStore) Get(ctx context.Context, req *configuration.GetRequ
...
-				}
-				delete(p.ActiveSubscriptions, k)
-				return nil
-			}
+	if sub, ok := p.ActiveSubscriptions[req.ID]; ok {
+		if oldStopChan, ok := p.subscribeStopChanMap[req.ID]; ok {
+			delete(p.subscribeStopChanMap, req.ID)
+			close(oldStopChan)
 		}
+		pgChannel := fmt.Sprintf(unlistenTemplate, sub.channel)
```
ItalyPaleAle, line 186: Since the constants `unlistenTemplate` and `listenTemplate` are only used once, please inline them. It makes the code a lot more readable...

(Also consider using simple string concatenation: `pgChannel := "UNLISTEN " + sub.channel` )

configuration/postgres/postgres.go
```diff
@@ -162,42 +163,40 @@ func (p *ConfigurationStore) Get(ctx context.Context, req *configuration.GetRequ
...
-				defer conn.Release()
-				_, err = conn.Exec(ctx, pgChannel)
-				if err != nil {
-					p.logger.Errorf("error un-listening to channel:", err)
-					return fmt.Errorf("error un-listening to channel: %w", err)
-				}
-				delete(p.ActiveSubscriptions, k)
-				return nil
-			}
+	if sub, ok := p.ActiveSubscriptions[req.ID]; ok {
```
ItalyPaleAle, line 181: Please short-circuit this instead:

```go
if p.ActiveSubscriptions[req.ID] == "" {
    return fmt.Errorf("unable to find subscription with ID : %v", req.ID)
}

// Rest of the code
```

This too makes the code more readable by reducing the number of indented blocks and by making the exit condition more immediate.

tests/conformance/configuration/configuration.go
```diff
@@ -142,13 +147,21 @@ func ConformanceTests(t *testing.T, props map[string]string, store configuration
 	processedC3 := make(chan *configuration.UpdateEvent, keyCount*4)
 
 	t.Run("init", func(t *testing.T) {
-		err := store.Init(context.Background(), configuration.Metadata{
+		// Initializing config updater. It has to be initialized before the store to create the table
+		err := updater.Init(props)
+		assert.NoError(t, err)
```
ItalyPaleAle, line 152:
```suggestion
		require.NoError(t, err)
```

tests/conformance/configuration/configuration.go
```diff
@@ -142,13 +147,21 @@ func ConformanceTests(t *testing.T, props map[string]string, store configuration
...
+
+		// Creating trigger for postgres config updater
+		if component == postgres {
```
ItalyPaleAle, line 155: Please inline the constants here

tests/conformance/configuration/configuration.go
```diff
@@ -142,13 +147,21 @@ func ConformanceTests(t *testing.T, props map[string]string, store configuration
...
+			err = updater.(*postgres_updater.ConfigUpdater).CreateTrigger(pgNotifyChannel)
+			assert.NoError(t, err)
```
ItalyPaleAle, line 157:
```suggestion
			require.NoError(t, err)
```

tests/conformance/configuration/configuration.go
```diff
@@ -142,13 +147,21 @@ func ConformanceTests(t *testing.T, props map[string]string, store configuration
...
+
+		// Initializing store
+		err = store.Init(context.Background(), configuration.Metadata{
 			Base: metadata.Base{Properties: props},
 		})
-		assert.Nil(t, err)
-		// Initializing config updater
-		err = updater.Init(props)
-		assert.Nil(t, err)
+		assert.NoError(t, err)
```
ItalyPaleAle, line 164:
```suggestion
		require.NoError(t, err)
```

tests/conformance/configuration/configuration.go
```diff
@@ -142,13 +147,21 @@ func ConformanceTests(t *testing.T, props map[string]string, store configuration
...
 	})
```
ItalyPaleAle, line 166: Here you may also want to add a line that makes the conf test exit if the init failed. If the component couldn't be initialized, there's no point in trying to run the tests.

```go
if t.Failed() {
    t.Fatal("initialization failed")
}
```

tests/utils/configupdater/postgres/postgres.go
```diff
@@ -0,0 +1,186 @@
...
+
+	"github.com/dapr/components-contrib/configuration"
+	"github.com/dapr/components-contrib/tests/utils/configupdater"
+	"github.com/dapr/kit/logger"
+)
+
+const (
+	connectionStringKey          = "connectionString"
+	configtablekey               = "table"
+	ErrorMissingConnectionString = "missing postgreSQL connection string"
```
ItalyPaleAle, line 19: Please inline these constants since they're strings and used just once.

There's one scenario where it makes sense to define errors as global variables, and that's when:

1. They are defined as errors and not strings, so `var ErrMissingConnectionString = errors.New("missing postgreSQL connection string")`
2. This makes it possible for callers that receive an error to then check if the error matches (even if wrapped). So they can do `errors.Is(err, postgres.ErrorMissingConnectionString)`

I don't think this is necessary here so these can just be inlined as strings.

configuration/postgres/postgres.go
```diff
@@ -361,19 +363,8 @@ func buildQuery(req *configuration.GetRequest, configTable string) (string, []in
...
-				return sub.uuid, true
-			}
-		}
-	}
-	return " ", false
-}
-
 func (p *ConfigurationStore) isSubscribed(subscriptionID string, channel string, key string) bool {
-	if val, yes := p.ActiveSubscriptions[channel]; yes && val.uuid == subscriptionID && slices.Contains(val.keys, key) {
+	if val, yes := p.ActiveSubscriptions[subscriptionID]; yes && val.channel == channel && (slices.Contains(val.keys, key) || len(val.keys) == 0) {
```
ItalyPaleAle, line 367:
```suggestion
    val := p.ActiveSubscriptions[subscriptionID]
	if val.channel == channel && (len(val.keys) == 0 || slices.Contains(val.keys, key)) {
```

ItalyPaleAle (changes requested):

configuration/postgres/postgres.go
```diff
@@ -152,55 +179,51 @@ func (p *ConfigurationStore) Get(ctx context.Context, req *configuration.GetRequ
 	if err != nil {
 		return nil, fmt.Errorf("unable to parse response from configuration store - %w", err)
 	}
-	result := make(map[string]*configuration.Item)
-	for _, v := range items {
-		result[v.key] = v.item
-	}
+	result := getUniqueItemPerKey(items)
```
shivamkm07, line 182: This is to return the `latest` version of the key if there are multiple versions present in the config store as per discussed in https://github.com/dapr/components-contrib/issues/2187
cc @ItalyPaleAle

configuration/postgres/postgres.go
```diff
@@ -288,18 +320,18 @@ func parseMetadata(cmetadata configuration.Metadata) (metadata, error) {
...
 
 	if m.ConnectionString == "" {
-		return m, fmt.Errorf(ErrorMissingConnectionString)
+		return m, fmt.Errorf("missing postgreSQL connection string")
 	}
 
 	if m.ConfigTable != "" {
-		if !allowedChars.MatchString(m.ConfigTable) {
-			return m, fmt.Errorf("invalid table name : '%v'. non-alphanumerics are not supported", m.ConfigTable)
+		if !allowedTableNameChars.MatchString(m.ConfigTable) {
```
ItalyPaleAle, line 327: (Not going to block the PR on this) FYI you can use a single regex to validate the charset and length: https://regex101.com/r/hNxNfd/1

ItalyPaleAle:
Hi @shivamkm07 thanks for this PR. I will merge it for now to unblock you, but I'd like to ask that you please make a follow-up PR to address some perf issues.

**First**, in `getUniqueItemsPerKey` you are already iterating through the `res` array once. 

https://github.com/shivamkm07/components-contrib/blob/db91a2b742fae876a871ccebdfb9747927e21074/configuration/postgres/postgres.go

What you want to do is:

1. Initialize `items := make(map[string]*configurationItem)` before the first loop
1. While you iterate through `res` the first time (in `for _, r := range res`):
   - If `items[r.key] == nil` is true, then pick this item as the one for `r.key`
   - If there's already an item,  in `items[r.key]`, right there you check if `items[r.key].Version < r.Version`. If yes, then you replace `items[r.key]` with this item

This will allow both having a single iteration over the loop, and saves allocating a lot of temporary memory.

**Second** when you compare versions in `compareVersion`, every time you're converting from string to int:

https://github.com/shivamkm07/components-contrib/blob/db91a2b742fae876a871ccebdfb9747927e21074/configuration/postgres/postgres.go

`strconv.Atoi` is fast, but not free.

So when you iterate through the `res` array, you should also have a `map[string]int` where you store the integer value of the the latest version of the item. So then you don't have to re-convert the same version from string to int on each iteration.

berndverst (dismissed):
@ItalyPaleAle please do not merge until perf improvements have been made in *this* PR. Follow up PRs never get done in a timely manner and I don't see a reason for doing so.

shivamkm07:
> Hi @shivamkm07 thanks for this PR. I will merge it for now to unblock you, but I'd like to ask that you please make a follow-up PR to address some perf issues.
> 
> **First**, in `getUniqueItemsPerKey` you are already iterating through the `res` array once.
> 
> https://github.com/shivamkm07/components-contrib/blob/db91a2b742fae876a871ccebdfb9747927e21074/configuration/postgres/postgres.go
> 
> What you want to do is:
> 
> 1. Initialize `items := make(map[string]*configurationItem)` before the first loop
> 2. While you iterate through `res` the first time (in `for _, r := range res`):
>    
>    * If `items[r.key] == nil` is true, then pick this item as the one for `r.key`
>    * If there's already an item,  in `items[r.key]`, right there you check if `items[r.key].Version < r.Version`. If yes, then you replace `items[r.key]` with this item
> 
> This will allow both having a single iteration over the loop, and saves allocating a lot of temporary memory.
> 
> **Second** when you compare versions in `compareVersion`, every time you're converting from string to int:
> 
> https://github.com/shivamkm07/components-contrib/blob/db91a2b742fae876a871ccebdfb9747927e21074/configuration/postgres/postgres.go
> 
> `strconv.Atoi` is fast, but not free.
> 
> So when you iterate through the `res` array, you should also have a `map[string]int` where you store the integer value of the the latest version of the item. So then you don't have to re-convert the same version from string to int on each iteration.

Done

ItalyPaleAle (approved):

PR #2747, opened 2023-04-04
Refactor state store bulk operations and add some native implementations

ItalyPaleAle:
Companion PR in runtime: dapr/dapr#6186

# Description

This PR refactors how state store components handle bulk state store operations:

1. Changes the default BulkSet/BulkDelete implementation to use transactions (i.e. the Multi method) when the component supports that. This makes sure that operations are performed atomically when possible, and should allow for better performance. When the component is not transactional, it falls back to performing them in sequence.
2. Added a BulkGet implementation in the default bulk store.
   - Previously, the runtime implemented this in [three](https://github.com/dapr/dapr/blob/fa7a8d2989b701ceaa9bfba656fc98dd5c3cce1f/pkg/actors/actors.go#L1522) [different](https://github.com/dapr/dapr/blob/fa7a8d2989b701ceaa9bfba656fc98dd5c3cce1f/pkg/grpc/api.go#L611) [places](https://github.com/dapr/dapr/blob/fa7a8d2989b701ceaa9bfba656fc98dd5c3cce1f/pkg/http/api.go#L682), which had some differences and were hard to maintain (see bugs like [these](https://github.com/dapr/dapr/pull/6134)). At least one of them also had a TODO saying it should be refactored into components-contrib
3. The shared implementation gets all elements in sequence, but components that can perform bulk get operations are encouraged to do so.
   - Some components have the implementation added in this PR, others were added [previously](https://github.com/dapr/components-contrib/pull/2704) (SQLite, Postgres, CockroachDB)
4. Another change is in the internal interface used by Dapr to send the list of transactional requests, so we don't need to perform costly conversions using `mapstructure` anymore. 

It also adds support for native BulkGet and/or BulkSet/BulkDelete to these components:

- Cosmos DB (given its unique quirks, I've updated the certification tests to validate them)
- MongoDB
- In-memory

Lastly, it fixes some bugs with Multi() in the Redis state store.

The changes to the components have to be included in this PR or tests will fail.

## Issue reference

Related to dapr/dapr#6121


state/azure/cosmosdb/cosmosdb.go
```diff
@@ -228,43 +226,214 @@ func (c *StateStore) Get(ctx context.Context, req *state.GetRequest) (*state.Get
...
+					Key:   item.ID,
+					Error: err.Error(),
+				}
+				n++
+				continue
+			}
+
+			// If the item contains a partition key, ensure it's the correct one
+			// This should never happen unless there's some issue
+			if item.PartitionKey != "" && item.PartitionKey != pk {
```
ItalyPaleAle, line 304: Reviewers: this actually happened during testing: https://github.com/Azure/azure-sdk-for-go/issues/20543

Unclear why

state/in-memory/in_memory.go
```diff
@@ -174,78 +174,72 @@ func (store *inMemoryStore) BulkDelete(ctx context.Context, req []state.DeleteRe
...
-	}
-	return time.Now().UnixMilli() > *item.expire
-}
-
-func (store *inMemoryStore) BulkGet(ctx context.Context, req []state.GetRequest) (bool, []state.BulkGetResponse, error) {
-	return false, nil, nil
-}
-
-func (store *inMemoryStore) marshal(v any) (bt []byte, isBinary bool, err error) {
+func (store *inMemoryStore) marshal(v any) (bt []byte, err error) {
```
ItalyPaleAle, line 232: This code has been simplified since we don't need to marshal the data to JSON for then un-marshalling it when retrieving it. The data is kept in memory so that was an extra step.

go.mod
```diff
@@ -403,3 +406,5 @@ replace github.com/toolkits/concurrent => github.com/niean/gotools v0.0.0-201512
 
 // this is a fork which addresses a performance issues due to go routines
 replace dubbo.apache.org/dubbo-go/v3 => dubbo.apache.org/dubbo-go/v3 v3.0.3-0.20230118042253-4f159a2b38f3
+
+replace github.com/dapr/dapr => github.com/italypaleale/dapr v1.6.1-0.20230404165709-453bccb6c007
```
berndverst, line 410: Why here? I would understand `tests/certification/go.mod` but we never include and pin dapr/dapr here.

ItalyPaleAle: I just used the command in the Makefile to add the replace. I can remove it (I don't think it's used anyways)

ItalyPaleAle: Done

berndverst (approved):

artursouza:
I would expect a PR this big that makes some bold decisions to have some discussions before reaching the runtime. If it happened outside of GitHub, please, let's try to bring those here.

PR #2748, opened 2023-04-04
Cert tests: use a single go.mod for all certification tests

ItalyPaleAle:
# Description

As of today, there are 48 `go.mod` (and as many `go.sum`) files in this repository, which cause four consequences:

1. Makes it hard to replace dependencies when needed, for example to use a fork of dapr/dapr in cert tests
2. Causes all PRs that change any dependency to have hundreds of lines changed, which add noise and make the PRs harder to review
  - Case in point: this PR changes almost 46,000 LOCs! 
4. When there's a merge conflict in those go.mod files, resolving them manually is not fun
5. It makes gopls (the VS Code Go language server) very unhappy when you try to work on cert tests

The main cause for this is that there's a `go.mod` in each certification test. As we are adding more tests, this is starting to become hard to maintain.

This PR changes that to have a single `go.mod` for all cert tests. It brings the number of `go.mod` files in this repo down to just 7 (still a lot, but much easier to maintain)


mukundansundar (approved):
lgtm. 
one question, do we need to remove the diff check?

ItalyPaleAle:
> one question, any specific reason to remove diff check?

Tests would fail otherwise as it expects to find a `go.mod` inside each component's folder

That check wasn't really necessary anyways, since the main CI pipeline is already doing `make modtidy-all` and checking for diffs

PR #2751, opened 2023-04-05
Workflow management API cleanup / breaking changes for v1.11

cgillum:
# Description

This PR contains a set of breaking changes to improve the usability of the workflow management APIs. These changes are expected to go out with Dapr v1.11.

Summary of changes:

* (**breaking**) Updated APIs to follow standard request/response pattern
* (**breaking**) Switched JSON fields from _snake_case_ to _camelCase_
* (**breaking**) Changed data payload types from `interface{}` to `[]byte`
* Added more fields to `WorkflowState`
* Renamed "Pause" to "Suspend" to match existing request/response struct naming
* Updated Temporal component implementation

There is only one external component that implements this interface (Temporal) but this component has not yet shipped so it's not actually a breaking change for any shipped contributed components.

Here is the PR in `dapr/dapr` to implement the updated workflow management API: https://github.com/dapr/dapr/pull/6218

## Issue reference

Part 1 of [Improvements to existing workflow management APIs (dapr/dapr#5911)](https://github.com/dapr/dapr/issues/5911)

## Checklist

* [ ] Code compiles correctly
* [x] Created/updated tests


workflows/requests.go
```diff
@@ -1,20 +1,36 @@
...
 
-// RaiseEventRequest is the object describing a Raise Event request.
+// RaiseEventRequest is the struct describing a raise workflow event request.
 type RaiseEventRequest struct {
-	InstanceID string `json:"workflow_reference"`
-	EventName  string `json:"event_name"`
-	Input      []byte `json:"input"`
+	InstanceID string `json:"instanceID"`
+	EventName  string `json:"name"`
+	EventData  []byte `json:"data"`
```
ItalyPaleAle, line 25: Given the comment "We assume all inputs are JSON values" please consider using `json.RawValue` here. Two reasons:

First: It allows people to put valid JSON in there, but Go won't unmarshal it. So they don't need to encode it as string, they can pass valid JSON. For example:

```json
{
    "instanceID": "foo",
    "name": "bar",
    "data": {
        "some": ["random", "stuff", 42]
    }
}
```

Using the above with `[]byte` would fail because "data" is not a base64-encoded string. Users would have to input JSON as a base64-encoded string instead, which is not as nice.

Second, while json.RawMessage isn't parsed, it's still validated when doing json.Unmarshal to make sure it's valid JSON (so parentheses are closed, etc)

cgillum: Thanks for the review! This is actually really good info for me. However, the comment you're referring to is strictly for the Temporal component. It wasn't my intention that we assume this is the case for all workflow components.
One thing I haven't looked at closely, however, is how this impacts the HTTP API in `dapr/dapr`. I may revisit this if I think it makes the API too unweildy.

ItalyPaleAle: Do you think it may be a fair assumption to make all components accept JSON?

Yes, it would be a change in the runtime too. I would personally welcome this as a good step however :)

cgillum: After spending some time on the HTTP API handler, I feel that my use of `[]byte` here is correct.

The plan is to take the entire HTTP request payload as the event data without making any assumptions about its format. It could be JSON, it could be plain text, it could be binary data - the API doesn't care. The other fields you mentioned in your example (`instanceID` and `name`) will come from the request URL.

This means that the caller can send any payload they want in the HTTP request and don't have to do any encoding on their side to satisfy our API layer. They just need to make sure that any data they send matches what the component implementation expects. Only the component will attempt to decode the event data.

This is the same for my changes to the `StartRequest` type, which also makes no assumptions about the format of the workflow input.

ItalyPaleAle: Ok, that makes sense then and it's fine with me

cgillum:
FYI @RyanLettieri

workflows/temporal/temporal.go
```diff
@@ -106,42 +113,52 @@ func (c *TemporalWF) Terminate(ctx context.Context, req *workflows.WorkflowRefer
...
+			Properties: map[string]string{
+				"task_queue": resp.WorkflowExecutionInfo.GetTaskQueue(),
+			},
 		},
 	}
 
 	return &outputStruct, nil
 }
 
 func (c *TemporalWF) RaiseEvent(ctx context.Context, req *workflows.RaiseEventRequest) error {
```
RyanLettieri, line 144: Even though temporal is not currently implemented, should we still be updating the "LastUpdatedAt" field inside this method?

workflows/temporal/temporal.go
```diff
@@ -106,42 +113,52 @@ func (c *TemporalWF) Terminate(ctx context.Context, req *workflows.WorkflowRefer
...
 	outputStruct := workflows.StateResponse{
-		WFInfo:    workflows.WorkflowReference{InstanceID: req.InstanceID},
-		StartTime: resp.WorkflowExecutionInfo.StartTime.Format(time.RFC3339),
-		Metadata: map[string]string{
-			"task_queue": resp.WorkflowExecutionInfo.GetTaskQueue(),
-			"status":     lookupStatus(resp.WorkflowExecutionInfo.Status),
+		Workflow: &workflows.WorkflowState{
+			InstanceID:    req.InstanceID,
+			CreatedAt:     createdAtTime,
+			LastUpdatedAt: createdAtTime,
```
RyanLettieri, line 133: For the "LastUpdatedAt" field, what're your thoughts on including more than just the time? What if we included how it was updated? For instance, we could have the time as well as the 'culprit' such as "x event was raised" or "pause/resume was called at this time"

workflows/temporal/temporal.go
```diff
@@ -106,42 +113,52 @@ func (c *TemporalWF) Terminate(ctx context.Context, req *workflows.WorkflowRefer
...
+			RuntimeStatus: lookupStatus(resp.WorkflowExecutionInfo.Status),
+			Properties: map[string]string{
+				"task_queue": resp.WorkflowExecutionInfo.GetTaskQueue(),
```
ItalyPaleAle, line 136: Should this use camelCase too? Unless it needs to use snake_case for compatibility with some external service

workflows/requests.go
```diff
@@ -1,20 +1,36 @@
 package workflows
 
-type WorkflowReference struct {
-	InstanceID string `json:"instance_id"`
+// StartRequest is the struct describing a start workflow request.
+type StartRequest struct {
```
ItalyPaleAle, line 4: Just for my understanding, are these structs used in the runtime too?

cgillum: Yes, in the runtime's universal API handlers, we create instances of these structs and pass them to the component implementations.

workflows/temporal/temporal.go
cgillum: I think something like a "last action", or something else that gives a more description of the last thing the workflow did would be super useful. It wouldn't need to be included in `lastUpdatedAt`, though, and would probably be component-specific. My recommendation for now would be to include that kind of detailed info in the `Properties` dictionary.

workflows/temporal/temporal.go
cgillum: It can be any form that the component implementation wants. For this Temporal component, the actual form doesn't matter, as long as it's documented. I was just copying what we had previously for v1.10 (which didn't actually ship with this), but it can be changed. In fact, if we want to change it, I might even namespace it to something like `temporalio.taskQueue` instead of `task_queue` or `taskQueue` to make it more clear that it's a Temporal-specific property.

workflows/temporal/temporal.go
cgillum: `LastUpdatedAt` only applies to the `GetWorkflow` API. What will normally happen, however, is that someone will raise an event to a Temporal workflow, the Temporal workflow will respond to that event, and then the temporal engine will update the workflow's last updated time automatically. We don't need to do anything for that to happen here. The next time someone calls `GetWorkflow`, the returned workflow state struct will automatically have the updated `LastUpdatedAt` value.

workflows/temporal/temporal.go
berndverst: We will probably delete the temporal component code in the future given we are not using this component.

berndverst (changes requested):
This PR cannot currently be merged - you must not currently replace the Runtime replacement in certification tests that @ItalyPaleAle made due to crypto components. First this PR must be merged - https://github.com/dapr/dapr/pull/6197

Then you must include those changes in your runtime fork.

berndverst:
This PR is currently blocked by https://github.com/dapr/dapr/pull/6197

As a quick fix you can manually include the commits from https://github.com/dapr/dapr/pull/6197 in your runtime fork and then at the commits from your fork on top of that, then update the runtime replacement here.

tests/certification/go.mod
```diff
@@ -296,12 +295,13 @@ require (
...
+	sigs.k8s.io/controller-runtime v0.14.6 // indirect
 	sigs.k8s.io/json v0.0.0-20220713155537-f223a00ba0e2 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
 replace github.com/dapr/components-contrib => ../../
 
-replace github.com/dapr/dapr => github.com/italypaleale/dapr v1.6.1-0.20230404165709-453bccb6c007
+// TEMPORARY: Needed to unblock certification tests; can revert once https://github.com/dapr/dapr/pull/6218 is merged
```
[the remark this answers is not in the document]
cgillum: I've removed this now that the crypto changes have been merged in `dapr/dapr`.

berndverst (changes requested):
Sometimes the latest runtime being brought in has a different signature than our test package in `tests/certification/embedded/embedded.go`

You need to update this package to match the runtime options actually present in dapr runtime.

```
# github.com/dapr/components-contrib/tests/certification/embedded
Error: ../../embedded/embedded.go:151:3: unknown field AppSSL in struct literal of type "github.com/dapr/dapr/pkg/runtime".NewRuntimeConfigOpts
Error: ../../embedded/embedded.go:226:91: cannot use string(runtimeConfig.ApplicationProtocol) (value of type string) as bool value in argument to acl.ParseAccessControlSpec
```

cgillum:
@berndverst certification tests are happy now. :) Thanks for your help on getting this unblocked!

berndverst (approved):

PR #2753, opened 2023-04-06
Use consistent `Close` behaviour in crypto/jwks

JoshVanL:
Use `closeCh` channel to signal `Close()`. Wait for all go routines to return before returning from `Close()`.


crypto/jwks/component.go
```diff
@@ -159,21 +173,30 @@ func (k *jwksCrypto) initJWKSFromURL(ctx context.Context, url string) error {
...
+	loadedCh := make(chan error, 1) // Needs to be buffered to prevent an (unlikely, but possible) goroutine leak
+
+	k.wg.Add(2)
+
 	go func() {
-		watchErr := fswatcher.Watch(k.ctx, path, eventCh)
+		defer k.wg.Done()
+		watchErr := fswatcher.Watch(ctx, path, eventCh)
 		if watchErr != nil && !errors.Is(watchErr, context.Canceled) {
 			// Log errors only
```
JoshVanL, line 192: It feels like we should be failing the component when there is an error here.

crypto/jwks/component.go
```diff
@@ -94,11 +100,20 @@ func (k *jwksCrypto) Features() []contribCrypto.Feature {
...
 
 // Init the JWKS object from the metadata property
-func (k *jwksCrypto) initJWKS(ctx context.Context) error {
+func (k *jwksCrypto) initJWKS() error {
 	if len(k.md.JWKS) == 0 {
 		return errors.New("metadata property 'jwks' is required")
 	}
 
+	// Use context based on close channel.
+	ctx, cancel := context.WithCancel(context.Background())
```
mukundansundar, line 109: Shouldn't close also be based on incoming context?

JoshVanL: I think so, but this is what we have currently so left it.

JoshVanL: Whoops, that's not true at all. Sorry.

JoshVanL: Fixed

JoshVanL: I have reverted that change because we don't want the context by the `cache`, `fswatcher` etc. to be the same as the `Init` context. If it was and the runtime caller adds a timeout, for example, it will cancel the processes. By using a Background, the runtime can control the process shutdown with `Close` instead.

mukundansundar (commented):
@ItalyPaleAle and @berndverst 
`Init` method has a context passed in from runtime. Shouldn't that be stored in the `jwksCrypto` struct and that context be used for all ops in this component?

JoshVanL:
@mukundansundar stdlib explicitly calls out that you should not store contexts in structs.

```
Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx: 
```

https://pkg.go.dev/context

crypto/jwks/component.go
```diff
@@ -94,11 +100,20 @@ func (k *jwksCrypto) Features() []contribCrypto.Feature {
...
+	k.wg.Add(1)
+	go func() {
```
ItalyPaleAle, line 111: The problem with not using a component-scoped context is that now we have an extra goroutine (which apparently is a big deal lately :) )

crypto/jwks/component.go
```diff
@@ -82,10 +83,15 @@ func (k *jwksCrypto) Init(ctx context.Context, metadata contribCrypto.Metadata)
 
 // Close implements the io.Closer interface to close the component
 func (k *jwksCrypto) Close() error {
-	if k.cancel != nil {
-		k.cancel()
+	defer k.wg.Wait()
```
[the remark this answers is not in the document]
JoshVanL: Btw, that is the intended sequence. I can drop the `defer` and more to after `Clear()`.

crypto/jwks/component.go
ItalyPaleAle: Perhaps, but what if the error happens after the component has started?

crypto/jwks/component.go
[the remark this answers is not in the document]
ItalyPaleAle: Yes I understand that it's intended. But when defer's are used for things that are not cleanup related (like, `defer res.Body.Close()`) and actually impact the flow, I do think they hurt the readability of the code, as then someone has to remember that there was a deferred call.

crypto/jwks/component.go
JoshVanL: The component should be closed imo.

ItalyPaleAle: I don't think we should cause a panic in the entire runtime for this. And closing the component would mean that the app would stop working until it's manually restarted.

When this fails btw the component can continue to work with the keys that have been loaded in memory, it just can't pick up new keys.

JoshVanL: I'm not suggesting that the entire runtime to panic, just that this component to not be in a ready state and usable. The rest of the dapr runtime that does not rely on this should continue to function.

> When this fails btw the component can continue to work with the keys that have been loaded in memory, it just can't pick up new keys.

I am concerned that this is going to lead to silent failure modes and cause larger disruption downstream which are difficult to debug.

ItalyPaleAle: > I'm not suggesting that the entire runtime to panic, just that this component to not be in a ready state and usable. The rest of the dapr runtime that does not rely on this should continue to function.

Wouldn't that cause silent failures too?

JoshVanL: > > I'm not suggesting that the entire runtime to panic, just that this component to not be in a ready state and usable. The rest of the dapr runtime that does not rely on this should continue to function.
> 
> Wouldn't that cause silent failures too?

We should have a mechanism for reporting health on a per component basis. This can be some combination of kube resource status, logs, metrics, error codes. Only printing an error log line is very difficult for a human to pick up, and impossible for a machine to self heal automatically.

ItalyPaleAle:
@JoshVanL thanks for this contribution. Since most of the code for this component was moved to dapr/kit (see dapr/kit#46), I have included this PR in #2789 (keeping your comments so you'll still get credited), where I also brought in the cache from dapr/kit.

I'm going to close this as superseded.

PR #2777, opened 2023-04-13
fixup: even numbers must be strings in metadata

berndverst:
# Description

FYI @tmacam - this fixes up your metadata contribution.

cc @ItalyPaleAle

The command `make bundle-component-metadata` did not work because the number value was a number and not a string of a number.

### DIFF
--- a/state/azure/blobstorage/metadata.yaml
+++ b/state/azure/blobstorage/metadata.yaml
@@ -80,8 +80,8 @@ metadata:
     # getBlobRetryCount is a deprecated alias for this field
     type: number
     required: false
-    default: 3
-    example: 3
+    default: "3"
+    example: "3"
     description: |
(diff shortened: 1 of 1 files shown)


mukundansundar (approved):

tmacam:
> even numbers must be strings in metadata

I was not expecting this one :) I will add this verification on a PR tomorrow as part of `make check-component-metadata`. I think that should be the appropriate place to add this verification.

ItalyPaleAle:
Thanks for catching that

tmacam:
Created https://github.com/dapr/components-contrib/issues/2780 to track on automatic ways to catch such errors in the future. Adding `make bundle-component-metadata`  as a GH validation WF step would be an easy way to catch the particular issue that this PR addressed but I don't know if we want to go further and perform full schema validation on `metadata.yaml` files

PR #2778, opened 2023-04-13
fix postgres error handling for upsert with etag

### LINKED ISSUE #2773: Postgres State Store does not return expected ETag mismatch error
## Background

(From [Discord](https://discord.com/channels/778680217417809931/781589820128493598/1095792573861351545))

Hey folks. following on from this comment here pluggable-components, I'm looking to see if my analysis is correct or not.

So when using the dotnet SDK to save some state with the TrySaveStateAsync method, when an etag is provided, but said etag it is wrong due to a mismatch, the SDK throws an Exception. The exception feels/looks unexpected based on looking at this code in the SDK, the commentary looks like the Method should just return false (and not throw a DaprException) when a Etag mismatch occurs, no?

https://github.com/dapr/dotnet-sdk/blob/release-1.10/src/Dapr.Client/DaprClientGrpc.cs#L955

However,  the stack trace from the exception says "no item was updated" 

```
Dapr.DaprException : State operation failed: the Dapr endpoint indicated a failure. See InnerException for details.
---- Grpc.Core.RpcException : Status(StatusCode="Internal", Detail="failed saving state in state store standard-postgres: no item was updated")
```

which tracks to this code here in the 1.10 release ....

https://github.com/dapr/components-contrib/blob/release-1.10/state/postgresql/postgresdbaccess.go#L225

Notice how the etag mismatch code path is not being followed? (it's just a few lines above, on line 219)

So whats going on here? Not sure what is right and what is wrong 🙂

Interestingly, I've contrasted this with the behaviour for deleting a key with a mismatched Etag, and the SDK returns false and doesn't throw an Exception, and the postgres code for delete is different too, which looks correct to my untrained eyes.

Switching back to the update behaviour for a second.... I don't get why this this etag mismatch code is predicated on an error being raised from the db? Surely its going to attempt to perform an update and return 0 rows, not raise a db error.

```golang
if err != nil {
    if req.ETag != nil && *req.ETag != "" {
        return state.NewETagError(state.ETagMismatch, err)
    }
    return err
}
```

https://github.com/dapr/components-contrib/blob/a0041cd51767d07bf194d903e7cb9d03e0b15e55/state/postgresql/postgresdbaccess.go#L217 

As @philliphoff pointed out in the other conversation, SQL lite looks like its exhibiting correct behaviour, by not predicating the etag mismatch codepath on a db  error having occurred.

```golang
if err != nil {
    return err
}

// Count the number of affected rows
rows, err := res.RowsAffected()
if err != nil {
    return err
}
if rows == 0 {
    if req.ETag != nil && *req.ETag != "" {
        return state.NewETagError(state.ETagMismatch, nil)
    }
    return errors.New("no item was updated")
}
```

https://github.com/dapr/components-contrib/blob/5dad8a867c62a6100a939093619b0d5bc620120e/state/sqlite/sqlite_dbaccess.go#L462

looks like the postgres code was refactored in how the database update result is handled, and the behaviour change was introduced, as part of this commit :

[https://github.com/dapr/components-contrib/pull/1118/files#diff-787703635a6e79760dd7984698b15668deb2aa7f95178287d3f740ec2569780](https://github.com/dapr/components-contrib/pull/1118/files#diff-787703635a6e79760dd7984698b15668deb2aa7f95178287d3f740ec2569780f)

## Expected Behavior

A ETag mismatch error should be raised when the user attempts to set a value with an old (but valid) ETag using the postgres state store.  When using the .NET Dapr Client SDK, the `TrySaveStateAsync()` method should return `false`.

## Actual Behavior

A general "no item was updated" error is raised.  When using the .NET Dapr Client SDK, an `RpcException` to be thrown by the `TrySaveStateAsync()` method.

- Run a Dapr application configured to use a postgres state store
- Set and then get a value several times (noting the ETag returned during each get request)
- Attempt to set the value using the ETag returned on the first set/get requests. 

## Release Note

RELEASE NOTE:

berndverst:
# Description

Fixes #2773

During upsert with etag, if no DB error is returned and rows affected is 0 an ETAG Mismatch error should be thrown.

This fixes that.

### DIFF
--- a/internal/component/postgresql/postgresdbaccess.go
+++ b/internal/component/postgresql/postgresdbaccess.go
@@ -216,13 +216,12 @@ func (p *PostgresDBAccess) doSet(parentCtx context.Context, db dbquerier, req *s
 
 	result, err := db.Exec(parentCtx, query, params...)
 	if err != nil {
-		if req.ETag != nil && *req.ETag != "" {
-			return state.NewETagError(state.ETagMismatch, err)
-		}
 		return err
 	}
-
 	if result.RowsAffected() != 1 {
+		if req.ETag != nil && *req.ETag != "" {
+			return state.NewETagError(state.ETagMismatch, nil)
+		}
 		return errors.New("no item was updated")
 	}
 
--- a/tests/certification/state/cockroachdb/cockroachdb_test.go
+++ b/tests/certification/state/cockroachdb/cockroachdb_test.go
@@ -138,7 +138,7 @@ func TestCockroach(t *testing.T) {
 		return nil
 	}
 
-	//ETag test
+	// ETag test
 	eTagTest := func(ctx flow.Context) error {
 		etag1 := "1"
 		etag100 := "100"
@@ -156,14 +156,14 @@ func TestCockroach(t *testing.T) {
 			Value: "v3",
 			ETag:  &etag100,
 		})
-		assert.Equal(t, fmt.Errorf("no item was updated"), err)
+		assert.Equal(t, state.NewETagError(state.ETagMismatch, nil), err)
 
 		resp, err := stateStore.Get(context.Background(), &state.GetRequest{
 			Key: keyOneString,
 		})
 		assert.NoError(t, err)
-		assert.Equal(t, etag1, *resp.ETag)           //1 is returned since the default when the new data is written is a value of 1
-		assert.Equal(t, "\"v1\"", string(resp.Data)) //v1 is returned since it was the only item successfully inserted with the key of keyOneString
+		assert.Equal(t, etag1, *resp.ETag)           // 1 is returned since the default when the new data is written is a value of 1
+		assert.Equal(t, "\"v1\"", string(resp.Data)) // v1 is returned since it was the only item successfully inserted with the key of keyOneString
 
 		// This will update the value stored in key K with "Overwrite Success" since the previously created etag has a value of 1
 		// It will also increment the etag stored by a value of 1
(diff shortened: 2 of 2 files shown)


internal/component/postgresql/postgresdbaccess.go
```diff
@@ -216,13 +216,12 @@ func (p *PostgresDBAccess) doSet(parentCtx context.Context, db dbquerier, req *s
 
 	result, err := db.Exec(parentCtx, query, params...)
 	if err != nil {
-		if req.ETag != nil && *req.ETag != "" {
-			return state.NewETagError(state.ETagMismatch, err)
-		}
 		return err
 	}
-
 	if result.RowsAffected() != 1 {
```
berndverst, line 221: This code matches another place in this file that specifically checks `rows != 1`. We could make it `rows == 0` but it doesn't really matter since `rows > 1` isn't possible in these cases.

philliphoff (approved):

ItalyPaleAle (approved):

tests/certification/state/cockroachdb/cockroachdb_test.go
```diff
@@ -156,14 +156,14 @@ func TestCockroach(t *testing.T) {
 			Value: "v3",
 			ETag:  &etag100,
 		})
-		assert.Equal(t, fmt.Errorf("no item was updated"), err)
+		assert.Equal(t, state.NewETagError(state.ETagMismatch, nil), err)
```
berndverst, line 159: @philliphoff FYI this proves that the correct error is now returned -- cockroachdb uses the postgres implementation for this.

ItalyPaleAle (approved):

PR #2782, opened 2023-04-14
Add component metadata YAML for Azure CosmosDB output binding.

tmacam:
# Description

Add component metadata YAML for Azure CosmosDB output binding.

Verified with `make check-component-metadata && make bundle-component-metadata`

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/bindings/azure/cosmosdb/metadata.yaml
+++ b/bindings/azure/cosmosdb/metadata.yaml
@@ -1,0 +1,67 @@
+# yaml-language-server: $schema=../../../component-metadata-schema.json
+schemaVersion: v1
+type: bindings
+name: azure.cosmosdb
+version: v1
+status: stable
+title: "Azure Cosmos DB (SQL API)"
+urls:
+  - title: Reference
+    url: https://docs.dapr.io/reference/components-reference/supported-bindings/cosmosdb/
+binding:
+  output: true
+  operations:
+    - name: create
+      description: "Create an item."
+capabilities: []
+builtinAuthenticationProfiles:
+  - name: "azuread"
+authenticationProfiles:
+  - title: "Master key"
+    description: |
+      Authenticate using a pre-shared "master key".
+    metadata:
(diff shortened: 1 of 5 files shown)


bindings/azure/cosmosdb/metadata.yaml
```diff
@@ -0,0 +1,67 @@
...
+    example: '"Orders"'
+  - name: partitionKey
+    type: string
+    required: true
+    description: |
+      The name of the key to extract from the payload (document to be created) that is used as the
+      partition key. This name must match the partition key specified upon creation of the
+      Cosmos DB container.
+    example: '"OrderId"'
+  - name: azureEnvironment
```
ItalyPaleAle, line 57: I think this is only for Azure AD

tmacam: I was wondering if that is something that we should implicitly define under `builtinAuthenticationProfiles "azuread"`.

ItalyPaleAle: uhm good point. would you mind including this change too?

Also I think the value is now case-insensitive so we can put a "LESS AGGRESSIVE" (i.e. not all caps):

- `AzurePublic`
- `AzureChina`
- `AzureUSGovernment`

tmacam: But anyway, specifically for CosmosDB, [this field is only read if authentication falls back to Azure AD(https://github.com/dapr/components-contrib/blob/3a616c5678442f3bf3e8ebeaf820f4c0a696c8e7/bindings/azure/cosmosdb/cosmosdb.go#L84-L88).

ItalyPaleAle: Yeah I do think this is only for AAD.

ItalyPaleAle (approved):

PR #2783, opened 2023-04-14
Add component metadata YAML for Azure Event Grid input/output binding.

tmacam:
# Description

Add component metadata YAML for Azure Event Grid input/output binding.

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


bindings/azure/eventgrid/metadata.yaml
```diff
@@ -0,0 +1,68 @@
...
+  input: true
+  output: true
+  operations:
+    - name: create
+      description: "Create an event subscription."
+capabilities: []
+builtinAuthenticationProfiles:
+  - name: "azuread"
+metadata:
+  # Required Input Binding Metadata
```
ItalyPaleAle, line 21: For things that are only-input, please add:

```yaml
binding:
  input: true
  output: false
```

Same for output-only.

Otherwise, the default is to match the top-level `binding` property, which is input and output

ItalyPaleAle (changes requested):

bindings/azure/eventgrid/metadata.yaml
tmacam: I hope I fixed this right on d675ab99f6b21933eb56ad1f39aaba92d9ab7601

berndverst:
FYI - I think our SDK might not support the distinction between required fields in input and in output bindings right now :(

berndverst (approved):

PR #2790, opened 2023-04-18
state.sqlserver: Add support for Azure AD auth (+ metadata.yaml)

### LINKED ISSUE #2726: Issues with SQL Server state store and MSI auth
_Originally posted by @giovanidecusati in https://github.com/dapr/components-contrib/issues/1103#issuecomment-1489897854_

We are trying to use "state.sqlserver" component configured to use User-Managed-Identity to connect to Azure SQL Server Database instance. Is that supported? 

I'm facing this issue when trying to connect to SQL:
`
[INIT_COMPONENT_FAILURE]: initialization error occurred for <componentName>-statestore (state.sqlserver/v1): failed to create db database: mssql: login error: Login failed for user '<MyUserManagedIdentityClientId>'." app_id=<appid> instance=<revisionInstance> scope=dapr.runtime type=log ver=1.10.4-msft-2
`

SqlConnectionString:
`
Server=<MySqlServerInstance>.database.windows.net;Authentication=**Active Directory Managed Identity**;Encrypt=True;User Id=<MyUserManagedIdentityClientId>;Database=<MyDatabaseName>
`

We are deploying via Bicep Tempalte:
`
resource resource_daprSecretStore 'Microsoft.App/managedEnvironments/daprComponents@2022-06-01-preview' = {
  name: statestoreName
  parent: resource_environment
  properties: {
    componentType: 'state.sqlserver'
    version: 'v1'
    metadata: [
      {
        name: 'connectionString'
        secretRef: keyVaultConenctionStringSecretName
      }
      {
        name: 'tableName'
        value: tableName
      }
      {
        name: 'schema'
        value: schemaName
      }
    ]
    scopes: [
      daprAppName
    ]
    secretStoreComponent: secretStoreName
  }
}
`

Appreciate any help, thanks!

ItalyPaleAle:
Adds support for Azure AD authentication in the `state.sqlserver` component.

Fixes #2726

1. First, the library we used is now deprecated and replaced by an official, Microsoft-maintained fork: https://github.com/microsoft/go-mssqldb - updated to use that library
2. Second, although the library did already support Azure AD auth (by importing a separate package), it did in a way that was not compatible with the shared Azure auth library all other components use.
   - It's now possible to authenticate with Azure AD by setting `useAzureAD: true` in the metadata.
   - The other Azure AD-related options are required as usual (like "azureClientID", etc)
   - A `connectionString` is still required to include the URL of the database, the DB name, and any connection option. If it contains credentials, they are ignored
3. This required doing some refactoring on the metadata parsing and struct
4. Also added the metadata.yaml 

Note: I confirmed tests pass locally, but I can't easily add them to the certification tests because it appears I may need to create a separate Azure SQL instance to use Azure AD. I can look into doing that in a subsequent PR


internal/authentication/azure/services.go
```diff
@@ -0,0 +1,36 @@
...
+	ServiceAzureSQL cloud.ServiceName = "dapr.io/azuresql"
+)
+
+func init() {
+	// Register additional services in the SDK's clouds
+	cloud.AzureChina.Services[ServiceAzureSQL] = cloud.ServiceConfiguration{
+		Audience: "https://database.chinacloudapi.cn",
+	}
+	cloud.AzureGovernment.Services[ServiceAzureSQL] = cloud.ServiceConfiguration{
+		Audience: "https://	database.usgovcloudapi.net",
```
DeepanshuA, line 31: Does it have intentional space?

state/sqlserver/metadata.go
```diff
@@ -0,0 +1,298 @@
...
+	keyColumnName        = "Key"
+	rowVersionColumnName = "RowVersion"
+	databaseNameKey      = "databaseName"
+	cleanupIntervalKey   = "cleanupIntervalInSeconds"
+
+	defaultKeyLength       = 200
+	defaultSchema          = "dbo"
+	defaultDatabase        = "dapr"
+	defaultTable           = "state"
+	defaultMetaTable       = "dapr_metadata"
```
DeepanshuA, line 50: These are some common default names across other components as well, like postgresql etc.
Should we rather make these constants parts of common constants file, like in https://github.com/dapr/components-contrib/tree/master/state/utils ?

state/sqlserver/metadata.yaml
```diff
@@ -0,0 +1,101 @@
...
+capabilities:
+  # If actorStateStore is present, the metadata key actorStateStore can be used
+  - "actorStateStore"
+  - "crud"
+  - "transactional"
+  - "etag"
+authenticationProfiles:
+  - title: "Connection string"
+    description: |
+      Authenticate using a connection string.
```
DeepanshuA, line 20:
```suggestion
      Authenticates using a connection string.
```

state/sqlserver/metadata.yaml
```diff
@@ -0,0 +1,101 @@
...
+    metadata:
+      - name: connectionString
+        required: true
+        sensitive: true
+        description: |
+          The connection string used to connect.
+          If the connection string contains the database it must already exist. Otherwise, if the database is omitted, a default database named "Dapr" is created.
```
DeepanshuA, line 27:
```suggestion
          If the connection string contains the database, it must already exist. Otherwise, if the database is omitted, a default database named "Dapr" is created.
```

state/sqlserver/metadata.yaml
```diff
@@ -0,0 +1,101 @@
...
+        type: bool
+        description: |
+          Must be set to `true` to enable the component to retrieve access tokens from Azure AD.
+          This authentication method only works with Azure SQL databases.
+      - name: connectionString
+        required: true
+        sensitive: true
+        description: |
+          The connection string or URL of the Azure SQL database, without credentials.
+          If the connection string contains the database it must already exist. Otherwise, if the database is omitted, a default database named "Dapr" is created.
```
DeepanshuA, line 44:
```suggestion
          If the connection string contains the database, it must already exist. Otherwise, if the database is omitted, a default database named "Dapr" is created.
```

DeepanshuA (commented):
Some nitpicks

state/sqlserver/metadata.go
ItalyPaleAle: For these ones specifically, my preference is on not doing that because each component can choose a different implementation.

internal/authentication/azure/services.go
ItalyPaleAle: Uh, how did that happen 😳

DeepanshuA (approved):

giovanidecusati:
Hi Team,

Just want to say thank you for the work you are doing and have considered implementing support for AAD for the SQL server state store. Looking forward to version 1.11! We are using Dapr on Azure Container App. I hope the Azure team doesn't take long to update to the new version.
Cheers,

PR #2796, opened 2023-04-19
Add component metadata YAML for bindings.redis.

tmacam:
# Description

Add component metadata YAML for bindings.redis.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


bindings/redis/metadata.yaml
```diff
@@ -0,0 +1,207 @@
...
+      Allows specifying a default Time-to-live (TTL) in seconds that will be
+      applied to every state store request unless TTL is explicitly defined
+      via the request metadata.
+    example: "600"
+    type: number
+  - name: queryIndexes
+    required: false
+    description: Indexing schemas for querying JSON objects
+    example: "see Querying JSON objects"
+    type: string
```
tmacam, line 207: If I comment all this out and run `make check-component-metadata && make bundle-component-metadata` the command only complains about the following properties:

```
{
  "bindings/redis": {
    "maxretries": "int",
    "maxretrybackoff": "time.Duration",
    "queryindexes": "string",
    "ttlinseconds": "*int"
  }
}
```

Like 🤯. What is happening here?

DeepanshuA: But, these are optional attributes, as far as I understand. Am I missing something?

DeepanshuA: But, all these mentioned 4 metadata are optional. Am I missing something?

tmacam: If I only define those 4 metadata and nothing else that is in this metadata.yaml file, `make check-component-data` would pass.

What is puzzling me is that these fields (or a structure hosting only these fields) are not defined in any of the structs used for metadata and settings by bindings/redis. 

> Am I missing something?

That is exactly my question. I am even wondering: am I missing something really obvious and should I instead define only these 4 fields `make check-component-data`  is requiring instead ?

tmacam: ALrigh! We found the issue and it is addressed on #2803.

Pending its merge before procedding. Moving this back to draft.

berndverst (changes requested):
```
{
  "bindings/redis": {
    "maxlenapprox": "int64",
    "redismaxretries": "int",
    "redismaxretryinterval": "redis.Duration"
  }
}
```

berndverst (approved):

PR #2801, opened 2023-04-20
Remove replace from cert tests go.mod

ItalyPaleAle:
Not needed anymore


ItalyPaleAle:
Looks like the Hashicorp Vault test is failing and seems to be related to https://github.com/dapr/dapr/issues/5487 CC: @tmacam

PR #2807, opened 2023-04-25
Internal redis component in config store

shivamkm07:
# Description

This PR refactors redis config store implementation to use redis internal component. 

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


configuration/redis/metadata.yaml
```diff
@@ -27,20 +27,25 @@ metadata:
     example: "false"
     default: "false"
     type: bool
-  - name: maxRetries
+  - name: redisMaxRetries
```
shivamkm07, line 30: Note that this PR modifies redis metadata from `maxRetries` to `redisMaxRetries`, `maxRetryBackoff` to `redisMaxRetryInterval` and Adds another metadata `redisMinRetryInterval`. Renaming these metadata would be breaking changes in metadata, but since redis config store is alpha right now and moving to stable this release(1.11), this should be fine.

Also Since we are using internal redis component, most of the generic metadata options are supported like `dialTimeout`, `readTimeout`, poolSize` etc. Do we also need to included them in the metadata description and so in the associated docs? cc @berndverst

configuration/redis/redis.go
```diff
@@ -198,8 +125,14 @@ func (r *ConfigurationStore) Get(ctx context.Context, req *configuration.GetRequ
...
 	if len(keys) == 0 {
-		if keys, err = r.client.Keys(ctx, "*").Result(); err != nil {
+		var res interface{}
+		if res, err = r.client.DoRead(ctx, "KEYS", "*"); err != nil {
 			r.logger.Errorf("failed to all keys, error is %s", err)
+			return nil, err
+		}
+		keyList := res.([]interface{})
+		for _, key := range keyList {
+			keys = append(keys, fmt.Sprint(key))
```
shivamkm07, line 135: Any better ways to handle this?

configuration/redis/metadata.yaml
```diff
@@ -3,7 +3,7 @@ schemaVersion: v1
 type: configuration
 name: redis
 version: v1
-status: alpha
+status: stable
```
tmacam, line 6: Is the configuration API already marked as Stable?

configuration/redis/metadata.yaml
```diff
@@ -20,27 +20,93 @@ metadata:
     description: Password for the Redis host
     example:  "KeFg23!"
     type: string
+  - name: redisUsername
+    required: false
+    description: Username for Redis host. Defaults to empty. Make sure your redis server version is 6 or above, and have created acl rule correctly.
+    example: "default"
+    type: string
```
tmacam, line 27: I think it would be less confusing to state the default and user something as an example than "default".

```suggestion
  - name: redisUsername
    required: false
    description: Username for Redis host. Make sure your Redis server version is 6 or above, and have created ACLrule correctly.
    example: "myUsername"
    default: ""
    type: string
```

configuration/redis/metadata.yaml
```diff
@@ -20,27 +20,93 @@ metadata:
...
+    description: |
+      Minimum backoff for redis commands between each retry. Default is "8ms"; "-1" disables backoff.
+    example: "8ms"
+    type: duration
+  - name: redisMaxRetryInterval
+    required: false
+    description: |
+      Maximum backoff for redis commands between each retry. Default is "2s";"-1" disables backoff.
+    example: "5s"
+    type: duration
```
tmacam, line 59: I think it is better to explicitly add a `default` field to these methods and remove the default from the description. It would add more value to automated tools and for auto-doc generation.

I just did some of these fields on https://github.com/dapr/components-contrib/pull/2796. Might be convenient to just copy and paste from there to keep consistency. Do notice that not every field used by bindings is required for configuration so there might be a few extra fields you might want to remove if you do copy from there.

I would copy some of these fields from there.

configuration/redis/redis.go
```diff
@@ -30,35 +28,38 @@ import (
...
+	infoReplicationDelimiter     = "\r\n"
+	host                         = "redisHost"
+	password                     = "redisPassword"
+	enableTLS                    = "enableTLS"
+	redisMaxRetries              = "redisMaxRetries"
+	redisMaxRetryInterval        = "redisMaxRetryInterval"
+	redisMinRetryInterval        = "redisMinRetryInterval"
+	failover                     = "failover"
+	sentinelMasterName           = "sentinelMasterName"
+	redisDB                      = "redisDB"
```
tmacam, line 47: Does it make sense to keep those constants if are decoding from the `internal/metadata.Settings` struct? Besides, they are only used _once_ and on the test files, if you do need this constant, might be better to defined it there -- or honestly, not define it at all.

configuration/redis/redis.go
```diff
@@ -30,35 +28,38 @@ import (
...
+	defaultBase                  = 10
+	defaultBitSize               = 0
+	defaultDB                    = 0
+	defaultRedisMaxRetries       = 3
+	defaultRedisMaxRetryInterval = time.Second * 2
+	defaultRedisMinRetryInterval = time.Millisecond * 8
+	defaultEnableTLS             = false
```
tmacam, line 54: There are common default among Redis components. Should we define them globally?

configuration/redis/redis.go
berndverst: Agreed - better to move to test only.

configuration/redis/metadata.yaml
tmacam, line 1: Not required but I would move `redisUsername` and `redisPassword` to an `authenticationProfile`

configuration/redis/redis.go
berndverst: You must not set global defaults for all fields because different components used different defaults and changing that would be a breaking change. This must be evaluated on a case by case basis.

However, the defaults should be defined in `internal/component/redis` in the metadata parsing code there, there is a switch statement that allows setting defaults specifically for Configuration store, and that's where these defaults must go.

configuration/redis/metadata.yaml
tmacam: You could, if you wanted, take the a similar path to the one taken by redis statestore and still read `maxRetries` and other legacy fields _but_ as you said, those are legacy. It's better to remove these legacy fields now and push for the new fields (which are consistent among all Redis component now) now.

configuration/redis/metadata.yaml
berndverst: It will be stable in this release.

configuration/redis/metadata.yaml
berndverst: If it's not required then there is no default.

berndverst (approved):

berndverst:
@tmacam I'll merge this for now - but maybe we can have a follow-up PR to apply your metadata suggestions.

shivamkm07:
@berndverst @tmacam I have included the review comments in PR #2825 . Please take a look. Thanks!

Issue #2712, opened 2023-03-27
[AWS DynamoDB] Support for transaction


artursouza:
DynamoDB itself does support transaction but not in Dapr, adding support to it can enable Actors to be used with DynamoDB too.

https://aws.amazon.com/about-aws/whats-new/2019/06/amazon-dynamodb-now-supports-up-to-25-unique-items-and-4-mb-of-data-per-transactional-request

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/transaction-apis.html

PR #2809, opened 2023-04-25
[AWS DynamoDB] Transaction Support

robertojrojas:
# Description

Adds Transaction Support to DynamoDB.

## Issue reference

Please reference the issue this PR will close: #2712

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


state/aws/dynamodb/dynamodb.go
```diff
@@ -393,6 +393,61 @@ func (d *StateStore) parseTTL(req *state.SetRequest) (*int64, error) {
...
+// Multi performs a transactional operation. succeeds only if all operations succeed, and fails if one or more operations fail.
+func (d *StateStore) Multi(ctx context.Context, request *state.TransactionalStateRequest) error {
+	twinput := &dynamodb.TransactWriteItemsInput{
+		TransactItems: []*dynamodb.TransactWriteItem{},
+	}
+
+	// Dedup ops where the last operation with a matching Key takes precedence
+	txs := map[string]int{}
+	for i, o := range request.Operations {
+		txs[o.GetKey()] = i
```
DeepanshuA, line 405: Just trying to think if it is safe to do dedup of ops here and if yes, then why we have been not doing similar in other state stores as well.

state/aws/dynamodb/dynamodb.go
```diff
@@ -393,6 +393,61 @@ func (d *StateStore) parseTTL(req *state.SetRequest) (*int64, error) {
...
+		}
+
+		twi := &dynamodb.TransactWriteItem{}
+		switch req := o.(type) {
+		case state.SetRequest:
+			value, err := d.marshalToString(req.Value)
+			if err != nil {
+				return fmt.Errorf("dynamodb error: failed to marshal value for key %s: %s", req.Key, err)
+			}
+			twi.Put = &dynamodb.Put{
```
DeepanshuA, line 421: Do we also need to take care of Update op?

state/aws/dynamodb/dynamodb.go
```diff
@@ -393,6 +393,61 @@ func (d *StateStore) parseTTL(req *state.SetRequest) (*int64, error) {
...
+		case state.DeleteRequest:
+			twi.Delete = &dynamodb.Delete{
+				TableName: aws.String(d.table),
+				Key: map[string]*dynamodb.AttributeValue{
+					d.partitionKey: {
+						S: aws.String(req.Key),
+					},
+				},
+			}
+		}
```
DeepanshuA, line 442: Do we need a default statement here passing back unsupported operation?

state/aws/dynamodb/dynamodb.go
```diff
@@ -393,6 +393,61 @@ func (d *StateStore) parseTTL(req *state.SetRequest) (*int64, error) {
 	return nil, nil
 }
 
+// Multi performs a transactional operation. succeeds only if all operations succeed, and fails if one or more operations fail.
+func (d *StateStore) Multi(ctx context.Context, request *state.TransactionalStateRequest) error {
+	twinput := &dynamodb.TransactWriteItemsInput{
+		TransactItems: []*dynamodb.TransactWriteItem{},
```
ItalyPaleAle, line 399: Please pre-allocate the array with a capacity of `len(request.Operation)`

ItalyPaleAle (dismissed):

state/aws/dynamodb/dynamodb.go
```diff
@@ -393,6 +393,61 @@ func (d *StateStore) parseTTL(req *state.SetRequest) (*int64, error) {
 	return nil, nil
 }
 
+// Multi performs a transactional operation. succeeds only if all operations succeed, and fails if one or more operations fail.
+func (d *StateStore) Multi(ctx context.Context, request *state.TransactionalStateRequest) error {
+	twinput := &dynamodb.TransactWriteItemsInput{
```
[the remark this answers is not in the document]
robertojrojas: That makes sense. Will do!

state/aws/dynamodb/dynamodb.go
robertojrojas: @DeepanshuA 

Good question!
I added this logic due to a DynamoDB error I was running into while with the conformance tests. Here is the error 
```
ValidationException: Transaction request cannot include multiple operations on one item
```
The error was coming due to duplicate `keys` within a set of operations.
As part of my investigations, I had conversations with folks that know DynamoDB well and came to the conclusion that adding this here would both avoid getting the error and should not cause any bad side effects.

If this turns out to be a general practice, maybe a similar logic can be added at a higher level in the State Multi Transaction execution.

state/aws/dynamodb/dynamodb.go
robertojrojas: When I looked at the conformance tests, I only saw `SetRequest` and `DeleteRequest` being tested and looks like is the same pattern across the State components that support Multiple Tx Ops.

state/aws/dynamodb/dynamodb.go
robertojrojas: @ItalyPaleAle please see my comment https://github.com/dapr/components-contrib/pull/2809#discussion_r1179148188

state/aws/dynamodb/dynamodb.go
robertojrojas: I don't know if it will provide any value currently as `Set` and `Delete` are the only possible operations. If in the future, another operation is added, IMO it should be driven by the conformance tests.

state/aws/dynamodb/dynamodb.go
robertojrojas: @ItalyPaleAle I thought about pre-allocating the slice, but I didn't see a lot of benefits by doing so, plus if it pre-allocates the slice, the logic to make sure the order if preserved could be a bit awkward.

state/aws/dynamodb/dynamodb.go
ItalyPaleAle: I don't believe we have an Update op

state/aws/dynamodb/dynamodb.go
ItalyPaleAle: Ok if this is needed for a DynamoDB-specific behavior, please add a comment explaining that this component is the only one behaving that way, and why this is necessary. In the docs, we should mention this behavior too.

It's good our conf tests include this behavior so we could validate it. I find reassurance in the fact that we don't believe many users actually rely on Multi to perform multiple operations on the same key.

state/aws/dynamodb/dynamodb.go
ItalyPaleAle: The benefits are that when you allocate a slice without initial capacity, it has capacity 0. Then, as you add the first item, the Go runtime allocates a slice large enough for "some" items. If you add more items than the capacity, it then re-allocates the slice. All of this adds work for the allocator and the GC.

You can allocate with an initial capacity rather than length: `make([]*dynamodb.TransactWriteItem, 0, len(request.Operation))`. You can then still use `append` in the rest of the code, but you have already allocated capacity for the worst-case scenario (nothing to dedupe) so you'll never have to re-allocate the slice.

robertojrojas: I see your point. Good one!

state/aws/dynamodb/dynamodb.go
DeepanshuA: Yup, makes sense then. Agree with @ItalyPaleAle for comment and docs.

state/aws/dynamodb/dynamodb.go
DeepanshuA: Ok, got it.

state/aws/dynamodb/dynamodb.go
robertojrojas: Absolutely! I was planning to update the AWS DynamoDB Component docs indicating this behavior.

berndverst (approved):

Issue #2814, opened 2023-04-28
Rabbitmq Pub/Sub: Quorum queues


aaguilartablada:
From RabbitMQ 3.8.0 we have the possibility to create quorum queues instead classic: [https://www.rabbitmq.com/quorum-queues.html](https://www.rabbitmq.com/quorum-queues.html)

It is the recommended type when deploying a RabbitMQ cluster, as it provides fault tolerance. The proposal is allowing Dapr user to tell Subscription to create a quroum queue if desired through metadata.

berndverst:
Duplicate of https://github.com/dapr/components-contrib/issues/2544

berndverst:
Thanks for the request here. Please move the discussion of this feature to #2544. And update your PR to refer to this issue please :)

PR #2816, opened 2023-04-28
[RabbitMQ pub/sub] Allow quorum queues

aaguilartablada:
# Description

Allow Dapr user creating quorum queues through Subscription metadata

## Issue reference

https://github.com/dapr/components-contrib/issues/2814
https://github.com/dapr/components-contrib/issues/2544

Please reference the issue this PR will close: #_[2814]_ #_[2544]_


berndverst (commented):
Please add an integration tests in `tests/certification/pubsub/rabbitmq` it should be a new test case there.

DeepanshuA:
A similar PR was raised earlier with multiple queue options, but that seems to have gone stale: https://github.com/dapr/components-contrib/pull/2556

berndverst:
> A similar PR was raised earlier with multiple queue options, but that seems to have gone stale: #2556

Great point @DeepanshuA.

@aaguilartablada please take a look at #2556 and incorporate the feedback we left on that PR. We will want to see those details here as well.

As you are probably the most knowledgeable about RabbitMQ (among maintainers and approvers) @DeepanshuA I'll defer to you for the review of this PR.

Unfortunately this PR here isn't quite ready and we are now approaching code freeze any day.

aaguilartablada:
Sorry for the delay. I was out for some days. I have just uploaded changes to allow (also) stream queues and apply x-max-length, x-max-length-bytes and x-max-age when defined.

To cover these metadata cases in certification tests I have included x-queue-type and x-max-length arguments in the subscription generated to run TestRabbitMQ.

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -401,6 +406,57 @@ func (r *rabbitMQ) prepareSubscription(channel rabbitMQChannelBroker, req pubsub
...
+		parsedVal, pErr := strconv.ParseUint(val, 10, 0)
+		if pErr != nil {
+			r.logger.Errorf("%s prepareSubscription error: can't parse %s value on subscription metadata for topic/queue `%s/%s`: %s", logMessagePrefix, argMaxLength, req.Topic, queueName, pErr)
+			return nil, pErr
+		}
+		args[argMaxLength] = parsedVal
+	}
+
+	// Applying x-max-age if definde, but only for stream queues
+	if val, ok := req.Metadata[metadataMaxAgeKey]; ok && val != "" {
```
berndverst, line 446: If this is not defined at the request level do you want to use the global default you define in the component metadata? `r.rabbitMQMetadata.MaxAge` -- you made that value configurable but don't use it anywhere.

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -401,6 +406,57 @@ func (r *rabbitMQ) prepareSubscription(channel rabbitMQChannelBroker, req pubsub
 		args[argMaxPriority] = mp
 	}
 
+	// queue type is classic by default, but we allow user to create quorum queues if desired
+	if val, ok := req.Metadata[metadataQueueType]; ok && val != "" {
```
berndverst, line 410: If queue type isn't specified at the request level, do you want to use the global component queue type you made configurable? `r.rabbitMQMetadata.QueueType`

pubsub/rabbitmq/metadata.go
```diff
@@ -45,11 +45,13 @@ type rabbitmqMetadata struct {
 	ReconnectWait        time.Duration          `mapstructure:"reconnectWaitSeconds"`
 	MaxLen               int64                  `mapstructure:"maxLen"`
 	MaxLenBytes          int64                  `mapstructure:"maxLenBytes"`
+	MaxAge               string                 `mapstructure:"maxAge"`
 	ExchangeKind         string                 `mapstructure:"exchangeKind"`
 	PublisherConfirm     bool                   `mapstructure:"publisherConfirm"`
 	SaslExternal         bool                   `mapstructure:"saslExternal"`
 	Concurrency          pubsub.ConcurrencyMode `mapstructure:"concurrency"`
 	DefaultQueueTTL      *time.Duration         `mapstructure:"ttlInSeconds"`
+	QueueType            string                 `mapstructure:"queueType"`
```
berndverst, line 54: Currently unused.

pubsub/rabbitmq/metadata.go
```diff
@@ -45,11 +45,13 @@ type rabbitmqMetadata struct {
 	ReconnectWait        time.Duration          `mapstructure:"reconnectWaitSeconds"`
 	MaxLen               int64                  `mapstructure:"maxLen"`
 	MaxLenBytes          int64                  `mapstructure:"maxLenBytes"`
+	MaxAge               string                 `mapstructure:"maxAge"`
```
berndverst, line 48: Currently unused.

tests/certification/pubsub/rabbitmq/rabbitmq_test.go
berndverst, line 1: More is needed here -- please revert this change here.

If possible I'd like to see something that actually **additionally** shows either quorum or stream queues in use. Do you think that is possible?
If so create an entirely separate new `flow` run block. And if necessary you may need to define a different docker compose file -- depending on whether RabbitMQ needs to be initialized with different settings for different queue types.

if there is a behavior change between classic queues and one of the other queue types which we can assert -- that would be ideal!

berndverst (changes requested):

pubsub/rabbitmq/rabbitmq.go
aaguilartablada: I will change it to indicate queue type on req level instead of component

pubsub/rabbitmq/rabbitmq.go
aaguilartablada: You are right. My fault. As it only applies to stream queues, I'm going to change it to be defined in req metadata

tests/certification/pubsub/rabbitmq/rabbitmq_test.go
aaguilartablada: Ok! I will work in it

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -46,12 +46,17 @@ const (
 	publishRetryWaitSeconds = 2
 
 	argQueueMode          = "x-queue-mode"
-	argMaxLength          = "x-max-length"
+	argMaxLength          = "x-max-length" // only for quorum and classic queues
 	argMaxLengthBytes     = "x-max-length-bytes"
+	argMaxAge             = "x-max-age" // only for stream type queues
```
DeepanshuA, line 51: Should we rename parameter name to reflect that it is for stream specifically? Something like, StreamMaxAge.

Just thinking, if we should use directly amqp constant here as well? 
https://github.com/rabbitmq/amqp091-go/blob/09d223a6803ba6824f3c311c82a3f630c1b162a9/types.go#L255

aaguilartablada:
Hi again @DeepanshuA. 

Finally I discarded stream queues as I've never worked with them and their functionality is radically different from classic and quorum queues. Furthermore, it requires another package (github.com/rabbitmq/rabbitmq-stream-go-client/pkg/amqp) and several code changes to work. Maybe another contributor has experience whit this kind of queues and can create a dedicated PR.

The difference between classic and quorum queues is in the server and how it replicates and manage data in memory and disk. For this reason I didn't create a new flow for quorum queues. Instead, I made a change in an existing flow in which we have two consumers. One consumer creates quorum queues and the other creates classic queues. What do you think?

berndverst:
Unfortunately this will not make it into release 1.11 at this point.

aaguilartablada (commented):
can you review again, please?

aaguilartablada (commented):
@berndverst I don't know why the PR is stuck pending on requested changes that I already resolved.

pubsub/rabbitmq/metadata.go
aaguilartablada: deleted

pubsub/rabbitmq/metadata.go
aaguilartablada: deleted

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -402,6 +410,37 @@ func (r *rabbitMQ) prepareSubscription(channel rabbitMQChannelBroker, req pubsub
 		args[argMaxPriority] = mp
 	}
 
+	// queue type is classic by default, but we allow user to create quorum queues if desired
+	if val, ok := req.Metadata[reqMetadataQueueTypeKey]; ok && val != "" {
+		if !queueTypeValid(val) {
+			return nil, errors.New(fmt.Sprintf("invalid queue type %s. Valid types are %s and %s", val, amqp.QueueTypeClassic, amqp.QueueTypeQuorum))
```
ItalyPaleAle, line 416:
```suggestion
			return nil, fmt.Errorf("invalid queue type %s. Valid types are %s and %s", val, amqp.QueueTypeClassic, amqp.QueueTypeQuorum)
```

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -338,8 +342,12 @@ func (r *rabbitMQ) Subscribe(ctx context.Context, req pubsub.SubscribeRequest, h
 	select {
 	case <-time.After(time.Minute):
 		return fmt.Errorf("failed to subscribe to %s", queueName)
-	case <-ackCh:
-		return nil
+	case channelCrash := <-ackCh:
+		if channelCrash {
```
ItalyPaleAle, line 346:
```suggestion
	case failed := <-ackCh:
		if failed {
```

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -402,6 +410,37 @@ func (r *rabbitMQ) prepareSubscription(channel rabbitMQChannelBroker, req pubsub
 		args[argMaxPriority] = mp
 	}
 
+	// queue type is classic by default, but we allow user to create quorum queues if desired
+	if val, ok := req.Metadata[reqMetadataQueueTypeKey]; ok && val != "" {
```
ItalyPaleAle, line 414:
```suggestion
	if val := req.Metadata[reqMetadataQueueTypeKey]; val != "" {
```

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -402,6 +410,37 @@ func (r *rabbitMQ) prepareSubscription(channel rabbitMQChannelBroker, req pubsub
...
+		if !queueTypeValid(val) {
+			return nil, errors.New(fmt.Sprintf("invalid queue type %s. Valid types are %s and %s", val, amqp.QueueTypeClassic, amqp.QueueTypeQuorum))
+		} else {
+			args[amqp.QueueTypeArg] = val
+		}
+	} else if val == "" {
```
ItalyPaleAle, line 420:
```suggestion
	} else {
```

pubsub/rabbitmq/rabbitmq_test.go
```diff
@@ -98,10 +98,36 @@ func TestPublishAndSubscribeWithPriorityQueue(t *testing.T) {
 	assert.Equal(t, 1, messageCount)
 	assert.Equal(t, "hello world", lastMessage)
 
-	err = pubsubRabbitMQ.Publish(context.Background(), &pubsub.PublishRequest{Topic: topic, Data: []byte("foo bar")})
+	// subscribe using classic queue type
+	err = pubsubRabbitMQ.Subscribe(context.Background(), pubsub.SubscribeRequest{Topic: topic, Metadata: map[string]string{reqMetadataQueueTypeKey: "classic"}}, handler)
+	assert.Nil(t, err)
```
ItalyPaleAle, line 103:
```suggestion
	assert.NoError(t, err)
```

pubsub/rabbitmq/rabbitmq_test.go
```diff
@@ -98,10 +98,36 @@ func TestPublishAndSubscribeWithPriorityQueue(t *testing.T) {
...
+
+	// publish using classic queue type
+	err = pubsubRabbitMQ.Publish(context.Background(), &pubsub.PublishRequest{Topic: topic, Data: []byte("hey there"), Metadata: map[string]string{reqMetadataQueueTypeKey: "classic"}})
 	assert.Nil(t, err)
```
ItalyPaleAle, line 107:
```suggestion
	assert.NoError(t, err)
```

pubsub/rabbitmq/rabbitmq_test.go
```diff
@@ -98,10 +98,36 @@ func TestPublishAndSubscribeWithPriorityQueue(t *testing.T) {
...
 	assert.Equal(t, 2, messageCount)
+	assert.Equal(t, "hey there", lastMessage)
+
+	// subscribe using quorum queue type
+	err = pubsubRabbitMQ.Subscribe(context.Background(), pubsub.SubscribeRequest{Topic: topic, Metadata: map[string]string{reqMetadataQueueTypeKey: "quorum"}}, handler)
+	assert.Nil(t, err)
+
+	// publish using quorum queue type
+	err = pubsubRabbitMQ.Publish(context.Background(), &pubsub.PublishRequest{Topic: topic, Data: []byte("hello friends"), Metadata: map[string]string{reqMetadataQueueTypeKey: "quorum"}})
+	assert.Nil(t, err)
```
ItalyPaleAle, line 118:
```suggestion
	assert.NoError(t, err)
```

pubsub/rabbitmq/rabbitmq_test.go
```diff
@@ -98,10 +98,36 @@ func TestPublishAndSubscribeWithPriorityQueue(t *testing.T) {
...
+	<-processed
+	assert.Equal(t, 3, messageCount)
+	assert.Equal(t, "hello friends", lastMessage)
+
+	// trying to subscribe using invalid queue type
+	err = pubsubRabbitMQ.Subscribe(context.Background(), pubsub.SubscribeRequest{Topic: topic, Metadata: map[string]string{reqMetadataQueueTypeKey: "invalid"}}, handler)
+	assert.NotNil(t, err)
```
ItalyPaleAle, line 125:
```suggestion
	assert.Error(t, err)
```

ItalyPaleAle:
@aaguilartablada will merge when tests pass

Can you please open an issue in dapr/docs to get the docs for this component updated?

ItalyPaleAle (approved):

PR #2817, opened 2023-04-28
Changes wasm middleware and output binding config to url instead of path

### LINKED ISSUE #2700: Support WASM files being fetched from an http or https endpoint
## Describe the feature

Today the WASM Middleware component can fetch warm files from a path: https://github.com/dapr/components-contrib/blob/master/middleware/http/wasm/httpwasm.go#L34

https://docs.dapr.io/reference/components-reference/supported-middleware/middleware-wasm/

But it will be great if we can use an OCI registry to fetch wasm files, so we can leverage package distribution, enabling developers to share their wasm files as container images. 

I was looking at other projects, and something like the `oras-go` project can be used: https://github.com/oras-project/oras-go -> https://oras.land/client_libraries/0_go/

Creating an image with a WASM file should be as simple as creating a Dockefile like this: 
```
FROM scratch
COPY main.wasm /main.wams
```

Ideas? Suggestions? 

## Release Note

RELEASE NOTE:

codefromthecrypt:
# Description

This switches the wasm middleware and output binding config to use url instead of path to read the WebAssembly module. This allows us to later implement HTTP and OCI.

Note: The output binding passes STDIN as well as arguments. This means it can run intepreters such as python. I will be implementing this with the OCI scheme, tested with https://github.com/orgs/vmware-labs/packages/container/package/python-wasm

## Issue reference

related to, but won't close #2700

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
  * https://github.com/dapr/docs/pull/3358 (output binding)
  * https://github.com/dapr/docs/pull/3359 (middleware)


yaron2 (approved):

yaron2:
Thanks, LGTM overall. See lint failure.

codefromthecrypt:
thanks for the look. lint fixed now!

bindings/wasm/output_test.go
```diff
@@ -28,11 +28,15 @@ import (
 )
 
 const (
-	pathArgs    = "./testdata/args/main.wasm"
-	pathExample = "./testdata/example/main.wasm"
-	pathLoop    = "./testdata/loop/main.wasm"
+	urlArgsFile    = "file://testdata/args/main.wasm"
+	urlExampleFile = "file://testdata/example/main.wasm"
+	urlLoopFile    = "file://testdata/loop/main.wasm"
+	urlPythonOCI   = "oci://ghcr.io/vmware-labs/python-wasm:3.11.3"
```
salaboy, line 34: @codefromthecrypt is this an example? is that a python script? or some kind of python runtime?

codefromthecrypt: it is actually the python runtime. So there are basically two main ways to do wasm.

1. compile your source to wasm (c, go, etc.)
2. compile your interpreter to wasm (python, ruby, etc.)

So, in this case, the python script could be taken from stdin or args. For example, you could do an inline script in args, or if someone added a layer with a script, you could indicate that.
```
-c "import sys; import pprint; pprint.pprint(sys.path)"
```

The above are the ARGS used in the README of that image.
https://github.com/vmware-labs/webassembly-language-runtimes/blob/c6cf6e11ffaef1624108dfc9581911ed11f3f42d/python/README.md?plain=1#L72

Interpreters often need standard libraries mounted in some way, and this is where OCI comes in. There are layers for the root filesystem that also includes supporting files. This prevents users from having to deal with this or potentially add the wrong versions of these files (or mess up variables. For example, the OCI image linked has the correct PYTHONPATH/PYTHONHOME hard-coded into python.wasm and the files are in the image layers:

```bash
$  ./car -tvvf ghcr.io/vmware-labs/python-wasm:3.11.3
https://ghcr.io/v2/vmware-labs/python-wasm/manifests/3.11.3 platform=wasi/wasm32 totalLayerSize: 9584425
https://ghcr.io/v2/vmware-labs/python-wasm/blobs/sha256:2940f0fe18fb54307399ab895bdfaa2a45d13bec713bef86d07f205e9b223552 size=5516755
CreatedBy: COPY python/v3.11.3/bin/python.wasm /python.wasm # buildkit
-rw-r--r--	14756882	Apr 26 19:13:12	python.wasm
https://ghcr.io/v2/vmware-labs/python-wasm/blobs/sha256:8b1e220740c34229cc74ece6224070d2ca8f5d797b0a7d42dc62809575b2bf7f size=4067670
CreatedBy: COPY python/v3.11.3/usr/ /usr # buildkit
-rw-r--r--	0	Apr 26 19:13:12	usr/local/lib/python3.11/lib-dynload/.empty
-rw-r--r--	39504	Apr 26 19:13:12	usr/local/lib/python3.11/os.py
-rw-r--r--	4097136	Apr 26 19:13:12	usr/local/lib/python311.zip
```

Long answer, but hope it helps share the knowledge I accidentally have ;)

codefromthecrypt: PS I chose python for OCI because it uses the most features of OCI. A simple case OCI image will have only one layer, the wasm file. Interpreters will need to read the script from somewhere, so only really inlined stuff will be able to use simple case of a wasm file.

codefromthecrypt:
@yaron2 @salaboy I noticed 1.11 cutoff is tomorrow, so made a decision to just normalize the config property to "url" for both wasm middleware and the output binding.

middleware:
* https://github.com/dapr/docs/pull/3359 updates the current docs
* https://github.com/dapr/samples/pull/150 updates the sample

output binding:
* we need to merge this, so that I can update dapr/dapr go.mod and actually wire it up. There's https://github.com/dapr/docs/issues/3358 to document, but that doesn't make sense until it is integrated. I can do this as soon as first thing tomorrow asia time, depending on when this lands.

yaron2:
Is this a breaking change in terms of the path -> URL change?

codefromthecrypt:
> Is this a breaking change in terms of the path -> URL change?

yes, it is unless we want to back port the old one. As it is an alpha component and we broke it before not sure the rules.

codefromthecrypt:
ps I was wrong, it looks like the wasm binding was already integrated in https://github.com/dapr/dapr/pull/6269 so the only thing left is docs and just updating components-contrib version in dapr/dapr after merge

yaron2:
> > Is this a breaking change in terms of the path -> URL change?
> 
> yes, it is unless we want to back port the old one. As it is an alpha component and we broke it before not sure the rules.

We can technically break the spec as this is an Alpha component, but where possible we try to provide backward compatibility. For this early component I don't think it's required.

Have you tested the middleware locally with Dapr after these changes?

bindings/wasm/output_test.go
codefromthecrypt: hopefully clarified in https://github.com/dapr/docs/pull/3362

codefromthecrypt:
@yaron2 thx for understanding. I hadn't noticed a lot of use of middleware, so gut feel was to change the url before it becomes permanent confusion. I'll try middleware with a local build now and report back.

p.s. https://github.com/dapr/docs/pull/3362 for the docs on components.

codefromthecrypt:
@yaron2 tested and captured output [here](https://github.com/dapr/samples/pull/150#issuecomment-1529316534)

codefromthecrypt:
I added a commit to simplify the wasm output binding invocation, as well tighten the benchmark. We no longer return exit code zero on success (since 1.0)

berndverst:
@codefromthecrypt keep in mind the following deadlines in the future:

1. Feature Freeze -- any significant features should have complete PRs opened before this date (does not necessarily apply to this PR).
2. Code Freeze -- PRs need to have been fully reviewed and merged by this date. This is not the date for opening new PRs. It's best to allow 1-2 weeks before code freeze to give maintainers time to provide feedback and time for you to respond to feedback etc.

Of course we'll try to get this PR reviewed and merged -- but do keep these dates in mind in the future to make things easier for us maintainers. Thanks for your understanding.

codefromthecrypt:
@berndverst absolutely will bear in mind dates.. in fact I would have prioritized differently if I knew them. Where generally do I find these as I found out by accident of asking nervously.

yaron2 (approved):

berndverst:
> @berndverst absolutely will bear in mind dates.. in fact I would have prioritized differently if I knew them. Where generally do I find these as I found out by accident of asking nervously.

We always create an issue in `dapr/dapr` that we also pin in the issues list there called "[version] Release Planning", for example "v1.11 Release Planning" https://github.com/dapr/dapr/issues/5966

That's where we keep the timeline up to date and also discuss the major features we are including in that release.

Soon after we ship v1.11 we will create another such issue and pin it - so you can look for the "v1.12 Release Planning" issue then. The issue will be kept up to date with timelines.

PR #2822, opened 2023-05-01
Tls renegotiation

mdarii:
# Description

TLS Renegotiation is disabled by default in golang. (https://github.com/golang/go/issues/5742)
Some servers force TLS renegotiation and we need to enable it:
`Renegotiation: tls.RenegotiateFreelyAsClient`

## Checklist

* [ x] Code compiles correctly


bindings/http/http.go
```diff
@@ -91,6 +92,12 @@ func (h *HTTPSource) Init(_ context.Context, meta bindings.Metadata) error {
 			return err
 		}
 	}
+	if h.metadata.MTLSRenegotiation != "" {
```
berndverst, line 95:
```suggestion
	if h.metadata.MTLSRenegotiation {
```

bindings/http/metadata.yaml
```diff
@@ -63,6 +63,12 @@ metadata:
     example: "client.key"
     binding:
       output: true
+  - name: MTLSRenegotiation
+    required: false
+    description: "Set TLS renegotiation setting"
+    example: "RenegotiateOnceAsClient"
```
berndverst, line 69:
```suggestion
    allowedValues:
      - "RenegotiateNever"
      - "RenegotiateOnceAsClient"
      - "RenegotiateFreelyAsClient"
    example: "RenegotiateOnceAsClient"
```

bindings/http/http.go
```diff
@@ -64,6 +64,7 @@ type httpMetadata struct {
 	MTLSClientCert      string         `mapstructure:"mtlsClientCert"`
 	MTLSClientKey       string         `mapstructure:"mtlsClientKey"`
 	MTLSRootCA          string         `mapstructure:"mtlsRootCA"`
+	MTLSRenegotiation   string         `mapstructure:"mtlsRenegotiation"`
```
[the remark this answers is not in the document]
mdarii: @berndverst  to be honest I think we are interested more into this one: `RenegotiateOnceAsClient` - it should be enough for most use cases, but to be on safe side I think is better to have support to change any one from option provided by lib crypto.

berndverst:
Thanks for this contribution. It definitely makes sense to add.

Technically this feature is already well past our feature freeze deadline for the 1.11 release (see the pinned release planning issue in dapr/dapr repo). However, as this PR is small and I do not see any risk with this implementation I'll make an exception to include it - I certainly see the value.

Please open an issue for new features / capabilities in the future and wait for a maintainer or approver to give some feedback. Otherwise you may work on a PR that we will overlook or not merge - and that would be unfortunate.

berndverst:
@mdarii I opened this docs issue for you -- to accept this feature please also complete the documentation change: https://github.com/dapr/docs/issues/3368

On that docs issue, please comment with `/assign` so this issue is correctly assigned to you. Thanks!

PR #2831, opened 2023-05-06
Add Kafka Binding SASL Password cert test

berndverst:
# Description

Adds a certification test to the Kafka Binding covering the SASL Password Authentication (with TLS disabled)

This particular scenario broke when using this mechanism with Azure EventHubs Kafka compatibility mode. Downgrading sarama to 1.37.2 was the fix to this issue.


berndverst:
Failing pubsub test can be ignore - that's not touched in this PR and is just flakiness. Binding test passes :)

ItalyPaleAle (approved):

PR #2834, opened 2023-05-11
Remove native/transactional BulkSet/BulkDelete from state stores

ItalyPaleAle:
This PR is a follow-up from #2747 that removes from state stores the native or transactional BulkSet/BulkDelete.

- #2747 implemented BulkSet/BulkDelete so state stores that were transactional would use the Multi() method (i.e. transactions) for BulkSet and BulkDelete operations. This has been undone.
- A small number of components implemented "native" BulkSet and BulkDelete; all have been removed except RethinkDB (for which it was implemented correctly). In all but one case those were just wrappers around transactional operations. In one case, it was implemented in an actual, native way, but it did not support etags and returned an error if the method was invoked with an etag.

Instead, the default bulk store now implements BulkSet and BulkDelete in the same way for all components so that it performs the operations in parallel.

This change is needed to fix feedback in the PR in the runtime: https://github.com/dapr/dapr/pull/6186#pullrequestreview-1390814937

Additionally, this will be needed to fix the issue where etag errors in BulkSet/BulkDelete operations were not correctly sent to clients. See: https://github.com/dapr/dapr/pull/6186#issuecomment-1539262923


state/bulk.go
```diff
@@ -47,6 +57,31 @@ func NewDefaultBulkStore(base BaseStore) BulkStore {
...
+	return b.BulkDeleteWithOptions(ctx, req, BulkStoreOpts{})
+}
+
+// BulkDeleteWithOptions performs a bulk delete operation with options
+func (b *DefaultBulkStore) BulkDeleteWithOptions(ctx context.Context, req []DeleteRequest, opts BulkStoreOpts) error {
+	return DoBulkSetDelete(ctx, req, b.base.Delete, opts)
+}
+
+// DoBulkGet performs BulkGet.
+func DoBulkGet(ctx context.Context, req []GetRequest, opts BulkGetOpts, getFn func(ctx context.Context, req *GetRequest) (*GetResponse, error)) ([]BulkGetResponse, error) {
```
ItalyPaleAle, line 84: This method is exported so components that want to implement some Bulk* methods natively can pick-and-choose. Example is RethinkDB

bindings/zeebe/command/deploy_process.go
```diff
@@ -38,10 +38,9 @@ func (z *ZeebeCommand) deployProcess(ctx context.Context, req *bindings.InvokeRe
 		return nil, ErrMissingFileName
 	}
 
-	// TODO: Replace with a method that is not deprecated.
-	response, err := z.client.NewDeployProcessCommand(). //nolint:staticcheck
-								AddResource(req.Data, deployFileName).
-								Send(ctx)
+	response, err := z.client.NewDeployResourceCommand().
+		AddResource(req.Data, deployFileName).
```
ItalyPaleAle, line 42: The linter wasn't happy with `//nolint:staticcheck` and wasn't happy without it either. Rather than fighting the linter (lesigh) I just replaced this with the non-deprecated method.

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -414,16 +418,15 @@ func (c *StateStore) Set(ctx context.Context, req *state.SetRequest) error {
 	pk := azcosmos.NewPartitionKeyString(partitionKey)
 	_, err = c.client.UpsertItem(upsertCtx, pk, marsh, &options)
 	if err != nil {
+		resErr := &azcore.ResponseError{}
```
ItalyPaleAle, line 421: Looks like Cosmos DB was not reporting Etag failures as separate errors in Set operations. In a follow-up PR I'll update our conformance tests to make sure that all components report the correct errors.

ItalyPaleAle: The same changes was made for other components too.

Follow-up PR is #2843

state/bulk.go
```diff
@@ -16,21 +16,29 @@ package state
...
 
+// BulkStoreOpts contains options for the BulkSet and BulkDelete methods.
+type BulkStoreOpts struct {
+	// Number of requests made in parallel while storing/deleting values in bulk.
+	// When set to <= 0 (the default value), will perform all operations in parallel without limit.
+	Parallelism int
+}
+
 // BulkStore is an interface to perform bulk operations on store.
 type BulkStore interface {
```
artursouza, line 38: Does this break pluggable components?

state/oracledatabase/oracledatabaseaccess.go
```diff
@@ -194,16 +194,16 @@ func (o *oracleDatabaseAccess) doSet(ctx context.Context, db querier, req *state
...
-		}
 		return err
 	}
 	rows, err := result.RowsAffected()
 	if err != nil {
 		return err
 	}
 	if rows != 1 {
+		if req.ETag != nil && *req.ETag != "" {
+			return state.NewETagError(state.ETagMismatch, err)
```
artursouza, line 205: Wait. Was this another eTag issue detected?

state/sqlite/sqlite_dbaccess.go
```diff
@@ -297,42 +297,41 @@ func (a *sqliteDBAccess) BulkGet(parentCtx context.Context, req []state.GetReque
...
-	var (
-		n    int
-		etag string
-	)
+	var n int
 	res := make([]state.BulkGetResponse, len(req))
 	for ; rows.Next(); n++ {
 		r := state.BulkGetResponse{}
-		r.Key, r.Data, etag, err = readRow(rows)
+		r.Key, r.Data, r.ETag, err = readRow(rows)
```
artursouza, line 304: eTag issues here too?

state/sqlserver/migration.go
```diff
@@ -248,36 +244,12 @@ func (m *migration) ensureTypeExists(ctx context.Context, db *sql.DB, mr migrati
 	return runCommand(ctx, db, tsql)
 }
 
-/* #nosec. */
-func (m *migration) ensureBulkDeleteStoredProcedureExists(ctx context.Context, db *sql.DB, mr migrationResult) error {
```
artursouza, line 252: Why this is not needed?

artursouza (commented):
Overall LGTM. Some minor comments.

state/bulk.go
ItalyPaleAle: Actually.... It shouldn't. The [protos](https://github.com/dapr/dapr/pull/6342/files#diff-2179950bcc2e9039862417bd7c8e169b893f5a54dbdc2c62cf04f225502b6584) have changed but we've only added new properties, which should be backwards-compatible.

Btw my understanding was that pluggable components are not stable and we are allowed to break them if we need to change the APIs in contrib. Can you confirm that's the case?

state/sqlite/sqlite_dbaccess.go
ItalyPaleAle: This is another example of an etag error that's been fixed (and #2834 fixes more).

This was actually a much worse error because it caused BulkGet to return the wrong etag, as it was getting the address of a variable outside of the loop

state/oracledatabase/oracledatabaseaccess.go
ItalyPaleAle: Between this and #2834 I have fixed plenty of etag errors.

In most cases the errors were simply components not reporting etag failures as state.ETagError errors... In other cases it was much worse: https://github.com/dapr/components-contrib/pull/2834#discussion_r1197032949

state/sqlserver/migration.go
ItalyPaleAle: Removed native BulkDelete from this component. Otherwise that was behaving atomically, which is what we wanted to get rid of.

state/bulk.go
artursouza: We are allowed to break but it does not mean we can leave it broken. We can validate that in the RC.

artursouza (approved):

PR #2843, opened 2023-05-17
Add conformance tests for etags in Bulk operations

### LINKED ISSUE #2841: Update state store conformance tests to validate etag failures return the proper error
While working on #2834, I noticed that Cosmos DB did not report etag failures with the correct error, but rather the generic error. I would not be surprised if this weren't an isolated instance.

We should update our conformance tests to ensure etag failures are reported correctly.

ItalyPaleAle:
Fixes #2841 by expanding the conformance tests:

1. Adds conformance tests for etags in BulkSet and BulkDelete operations
2. Expands tests for etags and Bulk* operations
3. Changes state store conformance tests to require all components to have Get, Set, Delete, BulkGet, BulkSet, BulkDelete. Before components had to opt into those conformance tests, but since all components MUST implement those methods, it made no sense to do that and it added risks that some components were missing tests (it was happening)
4. Standardize how BulkGet works across components. Before, most state stores returned missing keys in the BulkGet response with empty values, but some omitted the keys that didn't exist. This has been standardized to always return all requested keys, with those that don't exist having empty values, so it's consistent across components and with the Get() method. Conformance tests have been updated to enforce this behavior.

In the process, a bunch of bugs have been fixed to ensure components pass the conformance tests.


ItalyPaleAle:
Ready for review now

state/cockroachdb/cockroachdb.go
```diff
@@ -28,12 +28,17 @@ func New(logger logger.Logger) state.Store {
...
 		SetQueryFn: func(req *state.SetRequest, opts postgresql.SetQueryOptions) string {
-			// String concat is required for table name because sql.DB does not
-			// substitute parameters for table names.
-			// Other parameters use sql.DB parameter substitution.
-			if req.ETag == nil || *req.ETag == "" {
+			// Sprintf is required for table name because the driver does not substitute parameters for table names.
+			if !req.HasETag() {
+				// We do an upsert in both cases, even when concurrency is first-write, because the row may exist but be expired (and not yet garbage collected)
+				// The difference is that with concurrency as first-write, we'll update the row only if it's expired
+				var whereClause string
```
ItalyPaleAle, line 35: This change was needed to ensure first-write wins works with this component

state/redis/redis.go
```diff
@@ -162,7 +160,11 @@ func (r *StateStore) Init(ctx context.Context, metadata state.Metadata) error {
 
 // Features returns the features available in this state store.
 func (r *StateStore) Features() []state.Feature {
-	return r.features
+	if r.clientHasJSON {
```
ItalyPaleAle, line 163: By doing this, we make the component report that query API is supported only if it's actually supported - if it's connected to a Redis that has RedisJSON

This was needed for the conformance tests too

tests/config/state/tests.yml
```diff
@@ -1,83 +1,106 @@
-# Supported operations: set, get, delete, bulkget, bulkset, bulkdelete, transaction, etag, first-write, query, ttl
+# Supported operations: transaction, etag, first-write, query, ttl
```
ItalyPaleAle, line 1: set, get, delete, bulkget, bulkset, bulkdelete have been removed because they're always enabled.

set, get, and delete are required in every state store (otherwise, it wouldn't be a valid state store).
bulk* are included automatically for all components, even when they don't implement them natively.

Removing these operations allowed for simplifying the YAML below. Importantly, some operations were missing even though they should have been present (see how memcached had bulkset and bulkdelete but not bulkget).

tests/config/state/tests.yml
```diff
@@ -1,83 +1,106 @@
...
-    operations: [ "set", "get", "delete", "bulkget", "bulkset", "bulkdelete", "transaction", "etag",  "first-write", "ttl" ]
+    operations: [ "transaction", "etag",  "first-write", "ttl" ]
   - component: aws.dynamodb.docker
     allOperations: false
-    operations: [ "set", "get", "delete", "etag", "bulkset", "bulkdelete", "transaction", "first-write" ]
+    operations: [ "transaction", "etag", "first-write" ]
   - component: aws.dynamodb.terraform
     allOperations: false
-    operations: [ "set", "get", "delete", "etag", "bulkset", "bulkdelete", "transaction", "first-write" ]
+    operations: [ "transaction", "etag", "first-write" ]
```
artursouza, line 97: Does it mean that only the non-required operations are now listed here? This is different from the other conformance tests. Is it a decision to change all conf tests to work this way? If so, this schema does not work well because it should be renamed to "optionalOperations" and allOperations should be removed.

tests/config/state/tests.yml
artursouza: I agree but this schema needs to be updated. We should rename "operations" to "optionalOperations"

artursouza: After looking into the README file we have for conf tests, I think we should either refactor all conf tests to have this new behavior or keep it as is. Having it half-way is not a good contributor experience. Also, it is not clear if there will be anyone volunteering to completing this refactoring for all components.

Long story short, I am trying to minimize half-finished work in our master branch - at least when there is no clear commitment of bringing it to the finish line any time soon.

artursouza: Let's do this then. I will merge this and for 1.12 we change the meaning of operations to only be optional operations and required operations are always included in the test suite.

artursouza (approved):

Issue #2857, opened 2023-05-24
Return Expire Time of TTL values


JoshVanL:
For **transaction** enabled DBs, return the expiration time of keys which have a TTL value set. This should be `expirationTime`, in the `GetResponse` `Metadata` field.

This value should be computed by the database to account for clock skew.

This field is required so that actor SDKs are able to recover state from a cold cache. Without this, actor state SDKs would always have to fetch from the DB when restoring keys that have a TTL, where that cache was not responsible for writing the data.

- [x] postgresql: https://github.com/dapr/components-contrib/pull/2864
- [x] cassandra: https://github.com/dapr/components-contrib/pull/2889
- [x] cockroachdb: https://github.com/dapr/components-contrib/pull/2864
- [x] in-memory: https://github.com/dapr/components-contrib/pull/2870
- [ ] ~cloudflare: *NOT SUPPORTED*~
- [x] sqlite: https://github.com/dapr/components-contrib/pull/2869
- [x] mysql: https://github.com/dapr/components-contrib/pull/2871
- [ ] redis: https://github.com/dapr/components-contrib/pull/2873
- [x] oracledatabase: https://github.com/dapr/components-contrib/pull/2897
- [x] mongodb: https://github.com/dapr/components-contrib/pull/2888
- [ ] ~memcached: *NOT SUPPORTED*~
- [x] azure: https://github.com/dapr/components-contrib/pull/2898
- [x] etcd: https://github.com/dapr/components-contrib/pull/2891
- [x] aws: https://github.com/dapr/components-contrib/pull/2899
- [x] sqlserver: https://github.com/dapr/components-contrib/pull/2896

JoshVanL:
All transactional DBs are supported in contrib so this can be closed.

PR #2863, opened 2023-05-26
Conformance Tests: State- ` ttlExpireTime`

JoshVanL:
Add state store conformance tests to ensure that the `ttlExpireTime` metadata key is returned when Getting keys which have a `ttlInSeconds` set.

`ttlExpireTime` is needed for state clients (like actors) to have a copy of the expiry time of TTL keys expire so that they can update their local cache when appropriate.

PR shouldn't be merged until all `TTL` enabled state stores implement this metadata response.

See https://github.com/dapr/dapr/pull/6400 https://github.com/dapr/components-contrib/issues/2857 for more info.

### DIFF
--- a/tests/conformance/state/state.go
+++ b/tests/conformance/state/state.go
@@ -1030,7 +1030,6 @@ func ConformanceTests(t *testing.T, props map[string]string, statestore state.St
 				return res.Data == nil
 			}, time.Second*3, 200*time.Millisecond, "expected object to have been deleted in time")
 		})
-
 	} else {
 		t.Run("ttl feature not present", func(t *testing.T) {
 			// We skip this check for Cloudflare Workers KV
--- a/tests/conformance/state/state.go
+++ b/tests/conformance/state/state.go
@@ -1020,6 +1020,7 @@ func ConformanceTests(t *testing.T, props map[string]string, statestore state.St
 				return res.Data == nil
 			}, time.Second*3, 200*time.Millisecond, "expected object to have been deleted in time")
 		})
+
 	} else {
 		t.Run("ttl feature not present", func(t *testing.T) {
 			// We skip this check for Cloudflare Workers KV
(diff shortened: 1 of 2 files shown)


tests/conformance/state/state.go
```diff
@@ -1016,6 +1016,46 @@ func ConformanceTests(t *testing.T, props map[string]string, statestore state.St
...
+				Key: key + "-ttl-expire-time",
+			})
+			require.NoError(t, err)
+			assertEquals(t, "⏱️", res)
+
+			require.Containsf(t, res.Metadata, "ttlExpireTime", "expected metadata to contain ttlExpireTime")
+			expireTime, err := time.Parse(time.RFC3339, res.Metadata["ttlExpireTime"])
+			require.NoError(t, err)
+			// Check the expire time is returned, and is in a 10 second window. This window should be _more_ than enough.
+			assert.InDelta(t, now.Add(time.Minute-(5*time.Second)), expireTime.Add(time.Second*5), float64(time.Second*10))
```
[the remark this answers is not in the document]
JoshVanL: Fair enough- I'll update the ttl to an hour.

tests/conformance/state/state.go
```diff
@@ -1016,6 +1026,174 @@ func ConformanceTests(t *testing.T, props map[string]string, statestore state.St
...
+			res, err := statestore.Get(context.Background(), &state.GetRequest{Key: key + "-no-ttl"})
+			require.NoError(t, err)
+			assertEquals(t, "⏱️", res)
+
+			assert.NotContains(t, res.Metadata, "ttlExpireTime")
+		})
+
+		t.Run("ttlExpireTime", func(t *testing.T) {
+			if !config.HasOperation("transaction") {
+				t.Skip("state store does not support transactions")
```
ItalyPaleAle, line 1048:
```suggestion
				// This test is only for state stores that support transactions
				return
```

ItalyPaleAle: Let's not pollute logs for things that are never going to change, it can cause confusion

tests/conformance/state/state.go
```diff
@@ -1016,6 +1026,174 @@ func ConformanceTests(t *testing.T, props map[string]string, statestore state.St
...
+			}
+
+			if config.ComponentName == "redis.v6" || config.ComponentName == "redis.v7" {
+				t.Skip("redis does not support ttlExpireTime")
```
ItalyPaleAle, line 1052:
```suggestion
				// Redis does not support ttlExpireTime
				return
```

ItalyPaleAle (commented):
Is this ready for being merged?

ItalyPaleAle (approved):

PR #2865, opened 2023-05-26
build: rm branch to clean up build

sicoyle:
# Description

I was searching for something in one of my other dapr repos in my VSCode workspace when I came across these lines that look like they should be cleaned up. Not sure if this was an addition prepping for 1.11 release, or if this should be removed now. Thoughts? @ItalyPaleAle @berndverst 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/.github/workflows/certification.yml
+++ b/.github/workflows/certification.yml
@@ -24,6 +24,7 @@ on:
       - 'release-*'
   pull_request:
     branches:
+      - 'master'
       - 'release-*'
 
 jobs:
--- a/.github/workflows/certification.yml
+++ b/.github/workflows/certification.yml
@@ -24,8 +24,6 @@ on:
       - 'release-*'
   pull_request:
     branches:
-      # TODO: REMOVE "master" BEFORE MERGING
-      - 'master'
       - 'release-*'
 
 jobs:


berndverst:
I am not sure what the comment was for but I'm inclined to leave everything in place, maybe removing the comment.

That is because we always merge the release branch back into master - and I don't want us to accidentally mess up the workflow in the master branch.

No harm in having both branches listed here. A PR against master could never trigger this workflow in `release-1.11` branch anyway, and a push to `release-1.11` could not trigger a run in master.

.github/workflows/certification.yml
```diff
@@ -24,8 +24,6 @@ on:
       - 'release-*'
   pull_request:
     branches:
-      # TODO: REMOVE "master" BEFORE MERGING
-      - 'master'
```
berndverst, line 28: Let's only remove the comment please. We don't know why the comment was there.

There is no harm in having master here -- and it's just too painful to merge release back into master everytime we have a release / hotfix etc otherwise

berndverst (approved):

berndverst:
I merged bypassing all tests -- because I know there is no need to run the workflows :)

sicoyle:
> No harm in having both branches listed here. A PR against master could never trigger this workflow in release-1.11 branch anyway, and a push to release-1.11 could not trigger a run in master.

Makes sense, thank you!

PR #2866, opened 2023-05-26
[Metadata] Clarify that MySQL components work with MariaDB too

ItalyPaleAle:
Follows a conversation on Discord with a user that wasn't clear that the MySQL components can be used with MariaDB too.

Also updated in docs: dapr/docs#3463

### DIFF
--- a/state/mysql/metadata.yaml
+++ b/state/mysql/metadata.yaml
@@ -4,7 +4,7 @@ type: state
 name: mysql
 version: v1
 status: stable
-title: "MySQL"
+title: "MySQL & MariaDB"
 urls:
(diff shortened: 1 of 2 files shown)


state/mysql/metadata.yaml
```diff
@@ -35,6 +35,7 @@ metadata:
     description: "Name of the table Dapr uses to store a few metadata properties"
     type: string
     default: "dapr_metadata"
+    example: '"dapr_metadata"'
```
ItalyPaleAle, line 38: The language server was complaining this was missing

berndverst (approved):

PR #2869, opened 2023-05-30
Draft: sqlite return ttlExpiryTime in GetResponse

JoshVanL:

### DIFF
--- a/state/sqlite/sqlite_dbaccess.go
+++ b/state/sqlite/sqlite_dbaccess.go
@@ -250,14 +250,14 @@ func (a *sqliteDBAccess) Get(parentCtx context.Context, req *state.GetRequest) (
 	}
 
 	// Concatenation is required for table name because sql.DB does not substitute parameters for table names
-	stmt := `SELECT key, value, is_binary, etag FROM ` + a.metadata.TableName + `
+	stmt := `SELECT key, value, is_binary, etag, expiration_time FROM ` + a.metadata.TableName + `
 		WHERE
 			key = ?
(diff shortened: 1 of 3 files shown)


state/sqlite/sqlite_dbaccess.go
```diff
@@ -250,25 +250,32 @@ func (a *sqliteDBAccess) Get(parentCtx context.Context, req *state.GetRequest) (
...
 		if errors.Is(err, sql.ErrNoRows) {
 			return &state.GetResponse{}, nil
 		}
 		return nil, err
 	}
 
+	var metadata map[string]string
+	if expireTime != nil {
+		metadata = map[string]string{
+			state.GetRespMetaKeyTTLExpireTime: strconv.FormatInt(expireTime.UnixMilli(), 10),
```
ItalyPaleAle, line 271: I would prefer this be a ISO-8601 formatted date. UNIX timestamp milliseconds are less clear to handle. They also do not fit on a int32 which causes a number of problems with parsers.

ItalyPaleAle (changes requested):

state/sqlite/sqlite_dbaccess.go
JoshVanL: The problem with text based formats like ISO-8601 is that it is a bit slower than pure number based. We could use int32 unix time, but this would put a shelf life on Dapr to `2038`.

ItalyPaleAle: it is a bit slower but i'm more concerned with correctness here. Also consistency: throughout Dapr we normally reference timestamps as ISO-8601 strings

PR #2871, opened 2023-05-30
mysql: return ttlExpiryTime in GetResponse

JoshVanL:

### DIFF
--- a/state/mysql/mysql.go
+++ b/state/mysql/mysql.go
@@ -546,7 +546,7 @@ func (m *MySQL) Get(parentCtx context.Context, req *state.GetRequest) (*state.Ge
 	var metadata map[string]string
 	if expireTime != nil {
 		metadata = map[string]string{
-			state.GetRespMetaKeyTTLExpireTime: strconv.FormatInt(expireTime.UnixMilli(), 10),
+			state.GetRespMetaKeyTTLExpireTime: expireTime.UTC().Format(time.RFC3339),
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


state/mysql/mysql.go
```diff
@@ -759,14 +773,24 @@ func (m *MySQL) BulkGet(parentCtx context.Context, req []state.GetRequest, _ sta
 	return res[:n], nil
 }
 
-func readRow(row interface{ Scan(dest ...any) error }) (key string, value []byte, etagP *string, err error) {
+func readRow(row interface{ Scan(dest ...any) error }) (key string, value []byte, etagP *string, expireTime *time.Time, err error) {
 	var (
 		etag     string
 		isBinary bool
+		expire   sql.NullString
```
ItalyPaleAle, line 780: Thoughts on making this just a string? Should be more efficient. If the value is NULL in the database, I presume this would be an empty string, and then we can skip the parsing.

JoshVanL: Unfortunately, this is not supported:

```
        	            	error getting state: rpc error: code = Internal desc = fail to get stable-certification-key1 from state store mysql: sql: Scan error on column index 4, name "expiredate": converting NULL to string is unsupported
```

ItalyPaleAle: Use `SELECT IFNULL(expiredate, "")`

ItalyPaleAle (approved):

PR #2883, opened 2023-06-03
feat(nameresolution/kubernetes): add ability for templating the dns name resolution

### LINKED ISSUE #2867: Templating for the DNS name resolution component
https://github.com/helpwave/services/issues/164

## Describe the feature
The schema of the [DNS name resolution component](https://github.com/dapr/components-contrib/tree/master/nameresolution/dns) should be configurable. The configuration key is optional and could be called `schema_template` with a possible value of `#app_id#.internal:#port#`. In that case, `#app_id#` and `#port#` are replaced during [`ResolveID()`](https://github.com/dapr/components-contrib/blob/master/nameresolution/dns/dns.go#L39). When the configuration is not set, we should fall back to the current schema of `#app_id#-dapr.#namespace#.svc:#port#` to be backward compatible.

If discussed, I would implement this change directly.

Another solution would be to develop a new name resolution component just for templating.

## Release Note
RELEASE NOTE: **ADD** templating to the DNS name resolution component

MaxSchaefer:
# Description

---

## Behaviour

When the configuration field `template` is set, on `.Init()` the kubernetes name resolution component will parse the provided template string via [`text/template`](https://pkg.go.dev/text/template) and will cache these for further executions when an address resolution gets requested via `.ResolveID()`.
The template gets populate on a resolve request with the passed [`ResolveRequest` struct](https://github.com/dapr/components-contrib/blob/1ac92a886a7c8534a9e07c5415c84f41297aa945/nameresolution/requests.go#L20).
The [`Data map[string]string`](https://github.com/dapr/components-contrib/blob/1ac92a886a7c8534a9e07c5415c84f41297aa945/nameresolution/requests.go#L24) part of the struct can also be used inside the template.
A successful resolve request will return a populated address string of the template to the caller.
When `template` is set, `clusterDomain` field gets ignored.

### Examples

| Template                                                     | [`ResolveRequest` struct](https://github.com/dapr/components-contrib/blob/1ac92a886a7c8534a9e07c5415c84f41297aa945/nameresolution/requests.go#L20) | Populated address           |
|--------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------|
| `{{.ID}}-dapr.{{.Namespace}}.svc:{{.Port}}`                  | ```{ ID: "my-id", Namespace: "abc", Port: 1234, Data: {} }```                                                                                      | my-id-dapr.abc.svc:1234     |
| `{{.ID}}-{{.Namespace}}.{{.Data.Region}}.internal:{{.Port}}` | ```{ ID: "my-id", Namespace: "abc", Port: 1234, Data: { Region: "land" } }```                                                                      | my-id-ns.land.internal:1234 |

## Spec configuration fields

| Field                | Required |                                                                                                                Type | Details                                                                                                                                                 | Examples |
|----------------------|:--------:|------------------------------------------------------------------------------------------:|:--------------------------------------------------------------------------------------------------------------------------------------------------------|----------|
| template             | N        | `string` | The template that gets parsed and executed by [`text/template`](https://pkg.go.dev/text/template)                                                       | `{{.ID}}-{{.Namespace}}.internal:{{.Port}}`

---

## Issue reference

Closes #2867

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


nameresolution/kubernetes/kubernetes.go
```diff
@@ -45,21 +58,34 @@ func (k *resolver) Init(metadata nameresolution.Metadata) error {
 	if err != nil {
 		return err
 	}
-	if config, ok := configInterface.(map[string]interface{}); ok {
-		clusterDomainPtr := config[ClusterDomainKey]
+	if cfg, ok := configInterface.(map[string]interface{}); ok {
+		clusterDomainPtr := cfg[ClusterDomainKey]
```
ItalyPaleAle, line 62: I know the name "ptr" was there before, but is it possible to rename it to something else? Since technically it's not a pointer, just an interface. Maybe "clusterDomainAny" or anything else? Same for "tmplStr"

nameresolution/kubernetes/kubernetes.go
```diff
@@ -14,28 +14,41 @@ limitations under the License.
 package kubernetes
 
 import (
+	"bytes"
+	"github.com/dapr/kit/config"
```
ItalyPaleAle, line 18: This import needs to be moved with the other ones github.com/dapr/

ItalyPaleAle (changes requested):
Thank you! Please just fix the linter errors. You can run the linter locally too with “make lint”

ItalyPaleAle (approved):

ItalyPaleAle:
Thanks! LGTM.

since it’s a more “foundational” thing, I’d like to get a thumbs up from @berndverst too before merging.

PR #2884, opened 2023-06-05
wasm: implements strictSandbox for output binding and HTTP middleware

codefromthecrypt:
# Description

This implements strictSandbox as an optional configuration for our two wasm features: output binding and HTTP middleware.

When enabled, this uses fake sources of system data, deterministic with the exception of walltime.

## Issue reference

https://github.com/dapr/dapr/issues/6422

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


codefromthecrypt:
ok @Taction @ItalyPaleAle if you like this approach I will backfill tests on the fake'ish clock.

internal/wasm/wasm.go
```diff
@@ -24,6 +29,19 @@ type InitMetadata struct {
...
+	//
+	// # Affected configuration
+	//
+	//   - sys.Walltime increments with a constant value when read, initially a
+	//     second resolution of the current system time.
+	//   - sys.Nanotime increments with a constant value when read, initially
+	//     zero.
+	//   - sys.Nanosleep returns immediately.
+	//   - Random number generators are seeded with a deterministic source.
+	StrictSandbox bool `mapstructure:"strictSandbox"`
```
ItalyPaleAle, line 43: Do you think this should be on by default, or off by default?

bindings/wasm/output.go
```diff
@@ -101,11 +90,16 @@ func (out *outputBinding) Init(ctx context.Context, metadata bindings.Metadata)
...
+		guestName = out.module.Name()
+	}
+
 	// Currently, concurrent modules can conflict on name. Make sure we have
 	// a unique one.
 	instanceNum := out.instanceCounter.Add(1)
-	instanceName := out.guestName + "-" + strconv.FormatUint(instanceNum, 10)
-	moduleConfig := out.moduleConfig.WithName(instanceName)
+	instanceName := guestName + "-" + strconv.FormatUint(instanceNum, 10)
+	moduleConfig := wasm.NewModuleConfig(out.meta).WithName(instanceName)
```
ItalyPaleAle, line 102: I think the config should be per-component and not per-invocation IMHO

ItalyPaleAle (commented):
Thanks for this, it looks good! Some comments

bindings/wasm/output.go
codefromthecrypt: you mean to cache the `wasm.NewModuleConfig(out.meta)` thing as a field? can to. the other parts are per-invocation.

internal/wasm/wasm.go
codefromthecrypt: TL;DR; I recommend starting with strict off until at least global config is available.

It is early enough that we can turn it on by default, but maybe we should get the higher level wiring in? otherwise, we'll get people having to make the fine grained config to turn it off vs teaching them to use the new knob discussed in https://github.com/dapr/dapr/issues/6422

Even though this avoids some problems, there will be questions like we get in slack. Like why does something run too fast in a bad way?  (sleep is turned off) Why is the date wrong? things like this. So if we turn it more strict, we should have our config story sorted out first.

bindings/wasm/output.go
Taction: It is configured on the component level(current by metadata and this may need to change depending on how we decide to pass this configuration). Because wasm binding creates a new wasm instance for every request, so the config needs to be used in Invoke method.

codefromthecrypt: yeah what @Taction said is the better answer. the base config from meta can be moved to a field, but it would really likely be moved back later

codefromthecrypt:
@Taction seems this can wait until your draft PR upstream? then no need for separate PRs or should we merge this first?

Taction:
@codefromthecrypt this pr should be merged first, but should be after deciding how to pass strictSandbox configuration.

My upstream pr passes the configuration via context, which could also probably be passed by metadata. So waiting for dapr maintainers who may have ideas on that. If the current draft pr is accepted, there will need some small changes to this pr(mainly about how to get the configuration).

codefromthecrypt:
ok I will clean this up for merge. thanks for the advice @Taction!

codefromthecrypt:
next step:

1. create internal/wasm/testdata/strict/main.go that hits all of these functions and compile with tinygo
2. add TestNewModuleConfig to internal/wasm/wasm_test.go and make the following 2 tests:
  * when strict the output is deterministic except wall time.
    * For example, using the random source twice should return the same value.
  * when not strict the output of running testdata/strict/main.wasm is not-deterministic at all.

Once that's done we should be good. If I don't have this by monday, please nag me!

codefromthecrypt:
ok I added the tests, so ready for real review I think. Thanks for the patience.

codefromthecrypt:
I also rebuilt all the wasm with the latest TinyGo 0.28.1. The more important release notes are on 0.28 which is mostly about portability with libraries like encoding/json.

ItalyPaleAle (approved):

ItalyPaleAle (approved):
Thanks for the ping, LGTM

ItalyPaleAle:
@codefromthecrypt it seems that the added unit test is not working on Windows, where it fails 100% of the time. The reported duration is always 0: https://github.com/dapr/components-contrib/actions/runs/5754171576/job/15598923556

Any thoughts?

ItalyPaleAle:
Thanks for the fix!

btw to you think maybe we could still enforce a minimum on Linux but not windows?

codefromthecrypt:
> btw to you think maybe we could still enforce a minimum on Linux but not windows?

We could, but it isn't high signal I think. Basically, we are testing it takes "some time" to sleep ;) Happy either way. If we want to test on non-windows, I would set the test time based on runtime.GOOS != "windows"

ItalyPaleAle:
> > btw to you think maybe we could still enforce a minimum on Linux but not windows?
> 
> We could, but it isn't high signal I think. Basically, we are testing it takes "some time" to sleep ;) Happy either way. If we want to test on non-windows, I would set the test time based on runtime.GOOS != "windows"

I think you're right, it's not very useful signal.

PR #2888, opened 2023-06-06
 Draft: Redis return ttlExpiryTime in GetResponse

JoshVanL:


tests/certification/state/mongodb/mongodb_test.go
```diff
@@ -67,6 +67,7 @@ func TestMongoDB(t *testing.T) {
 		item, err := client.GetState(ctx, stateStoreName, certificationTestPrefix+"key1", nil)
 		assert.NoError(t, err)
 		assert.Equal(t, "mongodbCert", string(item.Value))
+		assert.NotContains(t, item.Metadata, "ttlExpireTime")
```
ItalyPaleAle, line 70: For the future, no need to update certification tests to validate this new behavior, since we're going to have a conformance test for this. It's not that it's bad, it's just not necessary :)

ItalyPaleAle (approved):

PR #2890, opened 2023-06-06
memached: if tllInSeconds larger than 30 days, convert to unix timestamp

JoshVanL:
Currently, we are always passing the `ttlInSeconds` value as-is to memcached however, any value over 30 days will be treated as a UInix timestamp by memcached so will probably mark the key for expiration immediately.

PR changed client to conver `ttlInSeconds` to Unix timestamp when value is larger than 30 days.

https://github.com/memcached/memcached/wiki/Commands#standard-protocol

### DIFF
--- a/state/memcached/memcached.go
+++ b/state/memcached/memcached.go
@@ -141,7 +141,7 @@ func (m *Memcached) parseTTL(req *state.SetRequest) (*int32, error) {
 
 		// If ttl is more than 30 days, convert it to unix timestamp.
 		// https://github.com/memcached/memcached/wiki/Commands#standard-protocol
-		if parsedInt > 60*60*24*30 {
+		if parsedInt >= 60*60*24*30 {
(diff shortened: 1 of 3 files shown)


state/memcached/memcached.go
```diff
@@ -133,8 +136,15 @@ func (m *Memcached) parseTTL(req *state.SetRequest) (*int32, error) {
 		if err != nil {
 			return nil, err
 		}
+
 		parsedInt := int32(parsedVal)
 
+		// If ttl is more than 30 days, convert it to unix timestamp.
+		// https://github.com/memcached/memcached/wiki/Commands#standard-protocol
+		if parsedInt > 60*60*24*30 {
```
ItalyPaleAle, line 144:
```suggestion
		if parsedInt > (60*60*24*30)-1 {
```

Does this need a -1?

ItalyPaleAle (approved):

PR #2891, opened 2023-06-06
etcd: return ttlExpiryTime in GetResponse

JoshVanL:

### DIFF
--- a/state/aws/dynamodb/dynamodb.go
+++ b/state/aws/dynamodb/dynamodb.go
@@ -121,9 +121,10 @@ func (d *StateStore) Get(ctx context.Context, req *state.GetRequest) (*state.Get
 		return nil, err
 	}
 
-	var ttl int64
+	var metadata map[string]string
 	if d.ttlAttributeName != "" {
 		if val, ok := result.Item[d.ttlAttributeName]; ok {
+			var ttl int64
 			if err = dynamodbattribute.Unmarshal(val, &ttl); err != nil {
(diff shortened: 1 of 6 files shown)


state/etcd/etcd.go
```diff
@@ -141,9 +144,25 @@ func (e *Etcd) Get(ctx context.Context, req *state.GetRequest) (*state.GetRespon
 		return &state.GetResponse{}, nil
 	}
 
+	var metadata map[string]string
+	if resp.Kvs[0].Lease > 0 {
+		leaseResp, err := e.client.TimeToLive(ctx, clientv3.LeaseID(resp.Kvs[0].Lease))
```
ItalyPaleAle, line 149: Does this make a network call?

JoshVanL: Yes, but there is no other way around this.

ItalyPaleAle: I am concerned the perf impact in this case is not insigificant if we need to make an extra roundtrip to the database

JoshVanL: Indeed, I agree. It's also not even taking the now() clock time of the etcd process.

ItalyPaleAle:
Can this be closed in favor of #2932?

JoshVanL:
@ItalyPaleAle yeah I'm happy to do that if you don't mind the review :slightly_smiling_face: 

I'll update now.

/close

Issue #2568, opened 2023-02-22
Create component metadata for bindings.aws.s3


artursouza:
Create component metadata for bindings.aws.s3, like in https://github.com/dapr/components-contrib/blob/master/state/azure/cosmosdb/metadata.yaml

PR #2904, opened 2023-06-10
[AWS Bindings S3] Adds Component Metadata Schema

robertojrojas:
# Description
Adds AWS Bindings S3 Component Metadata Schema.

## Issue reference
this PR will close: #2568


bindings/aws/s3/metadata.yaml
```diff
@@ -0,0 +1,125 @@
...
+    - name: list
+      description: "List blob."
+capabilities: []
+authenticationProfiles:
+  - title: "AccessKey and secretKey"
+    description: "Authenticate using accessKey and secretKey."
+    metadata:
+      - name: accessKey
+        type: string
+        required: false
```
artursouza, line 29: Shouldn't this be required within this profile?

bindings/aws/s3/metadata.yaml
```diff
@@ -0,0 +1,125 @@
...
+        sensitive: true
+        description: |
+          ID of the AWS account with appropriate permissions to SNS and SQS.
+          Can be secretKeyRef to use a secret reference.
+        example:  "AKIAIOSFODNN7EXAMPLE"
+        default: ""
+      - name: secretKey
+        type: string
+        required: false
```
artursouza, line 38: Shouldn't this be required within this profile?

bindings/aws/s3/metadata.yaml
```diff
@@ -0,0 +1,125 @@
...
+      description: "Create blob."
+    - name: get
+      description: "Get blob."
+    - name: delete
+      description: "Delete blob."
+    - name: list
+      description: "List blob."
+capabilities: []
+authenticationProfiles:
+  - title: "AccessKey and secretKey"
```
ItalyPaleAle, line 24: As mentioned in the other PR, we should create a shared auth profile for these

robertojrojas: @ItalyPaleAle Yes, while I agree with adding the shared/common auth profile, I think it should be addressed as a separate PR.
I went ahead and created an issue for it: **[Component Metadata Schema] Extend to Support Shared Auth Profile** #2935 
At the time this feature is added the various components, like AWS, etc... can be modified to take advantage of the feature.

Unless there are any other changes to this current PR, I think we should just complete it.

ItalyPaleAle: Let’s do #2935 first. I’d rather have things done correctly right away and merge things once they’re ready. Adding a new auth profile shouldn’t be a lot of work: you need to add it here: https://github.com/dapr/components-contrib/blob/master/.build-tools/pkg/metadataschema/builtin-authentication-profiles.go

ItalyPaleAle (changes requested):
Waiting on #2935

bindings/aws/s3/metadata.yaml
```diff
@@ -0,0 +1,122 @@
...
+  output: true
+  operations:
+    - name: create
+      description: "Create blob."
+    - name: get
+      description: "Get blob."
+    - name: delete
+      description: "Delete blob."
+    - name: list
+      description: "List blob."
```
ItalyPaleAle, line 21: Nit:

```suggestion
    - name: create
      description: "Create blob"
    - name: get
      description: "Get blob"
    - name: delete
      description: "Delete blob"
    - name: list
      description: "List blob"
```

bindings/aws/s3/metadata.yaml
```diff
@@ -0,0 +1,122 @@
...
+      example: "AKIAIOSFODNN7EXAMPLE"
+      default: ""
+    - name: secretKey
+      type: string
+      required: true
+      sensitive: true
+      description: |
+        The secret key associated with the access key.
+      example: "wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY"
+      default: ""
```
ItalyPaleAle, line 41: These are included in the built-in authentication profile so need to be removed from here

```suggestion
```

bindings/aws/s3/metadata.yaml
```diff
@@ -0,0 +1,122 @@
...
+    description: |
+      The name of the S3 bucket to write to.
+    example: '"bucket"'
+    type: string
+  - name: region
+    required: true
+    description: |
+      The AWS region to the instance. See this page for valid regions:
+      https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.RegionsAndAvailabilityZones.html. 
+      Ensure that DynamoDB are available in that region.
```
ItalyPaleAle, line 54: DynamoDB?

bindings/aws/s3/metadata.yaml
```diff
@@ -0,0 +1,122 @@
...
+    example: '"true", "false"'
+  - name: encodeBase64
+    description: |
+      Configuration to encode base64 file content before return the content.
+      (In case of opening a file with binary content). true is the only allowed
+      positive value. Other positive variations like "True", "1" are not acceptable.  
+    type: bool
+    default: 'false'
+    example: '"true", "false"'
+  - name: disableSSL
```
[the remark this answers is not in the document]
ItalyPaleAle: Yes I mean can we rename those properties to `disableTLS` and `insecureTLS`? Maintain the old ones and aliases

bindings/aws/s3/metadata.yaml
ItalyPaleAle: Why do we need DynamoDB for S3?

robertojrojas: that's a typo. Fixing it.

bindings/aws/s3/metadata.yaml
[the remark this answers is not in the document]
robertojrojas: I'm curious. Will making these change hold this PR?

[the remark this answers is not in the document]
ItalyPaleAle: No, no! I asked if you can open an issue to track that however please :)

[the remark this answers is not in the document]
robertojrojas: no problem. I'll open the PR.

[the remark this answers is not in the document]
robertojrojas: https://github.com/dapr/components-contrib/issues/2983

bindings/aws/s3/metadata.yaml
```diff
@@ -0,0 +1,102 @@
...
+    required: true
+    description: |
+      The name of the S3 bucket to write to.
+    example: '"bucket"'
+    type: string
+  - name: region
+    required: true
+    description: |
+      The AWS region to the instance. See this page for valid regions:
+      https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.RegionsAndAvailabilityZones.html.
```
ItalyPaleAle, line 36: URLs should go in the `url` property not in the description.

Example:

https://github.com/dapr/components-contrib/blob/21885b306b5f46efb3f89c6ee90f78dec68e5f1b/configuration/redis/metadata.yaml#L87-L89

bindings/aws/s3/metadata.yaml
```diff
@@ -0,0 +1,102 @@
...
+  - name: bucket
+    required: true
+    description: |
+      The name of the S3 bucket to write to.
+    example: '"bucket"'
+    type: string
+  - name: region
+    required: true
+    description: |
+      The AWS region to the instance. See this page for valid regions:
```
ItalyPaleAle, line 35: "AWS region to the instance" sounds odd. Maybe "The AWS region where the bucket exists"?

bindings/aws/s3/metadata.yaml
```diff
@@ -0,0 +1,102 @@
...
+      metadata key on a create request that contains empty data fields.
+    example: '"mydir/my-test-file.txt"'
+    type: string
+  - name: presignTTL
+    required: false
+    description: |
+      To presign an object with a specified time-to-live, use the presignTTL metadata
+      key on a create request, as a Go duration.
+    example: '"2h45m"'
+    type: duration
```
ItalyPaleAle, line 102: According to the source code and the docs, `filePath` and `presignTTL` are keys used in the metadata of requests, not for the component. They should not be listed here.

https://docs.dapr.io/reference/components-reference/supported-bindings/s3/

ItalyPaleAle (changes requested):

bindings/aws/s3/metadata.yaml
```diff
@@ -0,0 +1,122 @@
...
+    example: '"http://localhost:4566"'
+    type: string
+  - name: sessionToken
+    required: false
+    sensitive: true
+    description: |
+      AWS session token to use. A session token is only required if you are using 
+      temporary security credentials.
+    example: '"TOKEN"'
+    type: string
```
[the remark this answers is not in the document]
artursouza: @robertojrojas

artursouza (approved):

artursouza:
@ItalyPaleAle Can we merge this?

PR #2910, opened 2023-06-13
State stores: conformace test to check special characters are not mangled

JoshVanL:
Conformance test for all state stores to ensure that all state stores are able to store and return keys who's values are string which contain special characters.

Adds new `state/utils.JSONStringify` function for converting a `[]byte` into a jsonified `[]byte` string. Update state stores which were incorrectly encoding special characters in string state values on set.

Question: Why are we doing anything to `Values` when Setting to state stores? Why do we not Set them as-is?


state/utils/utils.go
```diff
@@ -13,6 +13,26 @@ limitations under the License.
...
+	"fmt"
+	"strings"
+)
+
+func JSONStringify(value any) ([]byte, error) {
+	switch value := value.(type) {
+	case []byte:
+		return value, nil
+	case int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64, float32, float64, bool:
+		return []byte(fmt.Sprintf("%v", value)), nil
```
ItalyPaleAle, line 27: Maybe a bit of a nit, but I think this would be better (better perf) if each type were handled separately, such as:

```go
case int, int8, int16, int32, int64:
    return []byte(strconv.FormatInt(int64(value), 10))
case uint, uint8, uint16, uint32, uint64:
    return []byte(strconv.FormatUint(uint64(value), 10))
case bool:
    if value {
        return []byte("true")
    } else {
        return []byte("false")
    }
// Etc
```

state/utils/utils.go
```diff
@@ -13,6 +13,26 @@ limitations under the License.
...
+	case string:
+		return []byte(`"` + strings.ReplaceAll(value, `"`, `\"`) + `"`), nil
+	default:
+		return json.Marshal(value)
+	}
+}
+
+// Deprecated: use JSONStringify instead.
```
ItalyPaleAle, line 35: Should this be deleted then?

ItalyPaleAle (commented):
What I am concerned with: is this going to cause backwards-incompatibility?

1. What about reading values created with Dapr "vCurrent" using "vNext"?
2. What about reading values created with Dapr "vNext" into "vCurrent"?

state/in-memory/in_memory.go
```diff
@@ -377,15 +357,16 @@ func (store *inMemoryStore) Multi(ctx context.Context, request *state.Transactio
 
 	// step3: do really set
 	// these operations won't fail
+	var err error
 	for _, o := range request.Operations {
 		switch req := o.(type) {
 		case *innerSetRequest:
-			store.doSet(ctx, req.req.Key, req.data, req.ttl)
+			err = errors.Join(err, store.doSet(ctx, req.req.Key, req.req.Value, req.ttl))
```
ItalyPaleAle, line 364: Why the errors.Join here?

artursouza:
> What I am concerned with: is this going to cause backwards-incompatibility?
> 
> 1. What about reading values created with Dapr "vCurrent" using "vNext"?
> 2. What about reading values created with Dapr "vNext" into "vCurrent"?

This is a common debate of fixing a bug but having customers relying on the bug. We do have a way to workaround that, it is to bump the component version and keep the old version untouched. It would require to copy and paste the code and deprecate v1 of each component that has a breaking change.

ItalyPaleAle:
Even if we made a "v2", something I would consider a last-resort as it would double the maintenance burden on maintainers of this repo, we also need a solution for data that was already written. Otherwise, users (who, on average, never read docs, let alone changelogs ) will just switch "v1" to "v2" and then complain Dapr broke their data 🙃

@JoshVanL what are the "broken" components in 1.11.0?

ItalyPaleAle:
Some additional interesting findings. The interface in components accepts data as `any`, so I was wondering what kind of data actually gets stored from Dapr.

1. When data comes from the gRPC API, it's always received by the component as a `[]byte`. Some components base64-encode that before storing it.
2. When it comes from HTTP, it's whatever it is in the JSON that it's received. For example, if you pass a number, it's a `float64` (because JavaScript...), strings are strings, bools are bools, and objects/arrays are received by the component as-is so are serialized as JSON.

This seems like it can cause a lot of inconsistencies also depending on how the data is stored and retrieved.

_Perhaps the problem isn't really having a v2 of components, but a v2 of the state store APIs that removes all these inconsistencies..._

state/utils/utils.go
```diff
@@ -13,6 +13,26 @@ limitations under the License.
...
+func JSONStringify(value any) ([]byte, error) {
+	switch value := value.(type) {
+	case []byte:
+		return value, nil
+	case int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64, float32, float64, bool:
+		return []byte(fmt.Sprintf("%v", value)), nil
+	case string:
+		return []byte(`"` + strings.ReplaceAll(value, `"`, `\"`) + `"`), nil
+	default:
+		return json.Marshal(value)
```
ItalyPaleAle, line 31: Also here, consider using `json.NewEncoder` so you can [`SetEscapeHTML()`](https://pkg.go.dev/encoding/json#Encoder.SetEscapeHTML)

Maybe the HTML escaping is the root of all of our problems?

state/utils/utils.go
JoshVanL: It is still references in `dapr/dapr`. We can remove once we have merged this, updated go mod and migrated the caller in `dapr/dapr`

state/in-memory/in_memory.go
JoshVanL: We are inside a range, and want to catch all `doSet` errors.

ItalyPaleAle: However this is not really how `errors.Join` should be used, as the Go team told me too: https://github.com/golang/go/issues/60209
Should collect the errors in a slice and then call `errors.Join` once

ItalyPaleAle:
List here:

image

Aside from in-memory which doesn't matter, I am very surprised to see MariaDB in the list but not MySQL, since they should be the same...

PR #2911, opened 2023-06-13
Updated pinned runtime version in cert tests

ItalyPaleAle:
See: dapr/dapr#6301


DeepanshuA (approved):

ItalyPaleAle:
The SQL Server certification test is not working, but it seems to be broken in the master branch too. Merging this and will investigate the failure after.

PR #2912, opened 2023-06-13
Fixes for SQL Server state store

ItalyPaleAle:
Fixes SQL Server certification tests and integration tests.

1. Fixed stored procedure for SQL Server. Thanks @JoshVanL 
1. Updates the Docker image to a more modern base (rather than Ubuntu 16.04)
2. Fixes cert tests that were failing
3. Fixes integration tests

I had to manually delete the stored procedure that existed in the test database (and have it re-created) since it seems it was stuck at an old version


artursouza (approved):

berndverst:
@ItalyPaleAle @artursouza Can we open this against 1.11 instead and then merge back to master?

berndverst (approved):

PR #2920, opened 2023-06-19
[release-1.11] Cherry-pick sql server fixes #2912

berndverst:
# Description

[release-1.11] Cherry-pick sql server fixes #2912

### DIFF
--- a/state/sqlserver/migration.go
+++ b/state/sqlserver/migration.go
@@ -46,8 +46,8 @@ func newMigration(metadata *sqlServerMetadata) migrator {
 func (m *migration) newMigrationResult() migrationResult {
 	r := migrationResult{
 		itemRefTableTypeName:     fmt.Sprintf("[%s].%s_Table", m.metadata.Schema, m.metadata.TableName),
-		upsertProcName:           fmt.Sprintf("sp_Upsert_v3_%s", m.metadata.TableName),
-		getCommand:               fmt.Sprintf("SELECT [Data], [RowVersion] FROM [%s].[%s] WHERE [Key] = @Key AND ([ExpireDate] IS NULL OR [ExpireDate] > GETDATE())", m.metadata.Schema, m.metadata.TableName),
+		upsertProcName:           fmt.Sprintf("sp_Upsert_v4_%s", m.metadata.TableName),
+		getCommand:               fmt.Sprintf("SELECT [Data], [RowVersion], [ExpireDate] FROM [%s].[%s] WHERE [Key] = @Key AND ([ExpireDate] IS NULL OR [ExpireDate] > GETDATE())", m.metadata.Schema, m.metadata.TableName),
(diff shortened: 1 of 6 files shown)


state/sqlserver/migration.go
```diff
@@ -46,8 +46,8 @@ func newMigration(metadata *sqlServerMetadata) migrator {
 func (m *migration) newMigrationResult() migrationResult {
 	r := migrationResult{
 		itemRefTableTypeName:     fmt.Sprintf("[%s].%s_Table", m.metadata.Schema, m.metadata.TableName),
-		upsertProcName:           fmt.Sprintf("sp_Upsert_v3_%s", m.metadata.TableName),
-		getCommand:               fmt.Sprintf("SELECT [Data], [RowVersion] FROM [%s].[%s] WHERE [Key] = @Key AND ([ExpireDate] IS NULL OR [ExpireDate] > GETDATE())", m.metadata.Schema, m.metadata.TableName),
+		upsertProcName:           fmt.Sprintf("sp_Upsert_v4_%s", m.metadata.TableName),
+		getCommand:               fmt.Sprintf("SELECT [Data], [RowVersion], [ExpireDate] FROM [%s].[%s] WHERE [Key] = @Key AND ([ExpireDate] IS NULL OR [ExpireDate] > GETDATE())", m.metadata.Schema, m.metadata.TableName),
```
berndverst, line 50: should I remove `[ExpireDate]` here?

ItalyPaleAle: No, I would leave it because then the stored procedure is different so when using 1.12, we need to create a new version. There's no harm in returning an extra value IMHO

ItalyPaleAle (approved):

ItalyPaleAle (changes requested):
Looks like there is a syntax change and SQL Server is failing. Please take a look

state/sqlserver/migration.go
berndverst: This is causing issues after all during the Get() method where `Scan` now needs more params.

`err = rows.Scan(&data, &rowVersion)`

berndverst: Well the `getCommand` now is:
```go
getCommand:               fmt.Sprintf("SELECT [Data], [RowVersion], [ExpireDate] FROM [%s].[%s] WHERE [Key] = @Key AND ([ExpireDate] IS NULL OR [ExpireDate] > GETDATE())", m.metadata.Schema, m.metadata.TableName),
```

berndverst:
Do you think the scan should be something like this?

```go
	var data string
	var rowVersion []byte
	var ttlTime *time.Time
	err = rows.Scan(&data, &rowVersion, &ttlTime)
	if err != nil {
		return nil, err
	}
```

At that point maybe `ttlInSeconds` should actually be returned in the metadata also? Maybe as `{'ttlInSeconds': (time.Now() - *ttlTime).Seconds()}`.
Though that could run into problems with time zones perhaps.

ItalyPaleAle (dismissed):

ItalyPaleAle (approved):

ItalyPaleAle (approved):

DeepanshuA (approved):

PR #2925, opened 2023-06-21
Azure Key Vault crypto: fixed JSON marshalling of public keys

ItalyPaleAle:
When marshalling the object returned by Azure Key Vault as JSON, it is now correctly formatted as JWK

This is blocking the completion of E2E tests for crypto: https://github.com/dapr/dapr/pull/6546

### DIFF
--- a/crypto/key.go
+++ b/crypto/key.go
@@ -14,6 +14,7 @@ limitations under the License.
 package crypto
 
 import (
+	"encoding/json"
 	"time"
 
 	"github.com/lestrrat-go/jwx/v2/jwk"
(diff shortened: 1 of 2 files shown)


crypto/key.go
```diff
@@ -66,6 +67,12 @@ func (k Key) isValidAtTime(t time.Time) bool {
 	return true
 }
 
+// MarshalJSON implements the json.Marshaler interface
+func (k Key) MarshalJSON() ([]byte, error) {
```
DeepanshuA, line 71: Do we want to rename it to `MarshalKey`, if it is a specific thing OR if it is a common utility kind of thing, then probably move it to a common utils kinda file?

DeepanshuA (approved):

crypto/key.go
ItalyPaleAle: It has to be called MarshalJSON: https://pkg.go.dev/encoding/json#Marshaler

essentially what this method does is overriding what json.Marshal does. Without it, the result is a dictionary with the key “Key” and the JWK inside that. That’s incorrect. This overrides the behavior so when the object is marshaled as JSON, it only marshals the value of the Key property.

PR #2926, opened 2023-06-21
Add support for bulkpublish

arturotrenard:
# Description

The idea of this PR is to allow the custom implementation of the BulkPublish functionality in pubsub pluggable components.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


artursouza (changes requested):
I don't think this is needed. Pluggable components use a gRPC API and it does not need to impact the existing contrib implementation.

PR #2932, opened 2023-06-22
ETCD State store v2

### LINKED ISSUE #2922: ETCDv2
Context:  https://github.com/dapr/components-contrib/issues/2857 https://github.com/dapr/components-contrib/pull/2891

Add ETCD version 2.

`Etcd` will determine whether the client is using version 2 by inspecting the `components.dapr.io/version` request metadata. If this value is omitted or contains the value `v1`, then `Etcd` will use the existing `v1` implementation. If `v2`, `Etcd` will use a new `v2` implementation. Dapr runtime will need to be updated to send this metadata value to all state store requests. 

The `v2` implementation of `Etcd` will implement a new envelope schema to key values. This envelope is a JSON object with the following schema:

```json
{
  "ct": int64,
  "ttl": int64,
  "data": string
}
```

`ct` holds the created time of the last value write as a Unix Epoch number representing milliseconds. Clock is determined by components-contrib as ETCD does not expose the created time of key values. Note that this can be problematic if _Daprd_ instances have different clock times when evaluating TTL expire times but is the best we can  do. Required.

`ttl` stores the granted time to live in seconds. This field may be omitted if the value was written without a time to live. Optional.

`data` stores the data value from `Set` as a json encoded string. Required.

---

The above schema would allow ETCD to support returning `ttlExpireTime` by components-contrib calculating the expiry time with `ct` and `ttl`. `Etcd` would still rely on etcd itself to expire key values using leases.

---

Since `v1` and `v2` store values incompatibly, users should not point them to the same path prefix or database completely to avoid  data corruption.

Users define which version to use via the components definition `spec.version`, which will default to `v2`.

When an etcd component is enabled for actors with version `v1`, Dapr should log a warning that `ttlExpireTime` is not supported for that version.

JoshVanL:
Closes #2922 

Adds ETCD v2, see [here](https://github.com/dapr/components-contrib/issues/2922).


state/etcd/etcd.go
```diff
@@ -80,6 +81,15 @@ func (e *Etcd) Init(_ context.Context, metadata state.Metadata) error {
 
 	e.keyPrefixPath = etcdConfig.KeyPrefixPath
 
+	switch etcdConfig.Version {
```
ItalyPaleAle, line 84: I think the version should not come from the metadata. Instead, we should have a separate version of `NewEtcdStateStore` that returns a *Etcd configured with the version we want. Then, in the runtime we register the 2 components separately, and just use the different `New*` method

JoshVanL: What's the reason for doing that?
Having it in metadata means runtime always sends the version, rather than only sending this metadata on this specific state store specific component.

JoshVanL: @ItalyPaleAle I have now updated the PR so that the Version is passed in as part of the `metadata.Base` struct as a dedicated field. This seems a more correct place to be passing information.

After merging, dapr will need to be updated so that it always bases the `Version` string, along with the `Name` of the component.

ItalyPaleAle: I was hoping for an actual separate `New` method mostly to keep hygiene and force us to do things right.

1. It seems more appropriate to do the “version routing” decision in the runtime and not in the component. The `registry` code already supports [registering versioned packages](https://github.com/dapr/dapr/blob/master/pkg/components/state/registry.go#L42) (using `name/v2` format, although we haven’t tested that thoroughly)
1. In the future there’ll be a time where we will have a “v2” of a component that is an actual full rewrite, so it will be in a separate package. This seems like a good opportunity to do the work to make sure we can support that.

JoshVanL: I've updated the PR to now use two different `New` funcs to select the schema version. PTAL @ItalyPaleAle

state/etcd/schema.go
```diff
@@ -0,0 +1,91 @@
...
+limitations under the License.
+*/
+
+package etcd
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/gogo/protobuf/proto"
```
ItalyPaleAle, line 20: Just noticed it. This is a weird package, not the official protobuf one. (And gopls seems to like to auto-complete this)

Please switch to the official protobuf then we should be good.

state/utils/ttl.go
```diff
@@ -38,3 +38,19 @@ func ParseTTL(requestMetadata map[string]string) (*int, error) {
 	}
 	return nil, nil
 }
+
+// ParseTTL64 parses the "ttlInSeconds" metadata property.
+func ParseTTL64(requestMetadata map[string]string) (*int64, error) {
+	val, found := requestMetadata[MetadataTTLKey]
+	if found && val != "" {
```
ItalyPaleAle, line 45: Nit: the `found` check here is redundant. If the value doesn’t exist, `Val` will be an empty string already.

tests/conformance/common.go
```diff
@@ -577,8 +576,10 @@ func loadStateStore(tc TestComponent) state.Store {
 		store = s_awsdynamodb.NewDynamoDBStateStore(testLogger)
 	case "aws.dynamodb.terraform":
 		store = s_awsdynamodb.NewDynamoDBStateStore(testLogger)
-	case "etcd":
-		store = s_etcd.NewEtcdStateStore(testLogger)
+	case "etcd.v1":
```
ItalyPaleAle, line 579: Does this need to be added to the test_info.mjs file?

ItalyPaleAle (changes requested):

state/etcd/etcd.go
ItalyPaleAle: Thanks!

state/etcd/schema.go
JoshVanL: Thanks  :+1:

state/utils/ttl.go
JoshVanL: Agreed. This was a copy+paste from the `int` implementation in the same package. I'll update both.

tests/conformance/common.go
JoshVanL: I think so. Thanks for the catch. Will update the file to test both

JoshVanL:
@ItalyPaleAle Please take another look :slightly_smiling_face:

ItalyPaleAle (approved):

ItalyPaleAle:
@JoshVanL Merged! Please open a PR in dapr/dapr to register this, and an issue in docs

Issue #2902, opened 2023-06-09
Pubsub rabbitmq connection_name


gchadid:
Hi 

Can we set up the rabbitmq connection_name in the pubsub.rabbitmq component?

![image](https://github.com/dapr/components-contrib/assets/10290878/c3f6b9a5-bc35-4313-a5bb-dff13e5738dd)

as describe

https://www.rabbitmq.com/connections.html#client-provided-names

best regards,

berndverst:
Whoever wants to work on this: This needs to be introduced as new metadata property and then wire up with the rabbitmq client.

Code here:
https://github.com/dapr/components-contrib/tree/master/pubsub/rabbitmq

Taction:
/assign

Add an optional `clientName` metadata, users can set it to any value, especially `{uuid}`, `{podName}` or `{appID}` (these will be replaced by Dapr runtime to the real value).

PR #2933, opened 2023-06-23
Rabbitmq pubsub add client name

Taction:
# Description

Add an optional clientName metadata, users can set it to any value, especially {uuid}, {podName} or {appID} (these will be replaced by Dapr runtime to the real value).

## Issue reference

Please reference the issue this PR will close: #2902

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


yaron2:
@Taction please resolve the conflicts, otherwise LGTM

berndverst:
ping @Taction

Also do you want to make the locale and the heartbeat configurable as well?

Taction:
@berndverst made heartBeat configurable. According to the SDK comment, the locale is less likely to be changed:
```
        // Connection locale that we expect to always be en_US
	// Even though servers must return it as per the AMQP 0-9-1 spec,
	// we are not aware of it being used other than to satisfy the spec requirements
	Locale string
```

berndverst (approved):

PR #2947, opened 2023-06-29
Local storage binding: disable access to other system folders for security reasons

ItalyPaleAle:
@JoshVanL recently taught me that by reading the `/proc` folder, apps can get access to sensitive information, including env vars (that we disallow through the env var secret store).

This adds new denied paths:

- `/proc`
- `/sys`
- `/boot`


bindings/localstorage/localstorage.go
```diff
@@ -40,6 +40,7 @@ const (
 
 // List of root paths that are disallowed
 var disallowedRootPaths = []string{
+	filepath.Clean("/proc"),
```
JoshVanL, line 43: Probably also worth including `/sys`, `/boot`, and `/root`. There may be more that we might want to consider. `/tmp`?

JoshVanL (changes requested):

bindings/localstorage/localstorage.go
ItalyPaleAle: For `/root`:

- Default containers run as rootless so `/root` should not be readable.
- If someone is using the debug containers and want to run as root, then accessing `/root` is accessing the home folder, so it can make sense

I'll add `/sys` and `/boot`. I think `/tmp` needs to be allowed

berndverst (approved):

PR #2963, opened 2023-07-03
[WIP] OAuth2 Middleware: Store token in a cookie

### LINKED ISSUE #2635: Oauth2 Middleware uses in-memory session state for code grant flow and token storage
## Expected Behavior

## Actual Behavior

For scale:

For resilience:
2. Restart daprd

## Proposals

## Release Note

ItalyPaleAle:
Fixes #2635

Because cookies now store the token encrypted, users need to provide an encryption key via the **new metadata property `cookieKey`**:

  - This property is required.
  - *Technically speaking, `cookieKey` is not the actual encryption key used: the Dapr runtime deterministically derives both a signing key and an encryption key from that*

## Work In Progress

This PR is currently a WIP because of an issue with handling large tokens.

- Some IdPs can return tokens that are very large. Chief among them is Azure AD, which recently increased the size of the auth tokens it issues. I have now observed that a token from Azure AD can be well over 2KB without many claims; some users report seeing tokens much larger than that.
- Although the specs don't put any limit on the size of cookies, most browsers seem to silently reject cookies that are larger than 4KB (technically, [4093 bytes](https://stackoverflow.com/questions/640938/what-is-the-maximum-size-of-a-web-browsers-cookies-key/4604212#4604212))

Solutions:

- The most immediate fix was to enable compression on the JWT. Encrypted JWTs are based on JWE, and RFC7516 that defines JWE does include [support for compression](https://www.rfc-editor.org/rfc/rfc7516#section-4.1.3). Compression is enabled by default if the token is larger than 800 bytes. However, compression helps only to a certain point, and Azure AD can return [tokens that are much larger](https://techcommunity.microsoft.com/t5/microsoft-entra-azure-ad/azure-ad-b2c-authorization-code-and-refresh-token-size-increase/td-p/3770890).
- Two possible solutions have not been implemented yet:
   1. Currently the component stores tokens only in cookies. We could offer an option to use headers, such as the Authorization header (which has less problems with data size), instead. This would not work well if clients are browsers, but other apps should be fine.
   2. We could offer optional support for storing the token on the server-side, likely using an internal actor (see https://github.com/dapr/dapr/issues/6538). Actors seem well-suited for this use case because they allow using all supported state stores, and the use of reminders allows us to efficiently clean up expired sessions. In this case, if the token was too large to be self-contained in a cookie, we'd store it in an actor and only send a reference (a "session ID") to the user.


ItalyPaleAle:
Closing in favor of #2964

middleware/http/oauth2/oauth2_middleware.go
```diff
@@ -15,144 +15,299 @@ package oauth2
...
+		Secure:   isSecure,
+		Domain:   domain,
+		Path:     "/",
+		SameSite: http.SameSiteLaxMode,
+	}
+	cookieStr := cookie.String()
+
+	// Browsers have a maximum size of about 4KB, and if the cookie is larger than that (by looking at the entire value of the header), it is silently rejected
+	// Some info: https://stackoverflow.com/a/4604212/192024
+	if len(cookieStr) > 4<<10 {
```
drewby, line 289: What happens if the backend service also adds cookie values on subsequent requests? Is the total limit still 4k?

drewby (commented):
Overall, looks good given the cookie limitations.  I put some minor comments and a question inline.

Given the limitations, I think you still need a solution that uses a distributed cache or some shared session store server side.

ItalyPaleAle:
Thanks for the review! @drewby  This helps. I do want to implement support for storing tokens on the server, and that will likely involve using Dapr Actors. I am tracking that as a follow-up when #2967 is merged

middleware/http/oauth2/oauth2_middleware.go
ItalyPaleAle: Yes, it's still 4KB per domain in most browsers. If the cookies go over the limit, they are silently dropped

middleware/http/oauth2/oauth2_middleware.go
```diff
@@ -15,144 +15,299 @@ package oauth2
...
-	Scopes         string `json:"scopes" mapstructure:"scopes"`
-	AuthURL        string `json:"authURL" mapstructure:"authURL"`
-	TokenURL       string `json:"tokenURL" mapstructure:"tokenURL"`
-	AuthHeaderName string `json:"authHeaderName" mapstructure:"authHeaderName"`
-	RedirectURL    string `json:"redirectURL" mapstructure:"redirectURL"`
-	ForceHTTPS     string `json:"forceHTTPS" mapstructure:"forceHTTPS"`
-}
+const (
+	// Timeout for authenticating with the IdP
+	authenticationTimeout = 10 * time.Minute
```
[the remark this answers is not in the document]
ItalyPaleAle: Yes it's a timeout on authentication requests. Because we cannot influence the timeout on the IdP, we can only set an expiration on the cookie with the state token.

I am not sure this needs to be configurable. 10 mins should be plenty for someone to authenticate.

PR #2970, opened 2023-07-07
Add Azure AD support to Postgres state store component

### LINKED ISSUE #2595: Authenticate PostgreSQL components using Azure AD
We are able to access the configuration store from an Azure PostgreSQL database using a connection string with password. 

Subsequently, we configured the database server with AAD auth and the AKS cluster was setup with [Workload Identity](https://learn.microsoft.com/en-us/azure/aks/workload-identity-overview) to allow the Pods to leverage a User Assigned Managed Identity. This requires calling the [token issuer endpoint](https://learn.microsoft.com/en-us/azure/postgresql/single-server/how-to-connect-with-managed-identity) to retrieve the password for the managed identity and then setting that token as the password in the connection string that is used to access the PostgreSQL. The [configuration store binding for PostgreSQL](https://github.com/dapr/components-contrib/blob/master/configuration/postgres/postgres.go) does not do this token access, so it looks like it will have to be done by the app. 

Any thoughts on how this can be accomplished? I tried overwriting the _connectionString_ using the _GetConfigurationRequest_ class and calling _setMetadata()_ to set the connection string, but that did not seem to work. I don't see much documentation on this _setMetadata_ call - what is it used for?
  
## Components

- [X] State store: #2970
- [x] Binding: #2971
- [x] Configuration store: #2971

ItalyPaleAle:
Part of #2595

Adds Azure AD support to Postgres state store component.

It requires using a new metadata key `useAzureAD` (similar to how it works for the SQL Server component) set to `true`, and then pass the Azure AD credentials.

> Also added metadata option `maxConns` that indicates how many connections PGX can keep in a pool


.github/infrastructure/conformance/azure/setup-azure-conf-test.sh
```diff
@@ -229,6 +229,8 @@ SQL_SERVER_NAME_VAR_NAME="AzureSqlServerName"
 SQL_SERVER_DB_NAME_VAR_NAME="AzureSqlServerDbName"
 SQL_SERVER_CONNECTION_STRING_VAR_NAME="AzureSqlServerConnectionString"
 
+AZURE_DB_POSTGRES_CONNSTRING_VAR_NAME="AzureDBPostgresConnectionString"
```
ItalyPaleAle, line 232: I've already applied these changes (creating the Postgres DB) in our infra

ItalyPaleAle:
Tests passed on both Docker and Azure:
image

https://github.com/dapr/components-contrib/actions/runs/5482549862/

(No need to do /ok-to-test here since it's running from a branch in the repo and not a fork)

internal/component/postgresql/postgresql.go
```diff
@@ -31,9 +31,10 @@ type PostgreSQL struct {
 }
 
 type Options struct {
-	MigrateFn  func(context.Context, PGXPoolConn, MigrateOptions) error
-	SetQueryFn func(*state.SetRequest, SetQueryOptions) string
-	ETagColumn string
+	MigrateFn     func(context.Context, PGXPoolConn, MigrateOptions) error
+	SetQueryFn    func(*state.SetRequest, SetQueryOptions) string
+	ETagColumn    string
+	EnableAzureAD bool
```
ItalyPaleAle, line 37: This is needed because CockroachDB can't use Azure AD

DeepanshuA (approved):

PR #2971, opened 2023-07-07
Add Azure AD support to Postgres configuration store and bindings

### LINKED ISSUE #2595: Authenticate PostgreSQL components using Azure AD

ItalyPaleAle:
Fixes #2595

Completes the work started with #2970 and adds Azure AD support to the 2 remaining Postgres components: the binding and configuration store.

Also included:

- Conformance tests with the "azure" flavor that validate that connection with Azure AD works (for both components)
- Metadata options update:
  - Binding:
    - Renamed "url" to "connectionString" so it's the same as in the other components. "url" remains available for backwards-compatibility
    - Component got "free" support for "maxConns" and "connectionMaxIdleTime" metadata properties
  - Configuration store:
    - Renamed "connMaxIdleTime" to "connectionMaxIdleTime" for consistency with other components. "connmaxIdleTime" remains available for backwards-comaptibility
  - All components gained the "useAzureAD" metadata option to connect with Azure
- Some other small tweaks:
  -  Postgres binding had log statements that were printing out the entire queries, likely exposing PII
  - Fixed metadata.yaml for configuration store, including reporting the component as stable and correcting name to "postgresql"

PS: This PR is opened from a branch in the official repo, so no need to do /ok-to-test to test the new cloud conformance tests


.github/infrastructure/conformance/azure/setup-azure-conf-test.sh
```diff
@@ -693,6 +696,15 @@ AZURE_DB_POSTGRES_CONNSTRING="host=${PREFIX}-conf-test-pg.postgres.database.azur
 echo export ${AZURE_DB_POSTGRES_CONNSTRING_VAR_NAME}=\"${AZURE_DB_POSTGRES_CONNSTRING}\" >> "${ENV_CONFIG_FILENAME}"
 az keyvault secret set --name "${AZURE_DB_POSTGRES_CONNSTRING_VAR_NAME}" --vault-name "${KEYVAULT_NAME}" --value "${AZURE_DB_POSTGRES_CONNSTRING}"
 
+echo export ${AZURE_DB_POSTGRES_CLIENT_ID_VAR_NAME}=\"${SDK_AUTH_SP_APPID}\" >> "${ENV_CONFIG_FILENAME}"
```
ItalyPaleAle, line 699: These changes have already been applied to our infra

berndverst (approved):

berndverst:
@ItalyPaleAle resolve conflict please

berndverst (approved):

PR #2975, opened 2023-07-10
MySQL binding: allow passing parameters for queries

### LINKED ISSUE #2973: MySQL binding: add support for parameters in queries
## Describe the feature

The MySQL binding doesn't support parametrized queries. Using parameters is extremely important to prevent SQL injection attacks.

## Release Note

RELEASE NOTE: **ADD** Support for parametrized queries to MySQL binding

ItalyPaleAle:
Fixes #2973 

Adds support for query parameters in the MySQL binding. This allows executing parametrized queries (such as `SELECT * FROM table WHERE id = ?`) which are extremely important to prevent SQL injections.

PS: this is an alpha component and doesn't have certification tests. The feature is being validated in the integration tests (which don't run in the CI but pass locally)

### DIFF
--- a/bindings/mysql/mysql.go
+++ b/bindings/mysql/mysql.go
@@ -271,8 +271,9 @@ func initDB(url, pemPath string) (*sql.DB, error) {
 	}
 
 	if pemPath != "" {
+		var pem []byte
 		rootCertPool := x509.NewCertPool()
-		pem, err := os.ReadFile(pemPath)
+		pem, err = os.ReadFile(pemPath)
 		if err != nil {
 			return nil, fmt.Errorf("error reading PEM file from %s: %w", pemPath, err)
 		}
--- a/bindings/mysql/mysql.go
+++ b/bindings/mysql/mysql.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"reflect"
 	"strconv"
+	"sync/atomic"
 	"time"
 
 	"github.com/go-sql-driver/mysql"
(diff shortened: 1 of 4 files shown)


bindings/mysql/mysql.go
```diff
@@ -87,21 +90,22 @@ type mysqlMetadata struct {
 
 	// ConnMaxIdleTime is the maximum amount of time a connection may be idle.
 	ConnMaxIdleTime time.Duration `mapstructure:"connMaxIdleTime"`
-
-	// MaxRetries is the maximum number of retries for a query.
-	MaxRetries int `mapstructure:"maxRetries"`
```
ItalyPaleAle, line 92: This metadata property was completely unused and was removed

bindings/mysql/mysql.go
```diff
@@ -254,17 +282,25 @@ func initDB(url, pemPath string) (*sql.DB, error) {
...
+		})
 		if err != nil {
 			return nil, fmt.Errorf("error register TLS config: %w", err)
 		}
 	}
 
-	db, err := sql.Open("mysql", url)
+	// Required to correctly parse time columns
+	// See: https://stackoverflow.com/a/45040724
+	conf.ParseTime = true
```
ItalyPaleAle, line 296: Integration tests were failing without this

bindings/mysql/mysql.go
```diff
@@ -112,23 +116,29 @@ func (m *Mysql) Init(ctx context.Context, md bindings.Metadata) error {
...
+	m.db, err = initDB(meta.URL, meta.PemPath)
 	if err != nil {
 		return err
 	}
 
-	db.SetMaxIdleConns(meta.MaxIdleConns)
-	db.SetMaxOpenConns(meta.MaxOpenConns)
-	db.SetConnMaxIdleTime(meta.ConnMaxIdleTime)
-	db.SetConnMaxLifetime(meta.ConnMaxLifetime)
+	if meta.MaxIdleConns > 0 {
```
berndverst, line 124: possibly a behavior change here right? Not sure what setting the value to 0 (for unused metadata) would have done in all of those cases.

berndverst: The behavior now is correct - but unclear if something is changing for users or not.

berndverst (approved):

bindings/mysql/mysql.go
ItalyPaleAle: From the APIs:

image

image

image

PR #2977, opened 2023-07-10
Some linting in metadata.yaml files

ItalyPaleAle:
Some linting in metadata.yaml files

Also removed a superfluous log statement in Cosmos DB binding that could have logged PII.

### DIFF
--- a/bindings/azure/eventhubs/metadata.yaml
+++ b/bindings/azure/eventhubs/metadata.yaml
@@ -13,7 +13,7 @@ binding:
   output: true
   operations:
     - name: create
-      description: "Create an event subscription."
+      description: "Create an event subscription"
 capabilities: []
 authenticationProfiles:
   - title: "Connection string"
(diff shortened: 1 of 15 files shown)


bindings/azure/eventhubs/metadata.yaml
```diff
@@ -62,7 +66,7 @@ builtinAuthenticationProfiles:
       - name: subscriptionId
         type: string
         required: false
-        bindings:
+        binding:
```
daixiang0, line 69: Does it means the current version not work as expected? Do we have a way to verify metadata yaml files?

ItalyPaleAle: > Does it means the current version not work as expected?

Just the metadata.yaml, may not be 100% accurate. Doesn't alter the component's code

> Do we have a way to verify metadata yaml files?

Yes but it cannot catch these kinds of issues I'm afraid.

However, lines like [these](https://github.com/dapr/components-contrib/pull/2977/files#diff-959c9727aac0303712c78df4a173b367dd191017531b8bf449b60e4c8aed8f1aR1) make it so the language server in the editor (i.e. VS Code) can highlight these errors. That's how I caught these.

daixiang0 (approved):

PR #2958, opened 2023-06-30
[WIP][Component Metadata Schema] Adds Common/Shared Authentication Profile to Component Metadata Schema

robertojrojas:
# Description

[WIP][Component Metadata Schema] Adds Common/Shared Authentication Profile to Component Metadata Schema.

## Issue reference

 this PR will close: #2935

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #2984, opened 2023-07-11
[Metadata] Update validator and some other fixes

ItalyPaleAle:
This PR updates the metadata validator build tool:

1. `GetComponentMetadata` now returns a `MetadataMap` (a map of structs) so we can include more information about each field.
   - Required updating each component - that's the reason why this PR is so large
2. `GetComponentMetadata` was removed from the interface of the components (like state.Store). It's still required to be implemented, and the metadata checker CI task will fail if not.
3. Some tags have been added to fields:
   - `mdignore` tells the metadata validator to skip the field. This is used to ignore fields that are part of built-in authentication profiles, fixing the failures reported in #2904
   - `mdonly` replaces `only` (same behavior as before)
   - `mddeprecated` marks the field as deprecated (currently this has no impact on the generated metadata)
   - `mdaliases` is used to list aliases (currently this has no impact on the generated metadata)
4. Built-in authentication profiles are read from a YAML document rather than a Go struct. This allows fixing issues such as one introduced with #2958 where the AWS profiles did not support adding more metadata
5. `GetComponentMetadata()` is now included in its own interface which is then added to each component type's interface
6. When generating the metadata bundle, trim whitespaces at the start and end of description fields (before, many `description` fields ended with the string `\n`)
7. Fixed some components' metadata.
    - Most notably, `state.sqlserver` was not exporting any component metadata definition.


artursouza (dismissed):
Thanks for the description of what has been done. Feel free to wait for @berndverst's approval if he wants to review this.

ItalyPaleAle:
Thanks @artursouza . We need to wait for https://github.com/dapr/dapr/issues/6649 to be fixed or we can't pin an updated contrib.

ItalyPaleAle:
@berndverst  you ok with it?

berndverst (changes requested):
Please do not remove the `GetComponentMetadata` method from the interface. Instead update the interface with the new signature.

With your approach you would simply get a compile error if a component failed to provide the GetComponentMetadata method. That will be difficult to understand when we add a new component.

Or please update our unit tests to also do an assertion as ComponentWithMetadata struct so that nobody can ship a component without this method. We want this to fail early - in unit tests. The error should make it clear that the GetComponentMetadata method is missing.

ItalyPaleAle:
@berndverst I hear you.

I have made a change so that the `GetComponentMetadata` method now is required when running the linter. This should allow users to run "make lint" and immediately note if the method is not present. This is in addition to the fact that "make check-component-metadata" continues to fail if the GetComponentMetadata method isn't present.

Please take a look now.

> How this was implemented:
>
> 1. In the original PR I created an interface `metadata.ComponentWithMetadata` which included `GetComponentMetadata`. That interface was used by the metadata analyzer tool only
> 2. I now added the interface to all components' interfaces (like the `state.Store` interface).
> 3. BUT... Thanks to using build tags, if the `metadata` build tag isn't passed, `metadata.ComponentWithMetadata` is empty (so it doesn't require any method). When the linter runs, however, the `metadata` build tag is present so the method is required

One reason for doing this is so if we change the metadata interface (again), it doesn't break the runtime or the certification tests.

berndverst (approved):

PR #2987, opened 2023-07-12
State stores: expose TTL as a feature

ItalyPaleAle:
State stores that support TTLs now expose the TTL feature. This can be returned by the runtime when retrieving the metadata.

Updated conformance tests to ensure the feature is present when the "ttl" test is enabled, or missing otherwise (with the exception of CF Workers KV, for which TTLs are not tested in the conformance tests).


ItalyPaleAle:
Moving to draft until https://github.com/dapr/components-contrib/issues/2988 is fixed

PR #2989, opened 2023-07-13
Update a bunch of dependencies

ItalyPaleAle:
Updates a bunch of deps (not all - some require too many code changes).

This PR was opened from upstream so no need to run /ok-to-test


ItalyPaleAle:
Will move this to draft until dapr/dapr#6659 is merged. That updates an OTEL dependency which is used in our cert tests.

ItalyPaleAle:
Pinning an update runtime isn't currently working due to changes introduced in dapr/dapr#6382 and dapr/dapr#6584

PR #2992, opened 2023-07-16
Fix redefine flag

filintod:
# Description

This PR add something similar to https://github.com/kubernetes-sigs/controller-runtime/pull/1999 to avoid redefined kubeconfig flag issues when using this library plus others that might already have kubeconfig as a flag.

## Issue reference

Please reference the issue this PR will close: #2993 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


ItalyPaleAle (commented):
Although this PR LGTM, I was not aware of this code and this behavior. It does seem odd that we're defining CLI flags in components-contrib. This seems anti-pattern?

filintod:
It does seem strange maybe to ease tests but don't know the original use case

daixiang0:
> It does seem strange maybe to ease tests but don't know the original use case

Do we really register it in the CLI?

ItalyPaleAle:
Looks like it is getting registered. Running `daprd --help`:

```
  -kubeconfig string
    	(optional) absolute path to the kubeconfig file (default "/Users/alessandro/.kube/config")
```

I think this needs to change. We should allow setting the kubeconfig file using the metadata for each component.

@filintod can you change that please?

ItalyPaleAle (changes requested):
Let's refactor this to avoid adding CLI flags to daprd

filintod:
I can follow as how it is done in dapr https://github.com/dapr/dapr/blob/master/utils/utils.go#L50

with a custom kubeconfig env var `KUBE_CONFIG`

wdty?

ItalyPaleAle:
@filintod that method looks like is being used by Dapr control plane services only (operator, injector, sentry), and not by daprd.

I would prefer the Kubeconfig path to be a metadata option in all Kubernetes components. This is the pattern we use to configure all components.

ItalyPaleAle (approved):

filintod:
thanks, sorry I forgot about this.

ItalyPaleAle:
> thanks, sorry I forgot about this.

No worries! It became a bit more of a pressure because of dapr/dapr#6766

Do the changes look good to you? Mind doing a quick review?

I plan on doing the 2nd part, which is allowing setting kubeconfig path in components' metadata, in a separate PR.

ItalyPaleAle:
See #3060 for part 2

PR #3004, opened 2023-07-21
Azure App Config: fix time format to time.duration

### LINKED ISSUE #2976: Azure App Configuration component: use Go durations instead of nanoseconds for metadata options
The Azure App Configuration component has [4 metadata options](https://github.com/dapr/components-contrib/blob/6d6cb2af0a3a48b256c87a4f8412be3f4d38a442/configuration/azure/appconfig/metadata.go#L22-L25) that are durations, and those are defined as integers indicating nanoseconds (1 second =1_000_000_000 nanoseconds). This is a sub-par experience for end users.

The component should be updated to use `time.Duration` for those fields. The metadata parser automatically parses that as Go durations (e.g. "5s", or "2m"), which is a much better experience for users, and more aligned with the rest of Dapr.

This is a backwards-incompatible change that can be done to this alpha component.

pravinpushkar:
# Description

Changes nanosecond to time.Duration format

## Issue reference

Fixes #2976


configuration/azure/appconfig/appconfig.go
```diff
@@ -144,17 +144,17 @@ func parseMetadata(meta configuration.Metadata) (metadata, error) {
...
+		m.internalMaxRetryDelay = m.MaxRetryDelay
 	}
-	if m.RetryDelay != nil {
-		m.internalRetryDelay = time.Duration(*m.RetryDelay)
+	if m.RetryDelay != 0 {
+		m.internalRetryDelay = m.RetryDelay
 	}
-	if m.SubscribePollInterval != nil {
-		m.internalSubscribePollInterval = time.Duration(*m.SubscribePollInterval)
+	if m.SubscribePollInterval != 0 {
```
ItalyPaleAle, line 153:
```suggestion
	if m.SubscribePollInterval > 0 {
```

configuration/azure/appconfig/appconfig.go
```diff
@@ -144,17 +144,17 @@ func parseMetadata(meta configuration.Metadata) (metadata, error) {
...
+		m.internalSubscribePollInterval = m.SubscribePollInterval
 	}
-	if m.RequestTimeout != nil {
-		m.internalRequestTimeout = time.Duration(*m.RequestTimeout)
+	if m.RequestTimeout != 0 {
```
ItalyPaleAle, line 156:
```suggestion
	if m.RequestTimeout > 0 {
```

configuration/azure/appconfig/metadata.yaml
```diff
@@ -35,22 +35,22 @@ metadata:
...
+    default: '4s'
+    example: '5s'
   - name: maxRetryDelay
-    description: "Specifies the maximum delay allowed before retrying an operation, in nanoseconds. Typically the value is greater than or equal to the value specified in RetryDelay. Defaults to 120 seconds. -1 disables the limit."
-    type: number
-    default: '120000000000'
-    example: '180000000000'
+    description: "Specifies the maximum delay allowed before retrying an operation. Typically the value is greater than or equal to the value specified in RetryDelay. Defaults to 120 seconds."
+    type: duration
+    default: '120s'
```
ItalyPaleAle, line 45:
```suggestion
    description: "Specifies the maximum delay allowed before retrying an operation. Typically the value is greater than or equal to the value specified in RetryDelay. Defaults to 2 minutes."
    type: duration
    default: '2m'
```

ItalyPaleAle (changes requested):
Please add a “//nolint:goconst” at the beginning of appconfig_test.go

ItalyPaleAle:
Thanks for this PR @pravinpushkar !

berndverst:
- Why not refactor all configuration stores in the same PR since you know the problem is the exact same in the other components? :)

- Please note that the property still accepts pure integers - an integer without unit is interpreted as seconds. https://github.com/dapr/components-contrib/blob/master/metadata/duration.go#L69

berndverst (dismissed):

ItalyPaleAle:
I've pushed some changes to avoid duplicating properties with "external" and "internal" ones.

configuration/azure/appconfig/metadata.go
```diff
@@ -13,19 +13,63 @@ limitations under the License.
...
+	}
+	if m.RetryDelay > time.Duration(10^6)*time.Second {
+		log.Warnf("[WARN] Property 'retryDelay' is %v, which is probably incorrect. If you are upgrading from Dapr 1.11, please note that the property is now a Go duration rather than a number of nanoseconds", m.RetryDelay)
+	}
+	if m.SubscribePollInterval > time.Duration(10^6)*time.Second {
+		log.Warnf("[WARN] Property 'subscribePollInterval' is %v, which is probably incorrect. If you are upgrading from Dapr 1.11, please note that the property is now a Go duration rather than a number of nanoseconds", m.SubscribePollInterval)
+	}
+	if m.RequestTimeout > time.Duration(10^6)*time.Second {
+		log.Warnf("[WARN] Property 'requestTimeout' is %v, which is probably incorrect. If you are upgrading from Dapr 1.11, please note that the property is now a Go duration rather than a number of nanoseconds", m.RequestTimeout)
+	}
```
ItalyPaleAle, line 72: Thoughts on these @berndverst ? To make it better for users upgrading?

Feel free to remove if you don't like this

pravinpushkar: @ItalyPaleAle Where does things like `m.RetryDelay` are getting set to passed in values from component.yaml?

Secondly, If I do not pass any value then these will be set to default values like - m.MaxRetryDelay will be `2m0s` which is greater than `time.Duration(10^6)*time.Second`. Which does not seem right.

logs - 
```
WARN[0000] [WARN] Property 'maxRetryDelay' is 2m0s, which is probably incorrect. If you are upgrading from Dapr 1.11, please note that the property is now a Go duration rather than a number of nanoseconds  app_id=test component="pravin-cilium-app-config (configuration.azure.appconfig/v1)" instance=Pravins-MacBook-Pro.local scope=dapr.contrib type=log ver=edge
```

ItalyPaleAle: Duh, rookie mistake. `^` is a bitwise operator 😅 I'll fix it

> Where does things like m.RetryDelay are getting set to passed in values from component.yaml

When we call `DecodeMetadata` on the struct, they are populated. The method supports `time.Duration` automatically

ItalyPaleAle (approved):

Issue #2952, opened 2023-06-29
Implement Close() on Consul state store


artursouza:
## Describe the feature
Implement Close() on Consul state store

## Release Note

RELEASE NOTE:

PR #3020, opened 2023-07-30
Add closer to consul state

filintod:
# Description

Adds a consul closer to shutdown idle connections. Injected default transport into struct to be able to access the transport methods.

## Issue reference

Please reference the issue this PR will close: #2952 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -68,6 +68,7 @@ require (
 	github.com/grandcat/zeroconf v1.0.0
 	github.com/hamba/avro/v2 v2.13.0
 	github.com/hashicorp/consul/api v1.22.0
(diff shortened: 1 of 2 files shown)


ItalyPaleAle (changes requested):
Since this component is stateless and doesn't maintain any persistent connection, I am not sure a closer is needed.

I see the closer that was implemented is only concerned with closing potentially idle TCP sockets... Not sure there's much value in doing that. They'd be closed after a few seconds automatically anyways, and they don't really cause conflicts if the component is destroyed.

filintod:
tested that indeed the time is not even in the seconds range, so would not affect restarts in any meaningful way. will close

PR #3023, opened 2023-07-31
Adds OIDC authentication support for pubsub Apache Pulsar

### LINKED ISSUE #2591: Implement OIDC Authentication against Keycloak for Apache Pulsar
## Describe the feature

OIDC provides a more flexible, modern and standard-based approach to authentication. 
Currently, the Apache Pulsar PubSub [component](https://github.com/dapr/components-contrib/blob/master/pubsub/pulsar/pulsar.go) handles authentication by reading a [Pulsar-generated token](https://pulsar.apache.org/docs/2.11.x/security-jwt/#generate-tokens) defined in a PubSub Component [manifest](https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-pulsar/).

This request is to add an OIDC [Client Credentials grant](https://developers.onelogin.com/openid-connect/api/client-credentials-grant) authentication flow to the Apache Pulsar PubSub component. 

References:
[Keycloak - Identity and Access Management for Modern Applications - Packt Publishing](https://www.packtpub.com/product/keycloak-identity-and-access-management-for-modern-applications/9781800562493) 
[Go authentication example program from book above](https://github.com/PacktPublishing/Keycloak-Identity-and-Access-Management-for-Modern-Applications/tree/master/ch7/golang)

## Release Note

RELEASE NOTE:

JoshVanL:
Closes #2591 

PR adds support for OIDC authentication to the pubsub pulsar. This is configured via the following metadata fields:

```yaml
- oidcTokenCAPEM: CA PEM certificate bundle to connect to the OIDC issuer. If not defined, the system pool will be used.
- oidcTokenURL: URL to request the OIDC client_credentials token from.
- oidcClientID: token client ID.
- oidcClientSecret: token client secret.
- oidcAudiences: comma separated list of audiences to request for. Must not be empty.
- oidcScopes: comma separated list of scopes to request. If empty, defaults to `"openid"`. If defined, `"openid"` must be present.
```

Another field `authType` has been added which accepts `none`, `token`, and `oidc`. Defaults to `none`, however if undefined and the `token` metadata field has been defined, then this defaults to `token` to maintain backwards compatibility.

---

Adds a certification test which runs the current pulsar suite, but with OIDC authentication enabled. Uses pulsar v3 since that is the minimum version with OIDC authentication support. Runs a mock OIDC issuer server during the test. We template the CA PEM bundle, based on the generated self signed certificate from the mock server at test runtime.

---

PR doesn't move the kafka OIDC implementation to use the shared internal package since kafka OIDC auth isn't tested. Once a certificate test is written, we can move kafka to use the internal package.

Docs PR: https://github.com/dapr/docs/pull/3655

---

Question: Because the pulsar certification tests now take some time with running the suite in both non and oidc auth modes- do we want to split them into two separate workflows?

/cc @yaron2


internal/component/kafka/consumer.go
```diff
@@ -284,7 +284,21 @@ func (k *Kafka) Subscribe(ctx context.Context) error {
 		running: make(chan struct{}),
 	}
 
+	ctx, cancel := context.WithCancel(ctx)
```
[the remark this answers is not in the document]
yaron2: Thanks!

internal/authentication/oidc/clientcredentials.go
```diff
@@ -0,0 +1,209 @@
...
+		for {
+			select {
+			case <-c.closeCh:
+				return
+			case <-ctx.Done():
+				return
+			case <-c.clock.After(renewDuration):
+			}
+
+			c.log.Info("Renewing client credentials token")
```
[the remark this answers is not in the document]
JoshVanL: This is going to print every 1-24hrs but yeah we can make this a Debug

internal/authentication/oidc/clientcredentials.go
```diff
@@ -0,0 +1,209 @@
...
+		}
+	}()
+}
+
+func toConfig(opts ClientCredentialsOptions) (*ccreds.Config, *http.Client, error) {
+	scopes := opts.Scopes
+	if len(scopes) == 0 {
+		// If no scopes are provided, then the default is to use the 'openid' scope
+		// since that is always required for OIDC so implicitly add it.
+		scopes = []string{"openid"}
```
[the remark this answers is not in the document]
JoshVanL: Done

yaron2 (approved):

internal/authentication/oidc/clientcredentials.go
```diff
@@ -0,0 +1,213 @@
...
+			c.wg.Done()
+		}()
+
+		for {
+			select {
+			case <-c.closeCh:
+				return
+			case <-ctx.Done():
+				return
+			case <-c.clock.After(renewDuration):
```
ItalyPaleAle, line 104: This is anti-pattern for using OAuth and we should not refresh tokens in background. Instead, tokens should be fetched lazily and refreshed only if necessary. This is especially true when the consumer of the tokens is something like Pulsar where connections are long-lived so tokens don't need to be refreshed constantly.

internal/authentication/oidc/clientcredentials.go
```diff
@@ -0,0 +1,213 @@
...
+	if c.closed.CompareAndSwap(false, true) {
+		close(c.closeCh)
+	}
+}
+
+func (c *ClientCredentials) Token() (string, error) {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	if c.closed.Load() {
```
ItalyPaleAle, line 196: Here is where you check if the token that's in cache is valid, and if not you lazily request a new one.

Since multiple goroutines could be calling `Token()` at the same time, I recommend implementing some synchronization method (or use the "promise" library!)

internal/authentication/oidc/clientcredentials.go
```diff
@@ -0,0 +1,213 @@
...
+
+	return c.currentToken.AccessToken, nil
+}
+
+// tokenRenewTime returns the duration when the token should be renewed, which is
+// half of the token's lifetime.
+func (c *ClientCredentials) tokenRenewDuration() time.Duration {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	return c.currentToken.Expiry.Sub(c.clock.Now()) / 2
```
ItalyPaleAle, line 212: No need to halve the time, that's wasteful. Maybe subtract 1 min, but since tokens are used right away there's no need to halve this.

pubsub/pulsar/metadata.go
```diff
@@ -26,12 +26,20 @@ type pulsarMetadata struct {
...
 	Namespace               string                    `mapstructure:"namespace"`
 	Persistent              bool                      `mapstructure:"persistent"`
-	Token                   string                    `mapstructure:"token"`
 	RedeliveryDelay         time.Duration             `mapstructure:"redeliveryDelay"`
 	internalTopicSchemas    map[string]schemaMetadata `mapstructure:"-"`
 	PublicKey               string                    `mapstructure:"publicKey"`
 	PrivateKey              string                    `mapstructure:"privateKey"`
 	Keys                    string                    `mapstructure:"keys"`
+
+	AuthType         string   `mapstructure:"authType"`
```
ItalyPaleAle, line 35: Generally when a component has multiple auth mechanisms, we do not force users to specify that unless that's strictly necessary.

In this case, it is not necessary.

When the component is initialized, you can check if `token` is present and assume auth type is a shared token. Otherwise, check if the OIDC credentials are present.
If none is present, return an error.

This is more aligned with what we do for almost all other components, and saves users from having another metadata option.

pubsub/pulsar/pulsar.go
```diff
@@ -173,9 +178,37 @@ func (p *Pulsar) Init(_ context.Context, metadata pubsub.Metadata) error {
...
+			Logger:       p.logger,
+			TokenURL:     m.OIDCTokenURL,
+			CAPEM:        []byte(m.OIDCTokenCAPEM),
+			ClientID:     m.OIDCClientID,
+			ClientSecret: m.OIDCClientSecret,
+			Scopes:       m.OIDCScopes,
+			Audiences:    m.OIDCAudiences,
+		})
+		if err != nil {
+			return fmt.Errorf("could not instantiate oidc token provider: %v", err)
```
ItalyPaleAle, line 204: Total nit:


```suggestion
			return fmt.Errorf("could not instantiate oidc token provider: %w", err)
```

ItalyPaleAle (changes requested):
This PR should have not been merged this quickly…

ItalyPaleAle:
@JoshVanL please re-open this PR after addressing the feedback above, thanks

pubsub/pulsar/metadata.go
yaron2: What makes this component different from how we handle auth on Kafka for example? If we don't use auth explicitly there may be additional token auth mechanisms in the future that will conflict.

ItalyPaleAle: I am not sure about Kafka, but for example all Azure components automatically select shared token auth vs AzureAD depending on the metadata that's passed.

In the rare case of conflicts we have added one-off metadata keys such as `useAzureAD` for Azure SQL

internal/authentication/oidc/clientcredentials.go
JoshVanL: Why is this an anti-pattern? I have never heard of this. Fetching tokens lazily is strictly worse from the dapr perspective, increasing latency for the pubsub app which is particularly acute for publishing. I don't think we should be concerned with reducing load on the OIDC provider- signing JWTs is not resource intensive and is basically free for all intense and purposes. Regardless of long lived etc. we can assume that lazily requesting tokens will produce about roughly the same amount of requests as renewing in the background for a publishing app (bar the renewal time we agree on), however fetching tokens lazily will increase the latency on the unlucky publish(es).

internal/authentication/oidc/clientcredentials.go
JoshVanL: How about 2/3rds? I think 1 minute is far too short- node migrations in kube can take longer than a minute for example.

pubsub/pulsar/metadata.go
JoshVanL: My fear with not having an explicit auth field is that we end up with "magic config" schemas, which are both confusing and surprising to users. It wouldn't be clear to me what the behavior would be if I defined multiple auth types for example.

This is also particularly important for security folks who are responsible for auditing or writing policy over config. A clear auth type field gives them the confidence that things are setup correctly and the software is going to do what they expect.

internal/authentication/oidc/clientcredentials.go
ItalyPaleAle: Most IdP's I'm familiar with issue tokens with a lifetime of 1h by default. Cutting 30 or 20 mins out of that seems unnecessary.

5 mins?

Why are Kubernetes node migrations a concern here?

ItalyPaleAle: > It's very common to be running an internal provider in (or in another) cluster. A node migration is an example scenario where the issuer server becomes unavailable for a non-trivial amount of time, however the system should be able to cope with that as best it can.

That is an operational thing that should not be a concern of Dapr's. If an ops person runs their own IdP, they are responsible for ensuring its availability.

internal/authentication/oidc/clientcredentials.go
JoshVanL: > So refreshing automatically "just in case" seems really wasteful.

Fair enough- I still disagree with this statement though. I maintain that the latency improvements for daprd fair out way the small resource used to fetch more tokens than perhaps needed.

> It may help hitting rate-limits. It also can create additional noise in the audit logs that are sent to SIEMs, etc.

An IdP which rate limits clients for requesting tokens at half expiry is miss configured imo, and audit logs are built around the concept of clients renewing their identity documents. They will be indexed and filtered around this exact scenario.

internal/authentication/oidc/clientcredentials.go
JoshVanL: > That is an operational thing that should not be a concern of Dapr's. If an ops person runs their own IdP, they are responsible for ensuring its availability.

I agree, but it is standard practice as a client to have enough buffer time in the validity of your identity documents to maximize up time if the provider goes down. Using the document validity duration is a great metric to choose that buffer time, rather than using an arbitrary static value.

PR #3026, opened 2023-08-01
Pubsub pulsar authentication ~OIDC~ OAuth2

JoshVanL:
In an effort to ensure OIDC authentication support is added to pubsub pulsar, I have spun out https://github.com/dapr/components-contrib/pull/3023 with the background refresh routine and `authType` field removed. I am still keen for these two to be added back though! :slightly_smiling_face: 

/cc @ItalyPaleAle  @yaron2


internal/authentication/oidc/clientcredentials.go
```diff
@@ -0,0 +1,172 @@
...
+
+	return &ClientCredentials{
+		log:          opts.Logger,
+		currentToken: token,
+		httpClient:   httpClient,
+		fetchTokenFn: conf.Token,
+	}, nil
+}
+
+func toConfig(opts ClientCredentialsOptions) (*ccreds.Config, *http.Client, error) {
```
ItalyPaleAle, line 79: Nit: make this a method of ClientCredentialsOptions.

internal/authentication/oidc/clientcredentials.go
```diff
@@ -0,0 +1,172 @@
...
+	if c.currentToken.Valid() {
+		return nil
+	}
+
+	token, err := c.fetchTokenFn(context.WithValue(ctx, oauth2.HTTPClient, c.httpClient))
+	if err != nil {
+		return err
+	}
+
+	c.currentToken = token
```
ItalyPaleAle, line 170: Should check that `token.Valid` before saving it? As a sanity check

internal/authentication/oidc/clientcredentials.go
```diff
@@ -0,0 +1,172 @@
...
+	ccreds "golang.org/x/oauth2/clientcredentials"
+
+	"github.com/dapr/kit/logger"
+)
+
+const (
+	oidcScopeOpenID = "openid"
+)
+
+type ClientCredentialsOptions struct {
```
ItalyPaleAle, line 37: Can you add the `mapstructure` properties to this? So then components' metadata structs can inherit this, and we can help standardizing the metadata options should other components adopt OAuth2 authentication in the future.

internal/authentication/oidc/clientcredentials.go
```diff
@@ -0,0 +1,172 @@
...
+type ClientCredentials struct {
+	log          logger.Logger
+	currentToken *oauth2.Token
+	httpClient   *http.Client
+	fetchTokenFn func(context.Context) (*oauth2.Token, error)
+
+	lock sync.RWMutex
+}
+
+func NewClientCredentials(ctx context.Context, opts ClientCredentialsOptions) (*ClientCredentials, error) {
```
ItalyPaleAle, line 58: Thinking about it, is this actually OIDC? Seems like we're just using OAuth2 here, as we are not concerned with the claims that are returned?

internal/authentication/oidc/clientcredentials.go
```diff
@@ -0,0 +1,172 @@
...
+	"time"
+
+	"golang.org/x/oauth2"
+	ccreds "golang.org/x/oauth2/clientcredentials"
+
+	"github.com/dapr/kit/logger"
+)
+
+const (
+	oidcScopeOpenID = "openid"
```
ItalyPaleAle, line 34: The component should probably pass the scope it wants to use. Other components may have different scopes.

internal/authentication/oidc/clientcredentials.go
```diff
@@ -0,0 +1,172 @@
...
+)
+
+type ClientCredentialsOptions struct {
+	Logger       logger.Logger
+	TokenURL     string
```
berndverst, line 39: Expanding on what @ItalyPaleAle said - here's an example:

```suggestion
	TokenURL     string `mapstructure:"tokenURL"`
```

These spelling (case insensitive) as indicated by the mapstructure struct tag is the spelling which users would have to use in metadata.

berndverst (commented):
Additional clarification ☝️

internal/authentication/oidc/clientcredentials.go
JoshVanL: :+1:  Can't hurt :slightly_smiling_face:

internal/authentication/oidc/clientcredentials.go
JoshVanL: I'm actually very against this. I'm quite keen that we have a distinct and clean separation between external and internal APIs. Happy to create a shared metadata API package for this reason, but I don't think we should have external API definitions in this internal implementation package.

internal/authentication/oidc/clientcredentials.go
JoshVanL: Happy to add claim validation on the returned token if you think that makes sense?

internal/authentication/oidc/clientcredentials.go
JoshVanL: `openid` is always required for OIDC, but other scopes can be added with the `Scopes` option.

internal/authentication/oidc/clientcredentials.go
JoshVanL: Thanks @berndverst, please see my comment above

internal/authentication/oidc/clientcredentials.go
ItalyPaleAle: The problem that we've seen is that then components implement this using different metadata keys. Before we standardized the Azure AD metadata, some components used "azureClientID", some "spnClientID", some "clientID", for the exact same property.

If you'd rather have a separate struct, that's fine, but I'd still like to have a shared one in this package.

internal/authentication/oidc/clientcredentials.go
ItalyPaleAle: I don't think it's necessary TBH. The service will perform the validation when the token is presented, so us doing additional validation doesn't seem necessary here...

I am just thinking that maybe this package should be called "oauth2" rather than "oidc" since nothing here uses the OIDC standard, but just OAuth2.

internal/authentication/oidc/clientcredentials.go
ItalyPaleAle: But see my other comment. This package is for OAuth2 after all, and not OIDC. If a component wants the "openid" scope, it can request it.

internal/authentication/oidc/clientcredentials.go
JoshVanL: > I am just thinking that maybe this package should be called "oauth2" rather than "oidc" since nothing here uses the OIDC standard, but just OAuth2.

Why do you think this is the case?

ItalyPaleAle: I don't see anything OIDC-specific in here. We are using the OAuth2 protocol to request an access token. We are not validating the claims contained in the token, so it doesn't need to be OIDC.

Pulsar wants token requested with the "openid" scope, but aside from that this should work as-is for all OAuth2 client credential flows?

Unless I'm missing something

JoshVanL: I think this is OIDC because we are fetching an identity token document, requested with the `openid` scope, which we are then using to _authenticate_ with a service on behalf of the "user". Remember that OIDC is built on top of OAuth2, and basically a constrained subset.

I think it is quite important to keep the distinction that we are using OIDC here- we are proving an identity to authenticate with the service, not given proof of authorization to access data. I'll add some validation to ensure the returned token has the right claims to be conformant.

ItalyPaleAle: Alright, if we need OAuth2 we can always split it out in the future, creating a sub-class of this.

internal/authentication/oidc/clientcredentials.go
JoshVanL: Done [here](https://github.com/dapr/components-contrib/pull/3026/commits/b817c7c164f621d8b208c3e2874b71f15be31a8f)

internal/authentication/oidc/clientcredentials.go
```diff
@@ -34,6 +34,17 @@ const (
...
 
+// ClientCredentialsMetadata is the metadata fields which can be used by a
+// component to configure an OIDC client_credentials token source.
+type ClientCredentialsMetadata struct {
+	TokenCAPEM   string   `mapstructure:"oidcTokenCAPEM"`
+	TokenURL     string   `mapstructure:"oidcTokenURL"`
+	ClientID     string   `mapstructure:"oidcClientID"`
+	ClientSecret string   `mapstructure:"oidcClientSecret"`
+	Audiences    []string `mapstructure:"oidcAudiences"`
+	Scopes       []string `mapstructure:"oidcScopes"`
```
ItalyPaleAle, line 45: These are OAuth2 credentials not OIDC (In fact, it's called "Client Credentials" which is an OAuth2 flow :) ). Since these names are going to be used for OAuth2 (without OIDC) they should be generic.

Unsure if `oauth` or `oauth2` is better.

```suggestion
	TokenCAPEM   string   `mapstructure:"oauth2TokenCAPEM"`
	TokenURL     string   `mapstructure:"oauthT2okenURL"`
	ClientID     string   `mapstructure:"oauth2ClientID"`
	ClientSecret string   `mapstructure:"oauth2ClientSecret"`
	Audiences    []string `mapstructure:"oauth2Audiences"`
	Scopes       []string `mapstructure:"oauth2Scopes"`
```

JoshVanL: I've gone ahead and renamed it all to be OAuth2 and removed all OIDC compliance checks. It is now simply a generic client_credentials OAuth2 authenticator. The docs will need to be written so that the user is aware that they are responsible for making sure the authenticater is OIDC compliant. The certification tests have been kept basically the same since pulsar is expecting an OIDC provider be available and OIDC client tokens.

IMO, I think it is a mistake to not make this an ODIC authenticator but am keen to unblock users who want to use OIDC authentication in pulsar.

pubsub/pulsar/pulsar.go
```diff
@@ -173,9 +173,28 @@ func (p *Pulsar) Init(_ context.Context, metadata pubsub.Metadata) error {
 		ConnectionTimeout:          30 * time.Second,
 		TLSAllowInsecureConnection: !m.EnableTLS,
 	}
-	if m.Token != "" {
+
+	switch {
+	case len(m.Token) > 0:
```
ItalyPaleAle, line 178: I didn't notice that token was optional. I think in this case it makes even more sense to do the auth this way and not having an explicit "authType"?

pubsub/pulsar/pulsar.go
```diff
@@ -173,9 +173,28 @@ func (p *Pulsar) Init(_ context.Context, metadata pubsub.Metadata) error {
...
+			Logger:       p.logger,
+			TokenURL:     m.ClientCredentialsMetadata.TokenURL,
+			CAPEM:        []byte(m.ClientCredentialsMetadata.TokenCAPEM),
+			ClientID:     m.ClientCredentialsMetadata.ClientID,
+			ClientSecret: m.ClientCredentialsMetadata.ClientSecret,
+			Scopes:       m.ClientCredentialsMetadata.Scopes,
+			Audiences:    m.ClientCredentialsMetadata.Audiences,
+		})
+		if err != nil {
+			return fmt.Errorf("could not instantiate oauth2 token provider: %v", err)
```
ItalyPaleAle, line 192: Nit:
```suggestion
			return fmt.Errorf("could not instantiate oauth2 token provider: %w", err)
```

ItalyPaleAle (approved):
LGTM

@berndverst anything else from you?

berndverst (approved):

berndverst:
> LGTM
> 
> @berndverst anything else from you?

All good. approved.

PR #3027, opened 2023-08-01
Pubsub: pulsar OIDC authentication with `authType` metadata field & background token refresh

JoshVanL:
Branched from https://github.com/dapr/components-contrib/pull/3026

This reverts the changes back to the original PR https://github.com/dapr/components-contrib/pull/3023 to include the `authType` metadata field, as well as renewing the token in a background routine.


ItalyPaleAle:
Marking as draft until #3026 is ready.

pubsub/pulsar/metadata.go
```diff
@@ -26,12 +26,20 @@ type pulsarMetadata struct {
...
 	Namespace               string                    `mapstructure:"namespace"`
 	Persistent              bool                      `mapstructure:"persistent"`
-	Token                   string                    `mapstructure:"token"`
 	RedeliveryDelay         time.Duration             `mapstructure:"redeliveryDelay"`
 	internalTopicSchemas    map[string]schemaMetadata `mapstructure:"-"`
 	PublicKey               string                    `mapstructure:"publicKey"`
 	PrivateKey              string                    `mapstructure:"privateKey"`
 	Keys                    string                    `mapstructure:"keys"`
+
+	AuthType         string   `mapstructure:"authType"`
```
ItalyPaleAle, line 35: Although I'm open to adding the (optional) "authType" metadata back, I do not support the other change of refreshing tokens in background when not needed. Since clients are long-lived and cached, the perf benefit is negligible. At the same time, refreshing tokens in background when we are not going to use them is wasteful (requires another goroutine, etc) and not polite.

ItalyPaleAle (changes requested):

PR #3031, opened 2023-08-02
use rabbitmq message's header values as metadata values in the binding

### LINKED ISSUE #3030: [RabbitMQ] create metadata from message's header values
## Describe the feature
With the [PR](https://github.com/dapr/components-contrib/pull/2895) the ability has been added to create message header values from the passed in metadata values.
Now it's about to return these header values as metadata to the caller.

## Release Note
RELEASE NOTE: **ADD** [RabbitMQ] Fill binding's metadata with message's header values.

robert-oh:
Fixes #3030

### DIFF
--- a/bindings/rabbitmq/rabbitmq.go
+++ b/bindings/rabbitmq/rabbitmq.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"net/url"
 	"reflect"
 	"strconv"
 	"sync"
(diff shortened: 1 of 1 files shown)


bindings/rabbitmq/rabbitmq.go
```diff
@@ -464,8 +465,20 @@ func (r *RabbitMQ) handleMessage(ctx context.Context, handler bindings.Handler,
 				r.logger.Info("Input binding channel closed")
 				return
 			}
+
+			metadata := make(map[string]string)
```
ItalyPaleAle, line 469:
```suggestion
			metadata := make(map[string]string, len(d.Headers))
```

Small memory optimization

ItalyPaleAle (approved):

ItalyPaleAle:
Thanks for this PR!

bindings/rabbitmq/rabbitmq.go
robert-oh: Thanks for already applying the improvement.

PR #3039, opened 2023-08-03
Fixed flaky cert test for ASB Queue binding

ItalyPaleAle:
Also added prefix to all sent messages so tests don't fail if there are multiple concurrent runners (hopefully)


berndverst (approved):

ItalyPaleAle:
ASB Queue Binding cert test passed. Merging.

PR #3040, opened 2023-08-03
HTTP Binding: add "maxResponseBodySize" metadata option

ItalyPaleAle:
Updates the HTTP binding to add the "maxResponseBodySize" metadata option, which limits the amount of data read from response bodies.

This is necessary to prevent OOMs if the server returns too much data.

The default limit is 100 MiB. This _may_ be considered a breaking change if users were using the binding to fetch very large payloads (which may not be a very efficient thing to do anyways). The breaking change is justified by the security aspect of this fix.

The "maxResponseBodySize" metadata is parsed as a resource, using the same library used by K8s. A number without unit is interpreted as bytes (e.g. "100"); units can be added such as "Mi", "M", etc. To be able to support resource quantities, new methods have been added to the metadata parser.


.build-tools/pkg/metadataschema/schema.go
```diff
@@ -86,7 +86,7 @@ type Metadata struct {
 	Sensitive bool `json:"sensitive,omitempty" yaml:"sensitive,omitempty"`
 	// Type of the property.
 	// If this is empty, it's interpreted as "string".
-	Type string `json:"type,omitempty" yaml:"type,omitempty" jsonschema:"enum=string,enum=number,enum=bool,enum=duration"`
+	Type string `json:"type,omitempty" yaml:"type,omitempty" jsonschema:"enum=string,enum=number,enum=bool,enum=duration,enum=resource"`
```
berndverst, line 89: What's that

metadata/utils.go
```diff
@@ -250,16 +251,18 @@ func resolveAliases(md map[string]string, result any) error {
...
 func toTruthyBoolHookFunc() mapstructure.DecodeHookFunc {
+	boolPtrType := reflect.TypeOf(ptr.Of(true))
+
 	return func(
 		f reflect.Type,
 		t reflect.Type,
 		data any,
 	) (any, error) {
-		if f == reflect.TypeOf("") && t == reflect.TypeOf(true) {
+		if f.Kind() == reflect.String && t.Kind() == reflect.Bool {
```
berndverst, line 261: I remember there being a reason why we couldn't use kind. Anyways, make sure this doesn't introduce a bug. Don't refactor that which isn't broken.

ItalyPaleAle: I'll revert this

.build-tools/pkg/metadataschema/schema.go
ItalyPaleAle: It's parsed using the same library as resources in K8s, so you can set "100" for 100 bytes, or "1M" for 1MB.

berndverst (dismissed):

PR #3041, opened 2023-08-04
wasm: fixes test flake on Windows and strict mode

codefromthecrypt:
# Description

Windows clocks often don't give microsecond scope reading reliably. Our "strict mode" test was accidentally verifying a lower bound of time spent sleeping, when the important part is the upper bound. This removes the lower bound, so that tests pass on both unix and windows.

## Issue reference

See https://github.com/dapr/components-contrib/pull/2884#issuecomment-1664498860

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [n/a] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

### DIFF
--- a/internal/wasm/wasm_test.go
+++ b/internal/wasm/wasm_test.go
@@ -172,9 +172,14 @@ func TestNewModuleConfig(t *testing.T) {
 			maxDuration: 50 * time.Millisecond * 5,
 		},
 		{
-			name:        "strictSandbox = true",
-			metadata:    &InitMetadata{StrictSandbox: true, Guest: binStrict},
-			minDuration: 10 * time.Microsecond,
+			name:     "strictSandbox = true",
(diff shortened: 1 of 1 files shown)


ItalyPaleAle (approved):

ItalyPaleAle:
I’m only interested In the windows tests so I’ll stop all other tests

ItalyPaleAle (changes requested):
This doesn't seem to be working yet. It still fails because the delay is exactly 0 so it's never "greater than".

I can't leave a comment there, but you'll need to change L219 to:

```go
assert.GreaterOrEqual(t, duration, tc.minDuration)
assert.LessOrEqual(t, duration, tc.maxDuration)
```

codefromthecrypt:
ah thanks for the keen eye, you are right this expression is wrong.

ItalyPaleAle (approved):

PR #3046, opened 2023-08-05
[WIP] Distributed lock: add Lock (blocking) and RenewLock methods

ItalyPaleAle:
> This PR targets 1.13. It also includes #3044 and #3045

Implements the components-contrib part (and the implementation for the only lock store component, Redis) for dapr/dapr#3549

Adds two methods to the distributed lock components:

- Lock acquires a lock. Unlike TryLock, it blocks until it successfully acquires a lock if someone else owns it (or until the context is canceled)
- RenewLock renews an existing lock, extending its duration

Behavior is validated by conformance tests


sicoyle:
Hi! Just checking in — is this PR still in progress, or should we go ahead and close it out?

PR #3056, opened 2023-08-06
[Error Handling/Codes][State]Starts ETAG MISMATCH Error Handling

robertojrojas:
# Description

Starts State ETAG MISMATCH Error Handling. 
Redis is the first component to implement the Error codes. 

**_NOTE:_** This PR depends on KIT PR: https://github.com/dapr/kit/pull/57

## Issue reference
PR is part of the Dapr Issue: [ETAG MISMATCH](https://github.com/dapr/dapr/issues/6472)

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


state/redis/redis.go
```diff
@@ -117,6 +120,10 @@ func newStateStore(log logger.Logger) *StateStore {
 		json:                           jsoniter.ConfigFastest,
 		logger:                         log,
 		suppressActorStateStoreWarning: atomic.Bool{},
+		resourceInfoData: errorcodes.ResourceInfoData{
+			ResourceType: "state.redis/v1",
+			ResourceName: "REDIS",
```
ItalyPaleAle, line 125: Does this have to be ALL CAPS?

internal/errorcodes/errorcodes.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+var StateETagMismatchReason = Reason("DAPR_STATE_ETAG_MISMATCH")
+
+type ResourceInfoData struct {
+	ResourceType, ResourceName string
+}
+
+func FeatureEnabled(md map[string]string) bool {
+	if _, ok := md[ErrorCodesFeatureMetadataKey]; ok {
```
ItalyPaleAle, line 40: As per comment on other PR, our metadata parsing should ideally be case-insensitive.

internal/errorcodes/errorcodes.go
```diff
@@ -0,0 +1,87 @@
...
+		return true
+	}
+	return false
+}
+
+// NewStatusError returns a Status representing Code, error Reason, Message, and optional ResourceInfo and Metadata.
+// When successful, it returns a StatusError, otherwise returns the original error
+func NewStatusError(code codes.Code, err error, errDescription string, reason Reason, rid *ResourceInfoData, metadata map[string]string) error {
```
ItalyPaleAle, line 48: Why do we need both `err error` and `errDescription string`? Can't the two be inferred from one another?

internal/errorcodes/errorcodes.go
```diff
@@ -0,0 +1,87 @@
...
+
+func FeatureEnabled(md map[string]string) bool {
+	if _, ok := md[ErrorCodesFeatureMetadataKey]; ok {
+		return true
+	}
+	return false
+}
+
+// NewStatusError returns a Status representing Code, error Reason, Message, and optional ResourceInfo and Metadata.
+// When successful, it returns a StatusError, otherwise returns the original error
```
ItalyPaleAle, line 47: Please pass `rid` not as pointer, so it's a bit easier on the GC.

internal/errorcodes/errorcodes.go
```diff
@@ -0,0 +1,87 @@
...
+func NewStatusError(code codes.Code, err error, errDescription string, reason Reason, rid *ResourceInfoData, metadata map[string]string) error {
+	md := metadata
```
ItalyPaleAle, line 49: Why this?

internal/errorcodes/errorcodes.go
```diff
@@ -0,0 +1,87 @@
...
+	if md == nil {
+		md = map[string]string{}
+	}
+
+	messages := []protoiface.MessageV1{
```
ItalyPaleAle, line 54: Isn't `messagev1` the interface for old protobuf messages? Which are deprecated? Can we use the v2 please?

ItalyPaleAle (commented):
See comments

internal/errorcodes/errorcodes.go
robertojrojas: The idea behind passing both separate is to give the calling code flexibility on the `errDescription`.

internal/errorcodes/errorcodes.go
robertojrojas: I was hoping to avoid `panic: assignment to entry in nil map` down the road.

ItalyPaleAle: But you don't need to copy `md := metadata`, just modify metadata :)

internal/errorcodes/errorcodes.go
robertojrojas: Hmm... I used what `Status.WithDetails` takes which in my editor shows as `protoiface.MessageV1`

internal/errorcodes/errorcodes.go
robertojrojas: I like your suggestion better. will change it.

internal/errorcodes/errorcodes.go
robertojrojas: I wanted `rid` to be a pointer so it's _optional_

internal/errorcodes/errorcodes.go
ItalyPaleAle: I see that in the code you've written so far, in most cases the error is similar or derived.

how about creating a custom error object that has a "Description()" property that is optional?

```go
type ComponentError struct {
    err error
	code codes.Code
	description string
	resourceInfoData *ResourceInfoData
	metadata map[string]string
}

func NewComponentError(code codes.Code, err error) ComponentError {
	return ComponentError{
		code: code,
		err: err,
	}
}

// Error implements the error interface.
func (c ComponentError) Error() string {
	return c.err.Error()
}

// Unwrap implements the error unwrapping interface.
func (c ComponentError) Unwrap() error {
	return c.err
}

// Description returns the description of the error.
func (c ComponentError) Description() string {
	if c.description {
		return c.description
	}
	return c.err.Error()
}

func (c *ComponentError) SetDescription(description string) {
	c.description = description
}

// Add methods to access code, resourceInfoData, metadata
// Add methods to set resourceInfoData and metadata
```

Then you can pass a ComponetnError around as an `error`. In the code that handles errors, you can check:

```go
var compErr ComponentError
if errors.As(err, &compErr) {
// ...
}
```

Using custom errors is a more idiomatic thing in Go. It also allows accessing the `err` property for unwrapping.

Some links: https://www.digitalocean.com/community/tutorials/creating-custom-errors-in-go

internal/errorcodes/errorcodes.go
ItalyPaleAle: Ok

internal/errorcodes/errorcodes.go
robertojrojas: As I mentioned in the other PR, will be shared one both PRs are merged.

internal/errorcodes/errorcodes.go
robertojrojas: Yep, custom errors are idiomatic in Go, but the idea is to follow [The proposal and use gRPC error handling: 0009-BCIRS-error-handling-codes ](https://github.com/dapr/proposals/blob/main/0009-BCIRS-error-handling-codes.md)

ItalyPaleAle: This is still compliant with the proposal :)

You could also add the GRPCStatus method to make the custom error be a gRPC error too: https://github.com/dapr/dapr/blob/master/pkg/messages/api_error.go#L79-L83

internal/errorcodes/errorcodes.go
```diff
@@ -0,0 +1,87 @@
...
+
+	ste, stErr := status.New(code, errDescription).WithDetails(messages...)
+	if stErr != nil {
+		return err
+	}
+
+	return ste.Err()
+}
+
+func NewErrorInfo(reason Reason, md map[string]string) *errdetails.ErrorInfo {
```
artursouza, line 65: Make this private

internal/errorcodes/errorcodes.go
```diff
@@ -0,0 +1,87 @@
...
+	ei := errdetails.ErrorInfo{
+		Domain:   Domain,
+		Reason:   string(reason),
+		Metadata: md,
+	}
+
+	return &ei
+}
+
+func NewResourceInfo(rid *ResourceInfoData, description string) *errdetails.ResourceInfo {
```
artursouza: make this private

internal/errorcodes/errorcodes.go
artursouza: I like the idea of having an internal struct to hold all the error information. On the other hand, I am not a fan of having it as a component specific thing. I suggest we have an internal error struct defined in kit and that being the common object for all dapr internal code. The runtime then will have a logic to transform that object into the gRPC error we agreed in the proposal.

ItalyPaleAle: +1 on getting this in Kit. I like that

robertojrojas:
@ItalyPaleAle @artursouza I like your proposed implementation related to Customer Errors and hosting them in the `kit`. I've created the following issue https://github.com/dapr/kit/issues/56 
I'm working on a PR now.

PR #3060, opened 2023-08-07
Kubernetes components: add "kubeconfigPath" metadata

ItalyPaleAle:
Related to #2992

Passing a kubeconfig path should happen via metadata and not CLI flags, which are deprecated.

Also adds a "defaultNamespace" optional metadata to the Kubernetes secret store.


berndverst (approved):
Looks good to me. Let's keep it open at least until tomorrow so @artursouza and/or @yaron2 also have a chance to see this.

berndverst (approved):

Issue #3050, opened 2023-08-05
Create component metadata file for pubsub.rabbitmq


robertojrojas:
## Describe the feature
Create component metadata file for **pubsub.rabbitmq**, 
using the following metadata file as a reference of the YAML structure, metadata fields, authentication profiles, etc...
https://github.com/dapr/components-contrib/blob/master/pubsub/azure/servicebus/topics/metadata.yaml

Use the component's source code and documentation to populate the fields needed such metadata file.

PR #3069, opened 2023-08-10
[PubSub RabbitMQ] Adds Component Metadata Schema

robertojrojas:
# Description

[PubSub RabbitMQ] Adds Component Metadata Schema.

## Issue reference

This PR will close: #3050 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


pubsub/rabbitmq/metadata.yaml
```diff
@@ -0,0 +1,183 @@
...
+      both will apply; whichever limit is hit first will be enforced.
+      Defaults to no limit.
+    example: '"1000"'
+  - name: maxLenBytes
+    type: number
+    description: |
+      Maximum length in bytes of a queue and its dead letter queue (if dead
+      letter enabled). If both maxLen and maxLenBytes are set then both will 
+      apply; whichever limit is hit first will be enforced.
+    example: '"1048576"'
```
robertojrojas, line 58: Validate `"` are needed.

pubsub/rabbitmq/metadata.yaml
```diff
@@ -0,0 +1,183 @@
...
+    default: '"false"'
+    example: '"true", "false"'
+  - name: maxLen
+    type: number
+    description: |
+      The maximum number of messages of a queue and its dead letter queue
+      (if dead letter enabled). If both maxLen and maxLenBytes are set then
+      both will apply; whichever limit is hit first will be enforced.
+      Defaults to no limit.
+    example: '"1000"'
```
robertojrojas, line 51: Validate `"` needed.

ItalyPaleAle (changes requested):
Please review this metadata.yaml CAREFULLY before marking it as ready for review

robertojrojas:
@ItalyPaleAle is this PR ready to merge?

pubsub/rabbitmq/metadata.yaml
```diff
@@ -0,0 +1,168 @@
...
+    default: '"false"'
+    example: '"true", "false"'
+  - name: reconnectWaitSeconds
+    description: |
+      Reconnect wait in Seconds.
+    type: duration
+  - name: ttlInSeconds
+    description: |
+      Default Queue TTL.
+    type: duration
```
ItalyPaleAle, line 132: 1. Why is this called `ttlInSeconds` if it's a Go duration? Shouldn't it be `ttl`?
2. Missing the default value and the example

pubsub/rabbitmq/metadata.yaml
```diff
@@ -0,0 +1,168 @@
...
+  - name: protocol
```
ItalyPaleAle, line 133: Should probably be moved to the "Authenticate Using RabbitMQ properties" auth profile since it doesn't make sense otherwise. And you can remove "If both the connectionString and this property..."

pubsub/rabbitmq/metadata.yaml
```diff
@@ -0,0 +1,168 @@
...
+  - name: caCert
+    type: string
+    description: |
+      Certificate authority certificate, required for using TLS.
+    example: "-----BEGIN CERTIFICATE-----\n<base64-encoded DER>\n-----END CERTIFICATE-----"
+  - name: clientCert
+    type: string
+    sensitive: true
+    description: |
+      Client certificate, required for authType mtls.
```
ItalyPaleAle, line 150: I don't see mentions of "authType" set to "mtlS" anywhere in the component? And if it is an option, it should be a new auth profile.

pubsub/rabbitmq/metadata.yaml
```diff
@@ -0,0 +1,168 @@
...
+    example: "-----BEGIN CERTIFICATE-----\n<base64-encoded DER>\n-----END CERTIFICATE-----"
+  - name: clientKey
+    type: string
+    sensitive: true
+    description: |
+      Client key, required for authType mtls.
+    example: "-----BEGIN RSA PRIVATE KEY-----\n<base64-encoded DER>\n-----END RSA PRIVATE KEY-----"
+  - name: saslExternal
```
ItalyPaleAle, line 158: I guess we need a new auth proifle for using TLS?

ItalyPaleAle (changes requested):

pubsub/rabbitmq/metadata.yaml
```diff
@@ -0,0 +1,168 @@
...
+      value for production environments. The value of "0" means that
+      all available messages will be pre-fetched.
+    default: '0'
+    example: '2'
+  - name: exchangeKind
+    type: string
+    description: |
+      Exchange kind of the rabbitmq exchange.
+    default: '"fanout"'
+    example: '"fanout","topic"'
```
tmacam, line 98:
```suggestion
    default: '"fanout"'
    allowedValues:
      - "fanout"
      - "topic"
    example: '"fanout","topic"'
```

pubsub/rabbitmq/metadata.yaml
```diff
@@ -0,0 +1,168 @@
...
+    description: |
+      Whether or not to requeue when sending a negative acknowledgement
+      in case of a failure.
+    type: bool
+    default: '"false"'
+    example: '"true", "false"'
+  - name: reconnectWaitSeconds
+    description: |
+      Reconnect wait in Seconds.
+    type: duration
```
tmacam, line 128: Same comments as the ones @ItalyPaleAle  made for `ttlInSeconds`: 

1. this seems like a misnomer, the name induces a `number` but the type is `duration`?!
2. no default, 
3. no example

pubsub/rabbitmq/metadata.yaml
```diff
@@ -0,0 +1,168 @@
...
+      Default Queue TTL.
+    type: duration
+  - name: protocol
+    description: |
+      The RabbitMQ protocol. If the connectionString is specified,
+      this value will be set from URI scheme of the connectionString.
+      If both the connectionString and this property are specified,
+      the value specified here must match URI scheme of the connectionString.
+    type: string
+    example: '"amqp", "amqps"'
```
tmacam, line 140: Is there a list of valid values for this field? If so, then `allowedValues` should be used to limit/document what are the allowed values here.

tmacam (changes requested):

pubsub/rabbitmq/metadata.yaml
berndverst: This is not a misnomer - this is our standard implementation of durations. If you specify an integer without unit we interpret it as seconds! If you specify a duration string like "500ms" or "2m" we accept that too.

pubsub/rabbitmq/metadata.yaml
berndverst: Chances are we do not know the full list of values that can work. In that case using an example is acceptable until we know the finite set of allowed values. That's an iterative improvement that can be made and doesn't need to be blocking.

pubsub/rabbitmq/metadata.yaml
berndverst: That's our standard implementation- it got refactored in the past (probably by me) to use the duration type. This is because an integer without unit will be interpreted as seconds. But a duration string is now also supported (like 500ms or 2m)

pubsub/rabbitmq/metadata.go
```diff
@@ -28,7 +28,7 @@ import (
 
 type rabbitmqMetadata struct {
 	pubsub.TLSProperties `mapstructure:",squash"`
-	ConsumerID           string                 `mapstructure:"consumerID"`
+	ConsumerID           string                 `mapstructure:"consumerID" mdignore:"true"`
```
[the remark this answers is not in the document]
berndverst: I thought you did that elsewhere for consumerID specifically? Either way, we can try removing this.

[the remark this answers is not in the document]
robertojrojas: Removing the  `mdignore:"true"` tag attributes causes the following error:
```
{
  "pubsub/rabbitmq": [
    "consumerID"
  ]
}
make: *** [check-component-metadata] Error 1
```

pubsub/rabbitmq/metadata.yaml
artursouza: Components should agree on metadata names. We should not have one as `ttlInSeconds` accepting numbers and another one with `ttlInSeconds` accepting duration and another one using only `ttl` with duration.

This PR is about documenting the existing attributes. I don't think we should block for that, instead, we should have a separate issue to move as many of the attributes to support duration - except when the underlying client does not support it. For example, if a TTL is always in seconds for a given message broker, duration might be odd because it cannot support ms and the behavior cannot be honored (it would need to be 1s or 0s).

pubsub/rabbitmq/metadata.yaml
artursouza: Any attribute that contains "Seconds" or the unit in the name, should be documented as integer. This is the right behavior for the UI that consumes this IMO.

pubsub/rabbitmq/metadata.yaml
berndverst: As explained in Discord - this is an internal convenience implementation due to standardizing of metadata parsing.

Originally this field used an integer type. I changed it to `time.duration`, knowing that the duration parses we used falls back to trying integers and interpreting them as seconds.

Ideally we can rename this field to be just `ttl`, but keep `ttlInSeconds` arounds for backwards compatibility.

For now it's fine to document this field as either `integer` or `duration` - it doesn't matter to me.

pubsub/rabbitmq/metadata.go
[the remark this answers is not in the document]
ItalyPaleAle: Uh this is odd. It is a built-in property: https://github.com/robertojrojas/components-contrib/blob/bb2b4952e90f4216c81b12a3c6919a00844dd685/.build-tools/pkg/metadataschema/validators.go#L146-L156

Anyhow, ok, we can leave the mdignore here for now

ItalyPaleAle (approved):

Issue #3047, opened 2023-08-05
Create component metadata for pubsub.mqtt3


robertojrojas:
## Describe the feature
Create component metadata file for **pubsub.mqtt3**, 
using the following metadata file as a reference of the YAML structure, metadata fields, authentication profiles, etc...
https://github.com/dapr/components-contrib/blob/master/pubsub/azure/servicebus/topics/metadata.yaml

Use the component's source code and documentation to populate the fields needed such metadata file.

PR #3070, opened 2023-08-10
[PubSub MQTT3] Adds Component Metadata Schema

robertojrojas:
# Description

[PubSub MQTT3] Adds Component Metadata Schema.

## Issue reference

This PR will close: #3047 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/pubsub/mqtt3/metadata.yaml
+++ b/pubsub/mqtt3/metadata.yaml
@@ -1,0 +1,68 @@
+# yaml-language-server: $schema=../../component-metadata-schema.json
+schemaVersion: v1
+type: pubsub
+name: mqtt3
+version: v1
+status: stable
+title: "MQTT3"
+urls:
+  - title: Reference
+    url: https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-mqtt3/
+metadata:
+  - name: url
+    required: true
+    description: |
+      "Address of the MQTT broker. Can be secretKeyRef to use a secret reference.
+      Use the tcp:// URI scheme for non-TLS communication. Use the ssl:// URI scheme for TLS communication."
+    type: string
+    example: '"tcp://[username][:password]@host.domain[:port]"'
+  - name: consumerID
+    description: |
+      "The client ID used to connect to the MQTT broker. Defaults to the Dapr app ID."
+    type: string
+    example: '"myMqttClientApp"'
+  - name: retain
+    description: |
+      "Defines whether the message is saved by the broker as the last known good value for a specified topic.
+       Defaults to "false"."
+    type: bool
+    default: 'false'
+    example: '"true", "false"'
+  - name: cleanSession
+    description: |
+      "Sets the clean_session flag in the connection message to the MQTT broker if "true" (more info).
+       Defaults to "false"."
+    type: bool
+    default: 'false'
+    example: '"true", "false"'
+  - name: qos
+    description: |
+      "Indicates the Quality of Service Level (QoS) of the message (more info). Defaults to 1." 
+    type: number
+    default: 1
+    example: '0, 1, 2'
+  - name: caCert
+    description: |
+      "Certificate Authority (CA) certificate in PEM format for verifying server TLS certificates." 
+    type: string
+    example: |
+      '-----BEGIN CERTIFICATE-----
+        ...
+        -----END CERTIFICATE-----'
+  - name: clientCert
+    description: |
+      "TLS client certificate in PEM format. Must be used with clientKey." 
+    type: string
+    example: |
+      '-----BEGIN CERTIFICATE-----
+        ...
+        -----END CERTIFICATE-----'
+  - name: clientKey
+    description: |
+      "TLS client key in PEM format. Must be used with clientCert. Can be secretKeyRef to use a secret reference."  
+    type: string
+    example: |
+      '-----BEGIN RSA PRIVATE KEY-----
+        ...
+        -----END RSA PRIVATE KEY-----'
+
--- a/pubsub/mqtt3/metadata.yaml
+++ b/pubsub/mqtt3/metadata.yaml
@@ -8,61 +8,57 @@ title: "MQTT3"
 urls:
   - title: Reference
     url: https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-mqtt3/
+authenticationProfiles:
+  - title: "Connection string"
+    description: "The MQTT3 Connection/Authentication."
+    metadata:
+      - name: url
+        type: string
+        required: true
+        description: |
+          Address of the MQTT broker. Can be secretKeyRef to use a secret reference.
+          Use the tcp:// URI scheme for non-TLS communication. 
+          Use the ssl:// URI scheme for TLS communication.
+        example: '"tcp://[username][:password]@host.domain[:port]"'
 metadata:
-  - name: url
-    required: true
-    description: |
-      "Address of the MQTT broker. Can be secretKeyRef to use a secret reference.
-      Use the tcp:// URI scheme for non-TLS communication. Use the ssl:// URI scheme for TLS communication."
-    type: string
-    example: '"tcp://[username][:password]@host.domain[:port]"'
   - name: consumerID
-    description: |
-      "The client ID used to connect to the MQTT broker. Defaults to the Dapr app ID."
     type: string
+    description: |
+      The client ID used to connect to the MQTT broker. Defaults to the Dapr app ID.
     example: '"myMqttClientApp"'
   - name: retain
-    description: |
-      "Defines whether the message is saved by the broker as the last known good value for a specified topic.
-       Defaults to "false"."
     type: bool
+    description: |
+      Defines whether the message is saved by the broker as the last known good value for a specified topic.
     default: 'false'
     example: '"true", "false"'
   - name: cleanSession
-    description: |
-      "Sets the clean_session flag in the connection message to the MQTT broker if "true" (more info).
-       Defaults to "false"."
     type: bool
+    description: |
+      Sets the clean_session flag in the connection message to the MQTT broker if "true" (more info).
     default: 'false'
     example: '"true", "false"'
   - name: qos
-    description: |
-      "Indicates the Quality of Service Level (QoS) of the message (more info). Defaults to 1." 
     type: number
-    default: 1
+    description: |
+      Indicates the Quality of Service Level (QoS) of the message (more info).
+    default: '1'
     example: '0, 1, 2'
   - name: caCert
-    description: |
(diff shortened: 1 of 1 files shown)


pubsub/mqtt3/metadata.yaml
```diff
@@ -0,0 +1,64 @@
...
+urls:
+  - title: Reference
+    url: https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-mqtt3/
+authenticationProfiles:
+  - title: "Connection string"
+    description: "The MQTT3 Connection/Authentication."
+    metadata:
+      - name: url
+        type: string
+        required: true
```
artursouza, line 17: sensitive: true -> since this can contain a password.

pubsub/mqtt3/metadata.yaml
```diff
@@ -0,0 +1,67 @@
...
+        sensitive: true
+        description: |
+          Address of the MQTT broker. Can be secretKeyRef to use a secret reference.
```
ItalyPaleAle, line 20: Once again we don't need to say it can be a secretkeyRef here.

ItalyPaleAle (changes requested):
Please review this metadata.yaml carefully before marking it as ready for review

robertojrojas:
@ItalyPaleAle  is this PR ready for merge?

pubsub/mqtt3/metadata.yaml
```diff
@@ -0,0 +1,67 @@
...
+name: mqtt3
+version: v1
+status: stable
+title: "MQTT3"
+urls:
+  - title: Reference
+    url: https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-mqtt3/
+authenticationProfiles:
+  - title: "Connection string"
+    description: "The MQTT3 Connection/Authentication."
```
ItalyPaleAle, line 13: Description doesn't match title

pubsub/mqtt3/metadata.yaml
```diff
@@ -0,0 +1,67 @@
...
+    metadata:
+      - name: url
+        type: string
+        required: true
+        sensitive: true
+        description: |
+          Address of the MQTT broker.
+          Use the tcp:// URI scheme for non-TLS communication. 
+          Use the ssl:// URI scheme for TLS communication.
```
ItalyPaleAle, line 22:
```suggestion
          Use the `tcp://`` URI scheme for non-TLS communication. 
          Use the `ssl://`` URI scheme for TLS communication.
```

pubsub/mqtt3/metadata.yaml
```diff
@@ -0,0 +1,67 @@
...
+        example: '"tcp://[username][:password]@host.domain[:port]"'
+metadata:
+  - name: consumerID
```
ItalyPaleAle, line 25: This is a built-in property for all pubsub components, so no need to add it here. See .build-tools/pkg/metadataschema/validators.go

pubsub/mqtt3/metadata.yaml
```diff
@@ -0,0 +1,67 @@
...
+  - name: retain
+    type: bool
+    description: |
+      Defines whether the message is saved by the broker as the last known good value for a specified topic.
+    default: 'false'
+    example: '"true", "false"'
+  - name: cleanSession
+    type: bool
+    description: |
+      Sets the clean_session flag in the connection message to the MQTT broker if "true" (more info).
```
ItalyPaleAle, line 39: "more info"?

pubsub/mqtt3/metadata.yaml
```diff
@@ -0,0 +1,67 @@
...
+    default: 'false'
+    example: '"true", "false"'
+  - name: qos
+    type: number
+    description: |
+      Indicates the Quality of Service Level (QoS) of the message (more info).
```
ItalyPaleAle, line 45: Another "More info" that points to nowhere

pubsub/mqtt3/metadata.yaml
```diff
@@ -0,0 +1,67 @@
...
+    default: '1'
+    example: '0, 1, 2'
+  - name: caCert
+    type: string
+    description: |
+      Certificate authority certificate, required for using TLS. 
+      Can be secretKeyRef to use a secret reference".
```
ItalyPaleAle, line 52: We explained before many times that `Can be secretKeyRef to use a secret reference".` isn't needed (also why is there a `"` in there?)

pubsub/mqtt3/metadata.yaml
```diff
@@ -0,0 +1,67 @@
...
+    description: |
+      Sets the clean_session flag in the connection message to the MQTT broker if "true" (more info).
+    default: 'false'
+    example: '"true", "false"'
+  - name: qos
+    type: number
+    description: |
+      Indicates the Quality of Service Level (QoS) of the message (more info).
+    default: '1'
+    example: '0, 1, 2'
```
ItalyPaleAle, line 47: Should be an enum, with values "0", "1" or "2"

ItalyPaleAle (commented):
Please, make sure you double-check and triple-check metadata.yaml files. I see the same mistakes made over and over again in each PR, for which we've provided feedback multiple times.

This includes things like copy-pasting a description that includes "(more info") and not noticing there's a link...

ItalyPaleAle (changes requested):
Please, make sure you double-check and triple-check metadata.yaml files. I see the same mistakes made over and over again in each PR, for which we've provided feedback multiple times.

This includes things like copy-pasting a description that includes "(more info") and not noticing there's a link...

pubsub/mqtt3/metadata.yaml
ItalyPaleAle: Sorry but this didn't fix the issue

pubsub/mqtt3/metadata.yaml
ItalyPaleAle: I think the docs had a link, which is now removed

pubsub/mqtt3/metadata.yaml
ItalyPaleAle: I think the docs had a link, which is now removed

robertojrojas: The `more info` link was pointing to a non-TLS [URL](http://www.steves-internet-guide.com/mqtt-clean-sessions-example/).  I have not found a reliable URL to point to. I'm open to any suggestions.

ItalyPaleAle: What's the issue with the URL above?

pubsub/mqtt3/metadata.yaml
robertojrojas: @ItalyPaleAle what component metadata yaml I could use as an example to better understand what you mean by `enum` in this context?

ItalyPaleAle: https://github.dev/dapr/components-contrib/blob/master/bindings/http/metadata.yaml#L62-L70

pubsub/mqtt3/metadata.yaml
```diff
@@ -0,0 +1,63 @@
...
+    description: "Authenticate using a connection string."
+    metadata:
+      - name: url
+        type: string
+        required: true
+        sensitive: true
+        description: |
+          Address of the MQTT broker.
+          Use the `tcp://`` URI scheme for non-TLS communication. 
+          Use the `ssl://`` URI scheme for TLS communication.
```
tmacam, line 22: Based on the docs it would make sense to tie all the TLS-related/required  fields in this comment.

```suggestion
          Use the `ssl://`` URI scheme for TLS communication. Requires `caCert`, `clientKey`, `clientCert` to be defined.
```

tmacam (changes requested):

pubsub/mqtt3/metadata.yaml
robertojrojas: The URL is pointing to an unsecured location.  I believe the URLs in the Metadata yaml files should only be TLS/SSL.

ItalyPaleAle: Preferably, yes. But if there's no alternative, I'm fine with a non-TLS URL. Since these links are used for docs only, a link to a non-TLS page is better than no link.

pubsub/mqtt3/metadata.yaml
```diff
@@ -0,0 +1,71 @@
...
+    type: bool
+    description: |
+      Defines whether the message is saved by the broker as the last known good value for a specified topic.
+    default: 'false'
+    example: '"true", "false"'
+  - name: cleanSession
+    type: bool
+    description: |
+      When the value is set to "true", sets the clean_session flag in the connection message to the MQTT broker.
+      See MQTT Clean Sessions Example.
```
ItalyPaleAle, line 52:
```suggestion
```

pubsub/mqtt3/metadata.yaml
```diff
@@ -0,0 +1,71 @@
...
+    description: "Authenticate using a connection string."
+    metadata:
+      - name: url
+        type: string
+        required: true
+        sensitive: true
+        description: |
+          Address of the MQTT broker.
+          Use the `tcp://`` URI scheme for non-TLS communication. 
+          Use the `ssl://`` URI scheme for TLS communication. Requires `caCert`, `clientKey`, `clientCert` to be defined.
```
ItalyPaleAle, line 22:
```suggestion
          Use the `ssl://`` URI scheme for TLS communication (requires `caCert`, `clientKey`, `clientCert` to be defined).
```

pubsub/mqtt3/metadata.yaml
```diff
@@ -0,0 +1,71 @@
...
+    url:
+      title: "MQTT Clean Sessions Example"
+      url: "http://www.steves-internet-guide.com/mqtt-clean-sessions-example/"
+    default: 'false'
+    example: '"true", "false"'
+  - name: qos
+    type: number
+    description: |
+      Indicates the Quality of Service Level (QoS) of the message.
+      See MQTT Essentials - Part 6.
```
ItalyPaleAle, line 62:
```suggestion
```

pubsub/mqtt3/metadata.yaml
```diff
@@ -0,0 +1,71 @@
...
+          Use the `ssl://`` URI scheme for TLS communication. Requires `caCert`, `clientKey`, `clientCert` to be defined.
+        example: '"tcp://[username][:password]@host.domain[:port]"'
+      - name: caCert
+        type: string
+        description: |
+          Certificate authority certificate, required for using TLS.
+        example: '"-----BEGIN CERTIFICATE-----\n<base64-encoded DER>\n-----END CERTIFICATE-----"'
+      - name: clientCert
+        type: string
+        sensitive: true
```
ItalyPaleAle, line 31:
```suggestion
```

PR #3077, opened 2023-08-14
Fix wasm binding wasi http

Taction:
# Description

Found there are concurrent map writes errors and added a test case to show the error and ensure the fix work.

This is mainly related to the upstream sdk. Once the upstream work is done, I will change the draft status.

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


Taction:
waiting for https://github.com/stealthrocket/wasi-go/pull/93 to be merged, and update the upstream version.

brendandburns:
@Taction I think that the wasm output binding actually has a bug here.

WASM is intended to be single threaded, so parallel execution within a single WASM runtime is probably not the best idea. Given the current implementation where the `Runtime` is instantiated once, this is guaranteed to happen under heavy load.

I think that we need to either instantiate a pool of `Runtime` objects and/or use a Lock to prevent concurrent execution within a single `Runtime`.

wdyt?

ItalyPaleAle:
Thank you!

@berndverst ok to put this in the next 1.12 RC?

ItalyPaleAle (approved):

Issue #3054, opened 2023-08-05
Create component metadata file for state.cockroachdb


robertojrojas:
## Describe the feature
Create component metadata file for **state.cockroachdb**, 
using the following metadata file as a reference of the YAML structure, metadata fields, authentication profiles, etc...
https://github.com/dapr/components-contrib/blob/master/state/aws/dynamodb/metadata.yaml

Use the component's source code and documentation to populate the fields needed such metadata file.

PR #3082, opened 2023-08-15
[State CockroachDB] Continuation from issue 3054

robertojrojas:
# Description

Continuation from issue 3054 - updates as per PR review

## Issue reference

This PR will close: #3054

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/state/cockroachdb/metadata.yaml
+++ b/state/cockroachdb/metadata.yaml
@@ -21,9 +21,9 @@ authenticationProfiles:
       - name: connectionString
         required: true
         sensitive: true
-        description: The connection string for the PostgreSQL database
+        description: The connection string for the Cockroach database
         example: |
-          "host=localhost user=postgres password=example port=5432 connect_timeout=10 database=dapr_test"
+          "host=localhost user=cockroach password=example port=5432 connect_timeout=10 database=dapr_test"
         type: string
 metadata: 
   - name: timeoutInSeconds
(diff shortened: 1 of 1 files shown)


state/cockroachdb/metadata.yaml
```diff
@@ -21,9 +21,9 @@ authenticationProfiles:
       - name: connectionString
         required: true
         sensitive: true
-        description: The connection string for the PostgreSQL database
+        description: The connection string for the Cockroach database
```
ItalyPaleAle, line 24: The name is CockroachDB, which is a trademark, so we should respect that: https://en.wikipedia.org/wiki/CockroachDB

state/cockroachdb/metadata.yaml
```diff
@@ -21,9 +21,9 @@ authenticationProfiles:
...
         example: |
-          "host=localhost user=postgres password=example port=5432 connect_timeout=10 database=dapr_test"
+          "host=localhost user=cockroach password=example port=5432 connect_timeout=10 database=dapr_test"
```
ItalyPaleAle, line 26: According to https://docs.dapr.io/reference/components-reference/supported-state-stores/setup-cockroachdb/ the "root" account for CockroachDB is actually called "root"

ItalyPaleAle (changes requested):

ItalyPaleAle (approved):

PR #3084, opened 2023-08-15
tests/certification: Updates dapr/dapr to `master` HEAD

JoshVanL:
Updates certification tests dapr/dapr to master.

`runtime.Options` has been removed so embedding has been refactored to accept options which are applied to the `runtime.Config`

Updates the certification tests to use the new func signatures.


tests/certification/embedded/embedded.go
```diff
@@ -128,50 +133,88 @@ func WithProfilingEnabled(enabled bool) Option {
...
 			return nil, nil, err
 		}
+		runtimeConfig.DaprInternalGRPCPort = strconv.Itoa(port)
 	}
 
 	variables := map[string]string{
-		env.AppID:           runtimeConfig.ID,
-		env.AppPort:         fmt.Sprintf("%d", runtimeConfig.ApplicationPort),
+		env.AppID:           runtimeConfig.AppID,
+		env.AppPort:         fmt.Sprintf("%s", runtimeConfig.ApplicationPort),
```
ItalyPaleAle, line 212: Isn't this a nop?

tests/certification/embedded/embedded.go
```diff
@@ -104,21 +109,21 @@ func WithComponentsPath(path string) Option {
 	return WithResourcesPath(path)
 }
 
-func WithProfilePort(port int) Option {
+func WithProfilePort(port string) Option {
```
ItalyPaleAle, line 112: Why changing all ports from int to string?

tests/certification/flow/sidecar/sidecar.go
```diff
@@ -80,51 +83,37 @@ func (s Sidecar) ToStep() (string, flow.Runnable, flow.Runnable) {
...
+	go func() {
+		s.errCh <- rt.Run(ctx)
+	}()
 
-	if err = rt.Run(opts...); err != nil {
-		return err
-	}
-
-	daprClient, err := dapr.NewClientWithPort(fmt.Sprintf("%d", rtConf.APIGRPCPort))
+	daprClient, err := dapr.NewClientWithPort(fmt.Sprintf("%s", rtConf.DaprAPIGRPCPort))
```
ItalyPaleAle, line 116: nop too?

tests/certification/embedded/embedded.go
JoshVanL: :+1:

tests/certification/embedded/embedded.go
JoshVanL: This is what the runtime config accepts.

tests/certification/flow/sidecar/sidecar.go
JoshVanL: :+1:

tests/certification/embedded/embedded.go
ItalyPaleAle: But before we were accepting ints here and then converting to string in a single place. This change required changing all cert tests to convert from string to int themselves. Not the end of the world but just seems more verbose

JoshVanL: It’s because we have removed the runtime Options completely, so we are using the same struct which flags are being used against in dapr/dapr

ItalyPaleAle: Ok

ItalyPaleAle:
@JoshVanL looks like cert tests are all failing. PTAL

ItalyPaleAle:
Thanks for the fix. A lot of cert tests are still failing, even those that are not (normally) flaky.

berndverst:
@JoshVanL you need to update the certification and conformance tests also.

berndverst:
Manually ran `go test -v -tags=conftests --count=1 ./tests/conformance -run="TestBindingsConformance/azure.blobstorage"` which seems to work now, so rerunning all conformance tests and certification tests to see whether they now pass (minus the usual flakiness).

berndverst:
@JoshVanL my bad. I broke these tests. I will update the GitHub credentials and rerun! (cc @ItalyPaleAle )

berndverst:
@JoshVanL @ItalyPaleAle, ok those Azure tests are fixed. Not sure what's going on with the remaining tests though. Please take a look as they are not broken in master.

berndverst:
@JoshVanL sorry for the many comments:

It seems runtime changes are in fact breaking certification tests not broken in master:

```
bindings.kafka
middleware.http.bearer
middleware.http.ratelimit
pubsub.gcp.pubsub
pubsub.kafka
pubsub.mqtt3
pubsub.pulsar
pubsub.rabbitmq
state.redis
state.sqlite
state.gcp.firestore
```

Some of those changes are concerning, for example HTTP Middleware Bearer returning 200 when 401 is expected.
Rate limit middleware not working etc

berndverst:
My recommendation is to try to recreate this PR with an updated dapr/dapr in cert tests and the most minimal changes you can make.

There is a lot of refactoring here that isn't strictly necessary and can itself be the root cause of the problems.

JoshVanL:
@berndverst The tests are now green (as I understand mqtt3 is flakey). A summery of the issues which were causing the tests to fail:

- The metrics server was not running with port=0, so any test with multiple concurrent daprd's would fail.
- `Run` no longer blocks so tests relying on things like waiting for a component to init timeout would fail because the client would error.
- Previously, we were doing a weird check on a config const and loading in the config. Now we just set Config on the runtime config.
https://github.com/dapr/components-contrib/blob/3398bf407a2a83b9a7a011bfdf507c6244c3d058/tests/certification/embedded/embedded.go#L40
https://github.com/dapr/components-contrib/blob/3398bf407a2a83b9a7a011bfdf507c6244c3d058/tests/certification/embedded/embedded.go#L208

ItalyPaleAle:
Some tests are still failing, even beyond the "usual" flakiness

berndverst:
Fire store still seems to fail due to some auth issue:

```
        	Error Trace:	/home/runner/work/components-contrib/components-contrib/tests/certification/state/gcp/firestore/gcp_firestore_test.go:60
        	            				/home/runner/work/components-contrib/components-contrib/tests/certification/flow/flow.go:227
        	Error:      	Received unexpected error:
        	            	error saving state: rpc error: code = Internal desc = failed saving state in state store statestore-basic: rpc error: code = Unauthenticated desc = transport: per-RPC creds failed due to error: oauth2/google: unable to generate access token: Post "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/***:generateAccessToken": oauth2/google: invalid response when retrieving subject token: Get "https://pipelines.actions.githubusercontent.com/zkjY93hWfDqVFOaPHPOTK1OCjjOXg2vyQ25comzunJkaqH4UMR/00000000-0000-0000-0000-000000000000/_apis/distributedtask/hubs/Actions/plans/453118f3-3f2a-4edc-bcdb-26e57e1c3ebf/jobs/5362456e-705b-57be-e7e9-adaf7c774586/idtoken?api-version=2.0&audience=https%3A%2F%2Fiam.googleapis.com%2F***": context canceled
```

berndverst (approved):

berndverst:
GCP failures are a credential thing. We'll fix that separately.

PR #3093, opened 2023-08-21
Define DynamoDB Transaction limit

berndverst:
# Description

Defines the DynamoDB Transaction limit of 100 operations in a single transaction, similar to the way it is defined for CosmosDB.

This is necessary for Workflow, so it can be inferred what the max parallel operation batch size should be.


RyanLettieri (approved):
LGTM

yaron2 (dismissed):

yaron2:
Linter's complaining

ItalyPaleAle (approved):

PR #3094, opened 2023-08-21
Define ETCD Transaction limit

berndverst:
# Description

Define ETCD Transaction limit

Because the transaction limit in ETCD depends on the server (via server flag `--max-txn-ops`), but with a default of 128 if unset I have also exposed this configuration option via a new `maxTxnOps` component metadata option.

### DIFF
--- a/state/etcd/etcd.go
+++ b/state/etcd/etcd.go
@@ -43,6 +43,7 @@ type Etcd struct {
 	features      []state.Feature
 	logger        logger.Logger
 	schema        schemaMarshaller
+	maxTxnOps     int
 }
 
 type etcdConfig struct {
@@ -53,6 +54,8 @@ type etcdConfig struct {
 	CA        string `json:"ca"`
 	Cert      string `json:"cert"`
 	Key       string `json:"key"`
+	// Transaction server options
+	MaxTxnOps int `mapstructure:"maxTxnOps"`
 }
 
 // NewEtcdStateStoreV1 returns a new etcd state store for schema V1.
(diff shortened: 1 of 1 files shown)


state/etcd/etcd.go
```diff
@@ -53,6 +54,8 @@ type etcdConfig struct {
 	CA        string `json:"ca"`
 	Cert      string `json:"cert"`
 	Key       string `json:"key"`
+	// Transaction server options
```
ItalyPaleAle, line 57: Nit:

```suggestion
	// Transactional store options
```

state/etcd/etcd.go
```diff
@@ -134,7 +138,10 @@ func (e *Etcd) Features() []state.Feature {
 }
 
 func metadataToConfig(connInfo map[string]string) (*etcdConfig, error) {
-	m := &etcdConfig{}
+	m := &etcdConfig{
+		// This is the default value for maximum ops per transaction, configurtable via etcd server flag --max-txn-ops.
+		MaxTxnOps: 128,
```
ItalyPaleAle, line 143: Can you make this a constant?

state/etcd/etcd.go
berndverst: Not what I was going for. I want to be clear that these options must match options configured on the server. Other suggestions @ItalyPaleAle ?

state/etcd/etcd.go
berndverst: I guess - but this right here is how we initialize defaults other places too. I can point you to many examples in contrib.

state/etcd/etcd.go
ItalyPaleAle: It was just a nit.

state/etcd/etcd.go
ItalyPaleAle: Ok. It was just another nit too.

ItalyPaleAle (approved):

PR #3095, opened 2023-08-22
Add `\rerun <ID>` command for failed jobs in workflows

berndverst:
# Description

This adds a command which allows rerunning of just the failed jobs on an existing workflow run ID.

This can help with reducing unnecessary workflow runs. It also allows trusted community members who are not approvers to perform this action (the same people who can run `/ok-to-test`).

Usage:
`\rerun <ID>`, where the ID is obtained from URL of a given workflow run

### DIFF
--- a/.github/scripts/dapr_bot.js
+++ b/.github/scripts/dapr_bot.js
@@ -94,6 +94,10 @@ async function handleIssueCommentCreate({ github, context }) {
         case '/ok-to-test':
             await cmdOkToTest(github, issue, isFromPulls)
             break
+        case command.match(/^\/rerun /)?.input:
+            const workflowrunid = command.match(/\d+/)[0];
+            await rerunWorkflow(github, issue, workflowrunid)
+            break
         default:
             console.log(
                 `[handleIssueCommentCreate] command ${command} not found, exiting.`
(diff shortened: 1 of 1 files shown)


.github/scripts/dapr_bot.js
```diff
@@ -94,6 +94,10 @@ async function handleIssueCommentCreate({ github, context }) {
         case '/ok-to-test':
             await cmdOkToTest(github, issue, isFromPulls)
             break
+        case command.match(/^\/rerun /)?.input:
```
berndverst, line 97:
```suggestion
        case command.match(/^\/rerun \d+/)?.input:
```

.github/scripts/dapr_bot.js
```diff
@@ -234,3 +238,18 @@ async function cmdOkToTest(github, issue, isFromPulls) {
...
 }
+
+/**
+ * Rerun all failed jobs of a given workflow run ID.
+ * @param {*} github GitHub object reference
+ * @param {*} issue GitHub issue object
+ * @param {int} workflowrunid the workflow run ID for which to rerun all failed jobs
+ */
+async function rerunWorkflow(github, issue, workflowrunid) {
+    // Get pull request
```
berndverst, line 249:
```suggestion
    // Rerun all failed jobs of the specified workflow run
```

ItalyPaleAle (approved):

berndverst:
There is a 5% chance this will actually work given my Javascript skills, but it won't break anything. Testing in production :) - Let's see.

PR #3098, opened 2023-08-23
[PubSub][Metadata] The consumerID is Repeated in metadata-component-bundle.json for PubSub Components

robertojrojas:
The consumerID is Repeated in metadata-component-bundle.json for PubSub Components. This PR removes it from the `metadata.yaml` as it will be added automatically.

For example, the Pubsub/Redis conmponent metadata:
```
{
  {
      "name": "consumerID",
      "description": "The consumer group ID",
      "type": "string",
      "example": "myGroup"
  },
  {
      "name": "consumerID",
      "description": "Set the consumer ID to control namespacing. Defaults to the app's ID.",
      "type": "string",
      "example": "\"{namespace}\"",
      "url": {
          "title": "Documentation",
          "url": "https://docs.dapr.io/developing-applications/building-blocks/pubsub/howto-namespace/"
      }
  }
}
```

# Description

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/pubsub/azure/eventhubs/metadata.yaml
+++ b/pubsub/azure/eventhubs/metadata.yaml
@@ -97,9 +97,3 @@ metadata:
     description: |
       Storage container name.
     example: '"myeventhubstoragecontainer"'
-  - name: consumerId
-    type: string
-    required: true # consumerGroup is an alias for this field, let's promote this to default
-    description: |
-      The name of the Event Hubs Consumer Group to listen on.
-    example: '"group1"'
--- a/internal/component/azure/eventhubs/metadata.go
+++ b/internal/component/azure/eventhubs/metadata.go
@@ -28,7 +28,7 @@ import (
 type AzureEventHubsMetadata struct {
 	ConnectionString        string `json:"connectionString" mapstructure:"connectionString"`
 	EventHubNamespace       string `json:"eventHubNamespace" mapstructure:"eventHubNamespace"`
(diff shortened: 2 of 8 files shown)


berndverst:
I understand the problem you are trying to solve. And the duplication occurs because of the runtime governed use of `consumerID` and the component specific definition of this.

That being said, the idea behind the metadata is that one day we will generate our Dapr documentation straight from this metadata! And because the consumerID may behave slightly differently, or needs explanations to relate to the cloud service / technology specific terms it is imperative we maintain the current descriptions.

Therefore I do not think we should merge this PR. The duplication is better than losing this description. Perhaps in any tooling that processes the metadata you could special case the `consumerID` to deal with this instead.

pubsub/azure/eventhubs/metadata.yaml
```diff
@@ -96,10 +96,4 @@ metadata:
     required: true
     description: |
       Storage container name.
-    example: '"myeventhubstoragecontainer"'
-  - name: consumerId
-    type: string
-    required: true # consumerGroup is an alias for this field, let's promote this to default
-    description: |
-      The name of the Event Hubs Consumer Group to listen on.
```
berndverst, line 104: This description would be lost even though it's helpful to know that consumerID here means consumer group.

robertojrojas:
@berndverst Yeah, that makes sense. It's good that you clarified that the metadata will be used to generate the documentation. It's one of those things I was not completely clear on.

PR #3099, opened 2023-08-23
Updates github.com/choleraehyq/pid to v0.0.17 to support Go v1.21

JoshVanL:
This version support Go v1.21, see https://github.com/choleraehyq/pid/releases/tag/v0.0.17


berndverst:
@JoshVanL let's merge this after cutting the release branch. We have agreed not to use Go 1.21 for this particular release.

JoshVanL:
@berndverst sounds good! I’ve actually had quite a bit of unexpected pain with Go 1.21 with surprising things not compiling. This dep comes from Kitex which fails to compile with 1.21

Let’s close this for now, and revisit after the release branch to reduce noise.

PR #3106, opened 2023-08-28
Upgrade dependencies ahead of 1.12.0-rc.1

berndverst:
# Description

Upgrades dependencies ahead of 1.12.0-rc.1.

The following direct dependencies with newer versions have *not* been updated:

```
# We are using a custom fork due to a Goroutine performance issue
dubbo.apache.org/dubbo-go/v3 v3.0.3-0.20230118042253-4f159a2b38f3 [v3.1.0] => dubbo.apache.org/dubbo-go/v3 v3.0.3-0.20230118042253-4f159a2b38f3 [v3.1.0]
github.com/apache/dubbo-go-hessian2 v1.11.5 [v1.12.2]

# Upgrade causes a proto registration conflict
github.com/aerospike/aerospike-client-go/v6 v6.12.0 [v6.14.0]

# Unclear whether this version works with Azure Eventhubs in SASL auth mode
github.com/Shopify/sarama v1.37.2 [v1.41.0] => github.com/Shopify/sarama v1.37.2 [v1.41.0]

# Upgrading these together causes method signature conflicts
github.com/apache/thrift v0.13.0 [v0.18.1]
github.com/cloudwego/kitex v0.5.0 [v0.7.0]
github.com/cloudwego/kitex-examples v0.1.1 [v0.2.0]

# Upgrade causes significant method changes or perf issues
github.com/hazelcast/hazelcast-go-client v0.0.0-20190530123621-6cf767c2f31a [v1.4.1]

# Upgrade causes incompatibilities
github.com/stealthrocket/wasi-go v0.7.6-0.20230718231108-c3d30af59057 [v0.8.0]

# Currently unused
go.temporal.io/api v1.18.1 [v1.24.0]
go.temporal.io/sdk v1.21.1 [v1.24.0]

# Cassandra library upgrade causes test failures at this time
github.com/gocql/gocql v1.5.2 [v1.6.0]

# Kubernetes changes
k8s.io/api v0.26.3 [v0.28.1]
k8s.io/apiextensions-apiserver v0.26.3 [v0.28.1]
k8s.io/apimachinery v0.26.3 [v0.28.1]
k8s.io/client-go v0.26.3 [v0.28.1]
```

Issue #2874, opened 2023-05-30
Kafka Binding v1.10.7: consumer/broker/0 disconnecting due to error processing FetchRequest: EOF\n


Andemki:
## What version of Dapr?

1.10.7

## Expected Behavior

Dapr and Azure Event Hub integration using Kafka binding work without errors

## Actual Behavior

I've been using bindings.kafka component for recieving messages from Azure Event Hub in my application. Everything work properly in Dapr versions 1.8.7. But when I updated dapr components to version 1.10.7, my application stopped receiving new messages from EventHub. I've noticed the following error message in the sidecar container (daprd):

>2023-05-30 17:50:14 time="2023-05-30T16:50:14.5142882Z" level=debug msg="consumer/broker/0 disconnecting due to error processing FetchRequest: EOF\n" app_id=my-eventhublistener component="my-eventhub-receiver (bindings.kafka/v1)" instance=f3688891ba8e scope=dapr.contrib type=log ver=1.10.7

The last working version is 1.8.7. Starting from Dapr v.1.9.0 the Kafka binding component doesn't work for us.

Yaml output of my bindings.kafka component:

```
apiVersion: dapr.io/v1alpha1
auth:
  secretStore: my-kv-secrets
kind: Component
metadata:
  annotations:
    meta.helm.sh/release-name: my-eventhublistener
    meta.helm.sh/release-namespace: my-namespace
  labels:
    app.kubernetes.io/managed-by: Helm
  name: my-eventhub-receiver
  namespace: my-namespace
spec:
  metadata:
  - name: brokers
    value: my-eh-namespace.servicebus.windows.net:9093
  - name: topics
    value: my-eh-data
  - name: consumerGroup
    value: my-eventhublistener
  - name: authRequired
    value: "true"
  - name: saslUsername
    value: $ConnectionString
  - name: saslPassword
    secretKeyRef:
      key: eventhub-connectionstring
      name: eventhub-connectionstring
  type: bindings.kafka
  version: v1
```

## Steps to Reproduce the Problem

1. Create a kafka.binding component to receive messages from Event Hub in your ASP.Net Core application
3. Deploy your application and the component to the Kubernetes
4. Check that your application doesn't receive messages from Event Hub

berndverst:
Possibly related to https://github.com/Shopify/sarama/issues/913

berndverst:
The error itself comes from the `sarama` library

berndverst:
[...]

Issue #2755, opened 2023-04-05
Kafka Binding v1.10.4: kafka: client has run out of available brokers to talk to


Andemki:
## What version of Dapr?

1.10.4

## Expected Behavior

Dapr and Azure Event Hub integration using Kafka binding work without errors

## Actual Behavior

The same issue as I had for version 1.8: https://github.com/dapr/dapr/issues/5342

I've been using bindings.kafka component for recieving messages from Azure Event Hub in my application. Everything work properly in Dapr versions 1.8.6. But when I updated dapr components to version 1.10.4, the sidecar container (daprd) of my application's pod failed to start and I received the following error message:

> level=fatal msg="process component my-eventhub-receiver error: [INIT_COMPONENT_FAILURE]: initialization error occurred for my-eventhub-receiver (bindings.kafka/v1): kafka: client has run out of available brokers to talk to: read tcp 10.240.0.177:39836->20.40.127.97:9093: read: connection reset by peer

Yaml output of my bindings.kafka component:

```
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  annotations:
    meta.helm.sh/release-name: my-eventhublistener
    meta.helm.sh/release-namespace: mynamespace
  creationTimestamp: "2023-04-05T11:55:46Z"
  generation: 1
  labels:
    app.kubernetes.io/managed-by: Helm
  name: my-eventhub-receiver
  namespace: mynamespace
  resourceVersion: "217779032"
  uid: 28628338-0e88-4fe3-886a-0a3bb8dcce90
spec:
  metadata:
  - name: brokers
    value: my-eh-namespace.servicebus.windows.net:9093
  - name: topics
    value: my-eh-data
  - name: consumerGroup
    value: my-eventhublistener
  - name: authRequired
    value: "true"
  - name: saslUsername
    value: $ConnectionString
  - name: saslPassword
    secretKeyRef:
      key: connectionString
      name: my-eventhublistener-eventhub
  type: bindings.kafka
  version: v1
```

## Steps to Reproduce the Problem

1. Create a kafka.binding component to receive messages from Event Hub in your application
2. Deploy your application and the component to the Kubernetes
3. Check the daprd component is not in Running state.

ItalyPaleAle:
@mukundansundar @DeepanshuA @skyao you're the Kafka experts, could you please take a look at this?

skyao:
[...]

PR #3108, opened 2023-08-29
chore(deps): bump sarama to v1.41.0

### LINKED ISSUE #3104: Update Kafka sarama package to 1.41
## Expected Behavior

When consumer pods scale out, pods should automatically be included in the consumer groups and immediately begin consuming once all health checks have passed and the sidecar states that it has joined the consumer group.

## Actual Behavior

Often times pods do not ever join the consumer group. These pods reflect as healthy and log that they have joined the consumer group, but actually do not pick up any work. They do not reflect on Kafka as part of the consumer group. This appears to be related to various bugs in the shopify/sarama Go library for Kafka and by all accounts is fixed in version 1.41.0 which was released last week.

Related issues:
https://github.com/IBM/sarama/issues/1516
https://github.com/IBM/sarama/issues/2621

Fix version:
https://github.com/IBM/sarama/releases/tag/v1.41.0

1. Create a topic in Kafka with a high number of partitions (30 or more)
2. Define a consumer pod that is subscribed to the topic with a low number of pods by default with a flexible HPA config
3. Use some kind of publisher to generate a high volume of traffic (at least 1000 events per second)
4. Observe in K8s that pods have scaled out while consumer count in the consumer group does not match the number of pods.

Likely side effect is consumer lag grows dramatically and consumption remains fixed to only a handful of pods. In my case, I am running a topic with 50 partitions, 50 pods, but only have 29 consumers in the consumer group.

## Release Note

RELEASE NOTE:

dnwe:
# Description

Bump sarama to v1.41.0

Also remove old module replace pin to v1.37.2 as that shouldn't be necessary since https://github.com/dapr/docs/pull/3474 added documentation to add a config version pin for Azure EventHubs users instead.

Note: the module path has changed to github.com/IBM/sarama since ownership transitioned away from Shopify

## Issue reference

Fixes #3104

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


berndverst:
Please do not upgrade any deps at this time. I upgraded everything there was to upgrade safely yesterday and merged that this morning. We cannot safely upgrade sarama at this time.

See https://github.com/dapr/components-contrib/pull/3106

berndverst:
The original issue with `sarama` were these:
https://github.com/dapr/components-contrib/issues/2755
https://github.com/dapr/components-contrib/issues/2874

After upgrading sarama we encountered several issues. For this reason we will not take the risk to upgrade sarama at this time (just ahead of the 1.12 release).

To upgrade this library we will need to perform further manual testing and we do not have the bandwidth at this time.

berndverst:
@dnwe I'll see if we can run some additional tests - if we can get EventHubs to work with this version we can reopen this PR. That being said, I'm not sure anyone has time ahead of the 1.12 release unfortunately.

dnwe:
@berndverst yes no problem, I expect https://github.com/IBM/sarama/issues/2470 was the underlying issue
 
The problem with supporting Azure Event Hubs is that it isn't Apache Kafka under the covers, it's an intermediate proxy that supports a subset of the Kafka APIs at various versions and then maps them onto Event Hubs protocol(s) at the backend. As as result Sarama's current mechanism of specifying the KAFKA_VERSION to determine what protocol versions to support and use doesn't really work properly with Event Hubs because it supports an unusual set of protocols and even defines minimum versions for ProduceRequest (v3) and FetchRequest (v4). For some reason EventHubs is very behind on FetchRequest and MetadataRequest so the max configuration you can use in Sarama is V1_0_0_0.

Here are the equivalent max versions for Kafka 1.0 and EventHubs:

```
    APIs                                 1.0  EventHubs
0   ProduceRequest                       v5   v7
1   FetchRequest                         v6   v6
2   ListOffsetsRequest                   v2   v7
3   MetadataRequest                      v5   v5
8   OffsetCommitRequest                  v3   v8
9   OffsetFetchRequest                   v3   v8
10  FindCoordinatorRequest               v1   v1
11  JoinGroupRequest                     v2   v7
12  HeartbeatRequest                     v1   v4
13  LeaveGroupRequest                    v1   v4
14  SyncGroupRequest                     v1   v5
15  DescribeGroupsRequest                v1   v5
16  ListGroupsRequest                    v1   v4
17  SaslHandshakeRequest                 v1   v1
18  ApiVersionsRequest                   v1   v3
19  CreateTopicsRequest                  v2   v6
20  DeleteTopicsRequest                  v1   v2
22  InitProducerIdRequest                v0   v1
23  OffsetForLeaderEpochRequest          v0   v0
32  DescribeConfigsRequest               v0   v2
36  SaslAuthenticateRequest              v0   v1
37  CreatePartitionsRequest              v0   v1
42  DeleteGroupsRequest                       v0
47  OffsetDeleteRequest                       v0
```

berndverst (approved):

dnwe:
@berndverst ah thanks, I didn't realise you were going to re-open and merge this.

I followed up with https://github.com/dapr/components-contrib/pull/3109 to put an additional safety blanket in-place that should prevent you from getting issues raised by Event Hubs users

PR #3109, opened 2023-08-30
chore(deps): bump sarama to v1.41.1

### LINKED ISSUE #3104: Update Kafka sarama package to 1.41

dnwe:
# Description

Pulling in `fix(client): force Event Hubs to use V1_0_0_0` by @dnwe in https://github.com/IBM/sarama/pull/2633

This should avoid dapr hitting any issues with Azure Event Hubs users who might not have followed the instructions in dapr/docs#3474 to pin their version number.

A small number of other changes are included in the release, but nothing I'd deem to be any additional risk over the existing upgrade to v1.41.0:
https://github.com/IBM/sarama/releases/tag/v1.41.1

## Issue reference

Fixes #3104

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,7 +24,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/storage/azqueue v1.0.0
 	github.com/Azure/go-amqp v1.0.1
 	github.com/DATA-DOG/go-sqlmock v1.5.0
-	github.com/IBM/sarama v1.41.0
+	github.com/IBM/sarama v1.41.1
 	github.com/aerospike/aerospike-client-go/v6 v6.12.0
 	github.com/alibaba/sentinel-golang v1.0.4
 	github.com/alibabacloud-go/darabonba-openapi v0.2.1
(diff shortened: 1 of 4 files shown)


berndverst:
Oh very cool @dnwe - but will this actually work for us? We are setting a default protocol version of V2_0_0_0. Will this change override it regardless?

I reached out to the EventHubs team last night by the way to see if they will ever increase their supported Kafka protocol versions :)

dnwe:
@berndverst yes this will cause Sarama to override the config for event hubs at connection time if it is outside the supported range, regardless of what is set in dapr

berndverst (approved):

Issue #3111, opened 2023-09-04
Convert Zeebe integration tests to Dapr certification tests


akkie:
The Zeebe integration tests should be converted to Dapr certifications tests, so that they run in the pipeline and to stabilize the API.

## Release Note

RELEASE NOTE: **ADD**: Implement certification tests for Zeebe binding

PR #3112, opened 2023-09-04
Migrate Zeebe tests to certification tests

akkie:
# Description

Converte Zeebe integration tests to Dapr certifications tests.

Fixes: #3111

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#3712


bindings/zeebe/command/command.go
```diff
@@ -30,7 +30,7 @@ import (
 const (
 	// operations.
 	TopologyOperation         bindings.OperationKind = "topology"
-	DeployProcessOperation    bindings.OperationKind = "deploy-process"
```
ItalyPaleAle, line 33: If you want to change this, I'm ok, but `deploy-process` must be kept as an alias for backwars compatibility please

tests/certification/bindings/zeebe/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+# TLS
+
+Zeebe supports TLS connection between the client and Zeebe itself: https://docs.camunda.io/docs/self-managed/zeebe-deployment/security/secure-client-communication/
+
+The tests will test TLS and non-TLS connections. Therefore a cert and a private key needs to be created.
+
+## Generate cert
+
+For the test there is a predefined cert and key with a validaity of 10 years. The cert was generated with the following command:
```
cicoyle, line 32:
```suggestion
For the test there is a predefined cert and key with a validity of 10 years. The cert was generated with the following command:
```

tests/certification/bindings/zeebe/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+```bash
+cd tests/certification/bindings/zeebe/certs
+openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -sha256 -days 3650 --nodes --addext 'subjectAltName=IP:127.0.0.1, DNS:localhost'
+```
+
+### Automatic cert generation
+
+There was also an atempt to generate the cert for each test automatically. The problem is that when using the automatic generated cert, the connection could
```
cicoyle, line 41:
```suggestion
There was also an attempt to generate the cert for each test automatically. The problem is that when using the automatic generated cert, the connection could
```

berndverst:
Converted this to a draft PR. Since the title still says "Work in progress". Additionally, please address @ItalyPaleAle's concern.

tests/certification/bindings/zeebe/command/config.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: dapr.io/v1alpha1
+kind: Configuration
+metadata:
+  name: zeebecommandbindingconfig
```
cicoyle, line 4: In looking at the other bindings, it looks like the metadata name value can include `-`. It might lend towards a nicer user experience seeing the separated words there. Not sure if there was a reason there were no `-`s present?
```suggestion
  name: zeebe-command-binding-config
```

tests/certification/bindings/zeebe/jobworker/config.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: dapr.io/v1alpha1
+kind: Configuration
+metadata:
+  name: zeebejobworkerbindingconfig
```
cicoyle, line 4: same here
```suggestion
  name: zeebe-jobworker-binding-config
```

tests/certification/bindings/zeebe/command/components/standard/zeebe-command.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: zeebe-command
+spec:
+  type: bindings.zeebe.command
```
cicoyle, line 6: Should this file be used as a conformance test or can a conformance test be added as well? Ideally, for the components we will want both explicit `certification` and `conformance` tests. Following the [doc here](https://github.com/dapr/components-contrib/blob/master/tests/conformance/README.md), conformance tests would entail a `tests/config/bindings/zeebe/bindings.yml`.

ItalyPaleAle: Components of type bindings contain a  bit of everything, so conformance tests are often not much useful. If we have a good, comprehensive cert test, that's _usually_ enough for a binding, especially for something that's more "unique" like Zeebe

cicoyle: Oh, I see, so even though our [certification lifecycle doc](https://docs.dapr.io/operations/components/certification-lifecycle/) specified that for `beta` stage we have `conformance tests` and for `alpha` we have `certification tests` having it as is, and no _explicit_ conformance tests, is sufficient for the `beta` stage?

ItalyPaleAle: That document is correct, but we need to consider the specifics of each situation. A conformance test exists to ensure that all components behave similarly and can be used "interchangeably". This is very useful for things like state stores, secret stores, etc.

However, this binding has about a dozen operations that are unique to this component, so there's not much to "conform to".

Eventually, the decision to make this beta or stable will be at the maintainers' discretion.

cicoyle: Great, thanks for explaining. I'm looking into the issues related to moving the zeebe command binding & zeebe jobworker binding to `stable` is why I was asking.

bindings/zeebe/command/metadata.yaml
```diff
@@ -15,7 +15,9 @@ binding:
     - name: topology
       description: "Obtains the current topology of the cluster the gateway is part of."
     - name: deploy-process
-      description: "Deploys a single process to Zeebe."
+      description: "Deploys a single process to Zeebe. This is deprecated! Please use deploy-resource instead."
```
ItalyPaleAle, line 18:
```suggestion
      description: "Deprecated alias of 'deploy-resource'."
```

bindings/zeebe/jobworker/jobworker.go
```diff
@@ -104,27 +106,31 @@ func (z *ZeebeJobWorker) Read(ctx context.Context, handler bindings.Handler) err
...
-
+		// Wait for context cancelation or closure.
 		select {
-		case <-z.closeCh:
 		case <-ctx.Done():
+		case <-z.closeCh:
 		}
 
 		jobWorker.Close()
-		jobWorker.AwaitClose()
```
ItalyPaleAle, line 126: Why was this removed?

bindings/zeebe/jobworker/metadata.yaml
```diff
@@ -83,4 +83,9 @@ metadata:
     description: Indicates if a job should be autocompleted or not. If not set, all jobs will be auto-completed by default. Disable it if the worker should manually complete or fail the job with either a business error or an incident
     example: "true"
     type: bool
-                                    
\ No newline at end of file
+  - name: retryBackOff
+    required: false
+    description: The backoff timeout (in ms) for the next retry after a job failed. This will only be applied if autocomplete is set to true.
```
ItalyPaleAle, line 88: This is a Go duration so not in ms.

tests/certification/bindings/zeebe/certs/cert.pem
ItalyPaleAle, line 1: We prefer when private keys are not added into Git. If that's really necessary, and you can't generate them on-the-fly, please at least add a README file that says that these keys are used for testing and should not be used for anything else.

ItalyPaleAle (dismissed):

bindings/zeebe/command/create_instance.go
```diff
@@ -75,9 +80,19 @@ func (z *ZeebeCommand) createInstance(ctx context.Context, req *bindings.InvokeR
...
+	if payload.WithResult && payload.RequestTimeout.Duration != time.Duration(0) {
+		ctxWithTimeout, cancel := context.WithTimeout(ctx, payload.RequestTimeout.Duration)
+		defer cancel()
+		response, err = cmd3.WithResult().FetchVariables(payload.FetchVariables...).Send(ctxWithTimeout)
+	} else if payload.WithResult {
+		response, err = cmd3.WithResult().FetchVariables(payload.FetchVariables...).Send(ctx)
+	} else {
+		response, err = cmd3.Send(ctx)
+	}
+
```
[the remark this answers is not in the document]
akkie: The timeout has only an affect if withResult is used. Using `withResult` means that the operation is synchronous instead of asynchronous. And the request timeout defines how long the client should wait for the workflow/process to finish to get the result.

From a code perspective, there are two ´send` methods in the client. One if `withResult` was used and which gets the timeout from the context and another one which will not use any timeout.

# without request timeout
https://github.com/camunda/zeebe/blob/71d2e6accc9f0452c72c1de7c16f18cc24d22552/clients/go/pkg/commands/create_instance.go#L165

# with request timeout
https://github.com/camunda/zeebe/blob/71d2e6accc9f0452c72c1de7c16f18cc24d22552/clients/go/pkg/commands/create_instance.go#L174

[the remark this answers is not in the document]
akkie: I can add a comment to make this more clear

bindings/zeebe/jobworker/jobworker.go
akkie: Close uses AwaitClose internally: https://github.com/camunda/zeebe/blob/71d2e6accc9f0452c72c1de7c16f18cc24d22552/clients/go/pkg/worker/jobWorker.go#L48

tests/certification/bindings/zeebe/certs/cert.pem
akkie: I added an additional readme to the certs folder which points to the readme of the Zeebe certification test, where I have described how the cert was created and why it could not be created on-the-fly.

bindings/zeebe/jobworker/metadata.yaml
akkie: Thanks, fixed it

bindings/zeebe/command/metadata.yaml
akkie: @ItalyPaleAle What do you suggest regarding the documentation itself? Should we only document "deploy-resource" there or also the alias?

akkie: The reason why I have changed the operation is that Camunda has deprecated the DeployProcess API: https://github.com/camunda/zeebe/blob/71d2e6accc9f0452c72c1de7c16f18cc24d22552/clients/go/pkg/commands/deploy.go#L53

Only adding an alias will not work, because the response for the operation is also different. Actually the API is still alpha. Would it be possible to add a breaking change here and remove the deploy-process alias completely? I don't think it make sense to use a deprecated API in a stable API later.

ItalyPaleAle: Ok I see.

Perhaps, could it still be supported for now, but showing a warning that is deprecated and will be removed in the next Dapr version? (As a separate operation just like it does today).

Although we can deprecate things in alpha components, if we can avoid that (or do that gracefully) it's always best

bindings/zeebe/command/create_instance.go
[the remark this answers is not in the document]
ItalyPaleAle: Thanks for explaining

ItalyPaleAle:
@akkie Thanks for this contribution, so far looks good :)

Could I ask one favor, please, to not force-push after we've done a review? When you force-push, GitHub features such as "changes since your last review" don't work anymore, so it makes it harder to review PRs (especially larger ones like this).

We always squash-merge anyways, so no issues with keeping a lot of commits in the history.

akkie:
> @akkie Thanks for this contribution, so far looks good :)
> 
> Could I ask one favor, please, to not force-push after we've done a review? When you force-push, GitHub features such as "changes since your last review" don't work anymore, so it makes it harder to review PRs (especially larger ones like this).
> 
> We always squash-merge anyways, so no issues with keeping a lot of commits in the history.

Sorry, I usually squash my commits to easier rebase my changes on top of the latest master. Will not do it anymore.

bindings/zeebe/command/metadata.yaml
akkie: OK, I can do that

tests/certification/bindings/zeebe/command/components/standard/zeebe-command.yaml
artursouza: @cicoyle Please, add this binding to conformance test to keep us following the process. I understand that we had discussions about how much value conformance tests can actually add to bindings, but I would prefer to keep the process until we reach a formal decision. A change to the certification process should go into proposals IMO since it is big enough.

berndverst:
FYI @akkie I pushed two commits - one to enable the certification tests in CI (this is critical), and one to fix a linter issue!

tests/certification/bindings/zeebe/command/components/standard/zeebe-command.yaml
berndverst: @artursouza adding a conformance test doesn't add value here since there isn't any standardized behavior and interface being tested in bindings. It just means taking up more GitHub actions runners.

I think this PR here is good enough to make the component stable.

berndverst (approved):

bindings/zeebe/command/metadata.yaml
berndverst: I'm ok with dropping operations in going from Alpha to Stable. It's up to the two of you @ItalyPaleAle and @akkie. For now I will merge this PR and this can be addressed in a follow-up PR.

berndverst (changes requested):
@akkie 

There is a bug in unit tests now:

```
--- FAIL: TestInit (0.00s)
    --- FAIL: TestInit/jobType_is_mandatory (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x94ea24]

goroutine 7 [running]:
testing.tRunner.func1.2({0x9d3340, 0xf1[76](https://github.com/dapr/components-contrib/actions/runs/6115916495/job/16600289608?pr=3112#step:9:77)00})
	/opt/hostedtoolcache/go/1.20.7/x64/src/testing/testing.go:1526 +0x24e
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.20.7/x64/src/testing/testing.go:1529 +0x39f
panic({0x9d3340, 0xf17600})
	/opt/hostedtoolcache/go/1.20.7/x64/src/runtime/panic.go:8[84](https://github.com/dapr/components-contrib/actions/runs/6115916495/job/16600289608?pr=3112#step:9:85) +0x213
github.com/dapr/components-contrib/bindings/zeebe/jobworker.(*ZeebeJobWorker).Close(0xb59ce0?)
	/home/runner/work/components-contrib/components-contrib/bindings/zeebe/jobworker/jobworker.go:144 +0xe4
github.com/dapr/components-contrib/bindings/zeebe/jobworker.TestInit.func1(0x0?)
	/home/runner/work/components-contrib/components-contrib/bindings/zeebe/jobworker/jobworker_test.go:61 +0x1a5
testing.tRunner(0xc0001f5520, 0xc000010[87](https://github.com/dapr/components-contrib/actions/runs/6115916495/job/16600289608?pr=3112#step:9:88)0)
	/opt/hostedtoolcache/go/1.20.7/x64/src/testing/testing.go:1576 +0x10b
created by testing.(*T).Run
	/opt/hostedtoolcache/go/1.20.7/x64/src/testing/testing.go:1629 +0x3ea
	```

**Comment by berndverst** [comment]
Running the tests on Mac instead I got:

```
--- FAIL: TestOperations (0.00s)
    command_test.go:96:
        	Error Trace:	/Users/bverst/src/dapr/components-contrib/bindings/zeebe/command/command_test.go:96
        	Error:      	Not equal:
        	            	expected: 12
        	            	actual  : 13
        	Test:       	TestOperations
FAIL
FAIL	github.com/dapr/components-contrib/bindings/zeebe/command	0.794s
--- FAIL: TestInit (0.00s)
    --- FAIL: TestInit/jobType_is_mandatory (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x1542cac]
```

berndverst:
Please note that the contrib code freeze was technically weeks ago and this PR does change an existing component - not just tests. We can only include this PR if you can fix it up before we will be cutting the 1.12 release branch - which may be end of day tomorrow or on Monday.

bindings/zeebe/command/metadata.yaml
akkie: @ItalyPaleAle Sorry, my bad. The actual deploy-process command in master uses already the deployResource functionality under the hood. So an alias would be completely sufficient. So lets keep the alias for the moment and then remove it later as @berndverst suggested.

akkie:
@berndverst Fixed the issue

akkie:
Can someone please approve the workflows again. There was a test failing

berndverst (approved):

PR #3114, opened 2023-09-05
rm unused code

cicoyle:
Noticed in our yaml files that the `conformance.yml` workflow was referencing the `certification` test path, which is just confusing because that's the wrong test path. It's a copy pasta error. It should be removed to avoid confusion as well as since it's not even being used.


ItalyPaleAle (approved):
Thank you!

PR #3117, opened 2023-09-07
add zeebe command & jobworker yaml for conformance tests

cicoyle:
# Description

Adding zeebe command & jobworker component `bindings.yml` files to satisfy the conformance test requirement to move the components to `Beta` from `Alpha`.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/tests/config/bindings/zeebe/command/bindings.yml
+++ b/tests/config/bindings/zeebe/command/bindings.yml
@@ -1,0 +1,14 @@
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: zeebe-command
+spec:
+  type: bindings.zeebe.command
+version: v1
+  metadata:
+    - name: gatewayAddr
+      value: localhost:26500
+    - name: gatewayKeepAlive
+      value: 45s
+    - name: usePlainTextConnection
+      value: true
--- a/tests/config/bindings/zeebe/jobworker/bindings.yml
+++ b/tests/config/bindings/zeebe/jobworker/bindings.yml
@@ -1,0 +1,30 @@
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: zeebe-jobworker-test
+spec:
+  type: bindings.zeebe.jobworker
+  version: v1
+  metadata:
+    - name: gatewayAddr
+      value: localhost:26500
+    - name: gatewayKeepAlive
+      value: 45s
+    - name: usePlainTextConnection
+      value: true
+    - name: workerName
+      value: test-worker
+    - name: workerTimeout
+      value: 30s
+    - name: requestTimeout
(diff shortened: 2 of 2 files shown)


ItalyPaleAle:
Hi @cicoyle it's a bit more involved than that to enable a conformance test :)

In addition to creating the YAMLs there, you also need to:

1. Add the component to https://github.com/dapr/components-contrib/blob/master/tests/config/bindings/tests.yml
2. Load it in https://github.com/dapr/components-contrib/blob/0c2ce324b2bf174349c506d4644503075378f0eb/tests/conformance/common.go#L616
3. Update https://github.com/dapr/components-contrib/blob/master/.github/scripts/test-info.mjs so it runs in GitHub Actions

However, I don't think this component needs a conformance test, because it doesn't really have anything to conform to. The [bindings conformance test](https://github.com/dapr/components-contrib/blob/0c2ce324b2bf174349c506d4644503075378f0eb/tests/conformance/bindings/bindings.go#L113) tests the following operations: read, create, get, list, delete. This component implements neither. So there's not much to conform here - the test is currently only used by bindings that are more similar to pubsub or state stores: https://github.com/dapr/components-contrib/blob/master/tests/config/bindings/tests.yml

CC: @berndverst

berndverst:
@cicoyle I don't think we need this in conformance tests - we should just add it to certification tests. Bindings don't have a standard interface, so a conformance test for bindings is mostly meaningless.

@akkie recently started work on the certification test for this component - and that's really all that is required. Once we have the certification merged we can call this component stable - and I am comfortable skipping the beta stage entirely given this is a binding.

I agree with @ItalyPaleAle that we do not need a conformance test here - and that just unnecessarily takes up a Github task runner. The certification test that is in the works by @akkie will be sufficient. Ultimately stable certification is the real goal here.

Unfortunately I have to recommend that we close this PR @cicoyle

berndverst:
And adding to what @ItalyPaleAle said:
The missing pieces here are also actually spinning up the Zeebe service via Docker etc, and then defining the right Zeebe binding operations to invoke - and what the expected output should be.

I see you have also been involved in https://github.com/dapr/components-contrib/pull/3112 - which I just updated to actually activate the certification tests run.

I think we can skip this PR entirely - and enabling conformance tests for Zeebe would be quite a bit more work for you anyway!

berndverst:
Please note that the contrib code freeze was technically weeks ago and this PR does change an existing component - not just tests. We can only include this PR if you can fix it up before we will be cutting the 1.12 release branch - which may be end of day tomorrow or on Monday.

berndverst:
Ignore my other comment I deleted - I meant to comment on the other Zeebe issue.

ItalyPaleAle:
Closing - component now has a certification test which supersedes this

PR #3122, opened 2023-09-11
Metadata TryGetTTL: adds time.ParseDuration support

robertojrojas:
# Description

The [metadata/utils.go.TryGetTTL](https://github.com/dapr/components-contrib/blob/33a5fd20c0421744c15640fd3c74168935ae2202/metadata/utils.go#L53) doesn't appear to be parsing [time.Duration](https://pkg.go.dev/time#Duration) correctly.

Providing a value like: 
```yaml
    - name: ttlInSeconds
    - value: "20s"  
```
causes the function to return
```
"ttlInSeconds value must be a valid integer: actual is '20s'"
```
Based on my understanding, this function should be able to parse both valid integers and time.Duration values.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


metadata/utils.go
```diff
@@ -52,19 +52,24 @@ const (
...
-		valInt64, err := strconv.ParseInt(val, 10, 64)
+		duration, err := time.ParseDuration(val)
 		if err != nil {
-			return 0, false, fmt.Errorf("%s value must be a valid integer: actual is '%s'", TTLMetadataKey, val)
-		}
+			// failed to parse Duration string.
+			// let's try Integer and assume the value is in seconds
+			valInt64, err := strconv.ParseInt(val, 10, 64)
+			if err != nil {
+				return 0, false, fmt.Errorf("%s value must be a valid integer: actual is '%s'", TTLMetadataKey, val)
```
ItalyPaleAle, line 61:
```suggestion
				return 0, false, fmt.Errorf("%s value must be an integer or Go duration: actual is '%s'", TTLMetadataKey, val)
```

metadata/utils.go
```diff
@@ -52,19 +52,24 @@ const (
 // TryGetTTL tries to get the ttl as a time.Duration value for pubsub, binding and any other building block.
 func TryGetTTL(props map[string]string) (time.Duration, bool, error) {
 	if val, ok := props[TTLMetadataKey]; ok && val != "" {
-		valInt64, err := strconv.ParseInt(val, 10, 64)
+		duration, err := time.ParseDuration(val)
```
artursouza, line 55: We should have two different TTL attributes then. One "ttlInSeconds" and another one as "ttl" - this way, one is an integer and the other one is a duration. Components should agree and not mix them. Accepting unit in "ttlInSeconds"is not correct and the component should be changed. If it is stable, then it needs to follow a deprecation path.

artursouza (changes requested):

metadata/utils.go
```diff
@@ -52,19 +52,24 @@ const (
...
+			if valInt64 <= 0 {
+				return 0, false, fmt.Errorf("%s value must be higher than zero: actual is %d", TTLMetadataKey, valInt64)
+			}
 
-		duration := time.Duration(valInt64) * time.Second
-		if duration < 0 {
-			// Overflow
-			duration = math.MaxInt64
+			duration = time.Duration(valInt64) * time.Second
+			if duration < 0 {
```
ItalyPaleAle, line 69: Move this to outside of the `if err != nil {` block. Because the `duration` parsed by `time.ParseDuration` could be < 0 too

metadata/utils.go
```diff
@@ -52,19 +52,24 @@ const (
 // TryGetTTL tries to get the ttl as a time.Duration value for pubsub, binding and any other building block.
 func TryGetTTL(props map[string]string) (time.Duration, bool, error) {
 	if val, ok := props[TTLMetadataKey]; ok && val != "" {
```
ItalyPaleAle, line 54:
```suggestion
    val := props[TTLMetadataKey]
	if val != "" {
```

metadata/utils.go
berndverst: We should add `ttl` as an alias instead @artursouza. But that will also require an update to runtime.

berndverst:
@robertojrojas and I chatted about this before. This is a good change - but as we are beyond code freeze this will be merged in 1.13.

ItalyPaleAle (commented):
I think this is a good change.

However, I'd like to ask that you rename `ttlInSeconds` to `ttl` after this change. `ttlInSeconds` should remain as alias, but the main property should be `ttl` (and if you can, please make the parser case-insensitive: this is one of the few properties that are still case-sensitive)

metadata/utils.go
ItalyPaleAle: We were all writing the same thing.

I agree with @berndverst on aliasing `ttl` = `ttlInSeconds`. It's ok if `ttlInSeconds` accepts units too, it's not a breaking change.

metadata/utils_test.go
```diff
@@ -513,3 +515,80 @@ func TestResolveAliases(t *testing.T) {
...
+	tests := []struct {
+		name    string
+		md      map[string]string
+		result  any
+		wantOK  bool
+		wantErr bool
+		errStr  string
+	}{
+		{
+			name: "valid duration 20s",
```
ItalyPaleAle, line 529: Please add a test with a negative duration too, which should be converted to 0

berndverst:
FYI: Throughout contrib for every component metadata of type `duration` we support Duration strings, and also integers. If it's just an integer (string) then the value is interpreted as seconds. That exists primarily for backwards compatibility because duration strings are preferred instead. However, it's perfectly fine for only the duration string option to be documented.

Changing the TryGetTTL utility behavior here is consistent with the Duration behavior in Component Metadata parsing.

metadata/utils.go
artursouza: It is odd because you don't know the granularity of the client underneath. "ms" might not be supported, for example.

ItalyPaleAle:
@robertojrojas are you still working on this?

ItalyPaleAle (approved):

ItalyPaleAle (approved):

yaron2 (approved):

PR #3126, opened 2023-09-13
Minor Az Storage Binding improvements

berndverst:
# Description

Adds new response metadata property `pagesTraversed` for list blob operations that is purely informational. This is to provide confidence that Dapr indeed traversed all pages and no additional marker (for paging) is required.

Corrects the value returned by the `number` metadata property, which now sums up the blobs encountered across all pages. Previously this was only the blob count of the last page encountered.


bindings/azure/blobstorage/blobstorage.go
```diff
@@ -299,26 +301,30 @@ func (a *AzureBlobStorage) list(ctx context.Context, req *bindings.InvokeRequest
 	blobs := []*container.BlobItem{}
```
ItalyPaleAle, line 301: Just one small tweak please. On L300 (I can't leave a comment there), please replace:

```go
metadata := map[string]string{}
```

with:

```go
metadata := make(map[string]string, 3)
```

Since we're storing (up to) 3 keys, this improves the allocation a bit

berndverst: Making me change things unrelated to my PR I see :p

ItalyPaleAle: Not really. Before this was always empty. Now it includes 3 keys, so optimizing memory allocations is not necessarily unrelated :)

berndverst: It was not empty before!

ItalyPaleAle (approved):

ItalyPaleAle (approved):

PR #3134, opened 2023-09-15
Unify SQL DB migration subsystem (for Postgres and SQLite)

ItalyPaleAle:
This PR expands `internal/component/sql` to also add a unified subsystem to perform migrations for SQL databases. Currently, Postgres and SQLite are supported and have been migrated to that.

Related changes that are included:

- There is now an adapter used in `internal/component/sql` that works with both database/sql and pgx. This has been added to the garbage collector too, to simplify the code
- Some updates to `internal/component/postgresql` to remove the `dbAccess` layer which was adding complexity for no benefit (it didn't even help with unit tests)

Overall this PR moves around a bunch of code, but doesn't really change much how things work.


berndverst (changes requested):
Lint: Gofumpt

berndverst:
CockroachDB now fails @ItalyPaleAle

ItalyPaleAle:
@berndverst thanks. All issues (linter & tests) should be fixed now

berndverst (approved):

PR #3141, opened 2023-09-19
[Release 1.10 only] Fix Redis Binding Get Op for unknown keys

### LINKED ISSUE #3140: Redis:  Output Binding - Get operation does not return nil
## Output Binding Error

area runtime

## What version of Dapr?

CLI version: 1.10.0 
Runtime version: 1.10.2

## Expected Behavior

When using an Output Binding on a 'get' operation with Redis, return nil or null if the key does not exist.

## Actual Behavior

 Dapr throws an exception rather than returning nil.  The problem is that I have to look at the string message to determine if the value does not exist or if there was a network error, as the errors are very similar:

Value cannot be found error message:

== APP == io.grpc.StatusRuntimeException: INTERNAL: error invoking output binding account-database-binding: redis: nil

Redis is currently not running:

== APP == io.grpc.StatusRuntimeException: INTERNAL: error invoking output binding account-database-binding: read tcp [::1]:55697->[::1]:6377: i/o timeout

1.  When running an application that uses Output Binding with Redis, call the invokebinding command with the get operation and a key value that has not been added to the Redis database.

## Release Note

RELEASE NOTE:

berndverst:
# Description

The Redis SDK returns a special value `redis: nil` when a key does not exist. This was not handled correctly by the binding.

It was already fixed in Dapr 1.11, but not in Dapr 1.10.X

Fixes #3140

### DIFF
--- a/bindings/redis/redis.go
+++ b/bindings/redis/redis.go
@@ -82,6 +82,9 @@ func (r *Redis) Invoke(ctx context.Context, req *bindings.InvokeRequest) (*bindi
 		case bindings.GetOperation:
 			data, err := r.client.Get(ctx, key)
 			if err != nil {
+				if err.Error() == "redis: nil" {
+					return &bindings.InvokeResponse{}, nil
+				}
 				return nil, err
 			}
 			rep := &bindings.InvokeResponse{}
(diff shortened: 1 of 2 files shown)


berndverst:
Note that we refer to the string `redis: nil` vs a different constant because we use code that works with both the Redis v8 and v9 SDKs and using this string here is compatible with both SDKs. Of course this already works in Dapr 1.11.X and 1.12 RC / master. We are merely backporting a fix here.

yaron2 (approved):

Issue #3156, opened 2023-09-27
Error during Dapr Shutdown and Message Polling Behavior Issue (AWS)


tomatbay:
## Expected Behavior

I expect that as the number of subscriptions in Dapr increases, the following behavior should occur:

Dapr should only poll messages from subscriptions that have completed their registration process, ensuring no errors related to incomplete registration occur.
During shutdown, Dapr should gracefully stop polling messages and complete the shutdown process without errors.

## Actual Behavior

The actual behavior observed is as follows:

When Dapr initializes during startup and there are multiple subscriptions, it begins polling messages from the first subscription in the list. However, this behavior can include messages from subscriptions that may not have completed their registration process, leading to the occurrence of errors.

Occasionally, during shutdown, Dapr continues to attempt message polling even after the associated topics have been unregistered. This behavior leads to the same error related to message handling, and the error message is as follows:
```
error while handling received message. error is: handler for topic (sanitized):
```

## Steps to Reproduce the Problem

To replicate this problem, you can follow these steps:

1. Set up Dapr with multiple subscriptions.

2. Populate the SQS (Simple Queue Service) with a variety of messages.

3. Launch Dapr and monitor its startup behavior. Take note that Dapr might initiate message polling from subscriptions that haven't finished their registration, resulting in potential errors.

4. On occasion, during the shutdown process, observe Dapr's behavior as it may persist in attempting to poll messages even after the relevant topics have been unregistered. This persistence can trigger the aforementioned error message.

* Please note that the issue is intermittent and may not occur consistently during every run.

tomatbay:
@maintainers
could you please assign @amimimor for this bug

amimimor:
I'd like to highlight a potential additional issue with the current implementation of the subscription feature. Currently, the mapping between topics and handlers is managed by a synchronous standard library map, which is protected against race conditions using a shared lock instantiated as `RLock` for read operations. This occurs when messages come in from SQS and need to be dispatched to the appropriate handler. Simultaneously, during the creation of new subscriptions within the component, the shared lock is used as a `Lock` by writers to prevent stale reads.

[...]

PR #3174, opened 2023-10-12
FIx SQS: Error during Dapr Shutdown and Message Polling Behavior Issue (AWS)

amimimor:
# Description

In this PR, several significant changes have been made (please bear with me):

* Overhauled the `Subscribe` functionality, focusing on refining control constructs for categorizing messages by topic and ending short polling on SQS.
* Relocated the intricate poller cancelation logic to a new structure, `SubscriptionManager`.
* The `SubscriptionManager` oversees a control goroutine loop, monitoring subscribe/unsubscribe operations and context terminations.
* Those events are fanned-in to a channel the control loop consumes from. When a subscribe/unsubscribe event emerges, a lock is employed to determine if SQS operations should cease or if only handler removal is required. Upcoming PRs will allow these subscribe actions to govern the initiation of operations at boot, contingent on the availability of infrastructure components.
* Previously, the cancel function was derived from the initial subscription context and handed to the SQS poller. However, upon context cancel (unsubscribe) during shutdown, there was no guarantee this cancel function would be called last, thereby causing premature termination of consumption if called too early. This could have been prohibiting dynamic unsubscribes because if by chance a single subscription was  cancelled and unfortunately the consumption cancel function was 'pinned' to this topic, then the entire component consumption would be terminated until a new event of subscribe would be created
* Swapped the restrictive mutexes safeguarding the synchronous map with xsync.MapOf for lock-free thread-safe reads. Notably, this supports the `Size` method that is used to determine what/if actions like halting of SQS consumption should be taken.
* Introduced a straightforward Singleton Poller mechanism.
* Now, each `Subscribe` creates a listener unsubscribe goroutine improving readability and the overall sense the system makes. As simple as "a subscribe creates an unsubscribe listener watching for subscribe context cancellation and acting upon it"
* All termination procedures are now housed in the SubscriptionManager, while the primary component logic only manages its own status (closed or open) and invokes the manager's `Close` function.
* Post-refactoring, the remaining maps were transitioned to the thread-safe `sync.Map`.
* Standardized error messages to lowercase and restructured the frequently used "... topic (sanitized): abc" to "... (sanitized) topic: abc".
* Revamped getOrCreateTopic to align with linter standards, avoiding named return variables.
* Modernized consumeSubscription error handling using Errors.Is and Errors.As, and removed redundant waits on channel status or the termination of pollerRunning within the consumption process, simplifying understanding.

## Issue reference

Please reference the issue this PR will close: #3156 

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests
* [X] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#3820


ItalyPaleAle:
@amimimor looks like this PR is in a "broken" state where unrelated commits have been added to the branch. I'm not sure what happened.

Could you please try fixing it before we can review it?

pubsub/aws/snssqs/snssqs.go
```diff
@@ -251,40 +250,47 @@ func (s *snsSqs) getTopicArn(parentCtx context.Context, topic string) (string, e
 
 // get the topic ARN from the topics map. If it doesn't exist in the map, try to fetch it from AWS, if it doesn't exist
 // at all, issue a request to create the topic.
-func (s *snsSqs) getOrCreateTopic(ctx context.Context, topic string) (topicArn string, sanitizedName string, err error) {
-	s.topicsLock.Lock()
```
ItalyPaleAle, line 255: Although it may _look_ like using sync.Map is as good as a solution as the previous one that used locks, it isn't. 

Per the [official docs](https://pkg.go.dev/sync#Map):

> The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content. 

The TLDR is that sync.Map allows concurrent access to the map without issues, but it is not atomic, so in this case the code suffers from race conditions.

1. One goroutine starts and does `s.topicArns.Load`. It sees that the topic doesn't exist and starts creating one. After a delay, `s.topicsHandlers.Store` is invoked to store the topic.
2. If a separate goroutine is started at the same time, it may invoke `s.topicArns.Load` and see no topic, and then will create a new one and store it in `s.topicsHandlers.Store`.

That causes a race condition. Instead, using locks allows coordination in this case.

ItalyPaleAle (changes requested):

pubsub/aws/snssqs/snssqs.go
amimimor: ok, thanks. I'll think about maybe using xsync.MapOf (that I'm already using in the subscription manager) as I think it is lockless but safe (aka atomic). wdyt?

amimimor:
> @amimimor looks like this PR is in a "broken" state where unrelated commits have been added to the branch. I'm not sure what happened.
> 
> Could you please try fixing it before we can review it?

should be now aligned. I had colab with a teammate and forgot to squash (as he is not DCOed) so I had to rebase like 17 commits back and probably made a mess. git keeps biting me 17yrs in the business

pubsub/aws/snssqs/snssqs.go
ItalyPaleAle: No, it still wouldn't work. You need a lock that ensures coordination between multiple operations.

Imagine this were a SQL database. Using a lock is like using a transaction, so you can perform a SELECT first (reading from the map), do whatever processing you need to do (in this case, create a new connection if it doesn't exist), and then perform an INSERT to store the value.

Without a lock, it's like you don't use transactions in SQL databases. So even if there's no memory corruption because of sync.Map/xsync.MapOf, you still have the "double spending" problem.

ItalyPaleAle:
> git keeps biting me 17yrs in the business

I think most of us can relate 🤣

pubsub/aws/snssqs/snssqs.go
amimimor: First I want to apologize ahead of time on everything going around this PR as the situation in Israel is very hard and my level of focus is not at its best.

As you hinted with the SQL example, the problem here is transactional isolation.
So, here, there is a need for a call to the component `Subscribe` to behave as if it is atomic and avoid contentation with other calls for `Subscribe` for the same topic. If I'm correct, in that case, all mappings read/written during subscription creation (topic --> topicArn, queue --> queueInfo, topic --> subscriptionArn & the new call to the subs mgmt `Subscribe()`) should be isolated using a barrier (and an expensive one since there are many i/o ops especially if the component provisions the aws resources). 

I think that a single lock might be too brutal as the chances for same topic/queue/subscription contentation are rare cases to be locking the component out every time subscription ops are happening and we should think about a fine-grained lock-by-topic kind of locking. wdyt?

amimimor: @ItalyPaleAle ?

ItalyPaleAle: Sorry for the delay.

> First I want to apologize ahead of time on everything going around this PR as the situation in Israel is very hard and my level of focus is not at its best.

We are in no rush, and what's happening in Israel is terrible. I hope you and your loved ones are safe!

> I think that a single lock might be too brutal as the chances for same topic/queue/subscription contentation are rare cases to be locking the component out every time subscription ops are happening and we should think about a fine-grained lock-by-topic kind of locking. wdyt?

Good idea. I think this should work!

go.mod
```diff
@@ -324,6 +324,7 @@ require (
 	github.com/prometheus/common v0.42.0 // indirect
 	github.com/prometheus/procfs v0.11.0 // indirect
 	github.com/prometheus/statsd_exporter v0.22.7 // indirect
+	github.com/puzpuzpuz/xsync/v2 v2.5.1 // indirect
```
ItalyPaleAle, line 327: Please run `make modtidy-all`

pubsub/aws/snssqs/topics_locker.go
```diff
@@ -0,0 +1,48 @@
...
+
+var (
+	lockManager     *TopicsLockManager
+	lockManagerOnce sync.Once
+)
+
+// TopicsLockManager is a singleton for fine-grained locking, to prevent the component r/w operations
+// from locking the entire component out when performing operations on different topics.
+type TopicsLockManager struct {
+	xLockMap *xsync.MapOf[string, *sync.Mutex]
```
ItalyPaleAle, line 17: Why using xsync rather than just a sync.Map here?

pubsub/aws/snssqs/snssqs.go
```diff
@@ -243,48 +228,59 @@ func (s *snsSqs) getTopicArn(parentCtx context.Context, topic string) (string, e
...
-	defer s.topicsLock.Unlock()
+func (s *snsSqs) getOrCreateTopic(ctx context.Context, topic string) (string, string, error) {
+	var (
+		topicArn       string
+		loadOK         bool
+		sanitizedTopic string
+		err            error
+	)
```
ItalyPaleAle, line 245: I do really like named returns when there's > 2 returned values and have the smae type. The signature of the method should be clear as to what values are returned.

There's no difference in adding the names or initializing the variables right after, but the former approach is clearer for readers.

(Named returns don't require "naked returns", you can still return the variables by name)

```suggestion
func (s *snsSqs) getOrCreateTopic(ctx context.Context, topic string) (topicArn string, sanitizedTopic string, err error) {
	var loadOK bool
```

pubsub/aws/snssqs/snssqs.go
```diff
@@ -243,48 +228,59 @@ func (s *snsSqs) getTopicArn(parentCtx context.Context, topic string) (string, e
...
+		if len(topicArn) > 0 {
+			s.logger.Debugf("found existing topic ARN for topic %s: %s", topic, topicArn)
 
-	topicArnCached, ok := s.topicArns[sanitizedName]
-	if ok && topicArnCached != "" {
-		s.logger.Debugf("found existing topic ARN for topic %s: %s", topic, topicArnCached)
-		return topicArnCached, sanitizedName, nil
+			return topicArn, sanitizedTopic, err
+		} else {
+			err = fmt.Errorf("ARN for (sanitized) topic: %s was empty", sanitizedTopic)
```
ItalyPaleAle, line 255: Linter doesn't like when errors start with uppercase letter

pubsub/aws/snssqs/snssqs.go
```diff
@@ -243,48 +228,59 @@ func (s *snsSqs) getTopicArn(parentCtx context.Context, topic string) (string, e
...
+
+			return topicArn, sanitizedTopic, err
+		}
 	}
+
 	// creating queues is idempotent, the names serve as unique keys among a given region.
-	s.logger.Debugf("No SNS topic arn found for %s\nCreating SNS topic", topic)
+	s.logger.Debugf("no SNS topic ARN found for topic: %s. creating SNS with (sanitized) topic: %s", topic, sanitizedTopic)
```
ItalyPaleAle, line 262:
```suggestion
	s.logger.Debugf("No SNS topic ARN found for topic: %s. creating SNS with (sanitized) topic: %s", topic, sanitizedTopic)
```

pubsub/aws/snssqs/snssqs.go
```diff
@@ -243,48 +228,59 @@ func (s *snsSqs) getTopicArn(parentCtx context.Context, topic string) (string, e
...
+		sanitizedTopic string
+		err            error
+	)
+
+	sanitizedTopic = nameToAWSSanitizedName(topic, s.metadata.Fifo)
 
-	sanitizedName = nameToAWSSanitizedName(topic, s.metadata.Fifo)
+	if topicArn, loadOK = s.topicArns[sanitizedTopic]; loadOK {
+		if len(topicArn) > 0 {
+			s.logger.Debugf("found existing topic ARN for topic %s: %s", topic, topicArn)
```
ItalyPaleAle, line 251:
```suggestion
			s.logger.Debugf("Found existing topic ARN for topic %s: %s", topic, topicArn)
```

pubsub/aws/snssqs/snssqs.go
```diff
@@ -601,12 +606,13 @@ func (s *snsSqs) consumeSubscription(ctx context.Context, queueInfo, deadLetters
 		// iteration. Therefore, a global backoff (to the internal backoff) is used (sqsPullExponentialBackoff).
 		messageResponse, err := s.sqsClient.ReceiveMessageWithContext(ctx, receiveMessageInput)
 		if err != nil {
-			if err == context.Canceled || err == context.DeadlineExceeded {
+			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) || ctx.Err() != nil {
```
ItalyPaleAle, line 609:
```suggestion
			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
```

pubsub/aws/snssqs/snssqs.go
```diff
@@ -824,65 +828,18 @@ func (s *snsSqs) Subscribe(ctx context.Context, req pubsub.SubscribeRequest, han
...
+	s.subscriptionManager.Subscribe(&SubscriptionTopicHandler{
+		topic:        sanitizedName,
+		requestTopic: req.Topic,
+		handler:      handler,
+		ctx:          ctx,
+	})
 
 	return nil
+
 }
```
ItalyPaleAle, line 843:
```suggestion
	return nil
}
```

pubsub/aws/snssqs/topics_locker.go
ItalyPaleAle: Also note v2 of the package is deprecated: https://github.com/puzpuzpuz/xsync

pubsub/aws/snssqs/subscription_mgmt.go
```diff
@@ -0,0 +1,192 @@
...
+	SubscriptionAction int
+)
+
+const (
+	Subscribe SubscriptionAction = iota
+	Unsubscribe
+)
+
+var (
+	subscriptionMgmtInst *SubscriptionManager
```
ItalyPaleAle, line 23: No singletons please. Each component should be completely self-contained, as one could instantiate multiple components.

pubsub/aws/snssqs/subscription_mgmt.go
```diff
@@ -0,0 +1,192 @@
...
+		case changeEvent := <-sm.topicsChangeCh:
+			topic := changeEvent.handler.topic
+			sm.logger.Debugf("subscription change event received with action: %v, on topic: %s", changeEvent.action, topic)
+			// topic change events are serialized so that no interleaving can occur
+			sm.lock.Lock()
+			// although we have a lock here, the topicsHandlers map is thread safe and can be accessed concurrently so other subscribers that are already consuming messages
+			// can get the handler for the topic while we're still updating the map without blocking them
+			current := sm.topicsHandlers.Size()
+
+			if changeEvent.action == Subscribe {
```
ItalyPaleAle, line 111: Please use a `switch`

pubsub/aws/snssqs/topics_locker.go
amimimor: > Why using xsync rather than just a sync.Map here?

xsync supports the `LoadOrCompute` and `Compute` methods which are crucial here

> Also note v2 of the package is deprecated: https://github.com/puzpuzpuz/xsync

v3.0.0 was released 3 days ago I don't know if it is as stable as the semver suggests

pubsub/aws/snssqs/snssqs.go
amimimor: weird I'm using the project' golangci-lint config and not getting this as error or something. personally, as the log is JSON formatted, using Uppercase doesn't make any sense because the message is an entry in that json, unlike classical access-log entries

pubsub/aws/snssqs/snssqs.go
amimimor: why? the context might be cancelled while we pull messages but still haven't started a new for loop iteration (that verifies the ctx isn't cancelled, but just at the top of the loop)

pubsub/aws/snssqs/subscription_mgmt.go
amimimor: cool. maybe I'm too defensive

pubsub/aws/snssqs/topics_locker.go
ItalyPaleAle: > v3.0.0 was released 3 days ago I don't know if it is as stable as the semver suggests

We have some time. SInce v2 is unmaintained now, let's use the new version please!

pubsub/aws/snssqs/subscription_mgmt.go
```diff
@@ -0,0 +1,180 @@
...
+type (
+	SubscriptionAction int
+)
+
+const (
+	Subscribe SubscriptionAction = iota
+	Unsubscribe
+)
+
+var initOnce sync.Once
```
ItalyPaleAle, line 22: I think this should be a property of the SubscriptionManger struct? Otherwise, if you have 2 components of the same type, one of them will not have the initOnce callback executed

ItalyPaleAle (approved):
LGTM, thanks!

amimimor:
thanks @ItalyPaleAle ! it wasn't an easy one and your comments had been very helpful. I'm now going to rush into adding documentation about the edge cases I came across with sns/sqs so that users could know what to expect. See you (maybe) in the coming PRs 🥳

ItalyPaleAle:
> thanks @ItalyPaleAle ! it wasn't an easy one and your comments had been very helpful. I'm now going to rush into adding documentation about the edge cases I came across with sns/sqs so that users could know what to expect. See you (maybe) in the coming PRs 🥳

Of course, thanks for the contribution and for your patience while we got this reviwwed :) Looking forward to see more from you!

PR #3178, opened 2023-10-13
Local file-based name resolver with SQLite

ItalyPaleAle:
Fixes #1380
Will fix (when runtime PR is merged) dapr/dapr#3256

Addresses a longstanding request to have an alternative name resolver component for local development, for situations where mDNS can't be used (and even in those situations, a local name resolver is more convenient for local development).

As discussed in #1380, this is implemented with SQLite, which is fast and efficient.

PR includes tests implemented via unit tests, since we don't have an infrastructure for writing conformance or certification tests for name resolvers.


ItalyPaleAle:
All cert tests are failing because the APIs changed and the pinned runtime needs to be updated. Usual problem :) However nothing changed here that would impact cert tests

nameresolution/metadata.go
```diff
@@ -33,3 +37,41 @@ type Metadata struct {
...
+	p, _ := strconv.Atoi(m.Properties[DaprHTTPPort])
+	if p < 0 {
+		p = 0
+	}
+	return p
+}
+
+// GetDaprPort returns the host address property.
+func (m Metadata) GetDaprPort() int {
+	p, _ := strconv.Atoi(m.Properties[DaprPort])
```
JoshVanL, line 58: Shouldn’t we error in these cases and fail the unit of the component? I’d expect an internal type which containing the parsed metadata with the target types.

nameresolution/sqlite/sqlite.go
```diff
@@ -0,0 +1,301 @@
...
+
+	connString, err := s.metadata.GetConnectionString(s.logger)
+	if err != nil {
+		// Already logged
+		return err
+	}
+
+	// Show a warning if SQLite is configured with an in-memory DB
+	if s.metadata.SqliteAuthMetadata.IsMemory() {
+		s.logger.Warn("Configuring name resolution with an in-memory SQLite database. Service invocation across differet apps will not work.")
```
JoshVanL, line 71:
```suggestion
		s.logger.Warn("Configuring name resolution with an in-memory SQLite database. Service invocation across different apps will not work.")
```

nameresolution/sqlite/sqlite.go
```diff
@@ -0,0 +1,301 @@
...
+		return err
+	}
+
+	// Register the host and update in background
+	err = s.registerHost(ctx)
+	if err != nil {
+		return err
+	}
+
+	go s.renewRegistration()
```
JoshVanL, line 102: We should have a wait group of similar here to ensure all resources are released  before returning from component close.

nameresolution/sqlite/sqlite.go
```diff
@@ -0,0 +1,301 @@
...
+	for {
+		select {
+		case <-s.closeCh:
+			// Component is closing
+			s.logger.Debug("Stopped renewing host registration: component is closing")
+			return
+
+		case <-t.C:
+			// Renew on the ticker
+			err := s.doRenewRegistration(context.Background(), addr)
```
JoshVanL, line 174: We should tie this context to `closeCh` else we could hang the system on close when request timeout is high.

nameresolution/sqlite/sqlite_migrations.go
```diff
@@ -0,0 +1,64 @@
...
+		// Migration 0: create the hosts table
+		func(ctx context.Context) error {
+			logger.Infof("Creating hosts table '%s'", opts.HostsTableName)
+			_, err := m.GetConn().ExecContext(
+				ctx,
+				fmt.Sprintf(
+					`CREATE TABLE %[1]s (
+						registration_id TEXT NOT NULL PRIMARY KEY,
+						address TEXT NOT NULL,
+						app_id TEXT NOT NULL,
```
JoshVanL, line 50: We also need to include namespace

nameresolution/sqlite/sqlite.go
```diff
@@ -0,0 +1,301 @@
...
+		// See: https://stackoverflow.com/a/24591696
+		`SELECT address
+		FROM %[1]s
+		WHERE
+			ROWID = (
+				SELECT ROWID
+				FROM %[1]s
+				WHERE
+					app_id = ?
+					AND unixepoch(CURRENT_TIMESTAMP) - last_update < %[2]d
```
JoshVanL, line 227: Add namespace here too.

nameresolution/sqlite/sqlite.go
```diff
@@ -0,0 +1,301 @@
...
+
+func (s *resolver) initGC() (err error) {
+	s.gc, err = internalsql.ScheduleGarbageCollector(internalsql.GCOptions{
+		Logger: s.logger,
+		UpdateLastCleanupQuery: func(arg any) (string, any) {
+			return fmt.Sprintf(`INSERT INTO %s (key, value)
+				VALUES ('nr-last-cleanup', CURRENT_TIMESTAMP)
+				ON CONFLICT (key)
+				DO UPDATE SET value = CURRENT_TIMESTAMP
+					WHERE (unixepoch(CURRENT_TIMESTAMP) - unixepoch(value)) * 1000 > ?;`,
```
JoshVanL, line 115: This means records have a TTL of 1000s in the db?

nameresolution/sqlite/sqlite.go
```diff
@@ -0,0 +1,301 @@
...
+	defer queryCancel()
+
+	res, err := s.db.ExecContext(queryCtx,
+		fmt.Sprintf("UPDATE %s SET last_update = unixepoch(CURRENT_TIMESTAMP) WHERE registration_id = ? AND address = ?", s.metadata.TableName),
+		s.registrationID, addr,
+	)
+
+	if err != nil {
+		return fmt.Errorf("database error: %w", err)
+	} else {
```
JoshVanL, line 201:
```suggestion
	}
```

nameresolution/sqlite/sqlite.go
```diff
@@ -0,0 +1,301 @@
...
+	if s.registrationID == "" {
+		// We never registered
+		return nil
+	}
+
+	queryCtx, queryCancel := context.WithTimeout(ctx, s.metadata.Timeout)
+	defer queryCancel()
+
+	res, err := s.db.ExecContext(queryCtx,
+		fmt.Sprintf("DELETE FROM %s WHERE registration_id = ? AND address = ?", s.metadata.TableName),
```
JoshVanL, line 257: Namespace

nameresolution/sqlite/sqlite.go
```diff
@@ -0,0 +1,301 @@
...
+		s.registrationID, s.metadata.GetAddress(),
+	)
+	if err != nil {
+		return fmt.Errorf("failed to unregister host: %w", err)
+	}
+	n, _ := res.RowsAffected()
+	if n == 0 {
+		return fmt.Errorf("failed to unregister host: no row deleted")
```
JoshVanL, line 265: nit

```suggestion
		return errors.New("failed to unregister host: no row deleted")
```

JoshVanL (changes requested):
Some comments from me

nameresolution/metadata.go
ItalyPaleAle: The components error later because the port is 0. Also worth pointing out that this value is set by the runtime, so it shouldn't really happen that this be invalid, in practice.

nameresolution/sqlite/sqlite.go
ItalyPaleAle: Will do

nameresolution/sqlite/sqlite.go
ItalyPaleAle: Will do

nameresolution/sqlite/sqlite_migrations.go
ItalyPaleAle: Will do

nameresolution/sqlite/sqlite.go
ItalyPaleAle: Ohhh that's a copy-paste mistake. Good catch.

```suggestion
					WHERE unixepoch(CURRENT_TIMESTAMP) - unixepoch(value)  > ?;`,
```

nameresolution/sqlite/sqlite.go
ItalyPaleAle: Will do

nameresolution/sqlite/sqlite.go
```diff
@@ -0,0 +1,301 @@
...
+	)
+
+	if err != nil {
+		return fmt.Errorf("database error: %w", err)
+	} else {
+		n, _ := res.RowsAffected()
+		if n == 0 {
+			return errRegistrationLost
+		}
+	}
```
ItalyPaleAle, line 206:
```suggestion
	}

	n, _ := res.RowsAffected()
	if n == 0 {
		return errRegistrationLost
	}
```

nameresolution/sqlite/sqlite_migrations.go
ItalyPaleAle: I just looked at the other components and it seems that no nameresolver besides the K8s supports the namespace. mDNS and Consul don't.

I think it should still be added to this component but it would require a lot of testing in the runtime.

ItalyPaleAle: I just looked at the other components and it seems that no nameresolver besides the K8s supports the namespace. mDNS and Consul don't.

It doesn't seem to be possible to implement namespacing support at this stage. The runtime itself doesn't pass the namespace to the nameresolvers (none of them) when they are registered.

If namespacing support is required in these components, it should be done carefully and in a separate PR, as it is a significant change for the runtime too.

ItalyPaleAle: I have created a separate issue to add namespacing support: https://github.com/dapr/components-contrib/issues/3179

It's not something we can do this in PR as it requires larger changes in the runtime too. 

However, in this PR I've added some "infra" already, including the constant that will then need to be used in the runtime.

nameresolution/consul/consul_test.go
```diff
@@ -184,25 +184,19 @@ func (m *mockRegistry) get(service string) *registryEntry {
 }
 
 func TestInit(t *testing.T) {
-	t.Parallel()
```
JoshVanL, line 187: Why are we removing `Parallel`s here?

nameresolution/sqlite/sqlite.go
```diff
@@ -0,0 +1,317 @@
...
+}
+
+// In backgrounds, periodically renews the host's registration
+// Should be invoked in a background goroutine
+func (s *resolver) renewRegistration() {
+	defer s.wg.Done()
+
+	addr := s.metadata.GetAddress()
+
+	d := s.metadata.UpdateInterval - s.metadata.Timeout
```
JoshVanL, line 170: We are we subtracting the Timeout here?

nameresolution/sqlite/sqlite.go
```diff
@@ -0,0 +1,317 @@
...
+		select {
+		case <-s.closeCh:
+			// Component is closing
+			s.logger.Debug("Stopped renewing host registration: component is closing")
+			return
+
+		case <-t.C:
+			// Renew on the ticker
+			go func() {
+				err := s.doRenewRegistration(ctx, addr)
```
JoshVanL, line 188:
```suggestion
			// Renew on the ticker
			s.wg.Add(1)
			go func() {
			  defer a.wg.Done()
				err := s.doRenewRegistration(ctx, addr)
```

nameresolution/sqlite/sqlite.go
```diff
@@ -0,0 +1,317 @@
...
+				if err != nil {
+					// Log errors
+					s.logger.Errorf("Failed to update host registration: %v", err)
+
+					if errors.Is(err, errRegistrationLost) {
+						// This means that our registration has been taken over by another host
+						// It should never happen unless there's something really bad going on
+						// Panicking here to force a restart of Dapr
+						s.logger.Fatalf("Host registration lost")
```
JoshVanL, line 197: Please don't panic. All active connections and resources will not be given the time to correctly gracefully close. We should shutdown the component instead.

nameresolution/sqlite/sqlite.go
```diff
@@ -0,0 +1,317 @@
...
+					}
+				}
+			}()
+		}
+	}
+}
+
+func (s *resolver) doRenewRegistration(ctx context.Context, addr string) error {
+	queryCtx, queryCancel := context.WithTimeout(context.Background(), s.metadata.Timeout)
```
JoshVanL, line 206:
```suggestion
	queryCtx, queryCancel := context.WithTimeout(ctx, s.metadata.Timeout)
```

nameresolution/sqlite/sqlite.go
```diff
@@ -0,0 +1,317 @@
...
+
+	res, err := s.db.ExecContext(queryCtx,
+		fmt.Sprintf("UPDATE %s SET last_update = unixepoch(CURRENT_TIMESTAMP) WHERE registration_id = ? AND address = ?", s.metadata.TableName),
+		s.registrationID, addr,
+	)
+	if err != nil {
+		return fmt.Errorf("database error: %w", err)
+	}
+
+	n, _ := res.RowsAffected()
```
JoshVanL, line 217: Check error, and return error message if it not nil.

JoshVanL (changes requested):

nameresolution/consul/consul_test.go
ItalyPaleAle: `Parallel` makes it harder to read test results because the logs are full of `CONT`. As a helper, it makes sense when the tests are long-running and perhaps I/O bound, but these tests complete in nanoseconds anyways. We don't gain anything by making them parallel and we make reading test results harder.

nameresolution/sqlite/sqlite.go
ItalyPaleAle: The status must be updated every `UpdateInterval`, let's say 10s. If the table hasn't been updated in 10s, callers querying it will consider the record expired. However, queries are given a `Timeout`, for example 1s. So if we were to just loop every `UpdateInterval`, and the query did take 1s, it would be updated in 11s.

There's a check in the metadata struct that ensures that `UpdateInterval > Timeout`.

nameresolution/sqlite/sqlite.go
ItalyPaleAle: Ok

nameresolution/sqlite/sqlite.go
ItalyPaleAle: I think that's redundant. If `err != nil`, then `n == 0` as a consequence. It will error anyways

nameresolution/sqlite/sqlite.go
ItalyPaleAle: Looking into it. I am not sure what the best way to cascade this error is in this case, since this runs on a background goroutine that has no way to communicate the error to a caller.

I am concerned that just calling `Close()` on the component could put the sidecar in an un-healthy state where all service invocation wouldn't work, but the sidecar would still be reported as healthy and so there'd be no way for the platform (like K8s or Docker) to detect the failure and restart.

Do you have any suggestion?

nameresolution/sqlite/sqlite.go
JoshVanL: Should we not have instances consider a record to be stale after a longer period than the `UpdateInterval`? Having the update interval the same as the TTL of the record means that there is no slack in the time windows and inherently leads to hosts being thrashed from the registry.

ItalyPaleAle: I'm not sure I understand, sorry.

Currently we consider as stale after `UpdateInterval`, but we update the record every `UpdateInterval - Timeout`. (I am thinking of adding a `-1` there too as well). So there is some natural slack there?

nameresolution/sqlite/sqlite.go
JoshVanL: Ideally, since there is always at most 1 name resolver we probably should have a single long running `Run` style function so that we can correctly propagate these errors back to runtime and shutdown appropriately. Are we in a position to change the nameresolver interface given that it is only imported by runtime and not a component, or do we need to do "v2"? Conversely, we can also create a new long running `healthz` type function which does the same.

I agree with you that since the name resolver is a special case "component", we need to propagate the error back and not just `Close`.

nameresolution/sqlite/sqlite.go
JoshVanL: If we consider `Timeout` to be a very small number (basically 0), then there is a race condition whereby one instance reads and one writes _just after_ `UpdateInterval`. If the read was first then the reader would consider the host record to be stale. We can alleviate this issue by considering a record stale when it is `1.3*` past the `UpdateInterval`.

nameresolution/sqlite/sqlite.go
ItalyPaleAle: > Are we in a position to change the nameresolver interface given that it is only imported by runtime and not a component, or do we need to do "v2"?

Yes, the APIs of components-contrib are not stable (meaning, we don't have a stable contract with third-party users).

Of course, if this is what we want to do, it's something that should not be done in this PR :)

> Conversely, we can also create a new long running healthz type function which does the same.

I am not sure if this issue with possible conflicts would be unique to this component. I need to look at the other implementations.

> I agree with you that since the name resolver is a special case "component", we need to propagate the error back and not just Close.

Let's resolve this for now and accept the panic as a temporary workaround? We can address the issue more properly with API changes in the runtime in a separate PR.

nameresolution/sqlite/sqlite.go
ItalyPaleAle: Adding a `-1` to the timeout should be enough to solve this (and making sure that `UpdateInterval > Timeout + 1`). There shouldn't be race conditions in that case, since the timeout controls the query timeout too.

ItalyPaleAle: Updated

nameresolution/sqlite/sqlite.go
```diff
@@ -0,0 +1,328 @@
...
+func (s *resolver) doRenewRegistration(ctx context.Context, addr string) error {
+	// We retry this query in case of database error, up to the timeout
+	queryCtx, queryCancel := context.WithTimeout(ctx, s.metadata.Timeout)
+	defer queryCancel()
+
+	// We use string formatting here for the table name only
+	//nolint:gosec
+	query := fmt.Sprintf("UPDATE %s SET last_update = unixepoch(CURRENT_TIMESTAMP) WHERE registration_id = ? AND address = ?", s.metadata.TableName)
+
+	b := backoff.WithContext(backoff.NewConstantBackOff(50*time.Millisecond), queryCtx)
```
DeepanshuA, line 219: Please consider defining it as a constant.

nameresolution/sqlite/sqlite.go
```diff
@@ -0,0 +1,328 @@
...
+	return &resolver{
+		logger:  logger,
+		closeCh: make(chan struct{}),
+	}
+}
+
+// Init initializes the name resolver.
+func (s *resolver) Init(ctx context.Context, md nameresolution.Metadata) error {
+	if s.closed.Load() {
+		return errors.New("component is closed")
```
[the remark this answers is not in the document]
ItalyPaleAle: All logs already include the name of the component that emits them, so this is not necessary

internal/authentication/sqlite/metadata.go
```diff
@@ -60,10 +63,14 @@ func (m *SqliteAuthMetadata) Validate() error {
 	return nil
 }
 
+func (m SqliteAuthMetadata) IsMemory() bool {
```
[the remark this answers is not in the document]
ItalyPaleAle: Done (both)

nameresolution/sqlite/sqlite.go
```diff
@@ -0,0 +1,328 @@
...
+		return err
+	}
+
+	connString, err := s.metadata.GetConnectionString(s.logger)
+	if err != nil {
+		// Already logged
+		return err
+	}
+
+	// Show a warning if SQLite is configured with an in-memory DB
```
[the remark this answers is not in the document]
ItalyPaleAle: I would not disable it outright, although yes, it kind of defeats the point of svc invocation across multiple apps (but can be used to test an app invoking itself). We also use that for unit testing.

nameresolution/mdns/mdns.go
```diff
@@ -269,34 +269,23 @@ func (m *Resolver) startRefreshers() {
...
 	}
-
-	if props[nameresolution.DaprPort] == "" {
-		return errors.New("port is missing")
-	}
-
-	port, err := strconv.Atoi(props[nameresolution.DaprPort])
-	if err != nil {
-		return errors.New("port is invalid")
+	if metadata.Instance.DaprInternalPort == 0 {
```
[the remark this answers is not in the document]
ItalyPaleAle: Done

berndverst:
If these new unit tests are slow we might want to move them to be some sort of pseudo certification tests.

berndverst (approved):

PR #3188, opened 2023-10-20
Provide OPA SDK integration option

anderseknert:
# Description

This PR introduces an alternative way to provide policy to the OPA middleware for Dapr, where rather than providing a single Rego file inlined in the manifest for the middleware, one may choose to instead provide an OPA configuration file, and use that to start an instance of the full OPA service, as provided by the OPA [Go SDK](https://www.openpolicyagent.org/docs/latest/integration/#integrating-with-the-go-sdk).

This allows decoupling policy management from enforcement, as the user is then allowed to point the OPA middleware at the same remote bundle endpoints where they'd normally fetch their policy and data. Fetching [policy from bundles](https://www.openpolicyagent.org/docs/latest/management-bundles/) also allows the user to have any number of policy files included in the decision rather than a single one.

The drawback is obviously that the service needs to run as a continuous process. Allowing the user to choose which model works best for them seems like a good option to me.

This is not a complete PR, but before I start fleshing out on things like documentation, I wanted to bounce this off of the Dapr maintainers for some feedback.

Would it be preferable to have the OPA configuration inlined in the config for the middleware? The current approach is to simply point to a file, but that obviously requires something like a configmap to have been mounted into the container. The drawback of inlining the config would be that it'll be "config in config" i.e. one YAML or JSON document embedded into another.

Some feedback much appreciated before I either proceed here, or drop this. Thanks!

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/middleware/http/opa/middleware.go
+++ b/middleware/http/opa/middleware.go
@@ -23,12 +23,14 @@ import (
 	"math"
 	"net/http"
 	"net/textproto"
+	"os"
 	"reflect"
 	"strconv"
 	"strings"
 	"time"
 
 	"github.com/open-policy-agent/opa/rego"
+	"github.com/open-policy-agent/opa/sdk"
 	"k8s.io/utils/strings/slices"
 
 	"github.com/dapr/components-contrib/internal/httputils"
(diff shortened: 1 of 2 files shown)


yaron2:
Thanks for this PR @anderseknert.

We have components today that support both file paths and inline configuration, so supporting an inline OPA policy should be perfectly acceptable.

anderseknert:
Thanks @yaron2 👍 Inlined policy is the only option currently available, so it's good to know a filepath is considered OK as well. Would this be a welcome addition to the OPA middleware? Notably this does not change the behavior for anyone currently using it, but would just add another option for how the OPA middleware should operate.

I started looking into this after an OPA user wanted to run a POC using Dapr with our control plane (Styra DAS), OPA is of course agnostic to what's on the other side, and this works equally well with the "control plane" being nginx, an S3 bucket or a custom HTTP server.

yaron2:
> Thanks @yaron2 👍 Inlined policy is the only option currently available, so it's good to know a filepath is considered OK as well. Would this be a welcome addition to the OPA middleware? Notably this does not change the behavior for anyone currently using it, but would just add another option for how the OPA middleware should operate.
> 
> I started looking into this after an OPA user wanted to run a POC using Dapr with our control plane (Styra DAS), OPA is of course agnostic to what's on the other side, and this works equally well with the "control plane" being nginx, an S3 bucket or a custom HTTP server.

I misread, thought you meant the other way around. But it's all the same, and both are viable.

yaron2:
If this retains backward compatibility then it's even better. So yes, another option for file based OPA is definitely acceptable

anderseknert:
Thanks @yaron2 👍 Yes, 100% backwards compatible and I'll make sure there are tests to verify that.
I'll switch this PR over to draft status while I flesh out docs and whatever else is needed for a merge. If you'd prefer I close it and come back with a new one when it's RFR, that works for me too. The timeframe of getting it to that stage is end of next week.

yaron2:
Fine to keep it in draft for this timeframe

anderseknert:
Hey folks! I was out sick for most of last week, but I'm getting back to this now. Sorry for the delay!

anderseknert:
Hey folks! Sorry for the delay here. I've been out with bronchitis for the last few weeks, and I've still not recovered fully. I'll come back to this once I do.

yaron2:
Hi @anderseknert would love to see this picked up again.

anderseknert:
Hi there @yaron2! Then I'll get to it :) But I won't be able to until after KubeCon.

yaron2:
> Hi there @yaron2! Then I'll get to it :) But I won't be able to until after KubeCon.

No worries, thanks. Also drop by and say Hi at KubeCon :)

PR #3197, opened 2023-10-27
pubsub/rabbitmq: return error if reconnect errors in Init

JoshVanL:
We should always return the error if we fail to connection during the Init of a Component. This is because failing to do so has downstream effects whereby a component is considered healthy when it is not.

Re-attempting an Init or otherwise should be handled by the runtime, rather than on a per-component basis.

### DIFF
--- a/pubsub/rabbitmq/rabbitmq.go
+++ b/pubsub/rabbitmq/rabbitmq.go
@@ -154,10 +154,10 @@ func (r *rabbitMQ) Init(_ context.Context, metadata pubsub.Metadata) error {
 
 	r.metadata = meta
 
-	r.reconnect(0)
-	// We do not return error on reconnect because it can cause problems if init() happens
-	// right at the restart window for service. So, we try it now but there is logic in the
-	// code to reconnect as many times as needed.
+	if err := r.reconnect(0); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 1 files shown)


pubsub/rabbitmq/rabbitmq.go
```diff
@@ -154,10 +154,10 @@ func (r *rabbitMQ) Init(_ context.Context, metadata pubsub.Metadata) error {
 
 	r.metadata = meta
 
-	r.reconnect(0)
-	// We do not return error on reconnect because it can cause problems if init() happens
```
ItalyPaleAle, line 158: There’s a comment here that explains why it was done this way. Would changing it cause issues?

berndverst:
I already merged this commit into master. https://github.com/dapr/components-contrib/pull/3198

berndverst:
You can see there are no files changed here now. So we can close this PR.

pubsub/rabbitmq/rabbitmq.go
berndverst: I think the only issue is that the component would fail to initialize and the sidecar would crash, then the next time it would initialize correctly.

IMO that is desirable behavior.

I think the original code was actually hiding some potential bugs.

PR #3198, opened 2023-10-27
pubsub/rabbitmq: return error if reconnect errors in Init

JoshVanL:
We should always return the error if we fail to connection during the Init of a Component. This is because failing to do so has downstream effects whereby a component is considered healthy when it is not.

Re-attempting an Init or otherwise whould be handled by the runtime, rather than on a per-component basis.


berndverst (approved):

berndverst:
oh.. now I see what happened here. Two PRs from the same branch.

PR #3199, opened 2023-10-27
[1.12] pubsub/rabbitmq: return error if reconnect errors in Init

JoshVanL:
Backport of https://github.com/dapr/components-contrib/pull/3198


berndverst (changes requested):
Please clean up this PR. There should only be a single commit included here. Somehow you created a very messy commit history here.

berndverst (approved):

PR #3201, opened 2023-10-30
postgres: add column name to the result of query function.

Jack397419:
add column name to the result of function “query”.

# Description

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


bindings/postgres/postgres.go
```diff
@@ -186,13 +186,21 @@ func (p *Postgres) query(ctx context.Context, sql string, args ...any) (result [
 		return nil, fmt.Errorf("error executing query: %w", err)
 	}
 
+    cols := rows.FieldDescriptions()
```
ItalyPaleAle, line 189: Please run the Go formatting tool (`gofumpt`).

bindings/postgres/postgres.go
```diff
@@ -186,13 +186,21 @@ func (p *Postgres) query(ctx context.Context, sql string, args ...any) (result [
...
 	for rows.Next() {
 		val, rowErr := rows.Values()
 		if rowErr != nil {
 			return nil, fmt.Errorf("error reading result '%v': %w", rows.Err(), rowErr)
 		}
-		rs = append(rs, val) //nolint:asasalint
+		
+		r := map[string]interface{}{}
+		for i, col := range cols {
+			r[string(col.Name)] = val[i]
```
ItalyPaleAle, line 200: Sadly, this is a breaking change. Currently, users expect to receive a slice, but with this, they will receive a map. This will break all existing users.

Perhaps this should be a configurable option?

ItalyPaleAle (changes requested):

bindings/postgres/postgres.go
Jack397419: Infact, the query function return a json, not a map or a slice: "result, err = json.Marshal(rs)", and the json should include column name，like the query function in mysql.go: "result, err := m.jsonify(rows)"

ItalyPaleAle: Yes, my point is: that is a breaking change. If users expect the JSON to contain an array, and you're now returning a dictionary, that will break applications.

Jack397419: I agree with your opinion, and we expect the JSON contain a dictionary，how to make a configurable option? Can you help me?

berndverst: You need to do this via request metadata which a user would need to specify to enable your new functionality, or introduce a new operation code. You can look at the code of other bindings to see how this can be done.

berndverst:
@Jack397419 can you fix the linter issue, and also fix DCO?

For DCO:
To add your Signed-off-by line to every commit in this branch:

Ensure you have a local copy of your branch by [checking out the pull request locally via command line](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally).
In your local branch, run: git rebase HEAD~7 --signoff
Force push your changes to overwrite the branch: git push --force-with-lease origin master

yaron2:
ping @Jack397419

ItalyPaleAle:
This change is now part of #3282 which is implemented more correctly. I'm going to close this PR.

PR #3203, opened 2023-10-31
Azure Blob Storage v2 state store

### LINKED ISSUE #3032: Azure Blob Storage state store strips the prefix
Reported by a user on Discord.

`keyPrefix` is ineffective with the Azure Blob Storage state store. It appears that the component explicitly removes the prefix from the file key:

https://github.com/dapr/components-contrib/blob/c2dbb03069d92ee666011ed638c6fe72d6439753/state/azure/blobstorage/blobstorage.go#L221-L228

Looking at the "git blame", it appears that this behavior has been present since the very first version of the component: #348

However, this is not something we can fix because it would break backwards-compatibility for all existing users. It may require a v2 of the component.

Note that it appears that `||` is a valid character in Azure Blob Storage file paths.

ItalyPaleAle:
The v2 of the component does not strip the app's prefix

Fixes #3032

PS: This PR was opened from this repo (not a fork), so cloud tests run automatically


state/azure/blobstorage/v1/metadata.yaml
ItalyPaleAle, line 1: THere are 2 separate folders for v1 and v2 because there can be only 1 `metadata.yaml` in each folder

actorstore/postgresql/test-queries.sql
berndverst, line 1: Please remove this file from this PR. Feel free to add in a separate PR. No unrelated change here please.

berndverst (changes requested):

actorstore/postgresql/test-queries.sql
ItalyPaleAle: Oh this was a total accident. Sorry about that! I will remove it

ItalyPaleAle: Done

berndverst (approved):

PR #3207, opened 2023-11-02
Fixes first-write-wins with SQLite and MySQL

ItalyPaleAle:
#3159 introduced a regression that caused sets with first-write-wins fail if the table was empty

The issue was due to a "FROM" statement in the CTE that should not have been there.

Also improved how the SQLite component detects etag failures when FWW is used

### DIFF
--- a/state/sqlite/sqlite_dbaccess.go
+++ b/state/sqlite/sqlite_dbaccess.go
@@ -337,10 +337,6 @@ func (a *sqliteDBAccess) doSet(parentCtx context.Context, db querier, req *state
 	}
 
 	// Only check for etag if FirstWrite specified (ref oracledatabaseaccess)
-	var (
-		res  sql.Result
-		stmt string
-	)
 	ctx, cancel := context.WithTimeout(context.Background(), a.metadata.Timeout)
 	defer cancel()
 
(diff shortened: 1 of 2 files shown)


yaron2:
Great find

state/sqlite/sqlite_dbaccess.go
```diff
@@ -382,22 +389,20 @@ func (a *sqliteDBAccess) doSet(parentCtx context.Context, db querier, req *state
...
-		return err
-	}
-	if rows == 0 {
-		if req.HasETag() || req.Options.Concurrency == state.FirstWrite {
+		rows, err := res.RowsAffected()
+		if err != nil {
+			return err
+		}
+		if rows == 0 {
+			// 0 affected rows means etag failure
```
yaron2, line 403: Won't an update statement that, for example, tries to change an existing value to itself also return 0 rows affected?

ItalyPaleAle: Looks like no:

```
sqlite> select * from test;
mykey9|myvalue9
sqlite> update test set value = "myvalue9" where key = "mykey9";
sqlite> select changes()
1
```

yaron2: Ok. For MySQL it will, but this is relevant here only to SQLite.

yaron2 (approved):

PR #3217, opened 2023-11-09
Allow Azure Auth order to be specified via `azureAuthMethods` component metadata

### LINKED ISSUE #3183: Azure: GetTokenCredential optionally disable certain auth methods
## Describe the feature

for all azure based components they use this function for authentication https://github.com/dapr/components-contrib/blob/master/internal/authentication/azure/auth.go#L116

this function tries several authentication methods sequentially and returns the token for the first auth method that succeeds... however there are certain usecases when a user or a system admin does not want to allow certain auth methods to even be tried... and with the current implementation on failure scenarios all auth methods will always be tried, which will produce noisy and confusing errors

this could be improved simply using environment variables, i.e `DISABLE_AZURE_MI_AUTH` to disable specific auth methods, so if `DISABLE_AZURE_MI_AUTH` is set to true the azure components would skip and not even try the azure managed identity auth method.

## Release Note

RELEASE NOTE: **ADD** Options to disable adhoc azure auth methods

berndverst:
# Description

Fixes #3183

Implements support for specifying the order of Azure Authentication methods to try. This allows an admin to also exclude authentication methods, or to complete disallow Azure authentication.

This is accomplished via a new global component metadata field for all Azure components `azureAuthMethods`. This field accepts a string of comma-separated values from the following list: `ServicePrincipal, Certificate, WorkloadIdentity, ManagedIdentity, CLI, None`

If the list contains `None` as one of the options, all Azure Auth will be disabled regardless of other options listed.

If `azureAuthMethods` is empty, or omitted (default behavior), then the current default behavior is maintained:
```
// 1. Client credentials
// 2. Client certificate
// 3. Workload identity
// 4. MSI (we use a timeout of 1 second when no compatible managed identity implementation is available)
// 5. Azure CLI
```

## Issue reference

Please reference the issue this PR will close: #3183

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/internal/authentication/azure/auth.go
+++ b/internal/authentication/azure/auth.go
@@ -105,94 +105,158 @@ func (s EnvironmentSettings) GetAzureEnvironment() (*cloud.Configuration, error)
 	}
 }
 
-// GetTokenCredential returns an azcore.TokenCredential retrieved from, in order:
+func (s EnvironmentSettings) addClientCredentialsProvider(creds *[]azcore.TokenCredential, errs *[]error) {
+	if c, e := s.GetClientCredentials(); e == nil {
+		cred, err := c.GetTokenCredential()
+		if err == nil {
+			*creds = append(*creds, cred)
+		} else {
+			*errs = append(*errs, err)
+		}
+	}
+}
+
+func (s EnvironmentSettings) addClientCertificateProvider(creds *[]azcore.TokenCredential, errs *[]error) {
+	if c, e := s.GetClientCert(); e == nil {
+		cred, err := c.GetTokenCredential()
+		if err == nil {
+			*creds = append(*creds, cred)
+		} else {
+			*errs = append(*errs, err)
+		}
+	}
+}
+
+func (s EnvironmentSettings) addWorkloadIdentityProvider(creds *[]azcore.TokenCredential, errs *[]error) {
+	// workload identity requires values for AZURE_AUTHORITY_HOST, AZURE_CLIENT_ID, AZURE_FEDERATED_TOKEN_FILE, AZURE_TENANT_ID
+	// The workload identity mutating admissions webhook in Kubernetes injects these values into the pod.
+	// These environment variables are read using the default WorkloadIdentityCredentialOptions
+	workloadCred, err := azidentity.NewWorkloadIdentityCredential(nil)
+	if err == nil {
+		*creds = append(*creds, workloadCred)
+	} else {
+		*errs = append(*errs, err)
+	}
+}
+
+func (s EnvironmentSettings) addManagedIdentityProvider(timeout time.Duration, creds *[]azcore.TokenCredential, errs *[]error) {
+	c := s.GetMSI()
+	msiCred, err := c.GetTokenCredential()
+
+	useTimeout := true
+	if _, ok := os.LookupEnv(identityEndpoint); ok {
+		// App Service, Functions, Service Fabric and Container Apps
+		useTimeout = false
+	} else {
+		if _, ok := os.LookupEnv(arcIMDSEndpoint); ok {
+			// Azure Arc
+			useTimeout = false
+		} else {
+			if _, ok := os.LookupEnv(msiEndpoint); ok {
+				// Cloud Shell
+				useTimeout = false
+			} else if isVirtualMachineWithManagedIdentity() {
+				// Azure VM with MSI enabled
+				useTimeout = false
+			}
+		}
+	}
+
+	// We need to use a timeout for MSI on environments where it is not available because the request for the default IMDS endpoint can hang for several minutes.
+	if useTimeout {
+		msiCred = &timeoutWrapper{cred: msiCred, authmethod: "managed identity", timeout: timeout}
+	}
+
+	if err == nil {
+		*creds = append(*creds, msiCred)
+	} else {
+		*errs = append(*errs, err)
+	}
+}
+
+func (s EnvironmentSettings) addCLIProvider(timeout time.Duration, creds *[]azcore.TokenCredential, errs *[]error) {
+	cred, credErr := azidentity.NewAzureCLICredential(nil)
+	if credErr == nil {
+		*creds = append(*creds, &timeoutWrapper{cred: cred, authmethod: "Azure CLI", timeout: 30 * time.Second})
+	} else {
+		*errs = append(*errs, credErr)
+	}
+}
+
+func (s EnvironmentSettings) addProviderByAuthMethodName(authMethod string, creds *[]azcore.TokenCredential, errs *[]error) {
+	switch authMethod {
+	case "serviceprincipal":
+		s.addClientCredentialsProvider(creds, errs)
+	case "certificate":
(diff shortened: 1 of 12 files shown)


internal/authentication/azure/metadata-properties.go
```diff
@@ -35,6 +35,9 @@ var MetadataKeys = map[string][]string{ //nolint:gochecknoglobals
 	// Identifier for the Azure environment
 	// Allowed values (case-insensitive): AzurePublicCloud/AzurePublic, AzureChinaCloud/AzureChina, AzureUSGovernmentCloud/AzureUSGovernment
 	"AzureEnvironment": {"azureEnvironment", "azureCloud"},
+	// Identifier for the Azure authentication methods to try (in order), comma-separated
+	// Allowed values (case-insensitive): ServicePrincipal, Certificate, WorkloadIdentity, ManagedIdentity, CLI, None
+	"AzureAuthMethods": {"azureAuthMethods", "azureADAuthMethods"},
```
ItalyPaleAle, line 40: Should it be "azureADAuthMethods" as default maybe?

internal/authentication/azure/metadata-properties.go
```diff
@@ -35,6 +35,9 @@ var MetadataKeys = map[string][]string{ //nolint:gochecknoglobals
 	// Identifier for the Azure environment
 	// Allowed values (case-insensitive): AzurePublicCloud/AzurePublic, AzureChinaCloud/AzureChina, AzureUSGovernmentCloud/AzureUSGovernment
 	"AzureEnvironment": {"azureEnvironment", "azureCloud"},
+	// Identifier for the Azure authentication methods to try (in order), comma-separated
+	// Allowed values (case-insensitive): ServicePrincipal, Certificate, WorkloadIdentity, ManagedIdentity, CLI, None
```
ItalyPaleAle, line 39: I think "Service Principal" and "Certificate" are bad names. All the methods above use a service principal (also MI uses a SP).

More appropriate IMHO would be "ClientCredentials" and "ClientCertificate"

ItalyPaleAle, line 39: Is there a way to have aliases for these? Such as "WI" for WorkloadIdentity?

internal/authentication/azure/auth_test.go
```diff
@@ -221,6 +221,80 @@ func TestAuthorizorWithMSI(t *testing.T) {
...
+	os.Setenv("MSI_ENDPOINT", "test")
+	defer os.Unsetenv("MSI_ENDPOINT")
+	settings, err := NewEnvironmentSettings(
+		map[string]string{
+			"azureClientId":    fakeClientID,
+			"vaultName":        "vaultName",
+			"azureAuthMethods": "serviceprincipal,certificate,workloadidentity,managedIdentity,cli,SUPERAUTH",
+		},
+	)
+	assert.NoError(t, err)
```
ItalyPaleAle, line 288:
```suggestion
	require.NoError(t, err)
```

internal/authentication/azure/auth_test.go
```diff
@@ -221,6 +221,80 @@ func TestAuthorizorWithMSI(t *testing.T) {
...
+
+	testCertConfig := settings.GetMSI()
+	assert.NotNil(t, testCertConfig)
```
ItalyPaleAle, line 291:
```suggestion
	require.NotNil(t, testCertConfig)
```

internal/authentication/azure/auth_test.go
ItalyPaleAle: A bit of a nit, but if there's an error (or if you expect a value to be non-nil), the test should abort, because it's in a bad state

internal/authentication/azure/auth.go
```diff
@@ -105,94 +105,158 @@ func (s EnvironmentSettings) GetAzureEnvironment() (*cloud.Configuration, error)
...
+		s.addWorkloadIdentityProvider(creds, errs)
+	case "managedidentity":
+		s.addManagedIdentityProvider(1*time.Second, creds, errs)
+	case "cli":
+		s.addCLIProvider(30*time.Second, creds, errs)
+	}
+}
+
+func getAzureAuthMethods() [6]string {
+	return [...]string{"serviceprincipal", "certificate", "workloadidentity", "managedidentity", "cli", "none"}
```
ItalyPaleAle, line 202: Just return a slice?
```suggestion
func getAzureAuthMethods() []string {
	return []string{"serviceprincipal", "certificate", "workloadidentity", "managedidentity", "cli", "none"}
```

ItalyPaleAle (changes requested):

internal/authentication/azure/metadata-properties.go
berndverst: No because of the entra ID nonsense. That's why I chose this name. Please see what I had written on the linked issue.

internal/authentication/azure/auth.go
berndverst: It's silly optimization but this makes the compiler allocate a fixed length slice. So why not. I'll keep it, don't want anyone to dynamically append to this.

internal/authentication/azure/metadata-properties.go
ItalyPaleAle: Ok. How about you already pre-add "entraIDAuthMethods" and "azureEntraIDAuthMethods" as aliases then?

berndverst: It's called MicrosoftEntraID - not AzureEntraID 😄

berndverst: That's why I chose "azureAuthMethods" as the default - because it is neither about Azure AD, nor about Microsoft Entra ID :)

ItalyPaleAle (approved):

internal/authentication/azure/metadata-properties.go
ItalyPaleAle: Damn marketers :D

internal/authentication/azure/auth.go
```diff
@@ -105,94 +105,158 @@ func (s EnvironmentSettings) GetAzureEnvironment() (*cloud.Configuration, error)
...
+
+		// 4. MSI with timeout of 1 second (same as DefaultAzureCredential)
+		s.addManagedIdentityProvider(1*time.Second, &creds, &errs)
+
+		// 5. AzureCLICredential
+		s.addCLIProvider(30*time.Second, &creds, &errs)
+	} else {
+		authMethodIdentifiers := getAzureAuthMethods()
+		authMethods := strings.Split(strings.ToLower(strings.TrimSpace(authMethods)), ",")
+		for _, authMethod := range authMethods {
```
ItalyPaleAle, line 242: QQ: What if one passes both "ManagedIdentity" and "MI"? Wouldn't that cauise the auth method to be added twice?

berndverst: That is correct - and that's totally ok! You might want it to be tried twice for some reason. You have full control :)

You can have write `mi,mi,mi,mi,mi` it's all up to you :smile

PR #3222, opened 2023-11-10
[pubsub/jetstream] Add support for concurrencyMode

bruth:
# Description

This adds support for the `concurrencyMode` metadata field which will send spawn a new goroutine per message like the RabbitMQ implementation.

## Issue reference

n/a

This was requested by a user of NATS/JetStream to support this parallel mode. /cc @Moscagus

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


pubsub/jetstream/metadata.go
```diff
@@ -116,5 +118,17 @@ func parseMetadata(psm pubsub.Metadata) (metadata, error) {
 		m.internalAckPolicy = nats.AckExplicitPolicy
 	}
 
+	// Explicit check to default to Single mode since this was the
+	// previous default behavior.
+	if psm.Properties[pubsub.ConcurrencyKey] != "" {
```
ItalyPaleAle, line 123: We strive to make all metadata keys case-insensitive. THe way here to set a default value would be to set the default value when initializing the metadata struct, such:

Replace

```go
var m metadata
```

with:

```go
m := metadata{
    Concurrency: pubsub.Single,
}
```

This way, if the property passed by the user is empty or unset, the default value is used

pubsub/jetstream/jetstream.go
```diff
@@ -255,7 +253,22 @@ func (js *jetstreamPubSub) Subscribe(ctx context.Context, req pubsub.SubscribeRe
 		subscription, err = js.jsc.QueueSubscribe(req.Topic, queue, natsHandler, nats.Bind(streamName, consumerInfo.Name))
 	} else {
 		js.l.Debugf("nats: subscribed to subject %s", req.Topic)
-		subscription, err = js.jsc.Subscribe(req.Topic, natsHandler, nats.Bind(streamName, consumerInfo.Name))
+		subscription, err = js.jsc.Subscribe(
+			req.Topic,
+			func(msg *nats.Msg) {
```
ItalyPaleAle, line 258: Nit: no need to capture the parameter in the closure since it's not in a loop

```suggestion
			func() {
```

ItalyPaleAle (changes requested):
Thank you for this PR, it's something very useful! A couple of small comments but overall the direction looks good

bruth:
Thanks for the quick review. Addressed the comments and rebased the branch.

pubsub/jetstream/metadata.go
```diff
@@ -119,5 +123,15 @@ func parseMetadata(psm pubsub.Metadata) (metadata, error) {
 		m.internalAckPolicy = nats.AckExplicitPolicy
 	}
 
+	// Explicit check to prevent overriding the Single default
+	// (the previous behavior) if not set.
+	if psm.Properties[pubsub.ConcurrencyKey] != "" {
```
ItalyPaleAle, line 128: This needs to be removed too? Still not case-insensitive

bruth: The `pubsub.Concurrency` function defaults to `pubsub.Parallel` when the value is empty/undefined. So if I remove this check, the function will default to `Parallel` which I want to prevent in this case since the previous behavior was not that.

bruth: And FWIW, it appears the [`Concurrency`](https://github.com/dapr/components-contrib/blob/main/pubsub/concurrency.go) does appear to check for case-insensitivity.

ItalyPaleAle: Ok, that's a bug we need to fix in the Concurrency function. Both the fact that it overrides the default value in the object if the metadata property is empty (it should respect what you set in `metadata{}`), and the fact that it isn't case-insensitive.

For now, I'll accept this, but can you add a comment with a TODO linking to this conversation please?

Moscagus:
@bruth thanks for the PR.

pubsub/jetstream/metadata.go
bruth: Added the TODO comment, so ready for merge from my perspective.

Moscagus:
@bruth for reasons of high availability and scalability, I use at least 2 pods per application. Because of this I must configure the "DeliverGroup" ("queueGroupName") property in "push" mode.
Looking at the "PR" changes I see that "concurrencyMode" is not applied when the "queueGroupName" is set.
Therefore, at the scalability level, in my use case I remain the same as before the change. That is, if I want to parallelize the app by 100, I must upload 100 pods.
Is it possible to make the same change to ?:
**subscription, err = js.jsc.QueueSubscribe(req.Topic, queue, natsHandler, nats.Bind(streamName, consumerInfo.Name))**
Thanks

bruth:
Bah, that was an oversight.. sorry about that, will open another PR in a minute.

bruth:
Here is the new PR: https://github.com/dapr/components-contrib/pull/3232

Moscagus:
@bruth thanks for the new PR.

bruth:
@Moscagus new PR just got merged, FYI.

Moscagus:
@bruth thanks

Issue #2526, opened 2023-02-13
Improve "File not Found" error for Azure BlobStorage and AWS S3 bindings


federicocodo:
As the previous issue (#2374) is it possible to add a similar line of code, added for LocalStorage binding, also for other storages like Azure BlobStorage, AWS S3, ecc...?
```go
if os.IsNotExist(err) {
    return nil, fmt.Errorf("file not found: %s", absPath)
}
```
Currently for Azure BlobStorage I get this error: 
![image](https://user-images.githubusercontent.com/62144863/218482695-ea8cf87f-dbf5-48e3-96fb-8689686830d0.png)
This is all because I'm running a multi-storage service and trying to manage all responses.
Thanks in advance.

berndverst:
Sure, contributions are welcome. Please note that in Dapr 1.10 and going forward we use a completely different Azure Blob Storage SDK. So you will not be able to reproduce this exact error message there.

artursouza:
Removing from 1.12, if anyone from the community creates a PR, we can bring it back.

PR #3223, opened 2023-11-10
enhancement: not found error handling

sadath-12:
# Description

Added specific error handling to object not found mentioning the key so user is not frustated with unnecessary long errors

## Issue reference

Please reference the issue this PR will close: #2526 _

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/bindings/aws/s3/s3.go
+++ b/bindings/aws/s3/s3.go
@@ -18,6 +18,7 @@ import (
 	"crypto/tls"
 	b64 "encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -300,6 +301,10 @@ func (s *AWSS3) get(ctx context.Context, req *bindings.InvokeRequest) (*bindings
 		},
 	)
 	if err != nil {
+		var awsErr s3.Error
+		if errors.As(err, &awsErr) && *awsErr.Code == s3.ErrCodeNoSuchKey {
+			return nil, fmt.Errorf("object not found: %s", key)
+		}
 		return nil, fmt.Errorf("s3 binding error: error downloading S3 object: %w", err)
 	}
 
--- a/bindings/gcp/bucket/bucket.go
+++ b/bindings/gcp/bucket/bucket.go
@@ -23,11 +23,16 @@ import (
 	"net/url"
 	"reflect"
 	"strconv"
+	"errors"
+	"net/http"
+
 
 	"cloud.google.com/go/storage"
 	"github.com/google/uuid"
 	"google.golang.org/api/iterator"
 	"google.golang.org/api/option"
+	"google.golang.org/api/googleapi"
+
 
 	"github.com/dapr/components-contrib/bindings"
 	"github.com/dapr/components-contrib/metadata"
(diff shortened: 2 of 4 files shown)


bindings/aws/s3/s3.go
```diff
@@ -300,6 +301,10 @@ func (s *AWSS3) get(ctx context.Context, req *bindings.InvokeRequest) (*bindings
 		},
 	)
 	if err != nil {
+		var awsErr s3.Error
+		if errors.As(err, &awsErr) && *awsErr.Code == s3.ErrCodeNoSuchKey {
+			return nil, fmt.Errorf("object not found: %s", key)
```
ItalyPaleAle, line 306: Let's not include the key in the result as it could contain PII (Personally Identifiable Information). Same for the other components

ItalyPaleAle (changes requested):
Thank you! This is a great start.

In addition to the comment I left, can you please update the certification tests to add a case of testing retrieving a file that doesn't exist?

- https://github.com/dapr/components-contrib/tree/main/tests/certification/bindings/azure/blobstorage
- https://github.com/dapr/components-contrib/tree/main/tests/certification/bindings/aws/s3
- There's no test for OBS so you can skip that

bindings/aws/s3/s3.go
sadath-12: Sure , Good catch !

bindings/gcp/bucket/bucket.go
```diff
@@ -213,6 +218,11 @@ func (g *GCPStorage) get(ctx context.Context, req *bindings.InvokeRequest) (*bin
 	var rc io.ReadCloser
 	rc, err = g.client.Bucket(g.metadata.Bucket).Object(key).NewReader(ctx)
 	if err != nil {
+		var apiErr *googleapi.Error
+		if errors.As(err, &apiErr) && apiErr.Code == http.StatusNotFound {
+			return nil, fmt.Errorf("object not found: %s", key)
```
ItalyPaleAle, line 223:
```suggestion
			return nil, errors.New("object not found")
```

bindings/huawei/obs/obs.go
```diff
@@ -219,6 +221,10 @@ func (o *HuaweiOBS) get(ctx context.Context, req *bindings.InvokeRequest) (*bind
 
 	out, err := o.service.GetObject(ctx, input)
 	if err != nil {
+		var obsErr obs.ObsError
+		if errors.As(err, &obsErr) && obsErr.StatusCode == http.StatusNotFound {
+			return nil, fmt.Errorf("object not found: %s", key)
```
ItalyPaleAle, line 226:
```suggestion
			return nil, errors.New("object not found")
```

bindings/gcp/bucket/bucket.go
berndverst, line 1: You are almost done - you still need to format this file properly - that is done with the `gofumpt` tool. There are other files that need to be formatted as well - see the linter error (`make lint` to run the linter locally)

berndverst (changes requested):

bindings/gcp/bucket/bucket.go
sadath-12: Thank you @berndverst 🙂

PR #3227, opened 2023-11-15
CosmosDBState - Add, as option, the partitionKey in QueryMethod

luigirende:
Opportunity to add the partitionKey field in the metadata of the  Query request (how it's done in the Set and Get method).
If the user adds the partitionKey in the query method, the database before filtering all the items that belongs to the same partitionKey value, used in the request; after applies the query defined in the request. In this case the cross partition will be  disabled, so the scan  of the item that the database will do in the container will be only for a subset of item, improving the performance

# Description

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


state/azure/cosmosdb/cosmosdb.go
```diff
@@ -92,8 +92,11 @@ func (p crossPartitionQueryPolicy) Do(req *policy.Request) (*http.Response, erro
 	// Check if we're performing a query
 	// In that case, remove the partitionkey header and enable cross-partition queries
 	if strings.ToLower(raw.Header.Get("x-ms-documentdb-query")) == "true" {
-		raw.Header.Add("x-ms-documentdb-query-enablecrosspartition", "true")
-		raw.Header.Del("x-ms-documentdb-partitionkey")
+		raw.Header.Add("x-ms-documentdb-query-enablecrosspartition", "True")
```
berndverst, line 95: probably should be lowercase `true` or else this would not match!

berndverst: Move this into the if-clause you added. This header should only be added when the partition key is fake. You do not want to run expensive cross partition queries otherwise.

state/feature.go
```diff
@@ -26,6 +26,8 @@ const (
 	FeatureQueryAPI Feature = "QUERY_API"
 	// FeatureTTL is the feature that supports TTLs.
 	FeatureTTL Feature = "TTL"
+	// FeaturePartitionKey is the feature that supports the partition
+	FeaturePartitionKey Feature = "PARTITION_KEY"
```
berndverst, line 30: Please remove from here. We do not want to introduce this as a feature for all state stores because it doesn't apply to all state stores. You don't have to (and should not) formally define this feature.

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -206,6 +209,7 @@ func (c *StateStore) Features() []state.Feature {
 		state.FeatureTransactional,
 		state.FeatureQueryAPI,
 		state.FeatureTTL,
+		state.FeaturePartitionKey,
```
berndverst, line 212: Please remove this per the other comment.

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -92,8 +92,11 @@ func (p crossPartitionQueryPolicy) Do(req *policy.Request) (*http.Response, erro
 	// Check if we're performing a query
 	// In that case, remove the partitionkey header and enable cross-partition queries
 	if strings.ToLower(raw.Header.Get("x-ms-documentdb-query")) == "true" {
-		raw.Header.Add("x-ms-documentdb-query-enablecrosspartition", "true")
-		raw.Header.Del("x-ms-documentdb-partitionkey")
+		raw.Header.Add("x-ms-documentdb-query-enablecrosspartition", "True")
+		// Only when the partitionKey is fake (true), it will be removed
+		if strings.ToLower(raw.Header.Get("x-ms-documentdb-partitionkey")) == "[true]" {
+			raw.Header.Del("x-ms-documentdb-partitionkey")
```
berndverst, line 98: You only want to add the cross partition query header here instead.

berndverst (changes requested):
Generally the idea is right, but some small correction on the details. When to add headers etc

Also please do not define the new feature key / capability. That is only intended for features which we (at least at some point in time) intend to support in all state stores.

tests/conformance/state/state.go
```diff
@@ -235,6 +242,64 @@ func ConformanceTests(t *testing.T, props map[string]string, statestore state.St
...
+			},
+			partitionOnly: true,
+			// The same query from previous test but return only item having the same partitionKey value (%s-struct-2) given in the metadata
+			results: []state.QueryItem{
+				{
+					Key:  fmt.Sprintf("%s-struct-2", key),
+					Data: []byte(fmt.Sprintf(`{"message":"%stest"}`, key)),
+				},
+			},
+		},
```
luigirende, line 302: @berndverst I added PartitionKey as feature for doing this test; this scenario will be ok only for CosmosDB, because for the others states will return 2 items instead of 1.

tests/conformance/state/state.go
```diff
@@ -300,10 +365,12 @@ func ConformanceTests(t *testing.T, props map[string]string, statestore state.St
 			// Check if query feature is listed
 			features := statestore.Features()
 			require.True(t, state.FeatureQueryAPI.IsPresent(features))
-
 			querier, ok := statestore.(state.Querier)
 			assert.True(t, ok, "Querier interface is not implemented")
 			for _, scenario := range queryScenarios {
+				if (scenario.partitionOnly) && (!state.FeaturePartitionKey.IsPresent(features)) {
+					break
+				}
```
luigirende, line 373: @berndverst the test will be executed only for store with partitionKey. We can remove the test with partition key or if you can suggest me some solution to execute this test without using the new feature PartitionKey.

berndverst (changes requested):
I finally got around to merging your query operators PR.

Could you resolve the merge conflict here now and then we'll get this one merged also? :)

luigirende:
@berndverst  done!

tests/conformance/state/state.go
berndverst: I see. Not ideal - but I can't at the moment think of a better way to do this either.

berndverst:
`/ok-to-test`

berndverst:
@luigirende I'll merge this. Can you please make sure to take ownership of this docs issue and make the docs PR? Otherwise nobody will know about this.

https://github.com/dapr/docs/issues/3897

Use `/assign` on that issue.

berndverst (approved):

berndverst (changes requested):
The CosmosDB conformance test is failing.

luigirende:
> The CosmosDB conformance test is failing.

Hi @berndverst 

I think there has been an issue with azure tests. In fact fail the azure postgresql test too and in this PR I haven't done any change on postgresql state. 
Looking at the test log seems the dbs (cosmosdb and postgresql) wasn't empty when run the tests in fact the query found other keys store previously.

> 
>         	Error Trace:	/home/runner/work/components-contrib/components-contrib/tests/conformance/state/state.go:478
>         	Error:      	Not equal: 
>         	            	expected: "1fa86a779a7d49f0ba1a15fe7ce75f01-struct-operations"
>         	            	actual  : "429bcb9db8d04c8da420d686492f0239-struct-operations"

Can you do a check or test again the PR?

berndverst:
> > The CosmosDB conformance test is failing.
> 
> Hi @berndverst
> 
> I think there has been an issue with azure tests. In fact fails also the azure postgresql tes and in this PR I haven't done any change on postgresql state. Looking at the test log seems the dbs (cosmosdb and postgresql) weren't empty, when run the tests, in fact, the query find other keys store previously.
> 
> > ```
> >     	Error Trace:	/home/runner/work/components-contrib/components-contrib/tests/conformance/state/state.go:478
> >     	Error:      	Not equal: 
> >     	            	expected: "1fa86a779a7d49f0ba1a15fe7ce75f01-struct-operations"
> >     	            	actual  : "429bcb9db8d04c8da420d686492f0239-struct-operations"
> > ```
> 
> Can you do a check or test again the PR?

@luigirende  I think this because of your other PR which is now merged, but you also must keep in mind that multiple tests might be running in parallel. Your test must work in that situation!

When the test starts it uses a unique prefix for the test run. So you need to write your test such that it will only look for the results with this particular prefix.

Also keep in mind you still have a conflict in `state/feature.go`.

luigirende:
@berndverst merged the conflict in `state/feature.go` and fix the issue in the conformance tests

berndverst (approved):

Issue #3088, opened 2023-08-17
Add support for Azure Cache for Redis in all Redis components


berndverst:
This request is similar to the existing Azure Postgres state store support. https://github.com/dapr/components-contrib/pull/2970

AAD auth needs to be supported by fetching an AAD token and using it as the password, the username is the object ID / client ID.

Due to the long lived TCP connections however we need to ensure to manually renew the AAD token in a background goroutine and reinit the connection.

berndverst:
cc @ItalyPaleAle

ItalyPaleAle:
> Due to the long lived TCP connections however we need to ensure to manually renew the AAD token in a background goroutine and reinit the connection.

I don't think this is needed as long as the connection is alive. The token should be refreshed only if the connection needs to be re-established (that's what we do for Postgres and want to do for MySQL)

shpathak-msft:
If the token expires, the Azure Cache for Redis server will terminate the connection and it expects a renewed token to keep the connection alive.

ItalyPaleAle:
> If the token expires, the Azure Cache for Redis server will terminate the connection and it expects a renewed token to keep the connection alive.

Since AAD tokens have a lifetime of 1hr, are you saying that connections with Azure Cache for Redis are always terminated every hour forcefully?

(This doesn't seem to be the case for Azure DB for Postgres or Azure SQL AFAICT)

shpathak-msft:
Right, the client applications create long-running, TCP connections to Redis which are expected to be alive for the duration of the application ideally. We typically see connections alive and running for weeks without interruption.

ItalyPaleAle:
@shpathak-msft Ok, I understand now.

In practice, does that mean we need to re-send the [AUTH](https://redis.io/commands/auth/) command periodically, when we get a new token (like every hours or so)?

Do you know if there's a Go SDK that implements this already?

shpathak-msft:
Looks like go-redis supports it: https://github.com/redis/go-redis/blob/c3098d5f7eba88330638a8b2b3e23d42cd5c9338/commands.go#L514

berndverst:
@shpathak-msft can you confirm sending the AUTH command periodically will actually work with your system? Or do we completely need to terminate the connection?

shpathak-msft:
[...]

PR #3238, opened 2023-11-20
feature: add redis token refresh support

sadath-12:
# Description

The changes proposed will support token refresh of redis using auth command every hour and maintains the connection without terminating as before

## Issue reference

Please reference the issue this PR will close: #3088 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


common/component/redis/v8client.go
```diff
@@ -440,3 +448,33 @@ func newV8Client(s *Settings) RedisClient {
...
 }
+
+func refreshTokenRoutineV8(ctx context.Context, redisClient *v8.ClusterClient, meta map[string]string) {
+	ticker := time.NewTicker(time.Hour)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			env, err := azure.NewEnvironmentSettings(meta)
```
daixiang0, line 459: Does it means only support azure? A good start is a generic way.

daixiang0:
ping @johnewart @ItalyPaleAle

sadath-12:
Sure, Thank you @daixiang0

common/component/redis/v9client.go
```diff
@@ -395,9 +400,10 @@ func newV9Client(s *Settings) RedisClient {
 				InsecureSkipVerify: s.EnableTLS,
 			}
 		}
-
+		client := v9.NewClusterClient(options)
+		go refreshTokenRoutineForRedis(context.Background(), ClientFromV9Client(client), properties)
```
JoshVanL, line 404: Is it the case that we _always_ want to be doing this refresh? Should this not be an option?

JoshVanL (changes requested):
Some comments from me :slightly_smiling_face:

common/component/redis/v9client.go
sadath-12: ya it seems the connection might terminate so we must be doing it continuously in some interval

common/component/redis/v9client.go
```diff
@@ -429,11 +435,46 @@ func newV9Client(s *Settings) RedisClient {
...
 		dialTimeout:  s.DialTimeout,
 	}
 }
+
+func ClientFromV9Client(client v9.UniversalClient) RedisClient {
+	return v9Client{client: client}
+}
+
+func refreshTokenRoutineForRedis(ctx context.Context, redisClient RedisClient, meta map[string]string) {
+	ticker := time.NewTicker(time.Hour)
```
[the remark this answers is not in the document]
sadath-12: In the issue @ItalyPaleAle had mentioned it, so I thought to go with 1 hour

common/component/redis/v9client.go
```diff
@@ -429,11 +435,46 @@ func newV9Client(s *Settings) RedisClient {
 			InsecureSkipVerify: s.EnableTLS,
 		}
 	}
-
+	client := v9.NewClient(options)
+	go refreshTokenRoutineForRedis(context.Background(), ClientFromV9Client(client), properties)
```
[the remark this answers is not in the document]
sadath-12: Actually I thought of it at start , but client needs to be returned for other calls , so if we wait it wun't return right and later anyways if the parent function of this finishes the go routine will also be terminated .. I'm I right ?

sadath-12:
Push the changes @JoshVanL . In order to make sure that background goroutine is terminated I have used ```closeCh``` .  As you mentioned in comments about wait group ? do we have to use it here necessarily ? Since after close that goroutine will either listen to the incoming channel and terminate or if it does not listen means it was terminated anyways

common/component/redis/v8client.go
```diff
@@ -27,6 +28,8 @@ type v8Pipeliner struct {
 	writeTimeout Duration
 }
 
+var v8logger = logger.NewLogger("dapr.components.redisv9")
```
ItalyPaleAle, line 31:
```suggestion
var v8logger = logger.NewLogger("dapr.components.redisv8")
```

common/component/redis/v8client.go
```diff
@@ -349,27 +354,32 @@ func newV8FailoverClient(s *Settings) RedisClient {
 
 	if s.RedisType == ClusterType {
 		opts.SentinelAddrs = strings.Split(s.Host, ",")
-
+		client := v8.NewFailoverClusterClient(opts)
+		go refreshTokenRoutineForRedis(context.Background(), ClientFromV8Client(client), properties, v8logger, closeCh)
```
ItalyPaleAle, line 358: I think this should only be used if using Azure AD?

common/component/redis/v8client.go
```diff
@@ -349,27 +354,32 @@ func newV8FailoverClient(s *Settings) RedisClient {
...
+			client:       client,
 			readTimeout:  s.ReadTimeout,
 			writeTimeout: s.WriteTimeout,
 			dialTimeout:  s.DialTimeout,
+			closeCh:      closeCh,
 		}
 	}
-
+	client := v8.NewFailoverClient(opts)
+	go refreshTokenRoutineForRedis(context.Background(), ClientFromV8Client(client), properties, v8logger, closeCh)
```
ItalyPaleAle, line 368: Here too, only if using Azure AD?

common/component/redis/v9client.go
```diff
@@ -429,11 +443,53 @@ func newV9Client(s *Settings) RedisClient {
...
+				logger.Debug("Failed to get Azure AD token credential:", err)
+				continue
+			}
+			at, err := tokenCred.GetToken(ctx, policy.TokenRequestOptions{
+				Scopes: []string{
+					env.Cloud.Services[azure.ServiceOSSRDBMS].Audience + "/.default",
+				},
+			})
+			if err != nil {
+				logger.Debug("Failed to get Azure AD token:", err)
```
ItalyPaleAle, line 482: Should be a higher-level error here, maybe `logger.Error`

common/component/redis/v9client.go
```diff
@@ -429,11 +443,53 @@ func newV9Client(s *Settings) RedisClient {
...
+				continue
+			}
+
+			// Authenticate with Redis using the refreshed token
+			err = redisClient.(v9Client).client.Pipeline().Auth(ctx, at.Token).Err()
+			if err != nil {
+				logger.Debug("Failed to authenticate with Redis using refreshed Azure AD token:", err)
```
ItalyPaleAle, line 489: Here too - `logger.Error` perhaps since the error isn't returned but just logged

common/component/redis/v9client.go
```diff
@@ -429,11 +443,53 @@ func newV9Client(s *Settings) RedisClient {
...
+					env.Cloud.Services[azure.ServiceOSSRDBMS].Audience + "/.default",
+				},
+			})
+			if err != nil {
+				logger.Debug("Failed to get Azure AD token:", err)
+				continue
+			}
+
+			// Authenticate with Redis using the refreshed token
+			err = redisClient.(v9Client).client.Pipeline().Auth(ctx, at.Token).Err()
```
ItalyPaleAle, line 487: What if it's v8 client?

common/component/redis/v9client.go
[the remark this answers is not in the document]
ItalyPaleAle: The token must be renewed every hour, but that means the loop must run every "a little bit less than 1hr". For example every 50 minutes, to have some buffer

sadath-12:
loop must run every "a little bit less than 1hr" ? why and how ? please

ItalyPaleAle:
> loop must run every "a little bit less than 1hr" ? why and how ? please

So [in here](https://github.com/dapr/components-contrib/issues/3088#issuecomment-1683015805) the person from the Azure Redis Cache team explained that the service expects a new AUTH command, with a new token, sent every hour _at most_.

If you run the loop every 1h exactly, you may be sending the updated command too late, maybe even by a few ms, so this would cause the session to expire.

When there's something that must be renewed periodically, the common pattern is to do that a little bit before the deadline. In this case, if the deadline is 1h, maybe do it every 50m. This also leaves us some buffer, so if requesting the token fails, it can be retried 1m later, and we are still < 1h so the session is kept active.

sadath-12:
> > loop must run every "a little bit less than 1hr" ? why and how ? please
> 
> So [in here](https://github.com/dapr/components-contrib/issues/3088#issuecomment-1683015805) the person from the Azure Redis Cache team explained that the service expects a new AUTH command, with a new token, sent every hour _at most_.
> 
> If you run the loop every 1h exactly, you may be sending the updated command too late, maybe even by a few ms, so this would cause the session to expire.
> 
> When there's something that must be renewed periodically, the common pattern is to do that a little bit before the deadline. In this case, if the deadline is 1h, maybe do it every 50m. This also leaves us some buffer, so if requesting the token fails, it can be retried 1m later, and we are still < 1h so the session is kept active.

Thank you 😊👌

ItalyPaleAle (changes requested):

sadath-12:
thank you @ItalyPaleAle  , following that

common/component/redis/redis.go
```diff
@@ -22,8 +22,12 @@ import (
 
 	"golang.org/x/mod/semver"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
```
ItalyPaleAle, line 25:
```suggestion
	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
	"golang.org/x/mod/semver"
```

common/component/redis/redis.go
```diff
@@ -252,3 +256,51 @@ type RedisError string
...
+				logger.Error("Failed to get Azure AD token credential:", err)
+				continue
+			}
+			at, err := tokenCred.GetToken(ctx, policy.TokenRequestOptions{
+				Scopes: []string{
+					env.Cloud.Services[azure.ServiceOSSRDBMS].Audience + "/.default",
+				},
+			})
+			if err != nil {
+				logger.Debug("Failed to get Azure AD token:", err)
```
ItalyPaleAle, line 289:
```suggestion
				logger.Error("Failed to get Azure AD token:", err)
```

common/component/redis/redis.go
```diff
@@ -252,3 +256,51 @@ type RedisError string
...
+			}
+
+			// Authenticate with Redis using the refreshed token
+			if version == "v8" {
+				err = redisClient.(v8Client).client.Pipeline().Auth(ctx, at.Token).Err()
+			} else if version == "v9" {
+				err = redisClient.(v9Client).client.Pipeline().Auth(ctx, at.Token).Err()
+			}
+			if err != nil {
+				logger.Error("Failed to authenticate with Redis using refreshed Azure AD token:", err)
```
ItalyPaleAle, line 300: I think if we are here, we should retry right away and until we succeed, without waiting for the next ticker.

common/component/redis/redis.go
ItalyPaleAle: I'ts just what our linter wants :) Ideally dependencies should be in 3 "blocks":

1. standard library
2. 3rd-party deps
3. Deps from "github.com/dapr"

sadath-12: oh ya a good way of making it clean

berndverst:
@sadath-12 you need to run your changed files through an autoformatter. `gofumpt`

DeepanshuA:
I did following to verify this:

1. Enabled Managed identity on my already existing AKS cluster: az aks update -g <myResourceGroup> -n <myManagedCluster> --enable-managed-identity
2. Created "Azure cache for Redis" - Here, Enabled "Managed Identity"
3. In `Authentication` tab of my Redis (Azure Portal), I provided my Service Principal (my AKS cluster ObjectId) 
4. Created new dapr Build with PR 3238 and deployed that on my AKS cluster
5. Created a new redis component:
```apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: statestore
spec:
  type: state.redis
  version: v1
  metadata:
  - name: useAzureAD
    value: true```
 
6. Applied nodeapp from hello-kubernetes as it would use redis statestore, but it fails with this error:
 
time="2024-01-29T12:50:50.191942482Z" level=fatal msg="Fatal error from runtime: process component statestore error: [INIT_COMPONENT_FAILURE]: initialization error occurred for statestore (state.redis/v1): [INIT_COMPONENT_FAILURE]: initialization error occurred for statestore (state.redis/v1): redis store: error connecting to redis at : dial tcp [::1]:6379: connect: connection refused" app_id=nodeapp instance=nodeapp-577cc79c45-4b7rs scope=dapr.runtime type=log ver=edge

Do we need to enable managed identity some other way to verify this?
When I had tried to add principalId in user identity section of "azure redis", then it was not  able to recognize that.
 
so, I had added to Authentication.

**Comment by berndverst** [comment]
> I did following to verify this:
> 
> 1. Enabled Managed identity on my already existing AKS cluster: az aks update -g  -n  --enable-managed-identity
> 2. Created "Azure cache for Redis" - Here, Enabled "Managed Identity"
> 3. In `Authentication` tab of my Redis (Azure Portal), I provided my Service Principal (my AKS cluster ObjectId)
> 4. Created new dapr Build with PR 3238 and deployed that on my AKS cluster
> 5. Created a new redis component:
> 
> ```
> kind: Component
> metadata:
>   name: statestore
> spec:
>   type: state.redis
>   version: v1
>   metadata:
>   - name: useAzureAD
>     value: true
> ```
> 
> 6. Applied nodeapp from hello-kubernetes as it would use redis statestore, but it fails with this error:
> 
> time="2024-01-29T12:50:50.191942482Z" level=fatal msg="Fatal error from runtime: process component statestore error: [INIT_COMPONENT_FAILURE]: initialization error occurred for statestore (state.redis/v1): [INIT_COMPONENT_FAILURE]: initialization error occurred for statestore (state.redis/v1): redis store: error connecting to redis at : dial tcp [::1]:6379: connect: connection refused" app_id=nodeapp instance=nodeapp-577cc79c45-4b7rs scope=dapr.runtime type=log ver=edge
> 
> When I had tried to add principalId in user identity section of "azure redis", then it was not able to recognize that. So, I had added to Authentication.
> 
> Do we need to enable managed identity some other way to verify this? @shpathak-msft if you could please confirm here?

@DeepanshuA please use Azure Workload Identity instead. For this you first create a managed identity - user assigned or system assigned. Obviously you need to give the correct Redis permissions (role assignment) to this identity.

Then you must follow these steps for your Dapr enabled deployment:
https://learn.microsoft.com/en-us/azure/aks/workload-identity-overview?tabs=dotnet

`azure.workload.identity/use: true`
and
`azure.workload.identity/client-id` which must be set to the client ID of your managed identity you wish to use.

If the AKS cluster has workload identity enabled what will happen is that a special service account token gets mounted in each container of the annotated deployment and env vars are also injected. This allows the Azure Identity SDK to then make a request to authenticate has the managed identity.

I suggest first making sure with another Azure Dapr component that you have workload identity working correctly if you run into issues.

File: common/component/redis/v9client.go [hunk 14]
```diff
@@ -20,13 +20,21 @@ import (
...
 
 type v9Pipeliner struct {
 	pipeliner    v9.Pipeliner
 	writeTimeout Duration
 }
 
+var v9logger = logger.NewLogger("dapr.components.redisv9")
+
+const (
+	tokenRefreshInterval = 50 * time.Minute
```
berndverst, line 35: Apparently we do not have to do this every hour -- instead we should examine the most recent token expiration / duration, and then refresh the token say 10 minutes before it expires.

The token validity is variable and can be longer than one hour.

There should be a function you can use to examine the validity of the token.

shpathak-msft:
@sadath-12 and @ItalyPaleAle apologies for the confusion in the other issue, adding some clarifications here. Redis server does not care what is the lifetime of the token - it can be 1 hour or 24 hours. The only expectation is that the redis server receives an auth command before the previous token expires. I would recommend not hard coding the "re-auth" loop to run every 50 minutes but it should run ~10 minutes before the token expiry.

common/component/redis/redis.go
```diff
@@ -252,3 +255,65 @@ type RedisError string
...
+
+				// Authenticate with Redis using the refreshed token
+				var authErr error
+				if version == "v8" {
+					authErr = redisClient.(v8Client).client.Pipeline().Auth(ctx, at.Token).Err()
+				} else if version == "v9" {
+					authErr = redisClient.(v9Client).client.Pipeline().Auth(ctx, at.Token).Err()
+				}
+				if authErr != nil {
+					logger.Error("Failed to authenticate with Redis using refreshed Azure AD token:", authErr)
```
philon-msft, line 309: "Azure AD" is outdated terminology - this should be "Microsoft Entra ID" everywhere

mikeee:
@sadath-12 This PR is being tracked for v1.14, a few comments are outstanding - do you have the bandwidth to pick this up still?

common/component/redis/redis.go
berndverst: for consistency with the rest of Dapr at this time we will keep it as AzureAD. In the future all of Dapr - not just this component - will need to rename things in a single batch update.

common/component/redis/redis.go
```diff
@@ -252,3 +255,65 @@ type RedisError string
 func (e RedisError) Error() string { return string(e) }
 
 func (RedisError) RedisError() {}
+
+func (s *Settings) refreshTokenRoutineForRedis(ctx context.Context, redisClient RedisClient, version string, meta map[string]string, logger logger.Logger) {
+	ticker := time.NewTicker(tokenRefreshInterval)
+	defer ticker.Stop()
+
+	if !s.useAzureAD {
```
berndverst, line 263: I would like to see this exit condition instead be used in the v8client and v9client only.

```go
if s.useAzureAd {
  authenticateWithAzureRedis()
}
```

or something like that

common/component/redis/redis.go
```diff
@@ -252,3 +255,65 @@ type RedisError string
 func (e RedisError) Error() string { return string(e) }
 
 func (RedisError) RedisError() {}
+
+func (s *Settings) refreshTokenRoutineForRedis(ctx context.Context, redisClient RedisClient, version string, meta map[string]string, logger logger.Logger) {
```
berndverst, line 259: I'd rename this to 
```suggestion
func (s *Settings) authenticateWithAzureRedis(ctx context.Context, redisClient RedisClient, version string, meta map[string]string, logger logger.Logger) {
```

common/component/redis/redis.go
```diff
@@ -252,3 +255,65 @@ type RedisError string
...
+	ticker := time.NewTicker(tokenRefreshInterval)
```
berndverst, line 260: The ticker value must be reset every time you get a new token. The value must be token expiration time minus 10 minutes (so that we refresh 10 minutes before expiration).

common/component/redis/redis.go
```diff
@@ -252,3 +255,65 @@ type RedisError string
...
+				if err != nil {
+					logger.Error("Failed to get Azure AD token:", err)
+					retry++
+					time.Sleep(5 * time.Second)
+					continue
+				}
+
+				// Authenticate with Redis using the refreshed token
+				var authErr error
+				if version == "v8" {
```
berndverst, line 303: No, this is bad.

Instead the `Auth` method should be added to the interface in `redis.go`, then an implementation should be added for the `v8client.go` and `v9client.go`

That way you can then simply call `client.Auth` without using the Pipeline client and without using version specific clients in this code. The version should not be required here.

References:
https://pkg.go.dev/github.com/go-redis/redis/v8#Conn.Auth
https://pkg.go.dev/github.com/go-redis/redis/v9#Conn.Auth

common/component/redis/redis.go
```diff
@@ -252,3 +255,65 @@ type RedisError string
...
+				tokenCred, err := env.GetTokenCredential()
+				if err != nil {
+					logger.Error("Failed to get Azure AD token credential:", err)
+					retry++
+					time.Sleep(5 * time.Second)
+					continue
+				}
+				at, err := tokenCred.GetToken(ctx, policy.TokenRequestOptions{
+					Scopes: []string{
+						env.Cloud.Services[azure.ServiceOSSRDBMS].Audience + "/.default",
```
berndverst, line 291: @ItalyPaleAle  does this `dapr.io/oss-rdbms` environment setting always exist? I don't think so?

berndverst (changes requested):

Issue #2539, opened 2023-02-16
GCP PubSub resulting in excessive "Administrator operations per minute" on publish


CafeLungo:
Our team is running into a problem when we start load testing Dapr using GCP PubSub to publish message. There is an admin action, which we believe is part of the `ensureTopic` call on every single publish when we are using `disableEntityManagement: False` on our `pubsub.gcp.pubsub` Component.

## Expected Behavior

* When publishing messages to GCP PubSub, the extra administrator action of `getTopic` should not be excessive.
* Normal limit in GCP is only 6000/min
<img width="548" alt="image" src="https://user-images.githubusercontent.com/11787050/219384214-d03eb8eb-e263-4219-bbbf-8f932c34acf4.png">

## Actual Behavior

* The call to `getTopic` is called multiple times in the process of publishing a message: https://github.com/dapr/components-contrib/blob/master/pubsub/gcp/pubsub/pubsub.go#L233-L261
  * It is called via `ensureTopic()`
  * And again in the after that to publish to the topic.
* Our team believes this is excessive:
  * `ensureTopic` may be too aggressive to check that the topic exists on every single publish.
  * The result can likely be cached
  * Or when publishing to the topic via the `Topic` reference, the `PublishResult` should have an error if the topic doesn't exist.

## Steps to Reproduce the Problem

* Perform some load testing with some dapr message publishes connected to GCP PubSub, enough to exceed 6k publishes in a minute. May require many parallel containers/pods to achieve this.
```
with DaprClient() as client:
    # Publish events using Dapr PubSub
    for e in _events:
          client.publish_event(
              pubsub_name="gcp-pubsub",
              topic_name=topic,
              data=serialize_to_json_str(e),
              data_content_type="application/json",
          )
```
* Eventually, you may get a warning from GCP for exceeding Administrator operations per minute:
```
ResourceExhausted desc = Quota exceeded for quota metric 'Administrator operations' and limit 'Administrator operations per minute' of service 'pubsub.googleapis.com' for consumer 'project_number:XXXXXXXXXXXXX'
```
* And this results in an error such as this when publishing:
```
[...]

PR #3241, opened 2023-11-20
GCP Pubsub: cache topics

sadath-12:
# Description

Now we don't call ensureTopic function unnecessarily , we cache it and continue if its called before

## Issue reference

Please reference the issue this PR will close: #2539 

## Checklist

* [x] Code compiles correctly
* [] Created/updated tests


common/component/azure/servicebus/client.go
```diff
@@ -38,14 +38,16 @@ type Client struct {
...
+	topicCache  map[string]bool
 }
 
 // NewClient creates a new Client object.
 func NewClient(metadata *Metadata, rawMetadata map[string]string) (*Client, error) {
 	client := &Client{
-		metadata: metadata,
-		lock:     &sync.RWMutex{},
-		senders:  make(map[string]*servicebus.Sender),
+		metadata:   metadata,
```
JoshVanL, line 47: The original issue was about GCP- why do we want to do it for Azure too?

pubsub/gcp/pubsub/pubsub.go
```diff
@@ -175,11 +176,12 @@ func (g *GCPPubSub) Publish(ctx context.Context, req *pubsub.PublishRequest) err
 		return errors.New("component is closed")
 	}
 
-	if !g.metadata.DisableEntityManagement {
+	if !g.metadata.DisableEntityManagement && !g.topicCache[req.Topic] {
```
JoshVanL, line 179: Don't we want to invalidate the cache after some time so that it still checks periodically?

JoshVanL (changes requested):

pubsub/gcp/pubsub/pubsub.go
```diff
@@ -175,11 +176,12 @@ func (g *GCPPubSub) Publish(ctx context.Context, req *pubsub.PublishRequest) err
...
 		err := g.ensureTopic(ctx, req.Topic)
 		if err != nil {
 			return fmt.Errorf("%s could not get valid topic %s, %s", errorMessagePrefix, req.Topic, err)
 		}
+		g.topicCache[req.Topic] = true
```
ItalyPaleAle, line 184: All of these calls that read or set in the cache are not concurrent-safe and have race conditions.

Please use a sync.RWMutex around them.

common/component/azure/servicebus/client.go
sadath-12: Its actually the function ```ensureTopic``` that has the potential to bottleneck right ? so if any service calling it can potentially endup bottle necking . Correct my thought process please

pubsub/gcp/pubsub/pubsub.go
sadath-12: Good Catch !. Any suggestions for how much time ?

pubsub/gcp/pubsub/pubsub.go
sadath-12: Good catch

pubsub/gcp/pubsub/pubsub.go
sadath-12: Does something like 5 hours work ?

common/component/azure/servicebus/client.go
```diff
@@ -38,14 +38,16 @@ type Client struct {
...
+		lock:       &sync.RWMutex{},
+		senders:    make(map[string]*servicebus.Sender),
+		topicCache: make(map[string]bool),
```
ItalyPaleAle, line 50: Use `map[string]struct{}` since the value doesn't matter. You will then check if the value is in the cache by doing:

`if _, ok := topicCache[key]; ok`

common/component/azure/servicebus/client.go
```diff
@@ -121,13 +123,14 @@ func (c *Client) GetSender(ctx context.Context, queueOrTopic string, ensureFn en
 	}
 
 	// Ensure the queue or topic exists, if needed
-	if ensureFn != nil {
+	if ensureFn != nil && !c.topicCache[queueOrTopic] {
```
ItalyPaleAle, line 126: Looks like this Azure component already has a cache. See above on L110 where it checks `c.senders` (btw this is a good example of how to use locks to prevent race conditions :) )

pubsub/gcp/pubsub/pubsub.go
```diff
@@ -54,9 +54,10 @@ type GCPPubSub struct {
 	metadata *metadata
 	logger   logger.Logger
 
-	closed  atomic.Bool
-	closeCh chan struct{}
-	wg      sync.WaitGroup
+	closed     atomic.Bool
+	closeCh    chan struct{}
+	wg         sync.WaitGroup
+	topicCache map[string]bool
```
ItalyPaleAle, line 60: See my other comment

```suggestion
	topicCache map[string]struct{}
```

common/component/azure/servicebus/client.go
sadath-12: lol how I dint look at that

sadath-12:
@ItalyPaleAle , Does the structure I used now fine ? supporting cache invalidation

pubsub/gcp/pubsub/pubsub.go
```diff
@@ -354,9 +416,21 @@ func (g *GCPPubSub) getTopic(topic string) *gcppubsub.Topic {
...
 
 func (g *GCPPubSub) ensureSubscription(parentCtx context.Context, subscription string, topic string) error {
-	err := g.ensureTopic(parentCtx, topic)
-	if err != nil {
-		return err
+	g.lock.RLock()
+	_, exists := g.topicCache[topic]
+	_, DExists := g.topicCache[g.metadata.DeadLetterTopic]
+	g.lock.RUnlock()
+	if !exists {
```
ItalyPaleAle, line 423: This is almost good. The problem here is if `ensureSubscription` is called by 2 callers in parallel:

1. Caller A will get a read lock to check that the topic exists
2. Caller B does the same
3. Both see the topic does NOT exist
4. Both will create the topic
5. Both will save it in `topicCache

This is a bit wasteful.

The common pattern is:

1. Acquire a read lock and check if the topic exists
2. If the topic exists, all good
3. If it does NOT exist, acquire a write lock **before** calling `g.ensureTopic`. This means that there won't be > 1 call to `ensureTopic` at the same time
4. After getting the write lock, **before** calling `g.ensureTopic`, check again if the value exists in the cache. This is because by the time we acquire the lock, someone else may have created the topic in the cache.

You can find multiple examples of this pattern in the codebase!

Same feedback applies to the other instances :)

sadath-12: Done ! Is it because this function is bit long so it might be prone to happening like that ? because or for every function we might endup doing like first ```RLock``` and then for updating we read it again ?

pubsub/gcp/pubsub/pubsub.go
```diff
@@ -76,9 +83,46 @@ type WhatNow struct {
 	Type string `json:"type"`
 }
 
+const (
+	cacheRefreshInterval = 5 * time.Hour
```
JoshVanL, line 87: Please be specific with variable names.

```suggestion
	topicCacheRefreshInterval = 5 * time.Hour
```

pubsub/gcp/pubsub/pubsub.go
```diff
@@ -76,9 +83,46 @@ type WhatNow struct {
...
+		case <-ticker.C:
+			g.lock.Lock()
+			// Clear or refresh the cache entries here
+			g.clearExpiredCacheEntries()
+			g.lock.Unlock()
+		}
+	}
+}
+
+func (g *GCPPubSub) clearExpiredCacheEntries() {
```
JoshVanL, line 120: Don't think this needs to be a dedicated function and can be moved inline.

pubsub/gcp/pubsub/pubsub.go
```diff
@@ -354,9 +416,25 @@ func (g *GCPPubSub) getTopic(topic string) *gcppubsub.Topic {
 }
 
 func (g *GCPPubSub) ensureSubscription(parentCtx context.Context, subscription string, topic string) error {
-	err := g.ensureTopic(parentCtx, topic)
-	if err != nil {
-		return err
+	g.lock.RLock()
+	_, TExists := g.topicCache[topic]
+	_, DExists := g.topicCache[g.metadata.DeadLetterTopic]
```
JoshVanL, line 421: We don't capitalise local variable names. It is also convention to use `ok` as the name for the bool returned when checking a map.

```suggestion
	_, topicOK := g.topicCache[topic]
	_, dlTopicOK := g.topicCache[g.metadata.DeadLetterTopic]
```

pubsub/gcp/pubsub/pubsub.go
```diff
@@ -54,9 +54,16 @@ type GCPPubSub struct {
...
-	wg      sync.WaitGroup
+	closed     atomic.Bool
+	closeCh    chan struct{}
+	wg         sync.WaitGroup
+	topicCache map[string]cacheEntry
+	lock       *sync.RWMutex
+}
+
+type cacheEntry struct {
+	Exists   bool
```
JoshVanL, line 65: This doesn't look used.

JoshVanL (changes requested):
Please add tests for this change.

pubsub/gcp/pubsub/pubsub.go
```diff
@@ -76,9 +82,37 @@ type WhatNow struct {
 	Type string `json:"type"`
 }
 
+var topicCacheRefreshInterval = 5 * time.Hour
```
ItalyPaleAle, line 85:
```suggestion
const topicCacheRefreshInterval = 5 * time.Hour
```

pubsub/gcp/pubsub/pubsub.go
```diff
@@ -76,9 +82,37 @@ type WhatNow struct {
...
+		logger:     logger,
+		closeCh:    make(chan struct{}),
+		topicCache: make(map[string]cacheEntry),
+		lock:       &sync.RWMutex{},
+	}
+	return client
+}
+
+func (g *GCPPubSub) periodicCacheRefresh() {
+	ticker := time.NewTicker(topicCacheRefreshInterval)
```
ItalyPaleAle, line 99: Should probably be something like half of the time or less, because otherwise values can be cached for a lot longer. For example, assume a value is created 1s after the ticker, then it will live for almost 10 hours (minus 1s)

ItalyPaleAle:
```suggestion
	ticker := time.NewTicker(topicCacheRefreshInterval / 5)
```

pubsub/gcp/pubsub/pubsub.go
```diff
@@ -174,12 +214,20 @@ func (g *GCPPubSub) Publish(ctx context.Context, req *pubsub.PublishRequest) err
 	if g.closed.Load() {
 		return errors.New("component is closed")
 	}
+	g.lock.RLock()
+	_, topicExists := g.topicCache[req.Topic]
+	g.lock.RUnlock()
 
-	if !g.metadata.DisableEntityManagement {
+	if !g.metadata.DisableEntityManagement && !topicExists {
```
ItalyPaleAle, line 221: Just adding a comment here for future people:

```suggestion
	// We are not acquiring a write lock before calling ensureTopic, so there's the chance that ensureTopic be called multiple times
    // This is acceptable in our case, even is slightly wasteful, as ensureTopic is idempotent
	if !g.metadata.DisableEntityManagement && !topicExists {
```

pubsub/gcp/pubsub/pubsub.go
```diff
@@ -210,12 +258,20 @@ func (g *GCPPubSub) Subscribe(parentCtx context.Context, req pubsub.SubscribeReq
...
```
ItalyPaleAle, line 265: Just adding a comment here for future people:

```suggestion
	// We are not acquiring a write lock before calling ensureTopic, so there's the chance that ensureTopic be called multiple times
    // This is acceptable in our case, even is slightly wasteful, as ensureTopic is idempotent
	if !g.metadata.DisableEntityManagement && !topicExists {
```

common/component/azure/servicebus/client.go
```diff
@@ -394,4 +394,4 @@ func notEqual(a, b *bool) bool {
 		return true
 	}
 	return *a != *b
-}
+}
```
ItalyPaleAle, line 397: Reverting this change
```suggestion
}

```

pubsub/gcp/pubsub/pubsub_test.go
```diff
@@ -128,3 +131,35 @@ func TestInit(t *testing.T) {
 		assert.ErrorContains(t, err, "connectionRecoveryInSec")
 	})
 }
+
+func TestTopicCache(t *testing.T) {
```
ItalyPaleAle, line 135: I suggest removing this test as it depends on a wall clock and a `sleep`, so it will be flaky (we have seen this in a lot of tests...). This can be added to the certification tests in the future.

ItalyPaleAle (changes requested):

pubsub/gcp/pubsub/pubsub.go
```diff
@@ -76,9 +82,37 @@ type WhatNow struct {
...
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-g.closeCh:
+			return
+		case <-ticker.C:
+			g.lock.Lock()
+			for key, entry := range g.topicCache {
+				if time.Since(entry.LastSync) > topicCacheRefreshInterval/5 {
```
ItalyPaleAle, line 109:
```suggestion
                // Delete from the cache if the last sync was longer than topicCacheRefreshInterval
				if time.Since(entry.LastSync) > topicCacheRefreshInterval {
```

ItalyPaleAle (commented):
Just some nits

ItalyPaleAle (dismissed):

ItalyPaleAle:
Thank you @sadath-12 !

Issue #3247, opened 2023-11-23
Kafka Sarama client, optimize by not loading all metadata


famarting:
## Describe the feature
The kafka client used for the kafka components implementation by default has a feature that loads all the topics metadata on initialization, and keeps in memory the metadata for all the topics in the kafka cluster. This is inefficient and memory hungry in scenarios where the kafka cluster used has many topics.

The config property can be found here https://github.com/IBM/sarama/blob/main/config.go#L157

I suggest Dapr always disables that feature in the Sarama client

## Release Note

RELEASE NOTE: **UPDATE** Optimize kafka components for clusters with many topics

RELEASE NOTE:

famarting:
not applicable

olitomlinson:
I think its worth re-opening this issue. 

I'm not entirely sure why the associated PR was closed tbh, it could be viable!

For context : we have a use case where thousands of partitions on the Kafka cluster are related to Kafka Topics that are never going to be used by Dapr sidecars. So having Dapr load all this metadata on init and refresh it is highly wasteful.

cc @artursouza

famarting:
I closed the issue because I realized the changes needed to make it work were much larger than what I anticipated...

My hope was that by simply changing a flag, sarama would lazyly read the topic metadata as they are used by produce or consume calls... however that was not the case since I discovered the hard way that sarama requires you to pass the list of topics you'll want to use, maybe after providing an initial list sarama can lazy load new topics... but I didnt get to that point

The issue could be reopened or a new one could be created with a more specific requirement or at least after some more research on what can be done with sarama in terms of optimizing loading the topics metadata

famarting:
I'm playing a bit with the sarama client again and I think I was suffering from this two issues

https://github.com/IBM/sarama/issues/2630
https://github.com/IBM/sarama/issues/2069

famarting:
this has not actually been addressed, I found a way of fixing this, it will require a contribution to the sarama client first

famarting:
in case anyone sees this, I already have all the code in place but first we need my contribution here to be accepted and released https://github.com/IBM/sarama/pull/2830

PR #3248, opened 2023-11-23
kafka - disable keeping metadata for all topics

famarting:
# Description

Small change to the config of the kafka client used in all kafka components

## Issue reference

Please reference the issue this PR will close: #3247

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/common/component/kafka/kafka.go
+++ b/common/component/kafka/kafka.go
@@ -78,6 +78,8 @@ func (k *Kafka) Init(_ context.Context, metadata map[string]string) error {
 	config := sarama.NewConfig()
 	config.Version = meta.internalVersion
 	config.Consumer.Offsets.Initial = k.initialOffset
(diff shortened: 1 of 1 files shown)


famarting:
seems like this config option cannot be used without providing a known list of kafka topics...

artursouza:
> seems like this config option cannot be used without providing a known list of kafka topics...

Can that be passed as a parameter? Maybe there are users that have that scenario.

artursouza:
> @artursouza @yaron2
> 
> 
> 
> I think its worth reconsidering if there is a possible implementation here.
> 
> 
> 
> For context : we have a use case where thousands of partitions on the Kafka cluster are related to Kafka Topics that are never going to be used by Dapr sidecars. So having Dapr load all this metadata on init and refresh it is highly wasteful.
> 
> 
> 
> 
> 
> 
> 
> 

I can consider one. Can you track your request in an issue? We can tag it and get some traction from the community.

Issue #2529, opened 2023-02-14
Support YugabyteDB as a State Store


ashetkar:
## Describe the feature
Support YugabyteDB as a State Store.

About YugabyteDB:
YugabyteDB is a transactional distributed SQL database for cloud-native applications. It is wire-compatible with Postgresql. More details [here](https://docs.yugabyte.com/preview/).

## Release Note
RELEASE NOTE: **ADD** support for YugabyteDB as a state store in Dapr.

berndverst:
If it is wire compatible with Postgresql, can you not simply use the Postgres state store and point it at YugabyteDB? That would be preferred @ashetkar

berndverst:
According to the docs the GO library `lib/pq` works to connect. We use `jackc/pgx` which supports the same required authentication method.

For this reason I do not think a dedicated Dapr component is required to connect to YugabyteDB and we also do not need to update our documentation.

Please just use the Postgres state store.

ashetkar:
@berndverst Thank you for the response.

The Postgresql component of Dapr makes use of `xmin` column for etags and YugabyteDB does not support `xmin`, hence it cannot use the Postgresql component to connect. And there may be other incompatibilities existing.
Also, we may be able to add YugabyteDB-specific optimisations like using our fork of jackc/pgx driver, etc.

Even though this issue was filed to request the support, it was to be followed up with a PR from us that we have already been working on. And we would like to work with Dapr team to complete this support.

Given this, please consider retaining this GH Issue. Do let me know if you need more details, thanks.

artursouza:
I am reopening the issue for further discussion.

How do you solve the etag in the new component? What ate the other incompatibilities? I am wondering if there is a way to make the existing Postgres component configurable.

Now, if you have a fork of the jackc/pgx driver that you would like to use, have you considered making this a pluggable component? We do have a go-sdk that make this simple if the component is written with the same interface as components-contrib: http://github.com/dapr-sandbox/components-go-sdk

berndverst:
@ashetkar I would prefer if we add whatever capabilities are needed to the existing Postgres state store.

@ItalyPaleAle created the TTL implementation. We can certainly change it, or conditionally enable  / disable it to create compatibility.

Duplicating the Postgres libraries in our code means more memory utilization for something which isn't strictly necessary.

If you must create a new component  I strongly suggest using the version of pgx we are already importing.

berndverst:
[...]

PR #3250, opened 2023-11-28
PostgreSQL state store v2

### LINKED ISSUE #2956: PostgreSQL state store v2
As we now have "v2" components, we should look into creating a v2 of the Postgres state store to address some long-standing issues:

- Do not use the xmin column for etags, but a more "traditional" column
- The implementation of the "v2" component will be 100% identical to the current CockroachDB implementation, and we can also expose it as YugabyteDB (#2529). We should be able to reuse pretty much all code, but simplify the migration code.
- We do not plan on deprecating the "v1" at present
- Tests (conformance and certification) should run for both v1 and v2

ItalyPaleAle:
Fixes #2956

This PR introduces the v2 of the PostgreSQL state store component, which is meant to live side-by-side with v1. The implementation is very different and the data stored in the state store will not be compatible. However, to use the "v2" of the state store, users will need to explicitly opt into that, so existing apps will not break.

> There's currently no way to update from a "v1" to "v2" state store. If that were the needed, we could look into doing that in the future.

The main changes compared to v1 are:

- There is only 1 implementation that is used by all PostgresSQL-compatible databases, including CockroachDB.
  - This will work also for YugabyteDB (#2529).
  - These components do not have their own folder anymore, but will be defined as aliases in the runtime
- The data is now stored as binary and not in a JSON column.
   - This should improve performance quite a bit when the data stored is binary, such as all the data that comes in through the gRPC API (previously, that was base64-encoded every time) . It also improves performance in general due to using an opaque column type rather than JSONB (since data in JSONB columns is always validated to be correct JSON).
   - The biggest consequence of the above is that the v2 of this component does not support the state store query APIs. This is by design, as those APIs are being deprecated
- All implementations now use a UUID as etag, rather than relying on the xmin column (which was always a bit "hacky" and only worked on PostgreSQL), or a deterministic integer.

The new component has conformance and certification tests enabled.

PS: In both "v2" and "v1", there are some aliases for metadata properties: `cleanupInterval = cleanupIntervalInSeconds` (and it accepts a Go duration), `timeout = timeoutInSeconds` (and it accepts a Go duration), `connectionString = url`


common/component/sql/migrations/postgres/postgres_migrations.go
```diff
@@ -37,33 +37,58 @@ type Migrations struct {
 
 // Perform the required migrations
 func (m Migrations) Perform(ctx context.Context, migrationFns []commonsql.MigrationFn) error {
-	// Use an advisory lock (with an arbitrary number) to ensure that no one else is performing migrations at the same time
-	// This is the only way to also ensure we are not running multiple "CREATE TABLE IF NOT EXISTS" at the exact same time
-	// See: https://www.postgresql.org/message-id/CA+TgmoZAdYVtwBfp1FL2sMZbiHCWT4UPrzRLNnX1Nb30Ku3-gg@mail.gmail.com
-	const lockID = 42
+	// Ensure the metadata table exists
```
ItalyPaleAle, line 40: Reviewers: I had to change how we lock before migrations because advisory locks aren't available in CockroachDB. This is functionally equivalent (and the certification tests are validating it), but because we call EnsureMetadataTable regardless of the lock, that operation may fail and may need to be retried automatically (see implementation for EnsureMetadataTable)

berndverst:
Fix the conflict please :)

Also just to confirm, v2 components don't have query API right?

ItalyPaleAle:
> Fix the conflict please :)

Done!

> Also just to confirm, v2 components don't have query API right?

Correct, they do not.

olitomlinson:
@ItalyPaleAle 

> The data is now stored as binary and not in a JSON column.

Does this mean there is no convenient way to inspect the value of what has been stored against a key using something like PgAdmin?

I mean, I get it, and I get the reasoning, but this is a -1 for convenience. 

My developers have often relied on inspecting the state directly when getting to grips with Dapr, and still do dive into the data layer at times when debugging. Sounds like this would not be possible on V2?

ItalyPaleAle:
> Does this mean there is no convenient way to inspect the value of what has been stored against a key using something like PgAdmin?

Still can, it shouldn't be significantly more inconvenient. In some cases it may even be more convenient.

First, one thing to keep in mind, is that Dapr is a bit of a mess in the way it's passing data to the state stores.

- HTTP APIs store data by passing objects that will be encoded as JSON
- gRPC APIs pass `[]byte` to the state store

In v1 of the component, data from gRPC would be interpreted as "binary" and stored in a JSONB column base64-encoded. So if you sent `"hi!"` using gRPC, in the database you'd get `"aGkh"` (exactly like that, with quotes around the base64-encoded value)

In v2 everything is stored as-is in a BYTEA column. If the data is a `[]byte` it is stored as-is. If it's an object, it's serialized as JSON, which produces `[]byte`, and stored as-is.

pgAdmin will show the column as binary by default, but assuming you know the value is UTF-8-encoded text (as it the case for JSON data), then you can do:

```sql
SELECT convert_from(value, 'utf-8') FROM state
```

state/postgresql/v2/postgresql.go
```diff
@@ -0,0 +1,506 @@
...
+	s := &PostgreSQL{
+		logger:        logger,
+		enableAzureAD: !opts.NoAzureAD,
+	}
+	s.BulkStore = state.NewDefaultBulkStore(s)
+	return s
+}
+
+// Init sets up Postgres connection and performs migrations
+func (p *PostgreSQL) Init(ctx context.Context, meta state.Metadata) error {
```
DeepanshuA, line 74: Can we include somewhere here (apart from the docs) - the diff with v1 and what are the benefits of using v2 - just, so that a developer can upfront see the reason for a v2 here.

state/postgresql/v2/postgresql.go
```diff
@@ -0,0 +1,506 @@
...
+	// Get all keys
+	keys := make([]string, len(req))
+	for i, r := range req {
+		keys[i] = r.Key
+	}
+
+	// Execute the query
+	query := `SELECT
+	key, value, etag, expires_at
+FROM ` + p.metadata.TableName(pgTableState) + `
```
DeepanshuA, line 357: nit: identation?

DeepanshuA (commented):
It mostly LGTM.
I have some small comments only.

common/component/sql/migrations/postgres/postgres_migrations.go
```diff
@@ -37,33 +37,58 @@ type Migrations struct {
 
 // Perform the required migrations
 func (m Migrations) Perform(ctx context.Context, migrationFns []commonsql.MigrationFn) error {
```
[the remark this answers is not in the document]
ItalyPaleAle: No, this is for schema migration. They are uncommon and normally happen on major Dapr updates only. We were already using it, just had to make some changes to support non-Postgres databases.

state/postgresql/v2/postgresql.go
ItalyPaleAle: Fixed

state/postgresql/v2/postgresql.go
ItalyPaleAle: Done, in the comment on the NewPostgreSQLStateStore method

berndverst (approved):

PR #3253, opened 2023-12-02
SQLite: allow enabling foreign keys in `GetConnectionString`

ItalyPaleAle:
The `_pragma=foreign_keys(1)` option should be added only if the component requires it. Users should not be allowed to specify that.


berndverst (approved):
Could this be a breaking change for some users?

ItalyPaleAle:
> Could this be a breaking change for some users?

I would say potentially but it would require people to have been doing something "odd". The current component doesn't require foreign keys, so if someone were enabling that, it would have been them trying to do something unusual.

Issue #3269, opened 2023-12-14
Pulsar Add More TLS Connection Options


toneill818:
## Describe the feature
Allow setting the following options when connecting to Apache Pulsar

- TLSAllowInsecureConnection
- TLSValidateHostname

## Release Note

RELEASE NOTE: **ADD** Pulsar TLS Options AllowInsecureConnection and ValidateHostname

Issue #1860, opened 2022-07-07
[pubsub][pulsar]Support pulsar+ssl:// URL format


Kyle2123:
## Describe the feature
Currently the pulsar pubsub component expects the Pulsar url to be in one of three formats:
- pulsar://
- http://
- https://

Please add support for the pulsar+ssl:// format to allow for connections to production Pulsar clusters that are configured to use TLS.

https://github.com/dapr/components-contrib/blob/3821c001315dfd2c24f8e1c6bdd19dd1ff21475d/pubsub/pulsar/pulsar.go#L49

https://github.com/dapr/components-contrib/blob/3821c001315dfd2c24f8e1c6bdd19dd1ff21475d/pubsub/pulsar/pulsar.go#L164-L167

ItalyPaleAle:
Hey @Kyle2123  thanks for opening this issue! Would you be able to submit a PR for this?

PR #3270, opened 2023-12-14
Pulsar PubSub: allow TLS configuration for hostname and insecure connections

toneill818:
# Description

When connecting to pulsar with TLS enabled set the prefix to be `pulsar+ssl://` if a prefix was not set when defining the host. Allow setting the `TLSAllowInsecureConnection` and `TLSValidateHostname` options when connecting to pulsar. They default to the pulsar client's default. 

## Issue reference

Please reference the issue this PR will close: #1860
Please reference the issue this PR will close: #3269

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#3903


berndverst (changes requested):
Thanks for the contribution.

This PR needs an integration test (certification test) at:
https://github.com/dapr/components-contrib/tree/main/tests/certification/pubsub/pulsar

You can take a look at some other certification tests that specifically test the TLS modes for other components. The hardest part will be creating a docker compose file / setup which spins up Pulsar with the TLS configuration that you need to test. Then you need a Dapr component configuration which actually sets the appropriate new TLS settings. For the actual operations you can just reuse the code that exists in the basic test today.

Note that the existing test run should not be updated to use TLS, this needs to be a separate test run.

toneill818:
@berndverst Integration tests have been added for pulsar

yaron2:
@toneill818 the Pulsar certification test is failing - see the build check

yaron2:
@toneill-newinnov the pulsar certification test is still failing

toneill818:
@yaron2 I'll keep looking into this. It isn't timing out while I run the tests locally.

pubsub/pulsar/pulsar.go
```diff
@@ -538,3 +547,18 @@ func isValidPEM(val string) bool {
...
 }
+
+func getPulsarURL(m *pulsarMetadata) string {
+	pulsarURL := m.Host
+	if !strings.HasPrefix(m.Host, "http://") &&
+		!strings.HasPrefix(m.Host, "https://") &&
+		!strings.HasPrefix(m.Host, "pulsar+ssl://") &&
+		!strings.HasPrefix(m.Host, "pulsar://") {
+		if m.EnableTLS {
+			pulsarURL = fmt.Sprintf("%s%s", pulsarSSLPrefix, m.Host)
```
ItalyPaleAle, line 558:
```suggestion
			pulsarURL = pulsarSSLPrefix + m.Host
```

pubsub/pulsar/pulsar.go
```diff
@@ -538,3 +547,18 @@ func isValidPEM(val string) bool {
...
+		} else {
+			pulsarURL = fmt.Sprintf("%s%s", pulsarPrefix, m.Host)
```
ItalyPaleAle, line 560:
```suggestion
			pulsarURL = pulsarPrefix + m.Host
```

ItalyPaleAle (commented):
Normally, test failures in cert tests like these, which happen only on the CI, are due to a race condition _somewhere_ in the code.

I am not sure where specifically. My advice would be to try running one of the tests at a time, multiple times, and see if you can at least identify _where_ this is happening :(

Thanks for looking into this!

tests/certification/pubsub/pulsar/pulsar_test.go
```diff
@@ -1243,6 +1256,15 @@ func (p *pulsarSuite) client(t *testing.T) (pulsar.Client, error) {
 	opts := pulsar.ClientOptions{
 		URL: "pulsar://localhost:6650",
 	}
+
+	if p.tlsEnabled {
+		opts = pulsar.ClientOptions{
+			URL:                        "pulsar+ssl://localhost:6651",
+			TLSAllowInsecureConnection: true,
```
berndverst, line 1263: Isn't this the opposite of what you want?

tests/certification/pubsub/pulsar/pulsar_test.go
```diff
@@ -190,6 +192,17 @@ func TestPulsar(t *testing.T) {
...
 			services:          []string{"zookeeper", "pulsar-init", "bookie", "broker"},
+			tlsEnabled:        false,
+		})
+	})
+
+	t.Run("TLS", func(t *testing.T) {
+		suite.Run(t, &pulsarSuite{
+			authType:          "none",
+			dockerComposeYAML: dockerComposeTLSYAML,
+			componentsPath:    "./components/tls",
```
berndverst, line 203: You have six very differently configured pulsar components loaded here. How do you verify that ALL of them actually worked?

And if your server (docker file) is configured to require a certificate, then you really shouldn't be able to connect using an insecure mode like in `consumer_four` for example.

berndverst (changes requested):

tests/certification/pubsub/pulsar/pulsar_test.go
toneill818: No, this PR add the ability to connect to pulsar over an SSL connection with the TLSAllowInsecureConnection (self signed cert)

tests/certification/pubsub/pulsar/pulsar_test.go
toneill818: All of the configs have `enableTSL: true`. They are connecting over a TLS connection to send and receive messages over a TLS connection but allow us to skip Hostname validation on the key and setting `tlsAllowInsecureConnection: true` since the certificate isn't signed by an approved CA.

yaron2:
@toneill818 v the certification tests are all failing, please take a look

Issue #2173, opened 2022-10-11
Pub Sub Polling interval for Kafka


rahulpoddar-fyndna:
The subscriber polls to get the messages from subscription topic in kafka. The interval at which request is made to get the new message is not configurable as of now. Making this parameter configurable will help in tuning the  application for performance needs.

@pravinpushkar

yaron2:
Moving to correct repo.

ctoomey:
There's a huge number of configuration settings for kafka [consumers](https://kafka.apache.org/documentation/#consumerconfigs) and [producers](https://kafka.apache.org/documentation/#producerconfigs). E.g. a couple more that we use are `partition.assignment.strategy` and `client.dns.lookup`.

Ideally all these settings should be configurable, instead of just a very limited fixed subset of them.

rabollin:
Currently Pub-sub configurations are handled by sarama library[https://pkg.go.dev/github.com/shopify/sarama] and there is not much scope for the App developers to edit the configurations. I think extending the implementation to allow configuration edits may help.

berndverst:
> Currently Pub-sub configurations are handled by sarama library - https://pkg.go.dev/github.com/shopify/sarama and there is not much scope for the App developers to edit the configurations. I think extending the implementation to allow configuration edits may help.

Allowing all configuration options becomes impossible to maintain and test for. We will want to add the minimum necessary options users actually need.

Dapr isn't intended to be a passthrough client that allows configuring every single setting.

ctoomey:
It'd be REALLY hard to decide which subset of config. settings should be supported and which should not, as they all are important in different situations. If the goal is to support developers moving from kafka-specific clients that support all config. settings (which it should be), I'd suggest allowing and passing through all config. settings. Suggest having them in a separate section in the config. file vs. having them mixed in with the dapr config. settings.

berndverst:
If there is a way to pass through the settings the best we can probably do is providing a single metadata field which would be a comma separated list of key value pairs. And you'd have to refer to the Sarama library to get a list of supported settings. We will not separately document the supported options, but we can probably link to it from the docs.

Still I do not know whether this is possible.

By the way, as a maintainer of the components I do not think that the goal is for those to replace the use of native clients for advanced use cases / configuration scenario. Dapr is an abstraction and you always lose some ability to customize things with each abstraction.

rabollin:
/ keep-alive

rabollin:
/ keep-alive

[...]

PR #3274, opened 2023-12-14
Add metadata configs to kafka pub/sub

yaron2:
Partially closes https://github.com/dapr/components-contrib/issues/2173#issuecomment-1835795781.


yaron2:
Based on wrong branch

PR #3275, opened 2023-12-14
Add kafka configuration to pub/sub

yaron2:
Partially closes https://github.com/dapr/components-contrib/issues/2173#issuecomment-1835795781.


artursouza (approved):

common/component/kafka/metadata.go
```diff
@@ -257,5 +266,32 @@ func (k *Kafka) getKafkaMetadata(meta map[string]string) (*KafkaMetadata, error)
 		m.internalVersion = version
 	}
 
+	if val, ok := meta[channelBufferSize]; ok && val != "" {
```
ItalyPaleAle, line 269: @yaron2 looking at this PR that was merged, this is not done correctly as it's not case-insensitive. Can you please create a quick fix-up that uses GetMetadataProperty? https://pkg.go.dev/github.com/dapr/kit/metadata#GetMetadataProperty

PR #3278, opened 2023-12-18
Chore: use a single package for all Feature structs

ItalyPaleAle:
Also replaces `golang.org/x/exp/slices` with `slices` from the Go 1.21 stdlib

> small spin-off from #3046


berndverst:
Where is the tracking issue for this? Why are we doing this? I would like some explanation - no PRs without discussion.

berndverst (changes requested):
Background needed

ItalyPaleAle:
> Where is the tracking issue for this?

Just a refactoring to avoid code duplication. Instead of having a `Features` struct defined in each package, and the `Contains` method repeated, it's now implemented just once (and re-exported using generics)

berndverst (approved):

Issue #1016, opened 2021-07-17
PostgreSQL binding - Numeric data type column and uuid column response bug fix


TalalMahmoodChaudhry:
## Expected Behavior

1. When using PostgreSQL binding to run a query that gets a column of numeric data type, example:
`select numeric_data_type_column from table`
Then the response should be:
`[[31680], [305300]]`

2. For uuid column type, example query:
`select uuid_type_column from table`
Then the response should be:
`[['c4d239d2-a9c1-41e8-b987-59b6fd228b6f']]`

## Actual Behavior

1. Response for numeric column actually is:
`[[{'Int': 316800, 'Exp': -1, 'Status': 2}], [{'Int': 305300, 'Exp': -1, 'Status': 2}]]`

2. Response for uuid column actually is:
`[[[196, 210, 57, 210, 169, 193, 65, 232, 185, 135, 89, 182, 253, 34, 139, 111]]]`

## Steps to Reproduce the Problem
Create a table with at least one column of numeric data type and another column of uuid. Populate with some data.. Create a PostgreSQL binding. Run a select query for that column.

TalalMahmoodChaudhry:
For now the workaround is:

1. Cast numeric column as float. Then response is like expected.
2. Cast uuid column as text. Then response is like expected.

daixiang0:
Good catch, are you willing to post a PR to fix it?

TalalMahmoodChaudhry:
Unfortunately, it will take me way too long to do this due to work and a part time masters. Additionally, I have never worked in GO so perhaps it's better someone else takes care of this.

For now I would recommend users to use the workaround

PR #3282, opened 2023-12-20
postgres binding: added fix for uuid and numeric datatype columns and added two new methods fetchOne and fetchAll

sridhar-shivraj:
postgres binding: added fix for uuid and numeric datatype columns and added two new methods fetchOne and fetchAll

# Description
Fixed the issue for uuid and numeric datatype columns, earlier uuid was coming as list of strings

Added two new methods

1. FetchOne: returns one row as map with column names
2. FetchAll: returns array of map with column names

## Issue reference

[https://github.com/dapr/components-contrib/issues/1016]

Please reference the issue this PR will close: #_[1016]_

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


bindings/postgres/postgres.go
```diff
@@ -67,6 +73,14 @@ func (p *Postgres) Init(ctx context.Context, meta bindings.Metadata) error {
 	}
 
 	poolConfig, err := m.GetPgxPoolConfig()
+
+	// Register uuid and decimal types with pgx
+	poolConfig.AfterConnect = func(ctx context.Context, conn *pgx.Conn) error {
+		pgxuuid.Register(conn.TypeMap())
```
ItalyPaleAle, line 79: I believe Postgres has built-in support for UUID via the google/uuid package that we use throughout Dapr. I'd rather not add another dependency here if possible

bindings/postgres/postgres.go
```diff
@@ -23,18 +23,24 @@ import (
...
 // List of operations.
 const (
-	execOperation  bindings.OperationKind = "exec"
-	queryOperation bindings.OperationKind = "query"
-	closeOperation bindings.OperationKind = "close"
+	execOperation     bindings.OperationKind = "exec"
+	queryOperation    bindings.OperationKind = "query"
+	closeOperation    bindings.OperationKind = "close"
+	fetchOneOperation bindings.OperationKind = "fetchOne"
+	fetchAllOperation bindings.OperationKind = "fetchAll"
```
ItalyPaleAle, line 43: What's the difference between fetchOne, fetchAll, and query?

ItalyPaleAle (changes requested):

bindings/postgres/postgres.go
sridhar-shivraj: Currently there is an issue with pgx for uuid and numeric columns which returns uuid or numeric columns as a array of strings
[](https://github.com/dapr/components-contrib/issues/1016)

You can find the details about the fix here
[](https://github.com/jackc/pgx/wiki/UUID-Support)

bindings/postgres/postgres.go
sridhar-shivraj: query: returns array of strings (with no column names)
fetchOne: returns map of key value pairs (with column names as key)
fetchAll: returns array of map of key value pairs (with column names as key)

bindings/postgres/postgres.go
ItalyPaleAle: I would rather using something that is based on https://pkg.go.dev/github.com/google/uuid please - that's the package we use throughout Dapr

bindings/postgres/postgres.go
```diff
@@ -212,6 +242,62 @@ func (p *Postgres) exec(ctx context.Context, sql string, args ...any) (result in
...
+	rows, err := p.db.Query(ctx, sql, args...)
+	if err != nil {
+		return nil, fmt.Errorf("error executing query: %w", err)
+	}
+
+	rs := make(map[string]any, 0)
+	for rows.Next() {
+		val, rowErr := rows.Values()
+		if rowErr != nil {
+			return nil, fmt.Errorf("error reading result '%v': %w", rows.Err(), rowErr)
```
ItalyPaleAle, line 256: `rows.Err()` and `rowErr` should be identical per docs

```suggestion
			return nil, fmt.Errorf("error reading row: %w", rowErr)
```

bindings/postgres/postgres.go
```diff
@@ -212,6 +242,62 @@ func (p *Postgres) exec(ctx context.Context, sql string, args ...any) (result in
...
+	rows, err := p.db.Query(ctx, sql, args...)
+	if err != nil {
+		return nil, fmt.Errorf("error executing query: %w", err)
+	}
+
+	rs := make([]map[string]any, 0)
+	for rows.Next() {
+		val, rowErr := rows.Values()
+		if rowErr != nil {
+			return nil, fmt.Errorf("error reading result '%v': %w", rows.Err(), rowErr)
```
ItalyPaleAle, line 283:
```suggestion
			return nil, fmt.Errorf("error reading row: %w", rowErr)
```

bindings/postgres/postgres.go
```diff
@@ -212,6 +242,62 @@ func (p *Postgres) exec(ctx context.Context, sql string, args ...any) (result in
...
 
+func (p *Postgres) fetchOne(ctx context.Context, sql string, args ...any) (result []byte, err error) {
+
+	rows, err := p.db.Query(ctx, sql, args...)
+	if err != nil {
+		return nil, fmt.Errorf("error executing query: %w", err)
+	}
+
+	rs := make(map[string]any, 0)
+	for rows.Next() {
```
ItalyPaleAle, line 253: Since "fetchOne" is meant to fetch a single row, this method should return an error if the query returns more than one row.

Within the `for` loop, add a check to see if this is the 2nd iteration, and stop with an error if it is.

ItalyPaleAle (changes requested):

daixiang0:
ping @sridhar-shivraj

ItalyPaleAle:
ping @sridhar-shivraj

yaron2:
@sridhar-shivraj are you planning to work on this PR further?

PR #3284, opened 2023-12-21
Nameresolvers: add support for ResolveIDMulti

ItalyPaleAle:
This is an optional method that returns a list of addresses.

It is implemented in the K8s nameresolver where it resolves the DNS name and returns a list of addresses

See: https://github.com/dapr/dapr/pull/7329


nameresolution/kubernetes/kubernetes.go
```diff
@@ -91,3 +100,33 @@ func (k *resolver) ResolveID(ctx context.Context, req nameresolution.ResolveRequ
...
+func (k *resolver) ResolveIDMulti(ctx context.Context, req nameresolution.ResolveRequest) (nameresolution.AddressList, error) {
+	// First, get the address from ResolveID, which is usually a DNS name
+	addr, err := k.ResolveID(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+
+	// Extract the port if present
+	var port string
+	idx := strings.LastIndexByte(addr, ':')
```
daixiang0, line 114: Please add comment for those cases.

berndverst (approved):

Issue #3190, opened 2023-10-24
aws components - validate credentials and connection on init or ping


famarting:
## Describe the feature

Existing component implementations such as postgres, mysql, redis,..., use init function or at least the optional ping function in order to validate as much as possible the provided configuration for the component and to guarantee that it's properly configured and it would work as expected during runtime.

To get a consistent experience, and to be able to properly validate components and avoid runtime errors because of misconfigurations, It would be desirable that as much components as possible properly validate the provided configuration and establish a connection, or at least somehow probe the connection, with the underlying backing infrastructure.

The ask for this issue is to try to implement that behavior for the aws components.

i.e I've had issues with the aws s3 binding component. I had an invalid configuration, with wrong access key and secret key, however the component initialization was successfull but during runtime I got this error using the binding API
```
{
    "errorCode": "ERR_INVOKE_OUTPUT_BINDING",
    "message": "error invoking output binding bucket-wrong-users: s3 binding error: uploading failed: InvalidAccessKeyId: The AWS Access Key Id you provided does not exist in our records.\n\tstatus code: 403, request id:xxxxx, host id: xxxxxxx"
  },
```

I assume there should be some aws API or function in the aws sdk which can be used to probe aws and test the provided configuration, credentials,.... If no adhoc validating API exists, maybe the component can perform a GET request to the s3 bucket or other metadata API... however I'm afraid that may become expensive for users.

## Release Note

RELEASE NOTE: **ADD** connection validation for aws components

RELEASE NOTE:

elena-kolevska:
@famarting we can easily check if the credentials are valid using [sts.GetCallerIdentity](https://docs.aws.amazon.com/sdk-for-go/api/service/sts/#STS.GetCallerIdentity). This will be shared across all aws components. 
In addition, we could do an extra check specific to the component, for example "Is the DynamoDb table present"? I couldn't find any specific api for that check, but we could do a SCAN with a limit of 1. Should I add that?

elena-kolevska:
So that's actually part of why I suggested the sts.GetCallerIdentity call, because it doesn't require permissions, as per the docs:

> No permissions are required to perform this operation. If an administrator attaches a policy to your identity that explicitly denies access to the sts:GetCallerIdentity action, you can still perform this operation. Permissions are not required because the same information is returned when access is denied.

famarting:
oh nice, good catch, ty, so then the perfect solution would be calling first to sts.GetCallerIdentity and then the specialized validation

[...]

PR #3285, opened 2023-12-22
state.dynamodb: validate AWS connection

elena-kolevska:
# Description

I added a Get check to a dummy key on component Init for aws Dynamodb. This will prevent Dapr from initialising if we have wrong component info (credentials/table name).

## Issue reference
https://github.com/dapr/components-contrib/issues/3190

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/state/aws/dynamodb/dynamodb.go
+++ b/state/aws/dynamodb/dynamodb.go
@@ -19,12 +19,14 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
-	"github.com/google/uuid"
 	"reflect"
 	"strconv"
 	"time"
 
-	"github.com/aws/aws-sdk-go/aws"
+	"github.com/google/uuid"
+
+	"github.com/dapr/kit/ptr"
+
 	"github.com/aws/aws-sdk-go/service/dynamodb"
 	"github.com/aws/aws-sdk-go/service/dynamodb/dynamodbattribute"
 	"github.com/aws/aws-sdk-go/service/dynamodb/dynamodbiface"
@@ -102,12 +104,23 @@ func (d *StateStore) Init(ctx context.Context, metadata state.Metadata) error {
 // validateConnection runs a dummy Get operation to validate the connection credentials,
 // as well as validating that the table exists, and we have access to it
 func (d *StateStore) validateTableAccess(ctx context.Context) error {
+	var tableName string
+	if random, err := uuid.NewRandom(); err == nil {
+		tableName = random.String()
+	} else {
+		// We would get to this block if the entropy pool is empty.
+		// We don't want to fail initialising Dapr because of it though,
+		// since it's a dummy table that is only needed to check access, anyway
+		// So we'll just use a hardcoded table name
+		tableName = "dapr-test-table"
+	}
+
 	input := &dynamodb.GetItemInput{
 		ConsistentRead: ptr.Of(false),
 		TableName:      ptr.Of(d.table),
 		Key: map[string]*dynamodb.AttributeValue{
 			d.partitionKey: {
-				S: aws.String(uuid.NewString()),
+				S: ptr.Of(tableName),
 			},
 		},
 	}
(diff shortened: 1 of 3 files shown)


JoshVanL (approved):

state/aws/dynamodb/dynamodb.go
```diff
@@ -74,25 +75,48 @@ func NewDynamoDBStateStore(_ logger.Logger) state.Store {
...
 
+// validateConnection runs a dummy Get operation to validate the connection credentials,
+// as well as validating that the table exists, and we have access to it
+func (d *StateStore) validateTableAccess(ctx context.Context) error {
+	input := &dynamodb.GetItemInput{
+		ConsistentRead: aws.Bool(false),
+		TableName:      aws.String(d.table),
+		Key: map[string]*dynamodb.AttributeValue{
+			d.partitionKey: {
+				S: aws.String(uuid.NewString()),
```
ItalyPaleAle, line 111: `uuid.NewString` can panic (if the kernel doesn't have entropy). Please use `uuid.NewRandom` and then call `.String()` on the result

state/aws/dynamodb/dynamodb.go
```diff
@@ -74,25 +75,48 @@ func NewDynamoDBStateStore(_ logger.Logger) state.Store {
...
+	}
+
 	return nil
 }
 
+// validateConnection runs a dummy Get operation to validate the connection credentials,
+// as well as validating that the table exists, and we have access to it
+func (d *StateStore) validateTableAccess(ctx context.Context) error {
+	input := &dynamodb.GetItemInput{
+		ConsistentRead: aws.Bool(false),
```
ItalyPaleAle, line 107: We use `ptr.Of`
```suggestion
		ConsistentRead: ptr.Of(false),
```

state/aws/dynamodb/dynamodb.go
```diff
@@ -74,25 +75,48 @@ func NewDynamoDBStateStore(_ logger.Logger) state.Store {
...
+		TableName:      aws.String(d.table),
```
ItalyPaleAle, line 108:
```suggestion
		TableName:      ptr.Of(d.table),
```

state/aws/dynamodb/dynamodb.go
ItalyPaleAle: Also, ptr.Of

state/aws/dynamodb/dynamodb_test.go
```diff
@@ -124,6 +131,23 @@ func TestInit(t *testing.T) {
...
+
+		s.client = &mockedDynamoDB{
+			GetItemWithContextFn: func(ctx context.Context, input *dynamodb.GetItemInput, op ...request.Option) (*dynamodb.GetItemOutput, error) {
+				return nil, errors.New("Requested resource not found")
+			},
+		}
+
+		err := s.Init(context.Background(), m)
+		require.Error(t, err)
+		require.Equal(t, "error validating DynamoDB table 'does-not-exist' access: Requested resource not found", err.Error())
```
ItalyPaleAle, line 149: Please use `require.EqualError`: https://pkg.go.dev/github.com/stretchr/testify/require#EqualError

state/aws/dynamodb/dynamodb.go
```diff
@@ -74,25 +75,48 @@ func NewDynamoDBStateStore(_ logger.Logger) state.Store {
...
-	client, err := d.getClient(meta)
-	if err != nil {
-		return err
+	// We have this check because we need to set the client to  a mock in tests
+	if d.client == nil {
+		client, err := d.getClient(meta)
+		if err != nil {
+			return err
+		}
+		d.client = client
```
ItalyPaleAle, line 90: This shadows `err` so the linter overlord doesn't like this
```suggestion
		d.client, err = d.getClient(meta)
		if err != nil {
			return err
		}
```

ItalyPaleAle (changes requested):
Please fix the linter issues

state/aws/dynamodb/dynamodb.go
elena-kolevska: Thanks @ItalyPaleAle . I see we're using `aws.String` and `aws.Bool` 15 other times in this file. Should I update all of them to keep it consistent?

ItalyPaleAle: Up to you, but let's at least use `ptr.Of` for new code :) It's a generic method so works with all types

state/aws/dynamodb/dynamodb.go
elena-kolevska: Thanks @ItalyPaleAle! Check out if you like how I did this. In case of error, I chose to use a fixed string, since I didn't want to have the initialisation of dapr fail because of it; it's just a dummy check anyway.

yaron2 (approved):

ItalyPaleAle (approved):

Issue #3262, opened 2023-12-11
Adding an Optional DeleteWithPrefix Method for State Stores to Allow Deletion of Actor State


RyanLettieri:
## Describe the proposal
This proposal stems from the following feature request inside dapr runimte: https://github.com/dapr/dapr/issues/6992
After a discussion with various members of the project, the following solution was proposed: https://github.com/dapr/dapr/issues/6992#issuecomment-1839557004

By adding in an optional interface called 'DeleteWithPrefix', we will allow users to delete an actor's state. The interface is optional since some databases will not be able to implement it as highlighted by ItalyPaleAle, an active contributor for dapr here: https://github.com/dapr/dapr/pull/7223#issuecomment-1823720095

Looking at sqlite as an example, a second migration will be added in order to generate a virtual "prefix" column. This column will contain all the information from the actor state entry up to but not including the key. For example, if the key entry is as follows: `app-id||actor-type||actor-id||key` the new virtual prefix column will contain `app-id||actor-type||actor-id`

berndverst:
Please include the actual interface method you propose adding. Without that this proposal is not meaningful.

berndverst:
I want to know what the new request object will be called and what it will look like, and what the new method is called.

https://github.com/dapr/components-contrib/blob/main/state/store.go#L38
https://github.com/dapr/components-contrib/blob/main/state/requests.go#L43

For example, ideally you repurpose the existing Delete Request object (I don't recommend introducing an entire new request object), but introduce a new optional property on that request.

That property then only gets used if this new `DeleteWithPrefix` thing was called.

RyanLettieri:
@berndverst updated the proposal to include both suggestions

berndverst:
> @berndverst updated the proposal to include both suggestions

LGTM

And for say Postgres this will eventually be translated to `DELETE FROM my_tbl WHERE ID LIKE '$1%'` where `$1` is the prefix? Is that the goal?

In other words delete every entry where the ID begins with prefix.

ItalyPaleAle:
> And for say Postgres this will eventually be translated to `DELETE FROM my_tbl WHERE ID LIKE '$1%'` where `$1` is the prefix? Is that the goal?

The goal is that, and the query is "like" that (see what I did there? :) )

[...]

PR #3288, opened 2023-12-22
Added conformance tests for DeleteWithPrefix

ItalyPaleAle:
Adds conformance tests for the DeleteWithPrefix method in state stores

See: #3262

CC: @RyanLettieri

### DIFF
--- a/tests/conformance/state/state.go
+++ b/tests/conformance/state/state.go
@@ -337,7 +337,7 @@ func ConformanceTests(t *testing.T, props map[string]string, statestore state.St
 		// so will only assert require.NoError(t, err) finally, i.e. when current implementation
 		// implements ping in existing stable components
 		if err != nil {
-			require.EqualError(t, err, "ping is not implemented by this state store")
+			require.ErrorIs(t, err, state.ErrPingNotImplemented)
 		} else {
 			require.NoError(t, err)
 		}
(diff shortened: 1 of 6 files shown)


tests/conformance/state/state.go
```diff
@@ -1302,6 +1307,109 @@ func ConformanceTests(t *testing.T, props map[string]string, statestore state.St
...
+		}
+		validateFn := func() func(t *testing.T) {
+			return func(t *testing.T) {
+				for key, exists := range keys {
+					res, err := statestore.Get(context.Background(), &state.GetRequest{Key: key})
+					require.NoErrorf(t, err, "Error retrieving key '%s'", key)
+					if exists {
+						require.NotEmptyf(t, res.Data, "Expected key '%s' to be not empty", key)
+					} else {
+						require.Emptyf(t, res.Data, "Expected key '%s' to be empty", key)
```
RyanLettieri, line 1327: Can we also log what was in the res.Data? i.e. Expected key '%s' to be empty, but '%s' data was found

ItalyPaleAle: Good idea. `require.Empty` will already print it out, but because `res.Data` is a `[]byte`, it won't be human-readable.

```suggestion
						require.Emptyf(t, res.Data, "Expected key '%s' to be empty, but contained data: %s", key, string(res.Data))
```

RyanLettieri:
Assuming the failing tests are unrelated, LGTM!

DeepanshuA (approved):

Issue #3144, opened 2023-09-20
Schema registry support for kafka.pubsub component


passuied:
## Describe the feature
New to DAPR and really eager to leverage it at my company. I was surprised that there is no Schema registry support or Avro/protobuf serialization in the kafka.pubsub component.
Is this on the roadmap?
The way it could work is enabling configuring schema registry URL/Authentication.
You could add metadata to the subscription to add specific overrides (e.g. schema name)

## Release Note
RELEASE NOTE: **ADD** New feature in Dapr.

passuied:
Seems there was a mention in https://github.com/dapr/components-contrib/issues/8#issuecomment-1191225512 but the issue was never created...

berndverst:
It's not supported or on the roadmap, but we will accept contributions for this!

passuied:
So updating this issue with findings. 
This work is not trivial as the current [sarama](https://github.com/IBM/sarama/issues/1914) Kafka Go implementation doesn't support Schema registry...
Based on convo with @berndverst, we cannot switch to the Confluent Kafka client as it's leveraging a C library `librdkafka` and C bindings are not allowed in Dapr.
So the solution might be to re-implement the `kafka.pubsub` component as a pluggable component, which doesn't have the same limitations...
Doesn't seem trivial either but maybe easier?

passuied:
Hi, so getting ready to help on this issue. I'm looking into adding Schema Registry to the `kafka-pubsub` component by using the https://github.com/riferrei/srclient library. I'm not experienced in Go but it seems to be pure Go and therefore compatible.
Worst case scenario I could just re-implement the SchemaRegistryClient and leverage the underlying `github.com/linkedin/goavro/v2` library for serialization/deserialization.
I would focus on avro first, since this is our use case.

In terms of interface, I would: 
- Add the following settings to the component:
  - `schemaRegistryUrl`: Schema Registry URL
  - `schemaRegistryApiKey`: Schema Registry credentials API Key
  - `schemaRegitryApiSecret`: Schema Registry credentials API Secret
  - `autoRegisterSchemas`: Whether the schema will get auto registered if missing upon publishing (default=True)
- Add the following metadata to the subscription metadata
  - `valueSchemaType`: Schema Type for value (`None`, `AVRO` only for now)
  
  
So a subscription would look something like this:
```
{'pubsubname': 'pubsub',
                      'topic': 'deathStarStatus',
                      'route': 'dsstatus',
                      'metadata': {
[...]

PR #3292, opened 2023-12-30
Avro Schema registry kafka pubsub implementation

passuied:
# Description

Add Avro Schema Registry  to Kafka PubSub.
- Leverages [srclient](https://github.com/riferrei/srclient) and [go-avro](https://github.com/linkedin/goavro) under the hood
- Please note that for the serialization/deserialization to fully compatible, dates/datetimes in the JSON data need to be set as their respective Unix int and long representations, instead of the more standard Iso8601 one. This is a limitation of the goavro library.

## Issue reference

#3144

# Checklist

 - [x] Code compiles correctly
 - [x] Created/updated tests
 - [x] Extended the documentation - Created issue


passuied:
Please let me know if it's preferable to add publisher in same PR or separate.

yaron2:
Great contribution, I really like how this is done and wonder if this approach can be used in the future to enable a generic schema validation feature for all Dapr pub/subs, implemented in the Dapr runtime itself.

For this PR: It'd be best to add a certification test for this functionality (or add this to an existing test).

passuied:
> For this PR: It'd be best to add a certification test for this functionality (or add this to an existing test).
 Thanks @yaron2, Is there doc about these tests and especially how to run them locally?

ItalyPaleAle:
The certification tests for Kafka PubSub are here: https://github.com/dapr/components-contrib/tree/main/tests/certification/pubsub/kafka

We don't have good docs (we should....), but for certifications test you just "cd" into the test's folder, and then run `go test -v -tags cerrttests`

yaron2:
@passuied please resolve the conflict

pubsub/kafka/metadata.yaml
```diff
@@ -259,4 +259,31 @@ metadata:
...
+      description: |
+        Enables caching for schemas.
+      example: '"true"'
+      default: '"true"'
+    - name: LatestSchemaCacheTTL
+      type: duration
+      description: |
+        The TTL for schema caching when publishing a message with latest schema available.
+      example: '"5min"'
+      default: '"5min"'
```
ItalyPaleAle, line 289: Incorrect format for Go durations:
```suggestion
      example: '"5m"'
      default: '"5m"'
```

bindings/kafka/metadata.yaml
```diff
@@ -268,4 +268,31 @@ metadata:
...
+    description: |
+      Enables caching for schemas.
+    example: '"true"'
+    default: '"true"'
+  - name: LatestSchemaCacheTTL
+    type: duration
+    description: |
+      The TTL for schema caching when publishing a message with latest schema available.
+    example: '"5min"'
+    default: '"5min"'
```
ItalyPaleAle, line 298: Incorrect format for Go durations:
```suggestion
      example: '"5m"'
      default: '"5m"'
```

common/component/kafka/kafka.go
```diff
@@ -55,6 +67,39 @@ type Kafka struct {
...
+)
+
+type SchemaCacheEntry struct {
+	schema         *srclient.Schema
+	codec          *goavro.Codec
+	expirationTime time.Time
+}
+
+func GetValueSchemaType(metadata map[string]string) (SchemaType, error) {
+	schemaTypeStr, ok := metadata[valueSchemaType]
```
ItalyPaleAle, line 84: Please use this method from dapr/kit which also supports case-insensitive matching: https://github.com/dapr/kit/blob/77f7f031c92afb648d24d6dea18cf80e07fe21d5/metadata/utils.go#L26

common/component/kafka/kafka.go
```diff
@@ -166,6 +223,128 @@ func (k *Kafka) Close() (err error) {
 	return err
 }
 
+func getSchemaSubject(topic string) string {
+	// For now assumes that subject is named after topic (e.g. `my-topic-value`)
+	return fmt.Sprintf("%s-value", topic)
```
ItalyPaleAle, line 228:
```suggestion
	return topic + "-value"
```

common/component/kafka/kafka.go
```diff
@@ -166,6 +223,128 @@ func (k *Kafka) Close() (err error) {
...
+}
+
+func (k *Kafka) DeserializeValue(message *sarama.ConsumerMessage, config SubscriptionHandlerConfig) ([]byte, error) {
+	if config.ValueSchemaType == Avro {
```
ItalyPaleAle, line 232: Nit: please use a switch statement here, it just reads better :) 

Alternatively, change this to be short-circuited, such as :

```go
if config.ValueSchemaType != Avro {
    return message.Value, nil
}
```

common/component/kafka/kafka.go
```diff
@@ -166,6 +223,128 @@ func (k *Kafka) Close() (err error) {
...
+
+func (k *Kafka) getLatestSchema(topic string) (*srclient.Schema, *goavro.Codec, error) {
+	srClient, err := k.getSchemaRegistyClient()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	subject := getSchemaSubject(topic)
+	if k.schemaCachingEnabled {
+		cacheEntry, ok := k.latestSchemaCache[subject]
```
ItalyPaleAle, line 274: This needs a lock around it or it's not safe for concurrent access

common/component/kafka/kafka.go
```diff
@@ -166,6 +223,128 @@ func (k *Kafka) Close() (err error) {
...
+		}
+		schema, errSchema := srClient.GetLatestSchema(subject)
+		if errSchema != nil {
+			return nil, nil, errSchema
+		}
+		// New JSON standard serialization/Deserialization is not integrated in srclient yet.
+		// Since standard json is passed from dapr, it is needed.
+		codec, errCodec := goavro.NewCodecForStandardJSONFull(schema.Schema())
+		if errCodec != nil {
+			return nil, nil, err
```
ItalyPaleAle, line 287:
```suggestion
			return nil, nil, errCodec
```

common/component/kafka/kafka.go
```diff
@@ -166,6 +223,128 @@ func (k *Kafka) Close() (err error) {
...
+	srClient, err := k.getSchemaRegistyClient()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	subject := getSchemaSubject(topic)
+	if k.schemaCachingEnabled {
+		cacheEntry, ok := k.latestSchemaCache[subject]
+		// Cache present and not expired
+		if ok && cacheEntry.expirationTime.Compare(time.Now()) < 0 {
```
ItalyPaleAle, line 276: Nit: please use [`time.Before`](https://pkg.go.dev/time#Time.Before) (or `time.After`) for readability

common/component/kafka/kafka.go
```diff
@@ -166,6 +223,128 @@ func (k *Kafka) Close() (err error) {
...
+	return k.srClient, nil
+}
+
+func (k *Kafka) SerializeValue(topic string, data []byte, metadata map[string]string) ([]byte, error) {
+	valueSchemaType, err := GetValueSchemaType(metadata)
+	if err != nil {
+		return nil, err
+	}
+
+	if valueSchemaType == Avro {
```
ItalyPaleAle, line 320: Nit: please use a switch statement for readability

common/component/kafka/kafka.go
```diff
@@ -166,6 +223,128 @@ func (k *Kafka) Close() (err error) {
...
+		}
+
+		valueBytes, err := codec.BinaryFromNative(nil, native)
+		if err != nil {
+			return nil, err
+		}
+		schemaIDBytes := make([]byte, 4)
+		binary.BigEndian.PutUint32(schemaIDBytes, uint32(schema.ID()))
+
+		var recordValue []byte
```
ItalyPaleAle, line 338: Nit: optimize memory allocation (hopefully I got the right count)

```suggestion
		recordValue := make([]byte, 0, len(schemaIDBytes)+len(valueBytes)+1)
```

common/component/kafka/metadata.go
```diff
@@ -87,6 +88,12 @@ type KafkaMetadata struct {
 	channelBufferSize    int    `mapstructure:"-"`
 	consumerFetchMin     int32  `mapstructure:"-"`
 	consumerFetchDefault int32  `mapstructure:"-"`
+	// schema registry
+	SchemaRegistryURL       string        `mapstructure:"schemaRegistryURL"`
+	SchemaRegistryAPIKey    string        `mapstructure:"schemaRegistryAPIKey"`
+	SchemaRegistryAPISecret string        `mapstructure:"schemaRegistryAPISecret"`
+	SchemaCachingEnabled    bool          `mapstructure:"schemaCachingEnabled"`
+	LatestSchemaCacheTTL    time.Duration `mapstructure:"LatestSchemaCacheTTL"`
```
ItalyPaleAle, line 96: I think I'd prefer if this started with "schema" for consistency with the other schema-related properties

```suggestion
	LatestSchemaCacheTTL    time.Duration `mapstructure:"latestSchemaCacheTTL"`
```

ItalyPaleAle (changes requested):
Thanks for this PR :) left some comments

common/component/kafka/kafka.go
passuied: Thanks! Great catch!

ItalyPaleAle (approved):

ItalyPaleAle:
LGTM! @yaron2 good to merge if you are good too

yaron2:
Thanks for this @passuied

passuied:
> Thanks for this @passuied

Thank you for a very smooth PR process ;) That was fun!

Issue #3263, opened 2023-11-24
Using kafka component to subscribe to multiple topics on dapr of k8s and reporting an error. Sidecar can't get up


Huawu777:
/area runtime

**What version of Dapr?**
1.12.2

**Expected Behavior**
The status of sidecar is running

**Actual Behavior**
![d5ff999418c949e28e7e719caed4f5df](https://github.com/dapr/dapr/assets/14233576/0372dcd1-3e9e-4398-b5d7-d226d9b218d2)

_my component yaml ：_
```
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: pubsub
spec:
  type: pubsub.kafka
  version: v1
  metadata:
  - name: brokers
    value: xxx.xxx.xxx.xxx:31092
  - name: authType 
    value: "none"
```

By the way, it works perfectly when subscribing to only one topic.

olitomlinson:
took **8 minutes** for the sidecar to come up in our kubernetes environment, with just 5 subscribers.

PR #3297, opened 2024-01-02
Fix Kafka consumer

yaron2:
Fixes https://github.com/dapr/components-contrib/issues/3263.

The way the Kafka component was handling subscription topics was not inline with how Sarama should be correctly used. Until this change, for every new subscription added, the consumer group was closed and started again which in many cases could (and did as reported by users) lead to a cluster-wide outage, as that action involved a rebalance of Kafka partitions. Depending on the number of topics, messages and consumers, Dapr processes would take sometimes minutes to load and in the worst case causing a bottleneck in the Kafka brokers.

This PR fixes that by cancelling and recreating the context for the `Consume` function every time a subscription is added instead of closing down the entire consumer group. This is the correct way of doing that according to Sarama committers and users.
In addition, the PR further optimizes the logic by using a single goroutine to consume topics instead of N goroutines as the number of topics. This is also aligned with proper Sarama usage.

This fix has been tested in a user's live environment. After applying this fix, the following was observed:

1. Dapr sidecar init time decreased from +12s to less than 300ms
2. Kafka cluster outage stopped with all messages successfully processed

### DIFF
--- a/common/component/kafka/consumer.go
+++ b/common/component/kafka/consumer.go
@@ -29,12 +29,15 @@ import (
 )
 
 type consumer struct {
-	k       *Kafka
-	ready   chan bool
-	running chan struct{}
-	stopped atomic.Bool
-	once    sync.Once
-	mutex   sync.Mutex
+	k             *Kafka
+	ready         chan bool
+	running       chan struct{}
+	stopped       atomic.Bool
+	once          sync.Once
+	mutex         sync.Mutex
+	skipConsume   bool
+	consumeCtx    context.Context
+	consumeCancel context.CancelFunc
 }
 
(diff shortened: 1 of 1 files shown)


ItalyPaleAle (approved):

ItalyPaleAle:
LGTM, please cherry-pick into release-1.12 when merged

yaron2:
> LGTM, please cherry-pick into release-1.12 when merged

Done here: https://github.com/dapr/components-contrib/pull/3297.

The file structure is different so couldn't cherry-pick.

olitomlinson:
Can confirm a big noticeable decrease in time taken to init. My local repro with 5 subscribers, this is now down from 15s to 3s. Which is really positive!

I will try to validate on a real cluster tomorrow.

Thanks so much!!

ParrySMS:
Thanks so much!!   yaron2, you are th true hero

PR #3300, opened 2024-01-02
Improve Multi perf on SQL state stores with 1 op only

ItalyPaleAle:
When Multi is invoked with a single operation, on SQL state stores (Postgres v1/v2, CockroachDB, MySQL, SQLite, MS SQL Server) we can omit starting a transaction, as a single operation is going to be executed atomically regardless.

This should offer some non-insignificant perf benefits thanks to:

- Reducing the number of round-trips to the DB from 3 to 1
- Reducing locks used inside the DB server

### DIFF
--- a/common/component/postgresql/v1/postgresql.go
+++ b/common/component/postgresql/v1/postgresql.go
@@ -451,7 +451,7 @@ func (p *PostgreSQL) Multi(parentCtx context.Context, request *state.Transaction
 	default:
 		_, err := pgtransactions.ExecuteInTransaction[struct{}](parentCtx, p.logger, p.db, p.metadata.Timeout, func(ctx context.Context, tx pgx.Tx) (res struct{}, err error) {
 			for _, op := range request.Operations {
-				err = p.execMultiOperation(parentCtx, op, tx)
+				err = p.execMultiOperation(ctx, op, tx)
 				if err != nil {
 					return res, err
 				}
--- a/common/component/postgresql/v1/postgresql.go
+++ b/common/component/postgresql/v1/postgresql.go
@@ -438,29 +438,40 @@ func (p *PostgreSQL) doDelete(parentCtx context.Context, db pginterfaces.DBQueri
 }
 
 func (p *PostgreSQL) Multi(parentCtx context.Context, request *state.TransactionalStateRequest) error {
-	_, err := pgtransactions.ExecuteInTransaction[struct{}](parentCtx, p.logger, p.db, p.metadata.Timeout, func(ctx context.Context, tx pgx.Tx) (res struct{}, err error) {
-		for _, o := range request.Operations {
-			switch x := o.(type) {
-			case state.SetRequest:
-				err = p.doSet(parentCtx, tx, &x)
+	if request == nil {
+		return nil
+	}
+
+	// If there's only 1 operation, skip starting a transaction
+	switch len(request.Operations) {
+	case 0:
+		return nil
+	case 1:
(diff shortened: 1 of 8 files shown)


common/component/postgresql/v1/postgresql.go
```diff
@@ -438,29 +438,40 @@ func (p *PostgreSQL) doDelete(parentCtx context.Context, db pginterfaces.DBQueri
...
+	// If there's only 1 operation, skip starting a transaction
+	switch len(request.Operations) {
+	case 0:
+		return nil
+	case 1:
+		return p.execMultiOperation(parentCtx, request.Operations[0], p.db)
+	default:
+		_, err := pgtransactions.ExecuteInTransaction[struct{}](parentCtx, p.logger, p.db, p.metadata.Timeout, func(ctx context.Context, tx pgx.Tx) (res struct{}, err error) {
+			for _, op := range request.Operations {
+				err = p.execMultiOperation(parentCtx, op, tx)
```
elena-kolevska, line 454: Should we use `ctx` instead of `parentContext` here, just for correctness' sake?

ItalyPaleAle:
```suggestion
				err = p.execMultiOperation(ctx, op, tx)
```

state/postgresql/v2/postgresql.go
```diff
@@ -468,29 +468,40 @@ func (p *PostgreSQL) doDelete(parentCtx context.Context, db pginterfaces.DBQueri
...
+	// If there's only 1 operation, skip starting a transaction
+	switch len(request.Operations) {
+	case 0:
+		return nil
+	case 1:
+		return p.execMultiOperation(parentCtx, request.Operations[0], p.db)
+	default:
+		_, err := pgtransactions.ExecuteInTransaction[struct{}](parentCtx, p.logger, p.db, p.metadata.Timeout, func(ctx context.Context, tx pgx.Tx) (res struct{}, err error) {
+			for _, op := range request.Operations {
+				err = p.execMultiOperation(parentCtx, op, tx)
```
ItalyPaleAle, line 484:
```suggestion
				err = p.execMultiOperation(ctx, op, tx)
```

state/mysql/mysql.go
```diff
@@ -768,40 +768,50 @@ func readRow(row interface{ Scan(dest ...any) error }) (key string, value []byte
...
-			if err != nil {
-				return err
+	// If there's only 1 operation, skip starting a transaction
+	switch len(request.Operations) {
+	case 0:
+		return nil
+	case 1:
+		return m.execMultiOperation(ctx, request.Operations[0], m.db)
+	default:
+		tx, err := m.db.Begin()
```
elena-kolevska, line 785: We should probably use `BeginTx` here like we do in sqlite.

```suggestion
		tx, err := m.db.BeginTx(ctx, nil)
```

elena-kolevska (commented):
LGTM. Love that we care about saving unnecessary roundtrips.

state/mysql/mysql.go
ItalyPaleAle: Possibly a good idea, but I'd like to not make this change in this PR and maybe do a follow-up PR

`Begin` and `BeginTx` have some difference in behavior, specifically (unlike in the pgx driver for Postgres) the context controls the lifecycle of the entire transaction. See: https://pkg.go.dev/database/sql#DB.BeginTx

yaron2 (approved):

PR #3303, opened 2024-01-04
Fix Redis Query for RediSearch 2.8+

JoshVanL:
In RediSearch 2.8 the returned data structure of an `FT.SEARCH` changed, breaking the state Query API.

PR updates the redis Query function to parse both the old and new returned structures. Adds a query certification test.

https://docs.redis.com/latest/stack/release-notes/redisearch/redisearch-2.8-release-notes/

### DIFF
--- a/state/redis/redis_query.go
+++ b/state/redis/redis_query.go
@@ -295,10 +295,49 @@ func (q *Query) execute(ctx context.Context, client rediscomponent.RedisClient) 
 	if err != nil {
 		return nil, "", err
 	}
-	arr, ok := ret.([]interface{})
+
+	res := []state.QueryItem{}
+
+	arr, ok := ret.([]any)
 	if !ok {
-		return nil, "", fmt.Errorf("invalid output")
+		aarr, ok := ret.(map[any]any)
+		if !ok {
+			return nil, "", fmt.Errorf("invalid output")
+		}
+
+		// Post redisearch 2.8
+
+		arr = aarr["results"].([]any)
+		if len(arr) == 0 {
+			return nil, "", errors.New("invalid output")
+		}
+		for i := 0; i < len(arr); i++ {
+			aarr, ok = arr[i].(map[any]any)
+			if !ok {
+				return nil, "", fmt.Errorf("invalid output")
+			}
+			exattr, ok := aarr["extra_attributes"].(map[any]any)
+			if !ok {
+				return nil, "", fmt.Errorf("invalid output")
+			}
+			item := state.QueryItem{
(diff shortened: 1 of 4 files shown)


tests/certification/state/redis/docker-compose.yml
```diff
@@ -1,7 +1,6 @@
 version: "3.3"
 services:
   redis:
-    image: 'redislabs/redisearch:latest'
+    image: 'redislabs/redisearch:edge'
```
JoshVanL, line 4: edge includes the later 2.8 redis search module.

tests/certification/state/redis/docker-compose.yml
```diff
@@ -1,7 +1,6 @@
...
     ports:
       - '6379:6379'
-    command: redis-server
```
JoshVanL, line 7: Adding this command prevented both the search and json modules from being loaded, which are needed for the query API.

tests/certification/state/redis/docker-compose.yml
ItalyPaleAle: Please don't pin to non-stable builds. I trust that the bug was there :)

state/redis/redis_query.go
```diff
@@ -295,10 +295,49 @@ func (q *Query) execute(ctx context.Context, client rediscomponent.RedisClient)
 	if err != nil {
 		return nil, "", err
 	}
-	arr, ok := ret.([]interface{})
+
+	res := []state.QueryItem{}
+
+	arr, ok := ret.([]any)
```
ItalyPaleAle, line 301: Can you extract the pre- and post- in 2 separate functions please, to keep the code more readable

ItalyPaleAle (changes requested):

tests/certification/state/redis/docker-compose.yml
JoshVanL: Makes sense, done!

state/redis/redis_query.go
JoshVanL: Good idea, done :slightly_smiling_face:

ItalyPaleAle:
@JoshVanL code LGTM but looks like the certification test for redis is failing: https://github.com/dapr/components-contrib/actions/runs/7409569254/job/20160102316?pr=3303 can you please take a look?

JoshVanL:
Thanks @ItalyPaleAle, now fixed. Was an issue of `require.Error` vs `require.NoError`

ItalyPaleAle (approved):

PR #3304, opened 2024-01-04
Create shared auth package for SQL Server

ItalyPaleAle:
Similarly to other databases like Postgres and SQLite, moving auth code for MS SQL Server to a shared package.


state/sqlserver/metadata.yaml
ItalyPaleAle, line 1: Reviewers: the names have been slightly changed to be more standardized with other components and other metadata properties. However, the old names remain available as aliases and continue to work

berndverst (approved):

PR #3305, opened 2024-01-05
Kafka: fix context use in AWS IAM auth

famarting:
# Description

The Kafka awsiam authentication type did not actually work since it was making incorrect use of the context that was passed to the Init function

As it can be seen here https://github.com/dapr/dapr/blob/70e3f699793bb138d927079682261999b010e40e/pkg/runtime/processor/processor.go#L381 dapr cancels the context passed to Init once it finishes, making this context not suitable for background tasks such as using the aws sdk to get the kafka credentials at runtime...

## Issue reference

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


common/component/kafka/auth.go
```diff
@@ -111,16 +110,16 @@ func updateAWSIAMAuthInfo(ctx context.Context, config *sarama.Config, metadata *
...
 	secretKey    string
 	sessionToken string
 	region       string
 }
 
+// this function can't use the context passed on Init because that context would be cancelled right after Init
 func (m *mskAccessTokenProvider) Token() (*sarama.AccessToken, error) {
 	if m.accessKey != "" && m.secretKey != "" {
-		token, _, err := signer.GenerateAuthTokenFromCredentialsProvider(m.ctx, m.region, aws2.CredentialsProviderFunc(func(ctx context.Context) (aws2.Credentials, error) {
+		token, _, err := signer.GenerateAuthTokenFromCredentialsProvider(context.Background(), m.region, aws2.CredentialsProviderFunc(func(ctx context.Context) (aws2.Credentials, error) {
```
ItalyPaleAle, line 122: I understand the bug and why this can't use the old context. However, we need a way to stop this operation if the component is closing. Perhaps in `Init()` we can set `m.ctx` to a new context, which is then canceled in the `Close()` method?

ItalyPaleAle (changes requested):

common/component/kafka/auth.go
eunicecompra: Hi folks! Chiming in. I have a [WIP branch](https://github.com/eunicecompra/components-contrib/pull/1/files) in my personal account to add support for IAM role auth. I'm new to Go so, still figuring out the best way to use `context`. I also encountered the context cancelled error, which relates to the bug that is to be fixed by this PR. 

I have initially implemented the IAM auth like this. But I reckon my final implementation will be to follow suit the implementation in this PR. Subscribing..

```go
	} else if m.awsIamRoleArn != "" {
		// Using a different context since the context from init may have been cancelled when token is refreshed
		ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
		defer cancel()

		token, _, err := signer.GenerateAuthTokenFromRole(ctx, m.region, m.awsIamRoleArn, "")
		return &sarama.AccessToken{Token: token}, err
	}
```

famarting: just addressed your feedback, let me know what you think

common/component/kafka/auth.go
```diff
@@ -111,16 +113,21 @@ func updateAWSIAMAuthInfo(ctx context.Context, config *sarama.Config, metadata *
...
+	ctx                  context.Context
+	generateTokenTimeout time.Duration
+	accessKey            string
+	secretKey            string
+	sessionToken         string
+	region               string
 }
 
+// this function can't use the context passed on Init because that context would be cancelled right after Init
 func (m *mskAccessTokenProvider) Token() (*sarama.AccessToken, error) {
```
ItalyPaleAle, line 125:
```suggestion
func (m *mskAccessTokenProvider) Token() (*sarama.AccessToken, error) {
    // this function can't use the context passed on Init because that context would be cancelled right after Init
```

ItalyPaleAle (dismissed):

common/component/kafka/auth.go
ItalyPaleAle: Thank you!

eunicecompra (approved):
changes look good. There's a linting error though.

ItalyPaleAle (approved):

PR #3307, opened 2024-01-06
Postgres v2: Add support for DeleteWithPrefix

ItalyPaleAle:
Related to #3262

Adds DeleteWithPrefix to the PostgreSQL state store v2.

Notes:

1. This cannot work with CockroachDB, because CockroachDB does not offer sufficient support for UDFs which are needed to implement this. We will investigate how to add support for CockroachDB later. For now, the component can be instantiated without support for DeleteWithPrefix, so it can be used with CockroachDB
   - I am confirming if this works with YugabyteDB, although it should _on paper_
2. Because of lack of CockroachDB support, this has not been implemented in v1, due to the fact that the components for v1 are more "intertwined". Will be done in a separate PR.

### DIFF
--- a/.github/infrastructure/docker-compose-cockroachdb.yml
+++ b/.github/infrastructure/docker-compose-cockroachdb.yml
@@ -1,9 +1,9 @@
 version: '2'
 services:
   cockroachdb:
-    image: cockroachdb/cockroach:v21.2.3
+    image: cockroachdb/cockroach:v23.1.13
     hostname: cockroachdb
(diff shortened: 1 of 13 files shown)


ItalyPaleAle:
> Enabling conformance tests for Yugabyte really should be a separate PR

Agree. I'll leave it here for now so I can test that the feature works with Yugabyte too, but I'll move it out before marking the PR as ready for review.

sicoyle:
Hi! Just checking in — is this PR still in progress, or should we go ahead and close it out?

PR #3309, opened 2024-01-09
Add ExecuteInTransaction method for db.SQL

ItalyPaleAle:
A similar method already existed for Postgres, and this adds it for the other databases that use database/sql (MySQL, SQLite, SQL Server).

Also fixes a bug in the SQL state store cleanup, where a transaction was started for no reason.

Doesn't add new features, just a small refactoring


DeepanshuA (approved):
LGTM.

PR #3320, opened 2024-01-19
FIX: Redis to handle event metadata.

artursouza:
# Description

Redis does not handle event metadata at all. We had a P0 issue in runtime that was discovered because of Pulsar pubsub and it could not be reproduced with Redis because of inconsistency of handling event metadata - in this case, Redis simply drops it. This PR brings the Redis component (stable) at the same level of compatibility we should have across beta and stable pubsub components (IMO) regarding event metadata handling. This is not a P0 (IMO), it will just block the E2E test to reproduce the P0 bug we will need to fix.

## Issue reference

Related to the following issues:
 - P0 in runtime that fixes conflict of event metadata for pubsub: https://github.com/dapr/dapr/issues/7413
 - non-P0 issue to discuss consistency across pubsub for event metadata handling: https://github.com/dapr/components-contrib/issues/3321

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

RELEASE_NOTE: **FIXED** Redis pubsub now handles event metadata.

### DIFF
--- a/pubsub/redis/redis.go
+++ b/pubsub/redis/redis.go
@@ -15,6 +15,7 @@ package redis
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"reflect"
@@ -101,7 +102,17 @@ func (r *redisStreams) Publish(ctx context.Context, req *pubsub.PublishRequest) 
 		return errors.New("component is closed")
 	}
 
-	_, err := r.client.XAdd(ctx, req.Topic, r.clientSettings.MaxLenApprox, map[string]interface{}{"data": req.Data})
+	redisPayload := map[string]interface{}{"data": req.Data}
+
+	if req.Metadata != nil {
+		serializedMetadata, err := json.Marshal(req.Metadata)
+		if err != nil {
+			return err
+		}
+		redisPayload["metadata"] = serializedMetadata
+	}
+
+	_, err := r.client.XAdd(ctx, req.Topic, r.clientSettings.MaxLenApprox, redisPayload)
 	if err != nil {
 		return fmt.Errorf("redis streams: error from publish: %s", err)
 	}
@@ -157,7 +168,7 @@ func (r *redisStreams) Subscribe(ctx context.Context, req pubsub.SubscribeReques
 // pick them up for processing.
 func (r *redisStreams) enqueueMessages(ctx context.Context, stream string, handler pubsub.Handler, msgs []rediscomponent.RedisXMessage) {
 	for _, msg := range msgs {
-		rmsg := createRedisMessageWrapper(ctx, stream, handler, msg)
+		rmsg := r.createRedisMessageWrapper(ctx, stream, handler, msg)
 
 		select {
 		// Might block if the queue is full so we need the ctx.Done below.
@@ -172,7 +183,7 @@ func (r *redisStreams) enqueueMessages(ctx context.Context, stream string, handl
 
 // createRedisMessageWrapper encapsulates the Redis message, message identifier, and handler
 // in `redisMessage` for processing.
-func createRedisMessageWrapper(ctx context.Context, stream string, handler pubsub.Handler, msg rediscomponent.RedisXMessage) redisMessageWrapper {
+func (r *redisStreams) createRedisMessageWrapper(ctx context.Context, stream string, handler pubsub.Handler, msg rediscomponent.RedisXMessage) redisMessageWrapper {
 	var data []byte
 	if dataValue, exists := msg.Values["data"]; exists && dataValue != nil {
 		switch v := dataValue.(type) {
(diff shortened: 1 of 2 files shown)


pubsub/redis/redis.go
```diff
@@ -183,11 +194,21 @@ func createRedisMessageWrapper(ctx context.Context, stream string, handler pubsu
 		}
 	}
 
+	var metadata map[string]string
+	if metadataValue, exists := msg.Values["metadata"]; exists && metadataValue != nil {
+		metadataBytes := metadataValue.(string)
+		err := json.Unmarshal([]byte(metadataBytes), &metadata)
+		if err != nil {
+			r.logger.Warnf("Could not extract metadata for Redis message %s: %v", msg.ID, err)
```
artursouza, line 202: This seems to be better than getting the message stuck because metadata could not be handled.

yaron2 (approved):

ItalyPaleAle (changes requested):
Since there's no linked issue, can you please explain what this PR does and why it's needed?

We are past the code freeze phase, so this should probably be merged when the 1.14 merge window opens, unless there's a P0 issue attached to it.

artursouza:
I can create a specific issue if this needs further discussion. I created two issues that this PR is part of. Basically, bringing having consistency across pubsub components regarding handling of pubsub metadata.

This will only block our integration or E2E tests to reproduce the bug I am fixing for 1.13 - which is a P0 as it makes pubsub unusable for some components with gRPC subscriber.

artursouza:
> Since there's no linked issue, can you please explain what this PR does and why it's needed?
> 
> We are past the code freeze phase, so this should probably be merged when the 1.14 merge window opens, unless there's a P0 issue attached to it.

See. See the updated description and my comment. This fix can wait until 1.14 but we will not have E2E tests to validate the hotfix for 1.13 - which I can do manually. Of course, we can add yet-another-pubsub to E2E tests, but I don't think it is the right path since this is a basic inconsistency in pubsub that should be fixed (not P0).

ItalyPaleAle:
@artursouza my .02:

- If this is needed primarily to unblock an E2E test in runtime, I'd prefer if it were added to the in-memory pubsub component for 1.13
- If we agree that all components should support metadata (or at least expose a feature flag for those who do support that or not), then we should have a conformance test to validate this, or the implementation could end up being "patchwork-y"

I'd prefer avoiding adding new features to a "tier 1" component (with a large user base like Redis) when we're already in endgame. But I can be open if other maintainers agree this is the best path forward.

artursouza:
> @artursouza my .02:
> 
> 
> 
> - If this is needed primarily to unblock an E2E test in runtime, I'd prefer if it were added to the in-memory pubsub component for 1.13
> 
> - If we agree that all components should support metadata (or at least expose a feature flag for those who do support that or not), then we should have a conformance test to validate this, or the implementation could end up being "patchwork-y"
> 
> 
> 
> I'd prefer avoiding adding new features to a "tier 1" component (with a large user base like Redis) when we're already in endgame. But I can be open if other maintainers agree this is the best path forward.

In-memory pubsub does not work for e2e because the subscriber is not the same app as the publisher. While the test can exist with another pubsub, I found it odd for such a basic feature not be supported by all beta and stable pubsub. If this behavior is not standard, then we should add a new column to the feature set of pubsub components. As you see, I opened an issue to discuss the expected behavior for pubsubs and see what others think.

In the worst case, I can add this feature behind a component metadata flag for those that need it.

pubsub/redis/redis.go
```diff
@@ -172,7 +183,7 @@ func (r *redisStreams) enqueueMessages(ctx context.Context, stream string, handl
 
 // createRedisMessageWrapper encapsulates the Redis message, message identifier, and handler
 // in `redisMessage` for processing.
-func createRedisMessageWrapper(ctx context.Context, stream string, handler pubsub.Handler, msg rediscomponent.RedisXMessage) redisMessageWrapper {
+func (r *redisStreams) createRedisMessageWrapper(ctx context.Context, stream string, handler pubsub.Handler, msg rediscomponent.RedisXMessage) redisMessageWrapper {
```
berndverst, line 186: This change is not necessary. It exists the way it is by design because this helper method does not operate on the struct at all. It is static.

berndverst, line 186:
```suggestion
func createRedisMessageWrapper(ctx context.Context, stream string, handler pubsub.Handler, msg rediscomponent.RedisXMessage) redisMessageWrapper {
```

berndverst: Ah, I see you did it because you wanted access to the logger. Ok that's fine then

berndverst (changes requested):
I will approve this, but I really think you should add conformance tests.

I will not block on conformance tests, but that is the right thing to do.

My only blocking comment is to please prefix the error message with the component name as I have suggested. This helps with troubleshooting or grepping for similar logs (to see prevalence of an issue etc etc).

Beyond that I am fine merging this.

ItalyPaleAle (approved):
Given that @berndverst  and @yaron2 both agree to include this, I'm fine too. Thanks for your patience :)

berndverst:
I took the liberty to apply my suggested log prefix. LGTM now.

berndverst (approved):

Issue #3254, opened 2023-12-05
Support Amazon Aurora within Dapr


sicoyle:
## Describe the feature
There currently is no support for Amazon Aurora within Dapr, ie. there is no `state.aws.aurora`. I wanted to create an issue to discuss the 2 options to introduce support for Amazon Aurora as a component.
1. Update the PostgreSQL component with a light wrapper adding the required AWS auth fields
2. Add a specific component for `state.aws.aurora`

What would be the preferred approach here?

## Release Note

RELEASE NOTE: Add Amazon Aurora support within Dapr

sicoyle:
Documenting the discussion on discord over to here:

Amazon Aurora supports multiple database technologies. 
Adding a generic Aurora component would introduce maintenance overhead that is unideal, and so there is a way to support it for the existing PostgreSQL component. Alternatively, this can become a new component that under the hood uses the shared postgres implementation (see CockroachDB for reference).

Reference code thanks to @berndverst 
- https://github.com/dapr/components-contrib/tree/main/common/component/postgresql

Additionally, Azure Postgres is supported using a new `enableAzureAD` metadata field. See below for reference for a corresponding "Amazon Postgres" implementation:
- https://github.com/dapr/components-contrib/pull/2970
- https://github.com/dapr/components-contrib/pull/2971


The above summary is compiled from the Discord discussion here:
https://discord.com/channels/778680217417809931/781589820128493598/1181670307065499678

yaron2:
Its best to use the existing component and add the required AWS configuration to it. If it can work transparently than great, if not then it can be set using metadata.

berndverst:
> Its best to use the existing component and add the required AWS configuration to it. If it can work transparently than great, if not then it can be set using metadata.

Yup. That was exactly my recommendation.

See entire thread starting here: https://discord.com/channels/778680217417809931/781589820128493598/1181677547818328134

PR #3324, opened 2024-01-29
feat(aws iam): support aws iam auth for postgresql components

sicoyle:
# Description

Allow for AWS IAM auth with postgres v2. These changes rotate the authentication token automatically before the 15 minute expiration on the authentication token.

Adding support for v1 and v2 postgres statestore and config and bindings postgres types.

Please note that I removed the master connection string setup I had initially. There is no reason to use aws iam enabled on postgres and a master connection string, as the dapr user could just use the regular postgres connection string auth for that. I now have it setup to where you just use an iam enabled username and the pwd is auto generated based on the aws config file or the secretkey/accesskey for the iam enabled user.

## Issue reference

https://github.com/dapr/components-contrib/issues/3254

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


common/authentication/postgresql/metadata.go
```diff
@@ -43,36 +43,38 @@ func (m *PostgresAuthMetadata) Reset() {
...
 		m.azureEnv, err = azure.NewEnvironmentSettings(meta)
 		if err != nil {
 			return err
 		}
-	} else {
-		// Make sure this is false
+	case awsIAMEnabled && m.UseAWSIAM:
+		// Do y'all want the logic moved to moreso here instead of where I have it?
+	case m.ConnectionString == "":
+		return errors.New("missing connection string")
```
ItalyPaleAle, line 64: At least for Azure, and I suspect AWS too, the connection string must not be empty even when using Azure AD

common/authentication/postgresql/metadata.go
```diff
@@ -43,36 +43,38 @@ func (m *PostgresAuthMetadata) Reset() {
...
+	switch {
+	case azureADEnabled && m.UseAzureAD:
 		m.azureEnv, err = azure.NewEnvironmentSettings(meta)
 		if err != nil {
 			return err
 		}
-	} else {
-		// Make sure this is false
+	case awsIAMEnabled && m.UseAWSIAM:
+		// Do y'all want the logic moved to moreso here instead of where I have it?
```
ItalyPaleAle, line 62: meant to be deleted? :)

configuration/postgres/metadata.go
```diff
@@ -55,8 +55,8 @@ func (m *metadata) InitWithMetadata(meta map[string]string) error {
 		return fmt.Errorf("invalid table name '%s'. non-alphanumerics or upper cased table names are not supported", m.ConfigTable)
 	}
 
-	// Azure AD auth is supported for this component
-	err = m.PostgresAuthMetadata.InitWithMetadata(meta, true)
+	// Azure AD & AWS IAM auth is supported for this component
+	err = m.PostgresAuthMetadata.InitWithMetadata(meta, true, true)
```
ItalyPaleAle, line 59: Please update the `metadata.yaml` for this component

state/postgresql/v2/metadata.go
```diff
@@ -41,9 +41,20 @@ type pgMetadata struct {
 	MetadataTableName string         `mapstructure:"metadataTableName"` // Could be in the format "schema.table" or just "table"
 	Timeout           time.Duration  `mapstructure:"timeout" mapstructurealiases:"timeoutInSeconds"`
 	CleanupInterval   *time.Duration `mapstructure:"cleanupInterval" mapstructurealiases:"cleanupIntervalInSeconds"`
+
+	// AWS IAM related
+	// Ignored by metadata parser because included in built-in authentication profile
+	// access key to use for accessing sqs/sns.
+	AWSAccessKey string `json:"awsAccessKey" mapstructure:"awsAccessKey" mdignore:"true"`
```
ItalyPaleAle, line 48: Please move these options to a common struct. For example for Azure we have a common "azure auth" package. We should have one for AWS too.

bindings/postgres/metadata.go
```diff
@@ -32,8 +32,8 @@ func (m *psqlMetadata) InitWithMetadata(meta map[string]string) error {
 	}
 
 	// Validate and sanitize input
-	// Azure AD auth is supported for this component
-	err = m.PostgresAuthMetadata.InitWithMetadata(meta, true)
+	// Azure AD & AWS IAM auth is supported for this component
+	err = m.PostgresAuthMetadata.InitWithMetadata(meta, true, true)
```
ItalyPaleAle, line 36: Should probably be `false` for AWS IAM if you don't enable it for this component
```suggestion
	err = m.PostgresAuthMetadata.InitWithMetadata(meta, true, false)
```

common/component/postgresql/v1/postgresql.go
```diff
@@ -52,6 +52,7 @@ type PostgreSQL struct {
 	setQueryFn    func(*state.SetRequest, SetQueryOptions) string
 	etagColumn    string
 	enableAzureAD bool
+	enableAWSIAM  bool
```
ItalyPaleAle, line 55: Unused?

ItalyPaleAle (changes requested):
Thanks for this PR.

Please move all the code for authenticating with AWS IAM to a shared package. If it's specific for Postgres, it should be in one of the common packages we have for Postgres.

This way, you can also enable it for all PG components (not just the state store v1, but also the binding and configuration store)

common/component/postgresql/awsIAM/aws.go
```diff
@@ -0,0 +1,206 @@
...
+	// Default to load default config through aws credentials file (~/.aws/credentials)
+	awsCfg, err := aws_config.LoadDefaultConfig(ctx)
+	if err != nil {
+		// otherwise use metadata fields
+		// Check if access key and secret access key are set
+		accessKey := accessKey
+		secretKey := secretKey
+
+		// Validate if access key and secret access key are provided
+		if accessKey == "" || secretKey == "" {
```
famarting, line 63: it should be possible to not provide accessKey and secretKey so the aws sdk picks the credentials from the environment and users can use "workload identity" from AWS

common/component/postgresql/awsIAM/aws.go
```diff
@@ -0,0 +1,206 @@
...
+		grantRoleCancel()
+		if err != nil {
+			return fmt.Errorf("failed to grant PostgreSQL user role: %v", err)
+		}
+	}
+	return nil
+}
+
+func InitAWSDatabase(ctx context.Context, config *pgxpool.Config, db pginterfaces.PGXPoolConn, timeout time.Duration, connString string, awsAccessKey, awsSecretKey string) error {
+	err := CreateDatabaseIfNeeded(ctx, timeout, connString, db)
```
famarting, line 177: can this create steps be opt in or opt out? so the component can be configured with a connectionString without password and then the logic here would just automatically provide the aws token for the user in the connection string

sicoyle: ohh, yes I can account for that. thank you for the feedback!

sicoyle: I've tried to get this to work without using the master username/password. However, I haven't been able to get it to work this way. I started by creating my IAM user, creating the same user name in my database and granting the rds_iam permissions within the specific database from the connection string. However, I keep seeing the `FATAL: PAM authentication failed for user <username>"` even after verifying that the user has the rds_iam role granted to it within the proper database. Not sure what to try next here. I even tried following this tutorial too to see if I missed anything and I don't think that I did: https://www.linkedin.com/pulse/authenticating-postgres-rds-logins-using-aws-iam-cross-devadoss/

sicoyle:
potentially flakey test and unrelated?
```
--- FAIL: Test_EndToEnd (0.40s)
    --- FAIL: Test_EndToEnd/http+gzip (0.22s)
        --- FAIL: Test_EndToEnd/http+gzip/consoleLog_stdout_and_stderr (0.09s)
            e2e_test.go:91: 
                	Error Trace:	/home/runner/work/components-contrib/components-contrib/middleware/http/wasm/internal/e2e_test.go:91
                	            				/home/runner/work/components-contrib/components-contrib/middleware/http/wasm/internal/e2e_test.go:221
                	Error:      	"time=\"2024-02-07T21:16:45.410473385Z\" level=info msg=\"main ConsoleLog\" instance=fv-az736-80 scope=Test_EndToEnd type=log ver=unknown\ntime=\"2024-02-07T21:16:45.411097546Z\" level=info msg=\"main ConsoleLog\" instance=fv-az736-80 scope=Test_EndToEnd type=log ver=unknown\ntime=\"2024-02-07T21:16:45.411[128](https://github.com/dapr/components-contrib/actions/runs/7821189411/job/21337610051?pr=3324#step:9:129)433Z\" level=info msg=\"request[0] ConsoleLog\" instance=fv-az736-80 scope=Test_EndToEnd type=log ver=unknown\ntime=\"2024-02-07T21:16:45.411139224Z\" level=debug msg=\"wasm stdout: main Stdout\\nmain Stdout\\nrequest[0] Stdout\\n\" instance=fv-az736-80 scope=Test_EndToEnd type=log ver=unknown\ntime=\"2024-02-07T21:16:45.41115311Z\" level=debug msg=\"wasm stderr: main Stderr\\nmain Stderr\\nrequest[0] Stderr\\n\" instance=fv-az736-80 scope=Test_EndToEnd type=log ver=unknown\n" does not contain "level=debug msg=\"wasm stdout: main Stdout\\nrequest[0] Stdout\\n\""
                	Test:       	Test_EndToEnd/http+gzip/consoleLog_stdout_and_stderr
```

also my PR is seeing the below err for a pubsub kafka conformance test err:
```
go: github.com/DataDog/zstd@v1.5.2: read "https:/proxy.golang.org/@v/v1.5.2.zip": stream error: stream ID 4899; INTERNAL_ERROR; received from peer

```

sicoyle:
readyyyyy 😁

common/authentication/postgresql/metadata.go
```diff
@@ -43,27 +44,34 @@ func (m *PostgresAuthMetadata) Reset() {
...
-	if azureADEnabled && m.UseAzureAD {
+	switch {
+	case opts.AzureADEnabled && m.UseAzureAD:
 		m.azureEnv, err = azure.NewEnvironmentSettings(meta)
 		if err != nil {
 			return err
 		}
-	} else {
-		// Make sure this is false
+	case opts.AWSIAMEnabled && m.UseAWSIAM:
```
famarting, line 67: aws region should be a required metadata if useAWSIAM is set, no?

sicoyle: yes, this is marked as required in the metadata yaml file

sicoyle:
@ItalyPaleAle any chance y'all can review again please?

sicoyle:
From my PR build. Again, totally unrelated:

```
=== FAIL: bindings/rabbitmq TestRabbitMQExclusive/rabbitmq_certification (11.91s)
    flow.go:220: Running step: rabbitmqcertification
    dockercompose.go:57: Creating network "rabbitmqcertification_rabbitmq_go_net" with driver "bridge"
        Creating rabbitmq ... 
        Creating rabbitmq ... done
        
    flow.go:228: Completed step: rabbitmqcertification
    flow.go:220: Running step: wait for rabbitmq readiness
    retry.go:33: Failure: Exception (501) Reason: "read tcp [::1]:60818->[::1]:5672: read: connection reset by peer"; Retrying in 1s
    retry.go:35: Success!
    flow.go:228: Completed step: wait for rabbitmq readiness
    flow.go:220: Running step: standardApp
    flow.go:228: Completed step: standardApp
    flow.go:220: Running step: standardSidecar
time="2024-02-08T14:16:17.800355388Z" level=warning msg="The default value for 'spec.metric.http.increasedCardinality' will change to 'false' in Dapr 1.14" instance=fv-az570-300 scope=dapr.runtime type=log ver=unknown
time="2024-02-08T14:16:17.800466546Z" level=error msg="[INIT_FAILURE]: initialization error occurred for metrics: runtime/actor/timers: cannot register view \"runtime/actor/timers\"; a different view with the same name is already registered\nruntime/actor/reminders: cannot register view \"runtime/actor/reminders\"; a different view with the same name is already registered" instance=fv-az570-300 scope=dapr.runtime type=log ver=unknown
time="2024-02-08T14:16:17.800535134Z" level=info msg="Watching directories: [./components/exclusive]" instance=fv-az570-300 scope=dapr.runtime.hotreload.loader.disk type=log ver=unknown
    flow.go:228: Completed step: standardSidecar
    flow.go:230: Fatal error in step standardSidecar: failed to create watcher: failed to add target ./components/exclusive: no such file or directory
```

sicoyle:
also see unrelated pubsub mqtt err:
```
panic: test timed out after 30m0s
running tests:
	TestMQTT (30m0s)
	TestMQTT/mqtt_certification (30m0s)

goroutine 40599 [running]:
testing.(*M).startAlarm.func1()
	/opt/hostedtoolcache/go/1.21.6/x64/src/testing/testing.go:2259 +0x3b9
created by time.goFunc
	/opt/hostedtoolcache/go/1.21.6/x64/src/time/sleep.go:176 +0x2d

goroutine 1 [chan receive, 30 minutes]:
testing.(*T).Run(0xc0004adba0, {0x2211fab?, 0x588b3c?}, 0x23353b8)
	/opt/hostedtoolcache/go/1.21.6/x64/src/testing/testing.go:1649 +0x3c8
testing.runTests.func1(0x380bfe0?)
	/opt/hostedtoolcache/go/1.21.6/x64/src/testing/testing.go:2054 +0x3e
testing.tRunner(0xc0004adba0, 0xc000529c38)
	/opt/hostedtoolcache/go/1.21.6/x64/src/testing/testing.go:1595 +0xff
testing.runTests(0xc0005a0640?, {0x37d3690, 0x1, 0x1}, {0x418dbf?, 0xc000529cf8?, 0x3809f80?})
	/opt/hostedtoolcache/go/1.21.6/x64/src/testing/testing.go:2052 +0x445
testing.(*M).Run(0xc0005a0640)
	/opt/hostedtoolcache/go/1.21.6/x64/src/testing/testing.go:1925 +0x[63](https://github.com/dapr/components-contrib/actions/runs/7830915945/job/21366093106?pr=3324#step:19:64)6
main.main()
	_testmain.go:[81](https://github.com/dapr/components-contrib/actions/runs/7830915945/job/21366093106?pr=3324#step:19:82) +0x1bf
```

sicoyle:
ping @ItalyPaleAle

sicoyle:
ping for review pls

common/component/postgresql/awsIAM/aws.go
```diff
@@ -0,0 +1,240 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to parse database name from connection string to create database %v", err)
+	}
+
+	// check if database exists in connection string
+	var dbExists bool
+	dbExistsCtx, dbExistsCancel := context.WithTimeout(ctx, timeout)
+	err = db.QueryRow(dbExistsCtx, databaseExistsTmpl, dbName).Scan(&dbExists)
+	dbExistsCancel()
+	if err != nil && err != pgx.ErrNoRows {
```
cicoyle, line 128: Should this be an OR?

```suggestion
	if err != nil || err != pgx.ErrNoRows {
```

common/component/postgresql/awsIAM/aws.go
```diff
@@ -0,0 +1,240 @@
...
+	dbUser, err := parseUserFromConnectionString(connectionString)
+	if err != nil {
+		return fmt.Errorf("failed to parse database user from connection string to create database %v", err)
+	}
+
+	var userExists bool
+	userExistsCtx, userExistsCancel := context.WithTimeout(ctx, timeout)
+	err = db.QueryRow(userExistsCtx, userExistsTmpl, dbUser).Scan(&userExists)
+	userExistsCancel()
+	if err != nil && err != pgx.ErrNoRows {
```
cicoyle, line 155: same here?

state/postgresql/v2/postgresql.go
```diff
@@ -68,28 +76,75 @@ func NewPostgreSQLStateStoreWithOptions(logger logger.Logger, opts Options) stat
...
-	err := p.metadata.InitWithMetadata(meta, p.enableAzureAD)
-	if err != nil {
-		p.logger.Errorf("Failed to parse metadata: %v", err)
-		return err
+	var (
+		useAWS   bool
+		useAzure bool
+		err      error
+	)
+	if meta.Properties["UseAWSIAM"] == "" {
```
cicoyle, line 92: nit: this would look cleaner as a switch

cicoyle: Its really a nit, but decreases the {}
```
switch meta.Properties["UseAWSIAM"] {
case "":
...
default:
...
}
```

cicoyle (approved):
I left a small suggestion, but LGTM. Thanks Sam 🚀

state/postgresql/v2/postgresql.go
ItalyPaleAle: We should make sure that all properties are read as case-insensitive. So we can't read from meta.Properties directly. Use the metadata.GetMetadataProperty method which is case-insensitive.

common/component/postgresql/awsIAM/aws.go
cicoyle: Actually, maybe you are right. Maybe it makes sense to err only if there is an err and the err is not no rows found. If the err is no rows found we probably shouldn't error out since the query just returned empty, like no rows. I'll dismiss my comment here.

common/component/postgresql/awsIAM/aws.go
```diff
@@ -0,0 +1,240 @@
...
+
+	// https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/UsingWithRDS.IAMDBAuth.Connecting.Go.html
+	// Default to load default config through aws credentials file (~/.aws/credentials)
+	awsCfg, err := aws_config.LoadDefaultConfig(ctx)
+	if err != nil {
+		// otherwise use metadata fields
+
+		// Validate if access key and secret access key are provided
+		if accessKey == "" || secretKey == "" {
+			return "", fmt.Errorf("failed to load default configuration for AWS using accessKey and secretKey")
```
ItalyPaleAle, line 62:
```suggestion
			return "", errors.New("failed to load default configuration for AWS using accessKey and secretKey")
```

ItalyPaleAle: Isn't there any error we can show? Can we include `err`?

common/component/postgresql/awsIAM/aws.go
```diff
@@ -0,0 +1,240 @@
...
+		}
+
+		// Set credentials explicitly
+		var awsCfg2 aws.CredentialsProvider
+		awsCfg2 = aws_credentials.NewStaticCredentialsProvider(accessKey, secretKey, "")
```
ItalyPaleAle, line 67:
```suggestion
		awsCfg2 := aws_credentials.NewStaticCredentialsProvider(accessKey, secretKey, "")
```

common/component/postgresql/awsIAM/aws.go
```diff
@@ -0,0 +1,240 @@
...
+		authenticationToken, err = aws_auth.BuildAuthToken(
+			ctx, dbEndpoint, region, pgCfg.User, awsCfg2)
+		if err != nil {
+			return "", fmt.Errorf("failed to create AWS authentication token: %v", err)
```
ItalyPaleAle, line 71: in `fmt.Errorf`, errors should have `%w`

```suggestion
			return "", fmt.Errorf("failed to create AWS authentication token: %w", err)
```

common/component/postgresql/awsIAM/aws.go
```diff
@@ -0,0 +1,240 @@
...
+		}
+
+		return authenticationToken, nil
+	}
+
+	authenticationToken, err = aws_auth.BuildAuthToken(
+		ctx, dbEndpoint, region, pgCfg.User, awsCfg.Credentials)
+	if err != nil {
+		return "", fmt.Errorf("failed to create AWS authentication token: %v", err)
```
ItalyPaleAle, line 80:
```suggestion
		return "", fmt.Errorf("failed to create AWS authentication token: %w", err)
```

common/component/postgresql/awsIAM/aws.go
```diff
@@ -0,0 +1,240 @@
...
+		return "", fmt.Errorf("unable to find database password field in the connection string")
+	}
+
+	// Return the value after "password="
+	return match[1], nil
+}
+
+func CreateDatabaseIfNeeded(ctx context.Context, timeout time.Duration, connectionString string, db pginterfaces.PGXPoolConn) error {
+	// Parse database name from connection string
+	dbName, err := parseDatabaseFromConnectionString(connectionString)
```
ItalyPaleAle, line 118: I'm really not a fan of using a regex for these parsing. The pgx library already has a connstring parser which is more reliable, and we should use that: https://pkg.go.dev/github.com/jackc/pgx/v5#ParseConfig

common/component/postgresql/awsIAM/aws.go
```diff
@@ -0,0 +1,240 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to parse database name from connection string to create database %v", err)
```
ItalyPaleAle, line 120:
```suggestion
		return fmt.Errorf("failed to parse database name from connection string to create database: %w", err)
```

common/component/postgresql/awsIAM/aws.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	// check if database exists in connection string
+	var dbExists bool
+	dbExistsCtx, dbExistsCancel := context.WithTimeout(ctx, timeout)
+	err = db.QueryRow(dbExistsCtx, databaseExistsTmpl, dbName).Scan(&dbExists)
+	dbExistsCancel()
+	if err != nil && err != pgx.ErrNoRows {
+		return fmt.Errorf("failed to check if the PostgreSQL database %s exists: %v", dbName, err)
```
ItalyPaleAle, line 129:
```suggestion
		return fmt.Errorf("failed to check if the PostgreSQL database %s exists: %w", dbName, err)
```

common/component/postgresql/awsIAM/aws.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	// Create database if needed using master password in connection string
+	if !dbExists {
+		createDBCtx, createDBCancel := context.WithTimeout(ctx, timeout)
+		_, err := db.Exec(createDBCtx, fmt.Sprintf(createDatabaseTmpl, dbName))
+		createDBCancel()
+		if err != nil {
+			return fmt.Errorf("failed to create PostgreSQL user: %v", err)
```
ItalyPaleAle, line 138:
```suggestion
			return fmt.Errorf("failed to create PostgreSQL user: %w", err)
```

common/component/postgresql/awsIAM/aws.go
```diff
@@ -0,0 +1,240 @@
...
+	match := databasePasswordRegex.FindStringSubmatch(connectionString)
+	if len(match) < 2 {
+		return "", fmt.Errorf("unable to find database password field in the connection string")
+	}
+
+	// Return the value after "password="
+	return match[1], nil
+}
+
+func CreateDatabaseIfNeeded(ctx context.Context, timeout time.Duration, connectionString string, db pginterfaces.PGXPoolConn) error {
```
ItalyPaleAle, line 116: These methods are not specific to AWS, right? We should move this to the generic Postgres package so they can be used even when not using AWS.

common/component/postgresql/awsIAM/aws.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+func InitAWSDatabase(ctx context.Context, config *pgxpool.Config, timeout time.Duration, connString string, region string, awsAccessKey, awsSecretKey string) error {
+	// Note: check if password supplied, otherwise auto generate the AWS access token immediately
+	var (
+		err       error
+		pass, pwd string
+		db        *pgxpool.Pool
+	)
+	pass, err = parsePasswordFromConnectionString(connString)
```
ItalyPaleAle, line 194: Same, please don't use regexs. Use https://pkg.go.dev/github.com/jackc/pgx/v5#ParseConfig

common/component/postgresql/awsIAM/aws.go
```diff
@@ -0,0 +1,240 @@
...
+	if err != nil || pass == "" {
+		pwd, err = GetAccessToken(ctx, config.ConnConfig, region, awsAccessKey, awsSecretKey)
+		if err != nil || pwd == "" {
+			return fmt.Errorf("failed to refresh access token for iam authentication with postgresql: %v", err)
```
ItalyPaleAle, line 198:
```suggestion
			return fmt.Errorf("failed to refresh access token for IAM authentication with PostgreSQL: %w", err)
```

common/component/postgresql/awsIAM/aws.go
```diff
@@ -0,0 +1,240 @@
...
+		}
+		config.ConnConfig.Password = pwd
+	}
+
+	db, err = pgxpool.NewWithConfig(ctx, config)
+	if err != nil {
+		return fmt.Errorf("postgres configuration store connection error: %w", err)
```
ItalyPaleAle, line 205: Configuration store?

common/component/postgresql/awsIAM/aws.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	// Create database and user with proper iam role if not using an already created iam user
+	if pass != "" {
+		err = CreateDatabaseIfNeeded(ctx, timeout, connString, db)
+		if err != nil {
+			return fmt.Errorf("failed create AWS database if needed %v", err)
```
ItalyPaleAle, line 212:
```suggestion
			return fmt.Errorf("failed create AWS database if needed: %w", err)
```

common/component/postgresql/awsIAM/aws.go
```diff
@@ -0,0 +1,240 @@
...
+		}
+
+		err = CreateUserAndRoleIfNeeded(ctx, timeout, connString, db)
+		if err != nil {
+			return fmt.Errorf("failed create AWS user and grant role if needed %v", err)
```
ItalyPaleAle, line 217:
```suggestion
			return fmt.Errorf("failed create AWS user and grant role if needed: %w", err)
```

common/component/postgresql/awsIAM/aws.go
```diff
@@ -0,0 +1,240 @@
...
+	// Note: this will refresh connections before the 15 min expiration on the IAM AWS auth token,
+	// while leveraging the BeforeConnect hook to recreate the token in time dynamically.
+	config.MaxConnLifetime = time.Minute * 14
+
+	// Setup connection pool config needed for AWS IAM authentication
+	config.BeforeConnect = func(ctx context.Context, pgConfig *pgx.ConnConfig) error {
+		// Manually reset auth token with aws and reset the config password using the new iam token
+		pwd, err := GetAccessToken(ctx, pgConfig, region, awsAccessKey, awsSecretKey)
+		if err != nil {
+			return fmt.Errorf("failed to refresh access token for iam authentication with postgresql: %v", err)
```
ItalyPaleAle, line 231:
```suggestion
			return fmt.Errorf("failed to refresh access token for IAM authentication with PostgreSQL: %w", err)
```

common/component/postgresql/awsIAM/aws.go
```diff
@@ -0,0 +1,240 @@
...
+func InitAWSDatabase(ctx context.Context, config *pgxpool.Config, timeout time.Duration, connString string, region string, awsAccessKey, awsSecretKey string) error {
+	// Note: check if password supplied, otherwise auto generate the AWS access token immediately
+	var (
+		err       error
+		pass, pwd string
+		db        *pgxpool.Pool
+	)
+	pass, err = parsePasswordFromConnectionString(connString)
+	if err != nil || pass == "" {
+		pwd, err = GetAccessToken(ctx, config.ConnConfig, region, awsAccessKey, awsSecretKey)
```
ItalyPaleAle, line 196: I don't think we need this, because you already have a `BeforeConnect` hook below, so that will set the password

common/component/postgresql/v1/metadata_test.go
```diff
@@ -29,7 +30,8 @@ func TestMetadata(t *testing.T) {
 		m := pgMetadata{}
 		props := map[string]string{}
 
-		err := m.InitWithMetadata(state.Metadata{Base: metadata.Base{Properties: props}}, false)
+		opts := postgresql.InitWithMetadataOpts{AzureADEnabled: false, AWSIAMEnabled: false}
```
ItalyPaleAle, line 33: Nit: no need to specify the default values.

```suggestion
		opts := postgresql.InitWithMetadataOpts{}
```

common/component/postgresql/v1/postgresql.go
```diff
@@ -80,23 +84,47 @@ func NewPostgreSQLStateStore(logger logger.Logger, opts Options) state.Store {
...
+	// Note: if AWS IAM enabled then must use master connection string to connect initially,
+	// otherwise connect using regular p.metadata.ConnectionString.
+	if p.enableAWSIAM {
+		config, err = p.metadata.GetPgxPoolConfig()
+		if err != nil {
+			p.logger.Error(err)
+			return err
+		}
+
+		err = awsiam.InitAWSDatabase(ctx, config, p.metadata.Timeout, p.metadata.ConnectionString, p.metadata.AWSRegion, p.metadata.AWSAccessKey, p.metadata.AWSSecretKey)
```
ItalyPaleAle, line 116: Same, this should be moved to the `GetPgxPoolConfig` method

configuration/postgres/metadata.go
```diff
@@ -55,8 +62,13 @@ func (m *metadata) InitWithMetadata(meta map[string]string) error {
 		return fmt.Errorf("invalid table name '%s'. non-alphanumerics or upper cased table names are not supported", m.ConfigTable)
 	}
 
-	// Azure AD auth is supported for this component
-	err = m.PostgresAuthMetadata.InitWithMetadata(meta, true)
+	opts := pgauth.InitWithMetadataOpts{
+		AzureADEnabled: m.UseAzureAD,
+		AWSIAMEnabled:  m.UseAWSIAM,
```
ItalyPaleAle, line 67: Always `true` here as it indicates where it _can_ be used
```suggestion
		AzureADEnabled: true,
		AWSIAMEnabled:  true,
```

configuration/postgres/metadata.yaml
```diff
@@ -28,6 +28,49 @@ builtinAuthenticationProfiles:
...
+        required: false
+        sensitive: true
+        description: |
+          The secret key associated with the access key.
+        example: '"wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY"'
+      - name: awsSessionToken
+        type: string
+        sensitive: true
+        description: |
+          AWS session token to use. A session token is only required if you are using\ntemporary security credentials.
```
ItalyPaleAle, line 72:
```suggestion
          AWS session token to use. A session token is only required if you are using temporary security credentials.
```

configuration/postgres/postgres.go
```diff
@@ -284,19 +286,41 @@ func (p *ConfigurationStore) handleSubscribedChange(ctx context.Context, handler
...
+		err    error
+	)
+	// Note: if AWS IAM enabled then must use master connection string to connect initially,
+	// otherwise connect using regular p.metadata.ConnectionString.
+	if p.metadata.UseAWSIAM {
+		config, err = p.metadata.GetPgxPoolConfig()
+		if err != nil {
+			return nil, fmt.Errorf("postgres configuration store connection error: %w", err)
+		}
+		err = awsiam.InitAWSDatabase(ctx, config, p.metadata.Timeout, connStr, p.metadata.AWSRegion, p.metadata.AWSAccessKey, p.metadata.AWSSecretKey)
```
ItalyPaleAle, line 302: Ditto, must be moved to the `GetPgxPoolConfig`

Let's strive for code reuse please, rather than copy-pasting, to improve maintainability

configuration/postgres/postgres.go
```diff
@@ -284,19 +286,41 @@ func (p *ConfigurationStore) handleSubscribedChange(ctx context.Context, handler
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to init AWS database: %v", err)
+		}
+	} else {
+		config, err = p.metadata.GetPgxPoolConfig()
+		if err != nil {
+			p.logger.Error(err)
+			return nil, fmt.Errorf("failed to parse state store pool config: %v", err)
```
ItalyPaleAle, line 310:
```suggestion
			return nil, fmt.Errorf("postgres configuration store connection error : %w", err)
```

state/postgresql/v1/metadata.yaml
```diff
@@ -35,6 +35,49 @@ builtinAuthenticationProfiles:
...
+        required: false
+        sensitive: true
+        description: |
+          The secret key associated with the access key.
+        example: '"wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY"'
+      - name: awsSessionToken
+        type: string
+        sensitive: true
+        description: |
+          AWS session token to use. A session token is only required if you are using\ntemporary security credentials.
```
ItalyPaleAle, line 79:
```suggestion
          AWS session token to use. A session token is only required if you are using temporary security credentials.
```

state/postgresql/v2/metadata.yaml
```diff
@@ -32,8 +32,51 @@ builtinAuthenticationProfiles:
...
+        required: false
+        sensitive: true
+        description: |
+          The secret key associated with the access key.
+        example: '"wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY"'
+      - name: awsSessionToken
+        type: string
+        sensitive: true
+        description: |
+          AWS session token to use. A session token is only required if you are using\ntemporary security credentials.
```
ItalyPaleAle, line 78:
```suggestion
          AWS session token to use. A session token is only required if you are using temporary security credentials.
```

state/postgresql/v2/metadata_test.go
```diff
@@ -29,7 +30,8 @@ func TestMetadata(t *testing.T) {
 		m := pgMetadata{}
 		props := map[string]string{}
 
-		err := m.InitWithMetadata(state.Metadata{Base: metadata.Base{Properties: props}}, false)
+		opts := postgresql.InitWithMetadataOpts{AzureADEnabled: false, AWSIAMEnabled: false}
```
ItalyPaleAle, line 33: Nit


```suggestion
		opts := postgresql.InitWithMetadataOpts{}
```

state/postgresql/v2/postgresql.go
```diff
@@ -68,28 +76,75 @@ func NewPostgreSQLStateStoreWithOptions(logger logger.Logger, opts Options) stat
...
+	// Note: if AWS IAM enabled then must use master connection string to connect initially,
+	// otherwise connect using regular p.metadata.ConnectionString.
+	if useAWS {
+		config, err = p.metadata.GetPgxPoolConfig()
+		if err != nil {
+			p.logger.Error(err)
+			return err
+		}
+
+		err = awsiam.InitAWSDatabase(ctx, config, p.metadata.Timeout, p.metadata.ConnectionString, p.metadata.AWSRegion, p.metadata.AWSAccessKey, p.metadata.AWSSecretKey)
```
ItalyPaleAle, line 131: Same, move this to `GetPgxPoolConfig`

tests/utils/configupdater/postgres/postgres.go
```diff
@@ -87,8 +87,9 @@ func (r *ConfigUpdater) Init(props map[string]string) error {
 	md := pgauth.PostgresAuthMetadata{
 		ConnectionString: props["connectionString"],
 		UseAzureAD:       utils.IsTruthy(props["useAzureAD"]),
+		UseAWSIAM:        utils.IsTruthy(props["useAWSIAM"]),
```
ItalyPaleAle, line 90: Here too we should use `metadata.GetMetadataProperty` (the previous code was incorrect to)

ItalyPaleAle (changes requested):
Thanks for this PR.

Please move all the code for authenticating with AWS IAM to a shared package. If it's specific for Postgres, it should be in one of the common packages we have for Postgres.

This way, you can also enable it for all PG components (not just the state store v1, but also the binding and configuration store)

common/component/postgresql/awsIAM/aws.go
sicoyle: not really sure what you're getting at here

ItalyPaleAle: The error message says "configuration store" - but this is for a generic component, not just a config store.

bindings/postgres/postgres.go
```diff
@@ -67,28 +66,10 @@ func (p *Postgres) Init(ctx context.Context, meta bindings.Metadata) error {
...
-		}
-	} else {
-		poolConfig, err = m.GetPgxPoolConfig()
-		if err != nil {
-			p.logger.Error(err)
-			return err
-		}
+	poolConfig, err := m.GetPgxPoolConfig()
+	if err != nil {
+		p.logger.Error(err)
```
ItalyPaleAle, line 71: Dapr will log errors already

```suggestion
```

common/authentication/postgresql/metadata.go
```diff
@@ -19,11 +19,15 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/internal/exported"
```
ItalyPaleAle, line 22: Note that you're trying to import an `internal` package, which cannot be imported (that's just a limitation in Go)

common/authentication/postgresql/metadata.go
```diff
@@ -120,7 +129,8 @@ func (m *PostgresAuthMetadata) GetPgxPoolConfig() (*pgxpool.Config, error) {
 		// This is because tokens expire, and connections can drop and need to be re-established at any time
 		// Fortunately, we can do this with the "BeforeConnect" hook
 		config.BeforeConnect = func(ctx context.Context, cc *pgx.ConnConfig) error {
-			at, err := tokenCred.GetToken(ctx, policy.TokenRequestOptions{
+			var at exported.AccessToken
+			at, err = tokenCred.GetToken(ctx, policy.TokenRequestOptions{
```
ItalyPaleAle, line 133: In this specific case, the old code was correct.

This is inside a "lambda", so we should use an `err` that is scoped within the function.

The current code, however, uses `err` from the scope of the parent. That can cause unexpected issues if the method is invoked concurrently

```suggestion
			at, err := tokenCred.GetToken(ctx, policy.TokenRequestOptions{
```

common/authentication/postgresql/metadata.go
```diff
@@ -133,6 +143,25 @@ func (m *PostgresAuthMetadata) GetPgxPoolConfig() (*pgxpool.Config, error) {
...
+			return nil, fmt.Errorf("failed to find AWSRegion metadata field")
+		}
+		awsAccessKey, ok := metadata.GetMetadataProperty(m.awsEnv.Metadata, "AWSAccessKey")
+		if !ok {
+			return nil, fmt.Errorf("failed to find AWSAccessKey metadata field")
+		}
+		awsSecretKey, ok := metadata.GetMetadataProperty(m.awsEnv.Metadata, "AWSSecretKey")
+		if !ok {
+			return nil, fmt.Errorf("failed to find AWSSecretKey metadata field")
+		}
```
ItalyPaleAle, line 158: Here, you want to check not only that the metadata property exists, but also that it's not empty

```suggestion
	if m.UseAWSIAM {
		awsRegion, _ := metadata.GetMetadataProperty(m.awsEnv.Metadata, "AWSRegion")
		if awsRegion == "" {
			return nil, errors.New("metadata property AWSRegion is missing")
		}
		awsAccessKey, _ := metadata.GetMetadataProperty(m.awsEnv.Metadata, "AWSAccessKey")
		if awsAccessKey == "" {
			return nil, errors.New("metadata property AWSAccessKey is missing")
		}
		awsSecretKey, _ := metadata.GetMetadataProperty(m.awsEnv.Metadata, "AWSSecretKey")
		if awsSecretKey == "" {
			return nil, errors.New("metadata property AWSSecretKey is missing")
		}
```

common/authentication/postgresql/metadata.go
```diff
@@ -133,6 +143,25 @@ func (m *PostgresAuthMetadata) GetPgxPoolConfig() (*pgxpool.Config, error) {
...
 
+		err = awsiam.InitAWSDatabase(context.Background(), config, m.ConnectionString, awsRegion, awsAccessKey, awsSecretKey)
+		if err != nil {
+			err = fmt.Errorf("failed to init AWS database: %v", err)
```
ItalyPaleAle, line 162:
```suggestion
			err = fmt.Errorf("failed to init AWS database: %w", err)
```

common/component/postgresql/awsIAM/aws.go
```diff
@@ -77,146 +55,14 @@ func GetAccessToken(ctx context.Context, pgCfg *pgx.ConnConfig, region, accessKe
...
-			return fmt.Errorf("failed create AWS database if needed %v", err)
-		}
-
-		err = CreateUserAndRoleIfNeeded(ctx, timeout, connString, db)
-		if err != nil {
-			return fmt.Errorf("failed create AWS user and grant role if needed %v", err)
-		}
-	}
+func InitAWSDatabase(ctx context.Context, config *pgxpool.Config, connString string, region string, awsAccessKey, awsSecretKey string) error {
+	timeout := 15 * time.Second
```
ItalyPaleAle, line 65:
```suggestion
	const timeout = 15 * time.Second
```

common/component/postgresql/awsIAM/aws.go
```diff
@@ -228,13 +74,30 @@ func InitAWSDatabase(ctx context.Context, config *pgxpool.Config, timeout time.D
...
+	if err != nil {
+		return fmt.Errorf("failed create AWS database if needed %w", err)
+	}
+
+	err = databases.CreateUserAndRoleIfNeeded(ctx, timeout, connString, db)
+	if err != nil {
+		return fmt.Errorf("failed create AWS user and grant role if needed %w", err)
+	}
+
 	// TODO: add clean up hooks on user/db?
```
ItalyPaleAle, line 101: What's this TODO for?

common/component/postgresql/databases/databases.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+const (
+	createDatabaseTmpl = "CREATE DATABASE %s"
+	deleteDatabaseTmpl = "DROP DATABASE IF EXISTS %s WITH (FORCE)"
+	databaseExistsTmpl = "SELECT EXISTS (SELECT 1 FROM pg_database WHERE datname = $1)"
+	userExistsTmpl     = "SELECT CAST((SELECT 1 FROM pg_roles WHERE rolname = $1) AS BOOLEAN)"
+	createUserTmpl     = "CREATE USER %v WITH PASSWORD '%v'"
+	checkRoleTmpl      = "SELECT 1 FROM pg_roles WHERE rolname = $1 AND 'rds_iam' = rolname"
+	grantRoleTmpl      = "GRANT %v TO %v"
```
ItalyPaleAle, line 19: Most of these queries use `%s` or other string formatting, so are potentially vulnerable to SQL injection attacks. Although this is a bit mitigated by the fact that the data comes from the component's metadata, so it's set by an admin, It would be great if we could try and use parametrized queries (e.g. queries that use `$1` et similia) instead. Otherwise, we may get this flagged in the next security audit!

If using parametrized queries is not possible (sometimes it's not for queries that change the system), we should sanitize the input to make sure it's limited to characters that can't cause SQL injections. Example, what we do for MySQL: https://github.dev/dapr/components-contrib/blob/49afc557e4d7b836e8077017c503248f8fb9d6aa/state/mysql/mysql.go#L825-L844

common/component/postgresql/databases/databases.go
```diff
@@ -0,0 +1,91 @@
...
+		return fmt.Errorf("failed to parse PostgreSQL config: %w", err)
+	}
+
+	var userExists bool
+	userExistsCtx, userExistsCancel := context.WithTimeout(ctx, timeout)
+	err = db.QueryRow(userExistsCtx, userExistsTmpl, config.User).Scan(&userExists)
+	userExistsCancel()
+	if err != nil && err != pgx.ErrNoRows {
+		// If user does not exist, create the user with pwd = user name for now.
+		// This will be rotated to use proper iam credentials before connecting again.
```
ItalyPaleAle, line 63: I really do not like this.

Although the code below changes the password to IAM credentials, if Dapr crashes half-way (between the time the user is created and its password is changed), you are stuck with a user that has password = user name. That is a HUGE security risk.

common/component/postgresql/v1/postgresql.go
```diff
@@ -103,28 +102,10 @@ func (p *PostgreSQL) Init(ctx context.Context, meta state.Metadata) error {
...
-		}
-	} else {
-		config, err = p.metadata.GetPgxPoolConfig()
-		if err != nil {
-			p.logger.Error(err)
-			return err
-		}
+	config, err := p.metadata.GetPgxPoolConfig()
+	if err != nil {
+		p.logger.Error(err)
```
ItalyPaleAle, line 107:
```suggestion
```

configuration/postgres/postgres.go
```diff
@@ -287,38 +286,20 @@ func (p *ConfigurationStore) handleSubscribedChange(ctx context.Context, handler
...
-		}
-	} else {
-		config, err = p.metadata.GetPgxPoolConfig()
-		if err != nil {
-			p.logger.Error(err)
-			return nil, fmt.Errorf("failed to parse state store pool config: %v", err)
-		}
+	config, err := p.metadata.GetPgxPoolConfig()
+	if err != nil {
+		p.logger.Error(err)
```
ItalyPaleAle, line 291:
```suggestion
```

state/postgresql/v2/postgresql.go
```diff
@@ -89,19 +88,23 @@ func (p *PostgreSQL) Init(ctx context.Context, meta state.Metadata) error {
 		useAzure bool
 		err      error
 	)
-	if meta.Properties["UseAWSIAM"] == "" {
+
+	awsIam, ok := metadata.GetMetadataProperty(meta.Properties, "UseAWSIAM")
```
ItalyPaleAle, line 92: `utils.IsTruthy` please :)

Replace the entire block with this (which handles the case of the property missing already, no need to check for `ok`):

```go
awsIam, _ := metadata.GetMetadataProperty(meta.Properties, "UseAWSIAM")
useAWS = utils.IsTruthy(awsIam)
```

state/postgresql/v2/postgresql.go
```diff
@@ -89,19 +88,23 @@ func (p *PostgreSQL) Init(ctx context.Context, meta state.Metadata) error {
...
+		useAWS, err = strconv.ParseBool(awsIam)
 		if err != nil {
 			p.logger.Error(err)
 			return err
 		}
 	}
-	if meta.Properties["UseAzureAD"] == "" {
-		useAWS = false
+
+	azureAd, ok := metadata.GetMetadataProperty(meta.Properties, "UseAzureAD")
```
ItalyPaleAle, line 103: Same as above, please replace the entire block with this 2 lines:

```go
azureAd, _ := metadata.GetMetadataProperty(meta.Properties, "UseAzureAD")
useAzure = utils.IsTruthy(azureAd)
```

state/postgresql/v2/postgresql.go
```diff
@@ -118,28 +121,10 @@ func (p *PostgreSQL) Init(ctx context.Context, meta state.Metadata) error {
...
-		}
-	} else {
-		config, err = p.metadata.GetPgxPoolConfig()
-		if err != nil {
-			p.logger.Error(err)
-			return err
-		}
+	config, err := p.metadata.GetPgxPoolConfig()
+	if err != nil {
+		p.logger.Error(err)
```
ItalyPaleAle, line 126:
```suggestion
```

ItalyPaleAle (changes requested):

common/authentication/postgresql/metadata.go
sicoyle: yeah i know we cant pull from internal. I think my IDE messed up by not telling me the err here lol

common/component/postgresql/awsIAM/aws.go
sicoyle: if we need to have clean up logic, but I suppose if the user is specifying a db and user, then clean up is outside of the realm of dapr?

ItalyPaleAle: Cleanup for what things?

common/component/postgresql/databases/databases.go
sicoyle: addressed

common/component/postgresql/databases/databases.go
sicoyle: addressed. thoughts on changes?

sicoyle:
ready

sicoyle:
ping @ItalyPaleAle ready :)

sicoyle:
ping for review please

sicoyle:
ready for review pls

sicoyle:
@JoshVanL another one for ya please whenever you're get a chance :)

common/authentication/postgresql/metadata.go
```diff
@@ -112,19 +128,38 @@ func (m *PostgresAuthMetadata) GetPgxPoolConfig() (*pgxpool.Config, error) {
...
+		awsAccessKey, _ := metadata.GetMetadataProperty(m.awsEnv.Metadata, "AWSAccessKey")
+		if awsAccessKey == "" {
+			return nil, errors.New("metadata property AWSAccessKey is missing")
+		}
+		awsSecretKey, _ := metadata.GetMetadataProperty(m.awsEnv.Metadata, "AWSSecretKey")
+		if awsSecretKey == "" {
+			return nil, errors.New("metadata property AWSSecretKey is missing")
+		}
 
+		err = awsiam.InitAWSDatabase(context.Background(), config, m.ConnectionString, awsRegion, awsAccessKey, awsSecretKey)
```
ItalyPaleAle, line 158: We need to pass a context so this can be canceled. `context.Background` isn't a valid option. Or @JoshVanL may come chasing after you :)

common/authentication/postgresql/metadata.go
```diff
@@ -112,19 +128,38 @@ func (m *PostgresAuthMetadata) GetPgxPoolConfig() (*pgxpool.Config, error) {
...
+			return nil, errors.New("metadata property AWSRegion is missing")
+		}
+		awsAccessKey, _ := metadata.GetMetadataProperty(m.awsEnv.Metadata, "AWSAccessKey")
+		if awsAccessKey == "" {
+			return nil, errors.New("metadata property AWSAccessKey is missing")
+		}
+		awsSecretKey, _ := metadata.GetMetadataProperty(m.awsEnv.Metadata, "AWSSecretKey")
+		if awsSecretKey == "" {
+			return nil, errors.New("metadata property AWSSecretKey is missing")
+		}
```
ItalyPaleAle, line 156: This should be moved to a common package, not done within the component. When managing a project like this, reducing code duplication is very important!

common/component/postgresql/awsIAM/aws.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		// otherwise use metadata fields
+
+		// Validate if access key and secret access key are provided
+		if accessKey == "" || secretKey == "" {
+			return "", fmt.Errorf("failed to load default configuration for AWS using accessKey and secretKey: %w", err)
+		}
+
+		// Set credentials explicitly
+		awsCfg2 := aws_credentials.NewStaticCredentialsProvider(accessKey, secretKey, "")
```
ItalyPaleAle, line 46: Using `2` is generally a code smell.

```suggestion
		awsCfg = aws_credentials.NewStaticCredentialsProvider(accessKey, secretKey, "")
```

common/component/postgresql/databases/databases.go
ItalyPaleAle: I see the change but I am still not sure it's a good idea. Now you're creating a user without a password, and if Daprd crashes half-way (and maybe never restarts!), you are left with a user without a password. This is a MASSIVE security risk.

I would recommend removing all code that creates the database from Dapr. We should add it to the docs as a pre-requisite instead.

ItalyPaleAle (changes requested):

sicoyle:
@ItalyPaleAle this should be ready for rereview please

sicoyle:
Ready for rereview please :)

sicoyle:
This is ready for review please. Looks to me like the test failures are unrelated

sicoyle:
can i get some feedback pretty please cc @JoshVanL @ItalyPaleAle

common/component/postgresql/awsIAM/aws.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func GetAccessToken(ctx context.Context, pgCfg *pgx.ConnConfig, region, accessKey, secretKey string) (string, error) {
+	dbEndpoint := pgCfg.Host + ":" + strconv.Itoa(int(pgCfg.Port))
+	var authenticationToken string
+
+	// https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/UsingWithRDS.IAMDBAuth.Connecting.Go.html
+	// Default to load default config through aws credentials file (~/.aws/credentials)
+	awsCfg, err := aws_config.LoadDefaultConfig(ctx)
+	if err != nil {
```
JoshVanL, line 35: Can this error for any other reason other than the credential not being available at the well known locations? i.e. is this returned error typed which we need to select over.

JoshVanL (changes requested):

state/postgresql/v2/postgresql.go
```diff
@@ -63,36 +71,57 @@ func NewPostgreSQLStateStore(logger logger.Logger) state.Store {
...
+	config, err := p.metadata.GetPgxPoolConfig(ctx)
 	if err != nil {
-		p.logger.Error(err)
 		return err
 	}
 
 	connCtx, connCancel := context.WithTimeout(ctx, p.metadata.Timeout)
 	p.db, err = pgxpool.NewWithConfig(connCtx, config)
-	connCancel()
+	defer connCancel()
```
berndverst, line 122: @ItalyPaleAle can you confirm whether this should be deferred here? I know sometimes we want to immediately cancel the context.

common/authentication/postgresql/metadata.go
```diff
@@ -43,34 +48,61 @@ func (m *PostgresAuthMetadata) Reset() {
...
+		// Make sure these are false
 		m.UseAzureAD = false
+		m.UseAWSIAM = false
 	}
 
 	return nil
 }
 
+func (m *PostgresAuthMetadata) ValidateAwsIamFields() (string, string, string, error) {
+	awsRegion, _ := metadata.GetMetadataProperty(m.awsEnv.Metadata, "AWSRegion")
```
berndverst, line 89: Do not use `GetMetadataProperty` except in testing.

Instead, in the original metadata options struct you can define a pointer type if you want to be able to check whether an option was set or not. Not set means nil value in that case.

DecodeMetadata will marshal the metadata into the appropriate struct, and that's the only way you should access the metadata - via the metadata struct.

state/postgresql/v2/postgresql.go
```diff
@@ -63,36 +71,57 @@ func NewPostgreSQLStateStore(logger logger.Logger) state.Store {
...
 func (p *PostgreSQL) Init(ctx context.Context, meta state.Metadata) error {
-	err := p.metadata.InitWithMetadata(meta, p.enableAzureAD)
+	var (
+		useAWS   bool
+		useAzure bool
+		err      error
+	)
+
+	awsIam, _ := metadata.GetMetadataProperty(meta.Properties, "UseAWSIAM")
+	useAWS = utils.IsTruthy(awsIam)
```
berndverst, line 101: There is no need for any of this - our custom DecodeMetadata function already performs the IsTruthy check via decodehooks.

Please do not use `metadata.GetMetadataProperty`.

Instead, make sure on the original component metadata you have a property `UseAWSIAM` of type bool (or *bool if you do not want a default value). Then only access the value via the metadata struct.

common/authentication/postgresql/metadata.go
```diff
@@ -32,9 +35,11 @@ type PostgresAuthMetadata struct {
 	ConnectionMaxIdleTime time.Duration `mapstructure:"connectionMaxIdleTime"`
 	MaxConns              int           `mapstructure:"maxConns"`
 	UseAzureAD            bool          `mapstructure:"useAzureAD"`
+	UseAWSIAM             bool          `mapstructure:"useAWSIAM"`
```
berndverst, line 38: This line right here is what already automatically handles the IsTruthy check etc when the metadata gets unmarshalled into the struct using DecodeMetadata.

Only access metadata via this struct - do not manual access metadata except in testing!

tests/utils/configupdater/postgres/postgres.go
```diff
@@ -84,11 +85,16 @@ func (r *ConfigUpdater) CreateTrigger(channel string) error {
 }
 
 func (r *ConfigUpdater) Init(props map[string]string) error {
+	connString, _ := metadata.GetMetadataProperty(props, "connectionString")
```
berndverst, line 88: Here the use of GetMetadataProperty and IsTruthy is ok :)

common/component/postgresql/awsIAM/aws.go
berndverst, line 1: I think this file should be moved to `common/authentication/postgresql/awsIAM/aws.go` instead, and general reusable AWS auth stuff that isn't component specific belongs in `common/authentication/aws`

bindings/postgres/metadata.go
```diff
@@ -14,26 +14,41 @@ limitations under the License.
...
 	kitmd "github.com/dapr/kit/metadata"
 )
 
+const (
+	defaultTimeout = 20 * time.Second // Default timeout for network requests
+)
+
 type psqlMetadata struct {
 	pgauth.PostgresAuthMetadata `mapstructure:",squash"`
+	awsiam.AWSIAM               `mapstructure:",squash"`
```
berndverst, line 30: Just FYI - by including this here each included metadata (which is merged into this struct) must be included in the `metadata.yaml` file. There is a CI task that will validate this.
You may have already noticed this because you updated the metadata.yaml - so no action on that here.

berndverst (changes requested):

common/component/postgresql/awsIAM/aws.go
sicoyle: I see that there are 5 options to choose from when it comes to aws config pkg errors, and only one does not apply here, so I've updated accordingly.

state/postgresql/v2/postgresql.go
sicoyle: There is a ping later on in the func so yes I think deferring is fine

common/authentication/postgresql/metadata.go
sicoyle: I did it this way based on feedback prior in this PR to do it this way, and am doing it the same way that the Azure Environment Settings get picked up. See here in `bindings/azure/eventgrid/eventgrid.go` for the same setup that I am using for consistency.
```

func (a *AzureEventGrid) ensureInputBindingMetadata() error {
	a.metadata.azureTenantID, _ = metadata.GetMetadataProperty(a.metadata.properties, azauth.MetadataKeys["TenantID"]...)
	if a.metadata.azureTenantID == "" {
		return errors.New("metadata field 'azureTenantID' is empty in EventGrid binding")
	}
	a.metadata.azureClientID, _ = metadata.GetMetadataProperty(a.metadata.properties, azauth.MetadataKeys["ClientID"]...)
	if a.metadata.azureClientID == "" {
		return errors.New("metadata field 'azureClientID' is empty in EventGrid binding")
	}
	a.metadata.azureSubscriptionID, _ = metadata.GetMetadataProperty(a.metadata.properties, "subscriptionID", "azureSubscriptionID")
	if a.metadata.azureSubscriptionID == "" {
		return errors.New("metadata field 'azureSubscriptionID' is empty in EventGrid binding")
	}
```

state/postgresql/v2/postgresql.go
sicoyle: doing this the same as the azure setup, see my comment above on this for environment settings:
https://github.com/dapr/components-contrib/pull/3324#discussion_r1563051803

common/authentication/postgresql/metadata.go
sicoyle: see https://github.com/dapr/components-contrib/pull/3324#discussion_r1563051803

tests/utils/configupdater/postgres/postgres.go
sicoyle: maybe this function should have a Dos / Don'ts comment in the docs so others know when to and to not use this if you feel so strongly about it

sicoyle:
Thank y'all for the feedback! All feedback has been addressed or additional comments have been provided. cc @berndverst @JoshVanL @ItalyPaleAle

tests/utils/configupdater/postgres/postgres.go
berndverst: Oh I think I can do one better -- I can probably add a custom linter which will now allow the use of this function for added code :)

berndverst (dismissed):
LGTM once linter passes

sicoyle:
linter issues have been resolved. This should be g2g for merge pls and thank you!

sicoyle:
@ItalyPaleAle I think I need your approval on here too for merging please :) Bernd gave the approval yesterday and I've fixed the linter issue.

sicoyle:
gentle ping @ItalyPaleAle please 😇

sicoyle:
Can I get help/advice on what y'all want for the dependencies pls? From what I see is that the certification tests go.mod dependencies are behind what the main module dependencies are for kit and dapr. However, the main module is using kit:
```
github.com/dapr/kit](http://github.com/dapr/kit) v0.13.1-0.20240306152601-e33fbab74548
```
And contrib in main branch also has dependencies on dapr/dapr v1.13.2 bc of fields brought into
```
http://github.com/forks/dapr/components-contrib/tests/certification/embedded/embedded.go
```
For:
```
DaprHTTPMaxRequestSize:    maxRequestBodySize,
DaprHTTPReadBufferSize:    runtime.DefaultReadBufferSize,
```
How should I proceed with fixing? bc my cert tests still fail but the main module dependencies if I bump the cert tests to are incompatible...

sicoyle:
this is ready pls

tests/certification/embedded/embedded.go
```diff
@@ -50,7 +50,7 @@ const (
 	maxConcurrency      = -1
 	enableMTLS          = false
 	sentryAddress       = ""
-	maxRequestBodySize  = 4
+	maxRequestBodySize  = 4 << 20
```
berndverst, line 53: what does this magic number do?

sicoyle: see here in that pr where they are using similarly 
image

sicoyle: @berndverst lmk if you need further clarification

berndverst:
@sicoyle I mentioned a few times to remove the patch file from your PR please. The `3346.patch` file should not be in the PR. I will not approve and merge until this file is gone.

berndverst (approved):

marcduiker:
@holopin-bot @sicoyle Enjoy the Dapr components badge!

PR #3330, opened 2024-02-02
Enable configurable concurrency for pulsar

yaron2:
Reported by users, the Pulsar component has hardcoded concurrency settings which provides very limited performance at moderate to high scale scenarios, causing bottlenecks and issues.

I'm targeting this for 1.13 as Pulsar is a stable component and this is an issue that prevents a lot of production scenarios. Will merge if we have consensus for a 1.13 release, otherwise i'll target the PR against main branch for 1.14.

### DIFF
--- a/pubsub/pulsar/metadata.go
+++ b/pubsub/pulsar/metadata.go
@@ -35,6 +35,7 @@ type pulsarMetadata struct {
 	PublicKey               string                    `mapstructure:"publicKey"`
 	PrivateKey              string                    `mapstructure:"privateKey"`
 	Keys                    string                    `mapstructure:"keys"`
+	Concurrency             uint                      `mapstructure:"concurrency"`
 
 	Token                            string `mapstructure:"token"`
 	oauth2.ClientCredentialsMetadata `mapstructure:",squash"`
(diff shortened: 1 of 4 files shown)


pubsub/pulsar/metadata.go
```diff
@@ -35,6 +35,7 @@ type pulsarMetadata struct {
 	PublicKey               string                    `mapstructure:"publicKey"`
 	PrivateKey              string                    `mapstructure:"privateKey"`
 	Keys                    string                    `mapstructure:"keys"`
+	Concurrency             uint                      `mapstructure:"concurrency"`
```
ItalyPaleAle, line 38: Suggest calling this `maxConcurrentHandlers` which is the same name used in Azure Service Bus PubSub (and note that the default value there is 0, meaning unlimited)

ItalyPaleAle (changes requested):

pubsub/pulsar/metadata.go
yaron2: Renamed

ItalyPaleAle: Thanks for the rename.

Yup, sounds good to keep 100 for now. As discussed on Discord, can you please open an issue for 1.14 to allow this to be "0" (i.e. unlimited) too? Right now it's using a buffered channel which was probably not the best implementation.

ItalyPaleAle (commented):
Please open a docs issue/PR

Issue #3332, opened 2024-02-05
Redis State Store: numeric operators do not work correctly on large numbers due to formatting logic in Go


gspadotto:
## Expected Behavior

Define a **numeric** index on a Redis State Store (i.e. an index with a Timestamp attribute), then search with criteria that involve it. 

The criteria should be correctly evaluated and applied.

## Actual Behavior

When the numeric is larger than https://pkg.go.dev/fmt configured threshold, scientific notation is used to represent the number and comparisons do not work anymore in the generated Redis Query.

## Steps to Reproduce the Problem

Create a Redis State Store with an index on a numeric attribute, as described [here](https://docs.dapr.io/reference/components-reference/supported-state-stores/setup-redis/#querying-json-objects-optional).

Then invoke the Query API on it with this template curl command: 

`
curl -v --location 'http://localhost:3500/v1.0-alpha1/state/redis-state-store/query?metadata.contentType=application/json&metadata.queryIndexName=myRedisIndexName' \
        --header 'Content-Type: application/json' \
        --data '{
           "filter": {
              "GT": {
                 "header.timestamp": NUMBER_HERE
              }
           },
           "page":{
              "limit": 2
           },
           "sort":[
              {
                 "key":"header.timestamp",
                 "order":"DESC"
              }
           ]
        }'
`
Connect to Redis with RedisCLI and launch the "MONITOR" command. 

When NUMBER_HERE is 1.2 or 100000 you'll see the following outputs: 

`
"FT.SEARCH" "rawEventsIndx" "@var1:[(1.2 +inf]" "SORTBY" "var1" "DESC" "LIMIT" "0" "2" "RETURN" "2" "$.data" "$.version"
`

`
"FT.SEARCH" "rawEventsIndx" "@var1:[(100000 +inf]" "SORTBY" "var1" "DESC" "LIMIT" "0" "2" "RETURN" "2" "$.data" "$.version"
`

and you get the correct results, but when NUMBER_HERE is 1000000, you get

`
"FT.SEARCH" "rawEventsIndx" "@var1:[(1e+06 +inf]" "SORTBY" "var1" "DESC" "LIMIT" "0" "2" "RETURN" "2" "$.data" "$.version"
`

and no result. 

The solution might be (I do not know GoLang at all) in all occurrences of these lines: 
https://github.com/dapr/components-contrib/blob/main/state/redis/redis_query.go#L107-L112

use "%d" as format specifier in the default branch and add a branch for float64, float32 types
[...]

PR #3334, opened 2024-02-05
Redis State Store query: numeric operators do not work correctly on large numbers

gspadotto:
# Description

Applied changes to the formatting of numeric values.

I have no deep knowledge of why all numeric values in the Unit Tests are considered as float, so I changed the unit tests in order to make them pass. (I know this is **_bad_**)

## Issue reference

Please reference the issue this PR will close: #3332

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#3332

### DIFF
--- a/pubsub/pulsar/metadata.go
+++ b/pubsub/pulsar/metadata.go
@@ -35,6 +35,7 @@ type pulsarMetadata struct {
 	PublicKey               string                    `mapstructure:"publicKey"`
 	PrivateKey              string                    `mapstructure:"privateKey"`
 	Keys                    string                    `mapstructure:"keys"`
+	MaxConcurrentHandlers   uint                      `mapstructure:"maxConcurrentHandlers"`
 
 	Token                            string `mapstructure:"token"`
 	oauth2.ClientCredentialsMetadata `mapstructure:",squash"`
--- a/pubsub/pulsar/metadata.yaml
+++ b/pubsub/pulsar/metadata.yaml
@@ -171,3 +171,9 @@ metadata:
           {"name": "Name","type": "string"}
         ]
       }
+  - name: maxConcurrentHandlers
+    type: number
+    description: |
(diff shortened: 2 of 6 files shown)


ItalyPaleAle:
> I have no deep knowledge of why all numeric values in the Unit Tests are considered as float, so I changed the unit tests in order to make them pass. (I know this is bad)

I _think_ (haven't investigated too much however) that it's because they come from JSON. In order to stay compatible with JS (where all numbers are floats), the JSON parser uses float64 when parsing numbers into `any`. See the note here: https://pkg.go.dev/encoding/json#Unmarshal

gspadotto:
Hi @ItalyPaleAle, I have no knowledge of the process by which a given PR is associated to a specific milestone, 
so do not "roast" me for the following request of mine but... 

could this fix be scheduled for v1.13 or - at least - be included in the nightly builds (like [this](https://hub.docker.com/layers/daprio/dapr/nightly-2024-02-06/images/sha256-5f46b2191e328f7516ca08f57d9ec885d11a1dadc6798c678eae08932113c322?context=explore))?

My use cases rely on this fix to be in place so I would **really** appreciate if I could test it as soon as possible. 
I'm aware though that my requests will have to yeld to the official release strategy. 

Thanks.

ItalyPaleAle:
Sadly we are really past the code freeze date for 1.13, which was many weeks ago. We can, and sometimes do, make exceptions for bugs that are critical - in this case, because state store querying is an alpha feature (with no path to ever become stable), I am afraid it may be too late for 1.13.

We can definitely include it in a nightly build. I can pin an updated version of this repo in dapr/dapr so it can start being published.

ItalyPaleAle:
@gspadotto buona notizia. I spoke with another maintainer, we are ok cherry-picking this into 1.13. It should be included in the next RC :)

Issue #3336, opened 2024-02-06
optimize kafka component to use a single sarama client


famarting:
## Describe the proposal

Currently the dapr components implementation instantiates two different sarama clients, one for the producer and another one for the consumer.

It should be possible to instantiate a single sarama client and then call `NewSyncProducerFromClient` and `NewConsumerGroupFromClient` instead of the current approach.

berndverst:
Please do not open PRs before giving maintainers time to provide feedback on an issue.

@famarting some apps are only publishers while others are only subscribers.

yaron2:
> Please do not open PRs before giving maintainers time to provide feedback on an issue.
> 
> @famarting some apps are only publishers while others are only subscribers.

Indeed, but in case they are both, they should be able to use the same client

berndverst:
> > Please do not open PRs before giving maintainers time to provide feedback on an issue.
> > @famarting some apps are only publishers while others are only subscribers.
> 
> Indeed, but in case they are both, they should be able to use the same client

@yaron2 it doesn't matter anyway in this instance - I did some digging and under the hood the Sarama ProducerClient allocates a regular client first in its internal struct. So in that case we might as well explictly allocate it ourselves and just reuse that.

PR #3337, opened 2024-02-06
kafka share sarama client between producer and consumers

famarting:
# Description

Changes the way the kafka component initializes the producer and consumer sarama clients, now both will reuse the same underlying sarama.Client , this should reduce the number of requests made on startup to the kafka brokers.

This PR also changes the pubsub functions `GetKafka() *kafka.Kafka` and `GetClient() sarama.Client` so if using this code as a library its possible to reuse the same kafka client for other purposes

## Issue reference

Please reference the issue this PR will close: #3336

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


artursouza:
Is there any disadvantage of sharing the client between pub and sub? If so, it might be good to have it as a config in component metadata to decide to share or not the client.

famarting:
> Is there any disadvantage of sharing the client between pub and sub? If so, it might be good to have it as a config in component metadata to decide to share or not the client.

main disadvantage from my point of view is that each client fetches metadata from all the brokers and for all topics in the cluster, so if you have two clients this process is done twice

ItalyPaleAle (approved):

ItalyPaleAle (changes requested):
A bunch of Kafka tests are now failing, can you please take a look?

https://github.com/dapr/components-contrib/actions/runs/7802040724/job/21280858011?pr=3337
https://github.com/dapr/components-contrib/actions/runs/7802040724/job/21280858321?pr=3337
https://github.com/dapr/components-contrib/actions/runs/7802040743/job/21280855464?pr=3337

pubsub/kafka/kafka.go
```diff
@@ -122,6 +122,10 @@ func NewKafka(logger logger.Logger) pubsub.PubSub {
 	}
 }
 
+func (p *PubSub) GetKafka() *kafka.Kafka {
```
berndverst, line 125: Only methods related to the PubSub interface should act on this struct here.

berndverst, line 125: Remove this method.

berndverst (changes requested):
Please remove the method acting on the PubSub struct. Everything else is fine.

I would have rejected the PR ordinarily because publisher and subscriber could should have independent lifecycles, however, I did some digging into the sarama SDK and saw that the ProducerClient under the hood instantiates a regular (producer) client anyway. So given that, this change is totally fine by me.

berndverst (changes requested):
Kafka certification tests must pass before this PR can be merged.

famarting:
> Please remove the method acting on the PubSub struct. Everything else is fine.
> 
> I would have rejected the PR ordinarily because publisher and subscriber could should have independent lifecycles, however, I did some digging into the sarama SDK and saw that the ProducerClient under the hood instantiates a regular (producer) client anyway. So given that, this change is totally fine by me.

could you point me to the place in the code where it instantiates a regular client anyway? ty

pubsub/kafka/kafka.go
famarting: I wanted a way of exposing the internal common implementation of the kafka component for users of this repo as a library, given I added `func (k *Kafka) GetClient() sarama.Client` I was looking for a way of accessing it

famarting: is there any other approach for achieving this that would be accepted? i.e the entire pubsub kafka implementation could be moved to `common/component/kafka/kafka.go` equivalent to [postgres](https://github.com/dapr/components-contrib/blob/main/state/postgresql/v1/postgresql.go#L24)

which then allows for 
```
st := NewPostgreSQLStateStore(...)
pgst := st.(*postgresql.PostgreSQLQuery)
pgst.GetDB()
...
```

ItalyPaleAle: IMHO I think adding a "private" method is fine. We do that in other components and it can be useful for certification tests

berndverst:
> > Please remove the method acting on the PubSub struct. Everything else is fine.
> > I would have rejected the PR ordinarily because publisher and subscriber could should have independent lifecycles, however, I did some digging into the sarama SDK and saw that the ProducerClient under the hood instantiates a regular (producer) client anyway. So given that, this change is totally fine by me.
> 
> could you point me to the place in the code where it instantiates a regular client anyway? ty

Look in the sarama client code within their github repo.

PR #3338, opened 2024-02-06
Azure auth: do not use CLI provider by default when running in a cloud service

ItalyPaleAle:
This seems to be causing some issues for customers, and in a cloud service the CLI credentials would not work anyways.

### DIFF
--- a/common/authentication/azure/auth.go
+++ b/common/authentication/azure/auth.go
@@ -143,27 +143,8 @@ func (s EnvironmentSettings) addManagedIdentityProvider(timeout time.Duration, c
 	c := s.GetMSI()
 	msiCred, err := c.GetTokenCredential()
 
-	useTimeout := true
-	if _, ok := os.LookupEnv(identityEndpoint); ok {
-		// App Service, Functions, Service Fabric and Container Apps
-		useTimeout = false
-	} else {
-		if _, ok := os.LookupEnv(arcIMDSEndpoint); ok {
-			// Azure Arc
-			useTimeout = false
-		} else {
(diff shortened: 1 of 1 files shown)


berndverst:
This is a breaking change sorry. I personally rely on this OSS behavior. So we should not merge this.

berndverst (changes requested):
This is a breaking change IMO.

ItalyPaleAle:
> This is a breaking change IMO.

Please look at the diff. This only impacts when running in a cloud service. outside of a cloud service, CLI credentials still work

berndverst (approved):

common/authentication/azure/auth.go
```diff
@@ -499,6 +483,23 @@ func (s EnvironmentSettings) GetEnvironment(key string) (val string, ok bool) {
 	return metadata.GetMetadataProperty(s.Metadata, MetadataKeys[key]...)
 }
 
+// Returns true if the application is running on a cloud service with Managed Identity, including: Azure App Service, Azure Functions, Azure Service Fabric, Azure Container Apps, Azure Arc, Azure Cloud Shell.
```
berndverst, line 486: You now eliminated support for services powered by IMDS but which uses the default IMDS endpoint.

berndverst (changes requested):

berndverst (approved):

common/authentication/azure/auth.go
ItalyPaleAle: No there are 2 separate functions

PR #3358, opened 2024-02-21
Fix Azure App Config not working with gRPC

berndverst:
# Description

gRPC forces lowercase metadata, but the appconfig component expects camel case metadata. This means AppConfig doesn't work with gRPC.

This is now fixed.

### DIFF
--- a/configuration/azure/appconfig/appconfig.go
+++ b/configuration/azure/appconfig/appconfig.go
@@ -32,6 +32,8 @@ import (
 	"github.com/dapr/components-contrib/configuration"
 	contribMetadata "github.com/dapr/components-contrib/metadata"
 
+	kitmd "github.com/dapr/kit/metadata"
+
 	"github.com/dapr/kit/logger"
 )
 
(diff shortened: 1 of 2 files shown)


yaron2 (approved):

configuration/azure/appconfig/appconfig.go
```diff
@@ -284,9 +292,16 @@ func (r *ConfigurationStore) handleSubscribedChange(ctx context.Context, handler
...
 func (r *ConfigurationStore) getSentinelKeyFromMetadata(metadata map[string]string) string {
-	if s, ok := metadata["sentinelKey"]; ok && s != "" {
-		return s
+	type sentinelKeyMetadata = struct {
+		SentinelKey string `mapstructure:"sentinelKey"`
 	}
+	var sentinelKey sentinelKeyMetadata
+	kitmd.DecodeMetadata(metadata, &sentinelKey)
+
+	if sentinelKey.SentinelKey != "" {
```
tlund101, line 301: Don't need this check if you're returning the empty string anyway

berndverst: We'll clean this up when we make the component stable in the future. We are simply so late in the release process that I don't want to hold up the release further by seeking another approval (any code change cancels the existing approval unfortunately)

PR #3363, opened 2024-02-23
PubSub Kafka: Respect Subscribe context

JoshVanL:
Updates the kafka consumer group to correctly respect context cancellation for individual topic subscriptions. The consumer will be reloaded for every topic subscription, or un-subscription, using the same consumer group. Ensures there are no infinite tight loops, and only a singular consumer is run at a time. Ensures all go routines have returned on close.

### DIFF
--- a/common/component/kafka/subscriber.go
+++ b/common/component/kafka/subscriber.go
@@ -1,0 +1,103 @@
+/*
+Copyright 2024 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kafka
+
+import (
+	"context"
+	"errors"
+	"time"
+)
+
+// Subscribe adds a handler and configuration for a topic, and subscribes.
+// Unsubscribes to the topic on context cancel.
+func (k *Kafka) Subscribe(ctx context.Context, handlerConfig SubscriptionHandlerConfig, topics ...string) {
+	k.subscribeLock.Lock()
+	defer k.subscribeLock.Unlock()
+	for _, topic := range topics {
+		k.subscribeTopics[topic] = handlerConfig
+	}
+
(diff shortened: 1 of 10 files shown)


ItalyPaleAle:
Please note that in components-contrib we open all PR against main, and then cherry-pick in the relevant release branch (this has worked better here).

common/component/kafka/subscriber.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/cenkalti/backoff/v4"
+)
+
+// Subscribe adds a handler and configuration for a topic, and subscribes.
+// Unsubscribes to the topic on context cancel.
+func (k *Kafka) Subscribe(ctx context.Context, handlerConfig SubscriptionHandlerConfig, topics ...string) {
+	k.subscribeLock.Lock()
+	defer k.subscribeLock.Unlock()
+	for _, topic := range topics {
+		k.subscribeTopics[topic] = handlerConfig
```
ItalyPaleAle, line 29: Should this check if there's already a value for `k.subscribeTopics`? Otherwise, we may be in a bad state, where one subscriber stopping causes the other subscriber to stop receiving messages (see the `delete` below)

pubsub/kafka/kafka.go
```diff
@@ -54,7 +54,9 @@ func (p *PubSub) Subscribe(ctx context.Context, req pubsub.SubscribeRequest, han
 		Handler:         adaptHandler(handler),
 		ValueSchemaType: valueSchemaType,
 	}
-	return p.subscribeUtil(ctx, req, handlerConfig)
+
+	p.subscribeUtil(ctx, req, handlerConfig)
```
ItalyPaleAle, line 58: Do you think we should batch this?

At startup, Dapr will likely invoke `Subscribe` multiple times for different topics. This will cause the connection to Kafka to be destroyed and re-created multiple times. Perhaps here we could add a 200ms delay to batch all subscribe calls

yaron2: The code in 1.13 right now doesn't destroy the connection to Kafka, it only recreates the context that the Consume function uses. This PR should adhere to the same behavior

JoshVanL: Yep, this code will only close the consumer group when the pubsub component is `Close()`ed.

I think adding a batcher is a good idea. I've added one for 200ms. It uses a new struct in `kit`, please see the PR here https://github.com/dapr/kit/pull/85

yaron2:
As discussed on today's call, moving to 1.14

yaron2 (approved):

yaron2:
@JoshVanL please cherry pick to 1.13.

PR #3366, opened 2024-02-26
Fix context used in kafka consume

jjcollinge:

# Description

Use the consume context in the backoff retry of consume as the context parameter will be cancelled and not renewed.

Without this change the subscription does not recover once the context has been refreshed and no more messages are consumed.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


ItalyPaleAle (changes requested):
Looks like the Kafka conformance tests are failing, can you PTAL?

yaron2:
> Looks like the Kafka conformance tests are failing, can you PTAL?

They still need fixing from an earlier PR. Some tests need to be removed as they checked for non conformant behavior where messages were still expected to be delivered after the subscription was cancelled

jjcollinge:
@yaron2 I ended up removing the stop subscribers test. It looked to me that the test didn't have much value once we removed the logic to receive messages after the subscribe contexts had been cancelled. Let me know if you want me to reinstate it.

ItalyPaleAle:
> @yaron2 I ended up removing the stop subscribers test. It looked to me that the test didn't have much value once we removed the logic to receive messages after the subscribe contexts had been cancelled. Let me know if you want me to reinstate it.

Wait please don't remove the entire test. There's a reason for that test, we want to make sure that context cancellation works!

berndverst:
This PR should be made against main branch and then cherry picked.

In contrib we do not merge release back to main.

Can you change the base branch @jjcollinge

berndverst:
@yaron2 advised this is not going to be merged for 1.13. So I will update your base branch for that reason also.

Issue #3350, opened 2024-02-10
[PubSub Kafka] Dapr didn't reconcile missing subscription. Manual restart of dapr was required.


olitomlinson:
**runtime 1.12.2** (yes - I'm aware that 1.12.3 contains a subscriber fix, but I'm pretty sure that isn't the issue here)
**EKS**

---

1. On Feb 9th, I noticed that one of our Apps wasn't working as expected, upon further investigation we found that a particular Consumer Group (`ds-application.authz`) was missing 1 (out of 3) Topic subscriptions in Kafka.

`ds-application.authz` Consumer Group was expected to be subscribed to the following Topics

`ds-applicationdatabricks-created` - _subscribed as expected_ 
`ds-applicationrealm-created` - _subscribed as expected_ 
`ds-applicationrealm-user-created` - **missing** 

note : The consuming service only has 1 replica/pod, and there are only 1 partitions in the 3 Topics above, so its a really simple set-up.

---

**Just a note that the  `ds-applicationrealm-user-created` Topic hasn't always been missing from the Consumer Group**

I have logs from the 26th and 29th of January where I can see messages being processed from the `ds-applicationrealm-user-created` Topic as expected

time="2024-01-29T19:11:59.061442928Z" level=debug msg="**Processing Kafka message: ds-applicationrealm-user-created/0/3** [key=]" app_id=authz component="general-purpose-pubsub (pubsub.kafka/v1)" instance=sciplat-svc-authz-deployment-664d59564c-228hh scope=dapr.contrib type=log ver=1.12.2

---

2. My first thought was to check the sidecar logs to ensure that dapr was aware of the Topic (even if was missing in the Consumer Group in Kafka). **It did**. Many logs have all three Topics listed as expected, an example : 

time="2024-02-04T12:21:26.071093619Z" level=debug msg="**client/metadata fetching metadata for [ds-applicationdatabricks-created ds-applicationrealm-created ds-applicationrealm-user-created] from broker** luma-kafka-kafka-bootstrap.dotm-services:9092
" app_id=authz component="general-purpose-pubsub (pubsub.kafka/v1)" instance=sciplat-svc-authz-deployment-664d59564c-228hh scope=dapr.contrib type=log ver=1.12.2

---

3. Given that other Apps in the system which make use of the same PubSub component were operating correctly, there was little more I could do here. No point sending a canary message given that the Consumer Group had lost its subscription.

Therefore, I restarted the deployment at approx **11:47am on the 9th**

[...]

PR #3371, opened 2024-03-01
feat(kafka): add producer config capabilities for connections

sicoyle:
# Description

For the sidecars using the existing kafka client configurations, after a connection has been closed by the client after the default connection idle timeout of 9 minutes, then if that sidecar goes to write and publish a message they will see a `write: broken pipe` err. This is due to the connection being deemed idle, but the sidecar has no idea, and so when it tries to publish, then it is doing so to a closed connection to which we see that error. It is quite confusing for end users, and so these metadata fields allow us to keep a connection alive and establish a refresh for the topic metadata such that the connection may be kept alive indefinitely if desired.

## Issue reference

## Checklist

* [X] Code compiles correctly
* [x] Created/updated tests


sicoyle:
I can update docs after I get initial feedback here pls :)

sicoyle:
Looks like main still has rabbit mq issues based on these test results

common/component/kafka/metadata.go
```diff
@@ -80,16 +89,23 @@ type KafkaMetadata struct {
 	Version                string              `mapstructure:"version"`
 	internalVersion        sarama.KafkaVersion `mapstructure:"-"`
 	internalOidcExtensions map[string]string   `mapstructure:"-"`
+
+	// configs for kafka client
+	clientConnectionRefreshIntervalMin int `mapstructure:"clientConnectionRefreshIntervalMin"`
+	clientConnectionMaxKeepAliveMin    int `mapstructure:"clientConnectionMaxKeepAliveMin"`
```
ItalyPaleAle, line 95: Please make these time.Duration-we try to avoid durations in seconds or minutes and standardize in supporting Go duration formatting (will also need to update `metadata.yaml`). Also, suggesting better naming

```suggestion
	clientConnectionRefreshInterval time.Duration `mapstructure:"clientConnectionRefreshInterval"`
	clientConnectionKeepAliveInterval    time.Duration `mapstructure:"clientConnectionKeepAliveInterval"`
```

common/component/kafka/metadata.go
```diff
@@ -126,8 +142,8 @@ func (k *Kafka) upgradeMetadata(metadata map[string]string) (map[string]string,
 	}
 
 	// if consumeRetryEnabled is not present, use component default value
-	consumeRetryEnabledVal, consumeRetryEnabledPres := metadata[consumeRetryEnabled]
-	if !consumeRetryEnabledPres || consumeRetryEnabledVal == "" {
+	consumeRetryEnabledVal, consumeRetryEnabledOk := metadata[consumeRetryEnabled]
```
ItalyPaleAle, line 145: Please use [`metadata.GetMetadataProperty`](https://pkg.go.dev/github.com/dapr/kit@v0.13.0/metadata#GetMetadataProperty) which is case-insensitive

And, please use `utils.IsTruthy` to check if the value is truthy

common/component/kafka/metadata.go
```diff
@@ -307,5 +323,26 @@ func (k *Kafka) getKafkaMetadata(meta map[string]string) (*KafkaMetadata, error)
...
+	if val, ok := meta[clientConnectionMaxKeepAliveMin]; ok && val != "" {
+		v, err := strconv.Atoi(val)
+		if err != nil {
+			return nil, err
+		}
+		m.clientConnectionMaxKeepAliveMin = v
+	} else {
+		m.clientConnectionMaxKeepAliveMin = defaultClientConnectionMaxKeepAliveMin
+	}
+
```
ItalyPaleAle, line 346: This entire block can be removed if you use `time.Duration` as I suggested for the properties...


Although in that case you may want to add some validation:

- Ensure that clientConnectionRefreshInterval is > 0
- Ensure that clientConnectionKeepAliveInterval is >= 0 (and if it's < 0, make it = 0)

ItalyPaleAle (changes requested):

common/component/kafka/metadata.go
sicoyle: updated, but not sure how IsTruthy is applicable here as it does not make reference to if empty string...

sicoyle:
unrelated failure due to stable components certification test for bindings.rabbitmq

sicoyle:
ready!

common/component/kafka/metadata.go
ItalyPaleAle: If it's empty, `IsTruthy` returns false

common/component/kafka/metadata.go
```diff
@@ -115,10 +115,10 @@ type KafkaMetadata struct {
...
 
 // upgradeMetadata updates metadata properties based on deprecated usage.
-func (k *Kafka) upgradeMetadata(metadata map[string]string) (map[string]string, error) {
-	authTypeVal, authTypePres := metadata[authType]
-	authReqVal, authReqPres := metadata["authRequired"]
-	saslPassVal, saslPassPres := metadata["saslPassword"]
+func (k *Kafka) upgradeMetadata(meta map[string]string) (map[string]string, error) {
+	authTypeVal, authTypePres := meta[authType]
+	authReqVal, authReqPres := meta["authRequired"]
+	saslPassVal, saslPassPres := meta["saslPassword"]
```
ItalyPaleAle, line 121: Please use `metadata.GetMetadataProperties` like I mentioned, it has to be case-insensitive.

common/component/kafka/metadata.go
```diff
@@ -129,25 +129,25 @@ func (k *Kafka) upgradeMetadata(metadata map[string]string) (map[string]string,
...
 					// User has specified saslPassword, so intend for password auth.
-					metadata[authType] = passwordAuthType
+					meta[authType] = passwordAuthType
 				} else {
-					metadata[authType] = mtlsAuthType
+					meta[authType] = mtlsAuthType
 				}
 			} else {
-				metadata[authType] = noAuthType
+				meta[authType] = noAuthType
```
ItalyPaleAle, line 137: Don't update `meta`, because it can cause issues if the key had different casing.

Pass the auth type in a separate variable.

common/component/kafka/metadata.go
```diff
@@ -129,25 +129,25 @@ func (k *Kafka) upgradeMetadata(metadata map[string]string) (map[string]string,
...
 		}
 	}
 
 	// if consumeRetryEnabled is not present, use component default value
-	consumeRetryEnabledVal, consumeRetryEnabledOk := metadata[consumeRetryEnabled]
+	consumeRetryEnabledVal, consumeRetryEnabledOk := metadata.GetMetadataProperty(meta, consumeRetryEnabled)
 	if !consumeRetryEnabledOk || consumeRetryEnabledVal == "" {
-		metadata[consumeRetryEnabled] = strconv.FormatBool(k.DefaultConsumeRetryEnabled)
+		meta[consumeRetryEnabled] = strconv.FormatBool(k.DefaultConsumeRetryEnabled)
 	}
```
ItalyPaleAle, line 148: The better way to do this is to set the default value in the metadata struct (not map) before calling `metadata.ParseMetadata`. It needs to be done in each component.

This way we correctly respect truthy values, and do not cause issues with casing.

common/component/kafka/metadata.go
```diff
@@ -324,24 +324,42 @@ func (k *Kafka) getKafkaMetadata(meta map[string]string) (*KafkaMetadata, error)
...
+			dur = time.Duration(intVal) * time.Minute
+		}
+		m.clientConnectionKeepAliveInterval = dur
+		if m.clientConnectionKeepAliveInterval < 0 {
+			m.clientConnectionKeepAliveInterval = defaultClientConnectionKeepAliveInterval
 		}
-		m.clientConnectionMaxKeepAliveMin = v
 	} else {
-		m.clientConnectionMaxKeepAliveMin = defaultClientConnectionMaxKeepAliveMin
+		m.clientConnectionKeepAliveInterval = defaultClientConnectionKeepAliveInterval
```
ItalyPaleAle, line 362: All this code is not necessary now that you have set the properties to be of type `time.Duration`. You can just remove all these lines...

ItalyPaleAle (changes requested):

common/component/kafka/metadata.go
sicoyle: this is all preexisting code... but i can update, sure

common/component/kafka/metadata.go
sicoyle: this too is preexisting. I just updated to account for metadata not clashing with the imported pkg called metadata that we're now consuming

ItalyPaleAle: Could you please make this change in here since we're improving the metadata?

common/component/kafka/metadata.go
ItalyPaleAle: Clarified in discord

common/component/kafka/metadata.go
ItalyPaleAle: You need to set the default values in the struct before calling `decodeMetadata`, here: https://github.dev/sicoyle/components-contrib/blob/b572ee936a38cc32ca609fe414247b3c18fd167e/common/component/kafka/metadata.go#L155-L161 Then `decodeMetadata` will overwrite the defaults if there's a value explicitly set in the metadata

sicoyle: setting the defaults where you've stated, and then removing these lines of code also fails. While it does take the defaults, if I set a value such as for my test:
```

	t.Run("setting producer values explicitly", func(t *testing.T) {
		k := getKafka()
		m := getCompleteMetadata()
		m[clientConnectionRefreshInterval] = "3m0s"
		m[clientConnectionKeepAliveInterval] = "4m0s"

		meta, err := k.getKafkaMetadata(m)
		require.NoError(t, err)
		require.Equal(t, 3*time.Minute, meta.clientConnectionRefreshInterval)
		require.Equal(t, 4*time.Minute, meta.clientConnectionKeepAliveInterval)
	})
```

Then relying upon the decodeMetadata fails. This to me means there is a bug in the decodeMetadata when it comes to parsing time durations. However, this too falls outside of the scope of the intended work for this PR. I would like that follow Github best practices and fall outside of the realm of this PR to keep the scope in check with what the PR is intended to be for.

sicoyle:
this is now a dependency: https://github.com/dapr/kit/pull/88

common/component/kafka/metadata.go
```diff
@@ -80,16 +89,23 @@ type KafkaMetadata struct {
 	Version                string              `mapstructure:"version"`
 	internalVersion        sarama.KafkaVersion `mapstructure:"-"`
 	internalOidcExtensions map[string]string   `mapstructure:"-"`
+
+	// configs for kafka client
+	clientConnectionRefreshInterval   time.Duration `mapstructure:"clientConnectionRefreshInterval"`
```
famarting, line 94: should this fields be public? `ClientConnectionRefreshInterval` and `ClientConnectionKeepAliveInterval`

common/component/kafka/kafka.go
```diff
@@ -143,6 +143,9 @@ func (k *Kafka) Init(ctx context.Context, metadata map[string]string) error {
 	config.Consumer.Fetch.Default = meta.consumerFetchDefault
 	config.ChannelBufferSize = meta.channelBufferSize
 
+	config.Net.KeepAlive = meta.clientConnectionKeepAliveInterval
+	config.Metadata.RefreshFrequency = meta.clientConnectionRefreshInterval
```
famarting, line 147: the config `config.Metadata.RefreshFrequency` has nothing to do with the tcp connections to kafka, it is a sarama client config to tweak the interval at which the topics metadata is reloaded from the kafka brokers

sicoyle:
ready

common/component/kafka/kafka.go
```diff
@@ -143,8 +143,8 @@ func (k *Kafka) Init(ctx context.Context, metadata map[string]string) error {
 	config.Consumer.Fetch.Default = meta.consumerFetchDefault
 	config.ChannelBufferSize = meta.channelBufferSize
 
-	config.Net.KeepAlive = meta.clientConnectionKeepAliveInterval
-	config.Metadata.RefreshFrequency = meta.clientConnectionRefreshInterval
+	config.Net.KeepAlive = meta.ClientConnectionKeepAliveInterval
+	config.Metadata.RefreshFrequency = meta.ClientConnectionRefreshInterval
```
jjcollinge: Should this be named something like `meta.ClientMetadataRefreshFrequency` instead to be clear on its purpose?

sicoyle: I renamed to ClientConnectionTopicMetadataRefreshFrequency

olitomlinson:
I've had a couple of issues with [`broken pipe`](https://github.com/dapr/components-contrib/issues/3350#broken-pipe) so this fix will be well received :)

sicoyle:
ready

cicoyle (approved):

famarting (approved):

sicoyle:
ping for review please

sicoyle:
ready for review please

pubsub/kafka/metadata.yaml
```diff
@@ -266,6 +266,18 @@ metadata:
...
+      description: |
+        The interval for the client connection's topic metadata to be refreshed with the broker in minutes.
+      example: '4m'
+      default: '9m'
+    - name: clientConnectionKeepAliveInterval
+      type: number
+      description: |
+        The max amount of time for the client connection to be kept alive with the broker, in minutes, before closing the connection. Zero defaults to keeping alive indefinitely.
+      example: '4m'
+      default: '0m'
```
JoshVanL, line 280: Same

JoshVanL (changes requested):

yaron2:
@sicoyle please see latest comments from @JoshVanL. We should be good to merge once addressed

sicoyle:
addressed feedback. Thanks @JoshVanL

bindings/kafka/metadata.yaml
```diff
@@ -222,6 +222,18 @@ metadata:
     example: '"group1"'
     binding:
       input: true
+  - name: clientConnectionTopicMetadataRefreshInterval
+    type: number
```
ItalyPaleAle, line 226:
```suggestion
    type: duration
```

bindings/kafka/metadata.yaml
```diff
@@ -222,6 +222,18 @@ metadata:
...
+    description: |
+      The interval for the client connection's topic metadata to be refreshed with the broker as a Go duration.
+    example: '4m'
+    default: '9m'
+  - name: clientConnectionKeepAliveInterval
+    type: number
```
ItalyPaleAle, line 232:
```suggestion
    type: duration
```

ItalyPaleAle (changes requested):

sicoyle:
ready @ItalyPaleAle Thanks for the extra feedback!

ItalyPaleAle (approved):

PR #3386, opened 2024-03-29
Replace unsupported zeebe form key header

### LINKED ISSUE #3385: Zeebe worker binding doesn't handle user tasks
## Expected Behavior

When I configure "io.camunda.zeebe:userTask" as jobType in the binding configuration it should forward user tasks to my microservice. 

## Actual Behavior

I receive an error in the logs

```
time="2024-03-29T07:17:25.7572877+01:00" level=error msg="Failed to complete job `2251799813701385` reason: error invoking app: Post \"http://127.0.0.1:8001/user-task-binding\": net/http: invalid header field name \"io.camunda.zeebe:formkey\"" app_id=controlpanel component="user-task-binding (bindings.zeebe.jobworker/v1)" instance=[Redacted] scope=dapr.contrib type=log ver=1.13.1
```

* Deploy Camunda 8 locally.
* Deploy the attached BPMN file and Form file. (both are in the zip)
* Use the binding configuration in the attached YAML file (it's in the zip)
* Add a dummy API endpoint that listens on `/bindings/user-task`

wmeints:
When handling user tasks, Zeebe sends a custom io.camunda.zeebe:formKey header that's not supported over HTTP. This fix replaces the custom header with a X-Zeebe-Form-Key header.

# Description

Add code to detect the `io.camunda.zeebe:formKey` header and replace it with `X-Zeebe-Form-Key` header instead so that
the callback to the microservice doesn't fail.

## Issue reference

Closes #3385

## Checklist

* [x] Code compiles correctly

### DIFF
--- a/bindings/zeebe/jobworker/jobworker.go
+++ b/bindings/zeebe/jobworker/jobworker.go
@@ -198,6 +198,13 @@ func (h *jobHandler) handleJob(client worker.JobClient, job entities.Job) {
 		return
 	}
 
+	// Make sure to replace the formKey header with a custom header.
+	// The HTTP Client doesn't like the custom header.
+	if headers["io.camunda.zeebe:formKey"] != "" {
+		headers["X-Zeebe-Form-Key"] = headers["io.camunda.zeebe:formKey"]
+		delete(headers, "io.camunda.zeebe:formKey")
+	}
+
 	headers["X-Zeebe-Job-Key"] = strconv.FormatInt(job.Key, 10)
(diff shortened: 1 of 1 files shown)


wmeints:
After reviewing the changes some more and checking out the docs in Camunda I believe that this change is not quite what I wanted. As a tip to others: Don't try to subscribe to user tasks from your microservices. You'll need to integrate the Tasks API of Camunda which is not at all available in Dapr right now. Also, the Camunda platform has a ready-made solution for user tasks which is more suited.

Issue #3387, opened 2024-03-29
New kafka-pubsub Avro serializer doesn't handle NULL messages


passuied:
## Expected Behavior

NULL message values are perfectly valid messages (aka tombstone messages). The latest change adding support for Avro serialization/deserialization didn't account for null values which shouldn't tried to be serialized/deserialized...

## Actual Behavior

Currently, when publishing/consuming a message with NULL value, it will fail with `value is too short`.

## Steps to Reproduce the Problem
- Topic with messages in Avro format
- Subscription with metadata set to `valueSchemaType=Avro`
- Messages in topic with `NULL` message values

## Release Note

RELEASE NOTE: **FIX** Bug in kafka-pubsub component when `"valueSchemaType": "Avro"` and messages consumed have NULL value or when publishing messages with NULL value .

passuied:
Bug fixed and merged to 1.13

PR #3388, opened 2024-03-29
Fixing bug with new Avro conversion when message values are NULL

passuied:
# Description

Fixing bug with new Avro conversion when message values are NULL

## Issue reference

#3387

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/common/component/kafka/kafka.go
+++ b/common/component/kafka/kafka.go
@@ -266,7 +266,7 @@ func getSchemaSubject(topic string) string {
 func (k *Kafka) DeserializeValue(message *sarama.ConsumerMessage, config SubscriptionHandlerConfig) ([]byte, error) {
 	// Null Data is valid and a tombstone record. It shouldn't be serialized
 	if message.Value == nil {
-		return json.Marshal(nil)
+		return []byte("null"), nil;
 	}
 
 	switch config.ValueSchemaType {
--- a/common/component/kafka/kafka.go
+++ b/common/component/kafka/kafka.go
@@ -16,6 +16,7 @@ package kafka
 import (
 	"context"
 	"encoding/binary"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
@@ -265,7 +266,7 @@ func getSchemaSubject(topic string) string {
 func (k *Kafka) DeserializeValue(message *sarama.ConsumerMessage, config SubscriptionHandlerConfig) ([]byte, error) {
 	// Null Data is valid and a tombstone record. It shouldn't be serialized
 	if message.Value == nil {
-		return message.Value, nil
+		return json.Marshal(nil)
 	}
 
 	switch config.ValueSchemaType {
--- a/common/component/kafka/kafka.go
+++ b/common/component/kafka/kafka.go
@@ -263,7 +263,6 @@ func getSchemaSubject(topic string) string {
 }
 
 func (k *Kafka) DeserializeValue(message *sarama.ConsumerMessage, config SubscriptionHandlerConfig) ([]byte, error) {
-
 	// Null Data is valid and a tombstone record. It shouldn't be serialized
 	if message.Value == nil {
 		return message.Value, nil
--- a/common/component/kafka/kafka.go
+++ b/common/component/kafka/kafka.go
@@ -263,6 +263,12 @@ func getSchemaSubject(topic string) string {
 }
 
 func (k *Kafka) DeserializeValue(message *sarama.ConsumerMessage, config SubscriptionHandlerConfig) ([]byte, error) {
+
+	// Null Data is valid and a tombstone record. It shouldn't be serialized
+	if message.Value == nil {
+		return message.Value, nil
+	}
+
 	switch config.ValueSchemaType {
 	case Avro:
 		srClient, err := k.getSchemaRegistyClient()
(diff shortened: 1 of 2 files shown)


passuied:
@yaron2 These certification tests are really brittle. I doubt my changes have caused rabbitmq and mqtt3 tests to fail... Any path forward other than retrying these tests over an over? Should they be skipped until made more resilient?

passuied:
The current issue in cert tests seems realted to a certificate issue: 
```
failed to verify certificate: x509: certificate has expired or is not yet valid
```
I remember this was an issue previously due to some certificates being hardcoded in tests and expiring...

common/component/kafka/kafka.go
```diff
@@ -263,6 +264,11 @@ func getSchemaSubject(topic string) string {
 }
 
 func (k *Kafka) DeserializeValue(message *sarama.ConsumerMessage, config SubscriptionHandlerConfig) ([]byte, error) {
+	// Null Data is valid and a tombstone record. It shouldn't be serialized
+	if message.Value == nil {
+		return json.Marshal(nil)
```
ItalyPaleAle, line 269: Since this is a constant value...

```suggestion
		return []byte("null"), nil;
```

common/component/kafka/kafka.go
```diff
@@ -349,6 +355,11 @@ func (k *Kafka) getSchemaRegistyClient() (srclient.ISchemaRegistryClient, error)
 }
 
 func (k *Kafka) SerializeValue(topic string, data []byte, metadata map[string]string) ([]byte, error) {
+	// Null Data is valid and a tombstone record. It shouldn't be serialized
+	if data == nil {
+		return data, nil
```
ItalyPaleAle, line 360:
```suggestion
		return nil, nil
```

common/component/kafka/kafka_test.go
```diff
@@ -79,10 +83,23 @@ func TestDeserializeValue(t *testing.T) {
...
 		}
-		act, _ := k.DeserializeValue(&msg, handlerConfig)
+		act, err := k.DeserializeValue(&msg, handlerConfig)
 		var actMap map[string]any
 		json.Unmarshal(act, &actMap)
 		require.Equal(t, testValue1, actMap)
+		require.NoError(t, err)
+	})
+
+	t.Run("Data null, return as Json Null", func(t *testing.T) {
```
ItalyPaleAle, line 93:
```suggestion
	t.Run("Data null, return as JSON null", func(t *testing.T) {
```

ItalyPaleAle (changes requested):
Small changes

common/component/kafka/kafka.go
passuied: I don't think these 2 are the same.. I want to return NULL not the string "null"...

passuied: > json.Marshal(nil) produces a proper JSON null value, while []byte("null") is just a byte slice containing the string “null”

passuied:
> The current issue in cert tests seems realted to a certificate issue:
> 
> ```
> failed to verify certificate: x509: certificate has expired or is not yet valid
> ```
> 
> I remember this was an issue previously due to some certificates being hardcoded in tests and expiring...
> 
> Yeah these hard-coded certs have expired... https://github.com/dapr/components-contrib/blob/main/tests/certification/bindings/rabbitmq/mtls_sasl_external/components/mtls_external/rabbitmq-mtls.yaml

@ItalyPaleAle how should we deal with these certification errors due to expired. They've been failing since Feb... In the past, you figured out a way to generate certs as part of the test (though I cannot find the code anymore)...

common/component/kafka/kafka.go
ItalyPaleAle: https://go.dev/play/p/Tj_PlstjW7K
Note that `uint8 == byte` (they are aliases)

ItalyPaleAle:
> @ItalyPaleAle how should we deal with these certification errors due to certs having expired. They've been failing since Feb... In the past, you figured out a way to generate certs as part of the test (though I cannot find the code anymore)...

That's for RabbitMQ not Kafka, right? If you want to contribute a PR to fix that, it'd be nice :)

I think accepting hardcoded certs was a mistake. We should have generated the certs on-the-fly.

common/component/kafka/kafka.go
passuied: You're right. Validated it works. That will teach me to teach an AI instead of a human ;)

passuied:
> > @ItalyPaleAle how should we deal with these certification errors due to certs having expired. They've been failing since Feb... In the past, you figured out a way to generate certs as part of the test (though I cannot find the code anymore)...
> 
> That's for RabbitMQ not Kafka, right? If you want to contribute a PR to fix that, it'd be nice :)
> 
> I think accepting hardcoded certs was a mistake. We should have generated the certs on-the-fly.

Don't really have the time right this minute but would like my PR to go through quickly as we're eager to make the change. I can try to contribute later but I'm struggling to run these tests locally...

ItalyPaleAle:
I am not concerned with the RabbitMQ tests failing in this PR. If Kafka tests pass, we can merge

ItalyPaleAle (approved):

ItalyPaleAle (approved):

Issue #3395, opened 2024-04-11
Kafka-pubsub - Ability to override the consumer group within the subscription


passuied:
## Describe the feature
The Kafka-pubsub component allows specifying a consumer group within the component metadata. However since this is at the component level, if an application intended to consume the same topic for 2 different use cases, it would require to register a separate component in order to use a different consumer group.

The consumer group drives the offset tracking in kafka so 2 separate endpoints using the same consumer group would step on each other's toes.

This is a bit of an inconvenience. It would be simple to optionally override the consumer group by introducing a consumer group override in the subscription metadata...

### Proposed usage
```python
@dapr_app.subscribe(pubsub="kafka-pubsub", topic="my-topic", metadata={"consumerGroup": "my-consumer-group-override"})
async def on_my_topic_received(message: dict):   
  print(event_data) 
```

### More details about use case
We're trying to avoid double writes and instead relying on eventual consistency. In the current usecase, a request should drive 2 actions:
- Saving some state in a State store. In our case DynamoDB
- Publish an event that can be consumed by other systems

### Alternatives:
- A simple alternative is to register 2 separate components (e.g. `kafka-pubsub` and `kafka-pubsub2`) and have 2 separate consumer groups defined there. This is probably what we will go through short term but seems a bit overkill to have to register 2 components for this use case... Feels like being able to set the consumer group from the subscription is more similar to other kafka consumer usages.
- Another alternative is to leverage the new outbox use case but it's not perfect either because:
  - the event format and object saved have to be identical, which is not our case
  - in some usecases, it might not be a great usecase to use the state store component when the data being saved is part of a larger set of changes that should be saved atomically and already leveraging an ORM.

 
## Release Note

RELEASE NOTE: **ADD** ability to override the consumer group of a kafka-pubsub consumer within the subscription metadata.

passuied:
After reviewing the code, the change would involve maintaining a map of consumer groups within the existing kafka instance. Currently it's a 1-1 relationship...

passuied:
As mentioned in PR, dapr PubSub was designed to support distinct topics. Until that changes, it doesn't make sense to force it at the component level.

PR #3397, opened 2024-04-11
kafka-pubsub: Add consumer group overrides in subscription metadata

passuied:
# Description

Add ability to override the consumer group using the subscription metadata `consumerGroup`. Avoids having to define 2 separate components  when having to consume the same topic in a single application.
Defining a different consumer group will register and additional consumer group within the component.
Some refactoring involved to support many consumer groups.

## Issue reference
#3395 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#4113


passuied:
Just a note that while tests currently pass, it doesn't seem to work as expected when I test it end to end... While the additional consumer group shows up in the log, only the new consumer group is processing messages... Investigating...

passuied:
Ok I think I understand now. The issue at the dapr level, as it doesn't allow having the same topic multiple times under a given component :(
https://github.com/dapr/dapr/blob/master/pkg/runtime/processor/pubsub/subscribe.go#L116-L127

Not sure it's worth pursuing given that generic limitation... One alternative would be to pass the consumer group within the topic string somehow (e.g. `my-topic-1[cg2]`) but that looks messy...
Thoughts/Ideas?

passuied:
Given recent discovey, I feel this is not worth it... Dapr Pubsub  was designed to support distinct topics. Until this changes, it doesn't make sense to try to shoe horn it at the

PR #3411, opened 2024-05-03
Fix CosmosDB for latest API version

berndverst:
# Description

CosmosDB changed how they indicate when an entity is not found.


state/azure/cosmosdb/cosmosdb.go
```diff
@@ -685,14 +683,16 @@ func populatePartitionMetadata(key string, requestMetadata map[string]string) st
 }
 
 func isNotFoundError(err error) bool {
+	fmt.Printf("Check isNotFoundError: %v\n", err)
```
yaron2, line 686: Why fmt.Println here?

yaron2:
@berndverst was this merged by mistake?

state/azure/cosmosdb/cosmosdb.go
```diff
@@ -685,14 +683,16 @@ func populatePartitionMetadata(key string, requestMetadata map[string]string) st
...
 	if err == nil {
 		return false
 	}
 
 	if requestError, ok := err.(*azcore.ResponseError); ok {
-		if requestError.ErrorCode == statusNotFound {
+		if requestError.StatusCode == 404 {
```
yaron2, line 692: Linter calls this line out as well to replace 404 with http.StatusNotFound.

It's better to wait for a review and approval..

ItalyPaleAle:
@berndverst this LGTM but please make a follow-up PR to fix the linter issues

PR #3413, opened 2024-05-06
fix: metadata capitalization

sicoyle:
# Description

We currently have consumerGroup, consumerID, and consumerId metadata fields on components. I think we should remove the consumerId metadata field as the capitalization difference is confusing, and there are very few components with this capitalization, so I'm wondering if this is an oversight and is fine to be corrected. I know that we do not strive for consistency in naming just for name sake, but I do think we should strive for capitalization consistency on fields with the same name across the board.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


pubsub/azure/eventhubs/metadata.yaml
```diff
@@ -97,7 +97,7 @@ metadata:
     description: |
       Storage container name.
     example: '"myeventhubstoragecontainer"'
-  - name: consumerId
+  - name: consumerID
```
ItalyPaleAle, line 100: All our fields are case-insensitive. If you want to update the `metadata.yaml` and the docs, that's all that needs to change :)

ItalyPaleAle (changes requested):

3346.patch
berndverst, line 1: Remove this patch file... it should not be part of this PR.

pubsub/azure/eventhubs/metadata.yaml
berndverst: And additionally, keep in mind that `metadata.yaml` is not what controls the accepted values - this is just documentation.

Our Kit/Decode Metadata function which is based on mitchellh/mapstructure uses the WeakDecode function which accepts any casing. There are several other convenience features, which is why all metadata should always be specified as struct tags and then decoded using this special function.

Also FYI, `consumerID` is a special metadata which actually is **NOT** defined in `components-contrib`. This is hardcoded in Dapr runtime, so you might need to verify in that repo what casing is accepted. Unlike in contrib, they don't use our metadata handling code in runtime.

berndverst (changes requested):

pubsub/azure/eventhubs/metadata.yaml
sicoyle: ahh, okay. Docs do reflect the changes I'm making here, so that works out great and will require no doc updates: https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-azure-eventhubs/

berndverst (approved):

Issue #3428, opened 2024-05-22
Dapr pubsub service bus topics: Add support for including `ApplicationProperties` in subscriber


vbuonagura:
Hello,
we are trying to integrate Dapr within our microservices architectures but we are experiencing some issues in accessing custom properties for messaged published through Service Bus Topics.

We are not using cloudEvents and to make events traceable, we will leverage ElasticAPM with the default Azure Service Bus capabilities based on Diagnostic-Id field. See more [here](https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-end-to-end-tracing?tabs=net-standard-sdk-2).

We are publishing the message from a .NET 8 API in the following way:
`
public async Task<IActionResult> CreateNotification([FromBody] Notification notification)
{
    if (notification is not null)
    {
        ITransaction transaction = Agent.Tracer.CurrentTransaction;

        var asyncResult = await transaction.CaptureSpan("SendEvent", ApiConstants.TypeMessaging, async(s) =>
        {
            //metadata needed to remove the cloudevent envelop
            var metadata = new Dictionary<string, string>
            {
                { "rawPayload", "true" },
                { "Diagnostic-Id", "00-"+ Agent.Tracer.CurrentTransaction.TraceId + "-" + Agent.Tracer.CurrentSpan.Id + "-" +"01"}
            };

            await _daprClient.PublishEventAsync("pubsub", "notification-topic", notification, metadata);

            return true;
        });

        return Ok("Notification Created");
    }

    return BadRequest();
}`

The rawPayload and Diagnostic-Id metadata are correctly set into the message custom properties, but when we receive the data in the following way, we cannot see those metadata into the http request header
`Dapr.Topic("pubsub", "notification-topic")]
[HttpPost("notificationreceived")]
public async Task<IActionResult> NotificationReceived([FromBody] Notification notification)
{ 
    foreach (var header in Request.Headers)
    {
        _logger.LogInformation(header.Key + " = " + header.Value);
    }

    if (notification is not null)
    {
        await _notificationsService.CreateAsync(notification);

        return Ok("Notification Received");
    }

    return BadRequest();
}`

the only metadata available are:

- deliverycount
- enqueuedtimeutc
- lockeduntilutc
- locktoken
- messageid
- sequencenumber

Is there a way to get also custom properties with metadata?

berndverst:
[...]

PR #3436, opened 2024-06-04
ASB: Add support for ApplicationProperties in subscriptions

berndverst:
# Description

Service Bus PubSub (and Bindings) Components: Add `ApplicationProperties` to metadata payload on subscription (in the same way these are set from metadata when publishing through Dapr)

## Issue reference

#3428

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
~* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_~

### DIFF
--- a/common/component/azure/servicebus/message_pubsub.go
+++ b/common/component/azure/servicebus/message_pubsub.go
@@ -15,7 +15,7 @@ package servicebus
 
 import (
 	"encoding/base64"
-	"fmt"
+	"github.com/spf13/cast"
 	"net/http"
 	"strconv"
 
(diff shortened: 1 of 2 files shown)


common/component/azure/servicebus/message_pubsub.go
```diff
@@ -58,6 +59,12 @@ func addMessageAttributesToMetadata(metadata map[string]string, asbMsg *azservic
 		metadata = map[string]string{}
 	}
 
+	for key, val := range asbMsg.ApplicationProperties {
+		metadata["metadata."+key] = fmt.Sprintf("%v", val)
```
ItalyPaleAle, line 63: Can we avoid `fmt.Sprintf` in a hot path? Worst case, how about using [`cast.ToString()`](https://pkg.go.dev/github.com/spf13/cast#ToString) which at least avoids some allocations?

ItalyPaleAle (approved):

common/component/azure/servicebus/message_pubsub.go
berndverst: Done.

berndverst:
Force merging as maintainer given this was approved and suggestion was applied.

Issue #3120, opened 2023-09-08
RabbitMQ binding enhancements required or am I missing something in configuration?


jwyackel:
## Ask your question here
Hi,

I just started getting a RabbitMQ binding working, but noticed that there is no way to pass the following parameters:
- x-queue-type
- x-single-active-consumer

If I let the Dapr binding declare the queue, it gets created as a classic queue.  
If I manually declare the queue using the management portal, I get an exception when the binding is initialized.

` msg="process component rabbitmq-input error: [INIT_COMPONENT_FAILURE]: initialization error occurred for rabbitmq-input (bindings.rabbitmq/v1): Exception (406) Reason: \"PRECONDITION_FAILED - inequivalent arg 'x-single-active-consumer' for queue 'daprtest' in vhost '/': received none but current is the value 'true' of type 'bool'\"" app_id=RabbitMQConsumerDapr`

As well, does the RabbitMQ binding handle connecting to a cluster of nodes?  I can only pass in a single host, I have three nodes, is there any type of support for setting up multiple bindings (hence the request to be able to use x-single-active-consumer), or at least handle multiple hostnames for the initial connection process.

I am just starting out with Dapr and am unsure if I am doing something wrong or what I am trying to do is just not supported at this time?

Jeremy

liuzhch-Orka:
Just not supported.
The SAC parameter `x-single-active-consumer`, among many others, has not yet been implemented.
https://github.com/dapr/components-contrib/blob/2cb3c1cce1de407913615aa0296cba6db34205f7/pubsub/rabbitmq/rabbitmq.go#L50-L62

DropSnorz:
As quorum queues are supported, would you accept contributions for RabbitMQ `x-single-active-consumer` support ?

PR #3437, opened 2024-06-06
Add RabbitMQ single active consumer argument

DropSnorz:
# Description

Added a new RabbitMQ subscription argument to enable [Single Active Consumer](https://www.rabbitmq.com/docs/consumers#single-active-consumer) (`x-single-active-consumer`)

Usage with declarative subscription
```yaml
apiVersion: dapr.io/v2alpha1
kind: Subscription
metadata:
  name: orders
spec:
  topic: orders
  routes:
    default: /orders
  pubsubname: orderpubsub
  metadata:
    queueName: "foo-queue-quorum"
    queueType: quorum
    singleActiveConsumer: "true"
scopes:
- order-processor-sdk

```

## Issue reference

Please reference the issue this PR will close: #3120

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/pubsub/rabbitmq/rabbitmq.go
+++ b/pubsub/rabbitmq/rabbitmq.go
@@ -31,6 +31,7 @@ import (
 	"github.com/dapr/components-contrib/metadata"
 	"github.com/dapr/components-contrib/pubsub"
 	"github.com/dapr/kit/logger"
+	"github.com/dapr/kit/utils"
 )
 
 const (
@@ -435,15 +436,8 @@ func (r *rabbitMQ) prepareSubscription(channel rabbitMQChannelBroker, req pubsub
 	}
 
 	// Applying x-single-active-consumer if defined at subscription level
-	if val := req.Metadata[reqMetadataSingleActiveConsumerKey]; val != "" {
-		parsedVal, pErr := strconv.ParseBool(val)
-		if pErr != nil {
-			r.logger.Errorf("%s invalid boolean value for %s on subscription metadata for topic/queue `%s/%s`: %s", logMessagePrefix, reqMetadataSingleActiveConsumerKey, req.Topic, queueName, pErr)
-			return nil, pErr
-		}
-		if parsedVal {
-			args[argSingleActiveConsumer] = parsedVal
-		}
+	if val := req.Metadata[reqMetadataSingleActiveConsumerKey]; utils.IsTruthy(val) {
+		args[argSingleActiveConsumer] = true
 	}
 
 	// Applying x-max-length-bytes if defined at subscription level
--- a/pubsub/rabbitmq/rabbitmq.go
+++ b/pubsub/rabbitmq/rabbitmq.go
@@ -48,17 +48,19 @@ const (
 	defaultHeartbeat        = 10 * time.Second
 	defaultLocale           = "en_US"
 
-	argQueueMode              = "x-queue-mode"
-	argMaxLength              = "x-max-length"
-	argMaxLengthBytes         = "x-max-length-bytes"
-	argDeadLetterExchange     = "x-dead-letter-exchange"
-	argMaxPriority            = "x-max-priority"
-	propertyClientName        = "connection_name"
-	queueModeLazy             = "lazy"
-	reqMetadataRoutingKey     = "routingKey"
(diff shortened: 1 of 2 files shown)


pubsub/rabbitmq/rabbitmq_test.go
```diff
@@ -129,6 +129,11 @@ func TestPublishAndSubscribeWithPriorityQueue(t *testing.T) {
 	<-processed
 	assert.Equal(t, 4, messageCount)
 	assert.Equal(t, "foo bar", lastMessage)
+
+	// subscribe using single active consumer
+	err = pubsubRabbitMQ.Subscribe(context.Background(), pubsub.SubscribeRequest{Topic: topic, Metadata: map[string]string{reqMetadataSingleActiveConsumerKey: "true"}}, handler)
```
berndverst, line 134: You need to actually publish a message after this, then assert that the message was received and the overall messageCount has been increased.

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -432,6 +434,18 @@ func (r *rabbitMQ) prepareSubscription(channel rabbitMQChannelBroker, req pubsub
 		args[amqp.QueueTypeArg] = amqp.QueueTypeClassic
 	}
 
+	// Applying x-single-active-consumer if defined at subscription level
+	if val := req.Metadata[reqMetadataSingleActiveConsumerKey]; val != "" {
+		parsedVal, pErr := strconv.ParseBool(val)
```
berndverst, line 439: Do not use `strconv.ParseBool` instead use our preferred method from package `dapr/kit/utils` called `IsTruthy` https://github.com/dapr/kit/blob/106329e5839f70f3234b9f49b7dabff0247c0a90/utils/strings.go#L23

That will allow a lot more values to be used to express the "true" / on state.

berndverst (changes requested):

berndverst:
A few small changes / additions, then we can merge this.

pubsub/rabbitmq/rabbitmq.go
DropSnorz: Thanks for highlighting this, the code is far more readable this way 👍

pubsub/rabbitmq/rabbitmq_test.go
DropSnorz: Test function updated.

yaron2 (approved):

yaron2 (approved):

yaron2:
@DropSnorz Thanks for this PR! Please open a docs issue in dapr/docs to update the RabbitMQ pub/sub spec with this setting.

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -48,17 +49,19 @@ const (
...
-	reqMetadataRoutingKey     = "routingKey"
-	reqMetadataQueueTypeKey   = "queueType" // at the moment, only supporting classic and quorum queues
-	reqMetadataMaxLenKey      = "maxLen"
-	reqMetadataMaxLenBytesKey = "maxLenBytes"
+	argQueueMode                       = "x-queue-mode"
+	argMaxLength                       = "x-max-length"
+	argMaxLengthBytes                  = "x-max-length-bytes"
+	argDeadLetterExchange              = "x-dead-letter-exchange"
+	argMaxPriority                     = "x-max-priority"
+	argSingleActiveConsumer            = "x-single-active-consumer"
```
ItalyPaleAle, line 57: Please open another PR to update https://github.com/dapr/components-contrib/blob/main/pubsub/rabbitmq/metadata.yaml
CC: @yaron2

DropSnorz: @ItalyPaleAle It's not clear to me what should be updated in the `metadata.yml` file. This file seems to describe metadata for the rabbitmq pubsub **Component** while the `singleActiveConsumer` parameter is set at the **Subscription** level (like `queueName`, `routingKey`, ...)

ItalyPaleAle: I see, I misunderstood this PR, thanks for clarifying.

marcduiker:
@holopin-bot @DropSnorz Thank you!

PR #3439, opened 2024-06-11
fix: set required to true for redisPassword in state>redis>metadata.yaml as per docs

jake-engelberg:


berndverst:
Sorry, the docs are incorrect. This field is not required. This metadata yaml file is the source of truth - the docs may be outdated.

Please open a docs PR instead.

bibryam:
@berndverst any idea why redis pubsub component has it as mandatory, but redis state component has it as optional? Can it be a discrepancy in the metadata files?

PR #3459, opened 2024-06-22
change topicArns to map safe for concurrent access

nenikola:
# Description

Changed native map to xsync.MapOf which is safe for concurrent access

## Issue reference

Please reference the issue this PR will close: #3430 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/pubsub/aws/snssqs/snssqs.go
+++ b/pubsub/aws/snssqs/snssqs.go
@@ -30,7 +30,6 @@ import (
 	"github.com/aws/aws-sdk-go/service/sns"
 	"github.com/aws/aws-sdk-go/service/sqs"
 	"github.com/aws/aws-sdk-go/service/sts"
-	"github.com/puzpuzpuz/xsync/v3"
 
 	"github.com/dapr/kit/retry"
 
@@ -45,7 +44,7 @@ import (
 type snsSqs struct {
 	topicsLocker TopicsLocker
 	// key is the sanitized topic name
-	topicArns *xsync.MapOf[string, string]
+	topicArns sync.Map
 	// key is the topic name, value holds the ARN of the queue and its url.
 	queues map[string]*sqsQueueInfo
 	// key is a composite key of queue ARN and topic ARN mapping to subscription ARN.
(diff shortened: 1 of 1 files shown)


yaron2:
Thanks for this PR. Can you please replace the usage of the xsync package with Go's sync.Map? It has the same methods. https://pkg.go.dev/golang.org/x/sync/syncmap

nenikola:
> Thanks for this PR. Can you please replace the usage of the xsync package with Go's sync.Map? It has the same methods. https://pkg.go.dev/golang.org/x/sync/syncmap

@yaron2 sure, I opted for xsync just because I saw it being used in [topicsLocker](https://github.com/dapr/components-contrib/blob/7fd5524c5842293c0e6d1d2427b56dbd7dddc100/pubsub/aws/snssqs/topics_locker.go#L20-L22).

Also, I saw `subscriptions` and `queues` are also of type regular map and seems like both could potentially face similar concurrent write issue. Would it make sense to change those to `sync.Map` too?

ItalyPaleAle (changes requested):
In this file we were doing synchronization using a mutex. For example, calls to `getOrCreateTopic` are wrapped around a lock:

https://github.com/dapr/components-contrib/blob/95a0a1339e2791f5d8cbdc071ba85062aa5be531/pubsub/aws/snssqs/snssqs.go#L762-L763

Rather than changing `topicArns`, it may make more sense to just ensure that wherever the access to the map was happening that wasn't wrapped in a log (and was causing the panic), uses the lock instead

nenikola:
@ItalyPaleAle so I did something similar to what is done in the `Subscribe`  method by locking the topic when invoking `getOrCreateTopic` but would it make more sense for the method `getOrCreateTopic` to do the locking itself instead of its caller doing so?

The way it is now, caller of `getOrCreateTopic` needs to know that `getOrCreateTopic` method does unsafe map write, which shouldn't be the caller's concern per my understanding.

ItalyPaleAle (dismissed):

nenikola:
@ItalyPaleAle sounds good, let me add comment and then merge the changes. Thanks!

ItalyPaleAle (approved):
Thank you for this contribution and for fixing the issue!

marcduiker:
@holopin-bot @nenikola Thank you Nikola!

PR #3466, opened 2024-06-27
bindings/http: Use Go Default Transport as base Transport

JoshVanL:
Updates the HTTP client's transport to be based from the Go's default transport to inherit the default settings.

See https://github.com/dapr/dapr/pull/7833


yaron2:
Linter

yaron2 (approved):

marcduiker:
@holopin-bot @JoshVanL Thanks Josh!

PR #3470, opened 2024-06-29
Adds EntraID auth support to all Redis Components

berndverst:
# Description

Implements EntraID / AAD support for Azure Cache for Redis across all components

For a component yaml file to use Azure Cache for Redis with Entra ID see this:

This will assume that either your UserPrincipal (via AzureCLICredential) or the SystemAssigned Managed Identity have the RedisDataOwner role permission. If a user-assigned identity is to be used the `azureClientID` property needs to be specified.

```
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: statestore
spec:
  type: state.redis
  initTimeout: 30s
  metadata:
    - name: redisHost
      value: MYHOSTNAME.redis.cache.windows.net:6380
    - name: useEntraID
      value: "true"
    - name: enableTLS
      value: "true"
```

## Issue reference

Adds #3088

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

## Testing

Manually modified certification and conformance tests to run again Azure Cache for Redis using modified component configurations to exercise the new auth mechanism.

### DIFF
--- a/bindings/redis/metadata.yaml
+++ b/bindings/redis/metadata.yaml
@@ -195,7 +195,8 @@ metadata:
     example: "10m"
   - name: useEntraID
     required: false
-    default: false
+    default: "false"
+    example: "true"
     type: bool
     description: |
       If set, enables authentication to Azure Cache for Redis using Azure Managed Identity
--- a/bindings/redis/metadata.yaml
+++ b/bindings/redis/metadata.yaml
@@ -193,12 +193,18 @@ metadata:
       "-1" disables idle timeout check.
     default: "5m"
     example: "10m"
-  - name: useEntraID
-    required: false
-    default: "false"
-    example: "true"
-    type: bool
-    description: |
-      If set, enables authentication to Azure Cache for Redis using Azure Managed Identity
-      or Azure CLI Credential. The Redis server must explicitly enable EntraID authentication. Note that
-      Azure Cache for Redis also requires the use of TLS, so `enableTLS` should be set.
+builtinAuthenticationProfiles:
+  - name: "azuread"
+    metadata:
+      - name: useEntraID
+        required: false
+        default: "false"
+        example: "true"
+        type: bool
+        description: |
+          If set, enables authentication to Azure Cache for Redis using Microsoft EntraID. The Redis server must explicitly enable EntraID authentication. Note that
+          Azure Cache for Redis also requires the use of TLS, so `enableTLS` should be set. No username or password should be set.
+      - name: enableTLS
+        required: true
+        description: Must be set to true if using EntraID
+        example: "true"
--- a/bindings/redis/metadata.yaml
+++ b/bindings/redis/metadata.yaml
@@ -193,3 +193,11 @@ metadata:
       "-1" disables idle timeout check.
     default: "5m"
     example: "10m"
+  - name: useEntraID
+    required: false
+    default: false
+    type: bool
+    description: |
+      If set, enables authentication to Azure Cache for Redis using Azure Managed Identity
+      or Azure CLI Credential. The Redis server must explicitly enable EntraID authentication. Note that
+      Azure Cache for Redis also requires the use of TLS, so `enableTLS` should be set.
--- a/common/component/redis/redis.go
+++ b/common/component/redis/redis.go
@@ -232,8 +232,8 @@ func StartEntraIDTokenRefreshBackgroundRoutine(client RedisClient, username stri
 		backoffConfig.Policy = kitretry.PolicyExponential
 
 		var backoffManager backoff.BackOff
-		var refreshGracePeriod time.Duration = 2 * time.Minute
-		var tokenRefreshDuration time.Duration = time.Until(nextExpiration.Add(-refreshGracePeriod))
+		var refreshGracePeriod = 2 * time.Minute
+		var tokenRefreshDuration = time.Until(nextExpiration.Add(-refreshGracePeriod))
 
 		(*logger).Debugf("redis client: starting entraID token refresh loop")
 
--- a/common/component/redis/redis.go
+++ b/common/component/redis/redis.go
@@ -24,8 +24,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	"github.com/cenkalti/backoff/v4"
-	"github.com/golang-jwt/jwt/v5"
-
+	"github.com/lestrrat-go/jwx/v2/jwt"
 	"golang.org/x/mod/semver"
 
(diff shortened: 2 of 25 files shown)


berndverst:
I will cherry pick this into Release 1.14 once merged.

common/component/redis/redis.go
```diff
@@ -187,11 +210,135 @@ func ParseClientFromProperties(properties map[string]string, componentType metad
...
+		var backoffManager backoff.BackOff
+		refreshGracePeriod := 2 * time.Minute
+		tokenRefreshDuration := time.Until(nextExpiration.Add(-refreshGracePeriod))
+
+		(*logger).Debugf("redis client: starting entraID token refresh loop")
+
+		for {
+			(*logger).Debugf("redis client: next entraID token refresh: %v", tokenRefreshDuration)
+			select {
+			case <-ctx.Done():
```
ItalyPaleAle, line 243: Because `ctx` is derived by `context.Background()`, the context is never canceled. So if the component is stopped, we don't stop refreshing tokens. Additionally, if the component is reloaded, we spin a new goroutine for each time

common/component/redis/redis.go
```diff
@@ -187,11 +210,135 @@ func ParseClientFromProperties(properties map[string]string, componentType metad
...
+func StartEntraIDTokenRefreshBackgroundRoutine(client RedisClient, username string, nextExpiration time.Time, cred *azcore.TokenCredential, logger *kitlogger.Logger) {
+	go func(cred *azcore.TokenCredential, username string, logger *kitlogger.Logger) {
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+		backoffConfig := kitretry.DefaultConfig()
+		backoffConfig.MaxRetries = 3
+		backoffConfig.Policy = kitretry.PolicyExponential
+
+		var backoffManager backoff.BackOff
+		refreshGracePeriod := 2 * time.Minute
```
ItalyPaleAle, line 235:
```suggestion
		const refreshGracePeriod = 2 * time.Minute
```

common/component/redis/redis.go
```diff
@@ -15,15 +15,24 @@ package redis
...
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/cenkalti/backoff/v4"
+	"github.com/golang-jwt/jwt/v5"
```
ItalyPaleAle, line 27: In the past we tried to standardize on using [github.com/lestrrat-go/jwx/v2/jwt](https://pkg.go.dev/github.com/lestrrat-go/jwx/v2/jwt) for all jw* things, to avoid adding (direct) dependencies on other packages that do similar things. Please consider switching to this library

common/component/redis/redis.go
```diff
@@ -187,11 +210,135 @@ func ParseClientFromProperties(properties map[string]string, componentType metad
...
+	}
+
+	parsedToken, _, parseErr := jwt.NewParser().ParseUnverified(token.Token, &CustomClaims{})
+	if parseErr != nil {
+		return nil, nil, fmt.Errorf("redis client configuration error: %w", parseErr)
+	}
+	if claims, ok := parsedToken.Claims.(*CustomClaims); ok {
+		s.Username = claims.Oid
+	} else {
+		return nil, nil, fmt.Errorf("redis client configuration error: %w", errors.New("could not parse object ID from Auth token"))
```
ItalyPaleAle, line 339:
```suggestion
		return nil, nil, errors.New("redis client configuration error: could not parse object ID from Auth token")
```

common/component/redis/redis.go
```diff
@@ -187,11 +210,135 @@ func ParseClientFromProperties(properties map[string]string, componentType metad
...
+	}
+
+	s.Password = token.Token
+
+	type CustomClaims struct {
+		Oid string `json:"oid"`
+		jwt.RegisteredClaims
+	}
+
+	parsedToken, _, parseErr := jwt.NewParser().ParseUnverified(token.Token, &CustomClaims{})
```
ItalyPaleAle, line 332: Nit: please add a comment explaining that we use ParseUnverified because the Azure SDK already validated the token

bindings/redis/metadata.yaml
```diff
@@ -193,3 +193,12 @@ metadata:
       "-1" disables idle timeout check.
     default: "5m"
     example: "10m"
+  - name: useEntraID
```
ItalyPaleAle, line 196: This should be a separate authentication profile. See how it's done in Postgres for example (and same on other metadata.yaml's)

common/component/redis/redis.go
berndverst: We should probably make that a linter rule then! :)

bindings/redis/metadata.yaml
berndverst: Ah right - let me take a look. I forgot about that.

common/component/redis/redis.go
```diff
@@ -132,9 +142,9 @@ func ParseClientFromProperties(properties map[string]string, componentType metad
 		settings.Concurrency = 10
 	}
 
-	err = settings.Decode(properties)
-	if err != nil {
-		return nil, nil, fmt.Errorf("redis client configuration error: %w", err)
+	decodeErr := settings.Decode(properties)
```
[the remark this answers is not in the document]
berndverst: It's habit for readability, but also because the "shadow" linter - which didn't necessarily trigger here, but that linter taught me to always have dedicated error names.

ItalyPaleAle (approved):
LGTM!

berndverst:
The feature will work like this for all Redis components:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: statestore
spec:
  type: state.redis
  initTimeout: 30s
  metadata:
    - name: redisHost
      value: MYHOSTNAME.redis.cache.windows.net:6380
    - name: useEntraID
      value: "true"
    - name: enableTLS
      value: "true"
```

(change the component type of course)

TLS must be enabled, and the `useEntraID` property must be set. Beyond that you only need to specify the hostname with port (6380 seems to be the Azure Redis default).

This will automatically support all forms of EntraID (AAD) authentication, including Workload Identity, Managed Identities and also Azure CLI Credential (User principal). For more details on all those see:
https://docs.dapr.io/developing-applications/integrations/azure/azure-authentication/authenticating-azure/

Of course **don't forget to enable Entra ID authentication** on your Azure Redis server itself! And enabling it takes a few minutes to propagate as well.

berndverst:
In my own testing I just assigned myself (my user) the `Azure Redis Data Owner` role, then refreshed my CLI Credential. Then I used the above component yaml locally (with my server name). Everything worked perfectly fine.

PR #3478, opened 2024-07-05
snssqs: fix consumer starvation

### LINKED ISSUE #3479: snssqs: fix consumer starvation
## Actual Behavior
The current implementation uses separate goroutines to process each message retrieved by sqsClient.ReceiveMessageWithContext, which is executed when concurrencyMode is set to parallel. The consumeSubscription function waits for all spawned goroutines to complete before receiving the next batch. However, this design has a critical flaw: it starves the consumer if new messages are published during the processing of an existing batch. These newly published messages won't be processed until all goroutines from the original batch have finished their work, effectively blocking the consumption of the new messages.

## Expected Behavior
Message handling should not be block until the batch is completely processed.

- Set a handler that blocks for a long time.
- Publish one sns message.
- Message will be handled.
- Publish another message
- Message will not be handled until the first finishes.

qustavo:
Closes #3479


ItalyPaleAle:
Thanks for this PR. I can see why the change is useful and agree the current implementation has some issues.

However, by removing all waiters, what I am concerned with now is that Dapr could be "overloaded" if too many messages are received at once, as it would create new goroutines without any limit.

Perhaps a better solution could be to limit the number of concurrent goroutines that are created? Some other components (like Azure Service Bus) have similar configuration options where you can set a max number of active messages. This way there's also some "natural backpressure" implemented.

qustavo:
> Thanks for this PR. I can see why the change is useful and agree the current implementation has some issues.
> 
> However, by removing all waiters, what I am concerned with now is that Dapr could be "overloaded" if too many messages are received at once, as it would create new goroutines without any limit.
> 
> Perhaps a better solution could be to limit the number of concurrent goroutines that are created? Some other components (like Azure Service Bus) have similar configuration options where you can set a max number of active messages. This way there's also some "natural backpressure" implemented.

Backpressure is definitively valuable, and from what I can see, implementations like Redis are already controlling the number of goroutines: https://github.com/dapr/components-contrib/blob/main/pubsub/redis/redis.go#L89.

However, this sounds like a global concern that could be implemented at the pubsub level and not as a driver feature. I think that a global concurrency control allow operators to switch from pubsub implementation keeping the same backpressure settings.
Does that makes sense?

ItalyPaleAle:
> However, this sounds like a global concern that could be implemented at the pubsub level and not as a driver feature. I think that a global concurrency control allow operators to switch from pubsub implementation keeping the same backpressure settings.

That does make sense, but it would require some major changes to a lot of components. Additionally integrating that would not be straightforward, since Dapr currently delegates concurrency control for this scenario to individual components.

If for now you could implement a `maxConcurrentMessages` (or `maxParallelism`) option, that would be nice and we can then fix this bug!

qustavo:
> That does make sense, but it would require some major changes to a lot of components. Additionally integrating that would not be straightforward, since Dapr currently delegates concurrency control for this scenario to individual components.

You are right. Is this something you think its worth working on?

> 
> If for now you could implement a `maxConcurrentMessages` (or `maxParallelism`) option, that would be nice and we can then fix this bug!

Sure, I'm happy to add the feature. Initially I thought about `concurrencyLimit` which matches the `concurrencyMode` prefix, but I don't have a string opinion.
Also, some questions:
- what do you think is a sensible default? 
- should we allow the operator to uncap the concurrency limit (let's say you want to control the limit in the caller's code (outside dapr) ?

ItalyPaleAle:
>  Initially I thought about concurrencyLimit which matches the concurrencyMode prefix, but I don't have a string opinion.

`concurrencyLimit` is a good value!

> what do you think is a sensible default?

I just looked at the Azure Service Bus component and looks like the default is actually 0, i.e. unlimited. I'm not a fan of unlimited, but for consistency it could make sense. However, I'll accept other suggestions too!

> should we allow the operator to uncap the concurrency limit (let's say you want to control the limit in the caller's code (outside dapr) ?

My gut initially said "no", but then I saw that other components do allow un-capping the limit (like Service Bus), so perhaps it does make sense. (although I would personally not recommend controlling the limit on the caller (i.e. the publisher) side)

qustavo:
> `concurrencyLimit` is a good value!

Ok, let'a use that one then.

> I just looked at the Azure Service Bus component and looks like the default is actually 0, i.e. unlimited. I'm not a fan of unlimited, but for consistency it could make sense. However, I'll accept other suggestions too!

While `Azure` and `--app-max-concurrency`  defaults to unlimited, [Redis](https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-redis-pubsub/) defaults to "10". However, since there is a wide range of use cases, I think default to unlimited makes more sense.

> My gut initially said "no", but then I saw that other components do allow un-capping the limit (like Service Bus), so perhaps it does make sense. (although I would personally not recommend controlling the limit on the caller (i.e. the publisher) side)

Right, by `caller's code` I mean the Subscriber's handler. An analogous example will be Go's http server, which does not impose any limitation on the number of handlers and it's up to the user to set such a limit.

In summary, I'd document this as follows:

|Field | Required | Details | Example|
|---      |---                |---           |---          |
|concurrencyLimit |N | Defines the maximum number of concurrent workers handling messages. This value is ignored when `concurrentMode` is set to "single". To avoid limiting the number of concurrent workers set this to "0". Default: "0" | "100" |

pubsub/aws/snssqs/metadata.yaml
```diff
@@ -138,9 +138,18 @@ metadata:
     default: '"parallel"'
     example: '"single", "parallel"'
     type: string
+  - name: concurrencyLimit
+    required: false
+    description: |
+      Defines the maximum number of concurrent workers handling messages.
+      This value is ignored when concurrentMode is set to “single“.
```
ItalyPaleAle, line 145: typo
```suggestion
      This value is ignored when "concurrencyMode" is set to “single“.
```

pubsub/aws/snssqs/snssqs.go
```diff
@@ -625,33 +631,34 @@ func (s *snsSqs) consumeSubscription(ctx context.Context, queueInfo, deadLetters
...
 			case pubsub.Single:
 				f(message)
 			case pubsub.Parallel:
-				wg.Add(1)
 				go func(message *sqs.Message) {
-					defer wg.Done()
+					// This is the back pressure mechanism.
+					// It will block until another goroutine frees a slot.
+					if sem != nil {
+						sem <- struct{}{}
```
ItalyPaleAle, line 654: I wonder if this line (that blocks) should be outside of the lambda. Currently, the goroutine is still spawned for each message, but then the goroutine is blocked. Perhaps it'd be wiser to not create the goroutine at all, and block the loop, since too many goroutines (even blocked) can "clog" the entire Dapr application (and cause a lot of pressure on the GC)?

berndverst:
@qustavo can you open an issue in this repo and also explain the problem there then link to it in this PR? We don't usually accept PRs from non-maintainers/approvers without an issue being open first (and I don't mean issues that have been opened at the same time as the PR)

Please note that the 1.14 release is already underway and we have already cut the components-contrib release branch for 1.14. So this will not make it into the 1.14 release most likely. We can assess the need for this fix to improvement to make it into the 1.14 release within the issue (to be created).

berndverst (dismissed):

qustavo:
> @qustavo can you open an issue in this repo and also explain the problem there then link to it in this PR? We don't usually accept PRs from non-maintainers/approvers without an issue being open first (and I don't mean issues that have been opened at the same time as the PR)

Sure, will create issue

berndverst:
I'm ok with merging now - but in the absence of an issue this definitely cannot be considered to be included in 1.14.

berndverst:
Unfortunately the SNS integration tests have been broken. I think this needs to be fixed by @sicoyle @artursouza or someone else with more AWS integration test experience (and the credentials to manage the test infrastructure).

While I'm sure nothing is inherently wrong with this PR, we just cannot be certain due to the broken tests (which were not broken by this PR).

qustavo:
@berndverst AFAICS the issue is with `pubsub.mqtt3 certification`. How is  this related to SNS/SQS?

qustavo:
any progress on this? This is currently stopping our sns/sqs deployment, please lmk if there's anything I can do to get it merged.

yaron2:
> any progress on this? This is currently stopping our sns/sqs deployment, please lmk if there's anything I can do to get it merged.

Taking a look today

yaron2:
@qustavo the build is failing, please see the errors

qustavo:
> @qustavo the build is failing, please see the errors

Sorry, I assumed it was a certification error, seems like it was a merge issue, will fix.

qustavo:
@ItalyPaleAle @yaron2  Trying to reproduce the CI error by running:
```bash
DAPR_TEST_MQTT_URL=tcp://localhost:1883 go test ./...  -v
```

But locally all my tests are passing, not really sure how to reproduce it, any suggestion?

berndverst (approved):

berndverst:
@qustavo please open a docs PR to document the concurrency limit property for the component. Once you link the docs PR here we can merge this.

qustavo:
> @qustavo please open a docs PR to document the concurrency limit property for the component. Once you link the docs PR here we can merge this.

Docs have been updated here: https://github.com/dapr/components-contrib/pull/3478/files#diff-337af9a04d4da7e737aeb7f517ae43a22a3125b4b65cebe94c6d7f39f3c3b06c where else would you document the properties?

Also, tests are still failing but I don't know how this change is relevant to those failures, any hint I can use to diagnose the problem?

qustavo:
> @qustavo please open a docs PR to document the concurrency limit property for the component. Once you link the docs PR here we can merge this.

ping @berndverst

yaron2:
Thanks for the contribution @qustavo! Please update this section of the docs: https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-aws-snssqs/

marcduiker:
@holopin-bot @qustavo Thank you Gustavo!

PR #3562, opened 2024-10-07
fix(metadata): make access/secret keys optional

sicoyle:
# Description

In the event someone is using AWS IAM authentication, then they can leverage the loading of default credentials here: https://github.com/dapr/components-contrib/blob/9012bdce7f0eb67c1c22fa1928a657b27d4bdd31/common/authentication/aws/aws.go#L105

Alternatively, they can supply the access key and secret key. Therefore, the accesskey and secretkey fields are optional, not required.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/common/authentication/postgresql/metadata.go
+++ b/common/authentication/postgresql/metadata.go
@@ -91,14 +91,10 @@ func (m *PostgresAuthMetadata) ValidateAwsIamFields() (string, string, string, e
 	if awsRegion == "" {
 		return "", "", "", errors.New("metadata property AWSRegion is missing")
 	}
+	// Note: access key and secret keys can be optional
+	// in the event users are leveraging the credential files for an access token.
 	awsAccessKey, _ := metadata.GetMetadataProperty(m.awsEnv.Metadata, "AWSAccessKey")
-	if awsAccessKey == "" {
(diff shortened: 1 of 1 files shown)


common/authentication/postgresql/metadata.go
```diff
@@ -91,14 +91,10 @@ func (m *PostgresAuthMetadata) ValidateAwsIamFields() (string, string, string, e
 	if awsRegion == "" {
 		return "", "", "", errors.New("metadata property AWSRegion is missing")
 	}
+	// Note: access key and secret keys can be optional
```
JoshVanL, line 94: Does the metadata yaml need to be updated?

JoshVanL (changes requested):

common/authentication/postgresql/metadata.go
famarting: no https://github.com/dapr/components-contrib/blob/main/state/postgresql/v2/metadata.yaml#L61

JoshVanL: I see, the yaml was previously wrong, now it's correct :laughing:

JoshVanL (approved):

yaron2 (approved):

Issue #3575, opened 2024-10-21
[Mongo] - Serialization issues with Transaction upsert operation


luigirende:
This issue is same described https://github.com/dapr/dapr/issues/2016 for Redis State store

## Expected Behavior

When a state is saved via the the executeTransaction method via GRPC, `Multi` method implementation for mongodb state store, and the state is again read in via the getState API, there is a deserialization issue.
In the collection the _value_ field is always stored as string also if the `contentType=application/json` is present.

## Actual Behavior

State saved via transactional API should be read correctly via the Get State API and save the value as bson.D and not as string if the contentType is equals to `application/json`

## Steps to Reproduce the Problem

1. Install Dapr 1.14.4
2. Create a GRPC server implementing the Execute Transaction call
3. Save State via the transaction API GRPC and set in the metadata `contentType=application/json`
4. Get State and deserialize

## Potential Issue

In the runtime, in the `ExecuteStateTransaction` method defined in `grpc.go` file the `req.Value` parameter of the `SetRequest` is always set as `byte[] `

```
			setReq := state.SetRequest{
				Key: key,
				// Limitation:
				// components that cannot handle byte array need to deserialize/serialize in
				// component specific way in components-contrib repo.
				Value:    req.GetValue(),
				Metadata: req.GetMetadata(),
			}
```
because the `req.GetValue` is _bytes_  in the proto request.

```
// StateItem represents state key, value, and additional options to save state.
message StateItem {
  // Required. The state key
  string key = 1;

  // Required. The state data for key
  bytes value = 2;

  // The entity tag which represents the specific version of data.
  // The exact ETag format is defined by the corresponding data store.
  Etag etag = 3;

  // The metadata which will be passed to state store component.
  map<string,string> metadata = 4;

  // Options for concurrency and consistency to save the state.
  StateOptions options = 5;
}
```
 At the end of `ExecuteStateTransaction` method the `Multi` interface is called.
Specifically in mongo component `mongodb.go`, the `byte[]` value is marshaled as string, in fact the steps are _Multi_ function call _doTransaction_ and this call _setInternal_ where the value as `byte[]` are casted to string.

```
	var v interface{}
	switch obj := req.Value.(type) {
	case []byte:
		v = string(obj)
	case string:
		v = fmt.Sprintf("%q", obj)
	default:
		v = req.Value
	}
```

[...]

PR #3576, opened 2024-10-21
Mongo State: fix serialization value in the transaction method

luigirende:
# Description

This PR to fix the issue about the serialization of the value in the transactional request when the contentType is equals to `application/json` as described in https://github.com/dapr/components-contrib/issues/3575

## Issue reference

Please reference the issue this PR will close: _https://github.com/dapr/components-contrib/issues/3575_

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


berndverst:
I tried this many times but something was always broken with serialization and didn't have time to further investigate.

Mongo serialization is very very finnicky because BSON doesn't perfectly map to JSON and to Dapr's API.

Make sure to run both certification tests and conformance tests after making any changes!

`go test -v -tags=conftests -count=1 ./tests/conformance -run="TestStateConformance/mongodb"`

and also from the `tests/certification/state/mongodb` folder `go test -v --count=1 .`

berndverst (changes requested):
Please add a test case in conformance tests that can execute this particular scenario / new code path added here.

Make sure that test doesn't break other components however.

https://github.com/dapr/components-contrib/blob/main/tests/conformance/state/state.go

luigirende:
> I tried this many times but something was always broken with serialization and didn't have time to further investigate.
> 
> Mongo serialization is very very finnicky because BSON doesn't perfectly map to JSON and to Dapr's API.
> 
> Make sure to run both certification tests and conformance tests after making any changes!
> 
> `go test -v -tags=conftests -count=1 ./tests/conformance -run="TestStateConformance/mongodb"`
> 
> and also from the `tests/certification/state/mongodb` folder `go test -v --count=1 .`

About the test I did and were passed, in fact you can see the job in actions passed the test about confromance https://github.com/dapr/components-contrib/actions/runs/11499402962/job/32020955182?pr=3576

luigirende:
> Please add a test case in conformance tests that can execute this particular scenario / new code path added here.
> 
> Make sure that test doesn't break other components however.
> 
> https://github.com/dapr/components-contrib/blob/main/tests/conformance/state/state.go

@berndverst added the test in conformance

luigirende:
@berndverst I ran the tests locally and found that the issue with Redis 6 is related to the contentType I added in the Multi method. Specifically, the value stored in Redis is in Base64 format when the data request is represented as a byte array. I did a check in the response to decode the result and validate the content. As for Redis 7, when using the Multi method, the contentType is set to JSON; however, the version used in the conformance does not support this feature.

luigirende:
@berndverst any feedback about the PR?

yaron2 (approved):

marcduiker:
@holopin-bot @luigirende Thank you!

PR #3597, opened 2024-11-12
aws auth: add missing auth methods

famarting:
# Description

Testing AWS IAM auth with postgres and missing the assume role feature that its documented as part of the aws auth profile

It all started because this error check here is bad https://github.com/dapr/components-contrib/blob/main/common/authentication/aws/aws.go#L133
if there is no error, `errors.Is(err, config.SharedConfigAssumeRoleError{}.Err)` always evaluates to true... because when you do `config.SharedConfigAssumeRoleError{}.Err` then `Err` is nil!

I started fixing that but then looking at the kafka pubsub component https://github.com/dapr/components-contrib/blob/main/common/component/kafka/auth.go#L99 , kafka supports more fields, so it made sense to also support these for postgres.

docs updated here https://github.com/dapr/docs/pull/4429

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/common/authentication/aws/aws.go
+++ b/common/authentication/aws/aws.go
@@ -15,7 +15,6 @@ package aws
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"strconv"
 	"time"
@@ -23,7 +22,9 @@ import (
 	awsv2 "github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/config"
 	v2creds "github.com/aws/aws-sdk-go-v2/credentials"
+	"github.com/aws/aws-sdk-go-v2/credentials/stscreds"
 	"github.com/aws/aws-sdk-go-v2/feature/rds/auth"
+	"github.com/aws/aws-sdk-go-v2/service/sts"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/request"
@@ -108,8 +109,14 @@ type AWSIAM struct {
 	AWSAccessKey string `json:"awsAccessKey" mapstructure:"awsAccessKey"`
 	// Secret key to use for accessing PostgreSQL.
 	AWSSecretKey string `json:"awsSecretKey" mapstructure:"awsSecretKey"`
+	// Optional session token to use when using access key and secret key
+	AWSSessionToken string `json:"awsSessionToken" mapstructure:"awsSessionToken"`
 	// AWS region in which PostgreSQL is deployed.
 	AWSRegion string `json:"awsRegion" mapstructure:"awsRegion"`
+	// Optional role to assume
+	AWSIamRoleArn string `json:"awsIamRoleArn" mapstructure:"awsIamRoleArn"`
+	// Optional session name to use when assuming a role
+	AWSStsSessionName string `json:"awsStsSessionName" mapstructure:"awsStsSessionName"`
 }
 
 type AWSIAMAuthOptions struct {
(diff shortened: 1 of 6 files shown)


common/authentication/postgresql/metadata.go
```diff
@@ -86,16 +86,30 @@ func (m *PostgresAuthMetadata) InitWithMetadata(meta map[string]string, opts Ini
...
-	// in the event users are leveraging the credential files for an access token.
+	// only region is mandatory
+
 	awsAccessKey, _ := metadata.GetMetadataProperty(m.awsEnv.Metadata, "AWSAccessKey")
 	awsSecretKey, _ := metadata.GetMetadataProperty(m.awsEnv.Metadata, "AWSSecretKey")
-	return awsRegion, awsAccessKey, awsSecretKey, nil
+	awsSessionToken, _ := metadata.GetMetadataProperty(m.awsEnv.Metadata, "AWSSessionToken")
+
+	awsIamRoleArn, _ := metadata.GetMetadataProperty(m.awsEnv.Metadata, "awsIamRoleArn")
+	awsStsSessionName, _ := metadata.GetMetadataProperty(m.awsEnv.Metadata, "awsStsSessionName")
```
sicoyle, line 101:
```suggestion
	sessionName, _ := metadata.GetMetadataProperty(m.awsEnv.Metadata, "sessionName")
```

I don't think we need to expose to end users the underlying mechanisms leveraging Sts through which one can assume a new role with new creds

common/authentication/aws/aws.go
```diff
@@ -118,40 +119,66 @@ type AWSIAMAuthOptions struct {
 	Region           string          `json:"region" mapstructure:"region"`
 	AccessKey        string          `json:"accessKey" mapstructure:"accessKey"`
 	SecretKey        string          `json:"secretKey" mapstructure:"secretKey"`
+	SessionToken     string          `json:"sessionToken" mapstructure:"sessionToken"`
+
+	AssumeRoleArn         string `json:"assumeRoleArn" mapstructure:"assumeRoleArn"`
```
famarting, line 124: agreed to change this fields tags, but they are not used for anything

common/authentication/postgresql/metadata.go
```diff
@@ -86,16 +86,30 @@ func (m *PostgresAuthMetadata) InitWithMetadata(meta map[string]string, opts Ini
...
-	// Note: access key and secret keys can be optional
-	// in the event users are leveraging the credential files for an access token.
+	// only region is mandatory
+
 	awsAccessKey, _ := metadata.GetMetadataProperty(m.awsEnv.Metadata, "AWSAccessKey")
 	awsSecretKey, _ := metadata.GetMetadataProperty(m.awsEnv.Metadata, "AWSSecretKey")
-	return awsRegion, awsAccessKey, awsSecretKey, nil
+	awsSessionToken, _ := metadata.GetMetadataProperty(m.awsEnv.Metadata, "AWSSessionToken")
+
+	awsIamRoleArn, _ := metadata.GetMetadataProperty(m.awsEnv.Metadata, "awsIamRoleArn")
```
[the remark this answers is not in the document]
famarting: this cannot be changed, it must match with metadata.yaml , metadata.yaml has been set to match the existing metadata names for the kafka component, so we have consistency

common/authentication/postgresql/metadata.go
famarting: this cannot be changed, it must match with metadata.yaml , metadata.yaml has been set to match the existing metadata names for the kafka component, so we have consistency

sicoyle:
This PR adds "missing auth methods" that are not actually missing. These are for a different auth profile, that I've added for us to in this PR: https://github.com/dapr/components-contrib/pull/3606 when moving the Kafka component to use the built in AWS auth profiles. 

Please do not merge this PR until we get an official auth profile for assuming an IAM role. We should not continue down the path of adding new disjoint fields to kafka/postgres specific to IAM until they get moved to the common AWS auth profiles. Otherwise, we would have to keep deprecating old component specific IAM fields, adding to maintainer overhead.

famarting:
ok, closing

added feedback to the kafka and postgres PR. I'm missing the assume role functionality for the postgres components

Issue #3609, opened 2024-11-22
Kafka Pubsub avro serialization caching default not set


passuied:
## Expected Behavior

When leveraging Kafka Pubsub Avro serialization, caching should be enabled by default with a TTL of 5min

## Actual Behavior

Caching is disabled by default, making performance terrible.

## Steps to Reproduce the Problem

- Set up kafka pubsub component with
  - schemaRegistryUrl configured to point to a valid schema registry
  - SchemaRegistryAPIKey and SchemaRegistryAPISecret configured adequately
- Set up a subscription with this component with `valueSchemaType=avro` in subscription metadata or `publish_metadata`
- When consuming or publishing messages, schema caching will be disabled by default.

## Release Note

RELEASE NOTE: **FIX** Kafka pubsub Avro serialization/deserialization respects caching default = true

Issue #3529, opened 2024-09-03
Kafka pubsub message gets dropped despite default retry forever policy


passuied:
## Expected Behavior

Kafka pubsub invalid message gets retried forever when default retry policy in place

## Actual Behavior

Kafka pubsub invalid message gets dropped upon the pod terminating

## Steps to Reproduce the Problem

1. Use default retry policy (retry forever)
2. publish valid message to topic my_topic
3. publish an invalid message and another valid message to the same topic
5. set up a consumer consuming my_topic which returns 500 when consuming invalid message
6. Observe that dapr will retry processing the message over and over
7. Terminate pod
8. In log the following message will show: "Too many failed attempts at processing Kafka message:..."
9. Upon restart, the consumer will move to next message. I.e. the message was dropped...

## Release Note

RELEASE NOTE:  **FIX** Bug in Kafka Pubsub where poison pill message will get dropped upon pod termination despite retry forever policy

passuied:
Based on research, this issue is the result of [this code](https://github.com/dapr/components-contrib/blob/main/common/component/kafka/consumer.go#L75-L84):
- `retry.NotifyRecover()` will retry per the retry policy unless a Permanent error is encountered or the context is Done (see `doRetryNotify()`)
- When bubbled up, the ConsumeClaim() code will just assume the error is a permanent error and return null, causing the message to be ack'ed.

A solution maybe to add special handling in case the error returned was the result of Context.Done() and return the err instead, causing the offset to stay as is and guaranteeing NO data loss.

yaron2:
> Based on research, this issue is the result of [this code](https://github.com/dapr/components-contrib/blob/main/common/component/kafka/consumer.go#L75-L84):
> 
> * retry.NotifyRecover() will retry per the retry policy unless a Permanent error is encountered or the context is Done (see doRetryNotify())
> * When bubbled up, the ConsumeClaim() code will just assume the error is a permanent error and return null, causing the message to be ack'ed.
> 
> The solution is not super clear here as there is a lot of ping pong between the dapr and component-contrib code. An expert eye would be welcome. Specifically, it is not clear what acks the message when a permanent error occurs in general... The only place that seems to ack the message is in the doCallBack() if the handler doesn't return an error...

cc @halspang can you take a look?

passuied:
> cc @halspang can you take a look?
[...]

PR #3610, opened 2024-11-22
Kafka Pubsub fixes: Avro serialization caching and retries

passuied:
# Description

- When leveraging Kafka Pubsub Avro serialization, caching should be enabled by default with a TTL of 5min
- Also, addresses another issue where poison pill messages could be skipped in some weird circumstances after a context cancelation, which doesn't in turn seem to come from `session.Context()`
  - In that specific case, we're forcing the processing back into the retry policy for the same message

## Issue reference

#3609
#3529

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: N/A


common/component/kafka/kafka.go
```diff
@@ -321,8 +330,10 @@ func (k *Kafka) getLatestSchema(topic string) (*srclient.Schema, *goavro.Codec,
 
 		// Cache present and not expired
 		if ok && cacheEntry.expirationTime.After(time.Now()) {
+			k.logger.Debugf("Schema cache hit for subject %s", subject)
```
yevgen-el8, line 333: with many requests, wouldn't it be too chatty? I assume we can skip it.

common/component/kafka/kafka.go
```diff
@@ -208,14 +208,17 @@ func (k *Kafka) Init(ctx context.Context, metadata map[string]string) error {
 	k.consumeRetryInterval = meta.ConsumeRetryInterval
 
 	if meta.SchemaRegistryURL != "" {
+		k.logger.Debugf("Schema registry URL '%s' provided. Configuring the Schema Registry client.", meta.SchemaRegistryURL)
 		k.srClient = srclient.CreateSchemaRegistryClient(meta.SchemaRegistryURL)
 		// Empty password is a possibility
 		if meta.SchemaRegistryAPIKey != "" {
 			k.srClient.SetCredentials(meta.SchemaRegistryAPIKey, meta.SchemaRegistryAPISecret)
 		}
+		k.logger.Debugf("Schema caching enabled: %v", meta.SchemaCachingEnabled)
```
yevgen-el8, line 217: I would say we can log it with `Info` level as it's logged only during the `Init()` call

yevgen-el8 (approved):

common/component/kafka/kafka.go
```diff
@@ -313,6 +316,12 @@ func (k *Kafka) getLatestSchema(topic string) (*srclient.Schema, *goavro.Codec,
 		return nil, nil, err
 	}
 
+	// measure the time taken to get the schema
```
yaron2, line 319: Why do we need this?

passuied: just because originally I had some concerns about the time taken... I wish we could add a prometheus metric for serialization time but this was my closest proxy in case there are some future concerns about performance

yaron2 (approved):

Issue #3573, opened 2024-10-18
Azure Event Hub binding stops working if storage act returns 503


MattCosturos:
## Expected Behavior

Azure Event Hub binding will continue working after receiving a 503 error from the storage act used for checkpointing.

## Actual Behavior

The event hub binding dies after receiving a 503 error from table storage. The component does not appear to retry the calls to the storage act, and just gives up. Logs become filled with `Error from event processor: the Processor has been stopped. Create a new instance to start processing again`

We are using a standard storage act and not a premium act. I _could_ switch to a premium act for better throughput, but I would still have concerns that the component could still fail to process messages if it experiences any sort of transient error. 

Can we somehow configure a retry policy to restart this component instead of filling the logs with an error?

```
time="2024-10-17T19:48:21.031356514Z" level=warning msg="The default value for 'spec.metric.http.increasedCardinality' will change to 'false' in Dapr 1.15 or later" app_id=component-telemetry instance=component-telemetry-microservice-54454fbdd7-mxr8s scope=dapr.runtime.diagnostics type=log ver=1.14.4
time="2024-10-17T19:48:24.243756578Z" level=warning msg="The default value for 'spec.metric.http.increasedCardinality' will change to 'false' in Dapr 1.15 or later" app_id=component-telemetry instance=component-telemetry-microservice-54454fbdd7-mxr8s scope=dapr.runtime.http type=log ver=1.14.4
time="2024-10-18T01:25:52.104204839Z" level=error msg="Error processing events from partition client: error receiving events: client has been closed by user" app_id=component-telemetry component="component-telemetry-1 (bindings.azure.eventhubs/v1)" instance=component-telemetry-microservice-54454fbdd7-mxr8s scope=dapr.contrib type=log ver=1.14.4
time="2024-10-18T01:25:56.922083584Z" level=error msg="Error processing events from partition client: error receiving events: client has been closed by user" app_id=component-telemetry component="component-telemetry-0 (bindings.azure.eventhubs/v1)" instance=component-telemetry-microservice-54454fbdd7-mxr8s scope=dapr.contrib type=log ver=1.14.4
[...]

PR #3614, opened 2024-11-26
Recreate AEH Processor in the event of an error before retrying the processing operation

MattCosturos:
Recreate AEH Processor in the event of an error before retrying the processing operation

# Description

Moved the creation and setup of the Azure Event Hub Processor inside the main loop. 
The processor.run only returns a non-recoverable error. Therefore we must re-create new processor and setup the subscription.

## Issue reference

Please reference the issue this PR will close: #3573 
## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


yaron2:
See failing tests

MattCosturos:
Line 280, can I simply log the error and return nil?
or do we need to return that error out of the goroutine and back up the stack?

yaron2:
@artursouza for last review

artursouza (approved):

marcduiker:
@holopin-bot @MattCosturos Thank you!

Issue #3554, opened 2024-09-25
Redis Entra Id connection stops working after 24hours


skbar50:
## Expected Behavior

We started using Entra Id authentication to Redis with the dapr version 1.14.1. We are able to establish the connection from the AKS pod to Redis using the managed identity used by the pod. The pod runs for many days without restarting. It should be able to connect the Redis whenever there is a request from client.   

## Actual Behavior
However, after running for about 24hours the pod is failing to connect Redis with the error 'State operation failed: the Dapr endpoint indicated a failure. See InnerException for details. Status(StatusCode="Internal", Detail="fail to get PComSearchCacheProduct from state store statestore: ERR WRONGPASS invalid username-password pair") '

## Steps to Reproduce the Problem

Run the pod that connects Redis with Entra ID authentication for more then 24hours. It will fail with the above mentioned error after about 24hours. 

## Additional information
The AKS pod has been working with the same Redis statestore with the password without any issue. We faced this problem when we tried to implement Entra Id authentication.

skbar50:
We are still facing the issue. It has not been resolved yet. We need a solution to this.

swigerb:
@yaron2 Any way to escalate this issue? It appears to be a timeout bug in the code and Entra ID auth to Redis fails after 24 hours. Has anyone tested Entra ID auth to Azure Redis after a pod running for 24 hours?

shivamkm07:
@swigerb yes this was a bug in the implementation. #3632 would fix this.

skbar50:
The issue is not resolved  yet. The AKS pod that runs the redis client application fails after 24 hours. In the log we observed the token is now being refreshed. We suspect the token may not be applied in the connection. 

After 24 hours the application is giving this error - 'ERR WRONGPASS invalid username-password pair’

Attached is the application log.

[daprlogs.txt](https://github.com/user-attachments/files/19765834/daprlogs.txt)

tswanenberg:
Like @skbar50, I ended up here after encountering this exact issue. Initially the Azure Managed Redis setup with Dapr using managed identity works great but stops working after a day.. going back to access keys until this is fixed, which is a bummer.

georgii-sirotkin:
I encountered the same issue.
Dapr version: `1.15.9-rc.1`

Dapr sidecar logs:
```
[...]

PR #3632, opened 2024-12-13
Fix Redis EntraId Token Refresh

shivamkm07:
# Description

When authenticating using EntraId Authentication for Azure Cache for Redis, the token needs to be refreshed before the token expires as mentioned here: https://github.com/dapr/components-contrib/issues/3088#issuecomment-1683015805

However, the context being passed to the background routine gets cancelled once the component initialization is done (or timedout for which the default value is 5s) and so auth token is never refreshed once it expires: https://github.com/dapr/dapr/blob/6c488fdf7d2995d074413c875fc2e3fb4134a33b/pkg/runtime/processor/components.go#L215

This PR removes the dependency on initContext so that goroutine keeps running in the background.

## Issue reference

Please reference the issue this PR will close: #3554 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/common/component/redis/redis.go
+++ b/common/component/redis/redis.go
@@ -214,21 +214,21 @@ func ParseClientFromProperties(properties map[string]string, componentType metad
 	// start the token refresh goroutine
 
 	if settings.UseEntraID {
-		StartEntraIDTokenRefreshBackgroundRoutine(c, settings.Username, *tokenExpires, tokenCredential, ctx, logger)
+		StartEntraIDTokenRefreshBackgroundRoutine(c, settings.Username, *tokenExpires, tokenCredential, logger)
 	}
 	return c, &settings, nil
 }
 
-func StartEntraIDTokenRefreshBackgroundRoutine(client RedisClient, username string, nextExpiration time.Time, cred *azcore.TokenCredential, parentCtx context.Context, logger *kitlogger.Logger) {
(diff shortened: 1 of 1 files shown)


yaron2 (approved):

marcduiker:
@holopin-bot @shivamkm07 Thank you Shivam!

berndverst:
FYI the change here is somewhat problematic as it does not allow graceful component reloading. It needs to be connected to a different context object provided by Dapr runtime that is connected to the component runtime lifecycle. What that is however, I do not know. That applies to all components - but is particularly problematic for a background go routine.

skbar50:
We are still seeing this issue. The AKS pod that runs the redis client application is failling after 24 hours. In the log we observed the token is now being refreshed. We suspect the token may not be applied in the connection.

After 24 hours the application is giving this error - 'ERR WRONGPASS invalid username-password pair’

Attached is the application log.
[daprlogs.txt](https://github.com/user-attachments/files/19780612/daprlogs.txt)

I also posted this comment in the issue that I raised earlier - https://github.com/dapr/components-contrib/issues/3554

vijaytdh:
+1 Our developers are using this as well and have encountered the same issue with token expiry. I have to say I could not really follow the code `StartEntraIDTokenRefreshBackgroundRoutine` because `c` which is of type `RedisClient` is passed by value so how does the client get updated in the calling function? I would have expected a pointer needed to be used? 

Is anyone aware of a workaround for this issue?

Issue #3621, opened 2024-12-04
Dapr Sidecar fails to reconnect to rabbitmq cluster defined as pubsub component


sleom:
## Expected Behavior

The Dapr publisher should automatically reconnect to the remaining nodes in the RabbitMQ cluster after one node goes down, allowing messages to be published without manual intervention.

## Actual Behavior

When a node in the RabbitMQ cluster becomes unavailable:
The Dapr publisher fails to reconnect, and publishing to RabbitMQ fails.
Errors are repeatedly logged, with messages indicating that the connection is not open.

## Steps to Reproduce the Problem

[publisher-pod.yaml.txt](https://github.com/user-attachments/files/18003916/publisher-pod.yaml.txt)
[component.yaml.txt](https://github.com/user-attachments/files/18003917/component.yaml.txt)

1. Set up a publisher application with a Dapr pub/sub publisher as a sidecar.
2. Configure the publisher component to use RabbitMQ service created by RabbitMQ cluster deployment on Kubernetes
3. Find the RabbitMQ node that the publisher is connected to
4. Take down/restart this node.
5. Observe the logs and error messages in the publisher.

## Release Note

RELEASE NOTE:

vadim-tychonoff:
still an issue

sleom:
Why is it ignored?

PR #3637, opened 2024-12-23
update configuration property name to match documentation

sleom:
# Description

Align configuration property name with [component documentation](https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-rabbitmq/#:~:text=%2C%20%22false%22-,reconnectWait,-N)

## Issue reference

Please reference the issue this PR will close: #3621


sleom:
Meanwhile, the problem can resolved by adding `ReconnectWaitSeconds` property to the component definition.

Issue #3707, opened 2025-03-18
Support multiple keys per secret in AWS Secrets Manager secrets store


passuied:
## Describe the feature
AWS Secrets manager secrets store currently doesn't support multiple keys per secret. This would be a very useful feature.
We can make it backwards compatible by adding a new optional metadata value `MultipleKeysPerSecret`

The resulting secrets will be outputted as `{key}` e.g. `key1`
If the value is NOT a JSON, it would return the secret string as is (existing behavior)
If the metadata value is set to false, it would return the secret string as is (existing behavior)

## Release Note
RELEASE NOTE: **ADD** support for multiple keys per secret for AWS Secrets manager secrets store

passuied:
Just waiting for a review

passuied:
@JoshVanL any way to review this PR? It's pretty simple and additive... Not sure why it's closed as `Not Planned`

Issue #2155, opened 2022-10-05
Move to AWS SDK v2


yaron2:
We're currently using the v1.x of the AWS Go SDK. v2 SDKs recently became stable and is recommended to migrate over.

berndverst:
cc @robertojrojas

robertojrojas:
@berndverst I'm planning to work on this for v1.11.

joebowbeer:
@yaron2 end-of-support for AWS SDK for Go (v1) is effective July 31, 2025

https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/

yaron2:
> [@yaron2](https://github.com/yaron2) end-of-support for AWS SDK for Go (v1) is effective July 31, 2025
> 
> https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/

Thanks that's good info

passuied:
@yaron2 I've started playing with Cursor to see if it helps with the upgrade... So far it seems to de a decent job... Should I go all the way? Would be good to have it out for 1.16... Or is someone already working on this on your side?

yaron2:
> [@yaron2](https://github.com/yaron2) I've started playing with Cursor to see if it helps with the upgrade... So far it seems to de a decent job... Should I go all the way? Would be good to have it out for 1.16... Or is someone already working on this on your side?

It'd be great if you can open a PR for this

mikeee:
@passuied I've migrated sqs/sns with the common auth but won't have time blocked out for a little while. Feel free to open a PR and I'm happy to defer to you 🙂

passuied:
I'm using it as a way to gauge how good cursor is... (it should be in theory for these mindless updates...). Let me see how far I can get in between meetings... :)

PR #3708, opened 2025-03-18
Add support for multiple keys per secret

passuied:
# Description

AWS Secrets manager secrets store currently doesn't support multiple keys per secret. This would be a very useful feature.
We can make it backwards compatible by adding a new optional metadata value MultipleKeyValuesPerSecret

~~Also upgrade AWS sdk to v2~~

## Issue reference

#3707

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs/issues/4590


joebowbeer:
This looks like a reasonable approach.

The same approach can be used for AWS SSM ParameterStore?

Can this be applied to all secret store implementations that don't natively implement multiple values?

How performant is formatString on non-json values?

By the way, aws-sdk-go reaches end-of-support in a few months. #2155

passuied:
@joebowbeer 
Yes we can totally apply this to Ssm parameter store but wanted to keep the PR small. 
I'm sure we can apply this elsewhere too. 

Happy to contribute further to migrate AWS SDK. We heavily use the dynamodb state store and use other components integrating with AWS.

passuied:
> How performant is formatString on non-json values?

I don't have benchmarks but it should fail pretty fast if the first character is not a `{` or a `"`. 

Also I would expect that consumers would only use this option if they know their secrets have multiple k/v. The structure of the secrets should be known ahead.

joebowbeer:
On second look, it looks like this only supports a single level of string-valued keys. At least, I don't see tests for anything else.

Note however that arbitrarily nested JSON values are allowed, including values that are JSON objects and arrays. How can these values be extracted?

Since reviewing this, I've looked at the AWS provider for Secrets Store CSI Driver, which also supports extracting secrets from json-encoded values:

https://github.com/aws/secrets-store-csi-driver-provider-aws

In their solution, the user specifies which paths to extract using a JMES path:

> jmesPath: This optional field specifies the specific key-value pairs to extract from a JSON-formatted secret. You can use this field to mount key-value pairs from a properly formatted secret value as individual secrets.

I don't think Dapr needs to be concerned with name aliasing and mapping, but it occurs to me that it should be able to handle arbitrarily nested JSON. I also think it should have a strategy for creating the keys for the nested values. For example, combine the parts using a default separator such as '/' and allow the user to specify a different separator such as '.'.

Another place to look for ideas is the AWS provider for External Secrets Operator:

https://external-secrets.io/latest/provider/aws-secrets-manager/#json-secret-values

joebowbeer (changes requested):
Given that values can be arbitrarily nested JSON, the tests and docs should be extended to cover how nested values are extracted, including objects and arrays.

passuied:
> Given that values can be arbitrarily nested JSON, the tests and docs should be extended to cover how nested values are extracted, including objects and arrays.

@joebowbeer the DAPR secretstores interface only supports 1 level and any sub level will be left as a string. I can definitely add more tests to reflect but I don't think we can do anything else to support nested JSON use cases.

passuied:
@joebowbeer turns out your call was very important. given the unmarshalling to map[string]string, the setting would NOT work in case of nested values.
I now have improved the handling and of course added some tests. Ready for re-review.

joebowbeer (commented):
Is there a test for top-level array values? They are used with rotating secrets.

passuied:
> Is there a test for top-level array values? They are used with rotating secrets.

Wouldn't you use `versionId` and `versionStage` for that? I could add a test case for this if you know an array is a valid usecase... My expectation is that it would default to a single property with stringified json...

joebowbeer:
https://docs.aws.amazon.com/secretsmanager/latest/userguide/reference_secret_json_structure.html#reference_secret_json_structure_AD

```json
{
  "awsSeamlessDomainDirectoryId": "d-12345abc6e",
  "awsSeamlessDomainUsername": "<username>",
  "awsSeamlessDomainPassword": "<password>",
  "directoryServiceSecretVersion": 1,
  "schemaVersion": "1.0",
  "keytabArns": [
    "<ARN of child keytab secret 1>",
    "<ARN of child keytab secret 2>",
    "<ARN of child keytab secret 3>",
  ],
  "lastModifiedDateTime": "2021-07-19 17:06:58"
}
```

passuied:
> Here's some test input from the docs:
> 
> https://docs.aws.amazon.com/secretsmanager/latest/userguide/reference_secret_json_structure.html#reference_secret_json_structure_AD
> 
> ```json
> {
>   "awsSeamlessDomainDirectoryId": "d-12345abc6e",
>   "awsSeamlessDomainUsername": "<username>",
>   "awsSeamlessDomainPassword": "<password>",
>   "directoryServiceSecretVersion": 1,
>   "schemaVersion": "1.0",
>   "keytabArns": [
>     "<ARN of child keytab secret 1>",
>     "<ARN of child keytab secret 2>",
>     "<ARN of child keytab secret 3>",
>   ],
>   "lastModifiedDateTime": "2021-07-19 17:06:58"
> }
> ```

So really the secret is still an object and not a collection. So the nested test case would cover it...

joebowbeer (commented):
Make sure you request a review from one of the owners. I'm just trying to help.

As for me, I would like to see one of the AWS samples in the test data so that it's more clear what happens in a realistic sample with: numeric values (e.g., port numbers), nested arrays (see AD sample), and nested object (for completeness).

There are a lot samples on this page and I think the AD sample has the most variety.

https://docs.aws.amazon.com/secretsmanager/latest/userguide/reference_secret_json_structure.html

passuied:
@cicoyle any next steps on this PR? I can update the branch but without reviews from maintainers it seems moot

cicoyle:
Hey @passuied - mind fixing linter?

cicoyle:
Thanks for the implementation PR - Can you open a PR for this to the dapr/docs repo adding the new optional field to [this page](https://docs.dapr.io/reference/components-reference/supported-secret-stores/aws-secret-manager/)?

secretstores/aws/secretmanager/secretmanager.go
```diff
@@ -76,6 +79,41 @@ func (s *smSecretStore) Init(ctx context.Context, metadata secretstores.Metadata
...
+}
+
+func (s *smSecretStore) formatSecret(output *secretsmanager.GetSecretValueOutput) map[string]string {
+	result := map[string]string{}
+
+	if output.Name != nil && output.SecretString != nil {
+		if s.multipleKeyValuesPerSecret {
+			data := map[string]any{}
+			err := json.Unmarshal([]byte(*output.SecretString), &data)
+			if err == nil {
```
cicoyle: can you convert this to `if err != nil` to be more idiomatic go code?
```
if err := json.Unmarshal([]byte(*output.SecretString), &data); err != nil {
    result[*output.Name] = *output.SecretString
} else {
    result = convertMapAnyToString(data)
}
```

passuied: Addressed @cicoyle!

cicoyle: can you flip the == still?

cicoyle: to match 
```
if err := json.Unmarshal([]byte(*output.SecretString), &data); err != nil {
    result[*output.Name] = *output.SecretString
} else {
    result = convertMapAnyToString(data)
}
```

so instead of `if err := json.Unmarshal([]byte(*output.SecretString), &data); err == nil {` change it to `!= nil` and flip the logic in the if/else

passuied: fine... I didn't realize that was so important but I'm not a go developer so what do I know :)

passuied: done!

cicoyle (dismissed):
thx for the update! it just reads consistently with the err != nil, its a nit, but for overall readability :)

passuied:
@joebowbeer you seem to have some good AWS experience. Could you review again with the AWS SDK v2 implementation? Especially the x509 auth support... I cannot find a ton of doc online whether this approach is sound or not... Based on [this doc](https://docs.aws.amazon.com/sdk-for-go/v2/developer-guide/migrate-gosdk.html#credentials--credential-providers), it seems like v2 may handle refresh and a lot of the complexity introduced in x509.go file but I'm not sure this is the case...

joebowbeer:
@passuied I think the AWS v2 auth package needs to be thought through before using AWS v2 auth in secretsmanager.

Consider adding a separate package for AWS v2 auth.

https://github.com/dapr/components-contrib/blob/main/common/authentication/aws2 ?

https://github.com/dapr/components-contrib/blob/main/common/authentication/aws/client2.go ?

This decision should be made before accruing more v2 logic in the existing client.

Currently the AWS v2 support in client is isolated to MSK.

passuied:
Thanks @joebowbeer. Any thoughts on the x509 support in place? I had assumed that was what was involved in making auth work in EKS under a service account role but it doesn't seem to be connected... @yaron2 any guidance on needed support for aws sdk v2? Do we still need it? Should I be creating a separate x509 implementation for v2?

joebowbeer:
@passuied I suggest reverting v2 changes.

When migrating to v2, all aws refs should be migrated. But at the moment, I see v2 code calling some aws functions.

It's better to add your feature first, and then migrate to v2, which has a much bigger blast radius.

I don't know why the x509 changes were made. I can't see how they are related to your feature or v2 migration.

passuied:
@joebowbeer I took your advice and will be working on this separately. I still would like to be able to migrate these to v2 one component at a time to avoid a huge PR but I'd rather get this feature merged sooner.
cc @yaron2 @mikeee

mikeee:
https://github.com/dapr/components-contrib/pull/3856 V2 package PR - just investigating whether my changes to the snssqs component have caused the failures or the package which I doubt

passuied:
> just investigating whether my changes to the snssqs component have caused the failures or the package which I doubt

@mikeee not sure what you mean by failures here...
Thanks for sharing the PR. I'll definitely refrain from working on v2 upgrade since you seem to be way ahead....

passuied:
@mikeee if you can review and approve this PR I can create a subsequent one that will upgrade secrets manager to v2

passuied:
@yaron2 @JoshVanL any chance to get this PR reviewed/approved?

Issue #3664, opened 2025-02-11
Dapr PubSub API: Trim Redis stream when TTL is set


alicejgibbons:
## Describe the feature
Dapr should trim the pubsubs in redis streams honoring TTL.

When using redis streams, redis data is not purged once TTL expires. It's needed to write a custom cleaner, but this cleaner does not have visibility to the requested TTL for the message (and the cleaner would break once implementation details from dapr change).

## Release Note

RELEASE NOTE: **ADD** Redis streams TTL deletes messages.

elena-kolevska:
https://github.com/dapr/dapr/issues/6948

elena-kolevska:
The problem users reported here was high memory usage for Redis topics.
Redis streams work as an append-only-log and even though users have the possibility to delete messages from it, removal is not done automatically when a message is acknowledged. For reference, Kafka and AWS SQS behave the same, while RabbitMQ, ASB or Pulsar for example, do remove messagea after acknowledged.

I see two possible options:
- We add a component-level config option (specifically for Redis) to delete messages when acknowledged  (`deleteOnACK`)
- We add a component-level config option specifically for Redis to continuously trim both topics and the DLQ based on number of entries or time ([docs](https://redis.io/docs/latest/commands/xtrim/))

alicejgibbons:
> We add a component-level config option (specifically for Redis) to delete messages when acknowledged (deleteOnACK)
We add a component-level config option specifically for Redis to continuously trim both topics and the DLQ based on number of entries or time ([docs](https://redis.io/docs/latest/commands/xtrim/))

QQ on `deleteOnACK`:
- Moving to DLQ is considered an "ACK" correct?
- This method would allow a case where no consumer ever connects to the broker but if the TTL was set, still wouldn't fill up the memory, correct?

elena-kolevska:
> Moving to DLQ is considered an "ACK" correct?

It's actually the opposite, we move messages to the DLQ when they can't be processed (and thus acknowledged).

> This method would allow a case where no consumer ever connects to the broker but if the TTL was set, still wouldn't fill up the memory, correct?

Yes, if the trim option would be enabled, the trimming will happen every time a message is **added** to the queue.

elena-kolevska:
I'm gonna move forward implementing option number two above: "We add a component-level config option specifically for Redis to continuously trim both topics and the DLQ based on number of entries or time ([docs](https://redis.io/docs/latest/commands/xtrim/))"

[...]

PR #3710, opened 2025-03-19
Adds Redis stream trimming by time(stream ID)

elena-kolevska:
# Description

Adding the option to trim Redis streams based on timestamp. When `minIDApprox` is specified in the component file entries with IDs smaller than the specified threshold are evicted, keeping the stream trimmed to only contain entries with IDs greater than or equal to this value. 
This can be used for time-based eviction since Redis stream IDs are in the format {timestamp-sequence}, where timestamp is in milliseconds since epoch. 

## Issue reference

https://github.com/dapr/components-contrib/issues/3664

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: https://github.com/dapr/docs/pull/4591


yaron2 (dismissed):
LGTM

JoshVanL:
We should use a TTL time duration value as part of the component field, _not_ a timestamp- the timestamp should be computed at runtime.

Component field values should be evergreen.

elena-kolevska:
> We should use a TTL time duration value as part of the component field, not a timestamp- the timestamp should be computed at runtime.

Absolutely! That was my initial idea when i started doing this, I don't remember how it changed by the time I implemented it. I was just looking into it last week. I'll update the pr.

elena-kolevska:
Updated

common/component/redis/settings.go
```diff
@@ -127,6 +130,15 @@ func (s *Settings) SetCertificate(fn func(cert *tls.Certificate)) error {
 	return nil
 }
 
+func (s *Settings) GetMinID(now time.Time) string {
+	// If StreamTTL is not set, return empty string (no trimming)
+	if s.StreamTTL == 0 {
+		return ""
+	}
```
antontroshin, line 137:
```suggestion
	if s.StreamTTL <= 0 {
		return ""
	}
```

elena-kolevska: I think we shouldn't fail silently, and it would be too late to return the error here (we use this on message send). We should probably catch it on init.

nelson-parente (approved):

PR #3749, opened 2025-04-06
Update metadata for ./bindings/alicloud/dingtalk/webhook

jjcollinge:
This PR updates the metadata.yaml file for the ./bindings/alicloud/dingtalk/webhook component.

https://docs.dapr.io/reference/components-reference/supported-bindings/alicloud-dingtalk/


jjcollinge:
note this binding has the direction explicitly in the metadata.yaml as it is explicit in the docs. This is not always the case, for instance, https://docs.dapr.io/reference/components-reference/supported-bindings/alicloudsls/

Issue #3655, opened 2025-01-22
GCP Storage Bucket binding: Bulk file transfer


rahulpoddar-fyndna:
We move files from GCP object store to PV for processing and there are more then 100+ files (ranging from 15 to 30GB) to be transferred daily. 

Dapr supports one by one file transfer and hence not been able to use dapr object store binding for this. One by one file transfer takes long time to complete.

Using GCP SDK (java) completes the 100+ file transfers in around 1 minutes or less. 

We would like to go with dapr binding for this requirement as well. 

Can we have feature similar to how GCP sdk provides for transfer of files from a object store to PV.  Provision to transfer files in parallel from one object store to PV makes it very fast. 
Likewise, support for Azure and AWS will be helpful.

alicejgibbons:
An example here from Rahul ^ for implementation.

The java sample for the bulk file transfer from object store can be referred from below git url https://github.com/googleapis/java-storage/blob/HEAD/samples/snippets/src/main/java/com/example/storage/transfermanager/DownloadBucket.java

If we see above code, two important points are

```java
TransferManager transferManager = TransferManagerConfig.newBuilder().build().getService();
    ParallelDownloadConfig parallelDownloadConfig =
        ParallelDownloadConfig.newBuilder()
            .setBucketName(bucketName)
            .setDownloadDirectory(destinationDirectory)
            .build();

    List<DownloadResult> results =
        transferManager.downloadBlobs(blobs, parallelDownloadConfig).getDownloadResults();

```

So, from a given bucket(bucketName) files are downloaded in parallel (parallelDownloadConfig) and hence it is fast. 

Requirement for moving the files across buckets in a related issue: https://github.com/dapr/components-contrib/issues/3157

PR #3811, opened 2025-05-05
GCP Storage Bucket binding: Bulk file transfer

nelson-parente:
# Description

This PR adds bulk get with parallel file downloads.

docs PR: TODO

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/components-contrib/issues/3655

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


bindings/gcp/bucket/bucket.go
```diff
@@ -404,3 +411,74 @@ func (g *GCPStorage) signObject(bucket, object, ttl string) (string, error) {
...
+
+func (g *GCPStorage) bulkGet(ctx context.Context, req *bindings.InvokeRequest) (*bindings.InvokeResponse, error) {
+	var payload bulkGetPayload
+	err := json.Unmarshal(req.Data, &payload)
+	if err != nil {
+		return nil, fmt.Errorf("gcp bucket binding error while unmarshalling bulk get payload: %w", err)
+	}
+
+	if payload.DestinationPath == "" {
+		return nil, errors.New("gcp bucket binding error: required metadata 'destinationPath' missing")
```
antontroshin, line 427: `metadata` is confusing here as this is not a part of Binding's metadata, but a special payload of a specific operation.
```
"gcp bucket binding error: required request payload property 'destinationPath' missing"
```

bindings/gcp/bucket/bucket.go
```diff
@@ -404,3 +411,74 @@ func (g *GCPStorage) signObject(bucket, object, ttl string) (string, error) {
...
+			defer f.Close()
+
+			rc, err := g.client.Bucket(g.metadata.Bucket).Object(object.Name).NewReader(ctx)
+			if err != nil {
+				errCh <- err
+				return
+			}
+			defer rc.Close()
+
+			if _, err := io.Copy(f, rc); err != nil {
```
antontroshin, line 467: Most likely need to check and use the `encodeBase64` metadata setting, similar to a single `get` operation, in case the objects were stored with this binding to match the encoding setting.
Unless this entire operation explicitly states that encoding settings will be ignored.

bindings/gcp/bucket/bucket.go
```diff
@@ -404,3 +411,74 @@ func (g *GCPStorage) signObject(bucket, object, ttl string) (string, error) {
...
+
+	var allObjs []storage.ObjectAttrs
+	it := g.client.Bucket(g.metadata.Bucket).Objects(ctx, nil)
+	for {
+		attrs, err := it.Next()
+		if err == iterator.Done {
+			break
+		}
+		allObjs = append(allObjs, *attrs)
+	}
```
antontroshin, line 438: No need to wait to collect all objects into a slice; it can never end. GCP buckets can potentially have a quite literally unlimited number of objects.
For each object, start a "Download" Go routine, as coded below, in this iteration.

yaron2:
@nelson-parente linter is failing

bindings/gcp/bucket/bucket.go
```diff
@@ -404,3 +411,90 @@ func (g *GCPStorage) signObject(bucket, object, ttl string) (string, error) {
...
+
+	wg.Wait()
+	close(errCh)
+
+	for err := range errCh {
+		if err != nil {
+			return nil, err
+		}
+	}
```
JoshVanL, line 497: All errors should be collected. Not return just one.

bindings/gcp/bucket/bucket.go
```diff
@@ -404,3 +411,90 @@ func (g *GCPStorage) signObject(bucket, object, ttl string) (string, error) {
...
+			if metadata.EncodeBase64 {
+				encoded := b64.StdEncoding.EncodeToString(data)
+				data = []byte(encoded)
+			}
+
+			//nolint:gosec
+			if err := os.WriteFile(destPath, data, 0o644); err != nil {
+				errCh <- err
+				return
+			}
```
JoshVanL, line 486: This data should be returned, no written to a file, no? How is the data useful if written to the `daprd` filesystem.

JoshVanL (changes requested):

bindings/gcp/bucket/bucket.go
nelson-parente: I got a bit confused with the ask in the issue.. As @JoshVanL points out this would write the files to daprd filesystem and not the app filesystem.

Refactored to return all objects data PTAL

yaron2 (approved):

PR #3825, opened 2025-05-20
feat: add auth metadata options and connectionstring builder

nelson-parente:
# Description

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


common/authentication/postgresql/metadata.go
```diff
@@ -87,6 +102,117 @@ func (m *PostgresAuthMetadata) InitWithMetadata(meta map[string]string, opts Ini
 	return nil
 }
 
+// buildConnectionString builds the connection string from the metadata.
+// It supports both DSN-style and URL-style connection strings.
+// Metadata fields override existing values in the connection string.
+func (m *PostgresAuthMetadata) buildConnectionString() (string, error) {
+	if m.ConnectionString == "" {
+		return "", errors.New("connection string is required")
+	}
```
mikeee, line 111: Is an error here intended? The below method `buildDSNConnectionString` appears to build a suitable connection string

common/authentication/postgresql/metadata.go
```diff
@@ -87,6 +102,117 @@ func (m *PostgresAuthMetadata) InitWithMetadata(meta map[string]string, opts Ini
...
+	metadata := m.getConnectionStringMetadata()
+
+	if strings.HasPrefix(m.ConnectionString, "postgres://") || strings.HasPrefix(m.ConnectionString, "postgresql://") {
+		return m.buildURLConnectionString(metadata)
+	}
+	return m.buildDSNConnectionString(metadata)
+}
+
+func (m *PostgresAuthMetadata) buildDSNConnectionString(metadata map[string]string) (string, error) {
+	connectionString := ""
```
mikeee, line 122: Is it designed intentionally not to allow for - examples sake, a connection string with a host and/or password to override?

common/authentication/postgresql/metadata.go
```diff
@@ -87,6 +102,117 @@ func (m *PostgresAuthMetadata) InitWithMetadata(meta map[string]string, opts Ini
 	return nil
 }
 
+// buildConnectionString builds the connection string from the metadata.
+// It supports both DSN-style and URL-style connection strings.
```
alicejgibbons, line 106:
```suggestion
// It supports both DNS-style and URL-style connection strings.
```

alicejgibbons:
So the logic here should be the broken out (potentially sensitive) values can be used along with values in the dns connection string. If the values are present in their own metadata property then they should overwrite the *same* value in the DNS connection string. Of course we hope the user won't provide them twice but the precedence should always be the individual value over the connection string one. Looking at the other potential connection string values, we should also add `hostaddr`, looking [here](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS) - this is c lib but assuming its similar.

We also don't have to support this in v1, only v2 for State.

common/authentication/postgresql/metadata_test.go
```diff
@@ -0,0 +1,100 @@
...
+			expectError: true,
+		},
+		{
+			name: "valid DSN connection string",
+			metadata: PostgresAuthMetadata{
+				ConnectionString: "host=localhost user=postgres password=example port=5432 connect_timeout=10 database=my_db",
+			},
+			expected:    "host=localhost user=postgres password=example port=5432 connect_timeout=10 database=my_db",
+			expectError: false,
+		},
```
mikeee, line 29:
```suggestion
		{
			name: "valid DSN connection string",
			metadata: PostgresAuthMetadata{
				ConnectionString: "host=localhost user=postgres password=example port=5432 connect_timeout=10 database=my_db",
			},
			expected:    "host=localhost user=postgres password=example port=5432 connect_timeout=10 database=my_db",
			expectError: false,
		},
		{
			name: "valid DSN connection values",
			metadata: PostgresAuthMetadata{
				Host:           "localhost",
				User: 		"postgres",
				Password:       "example",
				Port:           "5432",
				Database:       "my_db",
			},
			expected:    "host=localhost user=postgres password=example port=5432 database=my_db",
			expectError: false,
		},
```

I would expect this test case should also pass

common/authentication/postgresql/metadata.go
nelson-parente: isn't it DSN?

common/authentication/postgresql/metadata.go
nelson-parente: Initially, I didn't add all the props so you couldn't build a valid connection with metadata only. Since I've now added more props to the metadata we can allow to build only from meta.

common/authentication/postgresql/metadata_test.go
nelson-parente: with metadata only the order isn't preserved.. Added a specific test for this case :)

common/authentication/postgresql/metadata.go
nelson-parente: wdym? the goal is to start building from connection string if it exists, override matching params passed in metadata and then append all the remaining

common/authentication/postgresql/metadata.go
alicejgibbons: ah yeah ignore me

common/authentication/postgresql/metadata.go
mikeee: The changelog at the time was confusing, this looks fine to me now!

alicejgibbons (approved):
Logic lgtm, only im unsure if we want to continue to support v1, would confirm with the community

nelson-parente:
@yaron2 can we merge this? I don't think merge queue is working so we need manual merge.

marcduiker:
@holopin-bot @nelson-parente  Thank you! Here's a digital badge as a small token of appreciation.

PR #3835, opened 2025-05-30
Attempt at fixing a weird edge case where an error interrupts the backoff policy but the session context is still active...

passuied:
# Description

...and the retry policy is set to retry forever...
In some rare occasions, the message would be skipped despite the process being still active...
This code should ensure the message is being retried for as long as the session context is active if the retry policy is set to retry forever.

## Issue reference

#3529

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/common/component/kafka/consumer.go
+++ b/common/component/kafka/consumer.go
@@ -14,7 +14,6 @@ limitations under the License.
 package kafka
 
 import (
-	"context"
 	"errors"
 	"fmt"
 	"net/url"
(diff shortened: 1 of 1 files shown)


common/component/kafka/consumer.go
```diff
@@ -45,9 +44,9 @@ func notifyRecover(consumer *consumer, message *sarama.ConsumerMessage, session
 			// If the retry policy got interrupted, it could mean that either
 			// the policy has reached its maximum number of attempts or the context has been cancelled.
 			// There is a weird edge case where the error returned is a 'context canceled' error but the session.Context is not done.
-			// This is a workaround to handle that edge case and reprocess the current message.
-			if err == context.Canceled && session.Context().Err() == nil {
```
passuied, line 49: the adjusted fix is 2-fold:
- the `err == context.Canceled` check was a bit too specific and if the error is wrapped it may not match and therefore the condition would NOT be met
- More importantly, the recovery strategy is focusing more on the intent: If the policy is focused on infinite retry (in case zero data loss is paramount), if the session.Context() is NOT done, we would re-enter the loop regardless of the error...

passuied:
Better fix provided...

PR #3846, opened 2025-06-17
refactor langchaing go common models

filintod:
# Description

This is a refactor to put common langchaingo-based models to use a simple base struct that is embedded on each. This is part of adding other functionality like streaming and tool-calling that will benefit from a single point of change for all the models that use langchaingo as its base.

Along the way we updated the langchaingo version to deal with an error while testing Mistral.  We also found out that HuggingFace is not properly been executed in langchaingo and there is no fix. In the tests we used OpenAI compatibility with huggingface API URL but we might need to change huggingface to either use openai by default and/or fix langchaingo.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


conversation/langchaingokit/message.go
```diff
@@ -0,0 +1,36 @@
+package langchaingokit
```
nelson-parente, line 15: missing copyright on added files

conversation/langchaingokit/model.go
```diff
@@ -0,0 +1,39 @@
+package langchaingokit
```
cicoyle, line 15: Can you add the license header?

cicoyle (dismissed):

cicoyle (approved):

PR #3864, opened 2025-06-26
fix(mqtt): ensure proper locking when subscribing and unsubscribing topics

antontroshin:
# Description

MQTT3 component panic when unsubscribe (delete from map) is called concurrently.
Revealed during MQTT3 certification test [here](https://github.com/dapr/components-contrib/actions/runs/15887328906/job/44802408693?pr=3790)
```
fatal error: concurrent map writes

goroutine 14718 [running]:
internal/runtime/maps.fatal({0x23aa4f8?, 0xc0007c9798?})
	/home/runner/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.3.linux-amd64/src/runtime/panic.go:1058 +0x18
internal/runtime/maps.(*Map).Delete(0xc000a70060, 0x202dee0, 0xc0007c96e8)
	/home/runner/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.3.linux-amd64/src/internal/runtime/maps/map.go:648 +0x4c
github.com/dapr/components-contrib/pubsub/mqtt3.(*mqttPubSub).Subscribe.func1()
	/home/runner/work/components-contrib/components-contrib/pubsub/mqtt3/mqtt.go:183 +0x130
created by github.com/dapr/components-contrib/pubsub/mqtt3.(*mqttPubSub).Subscribe in goroutine 14682
	/home/runner/work/components-contrib/components-contrib/pubsub/mqtt3/mqtt.go:171 +0x5ff
```

### Solution decision
I've opted to use a separate `deletionLock sync.RWMutex` for this special case due to this issue happening only inside the goroutine when topics are being unsubscribed, when calling `delete` on a `map`.
Previous attempts were using existing `subscribingLock sync.RWMutex`, which is used in many different places, and from the tests, it seemed like it was substantially extending the tests, which caused a timeout after 30 minutes, while the usual run is about 3-4 minutes.

## Issue reference

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/.build-tools/go.mod
+++ b/.build-tools/go.mod
@@ -12,7 +12,7 @@ require (
 )
 
 require (
-	github.com/dapr/kit v0.15.3-0.20250516121556-bc7dc566c45d // indirect
+	github.com/dapr/kit v0.15.3-0.20250616160611-598b032bce69 // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/iancoleman/orderedmap v0.0.0-20190318233801-ac98e3ecb4b0 // indirect
 	github.com/inconshreveable/mousetrap v1.0.1 // indirect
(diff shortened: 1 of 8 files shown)


yaron2:
Thanks!

Issue #3674, opened 2025-02-22
Dapr unable to connect to Azure Service Bus Emulator


connorjose:
## Expected Behavior

Dapr's Azure Service Bus binding should be able to connect to the ASB emulator.

## Actual Behavior

Dapr reports that it is unable to subscribe to the ASB emulator topics.
`WARN[0031] failed to subscribe to topics: failed to subscribe to topic topic-one: could not get topic topic-one: Get "https://localhost:5672/topic-one?api-version=2021-05": EOF
failed to subscribe to topic topic-two: could not get topic topic-two: Get "https://localhost:5672/topic-two?api-version=2021-05": EOF`

## Steps to Reproduce the Problem

1. [Setup](https://learn.microsoft.com/en-us/azure/service-bus-messaging/test-locally-with-service-bus-emulator?tabs=docker-linux-container) ASB emulator using docker
2. Configure Dapr to connect to the emulator
```
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: example-pub-sub
spec:
  type: pubsub.azure.servicebus.topics
  version: v1
  metadata:
    - name: connectionString
      value: "Endpoint=sb://localhost:5672;SharedAccessKeyName=RootManageSharedAccessKey;SharedAccessKey=SAS_KEY_VALUE;UseDevelopmentEmulator=true;"
```

Dapr runtime version: 1.14.4

Tanzy:
I'm getting this issue as well, could it be due to the emulator using a self-signed certificate?

oising:
It looks like there's special handling for the emulator in the Azure SDK for Go that was added last year. I'm going to assume that DAPR is using an SDK older than that. 

https://github.com/Azure/azure-sdk-for-go/blame/674ac20511182bcf01023ec918d2c7aa88be44d5/sdk/messaging/azservicebus/internal/conn/conn.go#L102

/cc @yaron2

oising:
bump

adam6878:
Hi all — I spent some time digging into this issue and wanted to share what I found:

 The root cause of the EOF errors is that Dapr tries to perform entity management (like checking or creating topics/subscriptions) using the azservicebus/admin SDK, which makes HTTPS requests to the Service Bus namespace (e.g. https://localhost:5672/...).
When using the Azure Service Bus emulator, this fails because the emulator only exposes AMQP (port 5672) and does not support the Azure Resource Manager (HTTPS) APIs.

adam6878:
[...]

PR #3871, opened 2025-07-01
update connection string handling

adam6878:
# Description

The emulator does not support the Service Bus Management APIs (over HTTPS), which causes Dapr to fail during topic or subscription creation. With this change:

- `disableEntityManagement` is set to `true` automatically
- A helpful warning is logged to guide users
- Prevents confusing `EOF` or `connection refused` errors during topic checks

This allows developers to test pub/sub functionality locally with fewer manual steps and less confusion.

## Issue reference

Please reference the issue this PR will close: #3674  
(https://github.com/dapr/components-contrib/issues/3674)

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/common/component/azure/servicebus/metadata.go
+++ b/common/component/azure/servicebus/metadata.go
@@ -16,6 +16,7 @@ package servicebus
 import (
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	sbadmin "github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/admin"
@@ -186,6 +187,17 @@ func ParseMetadata(md map[string]string, logger logger.Logger, mode byte) (m *Me
 	if m.MaxRetriableErrorsPerSec < 0 {
 		err = errors.New("must not be negative")
 		return m, err
+	}
+
+	if strings.Contains(strings.ToLower(m.ConnectionString), "usedevelopmentemulator=true") {
+		if !m.DisableEntityManagement {
+			logger.Warn(
+				"UseDevelopmentEmulator=true detected in connection string. " +
+					"Azure emulator does not support topic management APIs. " +
+					"Dapr will skip admin operations. " +
+					"To suppress this warning, explicitly set disableEntityManagement: true.")
+		}
+		m.DisableEntityManagement = true
 	}
 
 	/* Nullable configuration settings - defaults will be set by the server. */


yaron2:
Thanks for the PR, unfortunately we can't turn the value to true by default because that's a breaking change for existing users

adam6878:
@yaron2 Got it. So how about this
In client.go, where we initialize the adminClient, we could first check if UseDevelopmentEmulator=true is present in the connection string.
If it is, and disableEntityManagement is not already set, we skip initializing the admin client and instead log a helpful warning — something like:

    “You're using the Azure Service Bus emulator, which doesn’t support management APIs. Please set disableEntityManagement: true or manually create the topics/subscriptions.”

common/component/azure/servicebus/metadata.go
```diff
@@ -188,6 +189,17 @@ func ParseMetadata(md map[string]string, logger logger.Logger, mode byte) (m *Me
...
+	if strings.Contains(strings.ToLower(m.ConnectionString), "usedevelopmentemulator=true") {
+		if !m.DisableEntityManagement {
+			logger.Warn(
+				"UseDevelopmentEmulator=true detected in connection string. " +
+					"Azure emulator does not support topic management APIs. " +
+					"Dapr will skip admin operations. " +
+					"To suppress this warning, explicitly set disableEntityManagement: true.")
+		}
+		m.DisableEntityManagement = true
+	}
```
nelson-parente: can we unit test this?

common/component/azure/servicebus/metadata.go
```diff
@@ -188,6 +189,17 @@ func ParseMetadata(md map[string]string, logger logger.Logger, mode byte) (m *Me
 		return m, err
 	}
 
+	if strings.Contains(strings.ToLower(m.ConnectionString), "usedevelopmentemulator=true") {
```
nelson-parente: nit: extract this logic to a func `isAzureEmulator`

common/component/azure/servicebus/metadata.go
```diff
@@ -188,6 +189,17 @@ func ParseMetadata(md map[string]string, logger logger.Logger, mode byte) (m *Me
...
+		if !m.DisableEntityManagement {
+			logger.Warn(
+				"UseDevelopmentEmulator=true detected in connection string. " +
+					"Azure emulator does not support topic management APIs. " +
+					"Dapr will skip admin operations. " +
+					"To suppress this warning, explicitly set disableEntityManagement: true.")
+		}
+		m.DisableEntityManagement = true
```
nelson-parente: This isn’t my area of expertise, but I want to challenge whether this is really needed given that users can already set `disableEntityManagement=true`

adam6878: Obviously, users can define DisableEntityManagement, but those who don't will encounter the EOF error when using the emulator. Setting DisableEntityManagement is the most straightforward way to avoid this issue.

common/component/azure/servicebus/metadata.go
adam6878: We could, but since the if logic was extracted into a separate function, there isn't much to test here. Also, there's currently no test file for metadata.go.

yaron2:
> @yaron2 Got it. So how about this In client.go, where we initialize the adminClient, we could first check if UseDevelopmentEmulator=true is present in the connection string. If it is, and disableEntityManagement is not already set, we skip initializing the admin client and instead log a helpful warning — something like:
> 
> ```
> “You're using the Azure Service Bus emulator, which doesn’t support management APIs. Please set disableEntityManagement: true or manually create the topics/subscriptions.”
> ```

Sounds good

yaron2:
@adam6878 ping

yaron2 (approved):

Issue #3858, opened 2025-06-23
[Kaka Pubsub] Ability to inhibit certain  publish metadata from being converted from/to Kafka headers


passuied:
## Describe the feature
Currently, any publish metadata gets automatically converted to kafka headers and similarly, any kafka headers gets automatically converted to metadata.
This is in general very helpful but can have some negative side effects.
For instance:
- publish metadata, converted to headers can then be forwarded and used as publish metadata to DLQ within a downstream consumer. 
- certain publish metadata may include sensitive data that shouldn't be leaked into messages' headers.

We should be able to specify a list of metadata items to exclude from this conversion as a optional kafka pubsub component metadata config. E.g.
- name: `HeaderToMetadataExcludedKeysRegex`
- value: `rawPayload|valueSchemaType`

## UAC:
  1. Conversion from metadata to header
    - Given a Kafka component set with `HeaderFromToMetadataExcludedKeysRegex` configured
    - When publishing a message with metadata keys matching the regex
    - Then these metadata items are NOT converted to Kaka headers in the published message.
  2. Conversion from header to metadata
    - Given a Kafka component set with `HeaderFromToMetadataExcludedKeysRegex` configured
    - When consuming a message with header keys matching the regex
    - Then these header items are NOT converted to metadata consumed message.

## Release Note
RELEASE NOTE: **ADD**  ability to exclude metadata items from being converted from/to kafka headers in kafka pubsub

PR #3874, opened 2025-07-02
Implement ability to exclude metadata items from being converted to/from headers

passuied:
# Description

Implement ability to exclude metadata items from being added as headers when publishing, and similarly from being converted from headers to metadata when consuming messages.

Forwarding certain metadata as headers has some unwanted downstream side effects.

## Issue reference

#3858

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#4701

### DIFF
--- a/bindings/kafka/metadata.yaml
+++ b/bindings/kafka/metadata.yaml
@@ -372,3 +372,10 @@ metadata:
       - "range"
       - "sticky"
       - "roundrobin"
+  - name: headerFromToMetadataExcludedKeysRegex
+    type: string
+    required: false
+    description: |
+      A regular expression to exclude keys from being converted to/from headers from/to metadata to avoid unwanted downstream side effects.
+    example: '"^rawPayload|valueSchemaType$"'
+    default: '""'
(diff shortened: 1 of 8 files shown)


bindings/kafka/metadata.yaml
```diff
@@ -372,3 +372,10 @@ metadata:
       - "range"
       - "sticky"
       - "roundrobin"
+  - name: headerFromToMetadataExcludedKeysRegex
```
nelson-parente: this is a big one 😅

NIT: shorten this name. Some suggestions: metaHeadersToIgnoreRegex, excludeMetaFromHeadersRegex, excludeHeaderKeysRegex

nelson-parente (commented):
Looks good! I think we should shorten the prop name

passuied:
@nelson-parente any suggestions? I struggled to find a name. 
Should I just drop From to and make it just `headerMetadataExcludedKeysRegex`?

passuied:
> @nelson-parente any suggestions? I struggled to find a name. Should I just drop From to and make it just `headerMetadataExcludedKeysRegex`?

@nelson-parente saw your suggestions. Pushed a change. Please re-review and re-run build

passuied:
@nelson-parente do you think we could push this one through before 1.16?

PR #3705, opened 2025-03-14
add conversation components to component metadata schema and bundle

famarting:
# Description

_Please explain the changes you've made_

I realized of this because the metadata bundle included in the 1.15 release does not contain conversation components

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #_[issue number]_

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #3882, opened 2025-07-09
add conversation components to component metadata schema and bundle

famarting:
# Description

opening https://github.com/dapr/components-contrib/pull/3705 again

we really need this merged... otherwise the metadata schema does not contain the conversation components

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/conversation/deepseek/metadata.yaml
+++ b/conversation/deepseek/metadata.yaml
@@ -25,5 +25,5 @@ metadata:
     required: false
     description: |
       Max tokens for each request
-    type: int
-    example: 2048
+    type: number
+    example: "2048"
--- a/.build-tools/component-folders.yaml
+++ b/.build-tools/component-folders.yaml
@@ -9,6 +9,7 @@ componentFolders:
   - secretstores
   - state
   - workflows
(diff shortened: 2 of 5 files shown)


conversation/deepseek/metadata.yaml
```diff
@@ -25,5 +25,5 @@ metadata:
     required: false
     description: |
       Max tokens for each request
-    type: int
-    example: 2048
+    type: number
```
nelson-parente, line 28: why did this change?

nelson-parente (approved):

conversation/deepseek/metadata.yaml
nelson-parente: should it change in other conversation components

famarting: Wil check, but it was the only place that the tooling was complaining about, and after this the metadata bundle now builds fine

famarting: doublechecked, its the only place where this wrong type was being used

Issue #3915, opened 2025-07-25
Add Azure OpenAI support to the Conversation API


MyMirelHub:
**Problem:** The conversation API only supports standard OpenAI endpoints through `conversation.openai`. Enterprise users with Azure OpenAI Service cannot use the conversation API because Azure OpenAI uses different endpoint formats (`https://{resource}.openai.azure.com/`) and you need to refer to the deployment name rather than the underlying model name in API calls

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: azure-openai
spec:
  type: conversation.azure.openai
  version: v1
  metadata:
  - name: endpoint
    value: "https://myopenai.openai.azure.com"
  - name: deploymentId
    value: "gpt-4-turbo"
  - name: apiKey
    value: "your-key"
```

Dapr already supports a Azure Open AI bindings `bindings.azure.openai` component. So it can perhaps leverage for setup and auth patterns. 

## Release Note
RELEASE NOTE: ADD Azure OpenAI conversation component.

fabistb:
@MyMirelHub  thank you very much for opening the issue.
After digging a little bit I think we can enhance the current existing component to also support Azure OpenAi.

So the existing component could be enhanced with additional parameters based on which we could decided if it is the standard open ai or azure.

It should look something like this.
```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: openai
spec:
  type: conversation.openai
  metadata:
  - name: key
    value: mykey
  - name: model
    value: gpt-4-turbo
  - name: endpoint
    value: 'https://api.openai.com/v1'
  - name: cacheTTL
    value: 10m
- name: apiType
   value: openAi # azure, azureAd
- name: apiVersion
   value: 2025-01-01-preview # optional only required when apiType is azure / azureAd
```

PR #3918, opened 2025-07-29
Enhance OpenAI to be Azure OpenAi compatible

fabistb:
# Description
Enhance the existing `OpenAi` conversation api component to also support `Azure OpenAi`.
Introduces two new optional metadata parameters:
- apiType  --> Defines  if `openai` or `azure` is used. If empty `openai` is set as the default value
- apiVersion --> Defines the apiVersion. If `apiType` is set to `azure` this parameter becomes required.

## Issue reference

Please reference the issue this PR will close: #3915

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: [dapr/docs#_[4749]_](https://github.com/dapr/docs/issues/4749)


conversation/openai/metadata.go
```diff
@@ -0,0 +1,26 @@
+/*
+Copyright 2024 The Dapr Authors
```
yaron2: nit: 2025

yaron2:
@filintod These changes look fine to me, calling you for review

conversation/openai/metadata.go
fabistb: fixed.

conversation/openai/metadata.go
```diff
@@ -0,0 +1,26 @@
...
+package openai
+
+// OpenAILangchainMetadata extends LangchainMetadata with OpenAI-specific properties.
+type OpenAILangchainMetadata struct {
+	Key        string `json:"key"`
+	Model      string `json:"model"`
+	CacheTTL   string `json:"cacheTTL"`
+	Endpoint   string `json:"endpoint"`
+	APIType    string `json:"apiType"`
+	APIVersion string `json:"apiVersion"`
```
sicoyle:
```suggestion
type OpenAILangchainMetadata struct {
	LangchainMetadata
	APIType    string `json:"apiType"`
	APIVersion string `json:"apiVersion"`
```
can we embed instead of copy/paste the same fields over?

conversation/openai/metadata.yaml
```diff
@@ -40,3 +40,17 @@ metadata:
       A time-to-live value for a prompt cache to expire. Uses Golang durations
     type: string
     example: '10m'
+  - name: apiVersion
+    required: false
+    description: |
+      The API version to use for the Azure OpenAI service. This is required when using Azure OpenAI.
+    type: string
+    example: '2025-01-01-preview'
+    default: '2025-01-01-preview'
```
sicoyle:
```suggestion
    default: ''
```

conversation/openai/openai.go
```diff
@@ -65,6 +67,31 @@ func (o *OpenAI) Init(ctx context.Context, meta conversation.Metadata) error {
...
+	if md.APIType == "azure" && md.APIVersion == "" {
+		return errors.New("apiVersion must be provided when apiType is set to 'azure'")
+	}
+
+	// Return error when api Type is azure but the endpoint is not provided
+	if md.APIType == "azure" && md.Endpoint == "" {
+		return errors.New("endpoint must be provided when apiType is set to 'azure'")
+	}
+
+	// Set api version if provided
```
sicoyle: can we rm the comments pls as the code is very clear :)

filintod:
@fabistb thanks, could you run the conformance test and run locally (we don't have API keys in CI) to make sure everything is fine when running it.  Probably the docs for adding a new one are not there just looking at it now:

- Check https://github.com/dapr/components-contrib/tree/main/tests/config/conversation/README.md (it will tell you about adding api keys to a .env file that is gitignored)

- Update https://github.com/dapr/components-contrib/blob/main/tests/config/conversation/openai/openai.yml adding endpoint (as it looks like it is not there), apiVersion and apiType and with the values templated to `${OPENAI_END_POINT}`, `${OPENAI_API_VERSION}` and `${OPENAI_API_TYPE}` for example (or something nicer), add these variables to the .env file too or export them.  Another options is create a completely new component but this should actually work for all openai-compatible LLMs

- Run and Debug the conformance test as mentioned in the README.md above

- There is the DCO issue, follow the [error link ](https://github.com/dapr/components-contrib/pull/3918/checks?check_run_id=47061007853)to correct it. ALl commits needed to be signed-off (-s in git commit).

- Please update the conformance test README if you see issues following it.

- It might be good to check with a regular openai key (no endpoint/apiType) to make sure this does not bring a regression (again mainly because we don't have CI with real api tokens to assure us, that we probably need)

conversation/openai/openai.go
```diff
@@ -65,6 +67,31 @@ func (o *OpenAI) Init(ctx context.Context, meta conversation.Metadata) error {
 		options = append(options, openai.WithBaseURL(md.Endpoint))
 	}
 
+	// Identify correct API Type
+	switch md.APIType {
+	case "", "openai":
+		options = append(options, openai.WithAPIType(defaultAPIType))
```
sicoyle: did you confirm this works for regular openai? If a user puts "openai" and this option gets set does it still work? This is changing existing default behavior where this field would be empty string which means an api type option never gets set. Please update to make that the default case. I don't think we need an err here.

fabistb:
> @fabistb thanks, could you run the conformance test and run locally (we don't have API keys in CI) to make sure everything is fine when running it. Probably the docs for adding a new one are not there just looking at it now:
> 
> * Check https://github.com/dapr/components-contrib/tree/main/tests/config/conversation/README.md (it will tell you about adding api keys to a .env file that is gitignored)
> * Update https://github.com/dapr/components-contrib/blob/main/tests/config/conversation/openai/openai.yml adding endpoint (as it looks like it is not there), apiVersion and apiType and with the values templated to `${OPENAI_END_POINT}`, `${OPENAI_API_VERSION}` and `${OPENAI_API_TYPE}` for example (or something nicer), add these variables to the .env file too or export them.  Another options is create a completely new component but this should actually work for all openai-compatible LLMs
> * Run and Debug the conformance test as mentioned in the README.md above
> * There is the DCO issue, follow the [error link ](https://github.com/dapr/components-contrib/pull/3918/checks?check_run_id=47061007853)to correct it. ALl commits needed to be signed-off (-s in git commit).
> * Please update the conformance test README if you see issues following it.
> * It might be good to check with a regular openai key (no endpoint/apiType) to make sure this does not bring a regression (again mainly because we don't have CI with real api tokens to assure us, that we probably need)

@filintod  thanks for the insights.
Regarding the conformance tests. I used now the same approach which is already in place for other components like the configuration postgresql test. So I setup a specific test for _Azure OpenAI_ which contains the minimal metadata properties. For me it looks like the tests should always only contain the minimal required metadata configuration for one scenario e.g. _OpenAi_ / _Azure OpenAI_.

Hope that works and that should also make it easier to find potential regressions for the two different scenarios.

For the DCO topic, that should also be fixed now.

conversation/openai/metadata.yaml
fabistb: fixed.

conversation/openai/openai.go
fabistb: I check and it was working. This was just explicitly setting the default value as per method documentation:
image

Checking only for `azure` however is probably a better solution.
It doesn't change the existing behavior and doesn't add unnecessary code.

conversation/openai/openai.go
fabistb: done. Removed the comments.

conversation/openai/metadata.go
fabistb: That should work now.

conversation/openai/openai.go
```diff
@@ -65,6 +66,22 @@ func (o *OpenAI) Init(ctx context.Context, meta conversation.Metadata) error {
...
+		options = append(options, openai.WithAPIType(openai.APITypeAzure))
+	}
+
+	if md.APIType == "azure" && md.APIVersion == "" {
+		return errors.New("apiVersion must be provided when apiType is set to 'azure'")
+	}
+
+	if md.APIType == "azure" && md.Endpoint == "" {
+		return errors.New("endpoint must be provided when apiType is set to 'azure'")
+	}
```
filintod: These three could be coalesced (or appended to multiple errors in one, so one does not get hit with two consecutive failures if missing one or the other without knowing)

```suggestion
	if md.APIType == "azure" {
		options = append(options, openai.WithAPIType(openai.APITypeAzure))
		if md.APIVersion == "" || md.Endpoint == ""{
			return errors.New("apiVersion and endPoint must be provided when apiType is set to 'azure'")
		}
	}
```

Also, we do set default models, so maybe you could add a sensible default [apiVersion](https://learn.microsoft.com/en-us/azure/ai-foundry/openai/api-version-lifecycle?tabs=key) when `azure` is the APIType if it makes sense for quick starting using the component. But this is just if you think it makes sense and is general enough (not different for each azure client), endpoint is most likely different for everyone.

But leaving Endpoint as required (without default) is not bad either.

fabistb: I think for a default value the most sensible thing would probably be using the latest GA version.
Setting some preview version doesn't feel right and the error when a model isn't supported is clear.
image

But I'm wondering how do deal with this default value over the lifetime.
How would the update policy look like? I think making it mandatory to set for the user could avoid some issues here in the long run.

I changed now only the error behavior.

filintod: thanks. just to confirm you were you able to test with the real api keys plain openai and azure openai, correct?

fabistb: Yes, tested and received a valid response from openai and azure openai.

conversation/openai/openai.go
```diff
@@ -65,6 +66,15 @@ func (o *OpenAI) Init(ctx context.Context, meta conversation.Metadata) error {
 		options = append(options, openai.WithBaseURL(md.Endpoint))
 	}
 
+	if md.APIType == "azure" {
+		options = append(options, openai.WithAPIType(openai.APITypeAzure))
+
+		if md.Endpoint == "" || md.APIVersion == "" {
+			return errors.New("endpoint and apiVersion must be provided when apiType is set to 'azure'")
+		}
+		options = append(options, openai.WithAPIVersion(md.APIVersion))
```
filintod: LGTM. Final nit suggestion to put the appends together after the validation

```suggestion
		if md.Endpoint == "" || md.APIVersion == "" {
			return errors.New("endpoint and apiVersion must be provided when apiType is set to 'azure'")
		}

		options = append(options, openai.WithAPIType(openai.APITypeAzure))
		options = append(options, openai.WithAPIVersion(md.APIVersion))
```

fabistb: done.

yaron2 (approved):

marcduiker:
@holopin-bot @fabistb Thank you Fabian! Here's a digital badge as a small token of appreciation.

PR #3925, opened 2025-08-01
Fix default partition key for single-transaction operations in Cosmos

acroca:
# Description

There's a bug in Cosmos where when no partition key is configured, single operations default to use the same as the `key` but multi operations default to `""`.

Given this discrepancy in the defaults, it's impossible to operate on the same key with single and transactional operations, so a change in the defaults is necessary.

NOTICE this is a breaking change, transactional operations will now have a different default partition than before, so keys created in previous versions will not be accessible anymore because it'll now use the first key as a partition instead of `""`. We could also change the singe operations so use `""` by default, but this would also introduce a breaking change.

# Caveat

The individual `Set` operation defaults to use the same partition as the `key`, so if the user doesn't specify a partition all keys will live in it's own partition.
Cosmos transaction operations need to operate in the same partition.
If a transaction is created across multiple keys it's required for those keys to live in the same partition, and it's an scenario that was already unsupported before.
In the scenario where we need to update multiple keys, we need the user to make sure the partitions are correct.

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/state/azure/cosmosdb/cosmosdb.go
+++ b/state/azure/cosmosdb/cosmosdb.go
@@ -503,7 +503,14 @@ func (c *StateStore) Multi(ctx context.Context, request *state.TransactionalStat
 		return nil
 	}
 
-	partitionKey := request.Metadata[metadataPartitionKey]
+	partitionKey, partitionKeySet := request.Metadata[metadataPartitionKey]
+	if !partitionKeySet {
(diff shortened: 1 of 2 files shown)


sicoyle:
can we really make a "breaking change" to a stable component? https://docs.dapr.io/reference/components-reference/supported-state-stores/

PR #3930, opened 2025-08-01
Conversation Tool Calling fix on echo and other minor fixes

filintod:
# Description

The echo provider was responding to the Tool Response with a Tool Call when it should not had.

Added default endpoint for deepseek that was not correct

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


tests/config/conversation/huggingface/huggingface.yml
```diff
@@ -9,4 +9,4 @@ spec:
     - name: key
       value: "${{HUGGINGFACE_API_KEY}}"
     - name: model
-      value: "deepseek-ai/DeepSeek-R1-Distill-Qwen-32B" 
\ No newline at end of file
+      value: "HuggingFaceTB/SmolLM3-3B"
```
filintod, line 12: the other model was returning 404 now

tests/conformance/conversation/conversation.go
```diff
@@ -46,6 +46,8 @@ func NewTestConfig(componentName string) TestConfig {
 }
 
 func ConformanceTests(t *testing.T, props map[string]string, conv conversation.Conversation, component string) {
+	var finishStopReasonOfProviders = []string{"stop", "end_turn", "FinishReasonStop"}
```
filintod: gemini stop message added

conversation/echo/echo.go
```diff
@@ -85,15 +85,7 @@ func (e *Echo) Converse(ctx context.Context, r *conversation.Request) (res *conv
...
-				content = p.Content
-				toolCalls = append(toolCalls, llms.ToolCall{
-					ID:   p.ToolCallID,
-					Type: "function",
-					FunctionCall: &llms.FunctionCall{
-						Name:      p.Name,
-						Arguments: p.Content,
-					},
-				})
+				content += fmt.Sprintf("Tool Response for tool '%s' with ID '%s': %s", p.Name, p.ToolCallID, p.Content)
```
sicoyle: can you swap id and name around so id is first pls

tests/conformance/conversation/conversation.go
sicoyle:
```suggestion
	var providerStopReasons = []string{"stop", "end_turn", "FinishReasonStop"}
```

yaron2 (approved):

sicoyle:
please don't merge this until the mistral helper is updated based on feedback here https://github.com/dapr/dapr/pull/8937#discussion_r2248204455

conversation/mistral/mistral.go
```diff
@@ -110,13 +111,34 @@ func CreateToolCallPart(toolCall *llms.ToolCall) llms.ContentPart {
...
+
+	mistralContentParts := make([]string, 0, len(responses))
+	for _, response := range responses {
+		if resp, ok := response.(llms.ToolCallResponse); ok {
+			if toolID == "" {
+				toolID = resp.ToolCallID
+			}
+			if name == "" {
+				name = resp.Name
+			}
```
sicoyle: im confused - you set toolID and name to empty string but then you check if they are empty string to assign them?

filintod: well it is a loop, so assign the first time, that should be the only time required but just in case

cicoyle: Can you not just have `var toolID string`, then in the loop `toolID = resp.ToolCallID`? and the same for `name`? Seems very redundant to set it to `""` then check that its an empty string to assign it. Or are you meaning to check the resp and if `resp.ToolCallID != nil or ""` then assign? 

Either way, I believe we need to tweak this logic unless Im misunderstanding something.

conversation/echo/echo.go
```diff
@@ -61,67 +63,96 @@ func (e *Echo) GetComponentMetadata() (metadataInfo metadata.MetadataMap) {
...
+	// if we get tools, respond with tool calls for each tool
+	var toolCalls []llms.ToolCall
+	if r.Tools != nil && len(*r.Tools) > 0 {
+		// create tool calls for each tool
+		toolCalls = make([]llms.ToolCall, 0, len(*r.Tools))
+		for id, tool := range *r.Tools {
+			// extract argument names from parameters.properties
+			if tool.Function == nil || tool.Function.Parameters == nil {
+				continue // skip if no function or parameters
+			}
```
sicoyle: you cannot do this as sometimes a function will not have parameters but it would still be a tool call.

conversation/echo/echo.go
```diff
@@ -61,67 +63,96 @@ func (e *Echo) GetComponentMetadata() (metadataInfo metadata.MetadataMap) {
...
 		return &conversation.Response{
 			ConversationContext: r.ConversationContext,
 			Outputs:             []conversation.Result{},
 		}, nil
 	}
 
-	outputs := make([]conversation.Result, 0, len(*r.Message))
+	// if we get tools, respond with tool calls for each tool
+	var toolCalls []llms.ToolCall
+	if r.Tools != nil && len(*r.Tools) > 0 {
```
sicoyle: redundant. Pls check if r.Tools is nil and thats it

conversation/echo/echo.go
```diff
@@ -61,67 +63,96 @@ func (e *Echo) GetComponentMetadata() (metadataInfo metadata.MetadataMap) {
...
+				continue // skip if no function or parameters
+			}
+			// ensure parameters are a map
+			m, ok := tool.Function.Parameters.(map[string]interface{})
+			if !ok {
+				return nil, fmt.Errorf("tool function parameters must be a map, got %T", tool.Function.Parameters)
+			}
+			if len(m) == 0 {
+				return nil, fmt.Errorf("tool function parameters map cannot be empty for tool ID %d", id)
+			}
```
sicoyle: this can be removed as it is redundant to the check below pretty much, and also parameters can be empty on tools from what I saw

conversation/echo/echo.go
```diff
@@ -61,67 +63,96 @@ func (e *Echo) GetComponentMetadata() (metadataInfo metadata.MetadataMap) {
...
+			if !ok {
+				continue // skip if no properties
+			}
+			propMap, ok := properties.(map[string]interface{})
+			if !ok {
+				return nil, fmt.Errorf("tool function properties must be a map, got %T", properties)
+			}
+			if len(propMap) == 0 {
+				continue // skip if no properties
+			}
```
sicoyle: rm all of the properties check stuff pls. We should not err in this case. A tool call function can have no parameters and still be valid

conversation/echo/echo.go
```diff
@@ -61,67 +63,96 @@ func (e *Echo) GetComponentMetadata() (metadataInfo metadata.MetadataMap) {
...
+			// get argument names
+			argNames := make([]string, 0, len(propMap))
+			for argName := range propMap {
+				argNames = append(argNames, argName)
+			}
+			toolCalls = append(toolCalls, llms.ToolCall{
+				ID: strconv.Itoa(id),
+				FunctionCall: &llms.FunctionCall{
+					Name:      tool.Function.Name,
+					Arguments: strings.Join(argNames, ","),
```
sicoyle, line 113: this would make the last argument end in a comma which we do not want

conversation/echo/echo.go
```diff
@@ -61,67 +63,96 @@ func (e *Echo) GetComponentMetadata() (metadataInfo metadata.MetadataMap) {
...
-		for i, part := range message.Parts {
+		for _, part := range message.Parts {
 			switch p := part.(type) {
 			case llms.TextContent:
 				// end with space if not the first part
-				if i > 0 && content != "" {
-					content += " "
-				}
-				content += p.Text
+				contentFromMessaged = append(contentFromMessaged, p.Text)
```
sicoyle: if you are changing the code to just append then pls update the corresponding comments

conversation/echo/echo.go
sicoyle: can you just do something like

```
 if paramsMap, ok := tool.Function.Parameters.(map[string]interface{}); ok {
        if propsRaw, ok := paramsMap["properties"]; ok {
            if propsMap, ok := propsRaw.(map[string]interface{}); ok {
                argNames = make([]string, 0, len(propsMap))
                for argName := range propsMap {
                    argNames = append(argNames, argName)
                }
            }
        }
    }
```

then append the tool call pls? to simplify things

conversation/mistral/mistral.go
sicoyle: there should be no "just in case" imo. We don't do anything special or skip these values if they are empty string so then why even make a change on this part of the function? Sure, add the new line on the end of the content, but the rest of the changes here are unnecessary as we still print out the toolID and/or name if they are empty string

conversation/echo/echo.go
filintod: true, will remove the second side of `or` stmt

conversation/echo/echo.go
filintod: we can have a check for empy/nill and skip the type check but the type check should be there if not empty

conversation/echo/echo.go
filintod: no, strings.Join does not add a comma at the end, not sure what you mean

conversation/mistral/mistral.go
filintod: @cicoyle it is usually more idiomatic/concised to use empty assignment but I guess I could do `var toolID, name string` to keep it concise.

@sicoyle the changes are required to merged multiple parts as text via append, if they are provided, more than just appending a new line at the end. Hopefully this method can be removed in the future with better Mistral support in langchaingo. 

I could raise an error if toolID and name are empty as I guess it might not help the LLM not knowing what tool response it correspond too, maybe it could infer if it is only one though. Or if change the init state of them to `n/a`. Let me know

conversation/echo/echo.go
filintod: not completely redundant as with it all statements can be skip including the toolCalls map creation, but I guess this being only for tests is ok

conversation/echo/echo.go
```diff
@@ -61,67 +63,97 @@ func (e *Echo) GetComponentMetadata() (metadataInfo metadata.MetadataMap) {
...
+	for _, message := range *r.Message {
+		for _, part := range message.Parts {
 			switch p := part.(type) {
 			case llms.TextContent:
 				// end with space if not the first part
-				if i > 0 && content != "" {
-					content += " "
-				}
-				content += p.Text
+				contentFromMessaged = append(contentFromMessaged, p.Text)
```
sicoyle: the comment on line 124 is still there but should not be now with these changes... this one
```
// end with space if not the first part
```

sicoyle (approved):

yaron2:
Linter

yaron2:
Relevant unit test is failing

filintod:
@yaron2 can you retry the tests again

yaron2 (approved):

Issue #3934, opened 2025-08-05
ttlInSeconds not respected when publishing to Azure Service Bus Topic


aladd04:
## Expected Behavior

Azure Service Bus messages published via the runtime are not respecting the `ttlInSeconds` metadata key, even though the docs state that it should [here](https://docs.dapr.io/developing-applications/building-blocks/pubsub/pubsub-message-ttl/).

Note I first submitted this bug over in the `dotnet-sdk` [here](https://github.com/dapr/dotnet-sdk/issues/1590). But was redirected this way after confirming the behavior without using the dotnet SDK.

## Actual Behavior

A "custom property" key-value pair gets set on the message in Service Bus with your value, but the message never actually falls off or moves to the dead letter queue, as indicated by the message's "expiresAtUtc".

## Steps to Reproduce the Problem

I'm using Dapr CLI version `1.15.1` and runtime version `1.15.4`. This seems to behave the same way in self-hosted mode and when deployed to kubernetes.

Assuming some dapr config like below:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: mypubsub
spec:
  type: pubsub.azure.servicebus.topics
  version: v1
  metadata:
    - name: connectionString
      value: ""
    - name: maxConcurrentHandlers
      value: 1
    - name: maxActiveMessages
      value: 10000
    - name: defaultMessageTimeToLiveInSec
      value: 43200 # 12 hours
    - name: consumerID
      value: "{appID}"
```

Run the CLI command as recommend in [these docs](https://docs.dapr.io/developing-applications/building-blocks/pubsub/pubsub-message-ttl/#example).

`curl -X "POST" "http://localhost:3500/v1.0/publish/mypubsub/my-topic?metadata.ttlInSeconds=120" -H "Content-Type: application/json" -d '{"order-number": "345"}'`

Then go check your service bus. You will see that the "Message Properties" of the message has a "Custom Property" for "ttlInSeconds" with a value of "120". But the actual "Broker Properties" -> "expiresAtUtc" value either uses the default value configured for that subscription, or it will be the max value assigned if set to "unbounded" (no TTL).

## Release Note

RELEASE NOTE: **FIX** ttlInSeconds not respected when publishing messages to Azure Service Bus.

aladd04:
I did a little extra testing and found a workaround. If you set the metadata key `ttl` instead of `ttlInSeconds` it looks like messages appropriately expire! And it seems you can set the value in seconds still.

[...]

PR #3942, opened 2025-08-07
fix: 'ttlInSeconds' metadata key now respected when publishing to Azure Service Bus Topics.

aladd04:
# Description

The metadata key `ttlInSeconds` was being ignored when publishing messages to Azure Service Bus Topics. The fix was adding an alternative TTL constant to a switch statement.

[This line here](https://github.com/dapr/components-contrib/blob/20d73459f7a089704d9cc4cfe555bdaa450f43df/common/component/azure/servicebus/message.go#L125) was the culprit.

The switch statement was only referencing the TTL metadata constant `mdutils.TTLMetadataKey`, so I modified it to also include the other TTL metadata key constant too, `mdutils.TTLInSecondsMetadataKey`.

## Issue reference

Issue this PR will close: #3934

## Checklist

* [x] Code compiles correctly


aladd04:
Not sure why the `pubsub.kafka` certification tests failed? I'd think the `message.go` changes under `/common/component/azure/servicebus` wouldn't affect that at all. I'm not very familiar with the tests process here if anyone has any ideas.

sicoyle:
> Not sure why the pubsub.kafka certification tests failed? I'd think the message.go changes under /common/component/azure/servicebus wouldn't affect that at all. I'm not very familiar with the tests process here if anyone has any ideas.

Thanks for your contribution. Yeah, you can ignore the unrelated test failure. We're looking into those :)

sicoyle (approved):

marcduiker:
@holopin-bot @aladd04 Thank you Andy! Here's a digital badge as a small token of appreciation.

Issue #3948, opened 2025-08-07
automate release task of verifying component registration


sicoyle:
## Describe the feature

For each dapr release there is an endgame task to ensure that components are properly registered across contrib and dapr/dapr. This is something that should be automated and ran for every release as there are so many components here that it is easy to let some slip through the cracks. We need a script for:

Goal: Ensure that all new components are registered properly in runtime.

Checks:
- [ ] Documentation for each component type in docs repo
- [x] Metadata schema in contrib
    - [X] DONE MANUALLY This metadata.yaml file must follow the schema defined in component-metadata-schema.json
    - [X] DONE MANUALLY Must include required fields for: schemaVersion, type, name, version, status, title, urls
    - [X] DONE MANUALLY Must define authentication profiles and metadata options
    - [x] Ensure each component has a metadata.yaml file in components contrib
- [x] CHECKED AT COMPILE TIME SO NO NEED TO CHECK FURTHER Component implementation in contrib:
    - [ ] Component must implement the correct interface
    - [ ] Component must have a New constructor that takes a logger and returns the interface
    - [ ] Component must have Init, Close, and other required methods (which else?)
- [x] Registration file in dapr/dapr
    - [x] IMPLICITLY DONE File must import the component from components-contrib
    - [x] File must call {Type}.DefaultRegistry.RegisterComponent({Component}.New{Component}, “{name}”) in an init() function
    - [x] File must have //go:build all components
    - [x] Each component needs a registration file in dapr/cmd/daprd/components/{type}_{component}.go
- [x] Component type has a registry in dapr/dapr:
    - [x] Registry must have a Create method for instantiation
    - [x] Registry must have RegisterComponent method
    - [x] The registry must have a DefaultRegistry singleton
    - [x] Each component type needs a registry.go

## Release Note

RELEASE NOTE:
Add: CLI to ensure component registration within runtime

PR #3949, opened 2025-08-07
Endgame task 1.16 comp registration

sicoyle:
# Description

For each dapr release there is an endgame task to ensure that components are properly registered across contrib and dapr/dapr. I took a bit of time to create a new CLI with the other CLI cmds under `/build-tools` to automate this for myself (assigned this task), as well as for others. I also took the time to fix all the issues that I came across, cross referencing the docs. I have a list of components that need to be added to the docs. This took longer than one would hope; however, I noticed several  components not registered properly in runtime, or in the docs, and several component types with no metadata.yaml files at all. This PR is a first stab at fully automating this, and does not hit every check box on my issue I created below, but again, is a great starting point and can be iterated on for future releases.

## Issue reference

https://github.com/dapr/components-contrib/issues/3948

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


bindings/gcp/pubsub/pubsub.go
```diff
@@ -46,19 +46,20 @@ type GCPPubSub struct {
...
+	Type                    string `json:"type"`
+	ProjectID               string `json:"projectID"`
+	PrivateKeyID            string `json:"privateKeyID"`
+	PrivateKey              string `json:"privateKey"`
+	ClientEmail             string `json:"clientEmail"`
+	ClientID                string `json:"clientID"`
+	AuthURI                 string `json:"authURI"`
+	TokenURI                string `json:"tokenURI"`
+	AuthProviderX509CertURL string `json:"authProviderX509CertURL"`
+	ClientX509CertURL       string `json:"clientX509CertURL"`
```
sicoyle: pls note that i did update these intentionally. This is specific to bindings gcp pubsub which is not stable or listed as any component type from what I saw. I updated this to match the existing stable gcp pubsub component type and so that way i can use the builtin gcp auth provider properly.

WhitWaldo:
I wholly support this PR in its effort to ensure that the components-metadata-bundle.json file reflects accurate descriptions of what each of the supported components expect with all the schema validation.

Further, if this is something we can represent as a script that keeps these metadata outputs consistently up to date as part of the repository build/release pipeline, that would be fantastic for ensuring component parity with each tagged release!

bindings/alicloud/dingtalk/webhook/metadata.yaml
```diff
@@ -0,0 +1,31 @@
...
+    description: "The webhook ID"
+    example: '"your-webhook-id"'
+  - name: url
+    required: true
+    description: "The webhook URL"
+    example: '"https://oapi.dingtalk.com/robot/send?access_token=your-token"'
+  - name: secret
+    required: false
+    description: "The webhook secret for signature verification"
+    example: '"your-webhook-secret"'
```
JoshVanL: All files should include a new line at EOF

bindings/gcp/pubsub/pubsub.go
```diff
@@ -69,32 +89,74 @@ func NewGCPPubSub(logger logger.Logger) bindings.InputOutputBinding {
...
 
-func (g *GCPPubSub) parseMetadata(metadata bindings.Metadata) ([]byte, error) {
-	return json.Marshal(metadata.Properties)
+func (g *GCPPubSub) getPubSubClient(_ context.Context, metadata *pubSubMetadata) (*pubsub.Client, error) {
+	var pubsubClient *pubsub.Client
+	var err error
+
+	if metadata.PrivateKeyID != "" {
+		// TODO: validate that all auth json fields are filled
+		authJSON := &GCPAuthJSON{
```
JoshVanL: Where has this code come from?

state/zookeeper/zk.go
```diff
@@ -32,6 +32,7 @@ import (
 	"github.com/dapr/kit/ptr"
 )
 
+// TODO: I think we need more defaults set on these metadata fields.
```
JoshVanL: Why this TODO?

JoshVanL (changes requested):
There are a large number of changes in this PR spanning multiple groups and concepts which makes it tricky to review. I this makes it higher risk that something is added which is incorrect or breaks something.

It has been a long standing issue that the `metadata.yaml` files should be generated, however I think this should come from a generi  code type reflector which appropriate comment tag decoration of the input metadata.

As it stands, I don't want to move forward this all these additions where we rely on humans creating these files manually. Please can we at least split this PR into multiple PRs where each target a different change/`metadata.yaml` addition.

bindings/gcp/pubsub/pubsub.go
sicoyle: https://github.com/dapr/components-contrib/pull/3949#discussion_r2261532882

state/zookeeper/zk.go
sicoyle: I am open to suggestions on how to make this comment clearer. 

As an approver in this repo, I take ownership seriously, and I use TODO comments intentionally to highlight areas that need improvement but are outside the current scope of my changes.

In this case, the comment came from a thorough review of each metadata.yaml file I personally touched in this PR. I noticed that many fields — particularly in alpha/beta components — are missing default values that would improve clarity and usability. While addressing that across the board felt out of scope for this specific task, I still wanted to surface it as something worth revisiting.

Happy to reword the comment or track it elsewhere if you have a suggestion

sicoyle:
> There are a large number of changes in this PR spanning multiple groups and concepts which makes it tricky to review. I this makes it higher risk that something is added which is incorrect or breaks something.

> It has been a long standing issue that the metadata.yaml files should be generated, however I think this should come from a generi code type reflector which appropriate comment tag decoration of the input metadata.

> As it stands, I don't want to move forward this all these additions where we rely on humans creating these files manually. Please can we at least split this PR into multiple PRs where each target a different change/metadata.yaml addition.

I split this effort out across component types for each PR :)

sicoyle:
ready for review pls

yaron2 (approved):

PR #3950, opened 2025-08-11
feat: add binding metadata files plus fixes

sicoyle:
# Description

Separating the bindings changes from PR: 

For each dapr release there is an endgame task to ensure that components are properly registered across contrib and dapr/dapr. I took the time to fix all the issues that I came across, cross referencing the docs. This PR updates naming conventions on file directories, adds missing metadata.yaml files, and updates things like the gcp pubsub binding (alpha) to support the same logic as the stable pubsub gcp pubsub component so I could standardize their auth profiles in adding the metadata.yaml file and use the builtinAuthProfile for gcp.

You will also see TODO comments here and there. That is from me going through the code to manually create the metadata.yaml files and add todos i deemed necessary for the component. I also updated any `interface{}` to `any` that I came across along the way.

Yes, I did cross ref the docs too so the links should be 99.9% correct unless I made a typo or missed something accidentally.

## Issue reference

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests


bindings/gcp/pubsub/pubsub.go
```diff
@@ -46,19 +46,20 @@ type GCPPubSub struct {
...
+	ProjectID               string `json:"projectID"`
+	PrivateKeyID            string `json:"privateKeyID"`
+	PrivateKey              string `json:"privateKey"`
+	ClientEmail             string `json:"clientEmail"`
+	ClientID                string `json:"clientID"`
+	AuthURI                 string `json:"authURI"`
+	TokenURI                string `json:"tokenURI"`
+	AuthProviderX509CertURL string `json:"authProviderX509CertURL"`
+	ClientX509CertURL       string `json:"clientX509CertURL"`
 }
```
sicoyle: there are more changes here to align this alpha bindings component to what the stable pubsub version of it has for the auth profile so I could just use the builtinAuthenticationProfiles for gcp on its metadata yaml file

bindings/gcp/pubsub/pubsub.go
```diff
@@ -46,19 +46,20 @@ type GCPPubSub struct {
...
+	Subscription            string `json:"subscription"`
+	Type                    string `json:"type"`
+	ProjectID               string `json:"projectID"`
+	PrivateKeyID            string `json:"privateKeyID"`
+	PrivateKey              string `json:"privateKey"`
+	ClientEmail             string `json:"clientEmail"`
+	ClientID                string `json:"clientID"`
+	AuthURI                 string `json:"authURI"`
+	TokenURI                string `json:"tokenURI"`
+	AuthProviderX509CertURL string `json:"authProviderX509CertURL"`
```
yaron2: This change is good because it consolidates our metadata, and is allowed since the binding is Alpha, but I am worried about breaking changes since I'm aware of several users using this binding in production

sicoyle: i updated to use our mapstructurealiases in our kit decoder to support both approaches and added a unit test to double check :)

sicoyle:
this is ready pls

PR #3961, opened 2025-08-13
fix: invert state.sqlserver decoding logic

nelson-parente:
This pull request updates the way values are encoded before being stored in SQL Server by improving the handling of `[]byte` and non-`[]byte` types in the `executeSet` method. The main change is to ensure that all values are consistently base64-encoded, whether they are raw bytes or marshaled JSON.

**Improvements to value encoding:**

* In `sqlserver.go`, the `executeSet` method now base64-encodes `[]byte` values directly, and for other types, it marshals them to JSON and then base64-encodes the result, ensuring consistent encoding for all stored values.

**TLDR**
Logic is the same but order of operations was inverted and tests pass.

### DIFF
--- a/tests/certification/state/sqlserver/sqlserver_test.go
+++ b/tests/certification/state/sqlserver/sqlserver_test.go
@@ -187,9 +187,14 @@ func TestSqlServer(t *testing.T) {
 		data, err := json.Marshal(order)
 		assert.NoError(ctx.T, err)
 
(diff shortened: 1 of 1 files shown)


nelson-parente:
This seems to fix the state.sqlserver issues:

https://github.com/dapr/components-contrib/actions/runs/16935208301

PR #3990, opened 2025-08-22
improve redis getkeyversion

filintod:
# Description

After checking some profiling in one of our programs that makes use of this method, we noticed that a lot of the allocations and cpu time was sunk on this method

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -48,7 +48,6 @@ require (
 	github.com/aws/aws-sdk-go-v2/feature/rds/auth v1.3.10
 	github.com/aws/aws-sdk-go-v2/service/bedrockruntime v1.17.3
 	github.com/aws/aws-sdk-go-v2/service/dynamodb v1.43.4
-	github.com/aws/aws-sdk-go-v2/service/s3 v1.83.0
 	github.com/aws/aws-sdk-go-v2/service/servicediscovery v1.32.4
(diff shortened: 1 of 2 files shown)


state/redis/redis_test.go
```diff
@@ -521,3 +521,43 @@ func setupMiniredis() (*miniredis.Miniredis, rediscomponent.RedisClient) {
...
+	/*
+		On a Mac M1 Pro:
+		BenchmarkGetKeyVersion-10    	    13651144	        83.84 ns/op	      64 B/op	       6 allocs/op
+
+		// old getkeyversion method
+		BenchmarkGetKeyVersionOld-10    	 1631097	       729.1 ns/op	      96 B/op	      10 allocs/op
+
+		// ~8x speed - ~1/2 allocations
+
+		// unsafe comparison
```
filintod, line 548: this is just for comparison with a run changing toString method to use unsafe []byte->string. In many cases it would be fine but leaving for future if really needed it could be added as an option

filintod:
@yaron2 should I put this in the 1.16 branch? or leave for later

yaron2:
Linter needs go mod tidy

filintod:
@yaron2 I think that was coming from the main branch after I updated it here, but it seems fine after you updated it again

PR #3991, opened 2025-08-25
Fix sqlserver workflows encoding

javier-aliaga:
# Description

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


state/sqlserver/sqlserver.go
```diff
@@ -305,16 +313,27 @@ type dbExecutor interface {
...
+		if err != nil {
+			return err
+		}
+		reqValue = string(bt)
+	} else {
+		if json.Valid(bytes) {
+			reqValue = string(bytes)
+		} else {
+			reqValue = base64.StdEncoding.EncodeToString(bytes)
+		}
```
javier-aliaga, line 330: It can be the case that receiving []bytes the data is a json, in this cases encoding to base64 could break db indexes. For that reason we check if the bytes are valid or not

JoshVanL:
As I understand the PR- we are determining the encoding of the given value to determine the storage format, based on an expensive `json.IsValid` call. Rather than paying that cost, I think it would be better for us to create a `v2` of sqlserver which enforcing the data format from the get go, without needing to satisfy expensive backward compatible requirements.

This is largely what we have also done with the postgresql v2 implementation.

Issue #4000, opened 2025-09-01
Azure Event Hubs binding getAllMessageProperties doesn't sanitize header keys


nikoraes:
I'm using the azure event hub binding and I'm having issues with events coming from [Azure Digital Twins](https://learn.microsoft.com/en-us/azure/digital-twins/concepts-event-notifications?tabs=event-hubs-service-bus) because it's sending event properties with characters that aren't allowed in headers (cloudEvents:subject, ...)

## Expected Behavior

When event properties are included in headers, the header names should be escaped or sanitized. This could be optional or it could be the default.

## Actual Behavior

Event is ignored and an error is logged in the sidecar.
`time="2025-09-01T10:15:18.700039175Z" level=error msg="Failed to process Eventhubs message %!s(*string=<nil>) for topic evh-******: Error: error invoking app: Post \"http://127.0.0.1:8080/dapr/trigger/eventhub": net/http: invalid header field name \"cloudevents:specversion\"" app_id=dt-flows component="eventhub-dtevtweu (bindings.azure.eventhubs/v1)" instance=dt-flows-dc74b577b-2mqns scope=dapr.contrib type=log ver=1.15.11`

## Steps to Reproduce the Problem

Make sure getAllMessageProperties is enabled and consume an event with custom properties with unallowed characters.

## Potential fix

Escape the headers here
[https://github.com/dapr/components-contrib/blob/main/common/component/azure/eventhubs/events.go#L50](https://github.com/dapr/components-contrib/blob/main/common/component/azure/eventhubs/events.go#L50)

nikoraes:
This is an issue for all bindings and should probably be solved in dapr core, which would probably cause breaking changes.
By using GRPC I was able to avoid this issue.

PR #4001, opened 2025-09-01
Azure Event Hub: Add header key sanitization to ensure valid HTTP header names

nikoraes:
# Description

Add metadata key sanitization in addPropertyToMetadata. This allows to ge all message properties even if the original message properties have keys that wouldn't be allowed as header names. 

## Issue reference

#4000

## Checklist

* [x] Code compiles correctly

### DIFF
--- a/common/component/azure/eventhubs/events.go
+++ b/common/component/azure/eventhubs/events.go
@@ -17,6 +17,7 @@ package eventhubs
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


nikoraes:
Actually, this is a general problem for all bindings (specific for HTTP) and should probably be solved in Dapr core.
Using GRPC avoids this issue.

Issue #4003, opened 2025-09-02
Add Support for certificate-based authentication in Kafka OIDC flow,


MyMirelHub:
## Describe the feature
Enhance the Kafka pub/sub component to support certificate-based authentication (e.g., `private_key_jwt`) for `authType: oidc`, addressing setups requiring both mTLS (with a trusted certificate) and a valid OAuth2 token.

Support token requests with:
- `client_id`: X.509 API Client.
- `grant_type`: "client_credentials".
- `resourceID`: Resource URI of the protected API.
- `client_assertion_type`: `urn:ietf:params:oauth:client-assertion-type:jwt-bearer`.
- `client_assertion`:  Signed Base64-encoded JWT using certificates.

Enable a new auth method to combine mTLS and OAuth2 (SASL) for managed Kafka setups supporting Kerberos and OAuth.

## Release Note
RELEASE NOTE: ADD Support for certificate-based authentication in Kafka OIDC flow, including mTLS + OAuth2 combination.

olitomlinson:
I'm not entirely sure that the underlying [Kafka Go Client](https://github.com/IBM/sarama) supports everything you've described.

- It does support SASL via [KIP-255](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=75968876)
-  However I see no support for OIDC (which would have been introduced in [KIP-768](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=186877575)

acroca:
@olitomlinson maybe I'm oversimplifying the solution, but wouldn't this be a matter of adding the certificates in the component configuration (instead of the secret, so either one or the other) and use those certificates in the TLS connection when we [get a token](https://github.com/dapr/components-contrib/blob/a9c2ecd57bd08a46a9a932c48e539e69dbaa8491/common/component/kafka/sasl_oauthbearer.go#L107)?

acroca:
https://github.com/dapr/components-contrib/pull/4057 got merged and will be released in 1.17

PR #4057, opened 2025-10-06
Feat: Support client JWT auth method for kafka oidc

acroca:
Ref: https://github.com/dapr/components-contrib/issues/4003

At the moment the `oidc` auth type only supports authentication using `clientID` and `clientSecret`.

This PR adds a new auth type `oidc_private_key_jwt`that uses certificates instead. The component will use those certificates to build the client assertion and call the IdP to get a token.

This new auth type uses the following new fields in the kafka components:
- `oidcClientAssertionCert`
- `oidcClientAssertionKey`
- `oidcResource`
- `oidcAudience`

In this PR I also added tests for both oidc authentication mechanisms against a local kafka, using keycloak as an IdP.

### DIFF
--- a/bindings/kafka/metadata.yaml
+++ b/bindings/kafka/metadata.yaml
@@ -174,6 +174,19 @@ authenticationProfiles:
           Overrides the JWT client assertion audience (aud). If not set, the component uses the
           issuer derived from the token endpoint URL when available; otherwise, it falls back to the token URL.
         example: '"http://<idp-host>/realms/local"'
+      - name: oidcScopes
+        type: string
+        description: |
+          Comma-delimited list of OAuth2/OIDC scopes to request with the access token.
+          Although not required, this field is recommended.
+        example: '"openid,kafka-prod"'
+        default: '"openid"'
+      - name: oidcExtensions
+        description: |
+          String containing a JSON-encoded dictionary of OAuth2/OIDC extensions to request with the access token.
+        example: |
+          {"cluster":"kafka","poolid":"kafkapool"}
+        type: string
   - title: "SASL Authentication"
     description: |
       Authenticate using SASL.
--- a/pubsub/kafka/metadata.yaml
+++ b/pubsub/kafka/metadata.yaml
@@ -168,6 +168,19 @@ authenticationProfiles:
           Overrides the JWT client assertion audience (aud). If not set, the component uses the
           issuer derived from the token endpoint URL when available; otherwise, it falls back to the token URL.
         example: '"http://<idp-host>/realms/local"'
+      - name: oidcScopes
+        type: string
+        description: |
+          Comma-delimited list of OAuth2/OIDC scopes to request with the access token.
+          Although not required, this field is recommended.
+        example: '"openid,kafka-prod"'
+        default: '"openid"'
+      - name: oidcExtensions
+        description: |
+          String containing a JSON-encoded dictionary of OAuth2/OIDC extensions to request with the access token.
+        example: |
+          {"cluster":"kafka","poolid":"kafkapool"}
+        type: string
   - title: "SASL Authentication"
     description: |
       Authenticate using SASL.


common/component/kafka/metadata.go
```diff
@@ -235,8 +240,14 @@ func (k *Kafka) getKafkaMetadata(meta map[string]string) (*KafkaMetadata, error)
...
-			return nil, errors.New("kafka error: missing OIDC Client Secret for authType 'oidc'")
+		if m.OidcClientAuthMethod == "client_secret" && m.OidcClientSecret == "" {
+			return nil, errors.New("kafka error: missing OIDC Client Secret for authType 'oidc' (client_secret)")
+		}
+		if m.OidcClientAuthMethod == "client_jwt" && m.OidcClientAssertionCert == "" {
+			return nil, errors.New("kafka error: missing OIDC Client Assertion Cert for authType 'oidc' (client_jwt)")
+		}
+		if m.OidcClientAuthMethod == "client_jwt" && m.OidcClientAssertionKey == "" {
+			return nil, errors.New("kafka error: missing OIDC Client Assertion Key for authType 'oidc' (client_jwt)")
 		}
```
JoshVanL: Please can we use pointers to signal presence.

common/component/kafka/sasl_oauthbearer.go
```diff
@@ -131,6 +170,99 @@ func (ts *OAuthTokenSource) Token() (*sarama.AccessToken, error) {
...
+		if rsaKey, err2 := x509.ParsePKCS1PrivateKey(block.Bytes); err2 == nil {
+			pk = rsaKey
+		} else {
+			return nil, fmt.Errorf("unable to parse private key: %w", err)
+		}
+	}
+	rsaKey, ok := pk.(*rsa.PrivateKey)
+	if !ok {
+		return nil, errors.New("client_jwt requires RSA private key")
+	}
```
JoshVanL: Use `dapr/kit/crypto/pem`.

common/component/kafka/sasl_oauthbearer.go
```diff
@@ -131,6 +170,99 @@ func (ts *OAuthTokenSource) Token() (*sarama.AccessToken, error) {
...
+	if ts.Audience != "" {
+		audClaim = ts.Audience
+	}
+
+	token, err := jwt.NewBuilder().
+		Issuer(ts.ClientID).
+		Subject(ts.ClientID).
+		Audience([]string{audClaim}).
+		IssuedAt(now).
+		Expiration(now.Add(1 * time.Minute)).
```
JoshVanL: Why 1 minute?

common/component/kafka/sasl_oauthbearer.go
```diff
@@ -131,6 +170,99 @@ func (ts *OAuthTokenSource) Token() (*sarama.AccessToken, error) {
...
+		urlValues.Set("audience", ts.Audience)
+	}
+	if ts.Resource != "" {
+		urlValues.Set("resource", ts.Resource)
+	}
+	if len(ts.Scopes) > 0 {
+		urlValues.Set("scope", strings.Join(ts.Scopes, " "))
+	}
+
+	timeoutCtx, cancel := ctx.WithTimeout(ctx.TODO(), tokenRequestTimeout)
```
JoshVanL: Why `TODO`?

common/component/kafka/sasl_oauthbearer.go
```diff
@@ -131,6 +170,99 @@ func (ts *OAuthTokenSource) Token() (*sarama.AccessToken, error) {
...
+	}
+	var tr tokenResponse
+	if err := json.NewDecoder(resp.Body).Decode(&tr); err != nil {
+		return nil, err
+	}
+	if tr.AccessToken == "" {
+		return nil, errors.New("no access_token in response")
+	}
+	ts.CachedToken = oauth2.Token{AccessToken: tr.AccessToken, Expiry: time.Now().Add(time.Duration(tr.ExpiresIn) * time.Second)}
+	return ts.asSaramaToken(), nil
```
JoshVanL: Is there not something that does all this for us in a library?

tests/certification/pubsub/kafka/components/auth_oidc_certs/kafka.yaml
```diff
@@ -0,0 +1,81 @@
...
+      CSqGSIb3DQEBAQUAA4IBDwAwggEKAoIBAQDQIqGyGGlQRqXNDlSuHRDGmnJ4dQDb
+      jK/KxeC+wXQiDCSfHgP7NOjZOOCiQYKAUs9MRg+tKepxFRORZLAbRlU9kuU4Z7bg
+      E6wDMztCqTASUZtwZeN36iJDcQ++xnmAi8j0bvs7w00wrgPMXPnT09Sf9sQyEkR2
+      t26OLcQnZHwrUWHIbZeXqvRF7uYG00GKjFmqI+FjrmDx0hoENew+Jzpk7S/7m5t2
+      h/weVAr5REtGTIQxbX8nNmoO2JFbyILBcfP9M9R9zxiTEFVgP5sl2QfLFerPpULM
+      1tAiCO+oPk8CAFiD9TKe+HR/Us4uIxgRmxjgBnzmFSJjDflvK2DslpzPAgMBAAGj
+      ZDBiMAsGA1UdDwQEAwIFoDATBgNVHSUEDDAKBggrBgEFBQcDAjAdBgNVHQ4EFgQU
+      oCgedGjEdPHOBRq6OwKKGUmpUXIwHwYDVR0jBBgwFoAUImUtm30c0E0Z1Xsep9o/
+      nXsqyGYwDQYJKoZIhvcNAQELBQADggIBAKafBCz1EUny2pKOI8xTHVkoZjhTvtm2
+      Ly2DicAEwtESXlBta64bOug4mTq8RLdKxLa8xSBSZxzygTfb+q8IcuuYL0/DAttZ
```
JoshVanL: Please do not use hard coded certificates and keys in tests. They get copied and used in the real world, and they also expire.

Always generate them in the test, and use secret refs

common/component/kafka/metadata.go
acroca: I used empty strings for consistence with the rest of the metadata fields. I think it'd be better to change it all in a different PR, wdyt?

common/component/kafka/sasl_oauthbearer.go
acroca: It's arbitrary, it's just a short lived token used to communicate with the IdP. It could be even shorter in practice. Do you think we should make it shorter, like 5 seconds or so maybe?

common/component/kafka/sasl_oauthbearer.go
acroca: It was `TODO` before too, I don't know the reason. Other places use the background context 🤷 
The `sarama.AccessTokenProvider`([docs](https://pkg.go.dev/github.com/Shopify/sarama#AccessTokenProvider)) doesn't take a context, wiring it up with a provided context would change a bit too much. Should we use background instead?

common/component/kafka/sasl_oauthbearer.go
acroca: I couldn't find any unfortunately

bindings/kafka/metadata.yaml
```diff
@@ -121,6 +121,50 @@ authenticationProfiles:
...
           {"cluster":"kafka","poolid":"kafkapool"}
         type: string
+      - name: oidcClientAuthMethod
+        type: string
+        required: false
+        default: '"client_secret"'
+        description: |
+          Client authentication method at the token endpoint.
+          The "client_secret" method (default) uses the client_id and client_secret to authenticate.
+          The "client_jwt" method uses a JWT client assertion to authenticate.
```
sicoyle: would it make sense to instead of supporting a field for oidcClientAuthMethod to instead make the current oidc auth profile to be the client_secret approach, and then to add a secondary higher level auth profile for OIDC JWT Client assertion? That way then the metadata fields are cleanly separated according to the actual auth profile a user is going for so it is a cleaner DX? This would remove the fields such as `oidcClientAssertionCert` to not have conditionals on if it is required or not.

common/component/kafka/sasl_oauthbearer.go
JoshVanL: We should take a context which is passed through

JoshVanL (dismissed):

tests/certification/pubsub/kafka/components/auth_oidc_certs/kafka.yaml
```diff
@@ -0,0 +1,33 @@
...
+  - name: authType
+    value: "oidc_private_key_jwt"
+  - name: oidcTokenEndpoint
+    value: "http://localhost:8080/realms/local/protocol/openid-connect/token"
+  - name: oidcClientID
+    value: "dapr-kafka-client-jwt"
+  - name: oidcClientAssertionCert
+    secretKeyRef:
+      name: OICD_CLIENT_ASSERTION_CERT
+      key:  OICD_CLIENT_ASSERTION_CERT
```
sicoyle:
```suggestion
      name: OIDC_CLIENT_ASSERTION_CERT
      key:  OIDC_CLIENT_ASSERTION_CERT
```

tests/certification/pubsub/kafka/components/auth_oidc_certs/kafka.yaml
```diff
@@ -0,0 +1,33 @@
...
+  - name: oidcClientAssertionKey
+    secretKeyRef:
+      name: OICD_CLIENT_ASSERTION_KEY
+      key:  OICD_CLIENT_ASSERTION_KEY
```
sicoyle:
```suggestion
      name: OIDC_CLIENT_ASSERTION_KEY
      key:  OIDC_CLIENT_ASSERTION_KEY
```

tests/certification/pubsub/kafka/data/realm-export.json
```diff
@@ -0,0 +1,43 @@
...
+  "realm": "local",
+  "enabled": true,
+  "clients": [
+    {
+      "clientId": "dapr-kafka-client-jwt",
+      "secret": "dapr-kafka-secret",
+      "clientAuthenticatorType": "client-jwt",
+      "attributes": {
+        "token.endpoint.auth.signing.alg": "RS256",
+        "jwt.credential.certificate": "${OICD_CLIENT_ASSERTION_CERT_ONELINE}"
```
sicoyle:
```suggestion
        "jwt.credential.certificate": "${OIDC_CLIENT_ASSERTION_CERT_ONELINE}"
```

tests/certification/pubsub/kafka/docker-compose.auth.yml
```diff
@@ -0,0 +1,62 @@
...
+    container_name: keycloak
+    command: ["start-dev", "--http-enabled=true", "--health-enabled=true", "--import-realm"]
+    healthcheck:
+      test: ["CMD-SHELL", "exec 3<>/dev/tcp/127.0.0.1/8080; echo -e 'GET /health/ready HTTP/1.1\r\nHost: localhost:8080\r\nConnection: close\r\n\r\n' >&3;cat <&3 | grep -q '\"status\": \"UP\"' && exit 0 || exit 1"]
+      start_period: 20s
+      interval: 2s
+      timeout: 2s
+      retries: 5
+    environment:
+      - OICD_CLIENT_ASSERTION_CERT_ONELINE
```
sicoyle:
```suggestion
      - OIDC_CLIENT_ASSERTION_CERT_ONELINE
```

tests/certification/pubsub/kafka/kafka_test.go
```diff
@@ -454,14 +470,111 @@ func TestKafka(t *testing.T) {
...
+			Organization: []string{"Dapr"},
+		},
+	}
+	cert, err := x509.CreateCertificate(rand.Reader, template, template, &key.PublicKey, key)
+	require.NoError(t, err)
+	certPEM := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: cert})
+	keyPEM := pem.EncodeToMemory(&pem.Block{Type: "RSA PRIVATE KEY", Bytes: x509.MarshalPKCS1PrivateKey(key)})
+	os.Setenv("OICD_CLIENT_ASSERTION_CERT", string(certPEM))
+	os.Setenv("OICD_CLIENT_ASSERTION_KEY", string(keyPEM))
+	os.Setenv("OICD_CLIENT_ASSERTION_CERT_ONELINE", strings.ReplaceAll(string(certPEM), "\n", "\\n"))
```
sicoyle:
```suggestion
	os.Setenv("OIDC_CLIENT_ASSERTION_CERT", string(certPEM))
	os.Setenv("OIDC_CLIENT_ASSERTION_KEY", string(keyPEM))
	os.Setenv("OIDC_CLIENT_ASSERTION_CERT_ONELINE", strings.ReplaceAll(string(certPEM), "\n", "\\n"))
```

common/component/kafka/sasl_oauthbearer.go
sicoyle: pls ensure the docs reflect this as well when the time comes 🙏

common/component/kafka/metadata.go
```diff
@@ -251,6 +256,32 @@ func (k *Kafka) getKafkaMetadata(meta map[string]string) (*KafkaMetadata, error)
...
+		if m.OidcClientID == "" {
+			return nil, errors.New("kafka error: missing OIDC Client ID for authType 'oidc_private_key_jwt'")
+		}
+		if m.OidcClientAssertionCert == "" {
+			return nil, errors.New("kafka error: missing OIDC Client Assertion Cert for authType 'oidc_private_key_jwt'")
+		}
+		if m.OidcClientAssertionKey == "" {
+			return nil, errors.New("kafka error: missing OIDC Client Assertion Key for authType 'oidc_private_key_jwt'")
+		}
+		if m.OidcScopes != "" {
```
sicoyle, line 272: pls add this field to the auth profile in the metadata.yaml as I don't see it with the openid string default

common/component/kafka/metadata.go
```diff
@@ -251,6 +256,32 @@ func (k *Kafka) getKafkaMetadata(meta map[string]string) (*KafkaMetadata, error)
...
+			m.internalOidcScopes = strings.Split(m.OidcScopes, ",")
+		} else {
+			k.logger.Warn("Warning: no OIDC scopes specified, using default 'openid' scope only. This is a security risk for token reuse.")
+			m.internalOidcScopes = []string{"openid"}
+		}
+		if m.OidcExtensions != "" {
```
sicoyle, line 278: i think this is also missing in the metadata.yaml file. So this is an optional field?

common/component/kafka/metadata_test.go
```diff
@@ -243,6 +243,38 @@ func TestMissingOidcValues(t *testing.T) {
...
+	meta, err = k.getKafkaMetadata(m)
+	require.Error(t, err)
+	require.Nil(t, meta)
+	require.Equal(t, "kafka error: missing OIDC Client Assertion Key for authType 'oidc_private_key_jwt'", err.Error())
+
+	m["oidcClientAssertionKey"] = "sassapass"
+	meta, err = k.getKafkaMetadata(m)
+	require.NoError(t, err)
+	require.Contains(t, meta.internalOidcScopes, "openid")
+}
```
sicoyle, line 276: can you double check these check for all of the required fields you've set in the metadata please? I count 5 required but see 4 here

common/component/kafka/sasl_oauthbearer_private_key_jwt.go
```diff
@@ -0,0 +1,218 @@
...
+	ClientAuthMethod    string
+	ClientAssertionCert string
+	ClientAssertionKey  string
+	Resource            string
+	Audience            string
+}
+
+type tokenResponse struct {
+	AccessToken string `json:"access_token"`
+	ExpiresIn   int64  `json:"expires_in"`
```
sicoyle, line 57: why not camelCase?

acroca: This is the structure we get from the IdP

common/component/kafka/metadata_test.go
acroca: The 5th one is `authType`. It's the one used to select this authentication type, there's no need to test it directly.

Issue #4061, opened 2025-10-10
SFTP Binding seems to overload SFTP server with connections


SteffenEbert-HK:
## Expected Behavior
SFTP Binding can be used to connect to SFTP without having to restart the service to give up connections

## Actual Behavior
We see that after a couple of testcases have been run that the SFTP server declines communication due to too many open connections through dapr binding

## Steps to Reproduce the Problem
Connect to SFTP via binding and upload files until SFTP server declines communication

## Release Note
RELEASE NOTE:
 **FIX** Bug for SFTP Binding, releasing connections with SFTP Server.

PR #4075, opened 2025-10-27
chore: Handle sftp reconnections

javier-aliaga:
# Description

This PR introduces a re-connection mechanism to the sftp binding. 

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/components-contrib/issues/4061

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/bindings/sftp/client.go
+++ b/bindings/sftp/client.go
@@ -1,0 +1,255 @@
+/*
+Copyright 2025 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package sftp
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"sync"
+	"sync/atomic"
+
+	sftpClient "github.com/pkg/sftp"
+	"golang.org/x/crypto/ssh"
+
+	"github.com/dapr/kit/logger"
+)
+
+type Client struct {
+	sshClient      *ssh.Client
+	sftpClient     *sftpClient.Client
+	address        string
+	config         *ssh.ClientConfig
+	lock           sync.RWMutex
+	needsReconnect atomic.Bool
+	log            logger.Logger
+}
+
+func newClient(address string, config *ssh.ClientConfig, log logger.Logger) (*Client, error) {
+	if address == "" || config == nil {
+		return nil, errors.New("sftp binding error: client not initialized")
+	}
+
+	sshClient, err := newSSHClient(address, config)
+	if err != nil {
+		return nil, err
+	}
+
+	newSftpClient, err := sftpClient.NewClient(sshClient)
+	if err != nil {
+		_ = sshClient.Close()
+		return nil, fmt.Errorf("sftp binding error: error create sftp client: %w", err)
+	}
+
+	return &Client{
+		sshClient:  sshClient,
+		sftpClient: newSftpClient,
+		address:    address,
+		config:     config,
+		log:        log,
+	}, nil
+}
+
+func (c *Client) Close() error {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	// Close SFTP first, then SSH
+	var sftpErr, sshErr error
+	if c.sftpClient != nil {
+		sftpErr = c.sftpClient.Close()
+	}
+	if c.sshClient != nil {
+		sshErr = c.sshClient.Close()
+	}
+
+	// Return the first error encountered
+	if sftpErr != nil {
+		return sftpErr
+	}
+	return sshErr
+}
+
+func (c *Client) list(path string) ([]os.FileInfo, error) {
+	var fi []os.FileInfo
+
+	fn := func() error {
+		var err error
+		fi, err = c.sftpClient.ReadDir(path)
+		return err
+	}
+
+	err := c.withReconnection(fn)
+	if err != nil {
+		return nil, err
+	}
+
+	return fi, nil
+}
+
+func (c *Client) create(path string) (*sftpClient.File, string, error) {
+	dir, fileName := sftpClient.Split(path)
+
+	var file *sftpClient.File
+
+	createFn := func() error {
+		cErr := c.sftpClient.MkdirAll(dir)
+		if cErr != nil {
+			return cErr
+		}
+
+		file, cErr = c.sftpClient.Create(path)
+		if cErr != nil {
+			return cErr
+		}
+
+		return nil
+	}
+
+	rErr := c.withReconnection(createFn)
+	if rErr != nil {
+		return nil, "", rErr
+	}
+
+	return file, fileName, nil
+}
+
+func (c *Client) get(path string) (*sftpClient.File, error) {
+	var f *sftpClient.File
+
+	fn := func() error {
+		var err error
+		f, err = c.sftpClient.Open(path)
+		return err
+	}
+
+	err := c.withReconnection(fn)
+	if err != nil {
+		return nil, err
+	}
+
+	return f, nil
+}
+
+func (c *Client) delete(path string) error {
+	fn := func() error {
+		return c.sftpClient.Remove(path)
+	}
+
+	err := c.withReconnection(fn)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *Client) ping() error {
+	_, err := c.sftpClient.Getwd()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (c *Client) withReconnection(fn func() error) error {
+	err := c.do(fn)
+	if !c.shouldReconnect(err) {
+		return err
+	}
+
+	c.log.Debugf("sftp binding error: %s", err)
+	c.needsReconnect.Store(true)
+
+	rErr := c.doReconnect()
+	if rErr != nil {
+		c.log.Debugf("sftp binding error: reconnect failed: %s", rErr)
+		return errors.Join(err, rErr)
+	}
+	c.log.Debugf("sftp binding: reconnected to %s", c.address)
+
+	c.log.Debugf("sftp binding: retrying operation")
+	return c.do(fn)
+}
+
+func (c *Client) do(fn func() error) error {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	return fn()
+}
+
+func (c *Client) doReconnect() error {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	c.log.Debugf("sftp binding: reconnecting to %s", c.address)
+
+	if !c.needsReconnect.Load() {
+		return nil
+	}
+
+	pErr := c.ping()
+	if pErr == nil {
+		c.needsReconnect.Store(false)
+		return nil
+	}
+
+	sshClient, err := newSSHClient(c.address, c.config)
+	if err != nil {
+		return err
+	}
+
+	newSftpClient, err := sftpClient.NewClient(sshClient)
+	if err != nil {
+		_ = sshClient.Close()
+		return fmt.Errorf("sftp binding error: error create sftp client: %w", err)
+	}
+
+	if c.sftpClient != nil {
+		_ = c.sftpClient.Close()
+	}
+	if c.sshClient != nil {
+		_ = c.sshClient.Close()
+	}
+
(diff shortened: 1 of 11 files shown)


tests/utils/sftpproxy/proxy.go
```diff
@@ -0,0 +1,111 @@
+/*
```
JoshVanL, line 1: Please can we move this package to `tests/utils/sftpproxy`.

bindings/sftp/client.go
```diff
@@ -0,0 +1,254 @@
+/*
```
JoshVanL, line 1: Rename file to `client.go`.

bindings/sftp/client.go
```diff
@@ -0,0 +1,254 @@
...
+	err := c.ping()
+	if !shouldReconnect(err) {
+		return nil
+	}
+
+	sshClient, err := ssh.Dial("tcp", c.address, c.config)
+	if err != nil {
+		return fmt.Errorf("sftp binding error: error create ssh client: %w", err)
+	}
+
```
JoshVanL: Please can we move client creation to a single func?

bindings/sftp/client.go
```diff
@@ -0,0 +1,254 @@
...
+	"golang.org/x/crypto/ssh"
+)
+
+type Client struct {
+	sshClient  *ssh.Client
+	sftpClient *sftpClient.Client
+	address    string
+	config     *ssh.ClientConfig
+	lock       sync.RWMutex
+	rLock      sync.Mutex
```
JoshVanL: Why do we need two locks?

bindings/sftp/client.go
```diff
@@ -0,0 +1,254 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func doReconnect(c *Client) error {
+	c.rLock.Lock()
+	defer c.rLock.Unlock()
```
JoshVanL: It is hard to follow the logic when we have 2 locks

bindings/sftp/client.go
```diff
@@ -0,0 +1,254 @@
...
+	// Network/timeout conditions
+	if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) || errors.Is(err, os.ErrDeadlineExceeded) {
+		return true
+	}
+
+	// Common wrapped network error messages
+	msg := strings.ToLower(err.Error())
+	switch {
+	case strings.Contains(msg, "use of closed network connection"),
+		strings.Contains(msg, "connection reset by peer"),
```
JoshVanL: Can we not use typed errors here? `syscall.ECONNRESET`

JoshVanL (changes requested):

bindings/sftp/client.go
javier-aliaga: Because:

- `lock`
  - Controls that only one goroutin trying to reconnect and verifies with `ping()` if the connection is valid or not

- `rlock`
  - We do want to prevent multiple goroutines trying to use the client while swapping the connection and ensure the consistency of the client. We cannot swap the client until we have the full lock

bindings/sftp/client.go
javier-aliaga: I will add to the previous if(L230), but I will keep the message here just in case.

bindings/sftp/client.go
javier-aliaga: Added comment to explain why is each lock used, let me know what do you think

bindings/sftp/client.go
javier-aliaga: I modified the method, we will not reconnect if it is one of this errors:

```
if errors.Is(err, sftpClient.ErrSSHFxPermissionDenied) ||
		errors.Is(err, sftpClient.ErrSSHFxNoSuchFile) ||
		errors.Is(err, sftpClient.ErrSSHFxOpUnsupported) {
		return false
	}
```

Any other error we will try to reconnect

bindings/sftp/client.go
mikeee: Perhaps renaming the locks if this is the best path forwards would be helpful e.g.:
- reconnectLock
- operationsLock

bindings/sftp/sftp_integration_test.go
```diff
@@ -1,3 +1,16 @@
+/*
```
mikeee, line 3:
```suggestion
//go:build integration_test

/*
```

Could we use a build flag here and also spin up the required docker container in a setup function? I'm not a fan of dead tests that are skipped or not run without manual intervention.

bindings/sftp/client.go
JoshVanL: Why can we not have a single read write lock?

- rlock: when using the client in the happy path
- lock: when reconnecting

javier-aliaga: The problem with that is we do not know is the connection is valid, so inside the reconnect we have a ping method that does a simple request to the server. 

So imagine you have a connection broken and you pile up 1000 goroutins waiting to reconnect, once the first action heals the connection then the other 999 will just ping the server and exits (all with a lock). This 999 goroutins will not lock the readers. If you have a single lock for re-connections then every new goroutine will be blocked.

JoshVanL: We should be able to achieve the right behaviour (and simplify the logic) with the following pseudo code

```
r.RLock()
client.Call
if err != nil { needsReconnect.Store(true) }
r.RUnlock()
if err != nil {
  r.Lock()
  if needsReconnect.Load() { RECONECT(); needsReconnect.Store(false) }
  r.UnLock()
  GOTO BEGIN
}
```

bindings/sftp/client.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+func withReconnection(c *Client, fn func() error) error {
+	err := fn()
+	if err == nil {
+		return nil
+	}
+
+	if !shouldReconnect(err) {
+		c.needsReconnect.Store(false)
```
JoshVanL: Surely this should only be done be `doReconnect`?

bindings/sftp/client.go
```diff
@@ -0,0 +1,279 @@
...
+		newSftpClient = nil
+	}
+	c.lock.Unlock()
+
+	if oldSftp != nil {
+		_ = oldSftp.Close()
+	}
+	if oldSSH != nil {
+		_ = oldSSH.Close()
+	}
```
JoshVanL: Why is this done afterwards?

bindings/sftp/client.go
```diff
@@ -0,0 +1,279 @@
...
+	var oldSftp *sftpClient.Client
+	var oldSSH *ssh.Client
+	if c.needsReconnect.Load() {
+		oldSftp = c.sftpClient
+		oldSSH = c.sshClient
+		c.sftpClient = newSftpClient
+		c.sshClient = sshClient
+		c.needsReconnect.Store(false)
+		sshClient = nil
+		newSftpClient = nil
```
JoshVanL: Why?

bindings/sftp/client.go
```diff
@@ -0,0 +1,279 @@
...
+
+	err = fn()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func doReconnect(c *Client) error {
```
JoshVanL: Why isn't this all inside the lock?

javier-aliaga: the problem is the error may not be because of the connections, so we have the ping() to check if the connection is alive or not. I am going to move out the ping() from here and move it to the shouldReconnect func. Then we can lock all the doReconnect

bindings/sftp/client.go
javier-aliaga: Because the creation was not locked, it could happen that another goroutine fixed the connexction, so in that case no need to swap the new clients, but we need to close them

bindings/sftp/client.go
javier-aliaga: true

bindings/sftp/client.go
javier-aliaga: what do you mean? Why we close or why we close after unlock?

bindings/sftp/client.go
```diff
@@ -0,0 +1,257 @@
...
+	if !c.needsReconnect.Load() {
+		return nil
+	}
```
JoshVanL, line 200: Move below lock

bindings/sftp/client.go
```diff
@@ -0,0 +1,257 @@
...
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	pErr := c.ping()
+	if pErr != nil {
```
JoshVanL:
```suggestion
	err := c.ping()
	if err == nil {
	  return nil
	 }
```

bindings/sftp/client.go
```diff
@@ -0,0 +1,257 @@
...
+		}
+
+		newSftpClient, err := sftpClient.NewClient(sshClient)
+		if err != nil {
+			_ = sshClient.Close()
+			return fmt.Errorf("sftp binding error: error create sftp client: %w", err)
+		}
+
+		oldSftp := c.sftpClient
+		oldSSH := c.sshClient
```
JoshVanL: No need for these temp variables.


```suggestion
		if c.sftpClient != nil {
		c.sftpClient.Close()
		}
		if c.sshClient != nil {
		  c.sshClient.Close()
		}
```

bindings/sftp/client.go
```diff
@@ -0,0 +1,257 @@
...
+func withReconnection(c *Client, fn func() error) error {
+	c.lock.RLock()
+	err := fn()
+	if c.shouldReconnect(err) {
+		c.log.Debugf("sftp binding error: %s", err)
+		c.needsReconnect.Store(true)
+	} else {
+		c.lock.RUnlock()
+		return err
+	}
```
JoshVanL:
```suggestion
	if !c.shouldReconnect(err) {
		c.lock.RUnlock()
		return err
	}
	
		c.log.Debugf("sftp binding error: %s", err)
		c.needsReconnect.Store(true)
```

bindings/sftp/client.go
```diff
@@ -0,0 +1,257 @@
...
+	c.lock.RUnlock()
+
+	if c.needsReconnect.Load() {
```
JoshVanL:
```suggestion
```

bindings/sftp/client.go
```diff
@@ -0,0 +1,257 @@
...
+		c.log.Debugf("sftp binding: reconnecting to %s", c.address)
```
JoshVanL: Move log line inside `doReconnect`, and only after we have locked and checked `nedsReconnect`.

bindings/sftp/client.go
javier-aliaga: what if the c.ping works? should we set needsReconnect to false?

JoshVanL: Yep, I think so :)

JoshVanL (dismissed):

bindings/sftp/client.go
```diff
@@ -0,0 +1,259 @@
...
+
+func (c *Client) ping() error {
+	_, err := c.sftpClient.Getwd()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func withReconnection(c *Client, fn func() error) error {
```
cicoyle: can you make this a ptr receiver func and then we dont have to pass around c?

bindings/sftp/client.go
```diff
@@ -0,0 +1,259 @@
...
+	defer c.lock.RUnlock()
+	err = fn()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func doReconnect(c *Client) error {
```
cicoyle: Same here, can you make this a ptr receiver func and then we dont have to pass around c?

bindings/sftp/client.go
```diff
@@ -0,0 +1,259 @@
...
+func (c *Client) ping() error {
+	_, err := c.sftpClient.Getwd()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func withReconnection(c *Client, fn func() error) error {
+	c.lock.RLock()
```
cicoyle: This locking then calling to reconnect then locking again feels like maybe we should have an abstraction function with a for loop where we only lock in that one place versus in a few places. Maybe it could be cleaner if we had a `client.do` func that had a `for` loop where we check the client, then do the `shouldReconnect` and `reconnect` calls - this way we try reconnecting more than once otherwise, as is we only try once I believe.

javier-aliaga: I did not want to put more complex mechanism with retries and backoffs, I think this component it may need to be rewritten as some customers requested different improvements. I did a small refactor to introduce a do func where we lock the action https://github.com/dapr/components-contrib/pull/4075/commits/51c4e947a26b78f65a4101c279506d31f16782ad

what do you think?

cicoyle: I agree it can be rewritten and improved further. plz open an issue so we dont forget and it can be prioritized at a later date 👍🏻

PR #4079, opened 2025-10-29
Fix token renewal for pulsar oauth2

acroca:
Fixes the pulsar OAuth2 token renewal process.

### DIFF
--- a/common/authentication/oauth2/clientcredentials_test.go
+++ b/common/authentication/oauth2/clientcredentials_test.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"golang.org/x/oauth2"
 	ccreds "golang.org/x/oauth2/clientcredentials"
 )
@@ -109,6 +110,6 @@ func Test_TokenRenewal(t *testing.T) {
 	}
 
 	tok, err := c.Token()
-	assert.NoError(t, err)
+	require.NoError(t, err)
 	assert.Equal(t, "new-token", tok)
 }


MyMirelHub (approved):

nelson-parente (approved):

common/authentication/oauth2/clientcredentials.go
```diff
@@ -127,17 +127,22 @@ func (c *ClientCredentialsOptions) toConfig() (*ccreds.Config, *http.Client, err
...
+	currentAccessToken := c.currentToken.AccessToken
+	c.lock.RUnlock()
 
-	if !c.currentToken.Valid() {
+	if !isValid {
 		ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
 		defer cancel()
 		if err := c.renewToken(ctx); err != nil {
 			return "", err
 		}
```
javier-aliaga: Shall we lock this? what happens if multiple goroutines try to renew the token?

common/authentication/oauth2/clientcredentials.go
```diff
@@ -127,17 +127,22 @@ func (c *ClientCredentialsOptions) toConfig() (*ccreds.Config, *http.Client, err
...
+		c.lock.RLock()
```
JoshVanL: Why is the lock no longer a defer?
This is inherently racey

JoshVanL (changes requested):

common/authentication/oauth2/clientcredentials.go
acroca: the `renewToken` does lock, we were getting into the `renewToken` with the `RLock` locking it so it could never get the lock.

common/authentication/oauth2/clientcredentials.go
acroca: The `renewToken` has its own lock, and also has a check to make sure the token is not valid before renewing, all within its lock, so if two goroutines are renewing, the second one will find a valid token and will `return nil`

common/authentication/oauth2/clientcredentials.go
JoshVanL: This just means we will call renew extra times than is necessary

It's not clear to me what's being fixed here

acroca: It was not working at all before, the lock in the renew function was never able to get the lock because it was being called from the `Token` function, which has a lock during its whole execution.
I changed the Token to only lock for reading, but release early in case it needs to perform the renewal

JoshVanL: Renewing the token should be considered expensive. We should `.Lock()` the entire token cache to prevent a race whereby `renewToken` is called multiple times during the same token lifecycle.

acroca: Ok, changed it to be all within a full lock. With the full lock we don't need to double check for validity before triggering the renewal

JoshVanL (dismissed):

JoshVanL:
@acroca lint failing

JoshVanL (approved):

marcduiker:
@holopin-bot @acroca Here's a digital badge as a small token of appreciation.

Issue #3896, opened 2025-07-22
Upgrading the rest of the components using AWS to V2


mikeee:
## Describe the feature

Components outstanding are (with work size estimate):

- [x] (Common) Kafka aws path - L

Deprioritised:
- [ ] Bindings.s3 - L - #3910
- [ ] Bindings.ses - M - #3911 
- [ ] Bindings.kinesis - M - #4082
- [ ] Bindings.sns - M
- [ ] Bindings.sqs - M
- [ ] Configuration.postgres (migrate to new common provider) - S
- [ ] Conversation.bedrock (migrate to new common provider) - S - #3909
- [ ] SecretStore.parameterstore - S - #3908
- [ ] SecretStore.secretmanager - S - #3901

Stretch goals would be to migrate tests to v2 although as a lower priority, the tests work regardless of v1/v2 and the focus should be on stability as opposed to overhauling the tests as of yet

## Release Note

RELEASE NOTE: **UPDATE** All components interface with AWS using the V2 sdk

peterdalinis-scout:
Is github.com/vmware/vmware-go-kcl-v2 also part of this upgrade?

PR #4082, opened 2025-10-31
Feat/kinesis binding vmware go kcl v2 latest

swatimodi-scout:
# Description

This PR includes two key improvements to the AWS Kinesis input binding:

- **Upgrade to vmware-go-kcl-v2**
   - Updated the Kinesis consumer dependency from the legacy vmware-go-kcl to vmware-go-kcl-v2.
   - Motivation: The original library is no longer actively maintained and lacks critical bug fixes and performance improvements.
   - **Changes:**
        - Updated go.mod to reference github.com/vmware/vmware-go-kcl-v2.
        - Adjusted import paths and updated worker configuration to align with the new v2 APIs.
        - Verified compatibility with existing Kinesis input binding behavior through unit and integration tests.
- **Fix for nil pointer reference in Kinesis input binding**
     - Resolved a bug where the binding attempted to resolve the stream ARN unconditionally, even when KinesisConsumerMode was set to SharedThroughput (default mode).
     - Issue: The binding was calling authProvider.Kinesis().Stream(...) regardless of the configured consumer mode.
     - Impact: When running in SharedThroughput mode (such as with LocalStack), this caused nil pointer errors if the stream was unavailable or uninitialized.
     - Fix: Added a conditional check so that stream ARN resolution only occurs when KinesisConsumerMode is set to ExtendedFanout.

These changes improve stability and maintainability of the Kinesis input binding, ensuring compatibility with LocalStack and future AWS SDK updates.

## Issue reference

Please reference the issue this PR will close: #_[3980, 3985]_

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
    * [X] Created the dapr/docs PR: <[insert PR link here](https://github.com/dapr/docs/pull/4755)>


bindings/aws/kinesis/kinesis.go
```diff
@@ -50,9 +50,10 @@ type AWSKinesis struct {
 	logger       logger.Logger
 	consumerMode string
 
-	closed  atomic.Bool
-	closeCh chan struct{}
-	wg      sync.WaitGroup
+	closed          atomic.Bool
+	closeCh         chan struct{}
+	wg              sync.WaitGroup
+	applicationName string
```
sicoyle: can you please share some insights on why the additional field here?

common/authentication/aws/client.go
```diff
@@ -181,27 +182,32 @@ func (c *KinesisClients) Stream(ctx context.Context, streamName string) (*string
...
 	if c.Kinesis != nil {
 		if mode == sharedMode {
-			if c.Credentials != nil {
-				kclConfig := config.NewKinesisClientLibConfigWithCredential(consumer,
-					stream, c.Region, consumer,
-					c.Credentials)
-				return kclConfig
+			v1Creds, err := c.Credentials.Get()
+			if err != nil {
+				return nil
```
sicoyle: why are we ignoring the err here? If we are updating to use v2 then can you make the v2 creds provider be the default and if err then maybe try the v1 and then err from there? but pls reference the other aws components to see the auth flow so we can ensure that we standardize on this for when using v2 sdks.

sicoyle (commented):
thank you for your contribution to the Dapr project 🙌 - few comments :)

bindings/aws/kinesis/kinesis.go
swatimodi-scout: > can you please share some insights on why the additional field here?

applicationName is required for KCL (Kinesis Client Library) worker configuration in shared throughput mode. It identifies the consumer application and is used for DynamoDB table naming and checkpointing.

Without applicationName we were facing an error.

common/authentication/aws/client.go
swatimodi-scout: > why are we ignoring the err here? If we are updating to use v2 then can you make the v2 creds provider be the default and if err then maybe try the v1 and then err from there? but pls reference the other aws components to see the auth flow so we can ensure that we standardize on this for when using v2 sdks.

Done please verify this.

acroca:
The git history of this PR seems odd, lots of those commits are unrelated. Do you mind cleaning it up so it only contains commits related to this change?

swatimodi-scout:
> The git history of this PR seems odd, lots of those commits are unrelated. Do you mind cleaning it up so it only contains commits related to this change?

@acroca Thanks for the feedback! I’ve cleaned up the branch history — the PR now contains only the relevant commit for this change. Please let me know if anything else needs adjustment.

bindings/aws/kinesis/kinesis.go
```diff
@@ -158,19 +163,21 @@ func (a *AWSKinesis) Read(ctx context.Context, handler bindings.Handler) (err er
...
+	case SharedThroughput:
+		// initalize worker configuration
+		config := a.authProvider.Kinesis().WorkerCfg(ctx, a.streamName, a.metadata.Region, a.consumerMode, a.applicationName)
 		// Configure the KCL worker with custom endpoints for LocalStack
-		config := a.authProvider.Kinesis().WorkerCfg(ctx, a.streamName, a.consumerName, a.consumerMode)
 		if a.metadata.Endpoint != "" {
-			config.KinesisEndpoint = a.metadata.Endpoint
-			config.DynamoDBEndpoint = a.metadata.Endpoint
+			config.WithKinesisEndpoint(a.metadata.Endpoint)
+			config.WithDynamoDBEndpoint(a.metadata.Endpoint)
```
acroca: Even though this is correct, I think we mostly rely on the returned value from the `WithX` instead of assuming it changes its internal state.
```suggestion
			config = config.WithKinesisEndpoint(a.metadata.Endpoint)
			config = config.WithDynamoDBEndpoint(a.metadata.Endpoint)
```

bindings/aws/kinesis/kinesis.go
acroca: I'd move this field next to `consumerMode`. And the comment is not necessary if you ask me, I'd remove it for consistency, but feel free to keep if you think it's useful :)

common/authentication/aws/client.go
```diff
@@ -181,27 +183,41 @@ func (c *KinesisClients) Stream(ctx context.Context, streamName string) (*string
 		stream, err := c.Kinesis.DescribeStreamWithContext(ctx, &kinesis.DescribeStreamInput{
 			StreamName: aws.String(streamName),
 		})
-		if stream != nil {
-			return stream.StreamDescription.StreamARN, err
+		/**
+		 * If the error is not nil, do not proceed to the next step
+		 * as it may cause a nil pointer error on stream.StreamDescription.StreamARN.
+		 */
```
acroca: I think the comment is not necessary

common/authentication/aws/client.go
```diff
@@ -181,27 +183,41 @@ func (c *KinesisClients) Stream(ctx context.Context, streamName string) (*string
...
+		if err != nil {
+			return nil, err
 		}
+		return stream.StreamDescription.StreamARN, err
```
acroca:
```suggestion
		return stream.StreamDescription.StreamARN, nil
```

acroca: Also, is it possible for `stream` to be nil at this point?

common/authentication/aws/client.go
```diff
@@ -181,27 +183,41 @@ func (c *KinesisClients) Stream(ctx context.Context, streamName string) (*string
...
-				kclConfig := config.NewKinesisClientLibConfigWithCredential(consumer,
-					stream, c.Region, consumer,
-					c.Credentials)
+			// Try v2 default config first (standard approach for v2 components)
+			v2Config, err := awsv2config.LoadDefaultConfig(ctx, awsv2config.WithRegion(region))
+			if err == nil {
+				kclConfig := config.NewKinesisClientLibConfigWithCredential(applicationName, stream, region, "", v2Config.Credentials)
 				return kclConfig
 			}
+			// Fallback to v1 credentials if v2 fails
```
acroca: In which cases would v2 fail, but v1 work?

common/authentication/aws/client.go
swatimodi-scout: > I think the comment is not necessary

Done

common/authentication/aws/client.go
swatimodi-scout: > In which cases would v2 fail, but v1 work?

@acroca 

Good question! You're right to point that out. Looking at this more carefully, the v2 fallback to v1 scenario is actually quite rare and adds unnecessary complexity.

In practice, both v1 and v2 SDKs use the same underlying credential sources (env vars, credential files, IAM roles, etc.), so if v2 fails to load credentials, v1 would likely fail for the same reason.

Since we already have validated v1 credentials from the established session, I've simplified this to directly use those credentials and convert them to v2 format for KCL compatibility. This is more reliable and removes the redundant credential resolution logic.

Updated the code to use the existing session credentials directly - much cleaner approach! 👍

bindings/aws/kinesis/kinesis.go
swatimodi-scout: > I'd move this field next to `consumerMode`. And the comment is not necessary if you ask me, I'd remove it for consistency, but feel free to keep if you think it's useful :)

Done

common/authentication/aws/client.go
```diff
@@ -181,27 +183,25 @@ func (c *KinesisClients) Stream(ctx context.Context, streamName string) (*string
...
-					stream, c.Region, consumer,
-					c.Credentials)
-				return kclConfig
-			}
+	if c.Kinesis != nil && mode == sharedMode {
+		v1Creds, err := c.Credentials.Get()
+		if err != nil {
+			return nil
 		}
+		v2Creds := v2creds.NewStaticCredentialsProvider(v1Creds.AccessKeyID, v1Creds.SecretAccessKey, v1Creds.SessionToken)
```
acroca: It feels wrong to have both v1 and v2 living together. Can't we migrate completely to v2?

swatimodi-scout: > It feels wrong to have both v1 and v2 living together. Can't we migrate completely to v2?

Done please review it.

acroca: I don't think this works, it's not wired up to the refresh at all.
I think it was better before, but the conversion from v1 credentials to v2 should be done in the `KinesisClients.New`, which is called in the refresh operations.
This way, you can convert all kinesis to v2 and only do the v1->v2 in the `New` right?

javier-aliaga: thansks @swatimodi-scout! What if you fully migrate the aws client to v2? you do not need to migrate all other components, it would be something similar to this [PR](https://github.com/dapr/components-contrib/pull/3911/files) that @mikeee is working on.

swatimodi-scout: > thansks @swatimodi-scout! What if you fully migrate the aws client to v2? you do not need to migrate all other components, it would be something similar to this [PR](https://github.com/dapr/components-contrib/pull/3911/files) that @mikeee is working on.

@javier-aliaga Yes, that makes sense. However, we’re not migrating the entire AWS SDK version at this stage — my current change focuses only on updating vmware-go-kcl to vmware-go-kcl-v2.

Would you prefer that I continue implementing the AWS SDK v2 changes for Kinesis within this same PR, or should I move that work to a separate PR?

swatimodi-scout: > I don't think this works, it's not wired up to the refresh at all. I think it was better before, but the conversion from v1 credentials to v2 should be done in the `KinesisClients.New`, which is called in the refresh operations. This way, you can convert all kinesis to v2 and only do the v1->v2 in the `New` right?

@acroca I updated in KinesisClients.New. It is working i checked in local. Let me know if i need to revert back to previous version where we were supporting both v1 and v2.

javier-aliaga: I think you can do it in the same PR as changing to aws-sdk-v2 is related to update the vmware-go-kcl-v2

javier-aliaga: @mikeee  I updated the aws-sdk-v2 issue to track it  
https://github.com/dapr/components-contrib/issues/3896

swatimodi-scout: > I think you can do it in the same PR as changing to aws-sdk-v2 is related to update the vmware-go-kcl-v2

@javier-aliaga  Please review the code.

common/authentication/aws/client.go
```diff
@@ -116,9 +117,10 @@ type ParameterStoreClients struct {
 }
 
 type KinesisClients struct {
-	Kinesis     kinesisiface.KinesisAPI
-	Region      string
-	Credentials *credentials.Credentials
+	Kinesis       *kinesisv2.Client
+	Region        string
+	Credentials   *credentials.Credentials
+	V2Credentials aws.CredentialsProvider
```
javier-aliaga: To migrate to aws sdk V2 move this out to the kinesis.go file and create and store the client there. No references to sdk v2 should be here

bindings/aws/kinesis/kinesis.go
```diff
@@ -44,15 +43,15 @@ type AWSKinesis struct {
...
+	streamName      string
+	consumerName    string
+	consumerARN     *string
+	logger          logger.Logger
+	consumerMode    string
+	applicationName string
+	closed          atomic.Bool
+	closeCh         chan struct{}
+	wg              sync.WaitGroup
 }
```
javier-aliaga, line 59: Add kinesis client here and init with sdk v2

common/authentication/aws/client.go
swatimodi-scout: > To migrate to aws sdk V2 move this out to the kinesis.go file and create and store the client there. No references to sdk v2 should be here

@javier-aliaga All AWS resource clients are currently defined in client.go. Is there any specific reason we should move only the Kinesis client out of this file?

swatimodi-scout: > To migrate to aws sdk V2 move this out to the kinesis.go file and create and store the client there. No references to sdk v2 should be here

Done please review it

bindings/aws/kinesis/kinesis.go
swatimodi-scout: > Add kinesis client here and init with sdk v2

Done

bindings/aws/kinesis/kinesis.go
```diff
@@ -42,17 +45,19 @@ type AWSKinesis struct {
 	authProvider awsAuth.Provider
```
javier-aliaga: Do we still need the authProvider?

bindings/aws/kinesis/kinesis.go
```diff
@@ -132,6 +139,12 @@ func (a *AWSKinesis) Init(ctx context.Context, metadata bindings.Metadata) error
 		return err
 	}
 	a.authProvider = provider
```
javier-aliaga: Remove the authProvider and create config v2 directly like https://github.com/dapr/components-contrib/blob/d0e55205075140abfa656358d6d2a470edc2d183/common/component/kafka/kafka.go#L205C26-L205C37

bindings/aws/kinesis/kinesis.go
```diff
@@ -414,6 +415,46 @@ func (p *recordProcessor) Shutdown(input *interfaces.ShutdownInput) {
 	}
 }
 
+func (a *AWSKinesis) createKinesisClient(ctx context.Context) error {
```
javier-aliaga: This method probably need to receive the aws config and return the kinesis client. Assign the client in the Init method

rideshnath-scout: Done, please check

swatimodi-scout: @javier-aliaga  Is there anything that we need to fix on our side for lint or test failing?

go.mod
```diff
@@ -150,6 +150,11 @@ require (
 	sigs.k8s.io/yaml v1.4.0
 )
 
+require (
+	github.com/aws/aws-sdk-go-v2/service/kinesis v1.42.3
+	github.com/awslabs/kinesis-aggregation/go/v2 v2.0.0-20211222152315-953b66f67407 // indirect
+)
+
```
mikeee, line 161: Can you please run `make modtidy-all`

common/authentication/aws/client.go
mikeee, line 1: I think these changes to the old auth package should be reverted, as they are not necessary as part of this PR unless a full cleanup takes place.

bindings/aws/kinesis/kinesis.go
```diff
@@ -22,37 +22,41 @@ import (
...
 	"time"
 
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/request"
-	"github.com/aws/aws-sdk-go/service/kinesis"
+	awsv2 "github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/kinesis"
+	"github.com/aws/aws-sdk-go-v2/service/kinesis/types"
 	"github.com/cenkalti/backoff/v4"
+	aws "github.com/dapr/components-contrib/common/aws"
```
mikeee: Could this alias be changed to awsCommon?

bindings/aws/kinesis/kinesis.go
```diff
@@ -22,37 +22,41 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/request"
-	"github.com/aws/aws-sdk-go/service/kinesis"
+	awsv2 "github.com/aws/aws-sdk-go-v2/aws"
```
mikeee: I think this alias should be removed for clarity

bindings/aws/kinesis/kinesis.go
```diff
@@ -22,37 +22,41 @@ import (
...
-	streamName   string
-	consumerName string
-	consumerARN  *string
-	logger       logger.Logger
-	consumerMode string
-
-	closed  atomic.Bool
-	closeCh chan struct{}
-	wg      sync.WaitGroup
+	// authProvider awsAuth.Provider
```
mikeee:
```suggestion
```

bindings/aws/kinesis/kinesis.go
rideshnath-scout: Done

go.mod
rideshnath-scout: Done

swatimodi-scout:
@mikeee @javier-aliaga Requested changes are done. Please review it.

javier-aliaga:
@swatimodi-scout linting is failing, can you fix it?

rideshnath-scout:
> @swatimodi-scout linting is failing, can you fix it?

@javier-aliaga I have fixed the lint issue, please check

mikeee (changes requested):
~~Linter issue - complaining of the need to run `goimports` at the least (or `golangci-lint run --fix`)~~

The changes to `/common/authentication/aws` should also be reverted if the package is not being referenced in this PR.

swatimodi-scout:
@mikeee @acroca  Is there anything required from our side now?

swatimodi-scout:
@mikeee @acroca  Could you please merge this PR if everything looks good to you?

mikeee (changes requested):
My last review comment still stands

swatimodi-scout:
> The changes to `/common/authentication/aws` should also be reverted if the package is not being referenced in this PR.

Are you talking about this change? - The changes to /common/authentication/aws should also be reverted if the package is not being referenced in this PR.

rideshnath-scout:
> /common/authentication/aws

@mikeee I have reverted the changes.

swatimodi-scout:
> My last review comment still stands

Done please review the changes

bindings/aws/kinesis/kinesis.go
```diff
@@ -332,34 +338,19 @@ func (a *AWSKinesis) deregisterConsumer(ctx context.Context, streamARN *string,
...
+		}
+		if consumer.ConsumerDescription.ConsumerStatus == types.ConsumerStatusActive {
+			return nil
+		}
+		time.Sleep(10 * time.Second)
 	}
-	w.ApplyOptions(opts...)
-
-	return w.WaitWithContext(ctx)
+	return errors.New("consumer did not become active within timeout")
```
mikeee: Q:

I'm not sure on this but I believe all the aws service clients use the a default number of retries built-in which I'm not aware of us customising.

What is the significance of the magic number (18) used and does this mean for these requests the Nopretryer should be used here, and are we iterating more than the intended number of times?

rideshnath-scout: @mikeee You're absolutely right to question this! The magic number 18 appears to be implementing a custom polling mechanism.
Let me fix this by implementing a proper polling mechanism with exponential backoff instead of the this.

rideshnath-scout: @mikeee Could you please check and let me know if any changes are needed ?

mikeee (dismissed):
Lgtm, just noting that I haven't got an environment set up to test this and we don't have certification/conformance test coverage for this binding. A bonus definitely would be if you could write some e2e tests I can set up the resources on our aws account for them.

https://github.com/dapr/components-contrib/tree/main/tests/certification/bindings/aws/s3

swatimodi-scout:
> Lgtm, just noting that I haven't got an environment set up to test this and we don't have certification/conformance test coverage for this binding. A bonus definitely would be if you could write some e2e tests I can set up the resources on our aws account for them.
> 
> https://github.com/dapr/components-contrib/tree/main/tests/certification/bindings/aws/s3

@mikeee Thanks for the review. We’ve tested the changes locally, and everything is working as expected. Would it be possible to merge this PR at the earliest? We can follow up by adding the e2e test cases separately afterward.

mikeee (approved):

swatimodi-scout:
@javier-aliaga  Can we now merge this PR?

javier-aliaga (commented):
LGTM. Still this will not be available until 1.17 (planned for January)

swatimodi-scout:
> LGTM. Still this will not be available until 1.17 (planned for January)

Got it, thanks for the update.

sicoyle:
@acroca and @javier-aliaga any last feedback here before merge?

PR #4087, opened 2025-11-04
State: KeysLike

JoshVanL:
```go
// KeysLiker is an optional interface to list state keys with an
// optional SQL style wildcard pattern.
type KeysLiker interface {
	KeysLike(ctx context.Context, req *KeysLikeRequest) (*KeysLikeResponse, error)
}
```

```go
type KeysLikeRequest struct {
	// Pattern is the SQL LIKE pattern to match keys against.
	Pattern string `json:"pattern"`

	// ContinueToken is an optional parameter to indicate the key from which to
	// start the search.
	ContinueToken *string `json:"startKey,omitempty"`

	// PageSize is an optional parameter to indicate the maximum number of keys
	// to return.
	PageSize *uint32 `json:"pageSize,omitempty"`
}
```

```go
// KeysLikeResponse is the response object for getting keys like a pattern.
type KeysLikeResponse struct {
	Keys []string `json:"keys"`

	// ContinueToken is an optional token which can be used to continue the
	// search of keys. Usually only present if a `PageSize` was set on the
	// request.
	ContinueToken *string
}
```

PR adds a new state store API `KeysLike`, which allows listing keys
which match a given SQL LIKE style wildcard pattern ("%" and "_"). This
API is implemented currently solely to enable the workflow instance
listing functionality in Dapr, however could be exposed to the general
state APIs in future.

The request takes and requires a input `Pattern` field which is used to
match keys on. Also accepts an optional `PageSize` to limit the number of
results returned, and a `ContinueToken` to continue listing from a
previous request page.

Only returns the keys as a string slice. This function does not return
any values. Naturally, some state stores have better/more efficient
support for SQL LIKE style key queries.

Adds support for state stores:
- PostgreSQL (v1 and v2)
- MySQL
- SQLite
- SQL Server
- CockroachDB
- MongoDB
- Redis
- etcd
- In-Memory

Notes:

Both postgres v1 and v2 have a new migration to add a `row_id` column,
which is a unique identifier for each row. This is used to page results
efficiently and consistently.

The in-memory store now tracks an internal id for each key/value pair to
allow consistent paging when keys are added/removed during listing.

MySQL also introduces a new `row_id` column in migration for the same
purpose.

A new state feature `FeatureKeysLike` has been added to signal
`KeysLike` API support.

A conformance test has been added to test the `KeysLike` API for all
supported state stores.

### DIFF
--- a/state/cockroachdb/cockroachdb.go
+++ b/state/cockroachdb/cockroachdb.go
@@ -159,8 +159,8 @@ func ensureTables(ctx context.Context, db pginterfaces.PGXPoolConn, opts postgre
 	if !exists {
 		opts.Logger.Info("Creating CockroachDB metadata table")
 		_, err = db.Exec(ctx, fmt.Sprintf(`CREATE TABLE %s (
-  key   text NOT NULL PRIMARY KEY,
-  value text NOT NULL
+key   text NOT NULL PRIMARY KEY,
+value text NOT NULL
 );`, opts.MetadataTableName))
 		if err != nil {
 			return err


state/requests.go
```diff
@@ -161,3 +161,16 @@ type QueryRequest struct {
...
+	// Pattern is the SQL LIKE pattern to match keys against.
+	Pattern string `json:"pattern"`
+
+	// ContinueToken is an optional parameter to indicate the key from which to
+	// start the search.
+	ContinueToken *string `json:"startKey,omitempty"`
+
+	// PageSize is an optional parameter to indicate the maximum number of keys
+	// to return.
+	PageSize *uint32 `json:"pageSize,omitempty"`
```
acroca, line 175: I think we should set a maximum `PageSize` as there can be potentially lots of keys to return.

state/etcd/etcd.go
```diff
@@ -448,3 +450,180 @@ func NewTLSConfig(clientCert, clientKey, caCert string) (*tls.Config, error) {
...
+	// Build the etcd key prefix we need to scan, using the literal prefix
+	// (up to the first unescaped % or _) to keep scans narrow.
+	userPrefix := likeLiteralPrefix(req.Pattern)
+	etcdPrefix := strings.TrimSuffix(e.keyPrefixPath, "/") + "/" + userPrefix
+
+	// Fetch keys under that etcd prefix
+	// (we read values too to safely get revisions; KeysOnly omits CreateRevision on some clients).
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+	resp, err := e.client.Get(ctx, etcdPrefix, clientv3.WithPrefix())
```
acroca: Can we use `.WithLimit`? 
Also it'd be good to use `.WithFromKey` but I think we have to chose either `WithPrefix` or `WithFromKey`...

state/mysql/mysql_test.go
```diff
@@ -929,3 +931,8 @@ func TestValidIdentifier(t *testing.T) {
 		})
 	}
 }
+
+func Test_KeysLike(t *testing.T) {
+	m, _ := mockDatabase(t)
+	var _ state.KeysLiker = m.mySQL
```
acroca, line 937: For consistency with the postgres check:
```suggestion
	_, ok := m.mySQL.(state.KeysLiker)
	require.True(t, ok)
```

javier-aliaga:
LGTM

state/requests.go
JoshVanL: Sure, but I don't think that this belongs in contrib. It is a runtime concern. We have a default limit set in durabletask-go 

https://github.com/dapr/durabletask-go/pull/51/files#diff-2b0fbbb90d13e711efdcf3c8ba845ed7d2200f1f7b5b7db0b5812b6022ef9720R242

state/mysql/mysql_test.go
JoshVanL: We can't do this because `m.mySQL` is not an interface.

common/component/postgresql/v1/postgresql.go
```diff
@@ -531,3 +533,88 @@ func (p *PostgreSQL) GetComponentMetadata() (metadataInfo metadata.MetadataMap)
...
+
+	// Match with backslash-escaping for % and _
+	where := []string{
+		`key LIKE $1 ESCAPE '\'`,
+		`(expiredate IS NULL OR expiredate > CURRENT_TIMESTAMP)`,
+	}
+	args := []any{req.Pattern}
+
+	// Pagination: resume strictly AFTER the last returned row_id
+	if req.ContinueToken != nil && *req.ContinueToken != "" {
```
cicoyle: nit: in the protos in the proposal and durabletask we use `continuationToken` can we switch to that instead of `continue` for consistency?

state/cockroachdb/cockroachdb.go
```diff
@@ -78,42 +79,88 @@ func ensureTables(ctx context.Context, db pginterfaces.PGXPoolConn, opts postgre
...
 	if err != nil {
 		return err
 	}
-
 	if !exists {
 		opts.Logger.Info("Creating CockroachDB metadata table")
 		_, err = db.Exec(ctx, fmt.Sprintf(`CREATE TABLE %s (
-			key text NOT NULL PRIMARY KEY,
-			value text NOT NULL
+  key   text NOT NULL PRIMARY KEY,
```
cicoyle: nit the spacing looks off here

state/postgresql/v1/migrations.go
```diff
@@ -79,8 +81,119 @@ func performMigrations(ctx context.Context, db pginterfaces.PGXPoolConn, opts po
...
+		func(ctx context.Context) error {
+			opts.Logger.Infof("Ensuring row_id (identity) exists on '%s'", opts.StateTableName)
+
+			// Resolve schema+table and quoted FQ table name
+			var schema, table, fqtnQI, regName string
+			if err := db.QueryRow(ctx, `
+        SELECT n.nspname,
+               c.relname,
+               format('%I.%I', n.nspname, c.relname) AS fqtn_quoted,
+               n.nspname || '.' || c.relname         AS reg_name
```
cicoyle, line 98: nit: spacing after `c.relname`

state/postgresql/v2/postgresql.go
```diff
@@ -209,6 +211,113 @@ CREATE INDEX ON %[1]s (expires_at);
...
+		func(ctx context.Context) error {
+			p.logger.Infof("Ensuring row_id (identity) exists on '%s'", stateTable)
+
+			// Resolve schema + table and a safely quoted FQ table name
+			var schema, table, fqtnQI, regName string
+			if err := p.db.QueryRow(ctx, `
+        SELECT n.nspname,
+               c.relname,
+               format('%I.%I', n.nspname, c.relname) AS fqtn_quoted,
+               n.nspname || '.' || c.relname         AS reg_name
```
cicoyle, line 225: same nit here on spacing

cicoyle (approved):

PR #4099, opened 2025-11-25
Add `kid` header to kafka jwt auth

acroca:
Adds the possibility to configure the optional `kid` header for the JWT authentication flow in the kafka components.

### DIFF
--- a/common/component/kafka/sasl_oauthbearer_private_key_jwt.go
+++ b/common/component/kafka/sasl_oauthbearer_private_key_jwt.go
@@ -172,7 +172,7 @@ func (ts *OAuthTokenSourcePrivateKeyJWT) Token() (*sarama.AccessToken, error) {
 	var signOptions []jwt.Option
 	if ts.Kid != "" {
 		headers := jws.NewHeaders()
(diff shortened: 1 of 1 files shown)


common/component/kafka/sasl_oauthbearer_private_key_jwt.go
```diff
@@ -166,7 +169,15 @@ func (ts *OAuthTokenSourcePrivateKeyJWT) Token() (*sarama.AccessToken, error) {
 		return nil, fmt.Errorf("failed to build token: %w", err)
 	}
 
-	assertion, err := jwt.Sign(token, jwt.WithKey(jwa.RS256, rsaKey))
+	var signOptions []jwt.Option
+	if ts.Kid != "" {
+		headers := jws.NewHeaders()
+		if err := headers.Set("kid", ts.Kid); err != nil {
```
mikeee:
```suggestion
		if err = headers.Set("kid", ts.Kid); err != nil {
```

I think it's pretty unambiguous and doesn't require another named error here, just a nit to remove the shadowing

acroca:
@mikeee can you merge? I don't think I should as an approver.

PR #4105, opened 2025-12-03
ci: add create release wf

nelson-parente:
This pull request introduces automation for the release process by adding a GitHub Actions workflow and a supporting shell script, along with documentation to guide maintainers through the new workflow. The changes streamline how release branches and tags are created, reducing manual steps and potential errors.

**Release automation:**

* Added `.github/workflows/create-release.yaml` to define a GitHub Actions workflow that creates release branches and tags based on user input, including support for pre-releases and stable releases.
* Added `.github/scripts/create-release.sh`, a shell script that validates semantic versioning, creates the appropriate release branch and tag, and pushes them to the repository. Handles alpha releases and checks for existing branches/tags.

**Documentation:**

* Added `RELEASE.md` to document the new release process, describing how to use the workflow, steps for pre-releases, stable releases, and patch releases, and linking to broader project guidelines.


mikeee (changes requested):
I think there needs to be open discussion and planning for a release pipeline to de-risk the cutting of releases. Although this PR does introduce some reproducability into releases, I don't think a pipeline should reside in this repository.

https://github.com/dapr/community/issues/320

nelson-parente:
> I think there needs to be open discussion and planning for a release pipeline to de-risk the cutting of releases. Although this PR does introduce some reproducability into releases, I don't think a pipeline should reside in this repository.
> 
> [dapr/community#320](https://github.com/dapr/community/issues/320)

The goal is not to build a full release pipeline here. It just removes the manual tag step and makes the release flow match what we already do in dapr/dapr. It is only standardizing the existing process so releases are easier and less error-prone.

sicoyle:
>  I don't think a pipeline should reside in this repository.

@mikeee can you please share your thoughts a bit more here as to why?

.github/workflows/create-release.yaml
```diff
@@ -0,0 +1,49 @@
...
+          fetch-depth: 0
+      - name: Install required packages
+        run: |
+          sudo apt-get update
+          sudo apt-get install pcre2-utils
+      - name: Create release branch and tag
+        env:
+          GITHUB_TOKEN: ${{ secrets.DAPR_BOT_TOKEN }}
+        run: |
+          git config user.email "daprweb@diagrid.io"
```
mikeee:
```suggestion
          git config user.email "dapr@dapr.io"
```

.github/workflows/create-release.yaml
```diff
@@ -0,0 +1,49 @@
...
+    inputs:
+      rel_version:
+        description: 'Release version (examples: 1.16.0-rc.1, 1.16.0)'
+        required: true 
+        type: string
+
+permissions: {}
+
+jobs:
+  create-release:
```
mikeee, line 27: Could we consider gating the workflow dispatch to maintainers and the release team?

https://github.com/tspascoal/get-user-teams-membership Here's an action that I would consider reasonable to implement, wdyt?

mikeee (dismissed):
I think the release process needs to be well defined and tested. This also increases the attack surface and is detrimental to the project's security posture.

After some discussion yesterday I am happy to revise my stance and do want to push this forwards as an intermediary step albeit with a small suggestion for your consideration.

.github/workflows/create-release.yaml
nelson-parente: I think that already happens today, only certain people can run workflows

.github/workflows/create-release.yaml
nelson-parente: fixed!

.github/workflows/create-release.yaml
mikeee: Unfortunately the net is pretty wide with who can trigger workflows, the only actors that should be permitted to run these workflows are humans and a member of one of those groups (maintainers & release) for visibility purposes, this is not the case. Workflows are inherently problematic with a severe lack of accountability and having this would enhance our security posture in my opinion.

RELEASE.md
```diff
@@ -0,0 +1,45 @@
...
+
+4. Test the produced build.
+
+5. If there are issues, fix them in the release branch and trigger the workflow again by creating a new pre-release tag (for example `-rc.1`).
+
+6. Repeat until the build is good.
+
+## Stable Release
+
+Create a stable tag without the rc suffix (for example `v1.11.0`).
```
cicoyle: There have been several times while releasing I've manually had to update the release to set it to `latest`. Can you add a line to verify this, so ppl always check?
```suggestion
Create a stable tag without the rc suffix (for example `v1.11.0`). Ensure the new release is set to `latest` and not a `pre-release`
```

cicoyle (commented):
I believe there is one unresolved comment from Mikeee that still needs to be addressed, but otherwise lgtm 👍🏻

JoshVanL (approved):

mikeee:
@nelson-parente https://github.com/dapr/components-contrib/pull/4105/changes#r2623678218 I'd still feel like this would be a good thing to have to maintain a good security posture, what are your thoughts on this?

cicoyle:
> @nelson-parente https://github.com/dapr/components-contrib/pull/4105/changes#r2623678218 I'd still feel like this would be a good thing to have to maintain a good security posture, what are your thoughts on this?

Hey @mikeee - Can this be made a new/separate issue from this PR? As this PR aligns the release process with the rest of the dapr repos, I dont think we should block this PR

nelson-parente:
> @nelson-parente https://github.com/dapr/components-contrib/pull/4105/changes#r2623678218 I'd still feel like this would be a good thing to have to maintain a good security posture, what are your thoughts on this?

I’m not opposed to strengthening our security posture around these actions. However, I agree with Cassie that this should be done separately. Today, these actions are executed manually by the same group of people who can already trigger the workflows, so from an access or privilege perspective, nothing really changes.

cicoyle (approved):

cicoyle:
Im going to merge this. Lets open an issue to address @mikeee comments in a separate PR.

mikeee (approved):
I'm fine with this being merged with a follow-up issue 👍

PR #4111, opened 2025-12-09
Send audience as a single string in the JWT

acroca:
Some IdPs don't support array of audiences even though [the JWT spec](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3) says it's valid.
Since both approaches are valid, sending a single string would support those IdPs too.

### DIFF
--- a/common/component/kafka/auth_test.go
+++ b/common/component/kafka/auth_test.go
@@ -142,8 +142,7 @@ func TestAuth(t *testing.T) {
 		var receivedAssertion string
 
 		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			err := r.ParseForm()
-			require.NoError(t, err)
+			require.NoError(t, r.ParseForm())
 			receivedAssertion = r.FormValue("client_assertion")
 
 			w.Header().Set("Content-Type", "application/json")


nelson-parente (dismissed):

sicoyle:
is there a test supporting this change somewhere? also do you have a corresponding docs PR for this behavior? Thanks :)

acroca:
@sicoyle since keycloak supports both array and single value, the only way to validate we're sending a single string would be to place a proxy in front of keycloak and extract the JWT. I find it a bit too complicated given that both approaches are valid. what do you think?

cicoyle:
thx @acroca - mind fixing the linter issues?

common/component/kafka/sasl_oauthbearer_private_key_jwt.go
```diff
@@ -169,6 +169,9 @@ func (ts *OAuthTokenSourcePrivateKeyJWT) Token() (*sarama.AccessToken, error) {
 		return nil, fmt.Errorf("failed to build token: %w", err)
 	}
 
+	// Some IdPs require the audience to be set as a single string
+	token.Options().Enable(jwt.FlattenAudience)
```
cicoyle, line 173: Was this entirely broken and requires using the `flattenaudience` as a fix by default? Im questioning if this should be a user controlled toggle or if we should truly change the default behavior - want to ensure we dont break users by changing the default behavior.

acroca: This whole functionality is fairly new and will be released for the first time in 1.17, I think it's safe to change it now. We can always make it configurable, but I don't think it'll be a problem for any client to receive it flattened, and the complexity cost of adding a new attribute in the component I don't think it's worth. I think it's safe to assume all clients support flattened audiences. If a customer has a problem with this, we can promote the decision to the component configuration, but I would not do it until someone actually needs it. What do you think?

yaron2 (approved):

Issue #4102, opened 2025-11-27
PubSub Pulsar: Allow for OIDC `clientSecret` to be rotated when token is refreshed


alicejgibbons:
## Describe the feature
Security policy requires rotating the Pulsar OAuth2 client secret (not just the token) via a file path. The current [Pulsar Go client](https://github.com/apache/pulsar-client-go) loads this file only once at startup, breaking authentication on subsequent token refreshes. The Pulsar team will add a fix to re-read the secret file on every token refresh operation. The Dapr Pulsar component then just needs to support passing a file path for the `privateKey` (clientSecret) metadata so that this file can be re-read on token refresh events and then re-generated.

Currently `clientSecret` only supports a value but by allowing this to read from a filePath then the privateKey can be rotated and then the token subsequently regenerated. 

## Release Note

RELEASE NOTE: **ADD**

alicejgibbons:
Can be done together with this one: https://github.com/dapr/components-contrib/issues/4083

cicoyle:
Both issues will be fixed in the [linked PR](https://github.com/dapr/components-contrib/pull/4115).

cicoyle:
[PR to contrib for 1.16](https://github.com/dapr/components-contrib/pull/4115)
[PR to contrib for master](https://github.com/dapr/components-contrib/pull/4116)
[docs for 1.16](https://github.com/dapr/docs/pull/4984)
[docs for 1.17](https://github.com/dapr/docs/pull/4985)

alicejgibbons:
Closing as completed in 1.16.5

Issue #4083, opened 2025-10-31
Add better debug logs for PubSub Pulsar component with Oauth


alicejgibbons:
## Describe the feature

Today there are no debug logs when the Pulsar Pubsub component creates or refreshes a token. The following should be added.

```
time="2025-10-29T14:39:00.676470087Z" level=debug msg="renewToken() - fetching new token from OAuth server..." app_id=publisher component="pulsar-pubsub (pubsub.pulsar/v1)" instance=publisher-7cb45f6d49-cwt5b scope=dapr.contrib type=log ver=1.16.4

time="2025-10-29T14:39:00.688544837Z" level=debug msg="Token renewed successfully, new expiry: 2025-10-29 14:39:59.688447462 +0000 UTC m=+180.738594889" app_id=publisher component="pulsar-pubsub (pubsub.pulsar/v1)" instance=publisher-7cb45f6d49-cwt5b scope=dapr.contrib type=log ver=1.16.4
```

## Release Note

RELEASE NOTE:

alicejgibbons:
@MyMirelHub

alicejgibbons:
This was completed in https://github.com/dapr/components-contrib/pull/4115

PR #4115, opened 2025-12-18
[1.16] Fix: Support pulsar secret from filepath + govuln

cicoyle:
[cherrypick this initial commit](https://github.com/dapr/components-contrib/pull/4113). cleanup and add tests

This PR also ups the version for `paho.mqtt.golang` as I noticed in [dapr/dapr a failing CI step due to a vuln.](https://github.com/dapr/dapr/actions/runs/20342668178/job/58446135204?pr=9245#step:6:12)

This PR should be cherrypicked to master once merged && dapr/dapr should be updated accordingly for both the release branch and master.

[Fixes this issue](https://github.com/dapr/components-contrib/issues/4102)
[Fixes this logging issue too](https://github.com/dapr/components-contrib/issues/4083) cc @alicejgibbons

### DIFF
--- a/common/authentication/oauth2/clientcredentials.go
+++ b/common/authentication/oauth2/clientcredentials.go
@@ -154,6 +154,6 @@ func (c *ClientCredentials) renewToken(ctx context.Context) error {
 	}
 
 	c.currentToken = token
-	c.log.Debugf("Oauth token renewed successfully, new expiry: %s", token.Expiry)
+	c.log.Debugf("OAuth token renewed successfully, new expiry: %s", token.Expiry)
 	return nil
 }


nelson-parente (dismissed):
🚀

pubsub/pulsar/metadata.yaml
```diff
@@ -31,6 +31,12 @@ authenticationProfiles:
         sensitive: true
         description: |
           The OAuth Client Secret.
+      - name: oauth2ClientSecretPath
+        type: string
+        description: |
+          The path to the OAuth Client Secret.
+        example: |
+          "/path/to/oauth2/client_secret.json"
```
sicoyle:
```suggestion
        example: "/path/to/oauth2/client_secret.json"
```

sicoyle: do we need to denote absolute vs relative?

pubsub/pulsar/pulsar.go
```diff
@@ -118,6 +118,8 @@ type Pulsar struct {
 	wg       sync.WaitGroup
 }
 
+var newPulsarClient = pulsar.NewClient
```
sicoyle: is there anyway you can update this to instead of using a package level global var to instead use a client factor that you can then inject accordingly in tests pls?

something like:
```go
type pulsarClientFactory func(pulsar.ClientOptions) (pulsar.Client, error)

type Pulsar struct {
	logger       logger.Logger
	newClientFn  pulsarClientFactory
}

func NewPulsar(l logger.Logger) pubsub.PubSub {
	return &Pulsar{
		logger:      l,
		newClientFn: pulsar.NewClient, // default
	}
}

```
so init would be something like
```go
client, err := p.newClientFn(options)
if err != nil {
    return fmt.Errorf("could not instantiate pulsar client: %w", err)
}
```
and tests could be updated to use this too

common/authentication/oauth2/clientcredentials.go
```diff
@@ -142,6 +143,7 @@ func (c *ClientCredentials) Token() (string, error) {
 }
 
 func (c *ClientCredentials) renewToken(ctx context.Context) error {
+	c.log.Debug("renewing token: fetching new token from OAuth server...")
```
cicoyle, line 146: open to suggestions for diff log lines here :)

common/authentication/oauth2/clientcredentials.go
```diff
@@ -152,5 +154,6 @@ func (c *ClientCredentials) renewToken(ctx context.Context) error {
 	}
 
 	c.currentToken = token
+	c.log.Debugf("Token renewed successfully, new expiry: %s", token.Expiry)
```
alicejgibbons:
```suggestion
	c.log.Debugf("Oauth token renewed successfully, new expiry: %s", token.Expiry)
```

common/authentication/oauth2/clientcredentials.go
```diff
@@ -152,5 +154,6 @@ func (c *ClientCredentials) renewToken(ctx context.Context) error {
 	}
 
 	c.currentToken = token
+	c.log.Debugf("Oauth token renewed successfully, new expiry: %s", token.Expiry)
```
sicoyle:
```suggestion
	c.log.Debugf("OAuth token renewed successfully, new expiry: %s", token.Expiry)
```

sicoyle (approved):
LGTM - thank you!

PR #4129, opened 2025-12-20
feat: add resp format and timeout to conversation api

sicoyle:
# Description

This PR adds:
- timeout support for llm providers on the http client and to prevent idle connections from being closed
- response format field - this is something that users rely on. In dapr agents we do a workaround for this that I want to rm with this addition
- i also had to bump langchain versions for some of the new options to be supported (even with the workarounds)

API PR: https://github.com/dapr/dapr/pull/9241

I need this PR merged first to then merge my dapr/dapr one.

I tested conformance test on: mistral, openai, anthropic. 
I have a few fixes left for ollama, and aws (this one was broken to begin with...).

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


conversation/echo/echo.go
```diff
@@ -62,12 +64,25 @@ func (e *Echo) GetComponentMetadata() (metadataInfo metadata.MetadataMap) {
...
+// Reasoning behind 4 char per token:
+// - LLM tokens are subword units, not individual characters
+// - Text averages ~4-5 chars per token
+// ref: https://help.openai.com/en/articles/4936856-what-are-tokens-and-how-to-count-them
+// We round up division to avoid undercounting tokens.
+func approximateTokensFromLength(textLength int) int64 {
+	if textLength == 0 {
+		return 0
+	}
+	return int64((textLength + 3) / 4)
```
CasperGN: Looking over the OpenAI example, would it be better to iterate over the words in the input instead?

sicoyle:
> @sicoyle does this (or the linked pr) also carry bubbling out the model, tokens etc into the metadata api like we discussed?

The model was the only thing that I can remember that we want bubbled out into the metadata api. I can look to PR that later today or tomorrow :)

conversation/echo/echo.go
sicoyle: Thanks for the feedback! Sure, I can adjust. Echo provider is not the mostttttt useful so I just put something here for starters

conversation/aws/bedrock/bedrock.go
```diff
@@ -42,8 +42,9 @@ type AWSBedrockMetadata struct {
 	AccessKey    string `json:"accessKey"`
 	SecretKey    string `json:"secretKey"`
 	SessionToken string `json:"sessionToken"`
-	Model        string `json:"model"`
-	CacheTTL     string `json:"cacheTTL"`
+
+	Model            string `json:"model"`
+	ResponseCacheTTL string `json:"responseCacheTTL" mapstructure:"responseCacheTTL" mapstructurealiases:"cacheTTL"`
```
sicoyle: this field in component metadata used to be cacheTTL and I am updating to be more reflective that it is responseCacheTTL now while still supporting what it once was.

conversation/converse.go
```diff
@@ -36,23 +36,29 @@ type Conversation interface {
...
-	ToolChoice          *string
-	Parameters          map[string]*anypb.Any `json:"parameters"`
-	ConversationContext string                `json:"conversationContext"`
-	Temperature         float64               `json:"temperature"`
-
-	// from metadata
-	Key       string   `json:"key"`
-	Model     string   `json:"model"`
-	Endpoints []string `json:"endpoints"`
-	Policy    string   `json:"loadBalancingPolicy"`
```
sicoyle: The “from metadata” fields were bubbling the Metadata field from the Conversations API (see https://github.com/dapr/dapr/blob/master/dapr/proto/runtime/v1/ai.proto#L43
), which allowed all of these fields to be passed through the request. This is an incorrect use of metadata. Components already expose their own metadata field, so there’s no need to bubble this information up from the API—contrib already handles it.

As a result, we can remove the following fields: Key, Endpoints, and Policy.

I also removed Parameters and ConversationContext, since these were surfaced but never actually used. All of this traces back to the original implementation, which I’m now cleaning up.

sicoyle:
ready for feedback please @cicoyle and @JoshVanL 🙏

sicoyle:
ignore this please as I am WIP splitting this into different PRs

conversation/anthropic/metadata.yaml
```diff
@@ -34,3 +34,17 @@ metadata:
...
+    type: duration
+    example: '30s'
+    default: '0s'
+  - name: idleConnectionTimeout
+    required: false
+    description: |
+      Maximum duration an idle HTTP connection can remain open before being closed.
+    type: duration
+    example: '30s'
+    default: '90s'
```
JoshVanL: Should these not be controlled by Dapr Component resiliency policies? 

https://docs.dapr.io/getting-started/quickstarts/resiliency/resiliency-state-quickstart/

conversation/langchaingokit/model.go
```diff
@@ -37,43 +41,61 @@ func (a *LLM) Converse(ctx context.Context, r *conversation.Request) (res *conve
 
 	resp, err := a.GenerateContent(ctx, messages, opts...)
 	if err != nil {
+		if ctx.Err() == context.DeadlineExceeded {
+			return nil, context.DeadlineExceeded
+		}
```
JoshVanL: noop

```suggestion
```

conversation/anthropic/metadata.yaml
sicoyle: Resiliency policies do not control httpClientTimeout and idleConnectionTimeout. These are component-level transport settings that configure the HTTP client used by langchain that then propagates to the LLM provider. Resiliency policies wrap the component operation as a whole, but they don't configure the component's internal HTTP client within Langchain.

I need to configure the underlying HTTP clients to address end user feedback on Dapr Agents.

JoshVanL: What difference does setting the HTTP request timeout on the the component spec, vs setting the resiliency policy timeout here?

```
	resp, err := policyRunner(func(ctx context.Context) (*conversation.Response, error) {
		rResp, rErr := component.Converse(ctx, requestv2)
		return rResp, rErr
	})
```

sicoyle: There are many components in this repo that do not bubble resiliency policy timeouts down to the underlying clients. From a quick glance, for example:

- HTTP binding: responseTimeout

- Crypto JWKS: requestTimeout

- Crypto Azure Key Vault: requestTimeout

- Cloudflare Workers KV: timeoutInSeconds

- Postgres: timeout

- Azure App Configuration: requestTimeout

These fields configure low-level transport timeouts (HTTP clients, database drivers, etc.) that operate independently of the operation-level resiliency policies, which wrap the entire component method call.

The conversation components are following this established precedent by configuring SDK HTTP client settings as needed. Both sets of configuration are required, and in particular, idle connection timeouts are not part of resiliency policies—and should not be—since they are component-specific concerns.

sicoyle:
ready for review again please @JoshVanL @cicoyle

conversation/anthropic/metadata.yaml
yaron2: This has been a long running effort in Dapr to replace component-specific retries and timeouts with resiliency policies. There were some PRs opened for several components to do that where it works. We've seen issues with users regarding double retries (resiliency policy and component retry) that led to a pretty bad user experience (RabbitMQ as an example). So, while some components do have specific retries and timeouts, this is considered more as tech debt. Concretely for this PR: we should allow adding timeouts to the component only if:

1. The timeout setting is critical for users to set
2. The timeout cannot be set by using a resiliency policy

sicoyle: I see. Thanks for the context - I didn't know that. I removed the http client timeout field but must keep the idle connection timeout as that cannot be set by the resiliency policy.

conversation/anthropic/metadata.yaml
```diff
@@ -34,3 +34,10 @@ metadata:
       A time-to-live value for a prompt cache to expire. Uses Golang durations
     type: string
     example: '10m'
+  - name: idleConnectionTimeout
+    required: false
+    description: |
+      Maximum duration an idle HTTP connection can remain open before being closed.
+    type: duration
+    example: '30s'
+    default: '90s'
```
JoshVanL: Why would a user want to set this to something other than a default of 30s?

sicoyle: Users have seen a timeout error where the conversation was ended due to long requests, such as those using Thinking capabilities. Even something like 90s would be low depending on a conversation request. This is needed to directly unblock dapr agent users who see their connections deemed idle and dropped

JoshVanL: Go HTTP transport deems a connection idle only when there is no request-response in progress but there is a open connection with keep alives. The idle connection timeout should not effect the length of conversation result responses as the request-response is active so the connection is not idle.

sicoyle: thanks for questioning this @JoshVanL . I added this based on timeout issues I heard of but didn't have the full err logs so went with my best guess. I actually think the resiliency policy timeout bubbling over is sufficient for what we need, and if I need anything else then can add another PR in follow up.

JoshVanL (dismissed):

sicoyle:
merging even with flakey state.mysql cert test as my PR has no impact on state components at all

PR #4118, opened 2025-12-19
chore(deps): add dependency scanning with dependabot for git actions and gomod ecosystems

CasperGN:
# Description

Currently components contrib does not have dependabot dependency scanning. This PR introduces scanning for `gomod` and `github-actions` ecosystems.
Note that it ignores patch version to lower the noise. This will still pick up security fixes for patch versions.

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: N/A

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation
    * [x] Created the dapr/docs PR: N/A

**Note:** We expect contributors to open a corresponding documentation PR in the [dapr/docs](https://github.com/dapr/docs/) repository. As the implementer, you are the best person to document your work! Implementation PRs will not be merged until the documentation PR is opened and ready for review.

PR #4131, opened 2025-12-22
Chore(deps): dependabot modtidy

CasperGN:
# Description

Add missing `go.mod` files for dependabot + include PR CI action for running `make modtidy-all` when dependabot update dependencies.

## Issue reference

Please reference the issue this PR will close: N/A - comes out of: https://github.com/dapr/components-contrib/pull/4118

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
    * [x] Created the dapr/docs PR: <insert PR link here>

### DIFF
--- a/.github/workflows/dependabot-tidy.yml
+++ b/.github/workflows/dependabot-tidy.yml
@@ -15,7 +15,7 @@ jobs:
             - name: Checkout code
               uses: actions/checkout@v6
             - name: Set up Go
-              uses: actions/setup-go@v4
+              uses: actions/setup-go@v6
               with:
                   go-version-file: 'go.mod'
             - name: Run go mod tidy for all modules


CasperGN:
cc @mikeee I believe this would make it happen as we discussed in https://github.com/dapr/components-contrib/pull/4118

I've taken inspiration from https://github.com/dependabot/dependabot-core/issues/5962#issuecomment-1303781931 as dependabot simply don't support that out of the box and by the looks of it, wont

.github/workflows/dependabot-tidy.yml
```diff
@@ -0,0 +1,31 @@
...
+
+jobs:
+    modtidy-all:
+        if: github.actor == 'dependabot[bot]'
+        runs-on: ubuntu-latest
+        steps:
+            - name: Checkout code
+              uses: actions/checkout@v6
+            - name: Set up Go
+              uses: actions/setup-go@v4
```
mikeee:
```suggestion
              uses: actions/setup-go@v6
```

.github/workflows/dependabot-tidy.yml
```diff
@@ -0,0 +1,31 @@
...
+              with:
+                  go-version: '1.24'
```
mikeee:
```suggestion
                  go-version-file: 'go.mod'
```

I don't think pinning the version here is best, this would use the go version from the root package

CasperGN: Agreed!

CasperGN:
> My concern is that dependabot will not maintain the PRs following the workflow being run / external commits being pushed to the merge request. Would this workflow cause this to happen?
> 
> Also the realitiy of this is that it is a workaround and that expensive workflows are being triggered, once on opening and another with the push from this workflow.
> 
> Is this a case that switching to renovate may need to be considered?

Regarding the first point:
I believe it will still maintain the PR and will still respond to @\dependabot commands. It's "just" branch updates.

The trigger should only hit when the actor is dependabot. That should handle not having recursion on triggers. It conforms with the other examples I've seen.

I get your point regarding renovate @mikeee - one of the reasons I'd favor to stick with dependabot, even with workaround, is consistency across dapr repos. My stance would be that if this isn't really working and causes too much overhead then let's swap.

mikeee:
Also, just out of curiousity - I would have expected a lot more depedency updates. Is there a rate limit?

CasperGN:
> Also, just out of curiousity - I would have expected a lot more depedency updates. Is there a rate limit?

I deliberately set it to ignore patch versions to remove a lot of noise so it'll "only" bump on Major/Minor. If there's security in patch versions those should run through as well. I believe that's why there's so few. I was kinda surprised too!

mikeee:
https://github.com/Azure/go-amqp is a package I don't see a PR for, should be a minor update v1.0.5 -> v1.5.0

mikeee:
Nevermind, looks like there is some form of a rate limit with some more deps being bumped.

PR #4191, opened 2026-01-21
Remove in-memory pubsub retires

JoshVanL:
Remove the retry logic from the in-memory pubsub implementation as retries are and should be handled completely by dapr.

### DIFF
--- a/pubsub/in-memory/in-memory.go
+++ b/pubsub/in-memory/in-memory.go
@@ -73,13 +73,14 @@ func (a *bus) Subscribe(ctx context.Context, req pubsub.SubscribeRequest, handle
 		return errors.New("component is closed")
 	}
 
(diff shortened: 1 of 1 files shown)


pubsub/in-memory/in-memory.go
```diff
@@ -74,20 +73,10 @@ func (a *bus) Subscribe(ctx context.Context, req pubsub.SubscribeRequest, handle
 		return errors.New("component is closed")
 	}
 
-	// For this component we allow built-in retries because it is backed by memory
 	retryHandler := func(data []byte, md map[string]string) {
```
acroca: Do we need this handler at all? If we do, maybe rename it, as it's not retrying anything.

JoshVanL: Renamed

Issue #4174, opened 2026-01-14
Data loss in redis stream pubsub


fsedano:
## Expected Behavior

When using a dapr pubsub component, I would expect it to:
- Respect resiliency policies
- Not having data loss in the event of a app pod restart

## Actual Behavior

To be able to set policies etc, we need to set processingTimeout and redeliverInterval to zero. This works well, but causes message loss if our pod restarts, since it seems data is pulled from redis and stored in the sidecar.

So, either we set them to zero and face data loss if the pod restarts, or set them to non-zero and we have to implement policies in the app side, making dapr useless for us.

## Steps to Reproduce the Problem

- Create a redis streams pubsub, using k8s
- Create a k8s deployment with a health check
- Set processingTimeout=0 and redeliverInterval=0
- Start publishing and consuming messages successfully
- Simulate a failure: Return RETRY for some of the messages and after some time, make the app unhealthy. Make sure k8s health checks catches it and restart the pod.
- When the pod comes online again, the messages app returned as RETRY won't be redelivered again, they are lost.

## Release Note

RELEASE NOTE:

atrauzzi:
Hey @fsedano! We went through and did some checks on our end:

- With metadata settings `processingTimeout=0` and `redeliverInterval=0`, we [don't expect to see any retries of failures](https://github.com/dapr/components-contrib/blob/20ff9672fae165cfa41138c92170e104f2f3e2a2/pubsub/redis/redis.go#L296-L297). This is a design decision as Dapr will not look for any PENDING messages on the Redis consumer group when these two are set.
  - This is separate to messages where the application returns `RETRY`. By _explicitly_ returning `RETRY` this overrides the Dapr/component metadata settings and consistently retries the call forever. This is also expected behaviour.
- For scenarios where `processingTimeout` and `redeliverInterval` are non-zero, and `redisMaxRetries` is non-zero, we did notice a bug where the `maxRetries` value was being exceeded.

If the `maxRetries` value is respected would this solve your problem?

fsedano:
Hi @atrauzzi 

The expectation is to have a setting that:
- Honors policies: We can hace a finite number of retries, a well defined timeouts etc
- No data loss in case pod restarts

What would be such a setting?

atrauzzi:
Hey @fsedano -- Alright!  We managed to put together a clean repro on our and and can see the scenario where the finite number of retries aren't being honoured by the resiliency policy.

fsedano:
[...]

PR #4208, opened 2026-02-02
Make Redis Streams PubSub honor resiliency policy

cicoyle:
[Fix this issue](https://github.com/dapr/components-contrib/issues/4174)

Change Redis Streams to always ACK after the handler returns (including error/RETRY/DROP) so Redis redelivery stops once the dapr resiliency policy has exhausted its retries.

Previously, the handler errors were un-ACKed, keeping entires in the pending entries list, causing infinite redelivery - breaking parity with in-mem pubsub, etc, and the configured retry policy.

In my testing I set 1 as the `maxRetries` and noted back the initial + 1 retry && when 2 was set as `maxRetries` I got back the initial + 2 retries - confirming its respecting resiliency policy. Previously, the msg would retry indefinitely regardless of resiliency policy.

NOTE: Users should clear their existing pending entries list to avoid replaying old msgs.

### DIFF
--- a/pubsub/redis/redis.go
+++ b/pubsub/redis/redis.go
@@ -245,13 +245,11 @@ func (r *redisStreams) processMessage(msg redisMessageWrapper) error {
 	}
 	if err := msg.handler(ctx, &msg.message); err != nil {
 		r.logger.Errorf("Error processing Redis message %s: %v", msg.messageID, err)
-		// always ack to stop Redis redelivery after the runtime resiliency policy has been applied
 		if ctx.Err() != nil {
 			// If the subscription context is cancelled (shutdown/timeout), skip ACK so Redis can redeliver after restart.
 			return err
 		}
-		// Use the background context in case subscriptionCtx is already closed, ie a race condition between this line and the one above.
-		if err := r.client.XAck(context.Background(), msg.message.Topic, r.clientSettings.ConsumerID, msg.messageID); err != nil {
+		if err := r.client.XAck(ctx, msg.message.Topic, r.clientSettings.ConsumerID, msg.messageID); err != nil {
 			r.logger.Errorf("Error acknowledging Redis message %s: %v", msg.messageID, err)
 
 			return err


fsedano:
@cicoyle How is the case of sidecar restarting, after app asked for RETRY handled? If sidecar is ACKing them, aren't they lost once sidecar restarts?

cicoyle:
[Green run in CI](https://github.com/dapr/components-contrib/actions/runs/21610565542/job/62278120804) 🙌🏻 going to re-run everything just to verify no more flakes are showing up

javier-aliaga (dismissed):

pubsub/redis/redis.go
```diff
@@ -245,8 +245,13 @@ func (r *redisStreams) processMessage(msg redisMessageWrapper) error {
 	}
 	if err := msg.handler(ctx, &msg.message); err != nil {
 		r.logger.Errorf("Error processing Redis message %s: %v", msg.messageID, err)
+		// always ack to stop Redis redelivery after the runtime resiliency policy has been applied
+		if err := r.client.XAck(context.Background(), msg.message.Topic, r.clientSettings.ConsumerID, msg.messageID); err != nil {
+			r.logger.Errorf("Error acknowledging Redis message %s: %v", msg.messageID, err)
```
JoshVanL: Is it not the case that we want to not Ack the message is situations like shutting down (context cancelled)- and should we not being using the came context to Ack the message?

```suggestion
    if ctx.Err() != nil {
      return ctx.Err()
    }
		// always ack to stop Redis redelivery after the runtime resiliency policy has been applied
		if err := r.client.XAck(ctx, msg.message.Topic, r.clientSettings.ConsumerID, msg.messageID); err != nil {
			r.logger.Errorf("Error acknowledging Redis message %s: %v", msg.messageID, err)

```

JoshVanL (changes requested):

pubsub/redis/redis.go
cicoyle: I used background here bc the handler already finished at this point and the subscription ctx can be canceled during shutdown. Reusing that ctx when canceled means the ACK gets absorbed and the message is left in the pending entries list causing redelivery on restart. So, if daprd dies before the ACK completes, Redis keeps the entry pending to be redelivered. There is a comment below the code I added stating: `Use the background context in case subscriptionCtx is already closed.`

PR #4213, opened 2026-02-06
feat: Add optional redisPort metadata field for Redis components

### LINKED ISSUE #4212: feat: Support separate redisPort metadata field for Redis components
## Is your feature request related to a problem?

The Redis component metadata (`state.redis`, `pubsub.redis`, `binding.redis`, etc.) requires `redisHost` in combined `host:port` format (e.g., `redis-master.default.svc.cluster.local:6379`). There is no separate `redisPort` metadata field.

Many secret management systems and infrastructure tools store Redis connection details as **separate host and port values** — this is a widely adopted convention across the ecosystem:

- **AWS ElastiCache** exposes separate `Address` and `Port` attributes
- **Infisical / HashiCorp Vault** commonly store `REDIS_HOST` and `REDIS_PORT` as distinct secrets
- **Retool** documents them as separate env vars ([REDIS_HOST](https://docs.retool.com/self-hosted/reference/environment-variables/redis#property-REDIS_HOST), [REDIS_PORT](https://docs.retool.com/self-hosted/reference/environment-variables/redis#property-REDIS_PORT))
- **Kubernetes ConfigMaps** often split host and port for reusability across different consumers

When using `secretKeyRef` to pull values from Kubernetes secrets (a common Dapr pattern), users cannot concatenate two secret values within the component YAML. This forces one of these workarounds:

1. **Pre-combine** host and port into a single secret value — breaking the convention used by the rest of the infrastructure
2. **Init containers or scripts** to concatenate values into a derived secret
3. **Helm templating** to combine values at deploy time — but this doesn't work with external secret operators that sync individual fields

## Example: Current limitation

```yaml
# This is what we'd LIKE to do (but can't):
spec:
  type: state.redis
  version: v1
  metadata:
    - name: redisHost
      secretKeyRef:
        name: my-redis-secret
        key: host        # "redis-master.default.svc.cluster.local"
    - name: redisPort
      secretKeyRef:
        name: my-redis-secret
        key: port        # "6379"
```

```yaml
# What we're FORCED to do — pre-combine into a single "endpoint" secret:
spec:
  type: state.redis
  version: v1
  metadata:
    - name: redisHost
      secretKeyRef:
        name: my-redis-secret
        key: endpoint    # "redis-master.default.svc.cluster.local:6379"
```

## Describe the solution you'd like

Add an optional `redisPort` metadata field to the Redis `Settings` struct in `common/component/redis/settings.go`. When provided, it would be appended to `redisHost` during initialization:

- If `redisHost` already contains `:` (i.e., port is embedded) → ignore `redisPort` (backward compatible)
- If `redisHost` has no `:` and `redisPort` is set → combine as `host:port`
- If `redisHost` has no `:` and `redisPort` is not set → default to port `6379` (Redis default)

This is a **non-breaking, additive change**. All existing configurations continue to work unchanged.

## Describe alternatives you've considered

1. **String interpolation in component YAML** — Not supported by Dapr component spec
2. **Derived secrets via init containers** — Adds operational complexity and another failure point
3. **Helm-only concatenation** — Doesn't work with ExternalSecrets/Infisical operators that sync individual fields
4. **A generic "metadata value template" feature** — Much larger scope; a simple `redisPort` field solves the immediate need

## Additional context

The `Settings` struct in `common/component/redis/settings.go` currently maps `redisHost` via `mapstructure:"redisHost"`. The proposed change would add:

```go
Port string `mapstructure:"redisPort"`
```

And a small resolution step in the `Decode()` or client-creation path to combine host + port when needed.

This pattern of separate host/port fields is already used by other Dapr components (e.g., some SQL components accept separate host and port metadata).

lindner:
## Summary

Adds an optional `redisPort` metadata field to all Redis components (state, pubsub, bindings, lock, configuration). This allows host and port to be provided as separate metadata values, aligning with secret management systems that store them independently (AWS ElastiCache, Vault, Infisical).

Resolves #4212

## Changes

- Added `Port` field to `Settings` struct with `mapstructure:"redisPort"`
- Added `resolveHost`/`resolveAddr` using `net.JoinHostPort` (standard Go idiom used elsewhere in the repo)
- Host resolution runs in `Decode()` so all client types (v8, v9, failover) benefit automatically
- Conflict validation: errors if `redisHost` contains a port that differs from `redisPort`
- Updated all 5 Redis `metadata.yaml` files with the new field documentation
- No `default` set in metadata.yaml to avoid runtime injection breaking existing configs

## Backward compatibility

| `redisHost` | `redisPort` | Result |
|---|---|---|
| `redis:6380` | _(not set)_ | `redis:6380` — unchanged |
| `redis:6379` | `6379` | `redis:6379` — match OK |
| `redis:6379` | `6380` | Error — conflict |
| `redis` | `6380` | `redis:6380` — combined |
| `redis` | _(not set)_ | `redis:6379` — default |

## Test plan

- [x] 15 `TestResolveHost` unit tests (single, cluster, sentinel, IPv6, FQDN, conflict cases)
- [x] 6 `TestSettingsDecodeResolvesHostPort` integration tests through `Decode()`
- [x] All existing tests pass (`common/component/redis`, `state/redis`, `configuration/redis`, `lock/redis`)


lindner:
@acroca or @yaron2  WDYT? this feature would make our deployment much easier,

common/component/redis/redis_test.go
```diff
@@ -50,7 +50,7 @@ const (
 
 func getFakeProperties() map[string]string {
 	return map[string]string{
-		host:                  "fake.redis.com",
+		host:                  "fake.redis.com:6379",
```
acroca, line 53: This change shouldn't be necessary because it's the default port, right?

bindings/redis/metadata.yaml
```diff
@@ -69,9 +69,22 @@ authenticationProfiles:
...
+      provided separately via "redisPort".
     example: "redis-master.default.svc.cluster.local:6379"
     type: string
+  - name: redisPort
+    required: false
+    description: |
+      The Redis port. Optional: if "redisHost" already contains a port, this
+      field must either match or be omitted. When "redisHost" does not include
+      a port and this field is not set, the default Redis port 6379 is used.
+      Useful when host and port are stored as separate secrets.
```
acroca: I would remove this last line 😄
```suggestion
```

common/component/redis/settings_test.go
```diff
@@ -8,6 +8,181 @@ import (
...
+			host: "node1:6379,node2:6379,node3:6379",
+			port: "",
+			want: "node1:6379,node2:6379,node3:6379",
+		},
+		{
+			name: "cluster hosts without ports get separate port",
+			host: "node1,node2,node3",
+			port: "6380",
+			want: "node1:6380,node2:6380,node3:6380",
+		},
```
acroca, line 66: Do you think this behavior should be mentioned in the field description? It might not be obvious.

lindner:
reviewing shortly, thanks for the feedback!

bindings/redis/metadata.yaml
lindner: Fixed here and elsewhere.

common/component/redis/settings_test.go
lindner: Agreed, added more docs.

common/component/redis/redis_test.go
lindner: Correct, also adjusted the test.

lindner:
Addressed concerns, thanks for the review!

acroca:
Thanks for the contribution. I'd like to hear from other maintainers, since this is modifying a very important component I'm a bit cautious about introducing changes in there.

lindner:
> Thanks for the contribution. I'd like to hear from other maintainers, since this is modifying a very important component I'm a bit cautious about introducing changes in there.

Thanks!  I'm available to discuss anything if there are questions.

lindner:
Updating branch.

common/component/redis/redis_test.go
```diff
@@ -87,7 +87,7 @@ func TestParseRedisMetadata(t *testing.T) {
 
 		// assert
 		require.NoError(t, err)
-		assert.Equal(t, fakeProperties[host], m.Host)
+		assert.Equal(t, "fake.redis.com:6379", m.Host)
```
mikeee: Just for consistency could you please update the map value on #L53 rather than defining the host here

common/component/redis/settings_test.go
```diff
@@ -8,6 +8,181 @@ import (
...
+			want: "",
+		},
+		{
+			name: "cluster hosts all with ports",
+			host: "node1:6379,node2:6379,node3:6379",
+			port: "",
+			want: "node1:6379,node2:6379,node3:6379",
+		},
+		{
+			name: "cluster hosts without ports get separate port",
```
mikeee:
```suggestion
		{
			name: "cluster hosts without ports get the specified port applied",
```

Seems clearer with the test case to me

common/component/redis/settings_test.go
```diff
@@ -8,6 +8,181 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestResolveHost(t *testing.T) {
```
mikeee, line 11: Some more test cases and associated logic I would like to see just out of an abundance of clarity would be where there are a list of hosts given with some having ports and some without.

```
{
    name: "mixed cluster hosts: some with matching port, some without",
    host: "node1:6380,node2,node3:6380",
    port: "6380",
    want: "node1:6380,node2:6380,node3:6380",
},
{
    name: "mixed cluster hosts: some with port, some without, using default 6379",
    host: "node1:6379,node2,node3",
    port: "",
    want: "node1:6379,node2:6379,node3:6379",
},
{
    name:    "mixed cluster hosts: one entry conflicts with redisPort",
    host:    "node1:6380,node2,node3:9999",
    port:    "6380",
    wantErr: true,
},
```

common/component/redis/settings_test.go
mikeee:
```suggestion
		{
			name: "cluster hosts without ports get the specified port applied",
```

Seems clearer with the test case to me

lock/redis/metadata.yaml
```diff
@@ -58,8 +58,19 @@ authenticationProfiles:
 metadata:
   - name: redisHost
     required: true
-    description: "The Redis host address"
+    description: |
+      Connection-string for the redis host. The port can be included in the
+      host string (e.g. "host:6379") or provided separately via "redisPort".
```
mikeee:
```suggestion
      Connection-string for the redis host. If "redisType" is "cluster" it
      can be multiple hosts separated by commas or just a single host.
      The port can be included in the host string (e.g. "host:6379") or
      provided separately via "redisPort".
```

common/component/redis/settings.go
```diff
@@ -119,9 +125,68 @@ func (s *Settings) Decode(in interface{}) error {
...
 }
 
+// resolveHost ensures Host contains a port. If Host already includes a port
+// (contains ":"), it is returned unchanged. Otherwise, the separate Port value
+// is appended; when Port is empty the Redis default (6379) is used.
+// For comma-separated host lists (cluster/sentinel), each entry is resolved
+// individually.
+// Returns an error if any address already contains a port that conflicts with
+// a separately configured port value.
+func resolveHost(host, port string) (string, error) {
```
mikeee: The port should be a valid integer within bounds and if not, an error should be returned here.

common/component/redis/settings.go
```diff
@@ -119,9 +125,68 @@ func (s *Settings) Decode(in interface{}) error {
...
+	if host == "" {
+		return host, nil
+	}
+
+	// Comma-separated addresses (cluster or sentinel mode).
+	if strings.Contains(host, ",") {
+		addrs := strings.Split(host, ",")
```
mikeee:
```suggestion
		parts := strings.Split(host, ",")
		addrs:= make([]string, len(parts))
```

Create a sized slice here

mikeee (changes requested):

common/component/redis/redis_test.go
lindner: Added uint16 to the schema so it catches on parsing.

common/component/redis/settings_test.go
lindner: done

common/component/redis/settings.go
lindner: done

mikeee (approved):

marcduiker:
@holopin-bot @lindner components-badge Thank you! Here's a digital badge as a small token of appreciation.

PR #4217, opened 2026-02-11
Fix state redis nil

cicoyle:
repro-ed `redis: nil` here https://github.com/dapr/components-contrib/pull/4217/changes/98e4a22af2b8bcdb3571059cccc6ece3c890f0c6

fixing it and adding tests


pubsub/redis/redis.go
```diff
@@ -280,7 +280,7 @@ func (r *redisStreams) pollNewMessagesLoop(ctx context.Context, stream string, h
 		//nolint:gosec
 		streams, err := r.client.XReadGroupResult(ctx, r.clientSettings.ConsumerID, r.clientSettings.ConsumerID, []string{stream, ">"}, int64(r.clientSettings.QueueDepth), time.Duration(r.clientSettings.ReadTimeout))
 		if err != nil {
-			if !errors.Is(err, r.client.GetNilValueError()) && err != context.Canceled {
+			if err.Error() != r.client.GetNilValueError().Error() && err != context.Canceled {
```
acroca: I think we should keep using error type checks instead of string checks here. Should we create a specific type for Nil errors instead of using `RedisError`?

common/component/redis/redis.go
```diff
@@ -70,7 +70,7 @@ type RedisPipeliner interface {
 
 //nolint:interfacebloat
 type RedisClient interface {
-	GetNilValueError() RedisError
+	IsNilValueError(error) bool
```
acroca, line 73: I assume this interface is only used internally, right?

JoshVanL: Yep :+1:

acroca (approved):

yaron2 (approved):

PR #4227, opened 2026-02-18
fix(pubsub.snssqs): remove aws v1 by implementing a custom partition …

mikeee:
…resolver

# Description

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/tests/certification/bindings/kafka/components/consumer1/kafka.yaml
+++ b/tests/certification/bindings/kafka/components/consumer1/kafka.yaml
@@ -24,3 +24,5 @@ spec:
     value: 50ms
   - name: backOffDuration
     value: 50ms
+  - name: clientConnectionTopicMetadataRefreshInterval
+    value: 15s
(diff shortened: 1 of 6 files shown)


pubsub/aws/snssqs/metadata.go
```diff
@@ -70,6 +69,29 @@ func maskLeft(s string) string {
 	return string(rs)
 }
 
+// getPartitionFromRegion returns the AWS partition for a given region.
+// TODO: @mikeee - remove this partition acquisition.
+func getPartitionFromRegion(region string) string {
```
acroca, line 74: I see the `endpoints` package has been removed in the V1 SDK ([docs](https://docs.aws.amazon.com/sdk-for-go/v2/developer-guide/migrate-gosdk.html#endpoints)), but instead they recommend to use each service Resolver. Would it be possible to move to that approach? as we move to V2?

mikeee: That's right, eventually the component will be refactored to use the default service resolver. However, I don't want to make materially different changes to the component this close to a release, I think it would be best placed in a patch hence the TODO (appreciate it is less than ideal)

PR #4232, opened 2026-02-23
[1.17] fix(nameresolution.aws.cloudmap): quote numeric example in metadata.yaml

nelson-parente:
## Summary

- The `example` field in the component metadata schema must always be a `string`
- `defaultDaprPort` had an unquoted integer `3500` which caused the `bundle-component-metadata` build tool to panic at runtime
- Wrapping the value in quotes (`"3500"`) fixes the unmarshal error

## Root Cause

```
panic: failed to load metadata for component nameresolution/aws/cloudmap:
Failed to parse metadata file for component nameresolution/aws/cloudmap:
failed to unmarshal metadata: json: cannot unmarshal number into Go struct
field Metadata.metadata.example of type string
```

The `generate-component-metadata-for-tag.yml` workflow was failing on tag `20172e4` due to this.

## Change

```yaml
# before
- name: defaultDaprPort
  type: number
  example: 3500       # ← unquoted integer, fails JSON unmarshal into string

# after
- name: defaultDaprPort
  type: number
  example: "3500"     # ← quoted string, matches schema expectation
```

## Test plan

- [ ] Verify `make bundle-component-metadata` passes locally
- [ ] Confirm `generate-component-metadata-for-tag.yml` workflow passes

### DIFF
--- a/configuration/postgres/postgres_test.go
+++ b/configuration/postgres/postgres_test.go
@@ -138,8 +138,6 @@ func TestPostgresConfigurationWithIAM(t *testing.T) {
 
 	ctx := t.Context()
 
-	fmt.Println("TestPostgresConfigurationWithIAM")
-
 	// Testing use of moto to mock AWS services for IAM authentication
 	motoContainer, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
 		ContainerRequest: testcontainers.ContainerRequest{


configuration/postgres/postgres_test.go
```diff
@@ -126,6 +127,15 @@ func TestValidateInput(t *testing.T) {
 }
 
 func TestPostgresConfigurationWithIAM(t *testing.T) {
+	// testcontainers spins up Linux containers (moto, postgres) that rely on the
+	// bridge network driver. On Windows, Docker runs in Windows-container mode and
+	// does not ship the bridge plugin, so container creation fails even when the
+	// Docker daemon is reachable. Skip explicitly rather than letting the test hang
+	// or produce a confusing "plugin not found" error.
+	if runtime.GOOS == "windows" {
```
cicoyle: maybe we can tweak this to be != `linux_arm` bc it seems to only pass there.

nelson-parente:
TestPostgresConfigurationWithIAM uses testcontainers to spin up moto (AWS mock) and PostgreSQL containers, so it requires a Docker daemon to be available.

GitHub Actions macOS runners do not ship with Docker, causing testcontainers to panic with "rootless Docker not found" instead of failing gracefully. Linux runners have Docker pre-installed and are unaffected.

configuration/postgres/postgres_test.go
nelson-parente: yap! that worked, ty!

PR #4241, opened 2026-02-25
fix(configuration): ensure postgresql configuration reads pgNotifyChannel from component spec instead of requiring client code passing it

CasperGN:
# Description

Currently the PostgreSQL configuration requires a client to pass the `pgNotifyChannel` in client side code which bleeds the underlying infrastructure into the client code and renders it impossible to change the component underneath. This PR introduces reading the `pgNotifyChannel` from the component while still allowing clients to pass it through code ensuring backwards compatability.

## Issue reference

Please reference the issue this PR will close: N/A

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
    * [x] Created the dapr/docs PR: N/A - docs actually states this correctly, the implementation just don't conform


configuration/postgres/metadata.go
```diff
@@ -31,6 +31,7 @@ type metadata struct {
 	pgauth.PostgresAuthMetadata `mapstructure:",squash"`
 	Timeout                     time.Duration `mapstructure:"timeout" mapstructurealiases:"timeoutInSeconds"`
 	ConfigTable                 string        `mapstructure:"table"`
+	PgNotifyChannel             string        `mapstructure:"pgNotifyChannel"`
```
sicoyle: can this be unexported? also

```suggestion
	NotifyChannel             string        `mapstructure:"pgNotifyChannel" mapstructurealiases:"notifyChannel"`
```

We don't need the Pg prefix if the struct is already specific to postgres :)

configuration/postgres/metadata.go
```diff
@@ -63,6 +65,15 @@ func (m *metadata) InitWithMetadata(meta map[string]string) error {
 		return fmt.Errorf("invalid table name '%s'. non-alphanumerics or upper cased table names are not supported", m.ConfigTable)
 	}
 
+	if m.PgNotifyChannel != "" {
+		if len(m.PgNotifyChannel) > maxIdentifierLength {
+			return fmt.Errorf("pgNotifyChannel name is too long - '%s'. max allowed length is %d", m.PgNotifyChannel, maxIdentifierLength)
+		}
+		if !allowedTableNameChars.MatchString(m.PgNotifyChannel) {
+			return fmt.Errorf("invalid pgNotifyChannel name '%s'. non-alphanumerics or upper cased names are not supported", m.PgNotifyChannel)
```
sicoyle:
```suggestion
			return fmt.Errorf("pgNotifyChannel name is too long - %q. max allowed length is %d", m.PgNotifyChannel, maxIdentifierLength)
		}
		if !allowedTableNameChars.MatchString(m.PgNotifyChannel) {
			return fmt.Errorf("invalid pgNotifyChannel name %q. non-alphanumerics or upper cased names are not supported", m.PgNotifyChannel)
```

configuration/postgres/metadata.yaml
```diff
@@ -86,6 +86,14 @@ metadata:
     description: The table name for configuration information.
     example:  "configTable"
     type: string
+  - name: pgNotifyChannel
```
sicoyle: pls update this accordingly if you update with the alias in the metadata.go file

configuration/postgres/metadata.yaml
```diff
@@ -86,6 +86,14 @@ metadata:
...
+    required: false
+    description: |
+      The PostgreSQL NOTIFY channel name for configuration change notifications.
+      Must match the channel used in the database trigger function.
```
sicoyle, line 93: is this something we can test and enforce somewhere?

configuration/postgres/metadata.yaml
```diff
@@ -86,6 +86,14 @@ metadata:
...
+      Can also be set per-request in subscribe metadata.
```
sicoyle: pls state this will take precedence if someone sets both.

configuration/postgres/postgres.go
```diff
@@ -252,15 +252,21 @@ func (p *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.S
...
 	}
 	if pgNotifyChannel == "" {
-		return "", fmt.Errorf("unable to subscribe to '%s'. pgNotifyChannel attribute cannot be empty", p.metadata.ConfigTable)
+		return "", fmt.Errorf("unable to subscribe to '%s'. pgNotifyChannel must be set in component metadata or request metadata", p.metadata.ConfigTable)
+	}
+	if len(pgNotifyChannel) > maxIdentifierLength {
+		return "", fmt.Errorf("pgNotifyChannel name is too long - '%s'. max allowed length is %d", pgNotifyChannel, maxIdentifierLength)
+	}
+	if !allowedTableNameChars.MatchString(pgNotifyChannel) {
+		return "", fmt.Errorf("invalid pgNotifyChannel name '%s'. non-alphanumerics or upper cased names are not supported", pgNotifyChannel)
```
sicoyle:
```suggestion
		return "", fmt.Errorf("unable to subscribe to %q. pgNotifyChannel must be set in component metadata or request metadata", p.metadata.ConfigTable)
	}
	if len(pgNotifyChannel) > maxIdentifierLength {
		return "", fmt.Errorf("pgNotifyChannel name is too long - %q. max allowed length is %d", pgNotifyChannel, maxIdentifierLength)
	}
	if !allowedTableNameChars.MatchString(pgNotifyChannel) {
		return "", fmt.Errorf("invalid pgNotifyChannel name %q. non-alphanumerics or upper cased names are not supported", pgNotifyChannel)
```

sicoyle (commented):
thank you! few comments for ya :)

configuration/postgres/metadata.yaml
CasperGN: Not that I'm aware. We would need to read the inputted value and check against the db for a trigger of that name

sicoyle (dismissed):

configuration/postgres/postgres.go
```diff
@@ -252,17 +252,23 @@ func (p *ConfigurationStore) Subscribe(ctx context.Context, req *configuration.S
...
 
-	pgNotifyChannel := ""
+	notifyChannel := p.metadata.NotifyChannel
 	for k, v := range req.Metadata {
-		if strings.ToLower(k) == "pgnotifychannel" { //nolint:gocritic
-			pgNotifyChannel = v
-			break
+		switch strings.ToLower(k) {
+		case "notifychannel", "pgnotifychannel":
+			notifyChannel = v
```
mikeee: Whilst this approach looks very clean, the unfortunate matter is that it is non-deterministic.

Which one should take precedence?

Iterations over Go maps are random

configuration/postgres/postgres_test.go
mikeee, line 1: Please add tests for conflicting metadata where both methods for setting channels are provided

mikeee (changes requested):
https://github.com/dapr/components-contrib/blob/main/tests/certification/configuration/postgres/components/default/configstore.yml

Does the cert test config need to be updated also?

configuration/postgres/postgres.go
CasperGN: This is a very good point. Good catch. I'll add explicit lookup with preference for the setting in the component spec over the legacy

mikeee (approved):
lgtm, thanks for addressing

PR #4252, opened 2026-03-03
feat: warn when multiple redis stream trim strategies are specified

erj826:
# Description

This is a bug/feature that I encountered while working with Dapr's Redis pubsub component. Redis supports **either** MAXLEN or MINID for [XADD](https://redis.io/docs/latest/commands/xadd/) and [XTRIM](https://redis.io/docs/latest/commands/xtrim/)

The `go-redis` client quietly chooses MAXLEN when both are specified: 
- [go-redis v8](https://github.com/redis/go-redis/blob/v8.11.5/commands.go#L1749-L1764)
- [go-redis v9](https://github.com/redis/go-redis/blob/v9.18.0/stream_commands.go#L106-L118)

Dapr let's me declare both in the `pubsub.redis` component

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: pubsub
spec:
  type: pubsub.redis
  version: v1
  metadata:
    ...
    - name: maxLenApprox
      value: "100000"
    - name: streamTTL
      value: "10s"
```

and `streamTTL` is ignored. I think the warning log and/or a docs update would be helpful.

## Issue reference

None -- can open if appropriate.

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
    * [x] Created the dapr/docs PR: https://github.com/dapr/docs/pull/5061

### DIFF
--- a/pubsub/redis/metadata.yaml
+++ b/pubsub/redis/metadata.yaml
@@ -227,7 +227,10 @@ metadata:
     type: string
   - name: maxLenApprox
     required: false
-    description: Maximum number of items inside a stream. The old entries are automatically evicted when the specified length is reached, so that the stream is left at a constant size. Defaults to unlimited.
+    description: |
+      Maximum number of items inside a stream. The old entries are automatically evicted when the specified length is reached,
+      so that the stream is left at a constant size. Defaults to unlimited.
+      Cannot be used together with streamTTL; only one stream trimming strategy can be active at a time.
     example: "10000"
     type: number
   - name: streamTTL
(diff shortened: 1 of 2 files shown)


JoshVanL (approved):

erj826:
@JoshVanL Happy to make copilot's suggested changes if you think they're appropriate, but two thoughts

> Consider updating the warning to clearly state which setting will be honored

Right now go-redis will fall back to `MaxLenApprox`, but it's just a switch statement on their side that could easily get refactored and change the default behavior. (See go-redis links in the PR description)

> To make behavior deterministic and avoid depending on go-redis internals, consider enforcing a single strategy here (e.g., return an init error, or explicitly clear one setting).

I wanted to avoid introducing a potentially breaking change if users are unknowingly passing both arguments, and didn't want to refactor the logic entirely to create a single setting.

Let me know if this makes sense.

JoshVanL:
Hey @erj826! Totally agree with you :) we can resolve those. I've started auto requesting from CoPilot as a sanity check & second pair of "eyes".

JoshVanL (approved):

marcduiker:
@holopin-bot @erj826 components-badge Thank you! Here's a digital badge as a small token of appreciation.

PR #4253, opened 2026-03-04
fix(kafka): add safety timeouts, client health check, and cleanup fixes

nelson-parente:
## Summary

Addresses findings from the Kafka/Sarama RCA remediation plan (Phases 1 & 4):

**Phase 1 — Critical Path Fixes (P0):**
- **Init timeout**: Add configurable `initTimeout` metadata field (default `30s`) — prevents `Init()` from blocking indefinitely when brokers are unreachable
- **Close timeout**: Add configurable `closeTimeout` metadata field (default `15s`) — prevents `Close()` from hanging forever when consumer goroutines are stuck on dead brokers
- **Client health checking**: After 3 consecutive consume errors, force client re-creation instead of retrying with a cached broken client

**Phase 4 — Cleanup (P2):**
- **Double-close guard**: Guard consumer group close in subscriber's `postAction` with `k.closed.Load()` check and nil-out after close, preventing race with `kafka.Close()`
- **AWS shared config mutation**: Copy `sarama.Config` before setting producer-specific fields (`RequiredAcks`, `Retry.Max`, `Return.Successes`) to prevent bleeding into consumer config

### New metadata fields

| Field | Type | Default | Description |
|-------|------|---------|-------------|
| `initTimeout` | `duration` | `30s` | Max time for Init() to connect to brokers |
| `closeTimeout` | `duration` | `15s` | Max time for Close() to wait for goroutines |

## Test plan
- [x] All existing unit tests pass (`go test ./common/component/kafka/...`)
- [x] `go build` passes for `common/component/kafka`, `pubsub/kafka`, `bindings/kafka`
- [ ] Integration tests against a real Kafka cluster (init timeout, close timeout, client reconnection)
- [ ] Verify AWS IAM auth path with config copy fix

### DIFF
--- a/bindings/kafka/metadata.yaml
+++ b/bindings/kafka/metadata.yaml
@@ -451,6 +451,18 @@ metadata:
       - "range"
       - "sticky"
       - "roundrobin"
+  - name: initTimeout
+    type: duration
+    description: |
+      The timeout for establishing the initial connection to Kafka brokers during component initialization.
+    example: '"30s"'
+    default: '"30s"'
+  - name: closeTimeout
+    type: duration
+    description: |
+      The timeout for waiting on background goroutines to finish gracefully during component Close.
+    example: '"15s"'
+    default: '"15s"'
   - name: excludeHeaderMetaRegex
     type: string
     required: false
--- a/common/component/kafka/aws.go
+++ b/common/component/kafka/aws.go
@@ -98,16 +98,18 @@ func (m *mskTokenProvider) Token() (*sarama.AccessToken, error) {
 }
 
 func (c *AwsClients) getSyncProducer() (sarama.SyncProducer, error) {
-	// Add SyncProducer specific properties to copy of base config
-	c.config.Producer.RequiredAcks = sarama.WaitForAll
-	c.config.Producer.Retry.Max = 5
-	c.config.Producer.Return.Successes = true
+	// Copy the base config to avoid mutating the shared config pointer,
+	// which would bleed producer settings into the consumer.
+	producerConfig := *c.config
+	producerConfig.Producer.RequiredAcks = sarama.WaitForAll
+	producerConfig.Producer.Retry.Max = 5
+	producerConfig.Producer.Return.Successes = true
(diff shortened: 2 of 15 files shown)


nelson-parente:
### Code review

Found 2 issues:

1. **Race condition on `k.clients` writes without synchronization.** The `consume()` loop writes `k.clients = nil` (on consecutive errors) and `postAction` writes `k.clients.consumerGroup = nil`, both from goroutines without holding any lock. Meanwhile `latestClients()` reads `k.clients` concurrently from Publish, Init, and other consume goroutines. This introduces new unsynchronized mutations to a shared pointer that will be caught by the race detector and can cause nil-pointer dereferences in practice. Consider protecting `k.clients` access with a mutex, or using an atomic pointer.

https://github.com/dapr/components-contrib/blob/3173076ac368495f6d6efbdf2e5360f4108ab536/common/component/kafka/subscriber.go#L127-L132

https://github.com/dapr/components-contrib/blob/3173076ac368495f6d6efbdf2e5360f4108ab536/common/component/kafka/subscriber.go#L53-L58

2. **Missing zero-value guard for `initTimeout`.** The code comment on `InitTimeout` states "If zero, defaults to 30s", but this is not enforced at usage. If a user sets `initTimeout=0` in metadata, `time.After(0)` fires immediately and `Init()` always returns a timeout error. `Close()` correctly guards against this (`if closeTimeout <= 0 { closeTimeout = 15 * time.Second }`), but `Init()` does not have an equivalent guard.

https://github.com/dapr/components-contrib/blob/3173076ac368495f6d6efbdf2e5360f4108ab536/common/component/kafka/kafka.go#L270-L275

Compare with the guard in `Close()`:

https://github.com/dapr/components-contrib/blob/3173076ac368495f6d6efbdf2e5360f4108ab536/common/component/kafka/kafka.go#L356-L359

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #4266, opened 2026-03-11
fix(pubsub/pulsar): decode Avro binary payload on subscribe path

nelson-parente:
## Summary

PR #4244 added Avro schema validation on the **publish** path but did not update `handleMessage` on the **subscribe** side.

The Pulsar Go client returns raw Avro binary bytes from `msg.Payload()` even when a schema consumer is configured. The Dapr runtime then tries `json.Unmarshal` on those bytes and fails:

```
error deserializing cloud event in pubsub <component> and topic <topic>:
invalid character '\x06' looking for beginning of value
scope=dapr.runtime.processor.subscription
```

`\x06` is the first byte of the Avro binary-encoded payload (string length 3 for `specversion` = `"1.0"` encoded as zigzag varint). Every message on an Avro-enforced topic gets stuck in a permanent retry loop and is never delivered to the application.

## Root Cause

`handleMessage` always passes `msg.Payload()` (raw bytes) directly:

```go
pubsubMsg := pubsub.NewMessage{
    Data: msg.Payload(), // always raw bytes, never decoded
    ...
}
```

The `goavro.Codec` required for decoding is already compiled and cached in `p.metadata.internalTopicSchemas[topic].codec` (added by #4244). It is simply never used on the receive path.

## Fix

When an Avro schema is registered for the incoming topic, decode the binary payload to native Go types using the cached codec, then re-encode as JSON before passing to the handler. This mirrors the publish path:

| Direction | Operation |
|---|---|
| Publish (#4244) | `NativeFromTextual` (JSON → native) → `msg.Value` (Avro binary via SDK) |
| Subscribe (this PR) | `NativeFromBinary` (Avro binary → native) → `TextualFromNative` (→ JSON) |

No new dependencies. The fix is ~15 lines in `handleMessage` only.

## Tests

Adds three unit tests for the subscribe decode path:
- **Round-trip**: Avro binary → `handleMessage` → JSON delivered to handler ✅
- **Decode error**: truncated payload → Nack + error returned, handler not called ✅
- **No schema**: raw payload passed through unchanged (existing behaviour preserved) ✅

Fixes the subscribe-side gap introduced by #4244.

### DIFF
--- a/.github/infrastructure/docker-compose-secrets-manager.yml
+++ b/.github/infrastructure/docker-compose-secrets-manager.yml
@@ -3,7 +3,7 @@ version: "3.8"
 services:
   localstack:
     container_name: "conformance-aws-secrets-manager"
-    image: localstack/localstack
+    image: localstack/localstack:1.4.0
     ports:
       - "127.0.0.1:4566:4566"
     environment:
--- a/pubsub/pulsar/pulsar.go
+++ b/pubsub/pulsar/pulsar.go
@@ -669,8 +669,29 @@ func (p *Pulsar) listenMessage(ctx context.Context, req pubsub.SubscribeRequest,
 }
 
 func (p *Pulsar) handleMessage(ctx context.Context, originTopic string, msg pulsar.ConsumerMessage, handler pubsub.Handler) error {
+	data := msg.Payload()
+
+	// If an Avro schema is registered for this topic, decode the Avro binary
+	// payload to JSON before passing it to the Dapr runtime. The Pulsar Go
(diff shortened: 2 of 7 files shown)


MyMirelHub (changes requested):
Can we also add a certification test with pulsar avro - see one for equivalent json schema here https://github.com/dapr/components-contrib/blob/main/tests/certification/pubsub/pulsar/components/auth-none/consumer_four/pulsar.yaml

MyMirelHub (approved):

cicoyle (approved):

PR #4275, opened 2026-03-12
feat(config): add config map as configuration

CasperGN:
# Description

Add `ConfigMap` as available configuration store with both `get` and `subscribe` options. Especially in Dapr Agents this configuration would be tremendously valuable to hot-reload various configurations that inevitably would be operator-led.

## Issue reference

Please reference the issue this PR will close: N/A

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
    * [X] Created the dapr/docs PR: https://github.com/dapr/docs/pull/5070


configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,381 @@
...
+
+	return nil
+}
+
+func (s *ConfigurationStore) resolveNamespace(requestMetadata map[string]string) string {
+	if ns, ok := requestMetadata["namespace"]; ok && ns != "" {
+		if errs := validation.IsDNS1123Label(ns); len(errs) == 0 {
+			return ns
+		}
+		s.logger.Warnf("ignoring invalid namespace override %q in request metadata", ns)
```
JoshVanL: Let's hard error when this is wrong

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,381 @@
...
+	s.informerWg.Wait()
+
+	return nil
+}
+
+func (s *ConfigurationStore) resolveNamespace(requestMetadata map[string]string) string {
+	if ns, ok := requestMetadata["namespace"]; ok && ns != "" {
+		if errs := validation.IsDNS1123Label(ns); len(errs) == 0 {
+			return ns
+		}
```
JoshVanL: This doesn't seem necessary to do client side.

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,381 @@
...
+		s.logger.Warnf("ignoring invalid namespace override %q in request metadata", ns)
+	}
+	// Use the NAMESPACE env var only when the component metadata did not
+	// explicitly set a namespace value.
+	if !s.metadata.namespaceExplicit {
+		if ns := os.Getenv("NAMESPACE"); ns != "" {
+			return ns
+		}
```
JoshVanL:
```suggestion
		if ns, ok := os.LookupEnv("NAMESPACE"); ok {
			return ns
		}
```

configuration/kubernetes/metadata.go
```diff
@@ -0,0 +1,62 @@
...
+
+type metadata struct {
+	Namespace      string        `mapstructure:"namespace"`
+	ConfigMapName  string        `mapstructure:"configMapName"`
+	KubeconfigPath string        `mapstructure:"kubeconfigPath"`
+	ResyncPeriod   time.Duration `mapstructure:"resyncPeriod"`
+
+	// namespaceExplicit tracks whether the namespace was explicitly set in
+	// component metadata, to distinguish from the "default" fallback.
+	namespaceExplicit bool
```
JoshVanL: Use pointers for presence and doing defaults.

```suggestion
	Namespace      *string        `mapstructure:"namespace"`
	ConfigMapName  string        `mapstructure:"configMapName"`
	KubeconfigPath *string        `mapstructure:"kubeconfigPath"`
	ResyncPeriod   *time.Duration `mapstructure:"resyncPeriod"`
```

configuration/kubernetes/metadata.go
```diff
@@ -0,0 +1,62 @@
...
+
+	if errs := validation.IsDNS1123Subdomain(m.ConfigMapName); len(errs) > 0 {
+		return fmt.Errorf("configMapName %q is not a valid Kubernetes resource name: %s", m.ConfigMapName, strings.Join(errs, "; "))
+	}
+
+	m.namespaceExplicit = m.Namespace != ""
+	if m.Namespace == "" {
+		m.Namespace = "default"
+	}
+
```
JoshVanL: != nil

configuration/kubernetes/metadata.yaml
```diff
@@ -0,0 +1,45 @@
...
+    example: "/path/to/kubeconfig"
+    type: string
+  - name: resyncPeriod
+    required: false
+    description: |
+      How often the informer fully re-syncs the ConfigMap state from the API
+      server as a consistency safety net, independent of watch events.
+      Set to "0" (default) to disable periodic resync and rely solely on
+      watch events. Set a duration (e.g. "10m") if you want a periodic
+      full re-read as a guard against missed events.
```
JoshVanL, line 35: The kubernetes go client should be doing this for you. So this parameter should be passed through to the kube client

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,381 @@
...
+
+var _ configuration.Store = (*ConfigurationStore)(nil)
+
+// subscriber holds the state for a single subscription registered via Subscribe.
+type subscriber struct {
+	req     *configuration.SubscribeRequest
+	handler configuration.UpdateHandler
+	ctx     context.Context
+	cancel  context.CancelFunc
+}
```
JoshVanL: Please can we move the subscriber struct and it's func to a diff file

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,381 @@
...
+			return fmt.Errorf("failed to create Kubernetes client: %w", err)
+		}
+		s.kubeClient = client
+	}
+
+	ns := s.resolveNamespace(nil)
+	_, err := s.kubeClient.CoreV1().ConfigMaps(ns).Get(ctx, s.metadata.ConfigMapName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get ConfigMap %q in namespace %q: %w", s.metadata.ConfigMapName, ns, err)
+	}
```
JoshVanL: Doesn't seem necessary 

```suggestion
```

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,381 @@
...
+	// Start the shared informer on first subscription.
+	s.startInformer()
+
+	return subscribeID, nil
+}
+
+// startInformer lazily starts a single informer that watches the configured
+// ConfigMap and fans out change events to all registered subscribers.
+func (s *ConfigurationStore) startInformer() {
+	s.informerOnce.Do(func() {
```
JoshVanL: Why do we have a once here?

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,381 @@
...
+		ctx, cancel := context.WithCancel(context.Background())
```
JoshVanL: Why background? We have a subscription context

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,381 @@
...
+			"configmaps",
+			ns,
+			func(options *metav1.ListOptions) {
+				options.FieldSelector = fields.OneTermEqualSelector("metadata.name", s.metadata.ConfigMapName).String()
+			},
+		)
+
+		// Only UpdateFunc is registered: subscribers receive change-only notifications.
+		// AddFunc is intentionally omitted — the initial List populates the informer
+		// cache but does not notify, matching the behavior of the Redis and PostgreSQL
```
JoshVanL: I think we _should_ want the current state on first subscribe.. It's shouldn't be the case that a client needs to list then subscribe on every configuration API use.

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,381 @@
...
+				UpdateFunc: func(oldObj, newObj any) {
+					s.fanOutUpdate(oldObj, newObj)
+				},
+			},
+		})
+
+		s.informerWg.Add(1)
+		go func() {
+			defer s.informerWg.Done()
+			controller.Run(ctx.Done())
```
JoshVanL: Handling error?

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,381 @@
...
+// fanOutUpdate dispatches ConfigMap change events to all registered subscribers.
+func (s *ConfigurationStore) fanOutUpdate(oldObj, newObj any) {
+	oldCM, ok1 := oldObj.(*corev1.ConfigMap)
+	newCM, ok2 := newObj.(*corev1.ConfigMap)
+	if !ok1 || !ok2 {
+		s.logger.Warn("received non-ConfigMap object in update handler")
+		return
+	}
+
+	s.subscribers.Range(func(key, value any) bool {
```
JoshVanL: Please can we use a concurrency.RunnerManager to fan out the update concurrently

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,381 @@
...
+// computeChangedItems computes the set of configuration items that changed
+// between oldCM and newCM, filtered to only the given subscribedKeys (or all
+// keys if subscribedKeys is empty).
+func computeChangedItems(subscribedKeys []string, oldCM, newCM *corev1.ConfigMap) map[string]*configuration.Item {
+	changedItems := make(map[string]*configuration.Item)
+
+	// Detect added or modified keys in data.
+	for k, newVal := range newCM.Data {
+		oldVal, existed := oldCM.Data[k]
+		if !existed || oldVal != newVal {
```
JoshVanL, line 388: This can be really big! We are doing this for every subscriber, when we should only need to do it once on each inform.

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,381 @@
...
+
+// fanOutUpdate dispatches ConfigMap change events to all registered subscribers.
+func (s *ConfigurationStore) fanOutUpdate(oldObj, newObj any) {
+	oldCM, ok1 := oldObj.(*corev1.ConfigMap)
+	newCM, ok2 := newObj.(*corev1.ConfigMap)
+	if !ok1 || !ok2 {
+		s.logger.Warn("received non-ConfigMap object in update handler")
+		return
+	}
+
```
JoshVanL, line 291:
```suggestion
		// Skip no-op updates.
		if oldCM.ResourceVersion == newCM.ResourceVersion {
			return
		}

		// Only react when .data changes.
		if apiequality.Semantic.DeepEqual(oldCM.Data, newCM.Data) {
			return
		}
```

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,381 @@
...
+// between oldCM and newCM, filtered to only the given subscribedKeys (or all
+// keys if subscribedKeys is empty).
+func computeChangedItems(subscribedKeys []string, oldCM, newCM *corev1.ConfigMap) map[string]*configuration.Item {
+	changedItems := make(map[string]*configuration.Item)
+
+	// Detect added or modified keys in data.
+	for k, newVal := range newCM.Data {
+		oldVal, existed := oldCM.Data[k]
+		if !existed || oldVal != newVal {
+			if isSubscribedKey(subscribedKeys, k) {
```
JoshVanL: This is giving me vibes that the data structure isn't quite right, needing to do string matching. I think we should change the data structure to be ConfigMap informer -> map[key]handler where a single handler may be referenced by multiple key map indexes.

JoshVanL (changes requested):

configuration/kubernetes/configmap.go
CasperGN: Removed to let kube api refuse

configuration/kubernetes/configmap.go
CasperGN: Opted for waitgroup as per chat

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,391 @@
...
+			UpdateFunc: func(oldObj, newObj any) {
+				s.fanOutUpdate(oldObj, newObj)
+			},
+		},
+	})
+
+	s.informerWg.Add(1)
+	go func() {
+		defer s.informerWg.Done()
+		controller.Run(ctx.Done())
```
JoshVanL: Missing error return here?

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,391 @@
...
+			defer wg.Done()
+			if err := sub.handler(sub.ctx, &configuration.UpdateEvent{
+				ID:    subID,
+				Items: items,
+			}); err != nil {
+				s.logger.Errorf("subscription %s handler error: %v", subID, err)
+			}
+		}()
+	}
+	wg.Wait()
```
JoshVanL, line 380: This is going to cause a deadlock with `Unsubscribe`. You want to lock, collect all the subscribers, then unlock, then send to all the subscribers which you collected.

configuration/kubernetes/metadata.go
```diff
@@ -0,0 +1,55 @@
...
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/validation"
+
+	"github.com/dapr/components-contrib/configuration"
+	kitmd "github.com/dapr/kit/metadata"
+)
+
+type metadata struct {
+	Namespace      *string        `mapstructure:"namespace"`
```
JoshVanL: We do not (and should not) support cross namespace configuration fetching. The Namespace should be derived form the namespace the daprd is in. Namespaces are a strong security boundary we want to keep for Dapr.

```suggestion
```

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,391 @@
...
+	)
+
+	_, controller := cache.NewInformerWithOptions(cache.InformerOptions{
+		ListerWatcher: watchlist,
+		ObjectType:    &corev1.ConfigMap{},
+		ResyncPeriod:  resyncPeriod,
+		Handler: cache.ResourceEventHandlerFuncs{
+			UpdateFunc: func(oldObj, newObj any) {
+				s.fanOutUpdate(oldObj, newObj)
+			},
```
JoshVanL, line 216: Missing AddFunc- are we covering in tests whereby we watch for a configmap which is not available yet, then it created, and we receive updates?
The declarative/controller like nature of kube means we should support this scenario ^

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,391 @@
...
+		ctx:     childCtx,
+		cancel:  cancel,
+	}
+	s.registry.add(sub)
+
+	// Deliver current state asynchronously. The Dapr sidecar's gRPC handler
+	// blocks until Subscribe returns (to send the subscription ID first), so
+	// calling the handler synchronously here would deadlock.
+	go func() {
+		resp, err := s.Get(ctx, &configuration.GetRequest{
```
JoshVanL:
```suggestion
		resp, err := s.Get(childCtx, &configuration.GetRequest{
```

JoshVanL (changes requested):

sicoyle:
amazing! 🙌 pls be sure to queue up the corresponding dapr/dapr PR for this. You should have some registration file similar to the one here https://github.com/dapr/dapr/blob/master/cmd/daprd/components/configuration_redis.go along with the contrib go mod bump :)

tests/certification/configuration/kubernetes/components/default/configstore.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: configstore
+spec:
+  type: configuration.kubernetes
+  version: v1
+  metadata:
+  - name: configMapName
+    value: "dapr-cert-test"
+  - name: namespace
+    value: "default"
```
JoshVanL:
```suggestion
```

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,470 @@
...
+		}
+
+		client, err := kubeclient.GetKubeClient(kubeconfigPath)
+		if err != nil {
+			return fmt.Errorf("failed to create Kubernetes client: %w", err)
+		}
+		s.kubeClient = client
+	}
+
+	if err := s.startInformer(); err != nil {
```
JoshVanL:
```suggestion
	if err := s.startInformer(ctx); err != nil {
```

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,470 @@
...
+	})
+
+	s.informerWg.Add(1)
+	go func() {
+		defer s.informerWg.Done()
+		controller.Run(ctx.Done())
+		s.logger.Info("ConfigMap informer stopped")
+	}()
+
+	if !cache.WaitForCacheSync(ctx.Done(), controller.HasSynced) {
```
JoshVanL, line 233: This should be using the context from `Init`, rather than the background context. This may hang forever otherwise.

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,470 @@
...
+
+	return nil
+}
+
+func (s *ConfigurationStore) Get(ctx context.Context, req *configuration.GetRequest) (*configuration.GetResponse, error) {
+	cm, err := s.kubeClient.CoreV1().ConfigMaps(s.namespace).Get(ctx, s.metadata.ConfigMapName, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get ConfigMap %q: %w", s.metadata.ConfigMapName, err)
+	}
+
```
JoshVanL: This should be using the informer to get, not an out of band full client get.

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,470 @@
...
+		ListerWatcher: watchlist,
+		ObjectType:    &corev1.ConfigMap{},
+		ResyncPeriod:  resyncPeriod,
+		Handler: cache.ResourceEventHandlerFuncs{
+			AddFunc: func(obj any) {
+				s.fanOutAdd(obj)
+			},
+			UpdateFunc: func(oldObj, newObj any) {
+				s.fanOutUpdate(oldObj, newObj)
+			},
```
JoshVanL, line 216: I guess we should also be covering Deletes as well.

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,470 @@
...
+	s.registry.add(sub)
+
+	// Deliver current state asynchronously. The Dapr sidecar's gRPC handler
+	// blocks until Subscribe returns (to send the subscription ID first), so
+	// calling the handler synchronously here would deadlock.
+	go func() {
+		resp, err := s.Get(childCtx, &configuration.GetRequest{
+			Keys:     req.Keys,
+			Metadata: req.Metadata,
+		})
```
JoshVanL: We shouldn't swallow this, this is a real error that should error out the `Subscribe`.

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,470 @@
...
+	}
+
+	// Skip no-op updates.
+	if oldCM.ResourceVersion == newCM.ResourceVersion {
+		return
+	}
+	if maps.Equal(oldCM.Data, newCM.Data) &&
+		maps.EqualFunc(oldCM.BinaryData, newCM.BinaryData, bytes.Equal) {
+		return
+	}
```
JoshVanL: Redundant

```suggestion
```

JoshVanL (changes requested):

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,496 @@
...
+}
+
+// fanOutDelete notifies subscribers that all keys have been deleted when the
+// ConfigMap is removed from the cluster.
+func (s *ConfigurationStore) fanOutDelete(obj any) {
+	cm, ok := obj.(*corev1.ConfigMap)
+	if !ok {
+		s.logger.Warn("received non-ConfigMap object in delete handler")
+		return
+	}
```
JoshVanL, line 265: Need to handle `cache.DeletedFinalStateUnknown`.

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,496 @@
...
+
+	s.registry.add(sub)
+
+	// Deliver initial state asynchronously. The Dapr sidecar's gRPC handler
+	// blocks until Subscribe returns (to send the subscription ID first), so
+	// calling the handler synchronously here would deadlock.
+	if len(resp.Items) > 0 {
+		go func() {
+			if hErr := handler(childCtx, &configuration.UpdateEvent{
+				ID:    subscribeID,
```
JoshVanL, line 169: Adding to the registry and sending the initial state all need to happen inside the same lock/critical section, else an informer event can be sent _before_ the initial state causing the consumer to have the wrong view of the world. Have we considered using a `SharedIndexInformer` where then each Subscribe adds it's own handler? This prevents needing the registry and all this co-ordination.

JoshVanL (changes requested):

configuration/kubernetes/configmap.go
```diff
@@ -0,0 +1,436 @@
...
+func (s *ConfigurationStore) GetComponentMetadata() (metadataInfo contribMetadata.MetadataMap) {
+	metadataStruct := metadata{}
+	contribMetadata.GetMetadataInfoFromStructType(reflect.TypeOf(metadataStruct), &metadataInfo, contribMetadata.ConfigurationStoreType)
+	return metadataInfo
+}
+
+func (s *ConfigurationStore) Close() error {
+	s.closed.Store(true)
+
+	s.mu.Lock()
```
JoshVanL: Prevent race

```suggestion
	s.mu.Lock()
	s.closed.Store(true)
```

JoshVanL (changes requested):
Think it looks good to me, just one more nit

CasperGN:
> Think it looks good to me, just one more nit

Thanks @JoshVanL - I've merged the suggestion! 🚀

JoshVanL (approved):

sicoyle (approved):

marcduiker:
@holopin-bot @CasperGN components-badge Thank you! Here's a digital badge as a small token of appreciation.

PR #4277, opened 2026-03-12
return err if tool calls were not called when required

cicoyle:
ran this locally to confirm it at least will err instead of silently failing while completing the agent/wf successfully even tho no tool calls were made when required.


conversation/langchaingokit/model.go
```diff
@@ -39,6 +39,7 @@ func New(logger logger.Logger) LLM {
 }
 
 func (a *LLM) Converse(ctx context.Context, r *conversation.Request) (res *conversation.Response, err error) {
+	a.logger.Debugf("Calling LLM model: %s", a.model)
```
JoshVanL: Is this log really needed? Can end up being quite noisy when the model doesn't change (?)

conversation/langchaingokit/model.go
```diff
@@ -56,6 +57,27 @@ func (a *LLM) Converse(ctx context.Context, r *conversation.Request) (res *conve
...
+					hasUsefulResponse = true
+					break
+				}
+			}
+			if hasUsefulResponse {
+				break
+			}
+		}
+		if !hasUsefulResponse {
+			return nil, fmt.Errorf("LLM returned empty response with no tool calls despite %d tools being available", len(*r.Tools))
```
JoshVanL, line 76: Is it true a tool call must always be called?

JoshVanL (changes requested):

conversation/langchaingokit/model.go
cicoyle: Its not needed, but when I was switching models and testing things it was a nice to have.

conversation/langchaingokit/model.go
cicoyle: I had a case where I was passing in `tool_choice: required` as `AgentExecutionConfig` to ollama and ollama was ignoring this and returning an empty resp with no tool calls, no err, and it was just silently succeeding with nothing useful which is a silent failure. This check verifies when this occurs, uses dapr's resiliency policy retries.

This is needed imo bc models, like ollama, can silently return an empty resp with no tool calls, no errs and it looks like it succeeded but really it did not. This detects that and retries or surfaces an err to the user.

JoshVanL: I see- should it not be the case that we only error here if `tool_choice: required` is also there?

cicoyle: fair. let me add that rq

JoshVanL (approved):

JoshVanL:
@cicoyle see failures

JoshVanL (changes requested):
Test failures

JoshVanL (approved):

PR #4280, opened 2026-03-13
feat(bindings): add fromTemp metadata to localstorage

### LINKED ISSUE #3158: Cross-platform temp directory for local storage binding
The local storage binding is very useful in development scenarios where a cloud storage solution would be cumbersome.
In these development scenarios, you may want to use the system temp directory. 

However, there is currently no way to do this is in a cross-platform way.
The local storage component currently has a single "rootPath" metadata.  
To use the temp directory on Unix-based systems, rootPath can be set to "/tmp".
On Windows, either the system-wide temp directory (C:\Windows\Temp) or the user-specific temp directory (C:\Users\<User>\AppData\Local\Temp) can be used.

This may be a problem when sharing development environments between systems, especially in multi-app run scenarios.

To solve this, I'd like to suggest either adding a "fromTemp" boolean metadata that resolves rootPath from the system-wide tmp dir if true, and defaults to false (so as not to break any current deployments), or allowing the expansion of env variables in rootpath before creating the directory (but this may lead to unexpected behavior)

stywzn:
## Description

This PR adds the `fromTemp` boolean metadata to the `localstorage` binding component to allow resolving the `rootPath` from the system-wide temporary directory using `os.TempDir()`. This ensures true cross-platform compatibility.

This addresses the need for using system temp directories in development scenarios without hardcoding OS-specific paths (like `/tmp` or `C:\Windows\Temp`). Downward compatibility is strictly maintained (defaults to `false`). Unit tests and `metadata.yaml` have been updated accordingly.

Fixes #3158

### DIFF
--- a/bindings/localstorage/localstorage.go
+++ b/bindings/localstorage/localstorage.go
@@ -57,6 +57,7 @@ type LocalStorage struct {
 // Metadata defines the metadata.
 type Metadata struct {
 	RootPath string `json:"rootPath"`
+	FromTemp bool   `json:"fromTemp"`
 }
 
 type createResponse struct {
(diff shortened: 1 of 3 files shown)


stywzn:
@dapr-bot remove-stale

acroca:
@stywzn It'd be good if you improve the certification tests to use this new setting too

PR #4287, opened 2026-03-16
ollama: add missing endpoint metadata field to spec

lrascao:
# Description

Add missing `endpoint` field to ollama spec.

### DIFF
--- a/conversation/ollama/metadata.yaml
+++ b/conversation/ollama/metadata.yaml
@@ -27,4 +27,5 @@ metadata:
     description: |
       The URL of the Ollama server.
     type: string
-    example: 'http://127.0.0.1:11434/v1'
+    example: 'http://10.20.23.21:11434/v1'
+    default: 'http://localhost:11434/v1'


conversation/ollama/metadata.yaml
```diff
@@ -22,3 +22,9 @@ metadata:
       A time-to-live value for a prompt cache to expire. Uses Golang durations
     type: string
     example: '10m'
+  - name: endpoint
+    required: false
```
JoshVanL, line 26: Is this not required?

lrascao: defaults to `"http://localhost:11434/v1"`

conversation/ollama/metadata.yaml
```diff
@@ -22,3 +22,9 @@ metadata:
...
+    description: |
+      The URL of the Ollama server.
+    type: string
+    example: 'http://127.0.0.1:11434/v1'
```
JoshVanL: Please change this to the `default` and have a different example.

PR #4292, opened 2026-03-18
Bulk file transfer operations to AWS S3 binding

### LINKED ISSUE #3873: AWS Storage Bucket binding: Bulk file transfer
We move files from AWS object store to PV for processing and there are more then 100+ files (ranging from 15 to 30GB) to be transferred daily.

Dapr supports one by one file transfer and hence not been able to use dapr object store binding for this. One by one file transfer takes long time to complete.

Using AWS SDK (java) completes the 100+ file transfers in around 1 minutes or less.

We would like to go with dapr binding for this requirement as well.

Can we have feature similar to how AWS sdk provides for transfer of files from a object store to PV. Provision to transfer files in parallel from one object store to PV makes it very fast.

JoshVanL:
Add bulkGet, bulkCreate, and bulkDelete operations to enable parallel file transfers, addressing the performance gap when transferring 100+ large files.

- bulkGet: parallel downloads with optional filePath for streaming directly to disk (avoids loading large files into memory)
- bulkCreate: parallel uploads from inline data or filePath, using the existing transfer manager for multipart uploads
- bulkDelete: batch deletes using the S3 DeleteObjects API with automatic chunking at the 1000-key limit
- Bounded concurrency via configurable pointer field (default 10)
- Streaming base64 encode/decode throughout
- Per-item error reporting for partial success semantics

Fixes #3873

### DIFF
--- a/bindings/aws/s3/s3.go
+++ b/bindings/aws/s3/s3.go
@@ -538,7 +538,7 @@ func (s *AWSS3) bulkGet(ctx context.Context, req *bindings.InvokeRequest) (*bind
 			defer func() { <-sem }()
 
 			resp, gerr := s.s3Client.GetObject(ctx, &s3.GetObjectInput{
-				Bucket: ptr.Of(s.metadata.Bucket),
+				Bucket: ptr.Of(metadata.Bucket),
 				Key:    ptr.Of(it.Key),
 			})
 			if gerr != nil {
@@ -706,7 +706,7 @@ func (s *AWSS3) bulkCreate(ctx context.Context, req *bindings.InvokeRequest) (*b
 }
 
 func (s *AWSS3) bulkDelete(ctx context.Context, req *bindings.InvokeRequest) (*bindings.InvokeResponse, error) {
-	_, err := s.metadata.mergeWithRequestMetadata(req)
+	metadata, err := s.metadata.mergeWithRequestMetadata(req)
 	if err != nil {
 		return nil, fmt.Errorf("s3 binding error: error merging metadata: %w", err)
 	}
@@ -751,7 +751,7 @@ func (s *AWSS3) bulkDelete(ctx context.Context, req *bindings.InvokeRequest) (*b
 
 		var output *s3.DeleteObjectsOutput
 		output, err = s.s3Client.DeleteObjects(ctx, &s3.DeleteObjectsInput{
-			Bucket: ptr.Of(s.metadata.Bucket),
+			Bucket: ptr.Of(metadata.Bucket),
 			Delete: &s3types.Delete{
 				Objects: objects,
 				Quiet:   ptr.Of(false),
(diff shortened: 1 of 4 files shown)


JoshVanL:
AWS S3 certification test: https://github.com/dapr/components-contrib/actions/runs/23248555121/job/67583764860

bindings/aws/s3/s3.go
```diff
@@ -436,6 +469,310 @@ func (s *AWSS3) list(ctx context.Context, req *bindings.InvokeRequest) (*binding
...
+	jsonResponse, err := json.Marshal(results)
+	if err != nil {
+		return nil, fmt.Errorf("s3 binding error: error marshalling bulkCreate response: %w", err)
+	}
+
+	return &bindings.InvokeResponse{Data: jsonResponse}, nil
+}
+
+func (s *AWSS3) bulkDelete(ctx context.Context, req *bindings.InvokeRequest) (*bindings.InvokeResponse, error) {
+	var payload bulkDeletePayload
```
acroca, line 714: Don't we need to call the `metadata, err := s.metadata.mergeWithRequestMetadata(req)` as the other functions do?

bindings/aws/s3/s3.go
```diff
@@ -436,6 +469,310 @@ func (s *AWSS3) list(ctx context.Context, req *bindings.InvokeRequest) (*binding
...
+	var wg sync.WaitGroup
+
+	for i, item := range payload.Items {
+		if item.Key == "" {
+			results[i] = bulkItemResult{Key: "", Error: "key is required"}
+			continue
+		}
+
+		wg.Add(1)
+		sem <- struct{}{}
```
acroca: I think it'd be good for this to check `ctx.Done` as well, in case the context is cancelled half way

bindings/aws/s3/s3.go
JoshVanL: Done :+1:

bindings/aws/s3/s3.go
JoshVanL: Done :+1:

bindings/aws/s3/s3.go
```diff
@@ -436,6 +469,328 @@ func (s *AWSS3) list(ctx context.Context, req *bindings.InvokeRequest) (*binding
...
+			results[i] = bulkItemResult{Key: item.Key, Error: ctx.Err().Error()}
+			wg.Done()
+			continue
+		}
+		go func(idx int, it bulkGetItem) {
+			defer wg.Done()
+			defer func() { <-sem }()
+
+			resp, gerr := s.s3Client.GetObject(ctx, &s3.GetObjectInput{
+				Bucket: ptr.Of(s.metadata.Bucket),
```
cicoyle: Should this be `s.metadata.Bucket` or `metadata` that is set above?
```suggestion
				Bucket: ptr.Of(metadata),
```

cicoyle (changes requested):

bindings/aws/s3/s3.go
JoshVanL: Good catch- fixed to use the merged metadata.Bucket

cicoyle (approved):

PR #4299, opened 2026-03-19
bindings/azure/blobstorage: Adds presign

### LINKED ISSUE #3817: Trouble Generating SAS Token for Blob Storage
## area

> /area runtime

## Question

Hey folks, I ran into a bit of a roadblock while building my app with Dapr and wanted to share what’s going on — maybe you’ve seen something similar or have some guidance?

---

### Trouble Generating SAS Token for Blob Storage

**What’s the deal?:**

I am using Azure Blob Storage, which—as you probably know—needs a SAS token to access files securely. The bindings.azure.blobstorage binding in Dapr currently supports basic operations like `get`, `create`, `delete`, and `list`, but doesn’t seem to directly support generating SAS tokens.

**What I have tried:**

I gave it a shot by tweaking the metadata in our requests, hoping I could trigger SAS token generation that way. Here’s what I did:

- **1 with `get`** – here I receive the file data:
```json
{
  "operation": "get",
  "metadata": {
    "sasToken": "true",
    "blobName": "<filename>"
  }
}
```

- **2 with `create`** – here I receive the presigned URL:
```json
{
  "operation": "create",
  "metadata": {
    "sasToken": "true",
    "blobName": "<filename>"
  }
}
```

**Where I am at::**

It looks like Dapr doesn’t currently support SAS token generation directly, at least not in a straightforward way. If there’s a trick I am missing, I’d love to hear it. Otherwise, maybe this could be something to consider as a future enhancement?

Let me know what you think — any tips or workarounds are more than welcome!

---

JoshVanL:
Add SAS URL generation support to the Azure Blob Storage binding, enabling users to generate temporary read-only access URLs for blobs without downloading the blob content.

Add new "presign" operation requiring "blobName" and "signTTL" metadata. Optionally return a presigned SAS URL during "create" when "signTTL" is provided in request metadata.

Fixes #3817

### DIFF
--- a/bindings/azure/blobstorage/blobstorage.go
+++ b/bindings/azure/blobstorage/blobstorage.go
@@ -225,6 +225,17 @@ func (a *AzureBlobStorage) create(ctx context.Context, req *bindings.InvokeReque
 		blobName = id.String()
 	}
 
+	// Extract and validate signTTL before upload so we fail fast and don't
+	// persist it as blob metadata.
+	var signTTL string
+	if ttl, ok := req.Metadata[metadataKeySignTTL]; ok && ttl != "" {
+		if _, err := time.ParseDuration(ttl); err != nil {
+			return nil, fmt.Errorf("cannot parse signTTL duration %q: %w", ttl, err)
+		}
+		signTTL = ttl
+		delete(req.Metadata, metadataKeySignTTL)
+	}
+
 	blobHTTPHeaders, err := storagecommon.CreateBlobHTTPHeadersFromRequest(req.Metadata, nil, a.logger)
 	if err != nil {
 		return nil, err
@@ -256,11 +267,12 @@ func (a *AzureBlobStorage) create(ctx context.Context, req *bindings.InvokeReque
 	}
 
 	resp := createResponse{
-		BlobURL: blockBlobClient.URL(),
+		BlobURL:  blockBlobClient.URL(),
+		BlobName: blobName,
 	}
 
-	if ttl, ok := req.Metadata[metadataKeySignTTL]; ok && ttl != "" {
-		presignURL, presignErr := a.generateSASURL(blockBlobClient, ttl)
+	if signTTL != "" {
+		presignURL, presignErr := a.generateSASURL(blockBlobClient, signTTL)
 		if presignErr != nil {
 			return nil, fmt.Errorf("error generating SAS URL: %w", presignErr)
 		}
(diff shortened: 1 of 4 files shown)


bindings/azure/blobstorage/blobstorage.go
```diff
@@ -159,8 +185,18 @@ func (a *AzureBlobStorage) create(ctx context.Context, req *bindings.InvokeReque
 	}
 
 	resp := createResponse{
-		BlobURL: blockBlobClient.URL(),
+		BlobURL:  blockBlobClient.URL(),
+		BlobName: blobName,
+	}
+
+	if signTTL != "" {
+		presignURL, presignErr := a.generateSASURL(blockBlobClient, signTTL)
```
acroca: Shouldn't do this before the `UploadBuffer`?

bindings/azure/blobstorage/blobstorage.go
```diff
@@ -128,6 +143,17 @@ func (a *AzureBlobStorage) create(ctx context.Context, req *bindings.InvokeReque
...
 
+	// Extract and validate signTTL before upload so we fail fast and don't
+	// persist it as blob metadata.
+	var signTTL string
+	if ttl, ok := req.Metadata[metadataKeySignTTL]; ok && ttl != "" {
+		if _, err := time.ParseDuration(ttl); err != nil {
+			return nil, fmt.Errorf("cannot parse signTTL duration %q: %w", ttl, err)
+		}
+		signTTL = ttl
+		delete(req.Metadata, metadataKeySignTTL)
```
acroca: Is it ok to mutate the request object?

bindings/azure/blobstorage/blobstorage.go
```diff
@@ -344,6 +380,55 @@ func (a *AzureBlobStorage) list(ctx context.Context, req *bindings.InvokeRequest
 	}, nil
 }
 
+func (a *AzureBlobStorage) presign(ctx context.Context, req *bindings.InvokeRequest) (*bindings.InvokeResponse, error) {
```
acroca: Since this is non exposed and ctx is not used, I think we should drop it. 
```suggestion
func (a *AzureBlobStorage) presign(req *bindings.InvokeRequest) (*bindings.InvokeResponse, error) {
```

bindings/azure/blobstorage/blobstorage.go
JoshVanL: Done :+1:

bindings/azure/blobstorage/blobstorage.go
JoshVanL: Done :+1:

bindings/azure/blobstorage/blobstorage.go
JoshVanL: Done 👍

cicoyle (changes requested):
lgtm just conflicts

cicoyle (approved):

PR #4296, opened 2026-03-18
fwd port: ollama: add missing endpoint metadata field to spec (#4287)

cicoyle:
[cherrypick this PR to master](https://github.com/dapr/components-contrib/pull/4287)

PR #4314, opened 2026-03-25
fwd port: ollama: add missing endpoint metadata field to spec (#4287)

cicoyle:
supercede https://github.com/dapr/components-contrib/pull/4296 bc I ran the DCO commands and the history then looked off. Opened this one instead


sicoyle (approved):
thank you!

PR #4326, opened 2026-03-31
fix(pubsub/rabbitmq): use unique consumer tags to prevent subscription disruption on restart

javier-aliaga:
# Description

Fixes an issue where stopping and restarting a streaming subscription caused **all** active subscriptions to lose their RabbitMQ connection with `Exception (504) Reason: "channel/connection is not open"`.

**Root cause:** The component used the queue name as the AMQP consumer tag. When a subscription was restarted, `channel.Consume()` reused the same tag on the shared channel. RabbitMQ rejects duplicate consumer tags with a connection-level `not_allowed` exception, which kills the entire connection —  disrupting every subscription, not just the restarted one.

## Issue reference

Please reference the issue this PR will close: 
- https://github.com/dapr/components-contrib/issues/4328
- https://github.com/dapr/java-sdk/issues/1701

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/pubsub/rabbitmq/rabbitmq.go
+++ b/pubsub/rabbitmq/rabbitmq.go
@@ -564,13 +564,13 @@ func (r *rabbitMQ) subscribeForever(ctx context.Context, req pubsub.SubscribeReq
 			}
 
 			err = r.listenMessages(ctx, channel, msgs, req.Topic, handler)
+			// Always cancel the consumer server-side so RabbitMQ can
+			// release the registration. noWait=true avoids blocking if
+			// the channel is already closing.
+			if cancelErr := channel.Cancel(consumerTag, true); cancelErr != nil {
+				r.logger.Debugf("%s failed to cancel consumer %s: %v", logMessagePrefix, consumerTag, cancelErr)
+			}
 			if err != nil {
-				// Best-effort cleanup: cancel the consumer server-side so
-				// RabbitMQ can release the registration. noWait=true avoids
-				// blocking if the channel is already closing.
-				if cancelErr := channel.Cancel(consumerTag, true); cancelErr != nil {
-					r.logger.Debugf("%s failed to cancel consumer %s: %v", logMessagePrefix, consumerTag, cancelErr)
-				}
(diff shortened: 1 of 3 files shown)


pubsub/rabbitmq/rabbitmq.go
```diff
@@ -550,17 +565,25 @@ func (r *rabbitMQ) subscribeForever(ctx context.Context, req pubsub.SubscribeReq
...
-			ackCh <- true
+		if err != nil && strings.Contains(err.Error(), errorInvalidQueueType) {
+			if ackCh != nil {
+				ackCh <- true
+			}
 			return
 		}
 
-		if err == context.Canceled || err == context.DeadlineExceeded {
+		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
```
mikeee: More of a nit but I think for consistency and future proofing I might be worth adding an `err != nil` assertion to guard against future language/lib changes.

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -550,17 +565,25 @@ func (r *rabbitMQ) subscribeForever(ctx context.Context, req pubsub.SubscribeReq
 
 			err = r.listenMessages(ctx, channel, msgs, req.Topic, handler)
 			if err != nil {
+				// Best-effort cleanup: cancel the consumer server-side so
+				// RabbitMQ can release the registration. noWait=true avoids
+				// blocking if the channel is already closing.
+				if cancelErr := channel.Cancel(consumerTag, true); cancelErr != nil {
+					r.logger.Debugf("%s failed to cancel consumer %s: %v", logMessagePrefix, consumerTag, cancelErr)
+				}
```
mikeee: Should the consumer not always be cancelled regardless of error state - hence moved out of and above this error block or am I going vastly wrong?

javier-aliaga: I think you are right

mikeee (approved):

PR #4340, opened 2026-04-15
sentinel failover [repro]

MyMirelHub:
# Redis Sentinel Failover Certification Test Report (Repro)

This PR adds a test that reproduces how the Dapr `pubsub.redis` component behaves during a Redis Sentinel failover when the Redis client connection pool contains stale TCP connections.

It demonstrates a failure mode where `go-redis` can reuse a stale connection to the *old* master after a failover, causing publish calls to return errors such as `EOF` or `connection reset by peer` (depending on how the old master connection is terminated).

## How to run

Docker is required because the test spins up a containerized Redis master + replica + Sentinel.

```bash
cd tests/certification
go test -v -count=1 ./pubsub/redis/... -run TestRedisSentinelFailover
```
**Output**
```
--- FAIL: TestRedisSentinelFailover (96.25s) 
    --- FAIL: TestRedisSentinelFailover/Fails_With_Default_Timeouts (48.11s) 
        --- FAIL: TestRedisSentinelFailover/Fails_With_Default_Timeouts/Fails_With_Default_Timeouts (48.10s) 
    --- PASS: TestRedisSentinelFailover/Passes_With_10s_Timeouts (48.14s) 
        --- PASS: TestRedisSentinelFailover/Passes_With_10s_Timeouts/Passes_With_10s_Timeouts (48.14s) 
FAIL 
FAIL    github.com/dapr/components-contrib/tests/certification/pubsub/redis 96.893s 
```

## What this PR adds

- A single-container Redis+Replica+Sentinel setup:
  - `tests/certification/pubsub/redis/components/sentinel/docker-compose.yml`
- Two component configurations used by the test (the test selects the resources path):
  - `tests/certification/pubsub/redis/components/sentinel/fail_test/pubsub.yaml` (expected to fail)
  - `tests/certification/pubsub/redis/components/sentinel/pass_test/pubsub.yaml` (expected to pass; sets `idleTimeout` + `maxConnAge` to `10s`)
- The test itself:
  - `tests/certification/pubsub/redis/redis_test.go`

## Docker compose architecture

The compose file runs **three processes in one container**:
- Redis master on `6379`
- Redis replica on `6380`
- Redis Sentinel on `26379`

The component YAML points Dapr at Sentinel:
- `redisHost: "127.0.0.1:26379"`
- `failover: "true"`
- `sentinelMasterName: "mymaster"`

## How the test simulates the failure

For each test case, the flow is:

1. **Populate the pool**
   - Publishes 20 messages concurrently to encourage multiple connections in the client pool.
2. **Trigger Sentinel failover**
   - Runs (inside the container):
     - `redis-cli -p 26379 SENTINEL failover mymaster`
3. **Force socket closure of the old master**
   - Kills the process bound to port `6379` (the original master) to force TCP FIN/RST:
     - `kill -9 $(lsof -t -i:6379)`
4. **Wait**
   - Sleeps to allow failover completion + to make pool aging relevant.
5. **Publish again**
   - Attempts a second publish, which is expected to:
     - error in the default-timeouts case (stale connection reused), and
     - succeed in the `10s` timeouts case (pool prunes / connection re-dial happens).

Expected outcome is one subtest failing and one passing:
- `Fails With Default Timeouts` -> expects an error containing `EOF` or `connection reset`
- `Passes With 10s Timeouts` -> expects no error

## Why It Fails (Default Timeouts)

The first subtest (`Fails_With_Default_Timeouts`) uses `pubsub_fail.yaml`, which relies on the default `go-redis` connection pool settings.

*   **Default `idleTimeout`**: 5 minutes.
*   **Behavior**: When Dapr wakes up to publish the final message, the connections in its pool are only 15 seconds old. Because they haven't reached the 5-minute timeout, the internal `go-redis` reaper hasn't removed them. Dapr grabs a stale connection, attempts to write to it, and the OS immediately rejects it because the socket was closed by the `pkill` command. 
*   **Result**: The component crashes with an `EOF` or `connection reset by peer` error.

## Why It Passes (10s Mitigation)

The second subtest (`Passes_With_10s_Timeouts`) uses `pubsub_pass.yaml`, which explicitly configures the connection pool timeouts:

```yaml
- name: idleTimeout
  value: "10s" 
- name: maxConnAge
  value: "10s"
```

*   **Behavior**: By the time Dapr wakes up to publish the final message (15 seconds after the failover), the background `go-redis` reaper has identified that the connections have exceeded their 10-second maximum lifespan. The reaper automatically deletes the stale connections. When Dapr attempts to publish, it finds an empty pool, queries Sentinel for the new master address, and dials a fresh connection.
*   **Result**: The component successfully publishes the message without throwing an error.

## Notes / suggested next steps 

A potential mitigation to explore  would be to set more aggressive pool defaults automatically when `failover: true` is used (e.g., default `idleTimeout` / `maxConnAge` to ~10s when not provided), or to investigate an explicit pool reset strategy on failover.


MyMirelHub:
closing as this was only a repro demonstration

PR #4341, opened 2026-04-16
feat(state/oracledatabase): add BulkGet chunking to avoid Oracle IN-list limit

### LINKED ISSUE #4041: Oracle state store bulk querying improvements
### Problem
Currently the Oracle state store component will perform a single, unbounded retrieval of multiple keys using a(n) SQL `IN` condition.

For small numbers of keys, this is probably fine.  When dealing with larger numbers of items however, this approach can present some challenges:

- Can be rough on indexes
- Difficult to optimize
- (Might be difficult for execution plans?)

### Suggested improvements

While our general advice might be to suggest that users perform chunking on their end, there may still be a case for configurable chunking dapr-side as well:

- Allow a maximum page/chunk size to be configured
- When total number of keys exceeds the configured page size, dapr internally chunks
  - So, if page size is 100 and you have 1000 keys, it will do 100 keys 10 times...
- Perform chunking in series, not in parallel (this is important!)
- While we're here, it would be a good idea to do this Oracle specific tuneup to the query syntax: `WHERE id IN (SELECT COLUMN_VALUE FROM TABLE(:bind_array))`

## Release Note

RELEASE NOTE:

javier-aliaga:
# Description
## Summary
- Add configurable `bulkGetChunkSize` metadata option to the Oracle state store, which splits large `BulkGet` requests into sequential chunked SQL queries to avoid Oracle's hard 1000-expression IN-list limit ([#4041](https://github.com/dapr/components-contrib/issues/4041))
- Default chunk size is 1000 (Oracle's cap), so existing users with ≤1000 keys see **zero behavior change** — only the previously-crashing >1000-key path now works
- Chunks execute sequentially (not in parallel) per issue guidance; failed chunks don't abort remaining chunks (partial-success semantics preserved)

### Production code (`oracledatabaseaccess.go`)
- Added `BulkGetChunkSize` field to `oracleDatabaseMetadata` struct with `bulkGetChunkSize` metadata key
- Added `normalizeBulkGetChunkSize()` helper: defaults `<=0` to 1000, clamps `>1000` with a warning log
- Refactored `BulkGet` into an orchestrator (upfront key validation, fast-path single chunk, sequential chunking loop with `ctx.Err()` check between chunks) and `bulkGetChunk` (the existing single-query logic, extracted as-is)
- Called `normalizeBulkGetChunkSize` from `Init()` after metadata parsing

### Metadata (`metadata.yaml`)
- Added `bulkGetChunkSize` entry (type: number, default: 1000, range [1, 1000])
 
## Issue reference

Please reference the issue this PR will close: #4041

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
    * [ ] Created the dapr/docs PR: https://github.com/dapr/docs/pull/5122

### DIFF
--- a/state/oracledatabase/oracledatabase_integration_test.go
+++ b/state/oracledatabase/oracledatabase_integration_test.go
@@ -124,6 +124,10 @@ func TestOracleDatabaseIntegration(t *testing.T) {
 		testBulkGet(t, ods)
 	})
 
+	t.Run("Bulk get chunking exceeds Oracle IN-list limit", func(t *testing.T) {
+		testBulkGetChunkingExceedsLimit(t, ods)
+	})
+
 	t.Run("Update and delete with etag succeeds", func(t *testing.T) {
 		updateAndDeleteWithEtagSucceeds(t, ods)
 	})
@@ -730,6 +734,66 @@ func testBulkGet(t *testing.T, ods state.Store) {
 	}
 
 	err = ods.BulkDelete(t.Context(), deleteReq, state.BulkStoreOpts{})
+	require.NoError(t, err)
+}
+
+// testBulkGetChunkingExceedsLimit verifies that BulkGet succeeds when
+// requesting more keys than Oracle's IN-list expression limit (1000).
+// This exercises the sequential chunking path introduced to fix
+// https://github.com/dapr/components-contrib/issues/4041.
+func testBulkGetChunkingExceedsLimit(t *testing.T, ods state.Store) {
+	const totalKeys = 1100
+
+	// Seed keys
+	setReqs := make([]state.SetRequest, totalKeys)
+	for i := range setReqs {
+		setReqs[i] = state.SetRequest{
+			Key:   randomKey(),
+			Value: &fakeItem{Color: fmt.Sprintf("color-%d", i)},
+		}
+	}
+
+	err := ods.BulkSet(t.Context(), setReqs, state.BulkStoreOpts{})
+	require.NoError(t, err)
+
+	// Build BulkGet request plus one key that doesn't exist
+	getReqs := make([]state.GetRequest, totalKeys+1)
+	for i := range totalKeys {
+		getReqs[i] = state.GetRequest{Key: setReqs[i].Key}
+	}
+	getReqs[totalKeys] = state.GetRequest{Key: randomKey()} // non-existent
+
+	responses, err := ods.BulkGet(t.Context(), getReqs, state.BulkGetOpts{})
+	require.NoError(t, err)
+	require.Len(t, responses, totalKeys+1)
+
+	byKey := make(map[string]state.BulkGetResponse, len(responses))
+	for _, r := range responses {
+		byKey[r.Key] = r
+	}
+
+	// All seeded keys should be present with data
+	for i := range totalKeys {
+		expectedKey := setReqs[i].Key
+		r, ok := byKey[expectedKey]
+		require.True(t, ok, "expected bulk get response for key %s", expectedKey)
+		assert.Empty(t, r.Error, "key %s should not have error", expectedKey)
+		assert.NotNil(t, r.Data, "key %s should have data", expectedKey)
+		assert.NotNil(t, r.ETag, "key %s should have etag", expectedKey)
+	}
+
+	// Non-existent key should be present with no data
+	missingResp := byKey[getReqs[totalKeys].Key]
+	assert.Empty(t, missingResp.Error)
+	assert.Nil(t, missingResp.Data)
+	assert.Nil(t, missingResp.ETag)
+
+	// Clean up
+	deleteReqs := make([]state.DeleteRequest, totalKeys)
+	for i := range totalKeys {
+		deleteReqs[i] = state.DeleteRequest{Key: setReqs[i].Key}
+	}
+	err = ods.BulkDelete(t.Context(), deleteReqs, state.BulkStoreOpts{})
 	require.NoError(t, err)
 }
 
--- a/state/oracledatabase/oracledatabaseaccess.go
+++ b/state/oracledatabase/oracledatabaseaccess.go
@@ -39,6 +39,17 @@ const (
 	oracleWalletLocationKey    = "oracleWalletLocation"
 	errMissingConnectionString = "missing connection string"
 	defaultTableName           = "state"
+	// defaultBulkGetChunkSize is the default per-chunk size for BulkGet
+	// requests. Matches Oracle's documented IN-list expression limit so that
+	// requests with fewer keys than this default execute as a single query,
+	// preserving pre-chunking behavior.
+	// Invariant: defaultBulkGetChunkSize <= maxBulkGetChunkSize.
+	defaultBulkGetChunkSize = 1000
+	// maxBulkGetChunkSize is Oracle's hard limit on the number of
+	// expressions in a SQL IN list, per the Oracle SQL Language Reference.
+	// Values above this limit are rejected by Oracle with ORA-01795
+	// ("maximum number of expressions in a list is 1000").
+	maxBulkGetChunkSize = 1000
 )
 
 // oracleDatabaseAccess implements dbaccess.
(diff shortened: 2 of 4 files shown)


state/oracledatabase/oracledatabase_integration_test.go
```diff
@@ -733,6 +737,64 @@ func testBulkGet(t *testing.T, ods state.Store) {
...
+	}
+
+	err := ods.BulkSet(t.Context(), setReqs, state.BulkStoreOpts{})
+	require.NoError(t, err)
+
+	// Build BulkGet request plus one key that doesn't exist
+	getReqs := make([]state.GetRequest, totalKeys+1)
+	for i := range totalKeys {
+		getReqs[i] = state.GetRequest{Key: setReqs[i].Key}
+	}
```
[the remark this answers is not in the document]
javier-aliaga: not true

state/oracledatabase/oracledatabase_integration_test.go
```diff
@@ -733,6 +737,64 @@ func testBulkGet(t *testing.T, ods state.Store) {
...
+	missingResp := byKey[getReqs[totalKeys].Key]
+	assert.Empty(t, missingResp.Error)
+	assert.Nil(t, missingResp.Data)
+	assert.Nil(t, missingResp.ETag)
+
+	// Clean up
+	deleteReqs := make([]state.DeleteRequest, totalKeys)
+	for i := range totalKeys {
+		deleteReqs[i] = state.DeleteRequest{Key: setReqs[i].Key}
+	}
```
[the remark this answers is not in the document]
javier-aliaga: not true

state/oracledatabase/oracledatabase_integration_test.go
```diff
@@ -733,6 +737,64 @@ func testBulkGet(t *testing.T, ods state.Store) {
...
+	for _, r := range responses {
+		byKey[r.Key] = r
+	}
+
+	// All seeded keys should be present with data
+	for i := range totalKeys {
+		r := byKey[setReqs[i].Key]
+		assert.Empty(t, r.Error, "key %s should not have error", r.Key)
+		assert.NotNil(t, r.Data, "key %s should have data", r.Key)
+		assert.NotNil(t, r.ETag, "key %s should have etag", r.Key)
```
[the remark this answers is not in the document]
javier-aliaga: not true

state/oracledatabase/oracledatabaseaccess.go
```diff
@@ -312,20 +346,74 @@ func (o *oracleDatabaseAccess) Get(ctx context.Context, req *state.GetRequest) (
...
+
+	// Fast path: input fits in one chunk — no allocation overhead.
+	if len(req) <= chunkSize {
+		return o.bulkGetChunk(ctx, req)
+	}
+
+	// Chunked path: sequential chunks, results concatenated.
+	res := make([]state.BulkGetResponse, 0, len(req))
+	for start := 0; start < len(req); start += chunkSize {
+		// Check context between chunks so a cancelled context stops promptly
```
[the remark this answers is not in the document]
javier-aliaga: This is an internal struct — `BulkGet` is never called without `Init`, same as `Get`, `Set`, `Delete`, and `KeysLike` which all assume `o.db` is initialized. Adding a defensive guard here would be inconsistent with the rest of the component. Won't fix.

state/oracledatabase/oracledatabaseaccess.go
```diff
@@ -39,6 +39,17 @@ const (
 	oracleWalletLocationKey    = "oracleWalletLocation"
 	errMissingConnectionString = "missing connection string"
 	defaultTableName           = "state"
+	// defaultBulkGetChunkSize is the default per-chunk size for BulkGet
+	// requests. Matches Oracle's documented IN-list expression limit so that
+	// requests with fewer keys than this default execute as a single query,
```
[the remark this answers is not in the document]
javier-aliaga: Good catch — fixed in the latest push. The comment now reads "requests with this many keys or fewer".

state/oracledatabase/oracledatabaseaccess.go
```diff
@@ -312,20 +346,74 @@ func (o *oracleDatabaseAccess) Get(ctx context.Context, req *state.GetRequest) (
...
+	chunkSize := o.metadata.BulkGetChunkSize
+
+	// Fast path: input fits in one chunk — no allocation overhead.
+	if len(req) <= chunkSize {
+		return o.bulkGetChunk(ctx, req)
+	}
+
+	// Chunked path: sequential chunks, results concatenated.
+	res := make([]state.BulkGetResponse, 0, len(req))
+	for start := 0; start < len(req); start += chunkSize {
```
acroca, line 379: Would it be worth fetching each chunk in parallel instead of sequential?

state/oracledatabase/oracledatabaseaccess_test.go
```diff
@@ -563,3 +565,342 @@ func TestBulkGetEmptyKeyValidation(t *testing.T) {
...
+			expected:   500,
+		},
+		{
+			name:       "exact max boundary",
+			configured: 1000,
+			expected:   1000,
+		},
+		{
+			name:       "above max clamped to max",
+			configured: 2000,
```
acroca, line 606:
```suggestion
			configured: maxBulkGetChunkSize+1,
```

state/oracledatabase/oracledatabaseaccess_test.go
```diff
@@ -563,3 +565,342 @@ func TestBulkGetEmptyKeyValidation(t *testing.T) {
...
+		},
+		{
+			name:       "valid value 500",
+			configured: 500,
+			expected:   500,
+		},
+		{
+			name:       "exact max boundary",
+			configured: 1000,
+			expected:   1000,
```
acroca, line 602:
```suggestion
			configured: maxBulkGetChunkSize,
			expected:   maxBulkGetChunkSize,
```

state/oracledatabase/oracledatabaseaccess.go
javier-aliaga: No, it would add complexity but no too much benefit. We need the result to be the same order as the input so we would need to coordinate the chunks after the completion

acroca (approved):

state/oracledatabase/oracledatabaseaccess.go
```diff
@@ -312,20 +346,74 @@ func (o *oracleDatabaseAccess) Get(ctx context.Context, req *state.GetRequest) (
...
+// The default chunk size equals Oracle's IN-list limit (1000), so callers
+// with <= 1000 keys see identical behavior to the pre-chunking implementation.
 func (o *oracleDatabaseAccess) BulkGet(ctx context.Context, req []state.GetRequest) ([]state.BulkGetResponse, error) {
 	if len(req) == 0 {
 		return []state.BulkGetResponse{}, nil
 	}
 
+	// Validate all keys upfront — an empty key is a programmer bug, not a
+	// per-key data issue, so we fail fast before issuing any query.
+	for _, r := range req {
```
JoshVanL, line 364: Not sure why we need to be doing this here?

state/oracledatabase/oracledatabaseaccess.go
```diff
@@ -312,20 +346,74 @@ func (o *oracleDatabaseAccess) Get(ctx context.Context, req *state.GetRequest) (
...
+		end := start + chunkSize
+		if end > len(req) {
+			end = len(req)
+		}
+
+		chunkRes, err := o.bulkGetChunk(ctx, req[start:end])
+		if err != nil {
+			// bulkGetChunk only returns a hard error for programming bugs
+			// (e.g. more rows than expected). Propagate it.
+			return nil, err
```
JoshVanL, line 401: Above we get nice per key errors, but here we get a single error and no res

state/oracledatabase/oracledatabaseaccess.go
```diff
@@ -312,20 +346,74 @@ func (o *oracleDatabaseAccess) Get(ctx context.Context, req *state.GetRequest) (
...
+// bulkGetChunk executes a single IN-clause query for the given requests.
+// Pre-condition: req is non-empty and all keys are non-empty.
+func (o *oracleDatabaseAccess) bulkGetChunk(ctx context.Context, req []state.GetRequest) ([]state.BulkGetResponse, error) {
 	// Oracle supports the IN operator for bulk operations
 	// Build the IN clause with bind variables
 	// Oracle uses :1, :2, etc. for bind variables in the IN clause
 	params := make([]any, len(req))
 	bindVars := make([]string, len(req))
 	for i, r := range req {
-		if r.Key == "" {
```
JoshVanL, line 326: I see, it comes from here- still don't really get why we need to do this at all at this level

JoshVanL (changes requested):

state/oracledatabase/oracledatabaseaccess.go
javier-aliaga: mostly for consistency, the GET(line 301) does the same check, saying so, we can drop it

PR #4358, opened 2026-04-23
state: native bulk ops and DeleteWithPrefix

JoshVanL:
Improves the workflow and actor state paths by cutting avoidable round trips on the backends most commonly used with Dapr workflows. Each change is additive; existing call sites and defaults are unchanged.

  postgresql/v2:
  - Native BulkSet using a single multi-row UPSERT over UNNEST arrays (keys, values, TTL seconds). TTL is computed server-side via now() + make_interval(secs => $n) so expires_at has the same semantics as doSet's
  single-item path. Validates each request through state.CheckRequestOptions and rejects empty keys, matching the per-item path. Falls back to the default per-item path when any request carries an ETag or
  first-write concurrency option — per-row CAS can't be expressed as a multi-row upsert without losing atomicity.
  - Native BulkDelete with key = ANY($1), same ETag fallback; rejects empty keys.
  - DeleteWithPrefix — single DELETE ... WHERE key LIKE $1 server-side. LIKE metacharacters (%, _, \) in the caller-supplied prefix are escaped so a prefix containing them cannot widen the delete set.
  - Advertises FeatureDeleteWithPrefix.

  sqlserver/v2:
  - Native BulkGet builds a dynamic IN clause chunked at 2000 parameters (SQL Server's 2100-parameter ceiling with headroom). [Key] is in the SELECT projection so rows map back to request order without relying on
  driver row ordering. Duplicate keys in the input are supported — the implementation fetches distinct keys over the wire and builds the response by iterating the original request slice, so every occurrence receives
   the same data.
  - DeleteWithPrefix — single DELETE ... WHERE [Key] LIKE @Prefix ESCAPE '\'. Metacharacters (%, _, [, ], \) are escaped.
  - Advertises FeatureDeleteWithPrefix.

  mysql:
  - DeleteWithPrefix — single DELETE ... WHERE id LIKE ?. LIKE metacharacters are escaped; relies on MySQL's default \ LIKE escape char (no explicit ESCAPE clause needed since parameter binding delivers the pattern
  verbatim).
  - Advertises FeatureDeleteWithPrefix.

  sqlite:
  - DeleteWithPrefix threaded through the DBAccess interface and exposed on the store. Single DELETE ... WHERE key LIKE ? ESCAPE '\' (SQLite has no default LIKE escape char, so the clause is required).
  - Advertises FeatureDeleteWithPrefix.

  etcd:
  - Native BulkGet batches N OpGet into a single Txn RPC, respecting the configured maxTxnOps by chunking serially with a fresh per-chunk timeout context.
  - DeleteWithPrefix — single server-side range delete via client.Delete(prefix, WithPrefix()).
  - Advertises FeatureDeleteWithPrefix.

  in-memory:
  - DeleteWithPrefix simplified to match the server-side prefix semantics (straight HasPrefix match, no separator filter).

  Conformance / certification:
  - tests/conformance/state/state.go — the delete with prefix case now exercises the server-side prefix contract (keys whose names begin with the supplied prefix are all removed, including nested-looking keys
  sharing that prefix; unrelated keys whose names merely contain the prefix as a substring are unaffected).
  - tests/config/state/tests.yml — opts postgresql.v2.*, mysql.*, sqlite, etcd.v1/v2, and sqlserver.v2(.docker) into the existing delete-with-prefix conformance operation, so all the new implementations are
  exercised end-to-end against the docker-compose'd backends.

  Semantics note:
  - DeleteWithPrefix is defined as a literal-prefix server-side delete on every backend. The caller is expected to pass a fully-qualified prefix ending with DaprSeparator (||) — typically the actor's composite key
  appID||actorType||actorID|| — and every key beginning with that prefix is removed. LIKE metacharacters in the prefix are always escaped so they match literally.

### DIFF
--- a/state/in-memory/in_memory.go
+++ b/state/in-memory/in_memory.go
@@ -124,28 +124,18 @@ func (store *InMemoryStore) Delete(ctx context.Context, req *state.DeleteRequest
 }
 
 func (store *InMemoryStore) DeleteWithPrefix(ctx context.Context, req state.DeleteWithPrefixRequest) (state.DeleteWithPrefixResponse, error) {
-	// step1: validate parameters
-	err := req.Validate()
-	if err != nil {
+	if err := req.Validate(); err != nil {
 		return state.DeleteWithPrefixResponse{}, err
 	}
 
-	// step2 should be protected by write-lock
 	store.lock.Lock()
 	defer store.lock.Unlock()
 
-	// step2: do really delete
-	// this operation won't fail
 	var count int64
-
 	for key := range store.items {
 		if strings.HasPrefix(key, req.Prefix) {
-			// The string contains the prefix, now we check to make sure there aren't more || after
-			longerPrefix := strings.Contains(key[len(req.Prefix):], "||")
-			if !longerPrefix {
-				delete(store.items, key)
-				count++
-			}
+			delete(store.items, key)
+			count++
 		}
 	}
 	return state.DeleteWithPrefixResponse{Count: count}, nil
--- a/state/sqlserver/v2/sqlserver.go
+++ b/state/sqlserver/v2/sqlserver.go
@@ -439,32 +439,23 @@ func escapeSQLServerLikePattern(s string) string {
 	return s
 }
 
-// DeleteWithPrefix removes direct-children keys of the given prefix in one
-// server-side query. Matches in-memory "no-nested-||" semantics.
-//
-// The prefix is treated literally: LIKE metacharacters (%, _, [, ]) in the
-// caller-supplied prefix are escaped so a prefix containing those
-// characters cannot widen the delete set.
+// DeleteWithPrefix removes every row whose [Key] begins with the given
+// prefix in a single server-side DELETE. The prefix is treated literally:
+// LIKE metacharacters (%, _, [, ]) are escaped so a prefix containing
+// those characters cannot widen the delete set.
 func (s *SQLServer) DeleteWithPrefix(parentCtx context.Context, req state.DeleteWithPrefixRequest) (state.DeleteWithPrefixResponse, error) {
 	if err := req.Validate(); err != nil {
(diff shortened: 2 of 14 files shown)


state/in-memory/in_memory.go
```diff
@@ -124,28 +124,18 @@ func (store *InMemoryStore) Delete(ctx context.Context, req *state.DeleteRequest
...
 
-	// step2: do really delete
-	// this operation won't fail
 	var count int64
-
 	for key := range store.items {
 		if strings.HasPrefix(key, req.Prefix) {
-			// The string contains the prefix, now we check to make sure there aren't more || after
-			longerPrefix := strings.Contains(key[len(req.Prefix):], "||")
-			if !longerPrefix {
```
acroca, line 145: Why was it like that previously? And why don't we need it anymore? feels like a breaking change...

state/sqlserver/v2/sqlserver.go
```diff
@@ -307,6 +309,163 @@ func (s *SQLServer) Get(ctx context.Context, req *state.GetRequest) (*state.GetR
...
+				rows.Close()
+				return nil, err
+			}
+			var value []byte
+			if isBinary {
+				value = binaryData
+			} else if data.Valid {
+				value = []byte(data.String)
+			}
+			rowByKey[key] = row{
```
javier-aliaga, line 391: You have same problem as https://github.com/dapr/components-contrib/pull/4365#discussion_r3217455177

Default case for sqlserver is case-insensitive, here you are storing using the key returned by the scan but then on line 411 you uses the key from the request that may not match the same case.

Probably for dapr internals this is fine as we are consistent with the case, but may produce unexpected results for users. This also happens for mysql and sqlite.

tbh, this is a user missconfiguration, but as I was flagged in my PR I just wanted to rise it here

PR #4366, opened 2026-04-29
feat(pubsub/kafka): add PausableSubscriber + harden graceful shutdown

javier-aliaga:
# Description

## Summary

- Adds an optional `PausableSubscriber` capability to the pubsub interface so the runtime can stop the broker from feeding new messages into the local buffer during graceful shutdown without tearing down the session.
- Implements the new capability for Kafka via Sarama's `PauseAll` / `ResumeAll`.
- Hardens Kafka's existing graceful-shutdown path — fixes a multi-subscription race that was causing `consumer group was closed` log spam, and demotes a misleading retry-exhausted log to debug when the cause is shutdown rather than real exhaustion.

### New: `PausableSubscriber` interface (`pubsub/pubsub.go`)

Optional capability for pubsub components that can pause fetching from the broker without tearing down the session. While paused, the connection and partition assignments stay alive so messages already buffered locally can still be processed by handlers — this enables a clean drain-during-shutdown pattern in the runtime.

```go
type PausableSubscriber interface {
    Pause(ctx context.Context) error
    Resume(ctx context.Context) error
}
```

Detected by callers via interface assertion. Brokers that don't implement it fall back to existing behavior.

### Kafka implementation (`common/component/kafka/kafka.go`, `pubsub/kafka/kafka.go`)

`Kafka.Pause` / `Kafka.Resume` delegate to Sarama's `consumerGroup.PauseAll()` / `ResumeAll()`. The `pubsub/kafka.PubSub` wrapper exposes them through a closed-state guard. Compile-time interface assertion in `pubsub/kafka/kafka_test.go` ensures the wrapper stays in sync with the interface.

### `subscriber.go`: pause-before-close ordering on graceful shutdown

When the Subscribe context is cancelled with `pubsub.ErrGracefulShutdown` cause, the contrib's Subscribe goroutine now calls `consumerGroup.PauseAll()` before the existing `reloadConsumerGroup` + `Close` sequence. This stops Sarama from issuing further FetchRequests while the consume goroutine winds down — keeps the close path quieter and bounds any last-second claim-buffer growth.

### `subscriber.go`: only the last subscription closes the shared consumer group

When multiple Subscribe calls share a single Kafka pubsub component (one declarative Subscription per topic, all on the same component), graceful shutdown fires `sub.Stop(ErrGracefulShutdown)` for every Subscription in parallel. Each Subscribe goroutine then takes `subscribeLock` serially, deletes its own topic, calls `reloadConsumerGroup()`, and runs the postAction.

Before this fix, the postAction unconditionally called `consumerGroup.Close()` — so the FIRST goroutine to acquire the lock tore down the shared consumer group while sibling goroutines still held subscriptions. Each subsequent `reloadConsumerGroup` then started a fresh `Consume()` against the now-closed group, producing log spam like:

```
level=error msg="Error consuming [topic1 topic2 topic3]. Retrying...:
kafka: tried to use a consumer group that was closed"
```

Gate `consumerGroup.Close()` on `len(k.subscribeTopics) == 0` so only the LAST goroutine to exit closes the group. Earlier goroutines just delete their topic and let reload restart consume for the remaining topics on a still-alive group.

This race was pre-existing — introduced by #3907 (the original `postAction`) and only became visible when running a multi-topic Kafka graceful shutdown. The Pause-before-Close commit on this branch made it more visible (Pause runs before delete, so the broker is paused while Close is still racing with siblings) but did not introduce it.

### `consumer.go`: demote retry-exhausted log when shutdown is the cause

`retry.NotifyRecover`'s "exhausted" return value covers two cases:

1. **Real exhaustion** — a healthy session retried the configured number of times and the handler kept failing. Genuine processing failure worth `Errorf`.
2. **Session context canceled** — graceful shutdown or rebalance canceled the backoff context, so the retry returned with the last observed error. The message is left unmarked and will be redelivered to whoever takes over the partition. Shutdown noise, not a failure.

Distinguish by checking `session.Context().Err()` and the error itself for `context.Canceled`. If either matches, log `Debugf` with a clearer message ("aborted due to shutdown; will be redelivered"); otherwise keep the existing `Errorf` for true exhaustion. Same change applied to the bulk path.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/common/component/kafka/consumer.go
+++ b/common/component/kafka/consumer.go
@@ -91,14 +91,18 @@ func (consumer *consumer) ConsumeClaim(session sarama.ConsumerGroupSession, clai
 					}, func() {
 						consumer.k.logger.Infof("Successfully processed Kafka message after it previously failed: %s/%d/%d [key=%s]", message.Topic, message.Partition, message.Offset, asBase64String(message.Key))
 					}); err != nil {
-						consumer.k.logger.Errorf("Too many failed attempts at processing Kafka message: %s/%d/%d [key=%s]. Error: %v.", message.Topic, message.Partition, message.Offset, asBase64String(message.Key), err)
-						if errors.Is(session.Context().Err(), context.Canceled) {
-							// If the context is canceled, we should not attempt to consume any more messages. Exiting the loop.
-							// Otherwise, there is a race condition when this loop keeps processing messages from the claim.Messages() channel
-							// before the session.Context().Done() is closed. If there are other messages that can successfully be processed,
-							// they will be marked as processed and this failing message will be lost.
+						// When the session context is canceled (graceful
+						// shutdown / rebalance), the retry exits with the last
+						// observed error rather than a real "exhausted"
+						// outcome. The message is left unmarked and will be
+						// redelivered to whichever consumer takes over the
+						// partition, so this is not a processing failure —
+						// just shutdown noise. Demote the log accordingly.
+						if errors.Is(session.Context().Err(), context.Canceled) || errors.Is(err, context.Canceled) {
+							consumer.k.logger.Debugf("Kafka message processing aborted due to shutdown; will be redelivered: %s/%d/%d [key=%s]", message.Topic, message.Partition, message.Offset, asBase64String(message.Key))
 							return nil
 						}
+						consumer.k.logger.Errorf("Too many failed attempts at processing Kafka message: %s/%d/%d [key=%s]. Error: %v.", message.Topic, message.Partition, message.Offset, asBase64String(message.Key), err)
 					}
 				} else {
 					err := consumer.doCallback(session, message)
@@ -124,7 +128,13 @@ func (consumer *consumer) flushBulkMessages(claim sarama.ConsumerGroupClaim,
 			}, func() {
 				consumer.k.logger.Infof("Successfully processed Kafka message after it previously failed: %s", claim.Topic())
 			}); err != nil {
-				consumer.k.logger.Errorf("Too many failed attempts at processing Kafka message: %s. Error: %v.", claim.Topic(), err)
+				// Same shutdown-vs-exhausted distinction as the singular
+				// path: demote when session ctx is canceled.
+				if errors.Is(session.Context().Err(), context.Canceled) || errors.Is(err, context.Canceled) {
+					consumer.k.logger.Debugf("Kafka bulk message processing aborted due to shutdown; will be redelivered: %s", claim.Topic())
+				} else {
+					consumer.k.logger.Errorf("Too many failed attempts at processing Kafka message: %s. Error: %v.", claim.Topic(), err)
+				}
 			}
 		} else {
 			err := consumer.doBulkCallback(session, messages, handler, claim.Topic())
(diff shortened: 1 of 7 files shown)


common/component/kafka/kafka.go
```diff
@@ -319,6 +319,32 @@ func (k *Kafka) ValidateAWS(metadata map[string]string) (awsAuth.Options, error)
 	}, nil
 }
 
+// Pause stops fetching new messages from the broker for all active
+// subscriptions on this component. The Sarama session and partition
+// assignments are preserved so messages already buffered in claim queues can
+// still be processed by handlers. Idempotent.
+func (k *Kafka) Pause(ctx context.Context) error {
```
cicoyle, line 326: For this func `Pause` and `Resume` we take in the ctx but dont use it - can we either not accept the ctx in the funcs or use them? Should we check for `ctx.Err()` in these funcs?

common/component/kafka/subscriber.go
```diff
@@ -37,39 +37,51 @@ func (k *Kafka) Subscribe(ctx context.Context, handlerConfig SubscriptionHandler
...
+
+		// Only the last subscription to exit closes the consumer group.
+		// Closing while sibling Subscribe goroutines still hold subscriptions
+		// on the same component would race with the new consume goroutine
+		// reloadConsumerGroup just started for the remaining topics, leaving
+		// it to error out with "tried to use a consumer group that was
+		// closed".
+		if isGraceful && len(k.subscribeTopics) == 0 && k.clients != nil && k.clients.consumerGroup != nil {
+			k.logger.Debugf("Last subscription closing; closing consumer group.")
+			if err := k.clients.consumerGroup.Close(); err != nil {
```
cicoyle, line 81: Should we still have a non graceful `Close` path? I see `Close` is gated by `isGraceful`

cicoyle (changes requested):
mostly lgtm, just a few comments

common/component/kafka/subscriber.go
javier-aliaga: There is no need, it is closed when Kafka.Close()

common/component/kafka/consumer.go
```diff
@@ -124,7 +128,13 @@ func (consumer *consumer) flushBulkMessages(claim sarama.ConsumerGroupClaim,
...
 				consumer.k.logger.Infof("Successfully processed Kafka message after it previously failed: %s", claim.Topic())
 			}); err != nil {
-				consumer.k.logger.Errorf("Too many failed attempts at processing Kafka message: %s. Error: %v.", claim.Topic(), err)
+				// Same shutdown-vs-exhausted distinction as the singular
+				// path: demote when session ctx is canceled.
+				if errors.Is(session.Context().Err(), context.Canceled) || errors.Is(err, context.Canceled) {
+					consumer.k.logger.Debugf("Kafka bulk message processing aborted due to shutdown; will be redelivered: %s", claim.Topic())
+				} else {
+					consumer.k.logger.Errorf("Too many failed attempts at processing Kafka message: %s. Error: %v.", claim.Topic(), err)
+				}
```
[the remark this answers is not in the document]
javier-aliaga: This is not a problem introduced by this PR

cicoyle (approved):

PR #4395, opened 2026-06-04
add retriable/terminal error classification for pubsub publish

cicoyle:
Pubsub components return plain Go errors from `Publish`/`BulkPublish` (the raw broker SDK error or a `fmt.Errorf` wrap). These carry no status code, so Dapr's resiliency layer cannot distinguish a transient failure (broker unavailable timeout) from a terminal one (invalid topic, component closed) & it retries every publish error up to `maxRetries` and silently ignores any configured retry `matching` (httpStatusCodes / gRPCStatusCodes).
  
This PR adds a shared helper so components can classify their publish errors, which the runtime then reads via `status.FromError`


atrauzzi:
Closes: #4397

JoshVanL (approved):

PR #4413, opened 2026-06-17
fix(kafka): add connection timeouts, broker health check, and tunable producer config

nelson-parente:
## Summary

The shared Kafka client (`common/component/kafka/`) had three confirmed production gaps that could surface in customer RCAs:

1. **No init/connection timeout** — `Init()` could block indefinitely on unreachable brokers because Sarama's `Net.DialTimeout` / `Metadata.Timeout` were not set.
2. **No health check** — no method existed to validate broker connectivity after init; failures only surfaced later at publish/consume time.
3. **Untunable client config** — producer `RequiredAcks` and `Retry.Max` were hard-coded; Net timeouts relied on Sarama defaults with no override path.

### Changes

**`common/component/kafka/metadata.go`**
- New duration fields: `dialTimeout`, `readTimeout`, `writeTimeout` (defaults: 30 s each, matching Sarama's built-in defaults, fully backward-compatible when unset).
- New duration field: `metadataTimeout` (default: 0 / disabled, matching Sarama's built-in default).
- New string field: `producerRequiredAcks` — accepts `"all"` (default), `"local"`, `"none"`; maps to `sarama.WaitForAll` / `WaitForLocal` / `NoResponse`.
- New int field: `producerRetryMax` (default: 5, matching the previously hard-coded value).

**`common/component/kafka/kafka.go`**
- Applies `Net.DialTimeout`, `Net.ReadTimeout`, `Net.WriteTimeout`, and `Metadata.Timeout` from metadata during `Init()`.
- Stores a `ProducerConfig` on the `Kafka` struct for use by client factories.

**`common/component/kafka/producer.go`**
- Introduces `ProducerConfig` struct (RequiredAcks, RetryMax, MaxMessageBytes).
- `GetSyncProducer` signature changed from `(config, brokers, maxMessageBytes int)` to `(config, brokers, ProducerConfig)` — callers within this package updated accordingly.

**`common/component/kafka/clients.go`** / **`common/component/kafka/aws.go`**
- Both client creation paths (static and AWS IAM) now thread `ProducerConfig` so the new tunables are honoured.

**`common/component/kafka/health.go`** (new)
- `Ping(ctx context.Context) error` — opens a short-lived Sarama client against the configured brokers to verify connectivity. Respects context deadlines and clamps Net timeouts to the remaining deadline. Additive and does not modify internal state.

**`common/component/kafka/health_test.go`** (new)
- Unit tests for `Ping`: uninitialised component, cancelled context, deadline-already-exceeded, and unreachable-broker cases.

**`common/component/kafka/metadata_test.go`**
- Table-driven tests for: default Net timeouts, explicit Net timeouts, invalid (zero/negative) timeouts falling back to defaults, `producerRequiredAcks` all valid values + invalid, `producerRetryMax` defaults/explicit/zero/negative.

**`pubsub/kafka/metadata.yaml`** / **`bindings/kafka/metadata.yaml`**
- New fields added with descriptions, defaults, and examples matching existing entries.

### Backward compatibility

With no new metadata set, all fields default to the values that were previously hard-coded or Sarama's built-in defaults. Existing deployments require no configuration changes.

### Caveats

- `metadata-bundle` regeneration (normally triggered by `make bundle`) is skipped in this worktree — the `.build-tools` bundler panics when run from a git-worktree checkout (`.git` is a file, not a dir). CI will regenerate the bundle. The YAML schema files were updated manually and are consistent with the code.
- `Ping()` is exposed as a method on `*Kafka` but is not wired into any existing health-check hook (none exists in this package). Callers such as the runtime health endpoint can call it directly.

## Testing

✅ `go build ./common/component/kafka/...` — clean  
✅ `go test ./common/component/kafka/...` — 186 tests pass (was 166 before this PR; 20 new tests added)  
✅ `gofumpt -l -w` on all changed files — no remaining diffs  
✅ `goimports -w` on all changed files — clean  
✅ `golangci-lint run ./common/component/kafka/...` — no issues  
✅ `go vet ./common/component/kafka/...` — no issues  
☐ Integration / e2e tests against a real Kafka cluster — pending CI  
☐ `make bundle` (metadata bundle regen) — skipped (worktree limitation, see caveats)  

## Related

- Docs PR: (link to docs PR once open)


nelson-parente:
## Expert code review — OSS-1152 Kafka hardening

Reviewed by a Go correctness/concurrency pass and a security/robustness pass, with the High findings verified against the source. The **active changes (tunable net timeouts + producer config) are sound and backward-compatible** — net-timeout defaults (30s) match Sarama v1.48.1, `metadataTimeout=0` is Sarama's computed-default sentinel, and producer defaults reproduce the prior hardcoded `WaitForAll`/`5`. The issues are concentrated in the new `Ping()` health method (currently not yet wired into a runtime hook) plus one live-path config mutation.

**Findings (fixes being applied in this session, pushed to this branch):**

| Sev | Location | Issue | Status |
|---|---|---|---|
| High | `aws.go` `getSyncProducer()` | Mutates the **shared `*k.config`** pointer in place; the non-AWS `GetSyncProducer` correctly copies by value | fixing |
| High | `health.go` `Ping()` | Shallow config copy shares the SASL `TokenProvider` → docstring's "safe for concurrent use" is false under OIDC/MSK (token-cache race) | fixing (mutex on token sources) |
| High | `health.go` `Ping()` | `sarama.NewClient` is blocking/ctx-unaware; only Net.* clamped, `Metadata.Timeout`/`Retry` unbounded → "won't block indefinitely" is overstated without a deadline | fixing (deadline-bound + goroutine race on ctx) |
| Med/High | `health.go` `Ping()` | Inherits `Metadata.Full=true` → full-cluster topic scan + false-negative on least-privilege principals | fixing (`Metadata.Full=false`) |
| Med | metadata parse | `producerRequiredAcks=none` (silent data loss) has no warning log, unlike `skipVerify` | fixing (Warnf) |
| Low | `aws.go` / `clients.go` | Dead `KafkaOptions.MaxMessageBytes` field (superseded by `ProducerConfig`) | fixing (remove) |
| Low | `health_test.go` | Deadline test should `ErrorIs(DeadlineExceeded)`; no mid-flight-cancel test | fixing |

**Noted, intentionally out of scope here:**
- Exported `GetSyncProducer` signature changed (`maxMessageBytes int` → `ProducerConfig`) — a source break for external importers (e.g. downstream forks); unavoidable to deliver the feature.
- `producerRetryMax` has no upper cap; `dialTimeout: 0` silently falls back to 30s — minor UX, left as-is.

All fixes validated with `go test`, **`go test -race`**, and lint before push.

common/component/kafka/health.go
```diff
@@ -0,0 +1,130 @@
...
+	// Metadata.Full=false avoids a full topic scan which would be expensive and
+	// could fail on least-privilege principals that cannot list all topics.
+	cfgCopy := *k.config
+	cfgCopy.Metadata.Full = false
+
+	if deadline, ok := ctx.Deadline(); ok {
+		remaining := time.Until(deadline)
+		if remaining <= 0 {
+			return errors.New("kafka health check: context deadline already exceeded")
+		}
```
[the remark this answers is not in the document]
nelson-parente: Fixed in b8a90bba: the `remaining <= 0` branch now wraps `context.DeadlineExceeded` directly instead of `ctx.Err()`, so the error always satisfies `errors.Is(err, context.DeadlineExceeded)` even in the narrow window where `ctx.Err()` is not yet set. Resolving.

common/component/kafka/health_test.go
```diff
@@ -0,0 +1,143 @@
...
+	cfg.Net.ReadTimeout = 30 * time.Second
+	cfg.Net.WriteTimeout = 30 * time.Second
+	cfg.Metadata.Retry.Max = 0
+
+	k := &Kafka{
+		logger: logger.NewLogger("kafka_test"),
+		config: cfg,
+		// 192.0.2.0/24 is TEST-NET-1 (RFC 5737) — packets are blackholed.
+		brokers: []string{"192.0.2.1:9092"},
+	}
```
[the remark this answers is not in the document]
nelson-parente: Fixed in b8a90bba: the test no longer relies on TEST-NET (192.0.2.1) blackholing. It now stands up a local listener that accepts the connection but never responds, so the dial connects and then blocks on the broker handshake — making the in-flight cancellation deterministic across CI network stacks. Resolving.

common/component/kafka/health.go
```diff
@@ -0,0 +1,132 @@
...
+
+	if k.config == nil {
+		return errors.New("kafka health check: component not initialised")
+	}
+
+	// Build a minimal config copy that honours the context deadline.
+	// Metadata.Full=false avoids a full topic scan which would be expensive and
+	// could fail on least-privilege principals that cannot list all topics.
+	cfgCopy := *k.config
+	cfgCopy.Metadata.Full = false
```
acroca, line 66: The docstring says `Metadata.Full=false` makes this a lightweight probe with "no topic-level scan", but I don't think that's what happens on this code path.

In sarama (v1.48.1), `Metadata.Full` is only consulted in two places: the initial fetch in `NewClient`, and the background updater's `refreshMetadata()`, which narrows to topics already in use. The explicit no-args `RefreshMetadata()` call used here never checks it, per its interface contract: *"If no topics are provided, it will refresh metadata for all topics."*

So every `Ping()` issues a full-cluster topic metadata request. On clusters with thousands of topics that's an expensive call for a health probe that may be polled every few seconds. `Full=false` here only makes `NewClient` skip its construction-time fetch (as the inline comment below correctly notes).

common/component/kafka/health.go
```diff
@@ -0,0 +1,132 @@
+/*
+Copyright 2024 The Dapr Authors
```
acroca:
```suggestion
Copyright 2026 The Dapr Authors
```

common/component/kafka/health_test.go
```diff
@@ -0,0 +1,160 @@
+/*
+Copyright 2024 The Dapr Authors
```
acroca:
```suggestion
Copyright 2026 The Dapr Authors
```

common/component/kafka/health_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+	"github.com/IBM/sarama"
+	"github.com/stretchr/testify/require"
+
+	"github.com/dapr/kit/logger"
+)
+
+// TestPingUninitialised verifies that Ping returns an error when the
+// component has not been initialised (k.config == nil).
+func TestPingUninitialised(t *testing.T) {
```
acroca, line 31: Missing a test to exercise a successful Ping.

common/component/kafka/health.go
```diff
@@ -0,0 +1,132 @@
...
+		if cfgCopy.Net.ReadTimeout > remaining || cfgCopy.Net.ReadTimeout == 0 {
+			cfgCopy.Net.ReadTimeout = remaining
+		}
+		if cfgCopy.Net.WriteTimeout > remaining || cfgCopy.Net.WriteTimeout == 0 {
+			cfgCopy.Net.WriteTimeout = remaining
+		}
+		// Also bound the metadata request itself and disable retries so the
+		// total time is governed by the caller's deadline alone.
+		cfgCopy.Metadata.Timeout = remaining
+		cfgCopy.Metadata.Retry.Max = 0
```
acroca: This is only set when the `ctx` has a deadline right? If it doesn't, it uses the default Sarama retry, which seems to be 3 retries, 30s each if I'm not mistaken. That could be problematic in some situations, no?

bindings/kafka/metadata.yaml
```diff
@@ -458,3 +458,39 @@ metadata:
       A regular expression to exclude keys from being converted to/from headers from/to metadata to avoid unwanted downstream side effects.
     example: '"^rawPayload|valueSchemaType$"'
     default: '""'
+  - name: dialTimeout
+    type: duration
+    description: |
+      The maximum duration to wait for the initial TCP connection to a broker to be established. Bounds how long Init() can block on an unreachable cluster (OSS-1152).
```
acroca:
```suggestion
      The maximum duration to wait for the initial TCP connection to a broker to be established. Bounds how long Init() can block on an unreachable cluster.
```

common/component/kafka/health_test.go
```diff
@@ -0,0 +1,160 @@
...
+	cfg := sarama.NewConfig()
+	cfg.Net.DialTimeout = 200 * time.Millisecond
+	cfg.Net.ReadTimeout = 200 * time.Millisecond
+	cfg.Net.WriteTimeout = 200 * time.Millisecond
+	cfg.Metadata.Retry.Max = 0
+
+	k := &Kafka{
+		logger:  logger.NewLogger("kafka_test"),
+		config:  cfg,
+		brokers: []string{"127.0.0.1:19999"}, // non-routable port
```
acroca: Shouldn't we verify this is really non-routable?

common/component/kafka/aws.go
```diff
@@ -98,16 +98,18 @@ func (m *mskTokenProvider) Token() (*sarama.AccessToken, error) {
...
 func (c *AwsClients) getSyncProducer() (sarama.SyncProducer, error) {
-	// Add SyncProducer specific properties to copy of base config
-	c.config.Producer.RequiredAcks = sarama.WaitForAll
-	c.config.Producer.Retry.Max = 5
-	c.config.Producer.Return.Successes = true
-
-	if *c.maxMessageBytes > 0 {
-		c.config.Producer.MaxMessageBytes = *c.maxMessageBytes
+	// Apply SyncProducer-specific properties to a COPY of the shared config so
+	// the consumer-group's view of the config is not mutated (OSS-1152 fix).
```
cicoyle:
```suggestion
	// the consumer-group's view of the config is not mutated.
```

common/component/kafka/metadata_test.go
```diff
@@ -717,3 +717,147 @@ func TestGetEventMetadata(t *testing.T) {
 		require.NotContains(t, act, "rawPayload")
 	})
 }
+
+// OSS-1152: Net timeout tunables
```
cicoyle:
```suggestion
// Net timeout tunables
```

common/component/kafka/producer.go
```diff
@@ -37,14 +37,27 @@ func parsePartitionNumber(value string) (int32, error) {
...
+type ProducerConfig struct {
+	RequiredAcks    sarama.RequiredAcks
+	RetryMax        int
+	MaxMessageBytes int
+}
+
+// GetSyncProducer creates a new Sarama SyncProducer using the provided base
+// config and producer tunables. RequiredAcks and RetryMax are applied from
+// ProducerConfig so callers can override the previously hard-coded defaults
+// (WaitForAll, 5 retries) via component metadata (OSS-1152).
```
cicoyle:
```suggestion
// (WaitForAll, 5 retries) via component metadata.
```

pubsub/kafka/metadata.yaml
```diff
@@ -374,6 +374,42 @@ metadata:
         The max amount of time for the client connection to be kept alive with the broker, as a Go duration, before closing the connection. A zero value (default) means keeping alive indefinitely.
       example: '4m'
       default: '0'
+    - name: dialTimeout
+      type: duration
+      description: |
+        The maximum duration to wait for the initial TCP connection to a broker to be established. Bounds how long Init() can block on an unreachable cluster (OSS-1152).
```
cicoyle:
```suggestion
        The maximum duration to wait for the initial TCP connection to a broker to be established. Bounds how long Init() can block on an unreachable cluster.
```

common/component/kafka/health_test.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	// Deadline in the past.
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(-time.Second))
+	defer cancel()
+
+	err := k.Ping(ctx)
+	require.Error(t, err)
+	require.ErrorIs(t, err, context.DeadlineExceeded)
+}
```
javier-aliaga, line 76: Not sure what you want to test here, if you create a context that is already expired then it will fail in the first line of the Ping method https://github.com/dapr/components-contrib/pull/4413/changes#diff-2ed27feecf9428468b4533c38124d8111139c5fe955d7b46c4d69539cc606f2cR41

If you want to test this https://github.com/dapr/components-contrib/pull/4413/changes#diff-2ed27feecf9428468b4533c38124d8111139c5fe955d7b46c4d69539cc606f2cR59 I do not think this test is exercising it

common/component/kafka/health.go
```diff
@@ -0,0 +1,146 @@
...
+
+	ch := make(chan result, 1)
+	go func() {
+		// sarama.NewClient with Metadata.Full=false does NOT issue any network
+		// request — it only registers the seed broker addresses. We must call
+		// RefreshMetadata() explicitly to force an actual TCP connection; called
+		// with no topics it refreshes cluster-wide metadata. This is the real
+		// connectivity probe; it respects cfgCopy.Metadata.Timeout and
+		// .Retry.Max set above.
+		cl, err := sarama.NewClient(k.brokers, &cfgCopy)
```
cicoyle, line 109: This is building a new client every time we call `Ping()`. Can this instead create one health client in `Init()` and reuse it here and just call the `RefreshMetadata()`, so we don't set up a new client each time?

cicoyle (changes requested):
nice job, just a few things from me.

common/component/kafka/health.go
nelson-parente: Good catch, the comment was wrong. RefreshMetadata() with no topics pulls metadata for all topics regardless of Metadata.Full, so I dropped the "no topic-level scan" wording. Reworded it to say it's a cluster-wide refresh and that Full=false only skips NewClient's own fetch. Fixed in 64b4b08c.

common/component/kafka/health.go
nelson-parente: Done.

common/component/kafka/health_test.go
nelson-parente: Done.

common/component/kafka/health_test.go
nelson-parente: Added TestPingSuccess, it uses a mock broker that answers the metadata request. 64b4b08c

common/component/kafka/health.go
nelson-parente: Good point. Moved the retry disable out of the deadline block so it always applies. That way the probe fails fast instead of falling back to Sarama's default 3 retries when there's no deadline. 64b4b08c

bindings/kafka/metadata.yaml
nelson-parente: Done.

common/component/kafka/health_test.go
nelson-parente: Switched it to bind a loopback port and then close it, so the connection is guaranteed to be refused instead of relying on a hardcoded port being free. 64b4b08c

common/component/kafka/aws.go
nelson-parente: Done.

common/component/kafka/metadata_test.go
nelson-parente: Done.

common/component/kafka/producer.go
nelson-parente: Done.

pubsub/kafka/metadata.yaml
nelson-parente: Done, dropped it here too (already removed it in 64b4b08c).

common/component/kafka/health_test.go
nelson-parente: You're right, with an already-expired deadline it just returns at the ctx.Err() check at the top and never reaches the clamping code. I reworded the test to say that's what it's covering and added an assert that it returns right away without dialing. ca41da83

common/component/kafka/health.go
nelson-parente: We don't store the raw sarama.Client, only the producer and consumer group, and some cases (consumer-only, mocks) don't create one. 
Also, RefreshMetadata() has no context, so a shared client can't respect the per-call timeout. Kept one client per ping for now, can reuse later if you prefer.

cicoyle (approved):

PR #4422, opened 2026-06-25
Add topicsPattern regex subscriptions to Pulsar pub/sub

### LINKED ISSUE #2566: Multi-topic subscriptions using regular expressions for the Apache Pulsar Pub/Sub Building Block
## Describe the feature

When a consumer subscribes to a Pulsar topic, by default it subscribes to one specific topic, such as persistent://public/default/my-topic. However, Pulsar consumers can simultaneously subscribe to multiple topics using two methods:

a) By explicitly defining a list of topics. This is the only method Dapr supports today.
b) On the basis of a [regular expression](https://en.wikipedia.org/wiki/Regular_expression)) (regex), for example, persistent://public/default/finance-.*

Source: [Apache Pulsar - Multi-topic subscriptions](https://pulsar.apache.org/docs/2.11.x/concepts-messaging/#multi-topic-subscriptions)

This feature request is for implementing method b). The benefit is apparent when a consumer needs to subscribe to a large number of topics that comply with a naming convention. Apache Pulsar automatically discovers new topics meeting the regular expression to maintain consumer subscriptions current. The Apache Pulsar community is working on [PIP-145](https://github.com/apache/pulsar/issues/14505) to further improve performance of the topic discovery protocol thus making the use of regular expression more attractive. 

One possible way to implement this feature request is by adding logic to the Dapr [Subscription](https://github.com/apache/pulsar/issues/14505) custom resource controller to handle regular expressions in the "topic" clause.

Note: Although this feature is requested specifically for Apache Pulsar, other Pub/Sub products supporting regular expressions like Apache Kafka, would benefit as well.

## Release Note

RELEASE NOTE:

jslatten:
# Description

Adds support for **regex topic-pattern subscriptions** to the Apache Pulsar pub/sub component.

Previously a Dapr subscription mapped to a single explicit Pulsar topic. This adds an opt-in `topicsPattern` metadata field that subscribes a consumer to all topics matching a regular expression within the configured tenant/namespace, via the pulsar-client-go `ConsumerOptions.TopicsPattern`. An optional `autoDiscoveryPeriod` field maps to `ConsumerOptions.AutoDiscoveryPeriod`, making Pulsar's auto-discovery of newly created matching topics configurable.

Key points:
- Both fields can be set at the component level (applies to all subscriptions) or overridden per subscription via subscription metadata.
- In pattern mode, each message is delivered tagged with the **concrete topic it arrived on** (shortened to the local name) rather than the regex, so runtime route matching and observability see the real topic name.
- Per-topic schema validation (`<topic>.avroschema` / `.jsonschema`) is skipped in pattern mode, since the matching topic set is not known in advance.
- **Fully backward-compatible**: with `topicsPattern` unset, behavior is byte-for-byte unchanged. No runtime or CRD changes are required.

Invalid configuration (malformed regex, negative `autoDiscoveryPeriod`) is rejected at `Init` (component) and `Subscribe` (per-subscription override).

## Issue reference

Fixes #2566

(See the [comment on #2566](https://github.com/dapr/components-contrib/issues/2566#issuecomment-4802644223) describing why this component-local, no-runtime-change approach addresses the earlier maintainer concern.)

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
    * [x] Created the dapr/docs PR: dapr/docs#5226

### DIFF
--- a/pubsub/pulsar/metadata.go
+++ b/pubsub/pulsar/metadata.go
@@ -45,6 +45,8 @@ type pulsarMetadata struct {
 	SubscriptionInitialPosition      string                    `mapstructure:"subscribeInitialPosition"`
 	ReplicateSubscriptionState       bool                      `mapstructure:"replicateSubscriptionState"`
 	SubscriptionMode                 string                    `mapstructure:"subscribeMode"`
+	TopicsPattern                    string                    `mapstructure:"topicsPattern"`
+	AutoDiscoveryPeriod              time.Duration             `mapstructure:"autoDiscoveryPeriod"`
 	Token                            string                    `mapstructure:"token"`
 	CompressionType                  string                    `mapstructure:"compressionType"`
 	CompressionLevel                 string                    `mapstructure:"compressionLevel"`
--- a/pubsub/pulsar/metadata.yaml
+++ b/pubsub/pulsar/metadata.yaml
@@ -308,8 +308,8 @@ metadata:
     type: duration
     description: |
       When topicsPattern is set, controls how often Pulsar re-evaluates the pattern to pick up newly
-      created (or removed) matching topics. A value of "0" lets the Pulsar client use its own default
-      interval. Has no effect unless topicsPattern is set. Can be overridden per subscription via
-      subscription metadata.
-    default: '"1m"'
+      created (or removed) matching topics. The component default is "0", which lets the Pulsar client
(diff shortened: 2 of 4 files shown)


jslatten:
Committed fixes suggested above.

cicoyle (approved):

PR #4423, opened 2026-06-29
feat(conversation): add iFlytek Spark conversation component

FenjuFu:
## Description

Adds an **iFlytek Spark** conversation component.

iFlytek Spark exposes an OpenAI-compatible chat-completions endpoint
(`https://spark-api-open.xf-yun.com/v1`), so this component is a thin
declarative provider built on the shared `langchaingokit` LLM and
`BuildOpenAIClientOptions`, mirroring the existing `deepseek` provider.

Metadata:
- `key` (required, secret) – the iFlytek Spark HTTP API password from the open-platform console
- `model` (optional) – defaults to `4.0Ultra`
- `endpoint` (optional) – defaults to `https://spark-api-open.xf-yun.com/v1`

## Issue reference

N/A – new optional component, no behavior change to existing providers.

## Checklist

- [x] Code compiles correctly (mirrors `conversation/deepseek`)
- [x] Created/updated component metadata (`metadata.yaml`)
- [ ] Runtime registration in `dapr/dapr` (follow-up PR after this merges and `go.mod` is bumped)
- [ ] Reference docs in `dapr/docs` (`setup-spark.md`, follow-up)

## Follow-ups

1. Register the component in `dapr/dapr` (`cmd/daprd/components/conversation_spark.go`) once this is merged and the `components-contrib` dependency is updated.
2. Add the `setup-spark.md` reference doc under `dapr/docs`.


FenjuFu:
Gentle ping @dapr/maintainers-components-contrib — this has been open for about three weeks. The component mirrors conversation/deepseek (OpenAI-compatible thin wrapper) and is ready for review; the workflows still need first-time-contributor approval to run. Happy to address any feedback or rebase if needed.

mikeee (commented):
Could you please elaborate on what this would add if it's using an openai compatible api would using that component not already accommodate? 

Without extensive tests this would add a lot of maintenance burden on upgrades/updates.

Keen to hear your thoughts on this.

https://docs.dapr.io/reference/components-reference/supported-conversation/openai/

FenjuFu:
Thanks @mikeee, that's a fair question — let me be upfront about what this does and doesn't add.

You're right that technically the `openai` component already accommodates Spark: its endpoint is OpenAI-compatible, so pointing the existing `openai` component at `https://spark-api-open.xf-yun.com/v1` with an arbitrary `model` works. This component doesn't add any protocol handling on top of that — it's the same thin `langchaingokit` wrapper as `deepseek`, differing only in the two default constants (`generalv3.5` / the Spark base URL).

So the case for it is exactly the case for the existing per-vendor conversation components (`deepseek`, `mistral`, `googleai`, `huggingface`, `ollama`), which are all the same shape:
- **Discoverability** — Spark shows up in the supported-conversation list and docs, rather than users having to know it's "openai with this specific endpoint + model."
- **Sensible defaults** — the endpoint and a working default model are baked in.

I don't want to overstate it though: it's convenience + consistency, not new capability. So I'll defer to your read on the tradeoff:

1. If you'd like to keep it in line with the existing vendor components, I'm happy to **add unit tests** (metadata + init) — I can mirror what `openai`/`anthropic` have. Worth noting `deepseek`/`mistral`/`googleai`/`ollama` currently ship without tests, so this would actually raise the bar rather than match the (untested) precedent.
2. If the direction is instead to steer people toward the generic `openai` component + custom endpoint and *not* grow more thin per-vendor wrappers, that's completely reasonable — I'm fine closing this, and I'd be glad to instead send a small docs PR to the openai conversation reference noting that OpenAI-compatible Chinese endpoints (Spark, etc.) work via the `endpoint` field, so the discoverability gap is still addressed.

Let me know which you'd prefer and I'll follow up accordingly.

mikeee:
@FenjuFu If we can add certification/conformance tests to ensure continued functionality then I'd supportive of adding this component. This would probably involve having a staging/test api key if you could possibly supply one privately that we can add to infra?

FenjuFu:
Thanks, that makes sense. I'm happy to add certification/conformance coverage following the repository's existing test framework so continued Spark compatibility is exercised. For credentials, I need to confirm internally that we can provide a dedicated staging/test key; I won't use or share a production credential in the repository. Could you point me to the preferred private channel and any secret naming, scope, or rotation requirements for the test infrastructure? Once I have those details and key availability is confirmed, I'll update this PR with the conformance test changes.

mikeee:
> Thanks, that makes sense. I'm happy to add certification/conformance coverage following the repository's existing test framework so continued Spark compatibility is exercised. For credentials, I need to confirm internally that we can provide a dedicated staging/test key; I won't use or share a production credential in the repository. Could you point me to the preferred private channel and any secret naming, scope, or rotation requirements for the test infrastructure? Once I have those details and key availability is confirmed, I'll update this PR with the conformance test changes.

That would be very helpful, of course we don't store secrets in the repo itself. At the moment it'll be a case that you'll need to load the secret from github secrets - I think an appropriate secret name is `${{ secrets.COMPONENT_CONVERSATION_IFLYTEK_APIKEY }}`

An example of what you'll need to add (both for certification and conformance): https://github.com/dapr/components-contrib/blob/19911701bae02b8816ec9f77b6972dfc08efa93d/.github/workflows/certification.yml#L214

Once you're ready with a secret please email me with the staging/test key.

FenjuFu:
Thanks @mikeee — I’ve pushed commit f400ee89 with certification and conformance coverage for the IFLYTEK Spark conversation component.

The update includes:
- a live certification smoke test that initializes the component and verifies a real conversation response;
- Spark coverage in the shared conversation conformance suite, including message roles, tool calling, and conversation history;
- cloud-test matrix integration for both certification and conformance;
- loading the credential from the COMPONENT_CONVERSATION_IFLYTEK_APIKEY repository secret in both workflows;
- 4.0Ultra as the default model, matching the current Ultra 32K HTTP API.

I ran both suites against the real staging/test API:
- certification: passed;
- conformance: passed for the supported conversation capabilities.

The shared structured-output test sends OpenAI’s json_schema response format, while Spark’s HTTP API currently documents text and json_object; Spark returns HTTP 400 for json_schema, so that optional case is explicitly skipped with the precise capability reason. The remaining live conformance cases pass.

The dedicated non-production test credential is ready. I’ll send it separately by email so it can be added to the repository secret without exposing it in the PR.

conversation/iflytek/spark/spark.go
```diff
@@ -0,0 +1,87 @@
...
+
+type Spark struct {
+	langchaingokit.LLM
+	md SparkMetadata
+
+	logger logger.Logger
+}
+
+const (
+	defaultModel    = "4.0Ultra"
```
mikeee, line 39: Out of curiosity - how often does the default model change? I can see it has been updated already

FenjuFu: It should not change frequently. I initially used `generalv3.5` because it is still used in several examples in the HTTP API documentation, but that identifier maps to Spark Max.

iFlytek has since announced that the Max offering was retired on March 10, 2026, with its backend and quota migrated to Ultra. I therefore updated the default to `4.0Ultra`, which is the currently supported Ultra model and also supports the Function Call coverage exercised by the conformance tests.

The model remains configurable through the `model` metadata field, so users can override the default. I would expect us to change the default only when iFlytek deprecates or replaces the upstream model, rather than on a regular basis.

FenjuFu:
Thanks, Mike. I have sent the dedicated non-production staging/test API key to you by email. The PR now includes both certification and conformance coverage, and the workflows read the credential from `${{ secrets.COMPONENT_CONVERSATION_IFLYTEK_APIKEY }}` as requested. Please let me know if you need any additional configuration details.

tests/conformance/conversation/conversation.go
```diff
@@ -548,6 +548,10 @@ func ConformanceTests(t *testing.T, props map[string]string, conv conversation.C
 		})
 
 		t.Run("test response format returned", func(t *testing.T) {
+			if component == "spark" {
+				t.Skip("IFLYTEK Spark does not support the OpenAI json_schema response format")
+			}
+
```
mikeee: Could you please add this skip to the assertions below on L555

conversation/spark/metadata.yaml
```diff
@@ -0,0 +1,45 @@
+# yaml-language-server: $schema=../../../component-metadata-schema.json
+schemaVersion: v1
+type: conversation
+name: spark
```
mikeee: Could you please rename this component to 
```suggestion
name: iflytek.spark
```
This would also involve moving this into a subfolder of iflytek e.g. `iflytek/spark/`

mikeee (changes requested):
Could you also please update the copyright headers to 2026 along with double check if I'm right a few points mentioned

tests/conformance/conversation/conversation.go
FenjuFu: Addressed in 9d6737d5: I folded iflytek.spark into the existing provider skip branch, so the structured-output test now uses the shared assertion/skip path instead of a separate early skip.

conversation/spark/metadata.yaml
FenjuFu: Addressed in 9d6737d5: the component is now named iflytek.spark and moved to conversation/iflytek/spark, with certification tests, conformance config, imports, matrix source package, and documentation updated consistently. The new Go file headers are also updated to 2026.

FenjuFu:
Addressed the requested changes in commit 9d6737d5 and merged current upstream/main in f94470b1 without rewriting the branch:

- renamed the component to iflytek.spark and moved it under conversation/iflytek/spark;
- updated certification/conformance paths, imports, config, matrix source package, and docs;
- folded the Spark structured-output skip into the existing provider skip/assertion branch;
- updated the new Go source headers to 2026;
- preserved upstream's new generic authenticated matrix field while including the IFLYTEK credential flag.

Local verification:
- go test ./conversation/iflytek/spark: passed
- certification package compile from tests/certification: passed
- certification + conformance matrix generation includes conversation.iflytek.spark, authenticated: true, and require-iflytek-credentials: true
- metadata bundler loads conversation/iflytek/spark/metadata.yaml
- targeted conformance passed before the upstream merge; after the merge, the Windows host spent the full local timeout in the tagged dependency/compile phase without emitting a test failure, so the refreshed GitHub checks will be the authoritative post-merge run.

Both inline threads have been replied to and resolved. Please re-review when convenient.

FenjuFu:
Thanks @mikeee — addressed and double-checked both points.

**Copyright headers → 2026:** the three files this PR *adds* all carry 2026 headers:
- `conversation/iflytek/spark/spark.go`
- `conversation/iflytek/spark/metadata.go`
- `tests/certification/conversation/iflytek/spark/spark_test.go`

The two files this PR only *modifies* keep their original year — `tests/conformance/conversation/conversation.go` (2024, +1/-1) and `tests/conformance/conversation_test.go` (2021, +8/-0) — matching this repo's convention that the header reflects file creation rather than edit year (e.g. `conversation.go` has stayed 2024 across its edits). Happy to bump those two to 2026 as well if you'd prefer that for this PR — just say the word and I'll push it.

**Points you raised:**
- Component renamed to `iflytek.spark` and moved under `conversation/iflytek/spark/` (metadata `name: iflytek.spark`).
- The structured-output skip is folded into the shared assertion/skip branch in `tests/conformance/conversation/conversation.go` (`component == iflytek.spark` alongside echo/ollama/bedrock), not a separate early skip.

Re-requesting review — let me know on the two modified-file headers and I'll follow up right away.

FenjuFu:
@mikeee thanks for the review — all points are addressed in the current head:

- **Renamed to `iflytek.spark` and moved under `conversation/iflytek/spark/`** (`9d6737d5`). Component name, certification package, conformance config, imports, and the metadata matrix source were all updated consistently.
- **Conformance skip** folded `iflytek.spark` into the existing shared provider skip branch on the structured-output assertion, rather than a separate early skip.
- **Copyright headers** on the new Go files are set to `2026`.
- **Default model** stays `4.0Ultra` (Spark Max was retired 2026-03-10 with quota migrated to Ultra), which shouldn't change frequently.

Branch is also merged up to `main` and the go.mod fix (`47ca2f00`) promotes `langchaingo` to a direct require in the certification module so `check-mod-diff` stays green.

The only red check is the **certification** job, which fails with `IFLYTEK_API_KEY must be set for certification tests`. That's the expected missing-secret state for a brand-new provider — the test itself is correct, it just needs `IFLYTEK_API_KEY` provisioned as a repo/org Actions secret to run against the live endpoint (`https://spark-api-open.xf-yun.com/v1`). Happy to help however is easiest here: I can provide a key through whatever secret-handling process you prefer, or gate the cert test to skip cleanly when the secret is absent if you'd rather not wire one up yet. Let me know which you'd prefer.

Re-requesting review when you have a moment — thanks!

mikeee:
Looks good to merge but the API key has expired or has no budget.

FenjuFu:
@mikeee thanks for running it against the live API, and sorry for the slow reply. About the key: I'm getting a replacement with budget on the iFlytek side and will post here once it's ready. Please tell me how you'd like to receive it.

While waiting, I pushed f151111b to address the Copilot review:

- `Init` returns `spark api key is required` when `key` is empty. Before, users got langchaingo's `OPENAI_API_KEY` message.
- `responseCacheTTL` / `cacheTTL` now goes through `conversation.CacheResponses`, like the openai, mistral and ollama components do. `metadata.yaml` documents `cacheTTL` the same way as `openai`.
- `SparkMetadata` only embeds `LangchainMetadata` now. The duplicate `Key` field and the unused `MaxTokens` field are gone, so `GetComponentMetadata` no longer returns uppercase `Key` / `MaxTokens`. `maxTokens` is also removed from `metadata.yaml`.
- The default model is `4.0Ultra` on purpose. I fixed the PR description, which still said `generalv3.5`.
- New unit tests in `conversation/iflytek/spark/spark_test.go` cover the missing key, the defaults, the cache option and the metadata keys. `go test ./conversation/iflytek/spark/` passes locally.

PR #4426, opened 2026-06-29
docs(rabbitmq): list all supported exchangeKind values in metadata

MyMirelHub:
## Description

`exchangeKindValid` in `pubsub/rabbitmq/metadata.go` accepts four exchange kinds — `fanout`, `topic`, `direct`, and `headers` — but `pubsub/rabbitmq/metadata.yaml` only documented `fanout` and `topic`. This brings the metadata (which drives generated reference docs) in line with the actual validation logic by adding `direct` and `headers` to `allowedValues` and the example.

No behavior change — documentation/metadata only.

### DIFF
--- a/pubsub/rabbitmq/metadata.yaml
+++ b/pubsub/rabbitmq/metadata.yaml
@@ -145,7 +145,9 @@ metadata:
     allowedValues:
       - "fanout"
       - "topic"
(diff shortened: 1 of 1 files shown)


MyMirelHub:
Closed to fix branch name.

Issue #4281, opened 2026-03-13
RabbitMQ: `handleMessage` NACKs messages on `context.Canceled` during graceful shutdown, routing them to the dead-letter queue


kayvonkhosrowpour:
# RabbitMQ: `handleMessage` NACKs messages on `context.Canceled` during graceful shutdown, routing them to the dead-letter queue

## In what area(s)?

/area pubsub

## What component?

`pubsub.rabbitmq`

## What version of components-contrib / Dapr?

- components-contrib: v1.16.10 (and current `main`)
- Dapr runtime: v1.16.10 (also reproducible with dapr/dapr PR #9619 applied)

## Related Issues

- dapr/dapr#9604 — Pub/Sub: Messages NACKed to dead-letter queue during graceful shutdown due to premature subscription close
- dapr/dapr#9619 — fix: Prevent pub/sub messages from being NACKed during graceful shutdown

## Expected Behavior

When the Dapr runtime is shutting down and the subscription handler returns `context.Canceled` (because the subscription context was cancelled), the RabbitMQ component should **not** NACK the message. Instead, it should leave the message unacknowledged so that when the AMQP connection closes, RabbitMQ redelivers the message to another consumer.

## Actual Behavior

`handleMessage` in `pubsub/rabbitmq/rabbitmq.go` unconditionally NACKs any non-nil error returned by the handler — including `context.Canceled`. With `enableDeadLetter: true` and `requeueInFailure: false` (the default), this routes the message to the dead-letter queue.

This means that even after dapr/dapr#9619 fixes the runtime to block on `ctx.Done()` instead of returning `"subscription is closed"`, one message still gets NACKed to the DLQ during the narrow window when the context is cancelled and the AMQP connection hasn't yet closed.

### Relevant log from daprd

See reproducible small example here: https://github.com/kayvonkhosrowpour/dapr-repro/tree/components-contrib-bug

```
level=error msg="rabbitmq pub/sub error: handling message from topic 'incoming-events', context canceled"
```

Note: the error is `context canceled` (not `subscription is closed`), confirming the runtime fix from #9619 is working. But `handleMessage` still NACKs it.

## Root Cause

In [`pubsub/rabbitmq/rabbitmq.go` — `handleMessage`](https://github.com/dapr/components-contrib/blob/main/pubsub/rabbitmq/rabbitmq.go#L624-L656):

```go
[...]

PR #4450, opened 2026-07-13
fix(pubsub/rabbitmq): skip NACK on context cancellation in handleMessage

### LINKED ISSUE #4449: RabbitMQ: `handleMessage` still NACKs messages on `context.Canceled` during graceful shutdown in v1.18.1
# RabbitMQ: `handleMessage` still NACKs messages on `context.Canceled` during graceful shutdown in v1.18.1

## In what area(s)?

/area pubsub

## What component?

`pubsub.rabbitmq`

## What version of components-contrib / Dapr?

- components-contrib: **v1.18.1** (latest release)
- Dapr runtime: **v1.18.1** (latest release)

## History

This was previously filed as **[#4281](https://github.com/dapr/components-contrib/issues/4281)**
(March 2026) against v1.16.10, and was **closed as stale without being fixed**. The bug
persists in the latest v1.18.1 release. This is a re-filing with updated version confirmation
and a ready-to-merge fix.

## Related Issues

- [dapr/dapr#9604](https://github.com/dapr/dapr/issues/9604) — Original runtime-side bug report (closed/fixed)
- [dapr/dapr#9619](https://github.com/dapr/dapr/pull/9619) — Runtime fix (merged into v1.14+)
- [dapr/components-contrib#4281](https://github.com/dapr/components-contrib/issues/4281) — Previous filing (closed as stale, unfixed)

## Expected Behavior

When the Dapr runtime is shutting down and the subscription handler returns `context.Canceled`
(because the subscription context was cancelled), the RabbitMQ component should **not** NACK
the message. Instead, it should leave the message unacknowledged so that when the AMQP
connection closes, RabbitMQ redelivers the message to another consumer.

## Actual Behavior

`handleMessage` in `pubsub/rabbitmq/rabbitmq.go` unconditionally NACKs any non-nil error
returned by the handler — including `context.Canceled`. With `enableDeadLetter: true` and
`requeueInFailure: false` (the default), this routes the message to the dead-letter queue.

The dapr runtime fix from #9619 is working correctly — the error changed from
`"subscription is closed"` to `"context canceled"` — but `handleMessage` still NACKs it:

```
level=error msg="rabbitmq pub/sub error: handling message from topic 'incoming-events', context canceled"
```

This means **one message per graceful shutdown is unconditionally routed to the DLQ** when
using RabbitMQ with `enableDeadLetter: true`, even on the latest Dapr v1.18.1.

## Root Cause

In [`pubsub/rabbitmq/rabbitmq.go` — `handleMessage`](https://github.com/dapr/components-contrib/blob/main/pubsub/rabbitmq/rabbitmq.go):

```go
err := handler(ctx, pubsubMsg)

if err != nil {
    r.logger.Errorf("%s handling message from topic '%s', %s", errorMessagePrefix, topic, err)

    if !r.metadata.AutoAck {
        // NACKs unconditionally — including when err == context.Canceled
        if err = d.Nack(false, r.metadata.RequeueInFailure); err != nil { ... }
    }
}
```

The `if err != nil` block has no check for context cancellation. When the subscription context
is cancelled during graceful shutdown, `handler` returns `context.Canceled`, and the message
is NACKed with `requeue=false`, routing it to the dead-letter queue.

## Proposed Fix

Add a `ctx.Err()` guard before the NACK logic. If the context is done, skip both ACK and NACK —
the message stays unacknowledged, and RabbitMQ redelivers it to another consumer when the
connection closes:

```go
err := handler(ctx, pubsubMsg)

if err != nil {
    if ctx.Err() != nil {
        r.logger.Debugf("%s context done while handling message from topic '%s'; skipping ack/nack to allow redelivery", logMessagePrefix, topic)
        return err
    }

    r.logger.Errorf("%s handling message from topic '%s', %s", errorMessagePrefix, topic, err)

    if !r.metadata.AutoAck {
        r.logger.Debugf("%s nacking message '%s' from topic '%s', requeue=%t", logMessagePrefix, d.MessageId, topic, r.metadata.RequeueInFailure)
        if err = d.Nack(false, r.metadata.RequeueInFailure); err != nil {
            r.logger.Errorf("%s error nacking message '%s' from topic '%s', %s", logMessagePrefix, d.MessageId, topic, err)
        }
    }
} else if !r.metadata.AutoAck {
    // ack path unchanged
}
```

This is a **5-line change** with no new dependencies. I'm happy to submit a PR with this fix
and a unit test.

## Steps to Reproduce

Full reproduction repo (updated for v1.18.1):
**https://github.com/kayvonkhosrowpour/dapr-repro** (branch `components-contrib-bug`)

One-command setup:

```bash
minikube start --kubernetes-version=v1.33.0
skaffold run  # deploys Dapr v1.18.1, RabbitMQ, demo app
```

Then:

```bash
# Tail logs
stern -n dapr-repro dapr-repro

# Publish 8 events to create a backlog
for i in $(seq 1 8); do
  kubectl exec -n dapr-repro deploy/dapr-repro -c dapr-repro -- \
    wget -qO- --post-data="{\"id\":\"$i\"}" \
    --header='Content-Type: application/json' \
    http://localhost:3500/v1.0/publish/pubsub/incoming-events
done

# Wait for tick logs (first event processing), then delete the pod
kubectl delete pod -n dapr-repro -l app=dapr-repro
```

**Observe:**

| Scenario | DLQ messages | daprd error |
|---|---|---|
| Dapr v1.18.1 (stock) | **1** | `"context canceled"` |
| Dapr v1.18.1 + proposed fix | **0** | none |

Check the RabbitMQ management UI at `localhost:15672` (credentials `rabbit`/`rabbit`) —
`dlq-dapr-repro-incoming-events` queue will have 1 message after shutdown.

## Release Note

RELEASE NOTE: **FIX** RabbitMQ pub/sub `handleMessage` now skips NACK when the context is
cancelled (e.g. during graceful shutdown), leaving the message unacknowledged for redelivery
by the broker instead of routing it to the dead-letter queue.

kayvonkhosrowpour:
## Description

Fixes #4449 (re-filing of #4281, closed as stale).

When the subscription handler returns an error because its context was cancelled
(e.g. during Dapr graceful shutdown), `handleMessage` was unconditionally calling
`d.Nack()`. With `enableDeadLetter: true` and `requeueInFailure: false` (the
defaults), this routes the message to the dead-letter queue instead of letting
RabbitMQ redeliver it to another consumer when the AMQP connection closes.

### Root Cause

```go
// before — no context check before NACK
if err != nil {
    r.logger.Errorf(...)
    if !r.metadata.AutoAck {
        d.Nack(false, r.metadata.RequeueInFailure)  // fires on context.Canceled too
    }
}
```

### Fix

Add a `ctx.Err()` guard that returns early (without Ack or Nack) when the context
is done. The message stays unacknowledged and RabbitMQ redelivers it on connection
close — zero messages to DLQ during graceful shutdown.

### Relation to dapr/dapr#9619

The runtime fix in dapr/dapr#9619 changed the error from `"subscription is closed"`
to `"context canceled"`. This PR closes the remaining component-level gap.

### Reproduction

https://github.com/kayvonkhosrowpour/dapr-repro (branch `components-contrib-bug`) — updated for Dapr v1.18.1.

| Scenario | DLQ messages | daprd log |
|---|---|---|
| v1.18.1 without this fix | **1** | `context canceled` |
| v1.18.1 with this fix | **0** | (none) |

## Type of change

- [x] Bug fix

## Checklist

- [x] Tests updated — `TestHandleMessageSkipsNACKOnContextCanceled` with 3 subtests
- [x] All existing unit tests pass

### DIFF
--- a/pubsub/rabbitmq/rabbitmq.go
+++ b/pubsub/rabbitmq/rabbitmq.go
@@ -638,7 +638,9 @@ func (r *rabbitMQ) listenMessages(ctx context.Context, channel rabbitMQChannelBr
 				go func(d amqp.Delivery) {
 					defer r.wg.Done()
 					if err := r.handleMessage(ctx, d, topic, handler); err != nil {
-						r.logger.Errorf("%s error handling message: %v", logMessagePrefix, err)
+						if !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
+							r.logger.Errorf("%s error handling message: %v", logMessagePrefix, err)
+						}
 					}
 				}(d)
 			}
--- a/pubsub/rabbitmq/rabbitmq.go
+++ b/pubsub/rabbitmq/rabbitmq.go
@@ -660,6 +660,11 @@ func (r *rabbitMQ) handleMessage(ctx context.Context, d amqp.Delivery, topic str
 	err := handler(ctx, pubsubMsg)
 
 	if err != nil {
+		if ctx.Err() != nil {
+			r.logger.Debugf("%s context done while handling message from topic '%s'; skipping ack/nack to allow redelivery", logMessagePrefix, topic)
+			return err
+		}
+
 		r.logger.Errorf("%s handling message from topic '%s', %s", errorMessagePrefix, topic, err)
 
 		if !r.metadata.AutoAck {
--- a/pubsub/rabbitmq/rabbitmq_test.go
+++ b/pubsub/rabbitmq/rabbitmq_test.go
@@ -508,9 +508,9 @@ func TestHandleMessageSkipsNACKOnContextCanceled(t *testing.T) {
 		metadata: &rabbitmqMetadata{AutoAck: false},
 	}
 
-	t.Run("context canceled: no NACK and no ACK", func(t *testing.T) {
+	t.Run("context canceled: no NACK, no ACK, context error returned", func(t *testing.T) {
 		ctx, cancel := context.WithCancel(context.Background())
-		cancel() // already cancelled — simulates subscription context cancelled during shutdown
+		cancel() // already cancelled; simulates subscription context cancelled during shutdown
 
 		ack := &mockAcknowledger{}
 		d := amqp.Delivery{Acknowledger: ack, Body: []byte("test payload")}
@@ -520,8 +520,8 @@ func TestHandleMessageSkipsNACKOnContextCanceled(t *testing.T) {
 		}
 
 		err := r.handleMessage(ctx, d, "topic", handler)
-		require.Error(t, err)
-		assert.False(t, ack.nackCalled.Load(), "NACK must not be called when context is canceled — message should be redelivered, not DLQ'd")
+		require.ErrorIs(t, err, context.Canceled)
+		assert.False(t, ack.nackCalled.Load(), "NACK must not be called when context is canceled; message should be redelivered, not DLQ'd")
 		assert.False(t, ack.ackCalled.Load(), "ACK must not be called when context is canceled")
 	})
 
--- a/pubsub/rabbitmq/rabbitmq_test.go
+++ b/pubsub/rabbitmq/rabbitmq_test.go
@@ -475,6 +475,86 @@ func TestSubscribeReconnect(t *testing.T) {
 	assert.Equal(t, int32(4), broker.closeCount.Load())   // two counts for each connection closure - one for connection, one for channel
 }
 
+// mockAcknowledger tracks Ack/Nack calls on an amqp.Delivery for unit tests.
+type mockAcknowledger struct {
+	nackCalled atomic.Bool
+	ackCalled  atomic.Bool
+}
+
+func (m *mockAcknowledger) Ack(_ uint64, _ bool) error {
+	m.ackCalled.Store(true)
+	return nil
+}
+
+func (m *mockAcknowledger) Nack(_ uint64, _ bool, _ bool) error {
+	m.nackCalled.Store(true)
+	return nil
+}
+
+func (m *mockAcknowledger) Reject(_ uint64, _ bool) error {
+	return nil
+}
+
+// TestHandleMessageSkipsNACKOnContextCanceled verifies that handleMessage does
+// not NACK the delivery when the handler returns because the context was
+// cancelled (e.g. during graceful shutdown). Leaving the message unacknowledged
+// lets RabbitMQ redeliver it to another consumer when the connection closes,
+// rather than routing it to the dead-letter queue.
+//
+// Regression test for https://github.com/dapr/components-contrib/issues/4449
+func TestHandleMessageSkipsNACKOnContextCanceled(t *testing.T) {
+	r := &rabbitMQ{
+		logger:   logger.NewLogger("test"),
+		metadata: &rabbitmqMetadata{AutoAck: false},
+	}
+
+	t.Run("context canceled: no NACK and no ACK", func(t *testing.T) {
+		ctx, cancel := context.WithCancel(context.Background())
+		cancel() // already cancelled — simulates subscription context cancelled during shutdown
+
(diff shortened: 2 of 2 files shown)


mikeee (commented):
Just one Q: in parallel mode - the cancellation is still logged as an error:

https://github.com/dapr/components-contrib/pull/4450/changes#diff-6d69b213107781018930a58cb30f686338377af29091816676b2e22986bb84acL640-L641

Is this something you think should be addressed with this PR too?

pubsub/rabbitmq/rabbitmq.go
mikeee: Error still logged here

kayvonkhosrowpour:
**Checklist update:** Docs PR is now open at dapr/docs#5245.

**Summary of this PR:**
- Fixes the `handleMessage` NACK-on-`context.Canceled` bug (re-filing of #4281, which was closed as stale)
- **Confirmed live** against Dapr v1.18.1: stock build → 1 msg in DLQ; with this fix → 0 msgs. Reproduction: https://github.com/kayvonkhosrowpour/dapr-repro (branch `components-contrib-bug`)
- 3 unit tests added covering all three ack/nack paths
- DCO signed

cc @cicoyle — you merged the related Kafka graceful-shutdown PR (#4366). This is the same class of fix for RabbitMQ.

pubsub/rabbitmq/rabbitmq.go
kayvonkhosrowpour: Good catch — fixed. Changed `return err` to `return nil` so the caller (the parallel goroutine at this line) doesn't log a spurious error-level message during normal graceful shutdown. Updated the unit test subtest name and assertion accordingly (`require.NoError` instead of `require.Error`). Force-pushed.

mikeee: Is this change made to L665 - I'm not seeing it updated

mikeee (changes requested):
Need to address the parallel mode handler error logged

pubsub/rabbitmq/rabbitmq.go
kayvonkhosrowpour: My bad, it wasn't pushed... At `rabbitmq.go` L662-L665, `handleMessage` now returns `nil` when `ctx.Err() != nil`, so the parallel caller at L640-L641 won’t log the error during graceful shutdown. Let me know if this works. I think we want to keep the error log if it's an actual error. But since we return `nil` it shouldn't be an issue.

mikeee (approved):

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -660,6 +660,11 @@ func (r *rabbitMQ) handleMessage(ctx context.Context, d amqp.Delivery, topic str
 	err := handler(ctx, pubsubMsg)
 
 	if err != nil {
+		if ctx.Err() != nil {
+			r.logger.Debugf("%s context done while handling message from topic '%s'; skipping ack/nack to allow redelivery", logMessagePrefix, topic)
+			return nil
```
mikeee: Sorry for the confusion - I think returning a nil error when there is an error or ctx error isn't strictly correct. Perhaps updating the parallel mode assertion on L640 to exclude context.Canceled and context.DeadlineExceeded errors may be more idiomatic. Wdyt?

mikeee (changes requested):

pubsub/rabbitmq/rabbitmq.go
kayvonkhosrowpour: Fair comment, I agree. I changed it to log the error only for non `context.Canceled` and non `context.DeadlineExceeded` errors. Let me know if anything else is needed from me!

mikeee (approved):

mikeee:
@kayvonkhosrowpour thanks for addressing this!

marcduiker:
@holopin-bot @kayvonkhosrowpour components-badge Thank you! Here's a digital badge as a small token of appreciation.

PR #4461, opened 2026-07-21
fix(configuration/azure/appconfig): use retryDelay for initial retry delay

### LINKED ISSUE #4303: Azure App Configuration component ignores retryDelay when configuring Azure SDK retry options
## Expected Behavior

The Azure App Configuration component should correctly apply both `retryDelay` and `maxRetryDelay` from metadata when configuring the Azure SDK retry options.

- `retryDelay` should be used as the initial retry delay  
- `maxRetryDelay` should be used as the maximum retry backoff  

## Actual Behavior

Although both `retryDelay` and `maxRetryDelay` are parsed correctly from component metadata, the current implementation uses `maxRetryDelay` for both fields when constructing the Azure SDK retry options.

[configuration/azure/appconfig/appconfig.go](https://github.com/dapr/components-contrib/blob/fd97a1c1398468e8aacebdef0ce1d39400e055cc/configuration/azure/appconfig/appconfig.go#L97)
Image

1. Configure the Azure App Configuration component with retry settings:

```yaml
metadata:
  - name: maxRetries
    value: "3"
  - name: retryDelay
    value: "4s"
  - name: maxRetryDelay
    value: "120s"
```
2. Initialize the component.

3. Observe that:
   - the Azure SDK retry configuration does not use the configured `retryDelay`, and the initial retry delay is ignored.

## Release Note

RELEASE NOTE: **FIX** Azure App Configuration component ignores retryDelay when configuring retry options

pujitha24:
Motivation:
The Azure App Configuration component parsed both `retryDelay` and
`maxRetryDelay` metadata properties correctly, but Init() used
`maxRetryDelay` for both the `RetryDelay` and `MaxRetryDelay` fields
when building the Azure SDK's policy.RetryOptions. As a result, a
user-configured `retryDelay` value was silently ignored: the SDK used
`maxRetryDelay` as the initial per-attempt backoff delay instead. This
is a silent misconfiguration bug, not a crash or outage — retries still
happen, just with the wrong initial delay.

Approach:
Extract the retry-options construction into a small `newRetryOptions`
helper so the mapping is independently unit-testable, and fix the
mapping so `RetryDelay` comes from `metadata.RetryDelay` instead of
`metadata.MaxRetryDelay`. Added TestNewRetryOptions, which uses
distinct RetryDelay/MaxRetryDelay values and would have failed against
the previous code (both fields would have equaled MaxRetryDelay).

Validation:
go test ./configuration/azure/appconfig/... -v
All tests pass, including the new TestNewRetryOptions and the existing
TestInit/TestParseMetadata suites.

Fixes #4303

Signed-off-by: Pujitha Paladugu <venkat@dema.shop>
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>


mikeee (approved):
lgtm

marcduiker:
@holopin-bot @pujitha24 components-badge Thank you! Here's a digital badge as a small token of appreciation.

PR #4465, opened 2026-07-27
feat(bindings/localstorage): add fromTemp metadata for cross-platform temp dir

### LINKED ISSUE #3158: Cross-platform temp directory for local storage binding

pujitha24:
Motivation:
The localstorage output binding requires rootPath to be set explicitly,
and there is no cross-platform way to point it at the OS temp directory.
Users have to hardcode OS-specific paths (e.g. "/tmp" on Unix,
"C:\Windows\Temp" or "%LocalAppData%\Temp" on Windows), which breaks when
a component definition is shared across development environments running
different operating systems.

Approach:
Add an optional `fromTemp` boolean metadata field. When true, rootPath is
resolved relative to os.TempDir() before the existing path validation
(disallowed-location checks, symlink resolution, etc.) runs unchanged.
When fromTemp is unset (the default, false), behavior is identical to
before this change. This is the same approach the issue author proposed
and that a maintainer previously reacted to positively on a stale, closed
PR for this issue (#4280); this change reimplements it from scratch since
that PR was auto-closed for inactivity with no code merged.

Validation:
go build ./bindings/localstorage/...
go test ./bindings/localstorage/... -v
All tests pass, including the new TestParseMetadataFromTemp case.

Fixes #3158

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

### DIFF
--- a/bindings/localstorage/localstorage.go
+++ b/bindings/localstorage/localstorage.go
@@ -57,6 +57,7 @@ type LocalStorage struct {
 // Metadata defines the metadata.
 type Metadata struct {
 	RootPath string `json:"rootPath"`
+	FromTemp bool   `json:"fromTemp"`
 }
 
 type createResponse struct {
@@ -89,6 +90,10 @@ func (ls *LocalStorage) parseMetadata(meta bindings.Metadata) (*Metadata, error)
 	err := kitmd.DecodeMetadata(meta.Properties, &m)
 	if err != nil {
 		return nil, err
+	}
+
+	if m.FromTemp {
+		m.RootPath = filepath.Join(os.TempDir(), m.RootPath)
 	}
 
 	m.RootPath, err = validateRootPath(m.RootPath)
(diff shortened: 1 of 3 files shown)


pujitha24:
Just checking in on this one — it's rebased and all checks are green, happy to make any changes that would help move review along.

pujitha24:
Closing this to keep my open PR queue manageable - it has been open a while without review, and I would rather not leave stale PRs sitting in your queue. The change itself still applies; happy to reopen and rebase if it is useful to you.

PR #4478, opened 2026-07-28
Remove the workflows component abstraction

JoshVanL:
The workflows package only ever defined an interface and its request/response types: no concrete component implementation ever existed in this repository. The single implementation lives inside dapr/dapr, which now uses the durabletask-go client directly, so the abstraction serves no purpose and makes client changes require updates to two repositories.

Deletes the workflows package, its conformance test harness and config, and the workflows entries in the metadata utils, the metadata analyzer, the component metadata schema (regenerated), and the component folders list.


javier-aliaga (dismissed):

javier-aliaga (dismissed):

mikeee:
@JoshVanL conflicts

mikeee (approved):

PR #4485, opened 2026-07-30
Added sql server binding

### LINKED ISSUE #3414: [Bindings] Azure SQL DB / SQL Server Binding
## Describe the proposal

Customers who heavily leverage SQL Server installations in their existing on-premises data centres and are also moving to the cloud and adopting Azure SQL DB during the move to the cloud. Furthermore, these customers likely have relatively complex data structures, tables, views, and a host of stored procedures to make their systems work. 

Providing a Dapr binding for Azure SQL DB & SQL Server that looks the same as the [Microsoft SQL Server & Azure SQL state management component](https://docs.dapr.io/reference/components-reference/supported-state-stores/setup-sqlserver/#component-format) YAML component but gives them the ability to interact with their existing databases through the abstraction of the Dapr API will help progress Dapr's adoption for these customers.

The YAML would likely look as follows:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: <NAME>
spec:
  type: bindings.sqlserver
  version: v1
  metadata:
    # Authenticate using SQL Server credentials
    - name: connectionString
      value: Server=myServerName\myInstanceName;Database=myDataBase;User Id=myUsername;Password=myPassword;
    # Authenticate with Microsoft Entra ID (Azure SQL only)
    # "useAzureAD" be set to "true"
    - name: useAzureAD
      value: true
    # Connection string or URL of the Azure SQL database, optionally containing the database
    - name: connectionString
      value: sqlserver://myServerName.database.windows.net:1433?database=myDataBase
```

Given the intended use case, I'm not sure focusing on a specific table or indexing policy makes sense to include in the binding YAML. That would be for the database developer to configure on their database.

This would ultimately give developers the ability to (re)build their systems using Dapr without having to change their state stores and lowering the barrier of entry into Dapr for those who heavily leverage SQL Server / Azure SQL DB.

adam6878:
# Description

Adds a new bindings.sqlserver output binding for Microsoft SQL Server and Azure SQL

Dapr already has a state.sqlserver component, but it's a key-value state store: it owns its own schema (creates a fixed key/value/etag table) and isn't usable against a database that already has its own tables, views, and stored procedures. Users migrating an existing SQL Server database to Dapr need a way to run their own SQL/stored procedures through the sidecar — that's what an output binding is for, and no SQL Server one existed.

## Issue reference:

Please reference the issue this PR will close: #3414

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
    * [x] Created the dapr/docs PR: <insert PR link here>


bindings/sqlserver/sqlserver.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+
+	// Metadata property "params" contains JSON-encoded parameters, and it's optional
+	// If present, it must be unserializable into a []any object
+	var (
+		params []any
+		err    error
+	)
+	if paramsStr := req.Metadata[commandParamsKey]; paramsStr != "" {
+		err = json.Unmarshal([]byte(paramsStr), &params)
```
mikeee: I think the same issue exists as this looks like the code was derived from the postgres/mysql bindings - we're not handling numbers correctly here. All numbers are being converted to float64

bindings/sqlserver/sqlserver.go
```diff
@@ -0,0 +1,293 @@
+/*
+Copyright 2024 The Dapr Authors
```
mikeee:
```suggestion
Copyright 2026 The Dapr Authors
```

bindings/sqlserver/metadata.go
```diff
@@ -0,0 +1,53 @@
+/*
+Copyright 2024 The Dapr Authors
```
mikeee:
```suggestion
Copyright 2026 The Dapr Authors
```

bindings/sqlserver/sqlserver_test.go
```diff
@@ -0,0 +1,235 @@
+/*
+Copyright 2024 The Dapr Authors
```
mikeee:
```suggestion
Copyright 2026 The Dapr Authors
```

tests/certification/bindings/sqlserver/sqlserver_test.go
```diff
@@ -0,0 +1,296 @@
+/*
+Copyright 2024 The Dapr Authors
```
mikeee:
```suggestion
Copyright 2026 The Dapr Authors
```

mikeee (changes requested):

bindings/sqlserver/sqlserver.go
```diff
@@ -0,0 +1,330 @@
...
+			}
+		}
+
+		if value != nil {
+			r[ct.Name()] = value
+		}
+	}
+
+	return r
+}
```
JoshVanL, line 319: We need to preserve `Null` values.

```suggestion
func (s *SQLServer) jsonify(rows *sql.Rows) ([]byte, error) {
      columnTypes, err := rows.ColumnTypes()
      if err != nil {
              return nil, err
      }

      var ret []any
      for rows.Next() {
              values := make([]any, len(columnTypes))
              for i := range values {
                      values[i] = new(any)
              }
              err = rows.Scan(values...)
              if err != nil {
                      return nil, err
              }

              r := s.convert(columnTypes, values)
              ret = append(ret, r)
      }
      if err := rows.Err(); err != nil {
              return nil, err
      }

      return json.Marshal(ret)
}

func (s *SQLServer) convert(columnTypes []*sql.ColumnType, values []any) map[string]any {
      r := map[string]any{}

      for i, ct := range columnTypes {
              value := *(values[i].(*any))
              if value == nil {
                      continue
              }

              // go-mssqldb returns DECIMAL, NUMERIC, MONEY and SMALLMONEY values as a
              // []byte holding the number's string form; convert to string so JSON
              // marshalling does not base64-encode it.
              if b, ok := value.([]byte); ok {
                      switch ct.DatabaseTypeName() {
                      case "DECIMAL", "MONEY", "SMALLMONEY":
                              value = string(b)
                      }
              }

              r[ct.Name()] = value
      }

      return r
}
```

JoshVanL (changes requested):

JoshVanL (dismissed):

JoshVanL:
Hey @adam6878, thank you for the contribution!

CI is failing- think you just need to run a `$ make modtidy-all`

adam6878:
@JoshVanL to completely close the issue check pull request on the main dapr repo and the docs
https://github.com/dapr/dapr/pull/10315
https://github.com/dapr/docs/pull/5272

JoshVanL (dismissed):

tests/certification/bindings/sqlserver/docker-compose.yml
```diff
@@ -0,0 +1,9 @@
+version: "3.7"
+services:
+  sqlserver:
+    image: mcr.microsoft.com/mssql/server:2019-latest
```
mikeee, line 4: Can you please update to a later version of MSSQL Server that is not in its extended support phase?

bindings/sqlserver/sqlserver.go
```diff
@@ -0,0 +1,326 @@
...
+
+		r := s.convert(columnTypes, values)
+		ret = append(ret, r)
+	}
+	if err := rows.Err(); err != nil {
+		return nil, err
+	}
+
+	return json.Marshal(ret)
+}
```
mikeee, line 294: In the `metadata.yaml` you've specified that it'll return 'an array of row values' but `null` can be marshalled here. If you want an array of row values regardless of result, could you please initialise the ret slice

bindings/sqlserver/sqlserver.go
```diff
@@ -0,0 +1,326 @@
...
+func (s *SQLServer) Close() error {
+	if !s.closed.CompareAndSwap(false, true) {
+		// If this failed, the component has already been closed
+		// We allow multiple calls to close
+		return nil
+	}
+
+	if s.db != nil {
+		_ = s.db.Close()
+		s.db = nil
```
mikeee, line 212:
```suggestion
		return s.db.Close()
```

bindings/sqlserver/sqlserver.go
```diff
@@ -0,0 +1,326 @@
...
+		value := *(values[i].(*any))
+		if value == nil {
+			continue
+		}
+
+		// go-mssqldb returns DECIMAL, NUMERIC, MONEY and SMALLMONEY values as a
+		// []byte holding the number's string form; convert to string so JSON
+		// marshalling does not base64-encode it.
+		if b, ok := value.([]byte); ok {
+			switch ct.DatabaseTypeName() {
```
mikeee, line 309: I think also adding a case here to convert the GUID/UNIQUEIDENTIFIER is worth it from a usability perspective to make these columns usable from the binding.

mikeee (changes requested):

mikeee (approved):

PR #4491, opened 2026-07-30
Azure auth: report SPIFFE credential as unavailable without a JWT SVID source

### LINKED ISSUE #4467: Azure auth: SPIFFE credential hard-fails and halts the chain, breaking fallback
## Description

For Azure (Microsoft Entra ID) components, the SPIFFE workload-identity credential returns a fatal error when its prerequisite — a JWT-SVID source in the context — is not available. This halts `ChainedTokenCredential` and prevents any subsequent credential in the chain from being attempted.

This breaks fallback in two cases:

1. **Default chain (implicit).** The SPIFFE credential is added to the default chain whenever `azureClientId` + `azureTenantId` are present. If no SPIFFE source is available, the chain stops at the SPIFFE step and never reaches managed identity or Azure CLI. This surfaces when no client secret or certificate is configured, since client credentials would otherwise succeed earlier in the chain.
2. **Explicit multi-method config.** Setting `azureAuthMethods: "spiffeworkloadidentity,managedidentity"` (try SPIFFE, then fall back to managed identity) does not fall back — the SPIFFE step's fatal error still halts the chain before managed identity is attempted.

In both cases the expectation is that a credential whose prerequisites are unavailable defers to the next credential in the chain, rather than terminating it.

## Mechanism

In `common/authentication/azure/auth.go`, the chain (default order, or the order given in `azureAuthMethods`) adds the SPIFFE credential via `addSpiffeWorkloadIdentityProvider`. `GetSpiffeWorkloadIdentity()` adds it based only on `clientId` + `tenantId` being present — it does not check whether a SPIFFE source is actually available.

At token time, `common/authentication/azure/spiffe.go` returns a plain error when there is no source in context:

```go
tknSource, ok := spiffecontext.JWTFrom(ctx)
if !ok {
    return "", errors.New("failed to get JWT SVID source from context")
}
```

Because that is not a `credentialUnavailableError`, `azidentity.ChainedTokenCredential.GetToken` breaks the loop rather than continuing to the next credential:

```go
errs = append(errs, err)
// continue to the next source iff this one returned credentialUnavailableError
if !errors.As(err, &unavailableErr) {
    break
}
```

## Steps to reproduce

Either:

- Deploy an Azure component with `azureClientId` + `azureTenantId` set, no secret/certificate, and no SPIFFE/OIDC federation configured; or
- Set `azureAuthMethods: "spiffeworkloadidentity,managedidentity"` on such a component.

Start the sidecar.

## Actual

The chain stops at the SPIFFE credential; no subsequent credential is attempted:

```
ChainedTokenCredential: failed to acquire a token.
Attempted credentials:
	ClientAssertionCredential: failed to get JWT SVID source from context
```

The remaining credentials in the chain never appear in the "Attempted credentials" list.

## Expected

When a credential's prerequisites are unavailable, the remaining credentials in the chain are still attempted — for both the default chain and explicitly-ordered `azureAuthMethods` lists.

## Version boundary

- Not present on Dapr 1.15.x (components-contrib 1.15.x) — the SPIFFE credential is not in the chain.
- Present on Dapr 1.16.0+ (components-contrib 1.16.0+) — the SPIFFE credential was added to the default chain in #3797.

## References

- `common/authentication/azure/auth.go` — `GetTokenCredential`, `addSpiffeWorkloadIdentityProvider`
- `common/authentication/azure/spiffe.go`
- #3797 — introduced the SPIFFE credential

MyMirelHub:
## Description

Fixes the credential chain halting at the SPIFFE workload identity step. See #4467 for the failure modes.

The credential is now wrapped so that a missing JWT SVID source is reported as `credentialUnavailableError` before any token request is made, which is what `ChainedTokenCredential` requires in order to continue to the next credential.

## Issue reference

Closes #4467

## Testing

Added `common/authentication/azure/spiffe_test.go`, covering chain fall-through, the error not being a fatal `AuthenticationFailedError`, and the existing `clientId`/`tenantId` validation.

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] documented in dapr/docs#5269

### DIFF
--- a/common/authentication/azure/spiffe.go
+++ b/common/authentication/azure/spiffe.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/spiffe/go-spiffe/v2/svid/jwtsvid"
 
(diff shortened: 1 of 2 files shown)


mikeee (approved):
lgtm

PR #4496, opened 2026-08-01
fix(state/in-memory): avoid deadlock between Close and cleanup goroutine

### LINKED ISSUE #4495: state/in-memory: Close() can deadlock with the cleanup goroutine
Close() calls wg.Wait() while holding the write lock. The cleanup goroutine started by Init() needs that same lock in doCleanExpiredItems(), so if the cleanup ticker fires between Close() acquiring the lock and the goroutine reaching Lock(), the two wait on each other and Close() never returns — a permanent hang with no timeout or error. Reproduces reliably with concurrent Close() calls; details and a fix in the linked PR.

karprabha:
# Description

`Close()` calls `wg.Wait()` while holding the write lock. The cleanup goroutine
started by `Init()` needs that same lock in `doCleanExpiredItems()`. If the
cleanup ticker fires in the window between `Close()` acquiring the lock and the
goroutine reaching `Lock()`, the two wait on each other and `Close()` never
returns.

This is a permanent hang — no timeout, no error, no retry. Under Kubernetes it
surfaces as a sidecar stuck terminating until SIGKILL, with nothing in the logs.

**Fix:** wait for the cleanup goroutine to exit *before* acquiring the lock.
Once `wg.Wait()` returns, nothing can contend for it.

Introduced in d098e38d (#2474, Feb 2023).

### Reproduction

The window is microseconds wide, so it reproduces under scheduling pressure
rather than precise timing: many stores closed concurrently, timed to coincide
with their cleanup tick. It is not single-core specific — measured on a 10-core
machine, 6 rounds per configuration:

| GOMAXPROCS | stores | before | after |
|---|---|---|---|
| 1  | 2000 | 6/6 deadlocked | 0/6 |
| 2  | 2000 | 4/6 | 0/6 |
| 4  | 2000 | 5/6 | 0/6 |
| 10 | 2000 | 3/6 | 0/6 |
| 10 | 4    | 1/6 | 0/6 |

Goroutine dump at the deadlock:

    goroutine [sync.Mutex.Lock]:
      ...doCleanExpiredItems()  in_memory.go:444   <- blocked on store.lock
    goroutine [sync.WaitGroup.Wait]:
      ...Close()                in_memory.go:90    <- holds store.lock

### Cleanup ticker now uses the injected clock

`startCleanThread` used `time.After` directly, so the cleanup tick ignored the
clock injected into the store. It now uses `store.clock.After`, with the interval
hoisted into a package constant shared with the test.

This means a fake clock now governs the sweeper as well - `fakeClock.Step` in
`TestReadAndWrite` drives a sweep, so tests get full control of the component's
time rather than a background goroutine ticking on real time underneath them.
Existing tests pass unchanged.

## Issue reference

Closes #4495

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
    * [ ] Created the dapr/docs PR: N/A

No documentation changes are required: this is an internal concurrency fix with
no change to component behavior, configuration metadata, or public API.


mikeee (commented):
Could you please consider including the proposed changes to `startCleanThread` in this PR to reduce the CI run load?

karprabha:
Done, pushed. `startCleanThread` now uses `store.clock.After` with the interval as a package constant, so the sweeper follows the injected clock like the rest of the component. Tests get full control of the component's time now - `fakeClock.Step` in `TestReadAndWrite` drives a sweep. Existing tests pass unchanged; `-race` and `make lint` clean.

Separately, `pubsub.mqtt3 certification` failed. This only touches `state/in-memory` and `state.in-memory conformance` passed, so it looks unrelated.

state/in-memory/in_memory_test.go
```diff
@@ -178,6 +180,74 @@ func TestReadAndWrite(t *testing.T) {
...
+// rather than precision: many stores are closed at once, timed to coincide with
+// their cleanup tick, on a single processor so goroutines are forced to
+// interleave at exactly that point.
+//
+// This test changes GOMAXPROCS, so it must not call t.Parallel.
+func TestCloseDoesNotDeadlock(t *testing.T) {
+	const (
+		// A single store almost never lands in the window; a few thousand
+		// closing together reliably do.
+		storeCount = 2000
```
mikeee: Do 2000 routines need to still be launched?

mikeee (changes requested):

state/in-memory/in_memory_test.go
karprabha: No, not anymore .. since the ticker now runs on the injected clock, the test doesn't need to brute-force it. It waits for the cleanup goroutine to go to sleep waiting for its tick, fires the tick with the fake clock, then calls Close before the goroutine gets a chance to run.. so Close grabs the lock first and the goroutine gets stuck behind it, which is exactly the deadlock. One store, no sleeps, runs in a few ms.

karprabha:
Hi @mikeee I've made the said changes Kindly review it again

karprabha (commented):
Hi @mikeee Its implemented

karprabha:
Hi @sicoyle sorry for pinging directly but can you review the PR ? Its been here for sometime

karprabha:
Hi @JoshVanL can you review and merge this PR?

state/in-memory/in_memory.go
```diff
@@ -80,15 +84,18 @@ func (store *InMemoryStore) Close() error {
 		close(store.closeCh)
 	}
 
+	// Wait for the cleanup goroutine to exit before acquiring the lock. It
+	// needs the same lock in doCleanExpiredItems, so waiting on it while
+	// holding the lock would deadlock.
+	store.wg.Wait()
```
wasniksudesh: nit : wg.Wait() must stay last blocking step before return . 
if someone adds operationY() in Close() just before returning, this operationY might add goroutine to store.wg, but we are not waiting on store.wg to finish, and Close() exits.

```suggestion
	defer store.wg.Wait()
```
wdyt ?

karprabha: Yeah good point .. implemented that with a comment that .. it only stays deadlock free because the wg.Wait defer is registered before the lock's Unlock defer (they run in reverse order). Flip the two and the deadlock comes back.. the regression test catches it in seconds if tried flipping.

wasniksudesh:
looks good 👍 🚀

karprabha:
> looks good 👍 🚀

@wasniksudesh  can you approve the PR?

PR #4498, opened 2026-08-03
fix(common/redis): Pipeliner.Do queues every command twice when writeTimeout is set

### LINKED ISSUE #4464: Redis: Pipeliner.Do executes every pipelined command twice when writeTimeout is set (breaks transactional ETag/first-write CAS, doubles Lua load)
## Summary

`Pipeliner.Do` in the shared Redis client queues every command **twice** when `writeTimeout` is configured: the `writeTimeout > 0` branch queues the command with a timeout context but is missing a `return`, so execution falls through and queues the same command again. Every command sent through a transaction (`TxPipeline`) then executes **twice per `Exec`**.

## Exact code

Both clients, same defect (permalinks to current `main` @ cc03682):

- https://github.com/dapr/components-contrib/blob/cc03682550d86cceac2ada5f04de7f9d40f24492/common/component/redis/v9client.go#L36-L43
- https://github.com/dapr/components-contrib/blob/cc03682550d86cceac2ada5f04de7f9d40f24492/common/component/redis/v8client.go#L36-L43

```go
func (p v9Pipeliner) Do(ctx context.Context, args ...interface{}) {
	if p.writeTimeout > 0 {
		timeoutCtx, cancel := context.WithTimeout(ctx, time.Duration(p.writeTimeout))
		defer cancel()
		p.pipeliner.Do(timeoutCtx, args...)
	}
	p.pipeliner.Do(ctx, args...)   // <-- also runs when writeTimeout > 0: command queued twice
}
```

Compare with `DoWrite`/`DoRead` on the same clients, which have the intended `return` in the timeout branch.

## Impact

For any Redis state store component configured with `writeTimeout` metadata:

1. **State transactions execute every operation twice.** `StateStore.Multi` pipelines the CAS `EVAL` (`setDefaultQuery`) per operation, so a single transactional write runs the script twice — the `version` hash field increments by **2** per write, and ETag/`first-write` semantics break.
2. **Dapr Workflow / actor state stores fail hard on daprd >= 1.17.9.** The newer workflow engine's write pattern plants the Lua `first-write` flag; the duplicated second `EVAL` then hits the closed CAS check and every second workflow write fails with:
   ```
   ERR user_script:14: failed to set key <app>||dapr.internal.<ns>.<app>.workflow||<instance>||metadata
   ```
   (On daprd <= 1.17.6 the duplication is *silent* — versions inflate but the escape hatch in the Lua masks it.)
3. **~2x Redis CPU for transactional traffic** — the CAS `EVAL` is the most expensive operation the store issues, and it runs twice per op. On smaller managed SKUs this is a visible serverLoad tax.

We hit this in production (Azure Managed Redis, Dapr Workflow via actor state stores): every workflow `Submit` failed once daprd moved to 1.17.9, and Redis CPU dropped materially after de-arming the bug. It reproduces identically through built-in `state.redis` and through a pluggable component, since the defect is in the shared client.

## Reproduction (minimal)

1. Redis state store component with:
   ```yaml
   metadata:
   - name: writeTimeout
     value: "3s"
   ```
2. One transactional upsert:
   ```
   POST /v1.0/state/<store>/transaction
   {"operations":[{"operation":"upsert","request":{"key":"k","value":"v1"}}]}
   ```
3. `GET /v1.0/state/<store>/k` → **ETag is `2`** (expected `1`). Any subsequent transactional write carrying the correct ETag + `first-write` concurrency then fails CAS (`user_script:14`), because the duplicated second `EVAL` of *that* write sees the version its own first execution just bumped.

A single pipelined `INCR` shows it too: `TxPipeline` + `Do("INCR", "counter")` + `Exec` → counter is `2` when `writeTimeout` is set, `1` when unset.

## Fix

One line — add `return` after the timeout-context call (both `v8client.go` and `v9client.go`):

```go
	if p.writeTimeout > 0 {
		timeoutCtx, cancel := context.WithTimeout(ctx, time.Duration(p.writeTimeout))
		defer cancel()
		p.pipeliner.Do(timeoutCtx, args...)
		return
	}
```

We're running this patch with a regression test (miniredis: pipelined `INCR` must land on exactly `1` with and without `writeTimeout`) and can open a PR with both if useful.

paulstadler-mesh:
## Description

Fixes #4464.

`Pipeliner.Do` (both `v8client.go` and `v9client.go`) is missing a `return` in its `writeTimeout > 0` branch: after queueing the command with a timeout context, execution falls through and queues the same command **again** with the unbounded context. Every command sent through a `TxPipeline` then executes **twice per `Exec`** whenever the component sets `writeTimeout` metadata.

## Impact

- **State transactions execute every operation twice.** `StateStore.Multi` pipelines the CAS `EVAL` per operation, so a single transactional write runs the script twice — the `version` hash field increments by 2 per write, breaking ETag/`first-write` semantics.
- **Dapr Workflow / actor state stores fail hard on daprd ≥ 1.17.9** (whose workflow engine plants the Lua `first-write` flag): every second workflow write fails with `ERR user_script:14: failed to set key …||workflow||…||metadata`. On earlier daprd the duplication is silent — versions inflate and transactional Redis CPU is ~doubled.
- Encountered in production against Azure Managed Redis: all Dapr Workflow submissions failed once daprd moved to 1.17.9; Redis CPU dropped materially once the duplication was removed. Reproduces identically through built-in `state.redis` and pluggable components (shared client).

Minimal repro and full analysis in #4464 (e.g. one transactional upsert on a fresh key → `GET` returns ETag `2`).

## Fix

Add the missing `return` (mirrors the existing pattern in `DoWrite`/`DoRead` on the same clients).

## Testing

New regression test `TestPipelinerDoQueuesOnceWithWriteTimeout` drives a pipelined `INCR` through both v8 and v9 clients against miniredis and asserts single execution, with and without `writeTimeout` (fails on the previous code with counter=2).

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Extended the documentation — n/a (no user-facing behavior beyond the bug fix)

### DIFF
--- a/common/component/redis/pipeliner_timeout_test.go
+++ b/common/component/redis/pipeliner_timeout_test.go
@@ -1,0 +1,71 @@
+/*
+Copyright 2021 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package redis
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/alicebob/miniredis/v2"
+	"github.com/stretchr/testify/require"
+)
+
+// TestPipelinerDoQueuesOnceWithWriteTimeout is the regression test for the
+// double-queue bug: Pipeliner.Do used to queue the command TWICE when
+// writeTimeout was set (the writeTimeout branch was missing a return, so the
+// unbounded call below it also ran). Every pipelined command — including the
+// state store's transactional CAS EVALs — then executed twice per Exec,
+// inflating versions and breaking ETag/first-write semantics for any
+// component configured with writeTimeout (Dapr Workflow actor state being the
+// visible casualty: "ERR user_script:14: failed to set key").
+func TestPipelinerDoQueuesOnceWithWriteTimeout(t *testing.T) {
+	ctx := context.Background()
+
+	run := func(t *testing.T, s *Settings) {
+		t.Helper()
+		mr := miniredis.RunT(t)
+		s.Host = mr.Addr()
+		s.RedisType = NodeType
+
+		clients := map[string]func(*Settings) (RedisClient, error){
+			"v8": newV8Client,
+			"v9": newV9Client,
+		}
+		for name, newClient := range clients {
+			t.Run(name, func(t *testing.T) {
+				c, err := newClient(s)
+				require.NoError(t, err)
+				defer c.Close()
+
+				mr.FlushAll()
+				pipe := c.TxPipeline()
+				pipe.Do(ctx, "INCR", "counter")
+				require.NoError(t, pipe.Exec(ctx))
+
+				got, err := c.Get(ctx, "counter")
+				require.NoError(t, err)
+				require.Equal(t, "1", got,
+					"a single pipelined INCR must execute exactly once (twice means Do double-queued)")
+			})
+		}
+	}
+
+	t.Run("writeTimeout set (the arming condition)", func(t *testing.T) {
+		run(t, &Settings{WriteTimeout: Duration(3 * time.Second)})
+	})
+	t.Run("writeTimeout unset", func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


paulstadler-mesh:
@mikeee sorry for the direct ping — this one's a 2-line fix (+regression test) for #4464: `Pipeliner.Do` double-queues every pipelined command when `writeTimeout` is set, which breaks transactional ETag/first-write semantics (and fails Dapr Workflow state saves on daprd ≥ 1.17.9). We've been running the fix in production for a couple of weeks.

The CI workflows are sitting at *awaiting approval* (fork PR) — if you could approve the runs, everything else should go green from there. Happy to make any changes needed. Thanks!

PR #4501, opened 2026-08-04
Feat/conversation max tokens

### LINKED ISSUE #10317: ConverseAlpha2 `parameters` field on doesn't work
## In what area(s)?

/area runtime

## What version of Dapr?

1.18.2

## Expected Behavior

For the Conversation API parameters to work for input like `max_tokens`, etc.

## Actual Behavior

Conversation API parameters (map<string, google.protobuf.Any>) is documented as supporting provider-specific overrides like max_tokens, but it's never actually read anywhere in the request path in Dapr. The field is silently dropped. For all conversation component providers this means there's currently no way to bound output length — degenerate/repetitive generations can run 90–230+ seconds with no cap.

## Root cause (from Claude)

1.  pkg/api/universal/conversation.go, ConverseAlpha2 (lines 178–576) copies Temperature, Tools, ToolChoice, and PromptCacheRetention from the request into conversation.Request — but never reads req.GetParameters(). It's dropped before reaching the component.
2. This is a known dead field on both ends of the stack, not just an oversight:
- dapr/proto/runtime/v1/ai.proto:62 — "when we go stable we need to remove context_id and parameters in ConversationRequestAlpha2 as these are not used."
- components-contrib conversation/converse.go:38 — Parameters map[string]*anypb.Any tagged // TODO: rm these in future PR as they are not used.
3. Even if parameters were wired through, there's a second gap downstream: components-contrib's conversation/langchaingokit/model.go:147-180 (shared by all langchaingo-backed providers, including googleai) has // llms.WithMaxTokens() commented out under a // NOTE: we can add these in future! block. googleai.go has no GenerationConfig/max-tokens handling of its own.
4. No integration test catches this: tests/integration/suite/daprd/conversation/alpha2/grpc/basic.go:89-92 sends a max_tokens parameter but never asserts it had any effect, and the echo test component doesn't consume it either.

```
run("no-params", None)
run("max_tokens=10 (SDK int)", {"max_tokens": 10})
run("max_tokens=10 (explicit Int64Value)", {"max_tokens": <Any-wrapped Int64Value>})
```
Note that all three return byte-identical output — the cap has zero effect regardless of encoding.

## Release Note

RELEASE NOTE:

CasperGN:
# Description

Fixes dapr/dapr#10317

## Issue reference

Please reference the issue this PR will close: dapr/dapr#10317

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
    * [x] Created the dapr/docs PR: dapr/docs#5275


mikeee:
@CasperGN looking good -  could you please run `make modtidy-all` on this branch?

tests/conformance/conversation/conversation.go
mikeee, line 1: More of a nit - should there be a test case asserting `maxToken `is defined even when `resp == nil`?

CasperGN:
@mikeee done. Let me also increase the coverage a bit!

mikeee:
> @mikeee done. Let me also increase the coverage a bit!

Sure, I'll hold fire on reviewing until I'm tagged again 👍

CasperGN:
@mikeee I think it's good to go now. Added you nit too

conversation/langchaingokit/model.go
```diff
@@ -47,8 +51,51 @@ func (a *LLM) GetModel() string {
...
+// SetDefaultMaxTokens sets the component-level default cap on generated tokens.
+// It is applied to every request unless the request carries its own positive
+// MaxTokens, which then takes precedence (langchaingo applies call options in
+// order; later wins). A request-level MaxTokens of zero or a negative value is
+// treated as unset and leaves this default (if any) in effect. A non-positive
+// default is ignored with a warning.
+func (a *LLM) SetDefaultMaxTokens(maxTokens *int64) {
+	if maxTokens != nil && *maxTokens <= 0 {
+		a.logger.Warnf("ignoring non-positive maxTokens component default %d", *maxTokens)
+	}
```
JoshVanL, line 67: This isn't ignoring

JoshVanL (changes requested):

tests/conformance/conversation/conversation.go
```diff
@@ -95,6 +95,45 @@ func ConformanceTests(t *testing.T, props map[string]string, conv conversation.C
...
+			// The cap must be observable in the reported usage. Content may
+			// legitimately be empty when a reasoning model spends the whole
+			// budget thinking, so only usage is asserted; providers that omit
+			// usage data entirely are logged rather than failed, matching the
+			// other usage checks in this suite.
+			if resp.Usage != nil {
+				assert.LessOrEqual(t, resp.Usage.CompletionTokens, uint64(maxTokensLimit),
+					"completion tokens must not exceed the requested max_tokens")
+			} else {
+				t.Logf("%s: no usage data returned; cannot verify max tokens via usage", component)
```
mikeee: The tests shouldn't succeed if a provider returns no usage data - making no assertions against the `maxTokens` used

mikeee (changes requested):

conversation/anthropic/metadata.yaml
```diff
@@ -55,3 +55,10 @@ metadata:
       A time-to-live value for a prompt cache to expire. Uses Golang durations
     type: string
     example: '10m'
+  - name: maxTokens
+    required: false
+    description: |
+      Default maximum number of tokens the model may generate per request.
+      A request-level maxTokens value greater than zero overrides this default.
```
sicoyle, line 62: what is the default? Can you put that under the `default` field here pls?

sicoyle: pls apply this change to the rest of the metadata.yaml files in this PR

sicoyle (commented):
two minor comments and then LGTM 🙌

conversation/anthropic/metadata.yaml
CasperGN: Done!

sicoyle (commented):
LGTM - thank youu!!

sicoyle (approved):

sicoyle:
merging - the ravendb conformance test failures are unrelated

CasperGN:
@mikeee & @JoshVanL can you re-review so this can roll?

sicoyle (dismissed):

sicoyle:
bump @mikeee and @JoshVanL pls bc i believe this is g2g

conversation/openai/openai.go
```diff
@@ -99,6 +99,7 @@ func (o *OpenAI) Init(ctx context.Context, meta conversation.Metadata) error {
 
 	o.Model = llm
 	o.SetModel(model)
+	o.SetDefaultMaxTokens(md.MaxTokens)
```
mikeee, line 102: This I believe constructs a request with `max_completion_tokens` if I'm not mistaken. 

The current OpenAI component docs suggest that we are compatible with custom endpoints; thus max_token should still be a valid override if requested; similarly to the ollama component. 

Do you think this is an issue until the key is fully deprecated- I understand the streamlining of`max_tokens` here but perhaps a selector e.g. `maxTokensParameter: max_completion_tokens|max_tokens`  might aid compatibility, has maintaining the `max_completion_tokens` been considered rather than our own abstraction around maxtokens/maxcompletiontokens?

mikeee (changes requested):
I have some concerns on using the `maxTokens` key in its current state - and whether we should defer/conform to `maxCompletionTokens`.

Should we be normalising the `finishReason` across all providers?

cicoyle (approved):
lgtm, just want to confirm - Is all the feedback addressed @CasperGN? If so, mind resolving it?

PR #4508, opened 2026-08-04
Kafka pubsub: transactions and exactly-once consume-transform-produce

### LINKED ISSUE #4301: support kafka EOS (exactly only once semantics)
## Describe the feature
Support **Kafka exactly-once semantics (EOS)** in the Dapr Kafka pub/sub component so that applications can get exactly-once processing when enabled via component configuration. Users would be able to opt in (e.g. via metadata) and have the component ensure that each message's offset commit and downstream publish succeed or fail atomically within a Kafka transaction, so duplicate processing is avoided in all scenarios — including failures and retries.

Kafka has supported exactly-once semantics since 0.11 (transactional produce/consume and offset commit). (see e.g. [Plans to support upcoming Kafka v0.11 ? (#901)](https://github.com/IBM/sarama/issues/901) and [Exactly Once arrived with kafka 0.11 ? (#977)](https://github.com/IBM/sarama/issues/977)); the feature request is to expose and certify this capability in the Dapr component. At a high level, enabling EOS would mean: 

- producing within a transaction (so sends are atomic), 
- consuming with read-committed isolation
- committing consumer offsets in the same transaction so processing and offset commit succeed or fail together—avoiding duplicates and double-processing when the component is used as configured.

**Business value:** Applications that need exactly-once or no-duplicate processing over Kafka could rely on Dapr’s Kafka component instead of building their own transaction handling, which is especially useful for event-driven workflows, audit trails, and pipelines that must not double-process messages.

## Release Note

RELEASE NOTE: **ADD** Kafka pub/sub component supports exactly-once semantics (EOS) via optional metadata, enabling transactional produce and consume so applications can avoid duplicate processing.

javier-aliaga:
# Description

Adds Kafka transactions support to the Kafka pubsub component (in the shared `common/component/kafka` code), enabling exactly-once processing between Kafka topics through three independently configurable capabilities:

- **Transactional publishing** (`producerTransactionsEnabled`): every publish — single or bulk — is wrapped in a Kafka transaction on an idempotent producer. Bulk publishes become atomic (all entries commit or none do), and aborted publishes are never visible to consumers reading with `read_committed`.
- **Committed-only consumption** (`consumerIsolationLevel: "read_committed"`): subscriptions never receive records from open or aborted transactions.
- **Exactly-once consume-transform-produce** (`consumerTransactionsEnabled`): every delivery is processed inside a Kafka transaction. The component injects a `__txnToken` metadata entry (reaching the app as a header via the runtime's existing metadata pass-through — no dapr/dapr changes needed). Publishes from the handler that echo the token join the delivery's open transaction, and the consumer offset commits atomically with them (KIP-447, `AddMessageToTxnWithGroupMetadata`). Handler failure aborts the whole transaction and the delivery is redelivered.

## Design notes

- **One producer per partition.** Each consumed topic-partition gets its own transactional producer, with the `transactional.id` `<prefix>-<digest>-<partition>`. The digest is the first 8 bytes of SHA-256 over the consumer group, a NUL byte and the topic. Every replica builds the same id for a partition, which is what lets a new owner fence a zombie after a rebalance or a crash. The group and topic are digested rather than written into the id because `-` is legal in both, so writing them out would let unrelated components build the same id and fence each other. The prefix is kept as it is, so `<prefix>-*` broker ACLs still work. No lock is held across the app handler: deliveries are serial within a partition.

- **Kafka 2.5 or later is required for `consumerTransactionsEnabled`**, where the other two capabilities need 0.11. The offset commit only carries the consumer group member id and generation from 2.5 on, and the broker needs them to fence a member whose assignment has been revoked. The component's default `version` is 2.0.0, so it has to be set explicitly.

- **Fail-loud token contract**: a publish carrying a stale or unknown token errors instead of silently publishing outside the transaction. The token is stripped from produced headers. A publish *without* a token is an ordinary publish (never an error).

- **Consumer transactions imply `read_committed`**: leaving `consumerIsolationLevel` unset selects it automatically; an explicit `read_uncommitted` fails validation. Autocommit is disabled, and sarama's session offset manager is not used at all in this mode: committing through it flushes every partition of the session in one request, which would let one partition write a stale offset for another. A delivery whose handler published nothing ends its empty transaction and commits its own offset with a single-partition request carrying the group generation and member id (sarama silently skips offset commits for record-less transactions).

- **Commits with an unknown outcome**: if a commit runs out of retries without an answer from the broker, the transaction may or may not have been committed. Reading the committed offset at that point cannot tell the two apart, so the partition's producer is rebuilt first — reusing the same `transactional.id` makes the coordinator settle the pending transaction — and only then is the offset read. If it cannot be settled, the delivery is processed again: a duplicate is better than a lost message.

- **Bulk subscribe** is all-or-nothing: any entry failure aborts and redelivers the whole batch.

- **Bindings**: `producerTransactionsEnabled` is supported for output; `consumerTransactionsEnabled` is rejected at init — the bindings model (separate input and output instances) cannot echo the token through the same component instance. Consume-transform-produce is pubsub-only.

- **Two flags, not one**: producer and consumer transactions are configured separately, mirroring Kafka's own config split (`transactional.id`/idempotence are producer configs, `isolation.level` is a consumer config). A merged flag would force per-delivery transaction overhead onto consume-and-publish apps that don't need correlation, and serialized transactional publishing onto apps that only need transactional consume.

- **Shutdown**: `Close()` preserves the prompt consumer-group teardown contract from #3907 — the group closes immediately and never waits on an in-flight transactional publish. When a publish is in flight, the producer close is handed to a goroutine that waits for it to finish, so nothing is left behind when the component is reloaded.

**Scope note**: the guarantee covers Kafka records and consumer offsets. The app handler itself still runs at-least-once — non-Kafka side effects are outside the transaction.

## Issue reference

Closes #4301

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
    * [x] Created the dapr/docs PR: https://github.com/dapr/docs/pull/5276


common/component/kafka/kafka.go
```diff
@@ -44,6 +44,18 @@ type Kafka struct {
 	mockConsumerGroup sarama.ConsumerGroup
 	mockProducer      sarama.SyncProducer
 	clients           *clients
+	// clientsLock guards creating the client pair (and awsClients),
+	// invalidating the producer after a fatal transaction error, and
+	// detaching either client during teardown. It does NOT guard every read
```
JoshVanL: This comment documents a data race instead of closing it, and I do not think it should ship that way.

`Close` now writes `k.clients.consumerGroup = nil` under `clientsLock` (kafka.go:463), while three readers still touch the same field holding `subscribeLock` only: `Pause` (kafka.go:403), `Resume` (kafka.go:418) and the graceful unsubscribe (subscriber.go:62 and subscriber.go:81). `subscribeLock` does not exclude that write, so this is an unsynchronized read/write of an interface value, not just a logically stale read.

The comment's reasoning is that the worst case is "PauseAll or ResumeAll landing on a group that is closing, which sarama answers with an atomic store". That is the outcome if the read returns a valid pointer. The outcome it misses is a torn interface read: the type word set and the data word already nil, so `PauseAll` runs on a nil `*consumerGroup` and panics on `c.lock.Lock()`. The race detector will flag it too, whether or not it ever tears in production.

Agreed the raciness predates the PR (on main the field was written with no lock at all), but this PR is the one introducing `clientsLock`, so it is the natural place to finish the job. A read helper is enough:

```go
// currentConsumerGroup returns the cached consumer group, read under
// clientsLock. Close detaches the field under that lock, so every reader has
// to take it too: Pause, Resume and the graceful-unsubscribe path hold
// subscribeLock only, which does not exclude the detach.
func (k *Kafka) currentConsumerGroup() sarama.ConsumerGroup {
	k.clientsLock.Lock()
	defer k.clientsLock.Unlock()
	if k.clients == nil {
		return nil
	}
	return k.clients.consumerGroup
}
```

Then `if cg := k.currentConsumerGroup(); cg != nil { cg.PauseAll() }` in the four call sites. Lock order stays `subscribeLock` then `clientsLock` everywhere, and `Close` releases `subscribeLock` before it takes `clientsLock`, so this adds no nesting and no deadlock. Verified with `go test -race ./common/component/kafka/...`.

common/component/kafka/consumer.go
```diff
@@ -225,6 +273,365 @@ func (consumer *consumer) doCallback(session sarama.ConsumerGroupSession, messag
...
+	resp, err := coordinator.CommitOffset(req)
+	if err != nil {
+		return fmt.Errorf("commit offset: %w", err)
+	}
+	kerr, ok := resp.Errors[message.Topic][message.Partition]
+	if !ok {
+		return errors.New("commit offset: no result for the partition")
+	}
+	if kerr != sarama.ErrNoError {
+		return fmt.Errorf("commit offset: %w", kerr)
```
JoshVanL, line 522: `ErrNotCoordinatorForConsumer` here leaves the client's cached coordinator in place, so the redelivery retries cannot make progress.

sarama's `client.Coordinator(group)` (consumer.go:494) returns `cachedCoordinator` and only does a lookup when the cache is empty. Nothing invalidates it on a `NOT_COORDINATOR` response; sarama's own offset manager calls `RefreshCoordinator` for exactly this reason (sarama v1.60 `offset_manager.go`).

Sequence: the group coordinator migrates (broker restart, `__consumer_offsets` leader move). A record-less delivery reaches `commitOffset`, the old broker answers `NOT_COORDINATOR`, the delivery fails and `retry.NotifyRecover` reruns it. Every rerun reuses the claim producer, so the same client, so the same cached broker, so the same error. The partition's offset stops advancing and the app handler re-runs on each attempt until the session finally dies and a fresh claim builds a new client.

`fetchCommittedOffset` (consumer.go:629) has the same gap, and it matters more there: it is the path that decides whether an ambiguous transaction landed.

No retry loop needed, the redelivery path already retries. It just needs the cache dropped so the next attempt looks the coordinator up again:

```go
// dropStaleCoordinator clears the client's cached group coordinator when the
// broker answers that it is not the coordinator any more. sarama caches the
// coordinator until something asks for a refresh, so without this the
// redelivery retries keep sending to the migrated-away broker and the
// partition's offset never advances again.
func dropStaleCoordinator(client sarama.Client, group string, kerr sarama.KError) {
	switch kerr { //nolint:exhaustive // only the coordinator-moved errors matter
	case sarama.ErrNotCoordinatorForConsumer, sarama.ErrConsumerCoordinatorNotAvailable:
		_ = client.RefreshCoordinator(group)
	}
}
```

Called just before the `return` here and before the `block.Err` return in `fetchCommittedOffset`. That needs `owner.Client()` hoisted into a local at consumer.go:494 so both the lookup and the refresh use the same client.

common/component/kafka/transactions.go
```diff
@@ -0,0 +1,172 @@
...
+	partition int32
+	producer  sarama.SyncProducer
+}
+
+func (k *Kafka) newClaimTxn(topic string, partition int32) *claimTxn {
+	return &claimTxn{k: k, topic: topic, partition: partition}
+}
+
+func (ct *claimTxn) transactionalID() string {
+	return ct.k.txnIDPrefix + "-" + ct.k.consumerGroup + "-" + ct.topic + "-" + strconv.Itoa(int(ct.partition))
```
JoshVanL: Not blocking, but I would reconsider the "document it" resolution from dfadb17.

The argument in that commit is that escaping would change an id format operators set broker ACLs on. That holds for a released format, but this one ships for the first time in this PR, so nobody has ACLs on it yet. This is the cheapest moment it will ever be to change.

Weighed against that: the failure mode is two components fencing each other's producers on every transaction, forever, with no error that names the cause. Someone will lose a day to that, and the doc line in `metadata.yaml` only helps whoever thinks to go back and read it.

A digest keeps prefix ACLs working and removes the ambiguity entirely:

```go
func (ct *claimTxn) transactionalID() string {
	sum := sha256.Sum256([]byte(ct.k.consumerGroup + "\x00" + ct.topic))
	return fmt.Sprintf("%s-%x-%d", ct.k.txnIDPrefix, sum[:8], ct.partition)
}
```

It stays stable across replicas, stays inside Kafka's id limits, and `<prefix>-*` ACL patterns are unaffected. What it costs is the readable group and topic inside the id, which is the real trade-off and the reason this is your call rather than a request.

JoshVanL (changes requested):

JoshVanL (approved):

PR #4518, opened 2026-08-10
fix(pubsub/redis): bound XREADGROUP BLOCK duration to release canceled subscriptions

### LINKED ISSUE #4497: Redis PubSub not dropping subscriptions when main process drops them
## Expected Behavior

In .NET version of the runtime, when the main process (application) disposes of the subscription object, the side-car should also dispose/close the subscription connection to Redis Stream.

## Actual Behavior

Currently, when the subscription object is disposed in the main process, the subscription to Redis Stream in the sidecar remains. This leads to an exhaustion problem, where the sidecar hits it's limit for concurrent connection, because it never releases them.

Reproducing this issue requires a multi-component setup. You will need:

- Running main process
- Running daprd side-car
- Redis (for initial reproduction, Docker container redis:8.2-alpine was used

Configuration used for the DAPR Redis component:

```YAML
spec:
  type: pubsub.redis
  version: v1
  metadata:
    - name: redisHost
      value: "127.0.0.1:6399" # Use actual Redis address and port during the test
    - name: enableTLS
      value: "false"
```

The main process must call multiple subscribes using the .NET DaprPublishSubscribeClient. What I did in the initial test was to call DaprPublishSubscribeClient.SubscribeAsync() every 20 seconds, collect the resulting IAsyncDisposable and dispose all created disposables every minute.

This allowed to observe, using Redis Insight and Redis INFO CLIENTS command the amount of blockedClients which raises every subscribe call made by the main process. On release, it is possible to observe that this number does not go down.

## Further info

This was discovered in .NET ASP.NET Core application and then reproduced using simple .NET 10 console application. However I do believe the bug will be present in other SDKs as well as this seems to be a component bug.

### Possible root cause
Since the workaround for the bug is to simply set readTimeout to other time than 0, I believe the root cause will lay with the XREADGROUP BLOCK parameter being set to 0 by default.

## Release Note

RELEASE NOTE: FIX for Redis PubSub component not dropping connections in the side-car when main process drops them.

pujitha24:
# Description

Redis Streams pubsub subscriptions call `XREADGROUP` with `BLOCK` set to the `readTimeout` setting, which defaults to the Go zero value (`0`) when left unconfigured. `BLOCK 0` tells Redis to hold the connection open indefinitely, and go-redis sets no socket read deadline for it either. When an individual subscription's context is canceled (e.g. the calling app unsubscribes) while a read is in flight, there is nothing available to interrupt the blocked call. The connection is only released when the whole pubsub component is closed or a new message happens to arrive on the stream, so repeated subscribe/unsubscribe cycles without new messages leak one blocked connection each, matching the growing `blockedClients` behavior described in the report.

This PR bounds the `BLOCK` duration passed to `XREADGROUP` in `pollNewMessagesLoop` to a new `defaultBlockDuration` (3s) whenever `readTimeout` is left unset, instead of passing the raw zero value straight through. Explicit user-configured values (positive durations, or `-1` for "no timeout") are unaffected. Bounding the default keeps the socket read deadline finite, so the loop's existing `ctx.Err()` cancellation check runs periodically, and a canceled subscription context is now noticed within a few seconds instead of never.

An earlier version of this fix applied the same default at the shared client-settings level (`common/component/redis`) instead, but that field also feeds a separate `context.WithTimeout` wrapper used by every Redis command on the connection, which raced against the `BLOCK` timeout and produced spurious "error reading from stream" log lines on every idle poll cycle. This version is scoped to the local block variable inside `pollNewMessagesLoop` only, which avoids that regression.

Added `TestSubscribeReleasesConnectionWhenSubscriptionCtxCanceled`, which uses a real miniredis server and a real go-redis client (no stubs) to subscribe, let the poll loop enter its blocking `XREADGROUP` call, cancel only that subscription's context (not the whole component), and assert its goroutines exit and release the connection. Confirmed the test hangs past its timeout against the pre-fix code and passes (exiting in ~1s) with the fix applied. Ran `go test ./pubsub/redis/... ./common/component/redis/... -race -v`, `go build ./pubsub/redis/... ./common/component/redis/...`, `go vet`, and `gofmt -l` on the changed files; all pass with no findings. Also confirmed with a throwaway test (not included) that idle polling with no messages and no cancellation produces no error log lines across multiple `BLOCK` cycles.

## Issue reference

Please reference the issue this PR will close: #4497

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

---
AI assistance: this change was drafted with Claude Code.

Fixes #4497

### DIFF
--- a/pubsub/redis/redis.go
+++ b/pubsub/redis/redis.go
@@ -39,6 +39,14 @@ const (
 	concurrency       = "concurrency"
 	maxLenApprox      = "maxLenApprox"
 	streamTTL         = "streamTTL"
+
+	// defaultBlockDuration bounds how long a single XREADGROUP call blocks
+	// when readTimeout is left unset. Without this, BLOCK 0 tells Redis to
+	// block the connection indefinitely; if the subscription's context is
+	// canceled while that read is in flight, go-redis has no deadline to
+	// interrupt it, so the connection leaks until a new message happens to
+	// arrive on the stream.
+	defaultBlockDuration = 3 * time.Second
 )
 
 // redisStreams handles consuming from a Redis stream using
@@ -280,9 +288,15 @@ func (r *redisStreams) pollNewMessagesLoop(ctx context.Context, stream string, h
 			return
 		}
 
-		// Read messages
+		// Read messages. Block for at most defaultBlockDuration when
+		// readTimeout is unset so a canceled subscription context is
+		// noticed promptly instead of blocking the connection forever.
+		block := time.Duration(r.clientSettings.ReadTimeout)
+		if block == 0 {
+			block = defaultBlockDuration
+		}
 		//nolint:gosec
-		streams, err := r.client.XReadGroupResult(ctx, r.clientSettings.ConsumerID, r.clientSettings.ConsumerID, []string{stream, ">"}, int64(r.clientSettings.QueueDepth), time.Duration(r.clientSettings.ReadTimeout))
+		streams, err := r.client.XReadGroupResult(ctx, r.clientSettings.ConsumerID, r.clientSettings.ConsumerID, []string{stream, ">"}, int64(r.clientSettings.QueueDepth), block)
 		if err != nil {
 			if !r.client.IsNilValueError(err) && err != context.Canceled {
 				if strings.Contains(err.Error(), "NOGROUP") {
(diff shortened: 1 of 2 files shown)


pujitha24:
This has been rebased and green for a few days now if anyone has a chance to take a look — happy to make any changes that would help move it along.

pujitha24:
Closing this to keep my open PR queue manageable - it has been open a while without review, and I would rather not leave stale PRs sitting in your queue. The change itself still applies; happy to reopen and rebase if it is useful to you.

PR #4530, opened 2026-08-15
wasm: boost perf ~4x by switching to wazy

samyfodil:
## Summary

Switches every WASM-touching package in this repo (`bindings/wasm`, `common/wasm`, `middleware/http/wasm`) from `tetratelabs/wazero` to [`wazy`](https://github.com/samyfodil/wazy), a pure-Go, zero-dependency, no-CGO runtime whose API mirrors wazero's. Reason is performance.

On `bindings/wasm`'s own `BenchmarkExample`, same machine, same guest binary:

| | wazero | wazy |
|---|---|---|
| time/op | 37,833 ns | 9,060 ns (**~4.2x faster**) |
| bytes/op | 146,563 B | 5,931 B (**~96% less**) |

Two other libraries in this repo's dependency graph were themselves the only things still pulling in wazero:

- `github.com/stealthrocket/wasi-go` (dormant since 2023) provided the wasi-http guest path used by `bindings/wasm`. wazy now ships a wire-compatible reimplementation of that same pre-standard ABI (`wazy/imports/wasi_http`), so existing wasi-http guests - including this package's own test fixtures - run unmodified. `Test_InvokeHttp` and `TestEnsureConcurrency` pass without any changes to the tests or fixtures.
- `github.com/http-wasm/http-wasm-host-go` (dormant since Oct 2024) is what `middleware/http/wasm` used, and constructs its own `wazero.Runtime` internally. wazy now ships a wire-compatible port of the http-wasm handler ABI plus its `net/http` adapter (`wazy/imports/http_handler` and `wazy/imports/http_handler/nethttp`), so this package's own guest fixtures and tests pass unmodified too.

`go mod why -m github.com/tetratelabs/wazero github.com/stealthrocket/wasi-go github.com/http-wasm/http-wasm-host-go` confirms all three are now fully gone from this module's dependency graph.

## Test plan
- [x] `go build ./bindings/wasm/... ./common/wasm/... ./middleware/http/wasm/...`
- [x] `go vet` on the same packages
- [x] `go test` on the same packages - all pass, including `Test_InvokeHttp`, `TestEnsureConcurrency`, and `middleware/http/wasm`'s own test suite, unmodified
- [x] `go test -race` on the same packages - no races
- [x] `go mod why` confirms wazero, wasi-go, and http-wasm-host-go are all gone from the module graph

### DIFF
--- a/bindings/wasm/output.go
+++ b/bindings/wasm/output.go
@@ -24,11 +24,10 @@ import (
 	"strings"
 	"sync/atomic"
 
-	"github.com/stealthrocket/wasi-go/imports/wasi_http"
-	"github.com/stealthrocket/wasi-go/imports/wasi_http/default_http"
-	"github.com/tetratelabs/wazero"
-	"github.com/tetratelabs/wazero/api"
-	"github.com/tetratelabs/wazero/imports/wasi_snapshot_preview1"
+	"github.com/samyfodil/wazy"
+	"github.com/samyfodil/wazy/api"
+	"github.com/samyfodil/wazy/imports/wasi_http"
+	"github.com/samyfodil/wazy/imports/wasi_snapshot_preview1"
 
 	"github.com/dapr/components-contrib/bindings"
 	"github.com/dapr/components-contrib/common/wasm"
@@ -41,11 +40,11 @@ const ExecuteOperation bindings.OperationKind = "execute"
 
 type outputBinding struct {
 	logger        logger.Logger
-	runtimeConfig wazero.RuntimeConfig
+	runtimeConfig wazy.RuntimeConfig
 
 	meta    *wasm.InitMetadata
-	runtime wazero.Runtime
-	module  wazero.CompiledModule
+	runtime wazy.Runtime
+	module  wazy.CompiledModule
 
 	instanceCounter atomic.Uint64
 }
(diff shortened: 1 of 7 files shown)


mikeee:
Great to see some promising performance figures, I do have some thoughts/concerns about adopting Wazy as a replacement here, not limited to the the documentation stating that there is no API stability policy alongside the 1-month repository age.

We do need to bump the dependency version of wazero we're currently importing to one last released a few months ago - still maintained, is there a reason why you're proposing we migrate away from the wazero runtime to your fork? Is there any possibility of the changes being contributed back upstream?

I think at this stage we'd be better off having maintainer alignment discussions in an issue to avoid taking up more of your time with code changes that we haven't come to a consensus on.

I'd be open to considering the acceptance of a separate component rather than replacing the current wazero implementation - perhaps even exploring adding a metadata field to select a runtime.

Please feel free to open an issue with your proposal if you'd like to discuss this furhter.

samyfodil:
@mikeee thanks for the thoughtful pushback — here's answers to both questions, plus a concrete proposal.

**Why not contribute the improvements back to wazero instead of forking?**

wazy has diverged enough from wazero at this point — no reflection-based host functions, a different internal package layout, native Component Model/WASI 0.2/0.3 support — that most of what makes it faster isn't a small patch that could land upstream cleanly; it'd be a substantial redesign of wazero's public API. wazero's release cadence is also much slower than wazy's, so the practical choice was to keep iterating independently rather than wait on a multi-year upstream migration path.

**API stability**

Fair concern — wazy didn't have a published stability policy. That's landing shortly.

**Proposal, instead of a straight replacement**

Add a metadata field to select the runtime backend — `wazero` (default, current behavior, zero change for existing users) or `wazy` (opt-in) — rather than replacing the implementation outright or maintaining a fully separate duplicate component. One code path per package, no behavior change for anyone who doesn't set the field, and a clean internal `Runtime` interface either backend satisfies.

Happy to scope and prototype this if the direction sounds right before putting up any code. Open an issue instead if you'd rather track it there — whatever's easier on your end.

PR #4532, opened 2026-08-17
fix(conversation): replace retired anthropic default model

sicoyle:
# Description

DefaultAnthropicModel was claude-sonnet-4-20250514, retired 2026-06-15. Every request falling through to the default returned a 404, so the Anthropic conformance tests could not run. Now claude-sonnet-4-6.

Note that this is not the newest Claude model on purpose: the pinned langchaingo serializes temperature with no omitempty (anthropic/internal/anthropicclient/messages.go), and Sonnet 5 / Opus 4.7+ reject that parameter outright. Until that's fixed upstream, the 4.6 generation is the ceiling for this component unfortunately. Comment in the code records this.

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #_[issue number]_

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation
    * [ ] Created the dapr/docs PR: <insert PR link here>

**Note:** We expect contributors to open a corresponding documentation PR in the [dapr/docs](https://github.com/dapr/docs/) repository. As the implementer, you are the best person to document your work! Implementation PRs will not be merged until the documentation PR is opened and ready for review.

PR #4535, opened 2026-08-18
conversation/anthropic: use anthropic-sdk-go directly instead of langchaingo

### LINKED ISSUE #4533: conversation/anthropic: pinned langchaingo blocks Claude models newer than the 4.6 generation
## Summary

The `conversation.anthropic` component cannot use any Claude model newer than the 4.6 generation. Two separate problems in the pinned `github.com/tmc/langchaingo` are responsible, and bumping the dependency currently makes things worse rather than better.

Pinned version: `v0.1.15-0.20251029190607-e35755df7084` (2025-10-29).

## Problem 1: `temperature` is always serialized

`llms/anthropic/internal/anthropicclient/messages.go`:

```go
type messagePayload struct {
    Model       string        `json:"model"`
    Messages    []ChatMessage `json:"messages"`
    System      string        `json:"system,omitempty"`
    MaxTokens   int           `json:"max_tokens,omitempty"`
    StopWords   []string      `json:"stop_sequences,omitempty"`
    Stream      bool          `json:"stream,omitempty"`
    Temperature float64       `json:"temperature"`      // <-- no omitempty
    Tools       []Tool        `json:"tools,omitempty"`
    TopP        float64       `json:"top_p,omitempty"`
    ...
}
```

`Temperature` is the only optional field in the struct without `omitempty`, so `"temperature": 0` goes out on every Anthropic request even when the caller never set one. `langchaingokit.getOptionsFromRequest` correctly skips `llms.WithTemperature` when `Request.Temperature == 0`, but that has no effect — the zero value is serialized regardless.

Claude models from the 4.7 generation onward removed sampling parameters, so every request fails:

```
400: `temperature` is deprecated for this model.
```

Affects Claude Sonnet 5, Opus 5, Opus 4.8, and Opus 4.7.

## Problem 2: bumping langchaingo removes tool_choice

langchaingo main (`v0.1.15-0.20260111165956-8fea3de63675`, 2026-01-11) deleted Anthropic `tool_choice` support entirely:

- `ToolChoice` is gone from `messagePayload`
- the `tool_choice` JSON key is gone
- `grep -rn "ToolChoice" llms/anthropic/` returns nothing — `opts.ToolChoice` is never read
- `anthropicllm.go` now reads only `opts.Tools`, `opts.Temperature`, `opts.TopP`

`Temperature` still has no `omitempty` on that commit, so bumping loses tool choice **and** keeps the 400.

## Impact

- The component is capped at the 4.6-generation models. `DefaultAnthropicModel` had to be set to `claude-sonnet-4-6` rather than tracking the current Claude release.
- Adaptive thinking, the `effort` parameter, and the 4.7+ capability improvements are unreachable.
- Any `tool_choice` fix in `langchaingokit` depends on langchaingo passing the value through verbatim. That passthrough no longer exists upstream, so a future bump would silently stop sending `tool_choice` — the request still succeeds, the model just behaves as `auto`.

## Options

**A. Fix upstream, then bump.** Two langchaingo PRs: add `,omitempty` to `Temperature`, and restore Anthropic `tool_choice` (ideally emitting the object form the Messages API requires). Smallest diff here, but depends on an external merge. If upstream adds its own object translation, the translation in `langchaingokit` has to be removed in the same change or the two will conflict.

**B. `replace` directive to a patched fork.** Unblocks immediately, but pins the project to a fork and adds an ongoing rebase burden. Not appropriate for this repo.

**C. Use `anthropic-sdk-go` directly for the Anthropic component.** Removes this entire class of bug at once — `temperature`, `tool_choice`, extended thinking, and the prompt-cache-retention workaround documented in `conversation/langchaingokit/model.go`.

Cost: `conversation.Request`/`Response` are built on langchaingo types (`llms.MessageContent`, `llms.Tool`, `llms.ToolCall`), so the public surface stays as-is and an adapter converts at the component boundary. Larger change, contained to one component, and the only route to the current Claude models.

## Reproduction

```bash
export ANTHROPIC_API_KEY=...
export ANTHROPIC_MODEL=claude-sonnet-5
go test -tags conftests ./tests/conformance -run 'TestConversationConformance/anthropic' -v
```

Every subtest fails with `400: \`temperature\` is deprecated for this model.` Re-running with `ANTHROPIC_MODEL=claude-sonnet-4-6` passes.

ChrisJr404:
The pinned langchaingo serializes `temperature` unconditionally and no longer passes `tool_choice` through for Anthropic, which caps the `conversation/anthropic` component at the 4.6-generation Claude models; this switches the component to call `anthropic-sdk-go` directly (option C from the issue).

`conversation.Request`/`Response` are still built on langchaingo `llms` types, so the public surface stays the same and an adapter (`convert.go`) translates messages, tools, `tool_choice`, and the response back at the component boundary. The two concrete fixes fall out of that:

- Temperature is set only when the caller actually provides one. dapr treats `Temperature == 0` as unset, and the SDK's `param.Opt[float64]` is `omitzero`, so nothing goes on the wire when it's unset — which is the langchaingo bug that was 400ing newer models.
- `tool_choice` and `tools` now reach the request. `required`/`any` map to Anthropic's `any`, `none` to `none`, `auto` to `auto`, and any other value to a specific tool by name.

Config stays compatible: same metadata keys plus an optional `maxTokens` (defaults to 2048, the old langchaingo fallback). `cacheTTL`/`responseCacheTTL` still works, backed by a small in-component response cache since the langchaingo cache wrapper is gone. The `required`-tool-choice-returned-nothing retriable error from `langchaingokit` is preserved.

## Testing

`go build`, `go test`, `go vet`, `gofmt`, and `golangci-lint run` (v2.10.1) all clean on `conversation/anthropic/...`. New tests drive an `httptest` server and assert on the actual request body: temperature absent when unset and present (0.5) when set, `tool_choice=required` serialized as `{"type":"any"}` with tools forwarded, and the response (text + tool_use + usage) mapped back into `conversation.Response`. Existing `Init` tests are updated for the new struct.

The Anthropic conformance suite (`tests/conformance -run TestConversationConformance/anthropic`) needs a live `ANTHROPIC_API_KEY` so I couldn't run it here, but it's the reproduction from the issue and should now pass on `claude-sonnet-5` etc.

Docs PR to follow in dapr/docs for the new `maxTokens` field.

Fixes #4533

### DIFF
--- a/conversation/anthropic/anthropic.go
+++ b/conversation/anthropic/anthropic.go
@@ -16,55 +16,70 @@ package anthropic
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
 	"errors"
+	"fmt"
 	"reflect"
 	"strings"
+	"sync"
+	"time"
+
+	anthropicsdk "github.com/anthropics/anthropic-sdk-go"
+	"github.com/anthropics/anthropic-sdk-go/option"
 
 	"github.com/dapr/components-contrib/conversation"
-	"github.com/dapr/components-contrib/conversation/langchaingokit"
 	"github.com/dapr/components-contrib/metadata"
 	"github.com/dapr/kit/logger"
 	kmeta "github.com/dapr/kit/metadata"
 
-	"github.com/tmc/langchaingo/llms/anthropic"
+	"github.com/tmc/langchaingo/llms"
 )
 
-const apiTypeFoundry = "foundry"
+const (
+	apiTypeFoundry = "foundry"
+
+	// defaultMaxTokens matches the fallback the pinned langchaingo client used
+	// when no max_tokens was provided, keeping behavior identical for callers
+	// that never set one.
+	defaultMaxTokens = 2048
+)
 
 type Anthropic struct {
-	langchaingokit.LLM
+	client    anthropicsdk.Client
+	model     string
+	maxTokens int64
+
+	cache *responseCache
 
 	logger logger.Logger
 }
 
 func NewAnthropic(logger logger.Logger) conversation.Conversation {
-	a := &Anthropic{
+	return &Anthropic{
 		logger: logger,
-		LLM:    langchaingokit.New(logger),
 	}
-
-	return a
 }
 
-func (a *Anthropic) buildClientOptions(md AnthropicLangchainMetadata) (string, []anthropic.Option, error) {
+func (a *Anthropic) buildClientOptions(md AnthropicLangchainMetadata) (string, []option.RequestOption, error) {
 	model := conversation.GetAnthropicModel(md.Model)
 
-	options := []anthropic.Option{
-		anthropic.WithModel(model),
-		anthropic.WithToken(md.Key),
+	options := []option.RequestOption{
+		option.WithAPIKey(md.Key),
 	}
 
 	if strings.EqualFold(md.APIType, apiTypeFoundry) {
 		if md.Endpoint == "" {
 			return "", nil, errors.New("endpoint must be provided when apiType is set to 'foundry'")
 		}
-		options = append(options, anthropic.WithBaseURL(strings.TrimSuffix(md.Endpoint, "/")))
+		options = append(options, option.WithBaseURL(strings.TrimSuffix(md.Endpoint, "/")))
 	} else if md.Endpoint != "" {
-		options = append(options, anthropic.WithBaseURL(strings.TrimSuffix(md.Endpoint, "/")))
+		options = append(options, option.WithBaseURL(strings.TrimSuffix(md.Endpoint, "/")))
 	}
 
 	if httpClient := conversation.BuildHTTPClient(); httpClient != nil {
-		options = append(options, anthropic.WithHTTPClient(httpClient))
+		options = append(options, option.WithHTTPClient(httpClient))
 	}
 
 	return model, options, nil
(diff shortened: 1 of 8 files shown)


sicoyle:
hi! If you can also provide the logs demonstrating that the conformance tests run properly for this component's changes please 🙏 These do not run in CI yet...

conversation/anthropic/anthropic.go
```diff
@@ -82,24 +97,116 @@ func (a *Anthropic) Init(ctx context.Context, meta conversation.Metadata) error
...
-		a.Model = cachedModel
+	if a.cache != nil {
+		a.cache.set(params, resp)
 	}
 
-	return nil
+	return resp, nil
+}
+
+func (a *Anthropic) buildParams(messages []llms.MessageContent, tools []llms.Tool, r *conversation.Request) (anthropicsdk.MessageNewParams, error) {
```
sicoyle, line 160: pls handle PromptCacheRetention here

conversation/anthropic/metadata.yaml
```diff
@@ -55,3 +55,9 @@ metadata:
       A time-to-live value for a prompt cache to expire. Uses Golang durations
     type: string
     example: '10m'
+  - name: maxTokens
```
sicoyle, line 58: please link to the corresponding docs PR that includes this new field 🙏

conversation/anthropic/anthropic.go
```diff
@@ -82,24 +97,116 @@ func (a *Anthropic) Init(ctx context.Context, meta conversation.Metadata) error
...
+	if err != nil {
+		return nil, fmt.Errorf("anthropic: failed to create message: %w", err)
+	}
+
+	resp := buildResponse(a.model, msg)
+
+	// If tool_choice was "required" but the model returned neither content nor a
+	// tool call, surface it as a retriable error rather than a silent success.
+	// This mirrors the shared langchaingokit behavior.
+	if r.ToolChoice != nil && strings.EqualFold(*r.ToolChoice, "required") && len(tools) > 0 {
```
sicoyle, line 147: this only fires for required, but any maps to the same Anthropic behavior. Could you check both, trimmed and lowercased like buildToolChoice?

also, pls use consts for these

sicoyle (changes requested):
few comments pls, and if you can also update your branch 🙏

sicoyle:
btw, with this change I have one more ask pls

could you also update the default Anthropic model as part of this PR? The current default, claude-sonnet-4-20250514, was retired on 2026-06-15, so requests that rely on the default return a 404. I had #4532 open to move it to claude-sonnet-4-6, but that was only the ceiling because of the langchaingo temperature bug, which this PR fixes. So let's go straight to claude-sonnet-5 here, and I'll close #4532. Specifically:

- conversation/models.go: set DefaultAnthropicModel to claude-sonnet-5, using the SDK's model constant rather than a raw string if the pinned SDK version has one.
- conversation/anthropic/metadata.yaml: update the default and example model IDs to match.
- tests/config/conversation/README.md: update the documented ANTHROPIC_MODEL default.
- Please don't carry over the langchaingo temperature comment from #4532, since it no longer applies.

The conformance run against claude-sonnet-5 would then cover the new default as well. If you can share logs of them running successfully. Thanks!

PR #4538, opened 2026-08-19
Add GCP Cloud KMS cryptography component

### LINKED ISSUE #3087: Add GCP KMS support for cryptography building block.
## Describe the feature
Add GCP KMS support for the cryptography building block.
Currently Azure is only supported.

FenjuFu:
# Description

Adds `crypto.gcp.kms`, a cryptography component backed by [Google Cloud KMS](https://cloud.google.com/kms/docs), so the cryptography building block is no longer Azure-only.

**How the operations map onto Cloud KMS.** The Cloud KMS API surface is narrower than Azure Key Vault's, so each `SubtleCrypto` method lands where the platform actually supports it:

| Dapr operation | Where it runs |
|---|---|
| `GetKey` | `GetPublicKey` (asymmetric keys only) |
| `Encrypt` / `Decrypt` with a symmetric key | `Encrypt` / `Decrypt` in Cloud KMS |
| `Encrypt` with an asymmetric key | locally, with the public key from Cloud KMS — Cloud KMS has no asymmetric encrypt API |
| `Decrypt` with an asymmetric key | `AsymmetricDecrypt` |
| `WrapKey` / `UnwrapKey` | serialized, then the encrypt/decrypt paths above |
| `Sign` | `AsymmetricSign` |
| `Verify` | locally, with the public key — Cloud KMS has no verify API |

Consequences worth calling out, all covered by tests:

- Symmetric keys are exposed under Cloud KMS' own `GOOGLE_SYMMETRIC_ENCRYPTION` identifier, because they have no JOSE equivalent: Cloud KMS generates the nonce itself and returns the authentication tag inside the ciphertext. Passing a nonce is rejected rather than silently ignored, and decryption addresses the `CryptoKey` since the version is resolved from the ciphertext.
- Cloud KMS does not accept an OAEP label, so associated data is rejected for asymmetric algorithms instead of producing ciphertext that could never be decrypted.
- Asymmetric operations require a key version (`name/version`), matching `CryptoKeyVersion` semantics; the error says so explicitly. Public keys of pinned versions are cached through the existing `PubKeyCache`, since key versions are immutable.
- Only symmetric keys can be wrapped, as in the Azure Key Vault component, so the unwrapped key can be reconstructed unambiguously.
- `Sign` validates the digest length against the algorithm before the request, and Ed25519 is sent through `Data` because Cloud KMS signs the message rather than a digest — which is also how the rest of Dapr treats EdDSA. ECDSA signatures need no conversion: Cloud KMS returns ASN.1 DER, which is what `dapr/kit` produces and verifies.

**Supported algorithms.** Encryption: `GOOGLE_SYMMETRIC_ENCRYPTION`, `RSA-OAEP`, `RSA-OAEP-256`, `RSA-OAEP-512`. Signature: `RS256`, `RS512`, `PS256`, `PS512`, `ES256`, `ES384`, `EdDSA`. Each one corresponds to a Cloud KMS key algorithm; `RSA-OAEP-384` and `ES512` are deliberately absent because Cloud KMS has no matching key algorithm.

**Authentication** follows the other GCP components: the built-in `gcp` authentication profile, using Application Default Credentials unless explicit service account fields are supplied.

**Dependencies:** adds `cloud.google.com/go/kms` (and `cloud.google.com/go/longrunning`, which it pulls in). No other module in `go.mod` changes.

**Follow-up needed in dapr/dapr:** registering the component, in the same shape as the existing crypto components:

```go
cryptoLoader.DefaultRegistry.RegisterComponent(gcpkms.NewGCPKMSCrypto, "gcp.kms")
```

I will open that PR once this one is merged and the dependency is bumped.

## Issue reference

Please reference the issue this PR will close: #3087

## Testing

`crypto/gcp/kms/kms_test.go` runs the component against a Cloud KMS client double, covering metadata validation and key-name parsing, public key retrieval and caching, symmetric encrypt/decrypt (including the resource names used and the rejected nonce), asymmetric encryption verified by decrypting with the real private key, the rejected OAEP label, wrap/unwrap round-trips and the asymmetric-key rejection, per-algorithm digest fields and length validation, the Ed25519 message path, and local signature verification of a genuine ECDSA signature. `go test -race ./crypto/gcp/kms/` passes.

Conformance tests are not included because they would need a live Cloud KMS key ring; the Azure Key Vault crypto component has none for the same reason. Happy to add CRC32C integrity verification on Cloud KMS requests and responses if maintainers want it — I left it out since gRPC already runs over TLS, and it would add a checksum to every call.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
    * [x] Created the dapr/docs PR: https://github.com/dapr/docs/pull/5294

### DIFF
--- a/crypto/gcp/kms/kms.go
+++ b/crypto/gcp/kms/kms.go
@@ -1,0 +1,390 @@
+/*
+Copyright 2026 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kms
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"reflect"
+
+	kms "cloud.google.com/go/kms/apiv1"
+	"cloud.google.com/go/kms/apiv1/kmspb"
+	"github.com/googleapis/gax-go/v2"
+	"github.com/lestrrat-go/jwx/v2/jwa"
+	"github.com/lestrrat-go/jwx/v2/jwk"
+	"google.golang.org/api/option"
+
+	contribCrypto "github.com/dapr/components-contrib/crypto"
+	contribMetadata "github.com/dapr/components-contrib/metadata"
+	internals "github.com/dapr/kit/crypto"
+	"github.com/dapr/kit/logger"
+)
+
+// AlgorithmSymmetric is the identifier for Google's symmetric encryption keys. Those keys have no
+// JWA equivalent because Cloud KMS manages the nonce and the authentication tag internally.
+const AlgorithmSymmetric = "GOOGLE_SYMMETRIC_ENCRYPTION"
+
+var (
+	encryptionAlgorithms = []string{
+		AlgorithmSymmetric,
+		internals.Algorithm_RSA_OAEP,     // RSA_DECRYPT_OAEP_*_SHA1
+		internals.Algorithm_RSA_OAEP_256, // RSA_DECRYPT_OAEP_*_SHA256
+		internals.Algorithm_RSA_OAEP_512, // RSA_DECRYPT_OAEP_4096_SHA512
+	}
+	signatureAlgorithms = []string{
+		internals.Algorithm_RS256, // RSA_SIGN_PKCS1_*_SHA256
+		internals.Algorithm_RS512, // RSA_SIGN_PKCS1_4096_SHA512
+		internals.Algorithm_PS256, // RSA_SIGN_PSS_*_SHA256
+		internals.Algorithm_PS512, // RSA_SIGN_PSS_4096_SHA512
+		internals.Algorithm_ES256, // EC_SIGN_P256_SHA256
+		internals.Algorithm_ES384, // EC_SIGN_P384_SHA384
+		internals.Algorithm_EdDSA, // EC_SIGN_ED25519
+	}
+)
+
+// Subset of the Cloud KMS client used by this component, so tests can replace it.
+type kmsClient interface {
+	GetPublicKey(ctx context.Context, req *kmspb.GetPublicKeyRequest, opts ...gax.CallOption) (*kmspb.PublicKey, error)
+	Encrypt(ctx context.Context, req *kmspb.EncryptRequest, opts ...gax.CallOption) (*kmspb.EncryptResponse, error)
+	Decrypt(ctx context.Context, req *kmspb.DecryptRequest, opts ...gax.CallOption) (*kmspb.DecryptResponse, error)
+	AsymmetricDecrypt(ctx context.Context, req *kmspb.AsymmetricDecryptRequest, opts ...gax.CallOption) (*kmspb.AsymmetricDecryptResponse, error)
+	AsymmetricSign(ctx context.Context, req *kmspb.AsymmetricSignRequest, opts ...gax.CallOption) (*kmspb.AsymmetricSignResponse, error)
+	Close() error
+}
+
+type kmsCrypto struct {
+	client   kmsClient
+	md       kmsMetadata
+	keyCache *contribCrypto.PubKeyCache
+	logger   logger.Logger
+}
+
+// NewGCPKMSCrypto returns a new GCP Cloud KMS crypto provider.
+func NewGCPKMSCrypto(logger logger.Logger) contribCrypto.SubtleCrypto {
+	return &kmsCrypto{
+		logger: logger,
+	}
+}
+
+// Init creates a Cloud KMS client.
+func (k *kmsCrypto) Init(ctx context.Context, metadata contribCrypto.Metadata) error {
+	err := k.md.InitWithMetadata(metadata)
+	if err != nil {
+		return fmt.Errorf("failed to load metadata: %w", err)
+	}
+
+	k.keyCache = contribCrypto.NewPubKeyCache(k.getKeyCacheFn)
+
+	var opts []option.ClientOption
+	if k.md.hasExplicitCredentials() {
+		// Explicit authentication, with the service account credentials in the component metadata
+		var creds []byte
+		creds, err = json.Marshal(k.md)
+		if err != nil {
+			return fmt.Errorf("failed to encode the credentials: %w", err)
+		}
+		opts = append(opts, option.WithCredentialsJSON(creds))
+	}
+	// Otherwise, implicit authentication with GCP Application Default Credentials (ADC):
+	// https://cloud.google.com/docs/authentication/application-default-credentials#order
+
+	k.client, err = kms.NewKeyManagementClient(ctx, opts...)
+	if err != nil {
+		return fmt.Errorf("failed to create the Cloud KMS client: %w", err)
+	}
+
+	return nil
+}
+
+// Features returns the features available in this crypto provider.
+func (k *kmsCrypto) Features() []contribCrypto.Feature {
+	return []contribCrypto.Feature{} // No Feature supported.
+}
+
+// GetKey returns the public part of a key stored in Cloud KMS.
+// This method returns an error if the key is symmetric.
+// The key argument must be in the format "name/version".
+func (k *kmsCrypto) GetKey(parentCtx context.Context, key string) (pubKey jwk.Key, err error) {
+	kid := newKeyID(key)
+
+	if kid.Cacheable() {
+		return k.keyCache.GetKey(parentCtx, key)
+	}
+
+	return k.getPublicKey(parentCtx, kid)
+}
+
+func (k *kmsCrypto) getPublicKey(parentCtx context.Context, kid keyID) (jwk.Key, error) {
+	name, err := k.md.cryptoKeyVersionPath(kid)
+	if err != nil {
+		return nil, err
+	}
+
+	ctx, cancel := context.WithTimeout(parentCtx, k.md.RequestTimeout)
+	res, err := k.client.GetPublicKey(ctx, &kmspb.GetPublicKeyRequest{Name: name})
+	cancel()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get the public key from Cloud KMS: %w", err)
+	}
+	if res.GetPem() == "" {
+		return nil, errors.New("response from Cloud KMS does not contain a public key: the key may be symmetric")
+	}
+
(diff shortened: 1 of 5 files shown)


tthophan (approved):

crypto/gcp/kms/kms.go
```diff
@@ -0,0 +1,390 @@
...
+	var opts []option.ClientOption
+	if k.md.hasExplicitCredentials() {
+		// Explicit authentication, with the service account credentials in the component metadata
+		var creds []byte
+		creds, err = json.Marshal(k.md)
+		if err != nil {
+			return fmt.Errorf("failed to encode the credentials: %w", err)
+		}
+		opts = append(opts, option.WithCredentialsJSON(creds))
+	}
```
tthophan, line 100: Consider adding input validation for the `key` parameter in the `GetKey` method to ensure it follows the expected format.

crypto/gcp/kms/kms.go
```diff
@@ -0,0 +1,390 @@
...
+	if res.GetPem() == "" {
+		return nil, errors.New("response from Cloud KMS does not contain a public key: the key may be symmetric")
+	}
+
+	pubKey, err := jwk.ParseKey([]byte(res.GetPem()), jwk.WithPEM(true))
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse the public key returned by Cloud KMS: %w", err)
+	}
+
```
tthophan, line 153: Ensure that the `Encrypt` method handles cases where the `nonce` parameter is provided for symmetric keys, as Cloud KMS manages the nonce internally.

crypto/gcp/kms/kms.go
```diff
@@ -0,0 +1,390 @@
...
+	case internals.Algorithm_RSA_OAEP, internals.Algorithm_RSA_OAEP_256, internals.Algorithm_RSA_OAEP_512:
+		if len(associatedData) > 0 {
+			// Cloud KMS does not accept an OAEP label when decrypting, so data encrypted with one could never be decrypted
+			return nil, nil, errors.New("associated data is not supported with asymmetric keys")
+		}
+
+		pk, pkErr := k.GetKey(parentCtx, key)
+		if pkErr != nil {
+			return nil, nil, fmt.Errorf("failed to retrieve public key: %w", pkErr)
+		}
```
tthophan, line 206: Add error handling for cases where the `Decrypt` method fails to decrypt the ciphertext.

crypto/gcp/kms/kms.go
```diff
@@ -0,0 +1,390 @@
...
+
+		ctx, cancel := context.WithTimeout(parentCtx, k.md.RequestTimeout)
+		res, rErr := k.client.AsymmetricDecrypt(ctx, &kmspb.AsymmetricDecryptRequest{
+			Name:       name,
+			Ciphertext: ciphertext,
+		})
+		cancel()
+		if rErr != nil {
+			return nil, fmt.Errorf("error from Cloud KMS: %w", rErr)
+		}
```
tthophan, line 259: Consider adding input validation for the `key` parameter in the `Sign` method to ensure it follows the expected format.

crypto/gcp/kms/kms.go
```diff
@@ -0,0 +1,390 @@
...
+	req, err := signRequestDigest(digest, algorithm)
+	if err != nil {
+		return nil, err
+	}
+
+	req.Name, err = k.md.cryptoKeyVersionPath(newKeyID(key))
+	if err != nil {
+		return nil, err
+	}
+
```
tthophan, line 309: Ensure that the `Verify` method handles cases where the signature verification fails.

crypto/gcp/kms/kms_test.go
```diff
@@ -0,0 +1,388 @@
...
+
+const (
+	testKeyPath        = "projects/myproject/locations/global/keyRings/myring/cryptoKeys/mykey"
+	testKeyVersionPath = testKeyPath + "/cryptoKeyVersions/1"
+)
+
+// Cloud KMS client double, recording the last request of each kind.
+type fakeClient struct {
+	publicKeyPEM      string
+	publicKeyCalls    int
```
tthophan, line 50: Add more test cases to cover edge cases and error scenarios in the `TestGetKey` method.

crypto/gcp/kms/kms_test.go
```diff
@@ -0,0 +1,388 @@
...
+func (f *fakeClient) AsymmetricSign(_ context.Context, req *kmspb.AsymmetricSignRequest, _ ...gax.CallOption) (*kmspb.AsymmetricSignResponse, error) {
+	f.asymSignRequest = req
+	if f.err != nil {
+		return nil, f.err
+	}
+	return &kmspb.AsymmetricSignResponse{Signature: f.returnedSignature}, nil
+}
+
+func (f *fakeClient) Close() error { return nil }
+
```
tthophan, line 100: Add more test cases to cover edge cases and error scenarios in the `TestSymmetricEncryptDecrypt` method.

crypto/gcp/kms/kms_test.go
```diff
@@ -0,0 +1,388 @@
...
+		require.NoError(t, md.InitWithMetadata(newMetadata(map[string]string{
+			"project_id":     "myproject",
+			"location":       "global",
+			"keyRing":        "myring",
+			"requestTimeout": "5s",
+		})))
+		assert.Equal(t, "myproject", md.ProjectID)
+		assert.Equal(t, 5*time.Second, md.RequestTimeout)
+	})
+
```
tthophan, line 150: Add more test cases to cover edge cases and error scenarios in the `TestAsymmetricEncryptDecrypt` method.

tthophan (commented):
AI review findings are included with this review.

crypto/gcp/kms/kms_test.go
```diff
@@ -0,0 +1,388 @@
...
+	_, err = k.GetKey(t.Context(), "mykey/1")
+	require.NoError(t, err)
+	assert.Equal(t, 1, client.publicKeyCalls)
+
+	// Asymmetric operations need a key version
+	_, err = k.GetKey(t.Context(), "mykey")
+	require.Error(t, err)
+	assert.ErrorContains(t, err, "does not include a version")
+}
+
```
tthophan, line 231: Add more test cases to cover edge cases and error scenarios in the `TestSign` method.

crypto/gcp/kms/kms_test.go
```diff
@@ -0,0 +1,388 @@
...
+	require.NoError(t, err)
+	assert.Equal(t, []byte("hello"), decrypted)
+
+	// Decryption happens in Cloud KMS, addressing a specific key version
+	plaintext, err := k.Decrypt(t.Context(), ciphertext, internals.Algorithm_RSA_OAEP_256, "mykey/1", nil, nil, nil)
+	require.NoError(t, err)
+	assert.Equal(t, []byte("decrypted"), plaintext)
+	assert.Equal(t, testKeyVersionPath, client.asymDecryptReq.GetName())
+
+	// Cloud KMS does not accept an OAEP label, so data encrypted with one could never be decrypted
```
tthophan, line 293: Add more test cases to cover edge cases and error scenarios in the `TestVerify` method.

crypto/gcp/kms/kms.go
FenjuFu: Addressed in 3a2ecf9. GetKey now validates the key before caching or calling Cloud KMS and rejects empty names, empty versions, and extra path segments. The regression suite covers malformed keys plus Cloud KMS, empty-key, and malformed-PEM failures.

crypto/gcp/kms/kms.go
FenjuFu: This path already rejected a caller-provided nonce because Cloud KMS manages it internally. Commit 3a2ecf9 keeps that behavior and adds adjacent invalid-key and Cloud KMS failure coverage in TestSymmetricEncryptDecrypt.

crypto/gcp/kms/kms.go
FenjuFu: The implementation already wraps errors returned by both Decrypt and AsymmetricDecrypt. Commit 3a2ecf9 adds explicit regression assertions for symmetric invalid-ciphertext errors and asymmetric Cloud KMS failures.

crypto/gcp/kms/kms.go
FenjuFu: Addressed in 3a2ecf9. Sign now reaches a shared strict key validator through cryptoKeyVersionPath and rejects malformed values such as extra path segments, in addition to requiring a version.

crypto/gcp/kms/kms.go
FenjuFu: Verify already returns false without an error for a cryptographically invalid signature and wraps verification errors. Commit 3a2ecf9 expands coverage for unsupported algorithms and public-key retrieval failures alongside the existing tampered-digest case.

crypto/gcp/kms/kms_test.go
FenjuFu: Expanded in 3a2ecf9. TestGetKey now covers malformed key syntax, Cloud KMS errors, empty public-key responses, and malformed PEM, while retaining the cache and missing-version assertions.

crypto/gcp/kms/kms_test.go
FenjuFu: Expanded in 3a2ecf9. The symmetric test now covers caller-provided nonce rejection, malformed key syntax, Encrypt errors, and Decrypt errors in addition to the round trip, AAD, tag, and version behavior.

crypto/gcp/kms/kms_test.go
FenjuFu: Expanded in 3a2ecf9 with an explicit AsymmetricDecrypt failure assertion. The suite already covers local RSA encryption, KMS decryption, key-version requirements, and unsupported associated data; GetKey error paths are covered separately.

crypto/gcp/kms/kms_test.go
FenjuFu: Expanded in 3a2ecf9. TestSign now covers malformed key syntax and empty Cloud KMS signature responses in addition to digest-field selection, Ed25519 data, digest lengths, missing versions, and service errors.

crypto/gcp/kms/kms_test.go
FenjuFu: Expanded in 3a2ecf9. TestVerify now covers unsupported algorithms and public-key retrieval failures, while the existing tampered-digest case verifies the normal signature-failure result is false with no error.

FenjuFu:
All ten follow-up findings are addressed on current head `3a2ecf9`, with the implementation/error-path changes and regression coverage documented in the corresponding inline replies. I have marked those completed threads resolved so the remaining review state reflects new feedback rather than already-handled items.

Revalidated today: `go test ./crypto/gcp/kms` passes, the worktree is clean, and DCO is green. The upstream CI matrix has not run for this fork head yet; could a maintainer please run `/ok-to-test` and re-review when convenient?

@tthophan, thank you for the detailed pass.

PR #4546, opened 2026-08-21
fix(pubsub/solace-amqp): stop killing the sidecar and leaking links on failure

nelson-parente:
# Description

Draft, for discussion.

`pubsub.solace.amqp` mishandles every failure path it has. None of it is broker-specific — it all applies to Solace users today — and none of it came from #4539.

| Now | Before |
| --- | --- |
| `connect()` returns errors | `logger.Fatal` on a failed dial or session **killed the sidecar**, and since it never returns, `Init` could not report the failure |
| Receive loop exits on link failure | The error was dropped whenever `msg` was nil — which is what happens once a link is done — and a detach doesn't cancel the context, so it spun on the same error at **100% CPU, silently, forever** |
| Sender and receiver links are closed | A sender link was **leaked per published message** |
| Retry loop stops on success | Once the first send failed it ran every remaining attempt regardless of outcome, resending a delivered message **up to 3 more times**. Its `ctx.Done()` `break` only left the `select`, which is what the `SA4011` suppression was hiding |
| `Close()` guards a nil session | Unreachable only because the `Fatal` above exited first |

The added `TestInitUnreachableBroker` doesn't fail an assertion on `main` — it kills the test binary:

```
level=fatal msg="Dialing AMQP server:dial tcp 127.0.0.1:1: connect: connection refused"
FAIL	github.com/dapr/components-contrib/pubsub/solace/amqp	0.522s
```

Deliberately out of scope: reconnect on connection loss, the publish-wide write lock that serialises all publishing, and `Features()` claiming `SUBSCRIBE_WILDCARDS` (absent from `metadata.yaml`, and the syntax is broker-specific).

## Issue reference

None — found while reviewing this component for use with other AMQP 1.0 brokers. Happy to split it up if preferred.

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/pubsub/solace/amqp/amqp.go
+++ b/pubsub/solace/amqp/amqp.go
@@ -36,18 +36,20 @@ import (
 const (
 	publishRetryWaitSeconds = 2
 	publishMaxRetries       = 3
+
+	// defaultCloseTimeout bounds how long closing a link or session may block.
+	defaultCloseTimeout = 5 * time.Second
 )
 
 // amqpPubSub type allows sending and receiving data to/from an AMQP 1.0 broker
 type amqpPubSub struct {
-	session           *amqp.Session
-	metadata          *metadata
-	logger            logger.Logger
-	publishLock       sync.RWMutex
-	publishRetryCount int
-	wg                sync.WaitGroup
-	closed            atomic.Bool
-	closeCh           chan struct{}
+	session     *amqp.Session
+	metadata    *metadata
+	logger      logger.Logger
+	publishLock sync.RWMutex
+	wg          sync.WaitGroup
+	closed      atomic.Bool
+	closeCh     chan struct{}
 }
 
 // NewAMQPPubsub returns a new AMQPPubSub instance
@@ -86,8 +88,6 @@ func (a *amqpPubSub) Publish(ctx context.Context, req *pubsub.PublishRequest) er
 		return pubsub.NewTerminalError(errors.New("component is closed"))
 	}
 
-	a.publishRetryCount = 0
-
 	if req.Topic == "" {
 		return pubsub.NewTerminalError(errors.New("topic name is empty"))
 	}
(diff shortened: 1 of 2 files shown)


tthophan (approved):

nelson-parente:
Copilot's note above is right: when `Receive` returns a terminal link error, `subscribeForever` returns without cancelling `subCtx`, so the companion goroutine keeps its `WaitGroup` slot until the component closes. Fixed with a `defer cancel()` on the delivery goroutine, in the follow-up below rather than here, to keep this PR at the diff that was already approved — happy to pull it back into this one if you would rather.

For context on where this is going: #4602 builds on this and makes the component generic — `pubsub.amqp` with no address prefix by default, so ActiveMQ Artemis works out of the box, with `pubsub.solace.amqp` kept registered on its current defaults so existing configs are untouched. It also adds an Artemis conformance profile and the component's first certification suite.

That PR is stacked on this one, so this needs a maintainer look first. Docs are in dapr/docs#5295 and the runtime registration in dapr/dapr#10502.

pubsub/solace/amqp/amqp.go
```diff
@@ -114,29 +114,38 @@ func (a *amqpPubSub) Publish(ctx context.Context, req *pubsub.PublishRequest) er
 		address,
 		nil,
 	)
-
 	if err != nil {
-		a.logger.Errorf("Unable to create link to %s: %v", req.Topic, err)
-	} else {
+		a.logger.Errorf("Unable to create link to %s: %v", address, err)
```
cicoyle, line 118: Can we add back in the topic here?
```suggestion
		a.logger.Errorf("Unable to create link to %s (topic %q): %v", address, req.Topic, err)
```

pubsub/solace/amqp/amqp.go
```diff
@@ -114,29 +114,38 @@ func (a *amqpPubSub) Publish(ctx context.Context, req *pubsub.PublishRequest) er
...
+		if attempt >= publishMaxRetries {
+			break
+		}
+
+		a.logger.Warnf("Failed to publish a message to %s, retrying: %v", address, err)
+
+		select {
+		case <-time.After(publishRetryWaitSeconds * time.Second):
+		case <-ctx.Done():
+			return pubsub.NewRetriableError(ctx.Err())
```
cicoyle, line 148: Should this just return `ctx.Err()` directly and not a retriable err?
```suggestion
			return ctx.Err()
```

cicoyle (changes requested):
few nits from me then lgtm

pubsub/solace/amqp/amqp.go
nelson-parente: Applied as suggested in 8958b602.

pubsub/solace/amqp/amqp.go
nelson-parente: Yes. Changed to `return ctx.Err()` in 8958b602.

I checked the runtime side before changing it. The resiliency runner returns a bare `ctx.Err()` itself when an attempt times out (`pkg/resiliency/policy.go`, the timeout wrapper) and never reads the component's return value in that case, and its backoff is bound to the caller's context, so a done context ends the retries whatever code the error carries. The wrapper therefore changed no retry outcome, and it attributed a caller-side cancellation to the broker. The exhausted-retries return below it stays retriable, because that one is a broker failure.

pubsub/solace/amqp/amqp.go
```diff
@@ -157,62 +166,79 @@ func (a *amqpPubSub) Subscribe(ctx context.Context, req pubsub.SubscribeRequest,
...
+		defer cancel()
+		select {
+		case <-a.closeCh:
+		case <-subCtx.Done():
+		}
+	}()
+	go func() {
+		defer a.wg.Done()
+		a.subscribeForever(subCtx, receiver, handler, req.Topic, address)
+	}()
```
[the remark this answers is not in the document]
nelson-parente: Pulled into this PR in 336d70c0, rather than leaving it to the follow-up: the delivery goroutine now cancels the subscription context on the way out, so a receiver-link failure releases both goroutines and the wait-group slot. Not unit tested, because reaching it needs a live receiver link and a hand-built `amqp.Session` panics inside go-amqp.

pubsub/solace/amqp/amqp.go
```diff
@@ -114,29 +114,38 @@ func (a *amqpPubSub) Publish(ctx context.Context, req *pubsub.PublishRequest) er
...
-					a.logger.Warnf("Failed to publish a message to the broker: %v", err)
-				}
-
-				select {
-				case <-time.After(publishRetryWaitSeconds * time.Second):
-				case <-ctx.Done():
-					break //nolint:staticcheck // SA4011: preserved for backwards compat
-				}
-			}
+		if err == nil {
```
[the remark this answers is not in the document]
nelson-parente: Added in 336d70c0. The loop moved into `sendWithRetry`, which takes the send step as a function and its pause from a field, so `TestSendWithRetry` drives it with a counting fake in milliseconds. It covers the three cases asked for:

- a failure followed by a success stops at exactly two sends,
- an exhausted budget is the first attempt plus three retries, reported as retriable with the last error in the chain,
- a cancelled caller ends the wait, with one send, and gets `context.Canceled` back unwrapped.

I checked that the test fails against both original defects. Without the stop on success it counts four sends instead of two. With the cancellation branch a no-op, as it was before this PR, it gets the broker error back instead of the context error.

PR #4552, opened 2026-08-25
[1.19] feat(pubsub/rabbitmq): support externally managed and consistent hash exchanges

### LINKED ISSUE #4553: RabbitMQ pub/sub: option to use an externally managed exchange, queue and binding instead of declaring them
## Describe the feature

The RabbitMQ pub/sub component always issues an **active** declare for the topology objects it uses: `exchange.declare` for the topic (exchange name = topic name) and `queue.declare` + `queue.bind` for each subscription. There is no way to tell it that the topology already exists and is owned by something else.

That makes the component incompatible with a broker whose topology is managed declaratively — for example the [RabbitMQ Cluster Kubernetes Operator's Topology Operator](https://www.rabbitmq.com/kubernetes/operator/using-topology-operator#exchanges-bindings), which owns `Exchange`, `Queue` and `Binding` custom resources, but equally Terraform, `rabbitmqadmin`, or any other provisioning step.

### Why it breaks

AMQP declaration is idempotent only when *every* property matches. Where it doesn't, RabbitMQ answers `PRECONDITION_FAILED (406)` and closes the channel. The defaults on each side differ on every property:

| | exchange kind | durable | autoDelete |
|---|---|---|---|
| Topology Operator `Exchange` CR defaults | `direct` | `false` | `false` |
| What the component declares | `fanout` | `true` | `true` |

So out of the box the two collide. Coexistence is possible today only by hand-aligning each CR to the component's exact declaration — name, kind, `durable`, `autoDelete`, arguments — which inverts the point of managing topology declaratively, and has to be redone whenever component defaults change.

For an exchange kind the component cannot declare at all (see #3423 and the linked issue for `x-consistent-hash`), there is no alignment that works: the declare can never match, so the component fails permanently on that topic.

### Proposed feature

A passive declare mode, so the component asserts that a topology object exists rather than creating it:

```yaml
spec:
  type: pubsub.rabbitmq
  version: v1
  metadata:
    - name: exchangeDeclareMode
      value: "passive"    # declare (default, today's behaviour) | passive
    - name: queueDeclareMode
      value: "passive"    # declare (default, today's behaviour) | passive
```

- `exchangeDeclareMode: passive` — assert the topic exchange (and the dead-letter exchange, when `enableDeadLetter` is set) exists; never create or modify it.
- `queueDeclareMode: passive` — assert the consumer queue (and dead-letter queue) exists, and leave `queue.bind` to the external owner, since a binding belongs to whoever owns the queue.

Two independent knobs rather than one switch, because the useful middle case is real: an operator-owned exchange paired with queues the component still declares, since consumer queue names are only known at runtime.

`amqp091-go` already exposes `ExchangeDeclarePassive` and `QueueDeclarePassive`, so this is additive. Both default to `declare`, leaving existing behaviour untouched.

Worth pairing with clearer errors: today a `406` surfaces as a raw AMQP exception with no indication that a topology conflict is what happened, or what to do about it.

## Release Note

RELEASE NOTE: **ADD** `exchangeDeclareMode` and `queueDeclareMode` to the RabbitMQ pub/sub component, for brokers whose topology is managed externally (for example by the RabbitMQ Topology Operator).

### LINKED ISSUE #4554: RabbitMQ pub/sub: support the x-consistent-hash exchange type for partitioned per-key ordering
## Describe the feature

Support the `x-consistent-hash` exchange type in the RabbitMQ pub/sub component, so a topic can be partitioned across consumers with per-key ordering preserved.

Previously requested in #3423, which was auto-closed as stale.

### Why the existing option isn't sufficient

The component's only in-the-box ordering control is `concurrencyMode: single`, which gives strict ordering via a *single serial consumer*. That is fine at modest throughput, but it does not scale out: ordering and parallelism become mutually exclusive.

The workload shape this matters for is a high-volume event stream — hundreds of millions of events a day and growing — where ordering is required *per key* (a tenant/device pair, say), not globally, and where processing must stay parallel. On Kafka that is partitioning. On RabbitMQ the equivalent is an `x-consistent-hash` exchange from the [`rabbitmq_consistent_hash_exchange`](https://github.com/rabbitmq/rabbitmq-server/tree/main/deps/rabbitmq_consistent_hash_exchange) plugin, which hashes a routing key to pick one bound queue, so every message for a key lands on the same queue and can be processed in order there while other keys proceed in parallel.

Today the component validates `exchangeKind` against the four built-in AMQP types only, so `x-consistent-hash` can be neither declared nor bound to.

### How it is used

Publish side is unchanged — the existing per-message `routingKey` metadata is the partition key, and the exchange hashes it:

```
client.PublishEvent(ctx, pubsub, topic, data,
    dapr.PublishEventWithMetadata(map[string]string{"routingKey": "tenant-42/serial-7"}))
```

Subscribe side has one thing worth stating explicitly, because it is easy to get wrong: for a consistent-hash exchange the **binding key is the bound queue's bucket weight**, an integer — not a pattern. Each consumer binds its own queue with a weight, and the exchange distributes the key space across the bound queues in proportion:

```yaml
metadata:
  routingKey: "10"   # this queue's bucket weight on the hash ring
```

So the same `routingKey` metadata name carries a partition key when publishing and a bucket weight when subscribing. That asymmetry is inherent to the plugin, and deserves validation and documentation — an empty or non-numeric binding key currently fails deep inside the AMQP channel with a message that does not point at the configuration.

### Client support

No SDK work is needed. `amqp091-go` declares exchanges by an arbitrary kind string (`Channel.ExchangeDeclare(name, kind, …)`), so `x-consistent-hash` is accepted as-is; the plugin lives on the broker, not in the client. The component already carries `routingKey` metadata on both publish and `queue.bind`. What is missing is the validation allowlist and the binding-key semantics above.

### Scope

- Accept `x-consistent-hash` as an `exchangeKind`.
- Validate the subscription binding key as a positive integer bucket weight, and fail with a message naming the metadata field.
- Note the broker prerequisite in the docs: the `rabbitmq_consistent_hash_exchange` plugin must be enabled.

Related: #4553 — where the exchange is created by an external topology manager the component must also be able to bind to it rather than declare its own, and for `x-consistent-hash` that is the only workable path, since a declare of an unsupported kind can never match. The two are useful together.

## Release Note

RELEASE NOTE: **ADD** Support for the `x-consistent-hash` exchange type in the RabbitMQ pub/sub component, for partitioned per-key ordering with multiple consumers.

MyMirelHub:
## Description

Adds two metadata options to the RabbitMQ pub/sub component, and support for the `x-consistent-hash` exchange kind.

| Option | Values | Effect |
|---|---|---|
| `exchangeDeclareMode` | `declare` (default) / `passive` | `passive` asserts the topic exchange already exists; never creates or modifies it. |
| `queueDeclareMode` | `declare` (default) / `passive` | `passive` asserts the consumer queue already exists, and leaves `queue.bind` and dead-lettering to its external owner. |

The two are independent rather than one "external topology" switch: an operator-owned exchange paired with queues the component still declares is the combination consistent hash scale-out needs, since consumer queue names are only known at runtime.

Dead-lettering follows `queueDeclareMode`, not `exchangeDeclareMode`. The DLX and DLQ are named from `consumerID` and topic at runtime, so no external owner could pre-create them; they are the component's own objects and belong with the queue. Under `queueDeclareMode: passive` they are left entirely to the queue's owner and `enableDeadLetter` has no effect.

Also here:

- `x-consistent-hash` accepted as an `exchangeKind`; in `passive` mode any broker-supported kind is accepted, since the component never declares it.
- The consistent-hash binding key is validated as a positive integer bucket weight and fails as a terminal subscription error naming the metadata field, rather than as an opaque channel error the subscription retries forever. Skipped when the binding is not ours to make.
- `PRECONDITION_FAILED` and passive-declare-on-missing-object are rewritten into messages that name what conflicted and which option resolves it. Every other AMQP error passes through untouched.

No behaviour change by default: both modes default to `declare`, and existing `exchangeKind` values are unaffected.

## Issue reference

Closes #4553
Closes #4554

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Component metadata (`metadata.yaml`) updated
- [x] Certification test coverage
- [x] Extended the documentation — companion `dapr/docs` PR: dapr/docs#5297

### Tests

Unit tests cover declare/passive dispatch for exchanges and queues, the decorated and pass-through error paths, bucket-weight validation, and metadata parsing for both new fields.

Two certification tests, run through the runtime against a real broker. The suite's `docker-compose.yml` now enables `rabbitmq_consistent_hash_exchange` via a mounted `enabled_plugins` file.

- `TestRabbitMQPassiveTopology` declares the exchange, queue and binding out-of-band, runs a fully passive component against them, then asserts **via the management API that the broker's set of exchanges and queues is unchanged** — no `<consumerID>-<topic>` queue, no dead-letter objects, no second exchange.
- `TestRabbitMQConsistentHashPartitioning` runs two independently scaled consumers against an operator-owned `x-consistent-hash` exchange and asserts every key is handled by exactly one consumer, the key space is spread across both, and per-key order holds. A fanout-like or unordered regression fails it.

### DIFF
--- a/pubsub/rabbitmq/metadata.go
+++ b/pubsub/rabbitmq/metadata.go
@@ -48,6 +48,7 @@ type rabbitmqMetadata struct {
 	MaxLenBytes                        int64                  `mapstructure:"maxLenBytes"`
 	ExchangeKind                       string                 `mapstructure:"exchangeKind"`
 	ExchangeDeclareMode                string                 `mapstructure:"exchangeDeclareMode"`
+	QueueDeclareMode                   string                 `mapstructure:"queueDeclareMode"`
 	ClientName                         string                 `mapstructure:"clientName"`
 	HeartBeat                          time.Duration          `mapstructure:"heartBeat"`
 	PublisherConfirm                   bool                   `mapstructure:"publisherConfirm"`
@@ -80,6 +81,7 @@ const (
 	metadataMaxLenBytesKey                        = "maxLenBytes"
 	metadataExchangeKindKey                       = "exchangeKind"
 	metadataExchangeDeclareModeKey                = "exchangeDeclareMode"
+	metadataQueueDeclareModeKey                   = "queueDeclareMode"
 	metadataPublisherConfirmKey                   = "publisherConfirm"
 	metadataSaslExternal                          = "saslExternal"
 	metadataMaxPriority                           = "maxPriority"
@@ -93,15 +95,22 @@ const (
 	protocolAMQP  = "amqp"
 	protocolAMQPS = "amqps"
 
-	// exchangeDeclareModeDeclare makes the component declare the exchange itself
-	// (an active AMQP exchange.declare). This is the default and the historical
-	// behavior.
-	exchangeDeclareModeDeclare = "declare"
-	// exchangeDeclareModePassive makes the component verify that the exchange
-	// already exists (a passive AMQP exchange.declare) without creating or
-	// modifying it. Use this when the topology is owned by something else, such
-	// as the RabbitMQ Cluster Kubernetes Topology Operator or Terraform.
-	exchangeDeclareModePassive = "passive"
+	// declareModeDeclare makes the component declare the topology object itself
+	// (an active AMQP declare). This is the default and the historical behavior.
+	declareModeDeclare = "declare"
+	// declareModePassive makes the component verify that the topology object
+	// already exists (a passive AMQP declare) without creating or modifying it.
+	// Use this when the topology is owned by something else, such as the
+	// RabbitMQ Cluster Kubernetes Topology Operator or Terraform.
+	declareModePassive = "passive"
+
+	// Kept as distinct names because the two knobs are independent: an
+	// operator-owned exchange is commonly paired with queues that the component
+	// still creates, since consumer queue names are only known at runtime.
+	exchangeDeclareModeDeclare = declareModeDeclare
+	exchangeDeclareModePassive = declareModePassive
+	queueDeclareModeDeclare    = declareModeDeclare
+	queueDeclareModePassive    = declareModePassive
 
 	// exchangeKindConsistentHash is the exchange type provided by the
 	// rabbitmq_consistent_hash_exchange plugin. It is not a built-in AMQP
@@ -121,6 +130,7 @@ func createMetadata(pubSubMetadata pubsub.Metadata, log logger.Logger) (*rabbitm
 		ReconnectWait:                      time.Duration(defaultReconnectWaitSeconds) * time.Second,
 		ExchangeKind:                       fanoutExchangeKind,
 		ExchangeDeclareMode:                exchangeDeclareModeDeclare,
+		QueueDeclareMode:                   queueDeclareModeDeclare,
 		PublisherConfirm:                   false,
 		SaslExternal:                       false,
 		HeartBeat:                          defaultHeartbeat,
@@ -160,8 +170,13 @@ func createMetadata(pubSubMetadata pubsub.Metadata, log logger.Logger) (*rabbitm
 	}
 
 	result.ExchangeDeclareMode = strings.ToLower(result.ExchangeDeclareMode)
-	if !exchangeDeclareModeValid(result.ExchangeDeclareMode) {
-		return &result, fmt.Errorf("%s invalid RabbitMQ exchange declare mode %q, accepted values are %q and %q", errorMessagePrefix, result.ExchangeDeclareMode, exchangeDeclareModeDeclare, exchangeDeclareModePassive)
+	if !declareModeValid(result.ExchangeDeclareMode) {
+		return &result, fmt.Errorf("%s invalid RabbitMQ %s %q, accepted values are %q and %q", errorMessagePrefix, metadataExchangeDeclareModeKey, result.ExchangeDeclareMode, declareModeDeclare, declareModePassive)
+	}
+
+	result.QueueDeclareMode = strings.ToLower(result.QueueDeclareMode)
+	if !declareModeValid(result.QueueDeclareMode) {
+		return &result, fmt.Errorf("%s invalid RabbitMQ %s %q, accepted values are %q and %q", errorMessagePrefix, metadataQueueDeclareModeKey, result.QueueDeclareMode, declareModeDeclare, declareModePassive)
 	}
 
 	if err := validateExchangeKind(result.ExchangeKind, result.ExchangeDeclareMode); err != nil {
@@ -204,8 +219,8 @@ func (m *rabbitmqMetadata) formatQueueDeclareArgs(origin amqp.Table) amqp.Table 
 	return origin
 }
 
-func exchangeDeclareModeValid(mode string) bool {
-	return mode == exchangeDeclareModeDeclare || mode == exchangeDeclareModePassive
+func declareModeValid(mode string) bool {
+	return mode == declareModeDeclare || mode == declareModePassive
 }
 
 // exchangeKindValid reports whether the component is able to declare an
@@ -238,10 +253,17 @@ func validateExchangeKind(kind string, declareMode string) error {
 	return nil
 }
 
-// isPassiveExchangeDeclare reports whether the exchange topology is managed
-// outside of Dapr.
+// isPassiveExchangeDeclare reports whether exchanges are managed outside of
+// Dapr.
 func (m *rabbitmqMetadata) isPassiveExchangeDeclare() bool {
 	return m.ExchangeDeclareMode == exchangeDeclareModePassive
+}
+
+// isPassiveQueueDeclare reports whether queues and their bindings are managed
+// outside of Dapr. In that case the component neither declares the queue nor
+// binds it: the external owner is responsible for both.
+func (m *rabbitmqMetadata) isPassiveQueueDeclare() bool {
+	return m.QueueDeclareMode == queueDeclareModePassive
 }
 
 func (m *rabbitmqMetadata) connectionURI() string {
--- a/pubsub/rabbitmq/metadata.go
+++ b/pubsub/rabbitmq/metadata.go
@@ -16,6 +16,7 @@ package rabbitmq
 import (
 	"fmt"
 	"net/url"
+	"strings"
 	"time"
 
 	amqp "github.com/rabbitmq/amqp091-go"
@@ -46,6 +47,7 @@ type rabbitmqMetadata struct {
 	MaxLen                             int64                  `mapstructure:"maxLen"`
 	MaxLenBytes                        int64                  `mapstructure:"maxLenBytes"`
 	ExchangeKind                       string                 `mapstructure:"exchangeKind"`
+	ExchangeDeclareMode                string                 `mapstructure:"exchangeDeclareMode"`
 	ClientName                         string                 `mapstructure:"clientName"`
 	HeartBeat                          time.Duration          `mapstructure:"heartBeat"`
 	PublisherConfirm                   bool                   `mapstructure:"publisherConfirm"`
@@ -77,6 +79,7 @@ const (
 	metadataMaxLenKey                             = "maxLen"
 	metadataMaxLenBytesKey                        = "maxLenBytes"
 	metadataExchangeKindKey                       = "exchangeKind"
+	metadataExchangeDeclareModeKey                = "exchangeDeclareMode"
 	metadataPublisherConfirmKey                   = "publisherConfirm"
 	metadataSaslExternal                          = "saslExternal"
 	metadataMaxPriority                           = "maxPriority"
@@ -89,6 +92,22 @@ const (
 
 	protocolAMQP  = "amqp"
 	protocolAMQPS = "amqps"
+
+	// exchangeDeclareModeDeclare makes the component declare the exchange itself
+	// (an active AMQP exchange.declare). This is the default and the historical
+	// behavior.
+	exchangeDeclareModeDeclare = "declare"
+	// exchangeDeclareModePassive makes the component verify that the exchange
+	// already exists (a passive AMQP exchange.declare) without creating or
+	// modifying it. Use this when the topology is owned by something else, such
+	// as the RabbitMQ Cluster Kubernetes Topology Operator or Terraform.
+	exchangeDeclareModePassive = "passive"
+
+	// exchangeKindConsistentHash is the exchange type provided by the
+	// rabbitmq_consistent_hash_exchange plugin. It is not a built-in AMQP
+	// exchange type, so it is only usable against a broker with that plugin
+	// enabled.
+	exchangeKindConsistentHash = "x-consistent-hash"
 )
 
 // createMetadata creates a new instance from the pubsub metadata.
@@ -101,6 +120,7 @@ func createMetadata(pubSubMetadata pubsub.Metadata, log logger.Logger) (*rabbitm
 		AutoAck:                            false,
 		ReconnectWait:                      time.Duration(defaultReconnectWaitSeconds) * time.Second,
 		ExchangeKind:                       fanoutExchangeKind,
+		ExchangeDeclareMode:                exchangeDeclareModeDeclare,
 		PublisherConfirm:                   false,
 		SaslExternal:                       false,
 		HeartBeat:                          defaultHeartbeat,
@@ -139,8 +159,13 @@ func createMetadata(pubSubMetadata pubsub.Metadata, log logger.Logger) (*rabbitm
 		return &result, fmt.Errorf("%s invalid RabbitMQ delivery mode, accepted values are between 0 and 2", errorMessagePrefix)
 	}
 
-	if !exchangeKindValid(result.ExchangeKind) {
-		return &result, fmt.Errorf("%s invalid RabbitMQ exchange kind %s", errorMessagePrefix, result.ExchangeKind)
+	result.ExchangeDeclareMode = strings.ToLower(result.ExchangeDeclareMode)
+	if !exchangeDeclareModeValid(result.ExchangeDeclareMode) {
+		return &result, fmt.Errorf("%s invalid RabbitMQ exchange declare mode %q, accepted values are %q and %q", errorMessagePrefix, result.ExchangeDeclareMode, exchangeDeclareModeDeclare, exchangeDeclareModePassive)
+	}
+
+	if err := validateExchangeKind(result.ExchangeKind, result.ExchangeDeclareMode); err != nil {
+		return &result, err
 	}
 
 	ttl, ok, err := metadata.TryGetTTL(pubSubMetadata.Properties)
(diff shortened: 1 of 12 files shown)


MyMirelHub:
Coverage gap closed in 3b713ff — the uncovered lines were the pass-through branches of `decorateExchangeDeclareError` (a non-AMQP error, an AMQP error of another code, and `NOT_FOUND` raised while the component is the one declaring). All three now have tests pinning that the error reaches the caller untouched; the new functions are at 100%.

Unrelated to this PR: the `pubsub.mqtt3 certification` failure on the previous commit is a `fatal error: concurrent map writes` at `pubsub/mqtt3/mqtt.go:185`, where `delete(m.topics, topic)` races with writers that don't take `deletionLock`. This PR touches only `pubsub/rabbitmq/**` and `tests/certification/pubsub/rabbitmq/**`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -484,25 +503,78 @@ func (r *rabbitMQ) prepareSubscription(channel rabbitMQChannelBroker, req pubsub
...
+	}
+
+	if r.metadata.ExchangeKind != exchangeKindConsistentHash {
+		return nil
+	}
+
+	// For a consistent hash exchange the binding key is the weight of the bound
+	// queue, which must be a positive integer.
+	for _, routingKey := range routingKeys {
+		weight, err := strconv.ParseUint(strings.TrimSpace(routingKey), 10, 32)
```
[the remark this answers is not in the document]
MyMirelHub: Now validates the exact value that gets bound, so a padded key like ` 10 ` is rejected instead of reaching the broker. Test case added.

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -578,7 +650,7 @@ func (r *rabbitMQ) subscribeForever(ctx context.Context, req pubsub.SubscribeReq
 			}
 		}
 
-		if err != nil && strings.Contains(err.Error(), errorInvalidQueueType) {
+		if err != nil && (errors.Is(err, errTerminalSubscription) || strings.Contains(err.Error(), errorInvalidQueueType)) {
 			if ackCh != nil {
 				ackCh <- true
 			}
```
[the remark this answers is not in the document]
MyMirelHub: ackCh is now `chan error`; the terminal error is logged and returned from Subscribe. Test asserts the message names `routingKey`.

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -484,25 +503,78 @@ func (r *rabbitMQ) prepareSubscription(channel rabbitMQChannelBroker, req pubsub
...
 
+// declareQueue creates the queue, or asserts that an externally managed one
+// exists when queueDeclareMode is passive. RabbitMQ ignores every argument but
+// the name on a passive declare, so queue arguments are the external owner's
+// responsibility in that mode.
+func (r *rabbitMQ) declareQueue(channel rabbitMQChannelBroker, queueName string, durable, autoDelete bool, args amqp.Table) (amqp.Queue, error) {
+	if !r.metadata.isPassiveQueueDeclare() {
+		r.logger.Infof("%s declaring queue '%s'", logMessagePrefix, queueName)
+
+		return channel.QueueDeclare(queueName, durable, autoDelete, false, false, args)
```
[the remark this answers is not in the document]
MyMirelHub: Queue PRECONDITION_FAILED is now decorated with the queue name and `queueDeclareMode: passive`, matching the exchange path.

pubsub/rabbitmq/metadata.yaml
```diff
@@ -140,14 +140,51 @@ metadata:
...
     description: |
-      Exchange kind of the rabbitmq exchange.
+      Exchange kind of the rabbitmq exchange. "x-consistent-hash" requires the
+      rabbitmq_consistent_hash_exchange plugin to be enabled on the broker; with
+      it, the "routingKey" publish metadata is the partition key and the
+      "routingKey" subscription metadata is the queue's bucket weight.
+      When "exchangeDeclareMode" is "passive" any exchange kind supported by the
+      broker is accepted, because the component does not declare the exchange.
     default: '"fanout"'
     allowedValues:
```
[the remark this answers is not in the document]
MyMirelHub: Tightened the runtime instead of loosening the schema: exchangeKind now uses the same allowlist in both modes, so metadata.yaml is accurate. Adding a plugin kind is a one-line allowlist change.

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -467,9 +486,9 @@ func (r *rabbitMQ) prepareSubscription(channel rabbitMQChannelBroker, req pubsub
 		args[argMaxLength] = parsedVal
 	}
 
-	q, err := channel.QueueDeclare(queueName, r.metadata.Durable, r.metadata.DeleteWhenUnused, false, false, args)
+	q, err := r.declareQueue(channel, queueName, r.metadata.Durable, r.metadata.DeleteWhenUnused, args)
```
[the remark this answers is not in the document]
MyMirelHub: Passive mode now returns before any queue-argument construction or validation, so an externally managed `stream` queue is consumable. Both directions covered by tests.

tthophan (approved):

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -578,9 +699,10 @@ func (r *rabbitMQ) subscribeForever(ctx context.Context, req pubsub.SubscribeReq
 			}
 		}
 
-		if err != nil && strings.Contains(err.Error(), errorInvalidQueueType) {
+		if err != nil && (errors.Is(err, errTerminalSubscription) || strings.Contains(err.Error(), errorInvalidQueueType)) {
```
JoshVanL, line 718: A passive-mode `404` is never classified as terminal, so this check lets it through to the retry loop. Two consequences:

1. The actionable "does not exist ... create it out-of-band" text from `declareQueue` / `decorateExchangeDeclareError` only ever reaches the logs. `Subscribe` blocks for the full minute and returns the generic `failed to subscribe to <queue>`, which is the opposite of what the PR description promises for this case.
2. A `404` on a passive declare is a channel exception, so the broker closes the shared channel. The decorated error string does not contain `channel/connection is not open`, so `mustReconnect` returns false, the loop spins once more on the dead channel, gets `ErrClosed`, and only then redials. `reset()` closes the connection shared by every subscription and publisher on this component, so one missing operator-owned object kicks every healthy topic off the broker every `ReconnectWait` for as long as the object is missing.

Suggest wrapping the passive `NotFound` (and the `PreconditionFailed`) decorations in `errTerminalSubscription`. `Subscribe` then fails fast with the real message and the runtime's subscription backoff handles the "operator has not created it yet" case, without tearing down the connection for everyone else. The in-memory broker never closes the channel on `404`, so nothing currently tests this path; a `Subscribe`-level test with `queueDeclareErr = &amqp.Error{Code: amqp.NotFound}` would cover it.

pubsub/rabbitmq/rabbitmq.go
JoshVanL, line 963: Related to the comment on line 702: any `*amqp.Error` returned from a declare is a channel exception and the channel is already dead, so treating only the `channel/connection is not open` string as "must reconnect" guarantees one wasted attempt on a closed channel (and, on the publish path, a 2s sleep before it). Consider `errors.As(err, &amqpErr)` here as well.

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -503,6 +523,107 @@ func (r *rabbitMQ) prepareSubscription(channel rabbitMQChannelBroker, req pubsub
 	return &q, nil
 }
 
+// preparePassiveSubscription binds to a queue that is managed outside of Dapr:
+// it asserts the queue (and, when dead lettering is enabled, the dead letter
+// queue) exists and applies QoS, leaving every declaration and binding to the
+// external owner.
+func (r *rabbitMQ) preparePassiveSubscription(channel rabbitMQChannelBroker, req pubsub.SubscribeRequest, queueName string) (*amqp.Queue, error) {
+	if r.metadata.EnableDeadLetter {
+		dlqName := fmt.Sprintf(defaultDeadLetterQueueFormat, queueName)
```
JoshVanL: In passive queue mode the component does not set `x-dead-letter-exchange`, does not declare the DLX and does not bind anything, so `enableDeadLetter` has no runtime effect except requiring a queue literally named `dlq-<queueName>` to exist. An operator who wired dead-lettering to their own queue name gets a `404` here and the reconnect churn described on line 702, for a queue the component would never touch. Suggest dropping this assertion in passive mode; dead-lettering is the owner's responsibility by definition here.

pubsub/rabbitmq/rabbitmq.go
JoshVanL: The DLX now follows `exchangeDeclareMode`, but its name is derived from `consumerID` and topic at runtime, which is the exact reason the PR gives for keeping queues component-declared. With `exchangeDeclareMode=passive` + `queueDeclareMode=declare` (the combination the consistent-hash certification tests use) and `enableDeadLetter=true`, the operator has to pre-create `dlx-<consumerID>-<topic>` for every consumer or every subscription fails. The DLX and DLQ belong with the queue; suggest declaring them under `queueDeclareMode` rather than `exchangeDeclareMode`. No test covers passive exchange + dead letter end to end.

pubsub/rabbitmq/metadata.go
```diff
@@ -179,8 +219,44 @@ func (m *rabbitmqMetadata) formatQueueDeclareArgs(origin amqp.Table) amqp.Table
...
+	default:
+		return false
+	}
+}
+
+// validateExchangeKind checks exchangeKind against the kinds this component
+// supports. The set is the same in both declare modes so that it matches the
+// allowedValues advertised in metadata.yaml, which the component metadata
+// schema treats as a hard allowlist.
+func validateExchangeKind(kind string) error {
```
JoshVanL: The PR description says "in `passive` mode any broker-supported kind is accepted, since the component never declares it", but this enforces the same allowlist in both modes. The comment justifies that by saying the metadata schema treats `allowedValues` as a hard allowlist; nothing at runtime enforces `allowedValues`, it is used for docs and the metadata analyzer only. Since RabbitMQ ignores the kind on a passive declare, the allowlist blocks `x-delayed-message` and other plugin kinds for no benefit. Either relax the check in passive mode or update the description and comment so they match the code.

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -339,7 +350,7 @@ func (r *rabbitMQ) Subscribe(ctx context.Context, req pubsub.SubscribeRequest, h
 	r.logger.Infof("%s subscribe to topic/queue '%s/%s'", logMessagePrefix, req.Topic, queueName)
 
 	// Do not set a timeout on the context, as we're just waiting for the first ack; we're using a semaphore instead
-	ackCh := make(chan bool, 1)
+	ackCh := make(chan error, 1)
 	defer close(ackCh)
```
JoshVanL: Not introduced here, but worth a follow-up: this `close(ackCh)` runs after the one-minute timeout while `subscribeForever` may still execute `ackCh <- nil` (line 685) or `ackCh <- err` (line 705). Sending on a closed channel panics the sidecar. Before, hitting the 60s path needed a broker outage; with passive mode, "the operator has not created the object yet" is an expected long-lived state, so the window is now routine. Making passive `404` terminal (line 702) also shrinks this window.

JoshVanL (changes requested):

pubsub/rabbitmq/rabbitmq.go
MyMirelHub: Both the passive NOT_FOUND and the PRECONDITION_FAILED decorations now wrap `errTerminalSubscription`, so Subscribe fails fast with the actionable message and the connection is never torn down for other subscriptions. Added `TestSubscribeFailsFastOnMissingPassiveObject` covering the missing-exchange and missing-queue cases at Subscribe level.

pubsub/rabbitmq/rabbitmq.go
MyMirelHub: `mustReconnect` now returns true for any `*amqp.Error` via errors.As, keeping the string check as a fallback. Covered by `TestMustReconnectOnAMQPError`.

pubsub/rabbitmq/rabbitmq.go
MyMirelHub: Dropped. Under a passive queue the component no longer touches the dead letter objects at all and logs that `enableDeadLetter` is ignored, rather than requiring `dlq-<queueName>` to exist.

pubsub/rabbitmq/rabbitmq.go
MyMirelHub: Moved: the DLX and DLQ now follow `queueDeclareMode`, and the DLX is always declared actively since its name is only known at runtime. `TestDeadLetterFollowsQueueDeclareMode` covers passive exchange + declared queue + dead letter, and the passive-queue counterpart.

pubsub/rabbitmq/metadata.go
MyMirelHub: You're right, and I had it right before a bot review talked me out of it: `AllowedValues` appears only under .build-tools, nothing enforces it at runtime. Reverted to accepting any kind in passive mode, and corrected the misleading comment.

pubsub/rabbitmq/rabbitmq.go
MyMirelHub: Fixed here rather than deferred: `ackCh` is buffered and written at most once, so the `defer close` was the only hazard and is gone.

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -503,6 +523,107 @@ func (r *rabbitMQ) prepareSubscription(channel rabbitMQChannelBroker, req pubsub
...
+// external owner.
+func (r *rabbitMQ) preparePassiveSubscription(channel rabbitMQChannelBroker, req pubsub.SubscribeRequest, queueName string) (*amqp.Queue, error) {
+	if r.metadata.EnableDeadLetter {
+		dlqName := fmt.Sprintf(defaultDeadLetterQueueFormat, queueName)
+		if _, err := r.declareQueue(channel, dlqName, true, r.metadata.DeleteWhenUnused, nil); err != nil {
+			r.logger.Errorf("%s prepareSubscription for topic/queue '%s/%s' failed in queue declare: %v", logMessagePrefix, req.Topic, dlqName, err)
+
+			return nil, err
+		}
+	}
```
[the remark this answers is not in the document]
MyMirelHub: Followed the maintainer's suggestion on the sibling thread instead: dead lettering now follows `queueDeclareMode`, and a passive queue leaves the DLX and DLQ entirely to their owner rather than asserting objects named by this component.

pubsub/rabbitmq/rabbitmq.go
```diff
@@ -503,6 +523,107 @@ func (r *rabbitMQ) prepareSubscription(channel rabbitMQChannelBroker, req pubsub
...
+
+	// For a consistent hash exchange the binding key is the weight of the bound
+	// queue, which must be a positive integer.
+	// Validate the exact values that will be bound: trimming here but binding
+	// the original would let " 10 " pass and still reach the broker as a
+	// non-integer binding key.
+	for _, routingKey := range routingKeys {
+		weight, err := strconv.ParseUint(routingKey, 10, 32)
+		if err != nil || weight == 0 {
```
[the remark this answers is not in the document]
MyMirelHub: Agreed, now requires exactly one value for this exchange kind. The multi-weight test case flipped to expecting an error.

nelson-parente:
### Code review

One observation, non-blocking.

`ensureExchangeDeclared` is shared by the publish and subscribe paths, and this PR wraps permanent declare failures with `errTerminalSubscription`. `subscribeForever` checks the sentinel and gives up:

https://github.com/dapr/components-contrib/blob/5171bf77e0362956a61c101335e6ee6fb2692fa6/pubsub/rabbitmq/rabbitmq.go#L717-L724

`Publish` has no equivalent check. A passive-mode exchange that does not exist yet burns all three attempts and then returns `pubsub.NewRetriableError`, which is the wrong classification for a configuration error:

https://github.com/dapr/components-contrib/blob/5171bf77e0362956a61c101335e6ee6fb2692fa6/pubsub/rabbitmq/rabbitmq.go#L315-L321

`Publish` is untouched by this PR and the retry churn predates it, so this is a follow-up rather than a blocker.

Nothing else came up. `metadata.yaml` validates against the repo's own metadata schema parser, both new fields default to the pre-existing behavior, the `ackCh` change to a buffered never-closed channel fixes a real send-on-closed-channel panic, and the consumer tag handling from #4326 is intact.

*Reviewed with Claude Code.*

MyMirelHub:
Filed the publish-path observation as #4596 so it is tracked rather than left in this thread — thanks @nelson-parente, the analysis there is yours.

Kept out of this PR deliberately: the `PRECONDITION_FAILED` branch of `decorateExchangeDeclareError` is not gated on passive mode, so having `Publish` check the terminal sentinel would flip a 406 on publish from retriable to terminal for users who never enable either declare mode. That contradicts this PR's no-behaviour-change-by-default guarantee, and the choice between narrowing the classification and accepting the change deserves its own review. Written up in the issue.

nelson-parente (approved):

cicoyle (approved):
Can you confirm there is a docs PR for this?

-------

I see it now 👍🏻

PR #4559, opened 2026-09-01
Adding binary state store support + providers

WhitWaldo:
# Description

Aiming to implement three providers for the [binary state store](https://github.com/dapr/proposals/pull/88) on initial release alongside v1.19:
1) Azure Blob Storage
2) Azure Data Lake Store v2
3) Generic S3-compatible provider
4) Google Cloud Storage
4) OCI object storage

This work pairs with [this PR](https://github.com/dapr/dapr/pull/10367).

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


Makefile
```diff
@@ -108,9 +93,30 @@ verify-linter-version:
...
+GO_TEST_EXEC := -exec="cmd /C exit 0"
+else
+GO_TEST_EXEC := -exec=true
+endif
+endif
+
 .PHONY: test
 test:
-	CGO_ENABLED=$(CGO) go test ./... $(COVERAGE_OPTS) $(BUILDMODE) -tags metadata --timeout=15m
+	CGO_ENABLED=$(CGO) go test $(GO_TEST_EXEC) ./... $(COVERAGE_OPTS) $(BUILDMODE) -tags metadata --timeout=15m
```
JoshVanL: `GO_TEST_EXEC` resolves to `-exec=true` whenever `GOARCH` differs from the host arch, which tells `go test` to run each test binary through `/usr/bin/true`. The binaries are compiled and then never executed, and every package still reports `ok`.

This is live on this PR. The matrix runs `macos-14`, which is an arm64 host, with `GOARCH: amd64`, and `make test` is gated only on `matrix.target_arch != 'arm'`, so the `darwin_amd64` job does run.

From this PR's `Build darwin_amd64 binaries` job:

```
CGO_ENABLED=0 go test -exec=true ./... -coverprofile=coverage.txt -covermode=atomic -tags metadata --timeout=15m
ok  	github.com/dapr/components-contrib/binarystore	0.004s
ok  	github.com/dapr/components-contrib/binarystore/aws/s3	0.004s
```

The same job on `main`:

```
CGO_ENABLED=0 go test ./... -coverprofile=coverage.txt -covermode=atomic -tags metadata --timeout=15m
ok  	github.com/dapr/components-contrib/bindings/apns	(cached)	coverage: 83.9% of statements
ok  	github.com/dapr/components-contrib/bindings/aws/s3	1.343s	coverage: 52.1% of statements
```

Note the missing coverage column and the uniform `0.004s`. One of the three test jobs stopped testing and stays green, so a macOS-specific regression now lands unnoticed.

bindings/azure/blobstorage/blobstorage.go
```diff
@@ -212,6 +212,14 @@ func (a *AzureBlobStorage) Operations() []bindings.OperationKind {
 	}
 }
 
+// blobName applies the configured prefix to a user-supplied blob name.
+func (a *AzureBlobStorage) blobName(name string) string {
```
JoshVanL, line 216: `blobName` prepends the component prefix to every caller-supplied name, but `list` returns Azure's blob names unchanged, and those already contain the prefix. A name that comes out of `list` and goes back into another operation therefore gets the prefix applied twice.

`list` marshals `resp.Segment.BlobItems` straight into the response with no prefix stripping:

```go
blobs = append(blobs, resp.Segment.BlobItems...)
...
jsonResponse, err := json.Marshal(blobs)
```

Meanwhile `get` (line 312), `delete` (line 430), `presign` (line 672), `create` (line 265), and `resolveBlobs` (line 541) all call `a.blobName(...)` unconditionally.

Concrete sequence with `prefix: tenantA`:

1. `list` returns an item with `Name: "tenantA/report.pdf"`.
2. The app passes that name to `get`.
3. `blobName` produces `tenantA/tenantA/report.pdf`, and the request fails with `BlobNotFound`.

`delete` is worse, because `BlobNotFound` is treated as success there, so the blob is silently never deleted.

Either strip the prefix from the names `list` returns, or treat names supplied by the caller as already absolute. The first keeps the binding's external contract prefix-free and is probably what users expect.

binarystore/aws/s3/s3.go
```diff
@@ -0,0 +1,251 @@
...
+// Features returns the optional features supported by this component.
+func (s *AWSS3) Features() []binarystore.Feature {
+	return []binarystore.Feature{}
+}
+
+// Set uploads binary data to S3.
+//
+// When req.Overwrite is false, an If-None-Match: * condition is applied so
+// that the server rejects the upload atomically if the object already
+// exists, returning binarystore.ErrFileAlreadyExists. Not all S3-compatible
```
JoshVanL, line 109: The doc comment says:

> Not all S3-compatible providers support conditional writes; in that case the underlying error is surfaced unless it maps to a PreconditionFailed-style response.

That is not what happens. A server that does not implement `If-None-Match` does not return an error, it ignores the unrecognised header and performs an ordinary `PutObject`. `UploadObject` returns nil, so `Set` returns nil at line 137 and the pre-existing object has been replaced.

So on an endpoint without conditional-write support:

```go
Set(ctx, &binarystore.SetRequest{
    FileName:  "invoice.bin",  // already exists
    Data:      newBytes,
    Overwrite: false,          // POST semantics: must not clobber
})
// returns nil, invoice.bin now holds newBytes
```

That is silent data loss on the exact path the caller chose in order to avoid it. Older MinIO builds and Ceph RADOS Gateway are the common cases, and `s3Metadata` explicitly supports custom endpoints, so this is reachable in normal use.

Options, roughly in order of cost:

- Probe conditional-write support once during `Init` and fail fast if the endpoint does not honour it.
- Document the endpoint requirement in `metadata.yaml` and keep the behaviour.

At minimum the comment should say what actually happens, since as written it implies a safety guarantee the code does not provide.

binarystore/oci/objectstorage/objectstorage.go
```diff
@@ -0,0 +1,426 @@
...
+			ObjectName:          &name,
+			ObjectStorageClient: c.objectClient,
+		},
+		StreamReader: data,
+	}
+	if !overwrite {
+		req.IfNoneMatch = common.String("*")
+	}
+
+	resp, err := transfer.NewUploadManager().UploadStream(ctx, req)
```
JoshVanL, line 341: An empty file fails to upload here whenever `req.Data` is not one of four concrete reader types, which is the normal case for a body arriving from the runtime.

`putObject` passes `req.Data` through as `StreamReader`, and `UploadManager.UploadStream` only short-circuits a zero-length body to a single `PutObject` for a fixed set of types. From `oci-go-sdk` v54.0.0, `objectstorage/transfer/upload_manager.go:123`:

```go
func isZeroLength(streamReader io.Reader) bool {
	switch v := streamReader.(type) {
	case *bytes.Buffer:
		return v.Len() == 0
	case *bytes.Reader:
		return v.Len() == 0
	case *strings.Reader:
		return v.Len() == 0
	case *os.File:
		...
	default:
		return false
	}
}
```

Any other reader hits `default` and takes the multipart path. `splitStreamToParts` in `multipart_manifest.go:167` then does:

```go
numberOfBytesRead, err := io.ReadFull(reader, buffer)
if err == io.EOF {
    break
}
```

On an empty stream the first `ReadFull` returns `io.EOF`, so zero parts are emitted and `CommitMultipartUpload` is called with an empty part list, which the service rejects.

The other five providers all create a zero-byte object for the same call, so this is a behaviour divergence as well as a failure. The conformance suite does not catch it because it uses `bytes.NewReader`, which is one of the four short-circuited types.

Simplest fix is to read the first part yourself and issue a plain `PutObject` when the read hits EOF immediately, falling back to `UploadStream` over an `io.MultiReader` otherwise. Worth adding an empty-payload case to the conformance suite either way.

binarystore/aws/s3/s3.go
```diff
@@ -0,0 +1,251 @@
...
+	if errors.As(err, &apiErr) {
+		switch apiErr.ErrorCode() {
+		case "PreconditionFailed", "ConditionalRequestConflict":
+			return true
+		}
+	}
+
+	var re interface{ HTTPStatusCode() int }
+	if errors.As(err, &re) {
+		return re.HTTPStatusCode() == http.StatusPreconditionFailed || re.HTTPStatusCode() == http.StatusConflict
```
JoshVanL, line 247: `isPreconditionFailed` matches on the bare status code:

```go
return re.HTTPStatusCode() == http.StatusPreconditionFailed || re.HTTPStatusCode() == http.StatusConflict
```

and `Set` consults it unconditionally at line 131, including when `req.Overwrite` is true and no `If-None-Match` was ever sent.

S3 returns 409 for conditions unrelated to object existence. `OperationAborted` ("A conflicting conditional operation is currently in progress against this resource") is the common one, returned when two writers race on the same key.

So two concurrent `Set(FileName: "big.bin", Overwrite: true)` calls give the loser a 409, and `Set` returns `binarystore.ErrFileAlreadyExists` for a PUT-semantics write whose contract is to replace. The caller sees a conflict that does not exist, and the real retryable error is discarded.

Two small changes:

```go
	_, err := s.tmClient.UploadObject(ctx, input)
	if err != nil {
		if !req.Overwrite && isPreconditionFailed(err) {
			return binarystore.ErrFileAlreadyExists
		}
		return fmt.Errorf("error uploading object %q: %w", req.FileName, err)
	}
```

and dropping `http.StatusConflict` from the status-code fallback, since `ConditionalRequestConflict` is already matched by error code above it.

binarystore/aws/s3/s3.go
```diff
@@ -0,0 +1,251 @@
...
+// binarystore.ErrFileNotFound is returned.
+func (s *AWSS3) Delete(ctx context.Context, req *binarystore.DeleteRequest) error {
+	if req.FileName == "" {
+		return binarystore.ErrMissingFileName
+	}
+
+	// S3 requires the object to exist to distinguish delete from no-op; check
+	// first so callers reliably receive ErrFileNotFound (S3's DeleteObject is
+	// idempotent and does not error when the key is missing).
+	_, err := s.s3Client.HeadObject(ctx, &s3.HeadObjectInput{
```
JoshVanL, line 175: The `HeadObject` here exists only to synthesise `ErrFileNotFound`, but it has two side effects worth weighing.

First, `HeadObject` is authorised by `s3:GetObject`, not `s3:DeleteObject`. The component now needs read permission in order to delete, which is not obvious from the component's documented behaviour.

Second, when the caller lacks `s3:ListBucket`, S3 answers `HeadObject` for a missing key with 403 `AccessDenied` rather than 404, specifically so that it does not leak key existence. `isNotFound` at line 212 matches only `NoSuchKey`, `NotFound`, and status 404, so the 403 falls through.

With an IAM policy granting just `s3:PutObject` and `s3:DeleteObject`:

```
Delete(FileName: "a.bin")
  -> HeadObject returns 403 AccessDenied
  -> isNotFound(err) == false
  -> returns: error checking object "a.bin": AccessDenied
```

`DeleteObject` is never reached, so deletes fail entirely under a delete-only policy.

If the `ErrFileNotFound` guarantee is worth keeping, classifying 403 from `HeadObject` as not-found would at least restore the delete path, though it conflates two conditions. Documenting the `s3:GetObject` requirement in `metadata.yaml` would help either way.

.github/scripts/test-info.mjs
```diff
@@ -6,6 +6,47 @@ import { writeFileSync } from 'node:fs'
...
+    'binarystore.gcp.bucket': {
+        conformance: true,
+        requireGCPCredentials: true,
+        requiredSecrets: ['GCPBucket'],
+        sourcePkg: ['binarystore/gcp/bucket'],
+    },
+    'binarystore.oci.objectstorage': {
+        conformance: true,
+        requiredSecrets: [
+            'DAPR_TEST_OCI_CONFIG_FILE_PATH',
```
JoshVanL, line 44: These three `requiredSecrets` names contain underscores, and Azure Key Vault secret names only allow letters, digits, and hyphens. They cannot exist in the vault, so the fetch cannot succeed.

`conformance.yml` passes each name straight to the CLI:

```bash
for secretName in $(echo -n $secrets | tr ',' ' '); do
  value=$(az keyvault secret show \
    --name $secretName \
    --vault-name $VAULT_NAME \
    ...
```

The step runs under `bash -e`, and a failed command substitution in an assignment aborts the step, so the whole `binarystore.oci.objectstorage` job fails at setup before any test runs.

Every other entry in this file uses hyphen-free CamelCase, for example `AzureBlobStorageAccount` and `AWSS3Bucket`; these are the only three names in the file with underscores. Renaming them to match, for example `OCIConfigFile`, `OCICompartmentOCID`, and `OCIBucketName`, would fix it. The component YAML in `tests/config/binarystore/oci/objectstorage/binarystore.yaml` and the `Setup OCI config file` step in `conformance.yml` reference the same names and would need the same rename.

This will only show up on an authenticated run, so it will not fail on this PR's checks.

common/component/azure/blobstorage/metadata.go
```diff
@@ -26,9 +26,10 @@ import (
 
 type BlobStorageMetadata struct {
 	ContainerClientOpts     `json:",inline" mapstructure:",squash"`
-	DecodeBase64            bool `json:"decodeBase64,string" mapstructure:"decodeBase64" mdonly:"bindings"`
-	PublicAccessLevel       azblob.PublicAccessType
-	DisableEntityManagement bool `json:"disableEntityManagement,string" mapstructure:"disableEntityManagement"`
+	Prefix                  string                  `json:"prefix" mapstructure:"prefix" mdonly:"state,binarystore"`
```
JoshVanL, line 29: The `mdonly:"state,binarystore"` tag and the binding's code disagree about whether `prefix` applies to `bindings.azure.blobstorage`.

`metadata/utils.go:231` uses the tag to drop the field from a component's advertised metadata when its type is not listed:

```go
if mdOnlyTag := currentField.Tag.Get("mdonly"); mdOnlyTag != "" {
	include := false
	for _, tag := range strings.Split(mdOnlyTag, ",") {
		if tag == string(componentType) {
			include = true
			break
		}
	}
	if !include {
		continue
	}
}
```

The binding passes `contribMetadata.BindingType` at `bindings/azure/blobstorage/blobstorage.go:1064`, so `prefix` is excluded from what it reports. But `blobName` at line 216 reads `a.metadata.Prefix` and applies it to every create, get, delete, list, presign, and bulk operation, because `mdonly` only affects the reported schema, not decoding.

Two consequences:

- `bindings/azure/blobstorage/metadata.yaml` has no `prefix` entry, and `make check-component-metadata` cannot flag the gap, because the field is hidden from it. This PR does add `prefix` to both `state/azure/blobstorage/v1/metadata.yaml` and `v2/metadata.yaml`, so the omission is specific to bindings.
- An existing binding component that carries a stray `prefix` key, previously ignored as an unknown field, silently relocates every blob to `<prefix>/<name>` after upgrade. Existing blobs become unreachable through `get`, `delete`, and `presign`.

If bindings should support `prefix`, please add it to `bindings/azure/blobstorage/metadata.yaml` and include `bindings` in the tag. If not, the binding should stop reading it.

tests/conformance/binarystore/binarystore.go
```diff
@@ -0,0 +1,279 @@
...
+		require.NoError(t, err)
+		require.NotNil(t, resp)
+		defer resp.Data.Close()
+
+		got, err := io.ReadAll(resp.Data)
+		require.NoError(t, err)
+		assert.Equal(t, payload, got)
+	})
+
+	t.Run("set without overwrite conflicts on existing file", func(t *testing.T) {
```
JoshVanL, line 125: The suite has exactly one `Overwrite: false` call, and it deliberately creates the file first:

```go
t.Run("set without overwrite conflicts on existing file", func(t *testing.T) {
	// Ensure the file exists from the previous test.
	err := store.Set(ctx, &binarystore.SetRequest{
		FileName:  fileName,
		Data:      bytes.NewReader([]byte("first")),
		Overwrite: true,
	})
	require.NoError(t, err)

	err = store.Set(ctx, &binarystore.SetRequest{
		FileName:  fileName,
		Data:      bytes.NewReader([]byte("second")),
		Overwrite: false,
	})
	require.ErrorIs(t, err, binarystore.ErrFileAlreadyExists)
```

The other seven `Set` calls all use `Overwrite: true`. Nothing calls `Set` with `Overwrite: false` against a name that does not yet exist.

That untested call is the one that carries each provider's server-side precondition:

- `s3.go:126` sets `IfNoneMatch: "*"`
- `bucket.go:194` sets `storage.Conditions{DoesNotExist: true}`
- `objectstorage.go:338` sets `IfNoneMatch: "*"`
- `blobstorage.go:75` sets `IfNoneMatch: ETagAny`
- `datalake.go:115` attaches the condition to the `Rename`

If any of those headers were rejected by the service, or scoped to the wrong resource, every create-only write would fail in production and this suite would still pass. A single subtest that writes a fresh name with `Overwrite: false` and asserts `require.NoError` would close the gap.

tests/config/binarystore/tests.yml
```diff
@@ -0,0 +1,14 @@
+componentType: binarystore
+components:
```
JoshVanL, line 2: `prefix` is declared in all six provider `metadata.yaml` files and is applied through `binarystore.ObjectPath` on `Set`, `Get`, and `Delete` in every implementation, but almost nothing exercises it.

None of the six conformance component configs set it:

```
tests/config/binarystore/aws/s3/binarystore.yaml
tests/config/binarystore/azure/blobstorage/binarystore.yaml
tests/config/binarystore/azure/datalake/binarystore.yaml
tests/config/binarystore/gcp/bucket/binarystore.yaml
tests/config/binarystore/in-memory/binarystore.yaml
tests/config/binarystore/oci/objectstorage/binarystore.yaml
```

A `grep -ri prefix tests/config/binarystore/` returns nothing. At the unit level, only `in_memory_test.go` and `s3_test.go` mention it at all; `blobstorage_test.go`, `datalake_test.go`, `bucket_test.go`, and `objectstorage_test.go` have zero mentions.

So a provider that forgot `ObjectPath` on one operation would pass everything. For instance, if `Delete` called `NewBlockBlobClient(req.FileName)` instead of `NewBlockBlobClient(binarystore.ObjectPath(prefix, req.FileName))`, it would target the wrong key, return `ErrFileNotFound`, and leave the real object in the container, with no test failing.

Adding a second entry to `tests.yml` for one provider with a prefix set, or a `prefix` profile alongside the existing configs, would cover the round trip cheaply.

binarystore/in-memory/in_memory.go
```diff
@@ -0,0 +1,128 @@
...
+// NewInMemoryBinaryStore returns a new in-memory BinaryStore.
+func NewInMemoryBinaryStore(log logger.Logger) binarystore.BinaryStore {
+	return &InMemoryBinaryStore{
+		logger: log,
+		files:  make(map[string][]byte),
+	}
+}
+
+func (s *InMemoryBinaryStore) Init(_ context.Context, md binarystore.Metadata) error {
+	s.prefix = md.Properties["prefix"]
```
JoshVanL, line 50: This reads the prefix with a case-sensitive map lookup:

```go
s.prefix = md.Properties["prefix"]
```

Every other provider declares it as a struct field and decodes it through the metadata decoder, which matches keys case-insensitively:

```go
Prefix string `json:"prefix" mapstructure:"prefix"`
```

So a component spec written as `- name: Prefix` or `- name: PREFIX` works on `binarystore.aws.s3`, `binarystore.gcp.bucket`, `binarystore.oci.objectstorage`, and both Azure providers, and is silently ignored here. Someone copying a working spec to `binarystore.in-memory` for local development gets different object names locally than in their deployed environment, with no error to explain it.

`binarystore/fake/fake.go:49` has the same lookup.

Decoding it the same way as the others removes the discrepancy:

```go
type inMemoryMetadata struct {
	Prefix string `json:"prefix" mapstructure:"prefix"`
}
```

binarystore/azure/blobstorage/blobstorage.go
```diff
@@ -0,0 +1,149 @@
...
+// When req.Overwrite is false an If-None-Match: * access condition is applied
+// so that the server rejects the upload atomically if the blob already exists,
+// returning binarystore.ErrFileAlreadyExists. When req.Overwrite is true the
+// blob is created or replaced without a condition check.
+func (a *AzureBlobStorage) Set(ctx context.Context, req *binarystore.SetRequest) error {
+	if req.FileName == "" {
+		return binarystore.ErrMissingFileName
+	}
+
+	opts := &blockblob.UploadStreamOptions{}
```
JoshVanL, line 71: `opts` is created with no `BlockSize` or `Concurrency`, so the SDK defaults apply. From `azblob` v1.8.0, `blockblob/models.go:350`:

```go
func (u *UploadStreamOptions) setDefaults() {
	if u.Concurrency == 0 {
		u.Concurrency = 1
	}
	if u.BlockSize < _1MiB {
		u.BlockSize = _1MiB
	}
}
```

One MiB blocks with one request in flight. Two consequences for a building block whose package doc promises files "ranging from a few bytes to several gigabytes":

- A 4 GiB upload becomes 4096 sequential `StageBlock` round trips, so throughput is bounded by per-request latency rather than bandwidth.
- Block blobs cap at 50,000 blocks, so 1 MiB blocks put a hard ceiling around 48.8 GiB on the object size.

The other providers in this PR are already an order of magnitude higher: S3 uses the `transfermanager` defaults, GCS uses the storage `Writer` default chunk, and OCI uses 10 MiB parts. Something like:

```go
opts := &blockblob.UploadStreamOptions{
	BlockSize:   8 * 1024 * 1024,
	Concurrency: 4,
}
```

is about 32 MiB of pooled buffers and still streams. `datalake.go:98` passes `nil` and has the same defaults, so it needs `file.UploadStreamOptions{ChunkSize, Concurrency}` too.

binarystore/oci/objectstorage/objectstorage.go
```diff
@@ -0,0 +1,426 @@
...
+		ObjectName:    &name,
+	})
+	if err != nil {
+		return nil, err
+	}
+	return resp.Content, nil
+}
+
+func (c *ociObjectStoreClient) deleteObject(ctx context.Context, name string) error {
+	_, err := c.objectClient.HeadObject(ctx, ociobjectstorage.HeadObjectRequest{
```
JoshVanL, line 372: `deleteObject` issues a `HeadObject` before `DeleteObject` purely to produce `ErrFileNotFound`, but OCI Object Storage already returns 404 `ObjectNotFound` from `DeleteObject` for a missing key, so the extra call is not needed.

Two costs:

- Every delete is two authenticated round trips instead of one.
- An object removed by another writer between the `HeadObject` at line 372 and the `DeleteObject` at line 380 makes `Delete` report success, when `binarystore/store.go` says it should return `ErrFileNotFound`.

`ObjectStorage.Delete` at line 225 already routes the error through `isNotFound`, so dropping the head call preserves the sentinel and closes the race:

```go
func (c *ociObjectStoreClient) deleteObject(ctx context.Context, name string) error {
	_, err := c.objectClient.DeleteObject(ctx, ociobjectstorage.DeleteObjectRequest{
		NamespaceName: &c.metadata.Namespace,
		BucketName:    &c.metadata.BucketName,
		ObjectName:    &name,
	})
	return err
}
```

Worth noting the S3 provider's equivalent head call is justified, because S3's `DeleteObject` is a no-op for a missing key. OCI's is not.

binarystore/azure/datalake/datalake.go
```diff
@@ -0,0 +1,208 @@
...
+			return fmt.Errorf("error checking file %q: %w", req.FileName, err)
+		}
+	}
+
+	if _, err := tempClient.Create(ctx, nil); err != nil {
+		return fmt.Errorf("error creating temporary file %q: %w", req.FileName, err)
+	}
+
+	if err := tempClient.UploadStream(ctx, req.Data, nil); err != nil {
+		_ = cleanupFileIfExists(ctx, tempClient)
```
JoshVanL, line 99: Both temp-file cleanup calls discard their error:

```go
	if err := tempClient.UploadStream(ctx, req.Data, nil); err != nil {
		_ = cleanupFileIfExists(ctx, tempClient)
		return fmt.Errorf("error uploading file %q: %w", req.FileName, err)
	}
```

and again at line 116 after a failed `Rename`.

`cleanupFileIfExists` already swallows `PathNotFound` internally, so the only errors it can return are genuine failures, and those now vanish. When one happens, `<objectPath>.tmp-<uuid>` stays in the filesystem indefinitely, accruing storage cost, with nothing operator-visible emitted. Each retry leaves another orphan behind.

The OCI provider handles the equivalent case at `objectstorage.go:352`:

```go
		if abortErr != nil {
			c.logger.Warnf("failed to abort OCI multipart upload for object %q: %v", name, abortErr)
		}
```

Same treatment here would make the leak diagnosable:

```go
		if cleanupErr := cleanupFileIfExists(ctx, tempClient); cleanupErr != nil {
			a.logger.Warnf("failed to remove temporary file %q: %v", tempPath, cleanupErr)
		}
```

binarystore/fake/fake.go
```diff
@@ -0,0 +1,121 @@
...
+	"reflect"
+	"sync"
+
+	"github.com/dapr/components-contrib/binarystore"
+	"github.com/dapr/components-contrib/metadata"
+	"github.com/dapr/kit/logger"
+)
+
+// Fake is an in-memory binary store intended for tests only.
+type Fake struct {
```
JoshVanL, line 33: `binarystore/fake` is a functional duplicate of `binarystore/in-memory`. Same struct fields, same map, same lock, same `Init`, `Set`, `Get`, `Delete`, and `GetComponentMetadata` bodies. The only behavioural difference is that `in-memory`'s `Close` clears the map and this one does not.

Nothing imports it. A grep for the import path across the repo returns only the package's own files, and the conformance suite imports `binarystore/in-memory` instead, at `tests/conformance/binarystore_test.go:31`. It is absent from `test-info.mjs` and from `tests/config/binarystore/tests.yml`, and CI reports 0.0% coverage for it since it has no tests.

Because `binarystore/fake/metadata.yaml` exists, the bundle generator picks it up, so the published component metadata bundle carries an entry named `fake`. `.build-tools/cmd/cmd-check-component-registrations.go:79` also runs with empty ignore lists:

```go
checkComponents("binarystore", []string{}, []string{})
```

so the release-time check will count it as a contrib component with no runtime registration.

No other building block in this repo ships a `fake` package; `state/in-memory` and `pubsub/in-memory` double as the test doubles for theirs. Deleting `binarystore/fake` and using `inmemory.NewInMemoryBinaryStore` wherever a fake is wanted would drop about 120 lines plus a `metadata.yaml`, and removes the need to keep two copies in sync.

binarystore/azure/datalake/datalake.go
```diff
@@ -0,0 +1,208 @@
...
+// the path already exists, returning binarystore.ErrFileAlreadyExists. When
+// req.Overwrite is true, the file is created or replaced without a condition
+// check.
+func (a *AzureDataLakeStorage) Set(ctx context.Context, req *binarystore.SetRequest) error {
+	if req.FileName == "" {
+		return binarystore.ErrMissingFileName
+	}
+
+	objectPath := binarystore.ObjectPath(a.metadata.Prefix, req.FileName)
+	if err := a.ensureParentDirectories(ctx, objectPath); err != nil {
```
JoshVanL, line 76: `ensureParentDirectories` runs on every `Set` and issues one `Create` per path segment:

```go
	for i := 1; i < len(parts); i++ {
		dirPath := strings.Join(parts[:i], "/")
		dirClient := a.fileSystemClient.NewDirectoryClient(dirPath)
		_, err := dirClient.Create(ctx, nil)
```

ADLS Gen2 has a hierarchical namespace and creates intermediate directories implicitly when a path is created, so the loop is not needed. With `prefix: a/b/c` it costs three `Create` calls per write, each returning 409 `PathAlreadyExists` after the first run.

The `GetProperties` pre-check a few lines below is also redundant:

```go
	if !req.Overwrite {
		_, err := fileClient.GetProperties(ctx, nil)
		switch {
		case err == nil:
			return binarystore.ErrFileAlreadyExists
```

The `Rename` at line 115 already carries `If-None-Match: *` and maps the conflict to `ErrFileAlreadyExists`, so the pre-check only adds a round trip and a TOCTOU window that the condition already closes.

Together that is a small `Overwrite: false` write costing six or more metadata round trips on a per-file hot path. Dropping the directory loop and the `GetProperties` takes it to three.

JoshVanL (changes requested):

Makefile
WhitWaldo: Resolved

bindings/azure/blobstorage/blobstorage.go
WhitWaldo: That's a really good catch - fixed

binarystore/aws/s3/s3.go
WhitWaldo: Fair

binarystore/azure/datalake/datalake.go
```diff
@@ -0,0 +1,294 @@
...
+
+	if err := c.createFile(ctx, tempClient, tempPath); err != nil {
+		return err
+	}
+
+	uploadOpts := &file.UploadStreamOptions{
+		// Chunk size and concurrency are set explicitly so buffering matches
+		// the other binary store providers rather than the SDK defaults.
+		ChunkSize:   binarystore.DefaultUploadPartSize,
+		Concurrency: binarystore.DefaultUploadConcurrency,
```
JoshVanL, line 190: `Concurrency: binarystore.DefaultUploadConcurrency` is `2` here, and that value triggers a data race inside the SDK that can silently truncate the stored file.

`UploadStreamOptions.Concurrency` defaults to **1** in `azdatalake v1.1.0` (`file/models.go`, `setDefaults()` sets `u.Concurrency = 1` when it is zero). The doc comment claiming a default of 5 is wrong, and the default of 1 is load-bearing: `copyFromReader` in `file/chunkwriting.go` declares the flush length at function scope

```go
actualSize := int64(0)
```

and then increments it *inside each chunk goroutine* with no lock and no atomic:

```go
go func(chunkNum uint32) {
    offset := int64(chunkNum) * options.ChunkSize
    appendDataOpts := options.getAppendDataOptions()
    actualSize += int64(len(buffer[:n]))   // unsynchronised
    _, err := dst.AppendData(ctx, offset, ...)
```

after which it flushes with that value once the goroutines are done (`wg.Wait()`, then `dst.FlushData(ctx, actualSize, flushOpts)`). With `Concurrency: 1` only one goroutine ever touches `actualSize`, so the bug is invisible. Setting it to 2 makes two run concurrently.

Concrete trigger, from your own conformance suite: the large-payload case uploads `2*DefaultUploadPartSize + 1MiB` = 33 MiB, which is 3 chunks of 16 MiB. Chunks 0 and 1 are in flight together, both read `actualSize == 0` and both store `16777216`, so the final flush length is roughly 24 MiB instead of 33 MiB. The data was appended, but `FlushData` commits a shorter file, `Set` returns `nil`, and the truncation is only visible on a later `Get`. It is a race, so it will present as an intermittent conformance failure rather than a deterministic one.

This is specific to Data Lake. `azblob` tracks committed size with an atomic `blockTracker`, and the S3, GCS and OCI paths have no equivalent shared counter, so `DefaultUploadConcurrency` is fine for them and does not need changing.

Suggested fix, keeping the shared constant for the other providers:

```go
uploadOpts := &file.UploadStreamOptions{
    ChunkSize: binarystore.DefaultUploadPartSize,
    // azdatalake's copyFromReader increments the flush length from inside
    // each chunk goroutine without synchronisation, so anything above 1
    // races and can commit a short file. See file/chunkwriting.go.
    Concurrency: 1,
}
```

If the throughput matters here, the alternative is to size the body in the component and drive `AppendData` / `FlushData` directly, where you control the length.

Makefile
```diff
@@ -25,29 +25,14 @@ GIT_VERSION = $(shell git describe --always --abbrev=7 --dirty)
...
-endif
-export GOOS ?= $(TARGET_OS_LOCAL)
-
-ifeq ($(GOOS),windows)
+HOST_GOOS := $(shell go env GOHOSTOS)
+HOST_GOARCH := $(shell go env GOHOSTARCH)
+export GOARCH ?= $(HOST_GOARCH)
+export GOOS ?= $(HOST_GOOS)
+TARGET_ARCH_LOCAL := $(GOARCH)
+TARGET_OS_LOCAL := $(GOOS)
```
JoshVanL, line 33: Moving to `go env GOHOSTOS` / `GOHOSTARCH` is a nice cleanup, but these two lines invert the meaning of `TARGET_OS_LOCAL` / `TARGET_ARCH_LOCAL`, and one existing recipe depended on the old meaning.

Before, they were derived from `uname` and were always the **host**, with `export GOOS ?= $(TARGET_OS_LOCAL)` flowing host to target. Now `TARGET_OS_LOCAL := $(GOOS)` flows the other way, so with `GOOS` set they become the **target**.

The `compile-build-tools` recipe builds a binary that the Makefile then executes locally:

```make
cd .build-tools; CGO_ENABLED=$(CGO) GOOS=$(TARGET_OS_LOCAL) GOARCH=$(TARGET_ARCH_LOCAL) go build -o $(BUILD_TOOLS_BIN) .
```

Trigger: on a Linux box with `GOOS=windows` exported (or `make GOOS=windows check-component-metadata`), this now cross-compiles `components-contrib-build-tools` for Windows and then immediately fails to run it. Previously it was always built for the host regardless of `GOOS`.

You already introduced the right variables, so the recipe just needs to use them:

```make
cd .build-tools; CGO_ENABLED=$(CGO) GOOS=$(HOST_GOOS) GOARCH=$(HOST_GOARCH) go build -o $(BUILD_TOOLS_BIN) .
```

You switched the two `ifeq (...,windows)` guards to `HOST_GOOS` for exactly this reason, so this looks like the one site that was missed.

tests/conformance/binarystore_test.go
```diff
@@ -0,0 +1,121 @@
...
+			t.Skipf("Skipping Azure Blob Storage conformance test: AzureBlobStorageAccount and AzureBlobStorageAccessKey environment variables must be set")
+			return true
+		}
+	case "azure.datalake":
+		if os.Getenv("AzureBlobStorageAccount") == "" || os.Getenv("AzureBlobStorageAccessKey") == "" {
+			t.Skipf("Skipping Azure Data Lake Storage conformance test: AzureBlobStorageAccount and AzureBlobStorageAccessKey environment variables must be set")
+			return true
+		}
+	case "aws.s3":
+		if os.Getenv("AWSS3Bucket") == "" {
```
JoshVanL, line 84: `shouldSkipBinaryStoreComponent` does not cover every env var the component YAMLs interpolate, so a partially configured developer machine gets a hard failure instead of a skip.

`tests/config/binarystore/aws/s3/binarystore.yaml` references `${{AWS_REGION}}`, but this case only checks `AWSS3Bucket`. `parseMetadataProperty` returns `required env var is not set` for an empty value, so setting `AWSS3Bucket` without `AWS_REGION` fails the test rather than skipping it.

The OCI case at line 94 has the same gap: the OCI YAMLs reference `${{OCIConfigProfile}}`, which is not in the `OCIConfigFile` / `OCICompartmentOCID` / `OCIBucketName` trio. CI happens to be safe only because the new workflow step defaults it:

```sh
if [ -z "${OCIConfigProfile:-}" ]; then
  echo "OCIConfigProfile=DEFAULT" >> "$GITHUB_ENV"
```

so this bites local runs, not the pipeline.

```go
case "aws.s3":
    if os.Getenv("AWSS3Bucket") == "" || os.Getenv("AWS_REGION") == "" {
        t.Skipf("Skipping AWS S3 conformance test: AWSS3Bucket and AWS_REGION environment variables must be set")
        return true
    }
```

and add `OCIConfigProfile` to the OCI condition.

.github/scripts/test-info.mjs
```diff
@@ -6,6 +6,47 @@ import { writeFileSync } from 'node:fs'
...
+    },
+    'binarystore.azure.blobstorage': {
+        conformance: true,
+        requiredSecrets: [
+            'AzureBlobStorageAccount',
+            'AzureBlobStorageAccessKey',
+        ],
+        sourcePkg: ['binarystore/azure/blobstorage', 'common/component/azure/blobstorage'],
+    },
+    'binarystore.azure.datalake': {
```
JoshVanL, line 21: `binarystore.azure.datalake` declares the same secrets as `binarystore.azure.blobstorage` (`AzureBlobStorageAccount` / `AzureBlobStorageAccessKey`), i.e. the same storage account.

Every `Set` in the Data Lake provider finishes with `tempClient.Rename(...)` (`binarystore/azure/datalake/datalake.go:209`), and rename on ADLS Gen2 requires hierarchical namespace to be enabled on the account. A flat Blob Storage account will reject it, so if the existing CI account does not have HNS enabled then 100% of `binarystore.azure.datalake` conformance runs fail on the first write, not intermittently.

Could you confirm the account behind those secrets is HNS-enabled? If it is not, this probably wants its own secret pair (for example `AzureDataLakeStorageAccount` / `AzureDataLakeStorageAccessKey`) pointing at a dedicated account, otherwise the new job will be red from day one.

JoshVanL (changes requested):

binarystore/fake/fake.go
WhitWaldo: Added per the Copilot feedback as well

binarystore/in-memory/metadata.yaml
```diff
@@ -0,0 +1,17 @@
+# yaml-language-server: $schema=../../component-metadata-schema.json
+schemaVersion: v1
+type: binarystore
+name: in-memory
```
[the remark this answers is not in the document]
WhitWaldo: Addressed in recent commit

binarystore/oci/objectstorage/objectstorage.go
WhitWaldo: Addressed in f5518614

binarystore/aws/s3/s3.go
WhitWaldo: Already addressed in latest - thank you!

binarystore/azure/blobstorage/blobstorage.go
WhitWaldo: Already addressed in latest - thank you!

binarystore/azure/datalake/datalake.go
WhitWaldo: Already addressed - thank you!

binarystore/azure/datalake/datalake.go
WhitWaldo: Already addressed - thank you!

binarystore/aws/s3/metadata.yaml
```diff
@@ -0,0 +1,61 @@
+# yaml-language-server: $schema=../../../component-metadata-schema.json
+schemaVersion: v1
+type: binarystore
+name: aws.s3
```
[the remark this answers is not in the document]
WhitWaldo: Addressed in the dapr/dapr PR

binarystore/oci/objectstorage/objectstorage.go
```diff
@@ -0,0 +1,450 @@
...
+	req := transfer.UploadStreamRequest{
+		UploadRequest: transfer.UploadRequest{
+			NamespaceName:       &c.metadata.Namespace,
+			BucketName:          &c.metadata.BucketName,
+			ObjectName:          &name,
+			ObjectStorageClient: c.objectClient,
+			// Set explicitly so buffering matches the other binary store
+			// providers rather than the SDK defaults (10 MiB parts, 5
+			// goroutines).
+			PartSize:           common.Int64(binarystore.DefaultUploadPartSize),
```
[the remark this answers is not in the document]
WhitWaldo: Raised to 16 MiB in previous commit

binarystore/in-memory/in_memory.go
```diff
@@ -0,0 +1,139 @@
...
+// InMemoryBinaryStore is an in-memory BinaryStore implementation.
+type InMemoryBinaryStore struct {
+	logger logger.Logger
+	mu     sync.RWMutex
+	files  map[string][]byte
+	prefix string
+}
+
+// NewInMemoryBinaryStore returns a new in-memory BinaryStore.
+func NewInMemoryBinaryStore(log logger.Logger) binarystore.BinaryStore {
```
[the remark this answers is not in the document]
WhitWaldo: Fixed already

binarystore/azure/datalake/datalake.go
JoshVanL, line 190: `Concurrency: binarystore.DefaultUploadConcurrency` is `2` here, and that value triggers a data race inside the SDK that can silently truncate the stored file.

`UploadStreamOptions.Concurrency` defaults to **1** in `azdatalake v1.1.0` (`file/models.go`, `setDefaults()` sets `u.Concurrency = 1` when it is zero). The doc comment claiming a default of 5 is wrong, and the default of 1 is load-bearing: `copyFromReader` in `file/chunkwriting.go` declares the flush length at function scope

```go
actualSize := int64(0)
```

and then increments it *inside each chunk goroutine* with no lock and no atomic:

```go
go func(chunkNum uint32) {
    offset := int64(chunkNum) * options.ChunkSize
    appendDataOpts := options.getAppendDataOptions()
    actualSize += int64(len(buffer[:n]))   // unsynchronised
    _, err := dst.AppendData(ctx, offset, ...)
```

after which it flushes with that value once the goroutines are done (`wg.Wait()`, then `dst.FlushData(ctx, actualSize, flushOpts)`). With `Concurrency: 1` only one goroutine ever touches `actualSize`, so the bug is invisible. Setting it to 2 makes two run concurrently.

Concrete trigger, from your own conformance suite: the large-payload case uploads `2*DefaultUploadPartSize + 1MiB` = 33 MiB, which is 3 chunks of 16 MiB. Chunks 0 and 1 are in flight together, both read `actualSize == 0` and both store `16777216`, so the final flush length is roughly 24 MiB instead of 33 MiB. The data was appended, but `FlushData` commits a shorter file, `Set` returns `nil`, and the truncation is only visible on a later `Get`. It is a race, so it will present as an intermittent conformance failure rather than a deterministic one.

This is specific to Data Lake. `azblob` tracks committed size with an atomic `blockTracker`, and the S3, GCS and OCI paths have no equivalent shared counter, so `DefaultUploadConcurrency` is fine for them and does not need changing.

Suggested fix, keeping the shared constant for the other providers:

```go
uploadOpts := &file.UploadStreamOptions{
    ChunkSize: binarystore.DefaultUploadPartSize,
    // azdatalake's copyFromReader increments the flush length from inside
    // each chunk goroutine without synchronisation, so anything above 1
    // races and can commit a short file. See file/chunkwriting.go.
    Concurrency: 1,
}
```

If the throughput matters here, the alternative is to size the body in the component and drive `AppendData` / `FlushData` directly, where you control the length.

Makefile
JoshVanL, line 33: Moving to `go env GOHOSTOS` / `GOHOSTARCH` is a nice cleanup, but these two lines invert the meaning of `TARGET_OS_LOCAL` / `TARGET_ARCH_LOCAL`, and one existing recipe depended on the old meaning.

Before, they were derived from `uname` and were always the **host**, with `export GOOS ?= $(TARGET_OS_LOCAL)` flowing host to target. Now `TARGET_OS_LOCAL := $(GOOS)` flows the other way, so with `GOOS` set they become the **target**.

The `compile-build-tools` recipe builds a binary that the Makefile then executes locally:

```make
cd .build-tools; CGO_ENABLED=$(CGO) GOOS=$(TARGET_OS_LOCAL) GOARCH=$(TARGET_ARCH_LOCAL) go build -o $(BUILD_TOOLS_BIN) .
```

Trigger: on a Linux box with `GOOS=windows` exported (or `make GOOS=windows check-component-metadata`), this now cross-compiles `components-contrib-build-tools` for Windows and then immediately fails to run it. Previously it was always built for the host regardless of `GOOS`.

You already introduced the right variables, so the recipe just needs to use them:

```make
cd .build-tools; CGO_ENABLED=$(CGO) GOOS=$(HOST_GOOS) GOARCH=$(HOST_GOARCH) go build -o $(BUILD_TOOLS_BIN) .
```

You switched the two `ifeq (...,windows)` guards to `HOST_GOOS` for exactly this reason, so this looks like the one site that was missed.

tests/conformance/binarystore_test.go
JoshVanL, line 84: `shouldSkipBinaryStoreComponent` does not cover every env var the component YAMLs interpolate, so a partially configured developer machine gets a hard failure instead of a skip.

`tests/config/binarystore/aws/s3/binarystore.yaml` references `${{AWS_REGION}}`, but this case only checks `AWSS3Bucket`. `parseMetadataProperty` returns `required env var is not set` for an empty value, so setting `AWSS3Bucket` without `AWS_REGION` fails the test rather than skipping it.

The OCI case at line 94 has the same gap: the OCI YAMLs reference `${{OCIConfigProfile}}`, which is not in the `OCIConfigFile` / `OCICompartmentOCID` / `OCIBucketName` trio. CI happens to be safe only because the new workflow step defaults it:

```sh
if [ -z "${OCIConfigProfile:-}" ]; then
  echo "OCIConfigProfile=DEFAULT" >> "$GITHUB_ENV"
```

so this bites local runs, not the pipeline.

```go
case "aws.s3":
    if os.Getenv("AWSS3Bucket") == "" || os.Getenv("AWS_REGION") == "" {
        t.Skipf("Skipping AWS S3 conformance test: AWSS3Bucket and AWS_REGION environment variables must be set")
        return true
    }
```

and add `OCIConfigProfile` to the OCI condition.

.github/scripts/test-info.mjs
JoshVanL, line 21: `binarystore.azure.datalake` declares the same secrets as `binarystore.azure.blobstorage` (`AzureBlobStorageAccount` / `AzureBlobStorageAccessKey`), i.e. the same storage account.

Every `Set` in the Data Lake provider finishes with `tempClient.Rename(...)` (`binarystore/azure/datalake/datalake.go:209`), and rename on ADLS Gen2 requires hierarchical namespace to be enabled on the account. A flat Blob Storage account will reject it, so if the existing CI account does not have HNS enabled then 100% of `binarystore.azure.datalake` conformance runs fail on the first write, not intermittently.

Could you confirm the account behind those secrets is HNS-enabled? If it is not, this probably wants its own secret pair (for example `AzureDataLakeStorageAccount` / `AzureDataLakeStorageAccessKey`) pointing at a dedicated account, otherwise the new job will be red from day one.

binarystore/fake/fake.go
```diff
@@ -0,0 +1,30 @@
...
+package fake
+
+import (
+	"github.com/dapr/components-contrib/binarystore"
+	inmemory "github.com/dapr/components-contrib/binarystore/in-memory"
+	"github.com/dapr/kit/logger"
+)
+
+// NewFake returns a fake BinaryStore backed by the in-memory implementation.
+func NewFake(log logger.Logger) binarystore.BinaryStore {
```
JoshVanL, line 28: Design note rather than a defect: this is a stub, not a fake in the sense the rest of Dapr uses the word, and as written it does not add anything over importing the in-memory store directly.

Dapr's convention (for example `dapr/pkg/channel/fake`, `dapr/pkg/actors/fake`) is a `*Fake` struct holding one function field per interface method, with `WithXxx` builders:

```go
type Fake struct {
    fnGetAppConfig func(ctx context.Context, appID string) (*config.ApplicationConfig, error)
    fnInvokeMethod func(...)
}

func New() *Fake { ... }
func (f *Fake) WithGetAppConfig(fn ...) *Fake { ... }
```

That shape exists so a test can make a method fail on demand. `NewFake` here returns `binarystore.BinaryStore`, so a caller gets the concrete in-memory behaviour with no way to override it, and there is no way to exercise a consumer's error handling for a failing `Set` or `Get`.

Two observations that follow from that:

- For the "I just need a working store in a test" case, `inmemory.NewInMemoryBinaryStore(log)` already does the job, so the extra package and its test are redundant with `binarystore/in-memory`.
- For the failure-path case, which is the usual reason a `fake` package exists, this one cannot help.

Worth noting this is the only `fake` package in components-contrib today, so it is setting the precedent for the repo rather than following one.

If dapr/dapr genuinely only needs a working store for wiring tests then this is fine as-is and worth a sentence in the package doc saying so. If the intent is to follow the fake convention, it wants the function-field shape and a concrete `*Fake` return type.

binarystore/fake/fake_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/dapr/components-contrib/binarystore"
+	"github.com/dapr/kit/logger"
+)
+
+func TestNewFake(t *testing.T) {
+	store := NewFake(logger.NewLogger("test"))
+	require.NotNil(t, store)
+
+	var _ binarystore.BinaryStore = store
```
JoshVanL, line 32: This line asserts nothing, and it is one of the four issues currently failing `Build linux_amd64 binaries` (staticcheck QF1011).

`NewFake` is declared as returning `binarystore.BinaryStore`, so `store` already has that type. The compiler checks the interface is satisfied at the `return inmemory.NewInMemoryBinaryStore(log)` inside `NewFake`, not here, so this adds no coverage.

Staticcheck's literal suggestion is to drop the type from the declaration, but that leaves a `var _ = store` that is just as empty. Deleting the line outright is the better fix:

```go
func TestNewFake(t *testing.T) {
	store := NewFake(logger.NewLogger("test"))
	require.NotNil(t, store)
}
```

A compile-time interface assertion is worth keeping when the constructor returns a concrete type, in which case it belongs in `fake.go` as `var _ binarystore.BinaryStore = (*Fake)(nil)`. That would apply if you take up the suggestion on `fake.go:28`.

JoshVanL (changes requested):

nelson-parente (commented):
I went through the whole PR at 9b85810 with Josh's threads in one hand, then looked for what they do not cover. First the thread state, because the PR page overstates what is left.

**Josh's threads.** Seven of his unique asks are already fixed at this head but the threads were never resolved: the S3 403 fallthrough on `Delete`, the OCI secret names, the `mdonly` tag on `prefix`, the create-without-overwrite test, the `profile: prefix` conformance configs, the in-memory `prefix` decode, and the redundant OCI `HeadObject`. @WhitWaldo, resolve those with a one-line pointer to the commit so the delta is visible. Six remain open and I agree with all six: the Data Lake `Concurrency: 2` truncation race, the Makefile `TARGET_OS_LOCAL` inversion, the `shouldSkipBinaryStoreComponent` env coverage, the Data Lake account sharing Blob Storage secrets without hierarchical namespace, the registration checker counting `binarystore/fake`, and the `fake_test.go:32` assertion that is failing the build. The build is red on four lint lines only:

```
binarystore/errors_test.go:50:4: error-is-as: use assert.NotErrorIs (testifylint)
binarystore/errors_test.go:68:2: expected-actual: need to reverse actual and expected values (testifylint)
binarystore/fake/fake_test.go:32:8: QF1011: could omit type binarystore.BinaryStore from declaration (staticcheck)
bindings/azure/blobstorage/blobstorage_prefix_test.go:129:37: fmt.Sprintf can be replaced with strconv.Itoa (perfsprint)
```

New findings, none of them in Josh's list:

**1. `WorkflowType` is back.** `metadata/utils.go` re-adds `WorkflowType ComponentType = "workflows"` and lists it in `IsValid`. Josh removed it on main in #4478 with the workflows component abstraction. It came back in the "Addressing Copilot reviews" commit after the merge from main, and nothing references it. Drop it.
https://github.com/dapr/components-contrib/blob/9b858106d10c3b30341ae5e3083fe30e0e1bf333/metadata/utils.go#L152-L165

**2. `ObjectPath` is the one place every provider agrees on a key, and it has two gaps.** With a prefix it strips a leading `/` from the file name. Without a prefix it returns the name untouched, so `/x.bin` is stored as `/x.bin` today and as `p/x.bin` the day someone adds `prefix: p`, which orphans the old objects. It also never rejects `..` or empty segments. For S3, GCS, OCI and Blob a key like `tenant-a/../tenant-b/secret` is a literal, ugly, contained string. Data Lake is a real hierarchical namespace, and that string goes straight into `NewFileClient`, `Create` and `Rename`. I did not confirm from the docs whether the service resolves `..` server-side, and that is the point: the interface promises an identifier, so validate it once here. Reject `..` and empty segments, apply the same leading-slash rule with and without a prefix, and add the tests. That also answers Mike's isolation thread on the proposal, which the `prefix` description in all six `metadata.yaml` files leaves open by calling it "an optional prefix" without saying whether it is a boundary.
https://github.com/dapr/components-contrib/blob/9b858106d10c3b30341ae5e3083fe30e0e1bf333/binarystore/metadata.go#L27-L33

**3. The Data Lake temp file survives a crash.** `putObject` uploads to `<name>.tmp-<uuid>` and renames with `IfNoneMatch`. The comment says a failed upload never leaves a partial file behind. That holds while the process is alive, because `cleanupFile` runs from the error branches. If daprd dies between the upload and the rename, the temp object stays forever, with no sweep and no mention in `metadata.yaml`. Say so in the docs and give the temp names a fixed prefix operators can list and purge.
https://github.com/dapr/components-contrib/blob/9b858106d10c3b30341ae5e3083fe30e0e1bf333/binarystore/azure/datalake/datalake.go#L169-L215

**4. The S3 IAM guidance names the wrong permission for the 404.** `metadata.yaml` says `s3:GetObject` is what `Delete` needs to report whether the object existed. `HeadObject` does need `s3:GetObject`, but what turns a missing key into a 404 instead of a 403 is `s3:ListBucket`, which is what `isAccessDenied` and the test comment say. A customer scoping a policy from the yaml gets the fallthrough path every time. List both.
https://github.com/dapr/components-contrib/blob/9b858106d10c3b30341ae5e3083fe30e0e1bf333/binarystore/aws/s3/metadata.yaml#L18-L25
https://github.com/dapr/components-contrib/blob/9b858106d10c3b30341ae5e3083fe30e0e1bf333/binarystore/aws/s3/s3.go#L308-L313

**5. Component type versus the proposal.** The yamls declare `type: binarystore` and the runtime registers `binarystore.<provider>`, while dapr/proposals#88 says `state.binarystore.<provider>` and the runtime's HTTP route is `state/binarystore/...`. The top-level type is consistent with `conversation.*` and `crypto.*`, so I lean that way, but then the proposal and the route need to follow. Decide it now while it is alpha.

**6. Scope.** Four changes here have nothing to do with the block: the Makefile host-detection rewrite (where the inversion lives), the `conformance.yml` secret-fetch rewrite to heredocs with masking (which changes secret retrieval for every component's conformance run), and the flake fixes in `lock/redis/standalone_test.go` and `common/authentication/kubernetes/client_test.go`. Split them into their own PRs. They can land this week, and this PR shrinks for Josh. The `prefix` support on the existing Blob binding and state store is fine to keep: the shared metadata struct forces it, and blob names are byte-identical when `prefix` is unset.

**7. What the unit tests prove.** Every provider test swaps the SDK client for a fake, so the real single-part, multipart and rename code runs only in the conformance job, which needs cloud secrets. The `storetest` comment about exercising chunked upload paths is not true against the fakes. Not a blocker. Just do not read green unit tests as coverage of the upload paths. The conformance run is the evidence, which makes Josh's Data Lake account question matter more.

The `Overwrite: false` path is the part I was most worried about, and it is right: all five cloud providers pass `If-None-Match: *` or `DoesNotExist` through to the server for both single and multipart uploads, so the check is atomic everywhere. No docs PR yet.

nelson-parente (commented):
Posted twice by mistake. The review is the one above: https://github.com/dapr/components-contrib/pull/4559#pullrequestreview-5308197493

PR #4569, opened 2026-09-03
feat(kafka): add maxPollInterval config option

### LINKED ISSUE #2957: Kafka allow configuring max_poll_interval_ms
Dapr should allow setting max_poll_interval_ms for kafka. It will be helpful for long time processing messages.

magic-peach:
# Description

Dapr's Kafka pubsub and binding components already expose heartbeatInterval and sessionTimeout, but there was no way to configure how long the consumer group coordinator waits for a member to rejoin before triggering a rebalance. This matters for consumers with long running message processing, since a slow handler can cause the consumer to be kicked out of the group and reprocessing to start elsewhere.

This PR adds a maxPollInterval metadata option to the Kafka pubsub and binding components, mapped to Sarama's Consumer.Group.Rebalance.Timeout, which is the closest equivalent Sarama exposes to Kafka's max.poll.interval.ms in the Java client. It defaults to 60s, the same as Sarama's own built in default, so existing deployments are unaffected unless they opt in.

## Issue reference

Please reference the issue this PR will close: #2957

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
    * [ ] Created the dapr/docs PR: not yet opened, the metadata.yaml files in this PR are the source for the generated component reference docs

### DIFF
--- a/bindings/kafka/metadata.yaml
+++ b/bindings/kafka/metadata.yaml
@@ -366,6 +366,13 @@ metadata:
       The maximum time between heartbeats before the consumer is considered inactive and will timeout.
     example: '"20s"'
     default: '"10s"'
+  - name: maxPollInterval
+    type: duration
+    description: |
+      The maximum time the consumer group coordinator will wait for a member to rejoin the group, allowing
+      for longer message processing times before the consumer is considered dead and a rebalance is triggered.
+    example: '"5m"'
+    default: '"60s"'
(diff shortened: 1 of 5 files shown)


cicoyle (commented):
Hi, this mostly lgtm, but can you link to the docs PR for it? That is a pre-req to getting merged. Thank you 🙏🏻

PR #4574, opened 2026-09-07
feat(pubsub/redis): add streamStartID metadata option

### LINKED ISSUE #3256: Support to set the last delivered entry for Redis Stream Pubsub
## Describe the feature

Currently redis stream pubsub use "0" as the last delivered entry ID in the stream, here is the related [code](https://github.com/dapr/components-contrib/blob/79adc565c17ad8936048896591cd205a6609ad67/pubsub/redis/redis.go#L113). So this means it hardcode the consumer MUST receive messages in the specified stream from the very first message. 

There are some sceneries that the consumer want to specify the ID of the last entry in the stream to start to sync from, for example to set the `$` but not 0 for the ID to ask sync from the last ID of the entry in the stream. 

The request is to add one more config parameter in the component of type `pubsub.redis` to allow the human to setup the ID of the last entry in the stream to sync from.

## Release Note

ADD one more parameter in the pubsub.redis config to support setup the ID of the last entry in the stream.

RELEASE NOTE:

magic-peach:
# Description

The redis streams pubsub always created new consumer groups starting from stream entry ID 0, so a consumer group always had to replay the entire stream history before receiving new messages. This adds a streamStartID metadata option so callers can set it to $ to only receive messages published after the group is created, while defaulting to 0 to keep existing deployments unaffected.

## Issue reference

Please reference the issue this PR will close: #3256

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
    * [ ] Created the dapr/docs PR: not yet opened, the metadata.yaml file in this PR is the source for the generated component reference docs

**Note:** We expect contributors to open a corresponding documentation PR in the dapr/docs repository. As the implementer, you are the best person to document your work! Implementation PRs will not be merged until the documentation PR is opened and ready for review.

### DIFF
--- a/common/component/redis/redis.go
+++ b/common/component/redis/redis.go
@@ -139,6 +139,7 @@ func ParseClientFromProperties(properties map[string]string, componentType metad
 		settings.RedeliverInterval = 15 * time.Second
 		settings.QueueDepth = 100
 		settings.Concurrency = 10
+		settings.StreamStartID = "0"
 	}
 
 	err := settings.Decode(properties)
--- a/common/component/redis/settings.go
+++ b/common/component/redis/settings.go
@@ -113,6 +113,8 @@ type Settings struct {
 	QueueDepth uint `mapstructure:"queueDepth" mdonly:"pubsub"`
 	// The number of concurrent workers that are processing messages
 	Concurrency uint `mapstructure:"concurrency" mdonly:"pubsub"`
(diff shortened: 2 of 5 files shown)


cicoyle (commented):
Hi, thanks for the PR! 

This mostly lgtm, but can you link to the docs PR for it? That is a pre-req to getting merged. Thank you 🙏🏻

PR #4587, opened 2026-09-15
feat(conversation/anthropic): migrate to anthropic-sdk-go v1.72.0

### LINKED ISSUE #4533: conversation/anthropic: pinned langchaingo blocks Claude models newer than the 4.6 generation

vaishaldsouza:
# Description

Migrates the `conversation.anthropic` component from the pinned `github.com/tmc/langchaingo` wrapper to `github.com/anthropics/anthropic-sdk-go` directly, bypassing `conversation/langchaingokit` for this component only.

## Problem

The pinned langchaingo dependency has two bugs that cap this component at Claude 4.6-generation models:

1. `temperature` is always serialized (missing `omitempty`), so every request sends `"temperature": 0` even when unset. Claude 4.7+ models reject this with `400: temperature is deprecated for this model`.
2. Bumping langchaingo to a newer commit removes `tool_choice` support entirely, so a fix for (1) would silently lose explicit tool choice.

langchaingo has no active maintainers, so fixing upstream and bumping is not viable (see discussion on the linked issue).

## Changes

- `conversation/anthropic/anthropic.go`: rewritten to use `*anthropic.Client` directly instead of embedding `langchaingokit.LLM`.
  - `Temperature` is only set on the request when `> 0`, otherwise omitted — fixes the 4.7+ 400 error.
  - `tool_choice` mapped explicitly to `anthropic.ToolChoiceUnionParam` (auto/any/none/named tool) — restores what langchaingo's newer commit dropped.
  - `PromptCacheRetention` (`*time.Duration`) mapped to the SDK's `CacheControlEphemeralParam` TTL, rounding to the nearest supported value (`5m` or `1h`) since the SDK only supports those two.
  - Response conversion (text/tool_use content blocks, stop reason, usage/cache token accounting) reimplemented against the SDK's native types.
- `conversation/anthropic/metadata.go`: updated for the new client config.
- `conversation/langchaingokit`: unchanged — other components (openai, bedrock, mistral, etc.) still depend on it.
- `go.mod`/`go.sum`: added `github.com/anthropics/anthropic-sdk-go`.

## Testing

- `go build ./...`, `go vet ./...` clean.
- Unit tests in `conversation/anthropic` pass.
- `golangci-lint run` (pinned v2.10.1, matching CI's `go.mod` Go version) reports 0 issues.
## Issue reference

Fixes #4533

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


sicoyle:
Hi @vaishaldsouza, thank you so much for taking the time to put this together, and for your interest in the Dapr open source project and this issue in particular! 🙏

This issue is already being addressed in #4535, which another contributor opened earlier and which takes the same approach. I'm going to close this one as a duplicate to keep the effort in one place.

One piece I really liked here is the PromptCacheRetention → cache-control TTL mapping. If you're interested, please feel free to leave a review or suggestion on #4535 for that part.

We'd love to have you contribute elsewhere too! The good first issue and help wanted labels are a great place to start. Before picking one up, commenting /assign on the issue and checking that there isn't already an open PR for it helps avoid overlap like this.

If you'd like help finding another good issue, just let me know. I'd be happy to help! I help maintain Dapr Agents, the Python SDK, and Components Contrib, and there are always plenty of fun problems to solve 🤗

Thanks again!

PR #4589, opened 2026-09-15
fix(state/sqlserver): use pointer receiver for KeysLike so *SQLServer actually implements state.KeysLiker

### LINKED ISSUE #4588: SQL Server state store's KeysLike doesn't actually implement state.KeysLiker (wrong receiver type)
## Expected Behavior

The SQL Server state store (`state/sqlserver/sqlserver.go`) should support wildcard key listing via `state.KeysLiker`, the same way `etcd`, `in-memory`, `mongodb`, `mysql`, `redis`, `sqlite`, and `postgresql/v2` all do.

## Actual Behavior

`*SQLServer` silently does not implement `state.KeysLiker` at all, even though `KeysLike` is fully implemented (a working SQL `LIKE` query with pagination, `state/sqlserver/sqlserver.go:392`).

`state.KeysLiker` declares:
```go
type KeysLiker interface {
	KeysLike(ctx context.Context, req *KeysLikeRequest) (*KeysLikeResponse, error)
}
```
— a **pointer** parameter. SQLServer's implementation takes a **value** parameter instead:
```go
func (s *SQLServer) KeysLike(ctx context.Context, req state.KeysLikeRequest) (*state.KeysLikeResponse, error) {
```

Go requires an exact method-signature match for interface satisfaction, so this mismatch means `*SQLServer` never actually satisfies `state.KeysLiker`, despite the method body looking complete and correct. Any Dapr app using the SQL Server state store cannot use wildcard key listing through the interface-based dispatch the runtime actually uses.

Confirmed SQL Server is the sole outlier — every other provider implementing `KeysLike` (`etcd`, `in-memory`, `mongodb`, `mysql`, `redis`, `sqlite`, `postgresql/v2`) uses the correct pointer signature.

This also explains why it went undetected: SQL Server's tests have no compile-time interface assertion (`var _ state.KeysLiker = ...`), unlike `mongodb`, `mysql`, `in-memory`, and `postgresql/v2`, which all have exactly that guard — the safety net that would have caught this at compile time.

1. In any code that type-asserts or type-switches on `state.KeysLiker` (e.g. the shared conformance suite in `tests/conformance/state/state.go`), pass a `*sqlserver.SQLServer` instance.
2. The type assertion fails — `*SQLServer` does not satisfy `state.KeysLiker` — even though `KeysLike` exists on the type.
3. Alternatively: add `var _ state.KeysLiker = (*SQLServer)(nil)` anywhere in the `sqlserver` package pre-fix; it fails to compile.

RELEASE NOTE: **FIX** SQL Server state store's `KeysLike` now correctly implements `state.KeysLiker` (was using the wrong receiver type, so it silently didn't implement the interface at all).

karthikchundi-commits:
# Description

Fixes #4588.

`state.KeysLiker` declares `KeysLike(ctx, *KeysLikeRequest)` — a pointer parameter. SQL Server's implementation took a value parameter (`state.KeysLikeRequest`) instead. Go requires an exact method-signature match for interface satisfaction, so `*SQLServer` silently did not implement `state.KeysLiker` at all, despite the method body being fully built and correct (a working SQL `LIKE` query with pagination). Any Dapr app using the SQL Server state store could not use wildcard key listing through the interface-based dispatch the runtime actually uses, even though the implementation was sitting right there.

Confirmed SQL Server was the sole outlier by checking every other provider implementing `KeysLike`: `etcd`, `in-memory`, `mongodb`, `mysql`, `redis`, `sqlite`, `postgresql/v2` all use the correct pointer signature.

This went undetected because SQL Server's tests had no compile-time interface assertion (`var _ state.KeysLiker = ...`), unlike `mongodb`, `mysql`, `in-memory`, and `postgresql/v2`, which all have exactly that guard. Added the same assertion here, plus a test that calls `KeysLike` through a `state.KeysLiker`-typed variable (not a concrete `*SQLServer` one) — the thing that would fail to compile at all against the pre-fix signature.

**Verification limitation, disclosed honestly:** no Go toolchain is available in the environment I authored this in, so I could not run `go build`/`go vet`/`go test` to confirm the change compiles. I verified instead by reading the interface declaration and comparing byte-for-byte against every other implementer's signature, confirming every real call site in the repo (`sqlite`'s delegation, the full conformance suite in `tests/conformance/state/state.go`) already passes a pointer so nothing else needed to change, and hand-checking gofmt-consistent formatting against the surrounding file. Would appreciate a maintainer or CI confirming the build before merge.

## Issue reference

Closes #4588.

## Checklist

* [ ] Code compiles correctly — **not independently verified locally (no Go toolchain available); please confirm via CI**
* [x] Created/updated tests — added the compile-time `state.KeysLiker` assertion and a test calling `KeysLike` through the interface type
* [ ] Extended the documentation — not applicable, no user-facing behavior change beyond the interface now being satisfied as originally intended


state/sqlserver/sqlserver.go
cicoyle, line 72: Can you add the feature to the feature list that gets advertised so this functionality is usable?
```suggestion
			state.FeatureTTL,
			state.FeatureKeysLike,
```

state/sqlserver/sqlserver_test.go
cicoyle, line 530: I would expect this test func needs to be updated as well to verify the feature is there and supported.

state/sqlserver/sqlserver.go
```diff
@@ -389,7 +389,7 @@ func (s *SQLServer) CleanupExpired() error {
 	return nil
 }
 
-func (s *SQLServer) KeysLike(ctx context.Context, req state.KeysLikeRequest) (*state.KeysLikeResponse, error) {
+func (s *SQLServer) KeysLike(ctx context.Context, req *state.KeysLikeRequest) (*state.KeysLikeResponse, error) {
```
cicoyle: The PR title says `use pointer receiver` - the receiver is already correct as `(s *SQLServer)`, its just the parameter that changed. Can you update the PR title?

cicoyle (changes requested):
Thanks for the PR here 🎉 

The signature mix match is a real issue, thanks for fixing it. I dont think the one line change is safe on its own though, especially given the title says `so *SQLServer actually implements state.KeysLiker` - this code hasnt been compiled against or tested yet.

The PR title says `use pointer receiver` - the receiver is already correct as `(s *SQLServer)`, its just the parameter that changed. Can you update the PR title? 

I would like to see more from this if you have the time 🙏🏻 

Can you add the feature (`"keyslike"`) to the following: https://github.com/dapr/components-contrib/blob/main/tests/config/state/tests.yml#L46-L65

We should have thorough conformance tests as well. You shouldn't have too write a ton of new tests, but rather extend the suite to test this as well to exercise the actual SQL against a real SQL server, which is the verification we want :)

karthikchundi-commits:
@cicoyle Both addressed:

- Title updated — you're right, the receiver was already `(s *SQLServer)`; it's the `KeysLike` *parameter* that went from value to pointer.
- Added `"keyslike"` to the `sqlserver` and `sqlserver.docker` conformance entries in `tests/config/state/tests.yml`, wiring this up to `tests/conformance/state/state.go`'s real KeysLike suite (pattern matching, pagination, continuation tokens) against an actual SQL Server instance. Left `sqlserver.v2`/`.v2.docker` alone — that's a separate package that doesn't implement `KeysLike` at all.

Wiring that in surfaced a second, related gap while I was at it: `SQLServer`'s features list never included `state.FeatureKeysLike`, even with the interface satisfied — the method was reachable but the store never told callers it supported it, same silent-gap shape as the original bug, one layer up. Fixed that too, with a test against the real `New()`-constructed store (the existing `TestSupportedFeatures` test builds its own hand-picked features slice, which is exactly why this didn't get caught).

Same disclosure as before: no Go toolchain in my environment, so still unverified against a live SQL Server or even compiled locally — would appreciate CI or a maintainer confirming.

PR #4591, opened 2026-09-15
fix(ci): start RavenDB without the expired license

nelson-parente:
# Description

The `state.ravendb` conformance and certification jobs fail on every pull request. They are not related to the change under test.

Both compose files carry a RavenDB license that **expired on 5 August 2026**, and both set `RAVEN_License_ThrowOnInvalidOrMissingLicense=true`. The server therefore refuses to start:

```
Raven.Server.Commercial.LicenseExpiredException: The RavenDB server cannot start due to a missing license.
- The license 'b75b6995-5f1f-440f-bcaa-40f7388ecf90' obtained from 'License' has an expiration date
  of '05 August 2026' and is also expired.
```

`docker compose up -d` still reports `Container RavenDb Started`, so the setup step passes and the tests then run against a server that is not listening. That is why the jobs report topology errors rather than a license error:

```
Failed to set up RavenDB database "testdapr": failed to load database record:
  There are no nodes in the topology at all
error saving changes No known topology and no previously known one, cannot proceed, likely a bug
```

## Why the jobs went red on 2 September, not on 5 August

The image tag `ravendb/ravendb` is not pinned. The scheduled conformance run of 30 August passed, and the run of 2 September failed. The image in use was built on **1 September 2026**:

```
ravendb/ravendb@sha256:6a091483a4caec3b365f089f5098d64658953ec1efcdfab621e508ac4b730035
created=2026-09-01T09:12:55Z
```

So the license had been expired and tolerated for four weeks. The 1 September image began to enforce it, and every run since then fails. An unpinned tag is what let this land overnight with no repository change.

## Change

Remove the expired license from both compose files, and set `ThrowOnInvalidOrMissingLicense` to `false`.

RavenDB then runs in AGPL open-source mode. Its own `/license/status` endpoint confirms this, and this mode does not expire:

```json
{"Status": "AGPL - Open Source", "Expired": false, "Expiration": null,
 "MaxMemory": 6, "MaxCores": 3, "MaxClusterSize": 1}
```

The state store uses one RavenDB feature beyond plain document storage: document expiration, for TTL. AGPL mode includes it. The conformance case `set_and_get_with_TTL` passes. Nothing in `state/ravendb/` touches encryption, ETL, replication, snapshot backups or clustering, which are the capabilities this mode withholds.

```diff
-      - RAVEN_LICENSE={"Id":"b75b6995-...","Name":"Vega IT","Keys":[...]}
       - RAVEN_DATABASE=testdapr
       - RAVEN_Setup_Mode=None
       - RAVEN_License_Eula_Accepted=true
       - RAVEN_Security_UnsecuredAccessAllowed=PrivateNetwork
-      - RAVEN_License_ThrowOnInvalidOrMissingLicense=true
+      - RAVEN_License_ThrowOnInvalidOrMissingLicense=false
```

## This does not change the edition under test

The expired license stopped having any effect on 5 August. RavenDB 7.1.11, the release the last green run used, starts with it and silently falls back to open-source mode:

```
image: ravendb/ravendb:7.1.11, with the expired license and ThrowOnInvalidOrMissingLicense=true
Status    : AGPL - Open Source
Expired   : False
LicensedTo: None
```

So the suites have already been running in AGPL mode for five weeks, and they passed. 7.1.12 and 7.2.6 changed only the handling of an invalid license: refuse to start, rather than fall back.

This PR therefore does not downgrade what is tested. It makes the existing behavior explicit and deletes a key that does nothing.

## Verification

Started through the CI script itself, `.github/scripts/components-scripts/docker-compose.sh ravendb`. The server reports a healthy topology about 10 seconds after start:

```
{"Promotables":[],"Nodes":[{"Url":"http://127.0.0.1:8080","ClusterTag":"A","ServerRole":"Member","Database":"testdapr"}],"Etag":3}
```

Conformance passes in full:

```
ok  github.com/dapr/components-contrib/tests/conformance  3.635s
```

Certification passes every RavenDB operation. Its last step, `Interrupt network`, calls `sudo pfctl` and cannot run on macOS, so that one step is unverified locally. It is unaffected by this change.

## Relationship to #4582

#4582 touches the same two compose files, so the two conflict.

The approaches differ. #4582 moves the license to the `RAVENDBLICENSE` secret and adds `requireRavenDBCredentials` to the cloud-only set, which stops RavenDB running on ordinary pull requests. That removes the red check by skipping the test. The `RAVENDBLICENSE` secret was last updated on 26 January 2026, before the license expiry date, so a `/ok-to-test` run can still fail unless a current key is loaded into it.

This PR keeps the tests running on every pull request and needs no secret. If a current license is obtained later, restoring the license variable is a one-line change.

I am happy to close this in favor of #4582 if the maintainers prefer the secret route.

## Issue reference

No separate issue. The failing jobs, the root cause and the local reproduction are described above.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests — no new tests. This repairs the environment for the existing conformance and certification suites.
* [ ] Extended the documentation — no user-facing change. The component itself is untouched.


sicoyle (approved):
lgtm - thank you!

mikeee:
This should be reverted in favour of https://github.com/dapr/components-contrib/pull/4582

PR #4597, opened 2026-09-17
Revert "fix(ci): start RavenDB without the expired license"

mikeee:
Reverts dapr/components-contrib#4591

Failing CI shouldn’t ordinarily be hidden as this test isn’t monitored or tracked in another issue. A new secret has been issued and #4582 fixes secret loading


mikeee:
Actually, this should still be reverted as we do not want tests to silently fail. This licence key failure would not be surfaced.

JoshVanL (approved):

nelson-parente (approved):

Issue #3568, opened 2024-10-16
EventHub component delivers out of order messages (pubsub, binding)


oising:
## Expected Behavior

When using Event Hubs as a pubsub or binding, messages should be delivered in the order they were posted (assume PartitionKey is set when publishing/posting to ensure ordering across partitions.)

## Actual Behavior

In the pubsub case, the sidecar delivers new events before the subscriber has completed handling the last one. This causes major problems when trying to ensure order sensitive work is executed correctly (e.g. starting a workflow to process subsequent events.) 

## Steps to Reproduce the Problem

We're publishing to our topic like this (dotnet sdk):

```csharp

await dapr.PublishEventAsync(
        WellKnownPubSubNames.Percept,
        WellKnownPubSubTopics.RecordingSessionFileEventV1, 
        data,
        new Dictionary<string, string>
        {
            { "partitionKey", data.DeviceId }
        }
    );
```

and receiving like this:

```csharp
pubsub.MapPost(
    WellKnownPubSubTopics.RecordingSessionFileEventV1,
    [Topic(WellKnownPubSubNames.Percept, WellKnownPubSubTopics.RecordingSessionFileEventV1)]
    async (
        [FromServices] ILogger<Program> logger,
        [FromServices] IRecordingSessionRepository recordingSessionRepository,
        HttpContext HttpContext,
        [FromBody] RecordingSessionFileEventV1 data
    ) => {
        try {
            var messageId = HttpContext.Request.Headers["Message-Id"];

            logger.LogInformation("Start message id: {messageId}", messageId!);

            await Task.Delay(500);

            logger.LogInformation("Stop message id: {messageId}", messageId!);

            return Results.Ok();
        }
        catch (Exception ex) {
            logger.LogError(ex);
            return Results.Problem(ex.Message);
        }
    }
);
```

The problem is clear when watching the logs: instead of seeing a constant start/stop/start/stop alternating sequence of log events, we're seeing start/stop/start/start/stop/stop interleaving. The sidecar should not be sending another event until the current one has completed processing, i.e. it receives a http 200 (in this case.)

The same issue likely occurs for the binding since the common code is the problem (according to @yaron2):

[...]

PR #4598, opened 2026-09-17
Fix Azure Event Hubs checkpoint safety

### LINKED ISSUE #4561: pubsub.azure.eventhubs: checkpoint advances regardless of app response, breaking the documented at-least-once guarantee
## Expected Behavior

Per the [pub/sub overview](https://docs.dapr.io/developing-applications/building-blocks/pubsub/pubsub-overview/#at-least-once-guarantee):

> Dapr guarantees at-least-once semantics for message delivery. [...] Even if the message fails to deliver, or your application crashes, Dapr attempts to redeliver the message until successful delivery.
>
> All Dapr pub/sub components support the at-least-once guarantee.

A subscriber that returns `RETRY` (or whose sidecar dies mid-processing) should therefore see the message again after a restart. `pubsub.azure.eventhubs` should not advance the consumer-group checkpoint past a message the application never successfully processed.

This is how the sibling components already behave:

- **Kafka** — [`common/component/kafka/consumer.go#L221-L223`](https://github.com/dapr/components-contrib/blob/ece1f255da3313670729c6ef11c6b65108537a5c/common/component/kafka/consumer.go#L221-L223): `session.MarkMessage()` is guarded on `err == nil`.
- **Service Bus** — per-message `CompleteMessage` / `AbandonMessage`, plus a native DLQ.

## Actual Behavior

The Event Hubs component advances the checkpoint **unconditionally**, ignoring the application's response. The consequence is **at-most-once** delivery across a sidecar restart: the unprocessed event is permanently lost, with no error surfaced to the publisher or the subscriber.

[`processEvents`](https://github.com/dapr/components-contrib/blob/ece1f255da3313670729c6ef11c6b65108537a5c/common/component/azure/eventhubs/eventhubs.go#L399-L422) — the handler's error is discarded in *both* branches, and `UpdateCheckpoint` then runs regardless:

```go
if aeh.metadata.EnableInOrderMessageDelivery {
    _ = aeh.handleAsync(subscribeCtx, config.Topic, events, config.Handler)   // error dropped
} else {
    go aeh.handleAsync(subscribeCtx, config.Topic, events, config.Handler)    // fire-and-forget
}

// Checkpointing disabled for CheckPointFrequencyPerPartition == 0
if config.CheckPointFrequencyPerPartition > 0 {
    if counter%config.CheckPointFrequencyPerPartition == 0 {
        err = partitionClient.UpdateCheckpoint(ctx, events[len(events)-1], nil)   // runs unconditionally
```

The error is available at the call site — [`handleAsync`](https://github.com/dapr/components-contrib/blob/ece1f255da3313670729c6ef11c6b65108537a5c/common/component/azure/eventhubs/eventhubs.go#L344-L358) logs it and returns it. It is simply not consulted.

Both delivery modes are affected, to different degrees:

- `enableInOrderMessageDelivery: false` (**default**) — delivery is `go handleAsync(...)`, so the checkpoint advances without waiting for the app at all. The cursor can run arbitrarily far ahead of a stalled subscriber.
- `enableInOrderMessageDelivery: true` — the component waits for the app, which bounds the loss to one batch, but a `RETRY` still checkpoints.

**There is no workaround.** No component or subscription metadata field makes checkpointing conditional on app success, and no Resiliency policy helps — the inbound retry runs entirely inside the handler, and the checkpoint follows whether it succeeded or not. Setting `checkPointFrequencyPerPartition: 0` looks like an escape hatch but is strictly worse: with no checkpoint written, the SDK's start position falls back to `@latest`, so a restart skips everything enqueued while the sidecar was down.

Reproduced end to end on the [Azure Event Hubs emulator](https://github.com/Azure/azure-event-hubs-emulator-installer) with Azurite as the checkpoint store, so this needs no Azure subscription. Dapr 1.18.4 self-hosted, single-partition hub `orders`, consumer group `cg1`, component defaults (`enableInOrderMessageDelivery: false`, `checkPointFrequencyPerPartition` unset → 1), and an app that returns HTTP 500 `{"status":"RETRY"}` for every delivery.

1. Publish **A**. The app receives it and returns `RETRY`:

```
[run1] APP-RECEIVED -> returning RETRY body={"data":{"orderId":"EH-MSG-1"}, ...}
level=warning msg="retriable error returned from app while processing pub/sub event ... status code returned: 500"
level=warning msg="Error processing EventHubs events for topic orders ... Retrying..."
```

2. Restart the sidecar. **A is never redelivered.**

3. Publish **B** on the restarted sidecar. It is delivered immediately:

```
[run2] APP-RECEIVED -> returning RETRY body={"data":{"orderId":"EH-MSG-2-control"}}
```

Step 3 is the control: it rules out a dead subscription or an `@latest` start-position artifact. The subscription is healthy — **A** was skipped because its checkpoint had already advanced.

4. The checkpoint blob confirms it. Reading Azurite's blob metadata directly for `localhost/orders/cg1/checkpoint/0`:

```json
{"offset": "528", "sequencenumber": "1"}
```

Both messages were checkpointed despite the app having acknowledged neither. The cursor tracks the newest event received while the application succeeded at nothing.

**Observed:** the app sees **B** only. **A** is unrecoverable, and no error surfaces to publisher or subscriber.
**Expected:** **A** is redelivered, because the app never acknowledged it.

### Deterministic coverage

Two tests that need no broker at all, if a regression harness is wanted:

- `TestHandleAsyncSurfacesHandlerError` — the app's failure does reach the caller; `handleAsync` returns it verbatim.
- `TestProcessEventsDiscardsHandlerResult` — `processEvents` binds that result to nothing (`go handleAsync(...)` in the default branch, `_ = handleAsync(...)` in the in-order branch), so no guard on `UpdateCheckpoint` can reference it. Asserted over the AST rather than behaviourally, because `processEvents` takes a concrete `*azeventhubs.ProcessorPartitionClient` with no seam to inject a fake through; it should be replaced with a behavioural test once the fix introduces one.

### Side note

The component logs `"...Retrying..."` on a `RETRY`, which is misleading: with no Resiliency CR there is no inbound retry at all. Worth a separate look.

## Proposed Fix

Make checkpointing conditional on acknowledgement, opt-in via subscription metadata to preserve today's behavior by default — e.g. `checkpointOnSuccess: true`.

## Release Note

RELEASE NOTE: **FIX** Azure Event Hubs pub/sub: advance the partition checkpoint only after the application acknowledges the message.

gargsajal9:
## Description

Fixes Azure Event Hubs checkpointing so a partition never advances past an unresolved event or batch in either delivery mode.

- retains `enableInOrderMessageDelivery: false` as the default and keeps the field ordering-only
- retries handler failures with configured backoff until success or subscription cancellation
- advances concurrent checkpoints only through the newest contiguous successful batch and serializes checkpoint writes per partition
- adds `maxConcurrentHandlers` (default `100`) to bound received-but-not-checkpoint-eligible work and pause receives at capacity
- starts both modes from the earliest retained event only when no stored checkpoint exists; stored checkpoints retain precedence
- retries partially failed bulk batches in full and preserves the documented at-least-once duplicate tradeoff
- avoids checkpointing after cancellation or observed ownership loss and surfaces checkpoint-store failures
- adds pub/sub and binding certification coverage for fail, sidecar restart, and redelivery

Azure Event Hubs SDK v1.2.1 exposes ownership loss only when `ReceiveEvents` returns the AMQP detach; it does not provide an independent ownership callback or token for `UpdateCheckpoint`. Once loss is observed, processing is canceled immediately and checkpoint writes are gated. A handler that succeeds before loss is observable may be duplicated by the new owner, but cannot cause an unresolved event to be skipped because checkpoints still require contiguous handler success. Strict lease fencing independent of receives would require SDK/checkpoint-store support.

This implements the direction requested by @cicoyle. Please confirm the proposed `maxConcurrentHandlers` default of `100` and the earliest-retained fallback for a consumer group with no checkpoint.

## Issue reference

Closes #4561

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
    * [x] Created the dapr/docs PR: https://github.com/dapr/docs/pull/5322

## Validation

- Go 1.26.5 focused metadata-tagged tests with race detection
- focused `go vet`
- `golangci-lint` v2.10.1: 0 issues
- component metadata schema, bundle generation, and full analyzer
- official Event Hubs emulator + Azurite: concurrent restart redelivery, contiguous watermark, bounded receive backpressure, ordered partial-bulk replay, and checkpoint assertions
- pub/sub and binding certification packages compile; Azure-backed certification requires repository credentials/workflows
- companion docs Hugo render (972 EN/5 ZH-HANS pages), tests, and `git diff --check`


bindings/azure/eventhubs/metadata.yaml
```diff
@@ -60,10 +60,15 @@ builtinAuthenticationProfiles:
       - name: enableInOrderMessageDelivery
         type: bool
         required: false
-        default: "false"
-        example: "false"
+        default: "true"
```
cicoyle: We cannot change the default behavior, as that would be a breaking change. Please leave the default as-is for both `metadata.yaml` files. This should be taken in as an option users can toggle, never hard coded to true.

bindings/azure/eventhubs/metadata.yaml
```diff
@@ -60,10 +60,15 @@ builtinAuthenticationProfiles:
...
+        example: "true"
         description: |
-          Enable in order processing of messages within a partition.
+          Enable in order processing of messages within a partition. When true,
+          checkpoints advance only after the input handler successfully processes an event.
+          If no checkpoint exists, consumption starts from the earliest retained event.
```
cicoyle: Can you revert the default flip in `metadata.go` and both `metadata.yaml` files? That field should only be about ordered processing.

In both modes the checkpoint must never advance past an event that hasn't succeeded yet, since new processors resume after it. Your ordered mode changes already do this. For `false` mode (handlers run concurrently), we should only checkpoint up to the newest event where everything before it has succeeded. ex: if 1, 2 and 4 are done but 3 is still failing, the checkpoint stays at 2. Checkpoint writes should also be serialized per partition so an out-of-order write can't move it backwards.

We should extract the retry logic into a handleWithRetry() helper used by both modes. It should retry with backoff until the handler returns nil or ctx is cancelled, and never give up. That, together with the gated checkpoint, is what gives us the at-least-once guarantee.

I think we will also want a new field called `maxConcurrentHandlers` [matching Service Bus](https://github.com/dapr/components-contrib/blob/main/bindings/azure/servicebusqueues/metadata.yaml#L80). Since a stuck event pins the checkpoint, without a cap we'd keep receiving and spawning goroutines forever, and replay all of it on restart. When the cap is hit we should stop calling ReceiveEvents until a slot frees up.

We also need thorough tests for this PR. Can you add to the [eventhub_tests.go](common/component/azure/eventhubs/eventhubs_test.go) unit test the non happy path, so out of order completion and lets ensure to test both `true` and `false` for this field: `enableInOrderMessageDelivery`

Can you also add to the certification test suite a fail, restart sidecar, redelivered test for both pubsub && bindings?

cicoyle (changes requested):
Thanks for this PR 🎉 

The ordered mode work is solid. I'd like to change the approach tho, so we fix the guarantee without changing the default, and so the `false` case is also at-least-once per our docs.

bindings/azure/eventhubs/metadata.yaml
gargsajal9: Addressed this in `9a8c31f7`: restored the `false` default, added shared retry-until-success/cancellation handling, serialized contiguous per-partition checkpoints for concurrent delivery, bounded outstanding work with `maxConcurrentHandlers`, and added ordered/concurrent unit coverage plus pub/sub and binding restart/redelivery certification tests. The emulator matrix also verifies a gap blocks both checkpoint advancement and further receives at capacity. I used `100` as the proposed finite default and earliest-retained startup only when no checkpoint exists; please confirm those two compatibility choices.

pubsub/azure/eventhubs/metadata.yaml
```diff
@@ -110,6 +122,16 @@ metadata:
...
     example: '"group1"'
+  - name: checkPointFrequencyPerPartition
+    type: number
+    required: false
+    default: "1"
+    description: |
+      Update the checkpoint for a partition starting with the first batch, then
+      every N contiguous successfully processed batches. Set to 0 to stop writing
+      checkpoints. This does not delete an existing checkpoint. If no checkpoint
+      exists, consumption starts from the earliest retained event.
```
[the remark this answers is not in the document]
gargsajal9: Fixed in `7cb41d36`: the description now explicitly identifies `checkPointFrequencyPerPartition` as subscription metadata and directs users to set it on the subscription rather than the component spec. Runtime behavior remains unchanged because both subscribe paths already read request metadata.

pubsub/azure/eventhubs/metadata.yaml
```diff
@@ -39,9 +39,21 @@ builtinAuthenticationProfiles:
...
+        type: number
+        required: false
+        default: "100"
+        example: "10"
         description: |
-          Enable in order processing of messages within a partition.
+          Maximum number of batches per partition that can be received but are not yet
+          covered by the contiguous success checkpoint. This also limits concurrent
+          handler calls and pauses receives at capacity. Applies when
+          `enableInOrderMessageDelivery` is false and must be greater than 0.
```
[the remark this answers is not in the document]
gargsajal9: Fixed in `7cb41d36`: moved `maxConcurrentHandlers` into authentication-independent top-level pub/sub metadata so it is available to both connection-string and Azure AD configurations.

bindings/azure/eventhubs/metadata.yaml
```diff
@@ -61,9 +61,21 @@ builtinAuthenticationProfiles:
...
+        type: number
+        required: false
+        default: "100"
+        example: "10"
         description: |
-          Enable in order processing of messages within a partition.
+          Maximum number of events per partition that can be received but are not yet
+          covered by the contiguous success checkpoint. This also limits concurrent
+          input handler calls and pauses receives at capacity. Applies when
+          `enableInOrderMessageDelivery` is false and must be greater than 0.
```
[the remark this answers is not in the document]
gargsajal9: Fixed in `7cb41d36`: moved `maxConcurrentHandlers` from the Azure AD profile into top-level binding metadata and marked it input-only, so connection-string bindings expose and schema-validate the same receive limit.

common/component/azure/eventhubs/eventhubs.go
```diff
@@ -397,20 +513,45 @@ func (aeh *AzureEventHubs) processEvents(subscribeCtx context.Context, partition
...
-				go aeh.handleAsync(subscribeCtx, config.Topic, events, config.Handler) //nolint:errcheck // legacy behavior preserved
+				index := tracker.add(events)
+				handlers.Add(1)
+				go func(events []*azeventhubs.ReceivedEventData, index uint64) {
+					defer handlers.Done()
+
+					if handleErr := aeh.handleWithRetry(processCtx, config.Topic, events, config.Handler); handleErr != nil {
+						return
+					}
+					advanced, checkpointErr := tracker.complete(processCtx, index, partitionClient)
```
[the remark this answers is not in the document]
gargsajal9: Addressed the actionable race in `86810eeb`: when `ReceiveEvents` reports ownership loss, processing is now canceled immediately through a checkpoint gate. Cancellation reaches an in-progress checkpoint write, while queued or later writes are rejected before calling `UpdateCheckpoint`. A deterministic concurrent test holds a handler through ownership loss and verifies that its later success cannot checkpoint; the focused test passes 100 runs and under `-race`.

The SDK has no independent ownership-loss callback or ownership token for `UpdateCheckpoint`; it exposes loss only as an AMQP detach returned by `ReceiveEvents`. While no receive is active, there is therefore no earlier lifecycle signal to wire to. This does not weaken the at-least-once invariant: every checkpoint still requires contiguous handler success, so a handler that succeeds before loss is observed has processed the event; a new owner can only duplicate that work. Strict lease fencing independent of receives would require an SDK/checkpoint-store API change rather than a component-side context check.

cicoyle:
lmk when to take another look. Thanks for the efforts here 🙏🏻

gargsajal9:
> lmk when to take another look. Thanks for the efforts here 🙏🏻

Hey @cicoyle  its ready to review now :)

common/component/azure/eventhubs/eventhubs.go
```diff
@@ -368,26 +481,60 @@ func (aeh *AzureEventHubs) processEvents(subscribeCtx context.Context, partition
...
+				if checkpointErr := checkpointError(); checkpointErr != nil {
+					return checkpointErr
+				}
+			}
 			// If we get an error like ErrorCodeOwnershipLost, it means that the partition was rebalanced and we lost it
 			// We'll just stop this subscription and return
 			eventHubError := (*azeventhubs.Error)(nil)
 			if errors.As(err, &eventHubError) && eventHubError.Code == azeventhubs.ErrorCodeOwnershipLost {
 				aeh.logger.Debugf("Client lost ownership of partition %s for topic %s", partitionClient.PartitionID(), config.Topic)
+				checkpointGate.loseOwnership(processCancel)
```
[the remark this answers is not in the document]
gargsajal9: Fixed in `fdb9b5f3`: the receive-error path now classifies ownership loss and calls `checkpointGate.loseOwnership(processCancel)` before releasing the receive slot or inspecting checkpoint errors. This closes the scheduling window where a completing handler could start a checkpoint after the SDK had already returned ownership loss but before the gate was marked. The focused ownership/checkpoint tests pass 500 runs, plus 10 full package race runs, vet, and lint.

bindings/azure/eventhubs/metadata.yaml
```diff
@@ -98,6 +91,31 @@ builtinAuthenticationProfiles:
           entity management is enabled.
 metadata:
   # Input-only metadata
+  - name: enableInOrderMessageDelivery
+    type: bool
+    required: false
+    default: "false"
+    example: "true"
```
cicoyle:
```suggestion
    example: "false"
```

We might wanna leave the example bool alone.

pubsub/azure/eventhubs/metadata.yaml
```diff
@@ -74,6 +66,27 @@ builtinAuthenticationProfiles:
           Number of partitions for the new Event Hub namespace. Used only when
           entity management is enabled.
 metadata:
+  - name: enableInOrderMessageDelivery
+    type: bool
+    required: false
+    default: "false"
+    example: "true"
```
cicoyle: same here
```suggestion
    example: "false"
```

cicoyle (changes requested):
mostly lgtm, just a few more tests I would like to see covered:

- concurrent mode test w/ `CheckPointFrequencyPerPartition` set to 0
- concurrent mode test w/ `CheckPointFrequencyPerPartition` set to > 1 combined with out of order completion
- another retry test that fails more than 3 times

bindings/azure/eventhubs/metadata.yaml
gargsajal9: Updated in `68fdd22c`: the binding metadata example now remains `"false"`, matching the unchanged default.

pubsub/azure/eventhubs/metadata.yaml
gargsajal9: Updated in `68fdd22c`: the pub/sub metadata example now remains `"false"`, matching the unchanged default.

gargsajal9:
> mostly lgtm, just a few more tests I would like to see covered:
> 
> * concurrent mode test w/ `CheckPointFrequencyPerPartition` set to 0
> * concurrent mode test w/ `CheckPointFrequencyPerPartition` set to > 1 combined with out of order completion
> * another retry test that fails more than 3 times

@cicoyle  please have a look again

cicoyle (dismissed):

cicoyle:
Thank you @gargsajal9!

common/component/azure/eventhubs/eventhubs.go
```diff
@@ -357,9 +344,135 @@ func (aeh *AzureEventHubs) handleAsync(ctx context.Context, topic string, messag
...
+	}
+	return advanced, nil
+}
+
+func (aeh *AzureEventHubs) processEvents(subscribeCtx context.Context, partitionClient processorPartitionClient, config SubscribeConfig) error {
+	processCtx, processCancel := context.WithCancel(subscribeCtx)
+	var handlers sync.WaitGroup
 	defer func() {
+		processCancel()
+		handlers.Wait()
```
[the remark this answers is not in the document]
gargsajal9: Fixed in `5195c0bb`: shutdown now cancels processing and fences checkpoint writes, closes the partition client before draining handlers, and bounds the handler wait so an application handler that ignores cancellation cannot block client cleanup indefinitely. Added regressions for the bounded wait and for closing the client while a cancellation-ignoring handler remains blocked. The focused tests pass 100 runs, plus 10 full package race runs, vet, and lint.

gargsajal9:
@cicoyle need one more review :)

cicoyle (approved):

olitomlinson:
@cicoyle some context on why `enableInOrderMessageDelivery` wasnt flipped to `true` by default

https://github.com/dapr/components-contrib/issues/3568#issuecomment-2423073564

PR #4611, opened 2026-09-22
bump wazero to v1.12.0 — fixes wasm middleware crashes on amd64 with Go >= 1.24

LiBaoxuan-Neil:
## Problem

On amd64, a `middleware.http.wasm` guest that processes a `guestConfig` larger than the
2048-byte shared read buffer fails on every request with:

```
HTTP/1.1 500 Internal Server Error
wasm error: out of bounds memory access
wasm stack trace:
    ...
```

The failure is not deterministic across builds: it moves around with the guest's
compiler/allocator settings and only appears above a size threshold (~2 KB for the
config channel), which makes it look like a guest bug. It is not.

## Root cause

`dapr/dapr` and `dapr/components-contrib` (all 1.18.x) pin `github.com/tetratelabs/wazero v1.8.0`.

wazero v1.8.x's optimizing compiler backend (wazevo) emits a call to the Go runtime
`memmove` for bulk-memory instructions (`memory.copy` / `memory.fill`, i.e. what LLVM
generates for `memcpy` / `memset`). It only marked `xmm0`–`xmm15` as clobbered by that
call. **Go 1.24 changed `runtime.memmove` on amd64 to also use `RDX`**
(`src/runtime/memmove_amd64.s:123`), so a guest value that the register allocator kept
live in `RDX` across the internal call is destroyed; the guest then computes a garbage
address and traps on the next memory access.

Since dapr 1.18 requires Go >= 1.26.3, released sidecars are built with a Go version
that has this new `memmove`, so every dapr 1.18.x sidecar is affected.

Upstream:

* Reported against wazero as wazero#2375 — *"invalid table access only with Go 1.24 and amd64"*
  (the reporter hit it through go-re2, i.e. the same class of random traps).
* Fixed in **wazero v1.9.0** by wazero#2378 — *"wazevo(amd64): reserve dx when calling
  memmove to support go 1.24"*.

## Change

```diff
-	github.com/tetratelabs/wazero v1.8.0 // indirect
+	github.com/tetratelabs/wazero v1.12.0 // indirect
```

plus the two `v1.12.0` hashes in `go.sum` (the `v1.8.0/go.mod` line is kept, as it is
still needed for module graph resolution).

The minimum fixed version is **v1.9.0**; v1.12.0 is the current release and is what I
tested below. If you prefer the smallest possible change for a patch release, I'm happy
to retarget this PR to v1.9.0 — it also fixes the crash (verified).

## Verification

amd64, Go 1.26.8, `CGO_ENABLED=0`, dapr's default build tags (`allcomponents`), **the
same source tree** for both binaries — only the `go.mod` wazero version differs:

```
go mod edit -require=github.com/tetratelabs/wazero@<version>
CGO_ENABLED=0 go build -tags allcomponents -o daprd ./cmd/daprd
```

| daprd built from | wazero in binary (`go version -m`) | request hitting a 5.3 KB `guestConfig` |
| --- | --- | --- |
| dapr 1.18.4 | v1.8.0 | **HTTP 500** `wasm error: out of bounds memory access` |
| dapr 1.18.4 | v1.12.0 | **HTTP 200**, expected routing headers set by the guest |

Same guest artifact (a real TinyGo wasm plugin, 560 KB) and same 5331-byte `guestConfig`
in both runs; the only difference is the wazero version.

Additional checks:

* wazero-level bisect with a minimal host (no dapr involved): the same guest module fails
  on v1.8.0 / v1.8.1 / v1.8.2 and passes on v1.9.0+.
* Same guest under wazero's **interpreter** engine passes on every version — the bug is
  strictly in the amd64 compiler backend.
* `go test ./pkg/middleware/...` passes with the bump.


LiBaoxuan-Neil:
Hi @cicoyle ,

This is my first contribution to the dapr/components-contrib repo, and I noticed the workflow for PR #4611 is in Action required status—likely waiting for maintainer approval to run CI.

The change bumps wazero to v1.12.0. Could you please approve the workflow run when you have a moment? Thank you!

